PR #2445, opened 2023-03-23
Create simple SQL view assets

bendnorman:
# PR Overview
This PR contains:

- A SQL view called `denorm_plants_utils_ferc1` which performs the simple inner join `pudl.output.ferc1.plants_utils_ferc1()` does in pandas. `PudlTabl.pu_ferc1()` has been updated to pull the SQL view from the database instead of calling `plants_utils_ferc1()`.
- A helper function called`sql_asset_factory` which returns an asset that sends SQL code to the db to be executed. The SQL code is stored in `pudl.output.sql`.
- An example python output table called `denorm_utilities_eia860` is just a dagsterified version of `pudl.output.eia860.utilities_eia860()`.
- Metadata for `denorm_utilities_eia860` and `denorm_plants_utils_ferc1`.
- A new parameter for SQLiteIOManagers that allow you to exclude certain table schemas for being created in the database. I made this change because we want the IO manager to throw an error if we haven’t created the resource metadata for a view but we also don’t want it to create a table schema in the db for the view.
-  A notebook that documents the output table conversion process.

```[tasklist]
- [x] Update `PudlTabl.pu_ferc1()` to read the viewd
- [x] Update the view name to use `denorm_` prefix
- [x] Add tests for `sql_asset_factory`
- [x] Figure out how to handle metadata of views. We don't need to create table schemas for them but we do need to add them to the resources dictionary so `apply_pudl_dtypes()` or `enforce_schema()` knows how to convert the types.
- [x] The IO Manager doesn't need to know the types for create the view, it should still throw an error so people can't add a view without adding them to the resources.
- [ ] Update release notes
- [x] Decide if we want to reorganize the data dictionary page to make a distinction between the output and normalized tables
- [x] Add deprecation warning to pudl.output.eia860.utilities_eia860()
- [x] Update `PudlTabl.pu_eia860` to read the `denorm_utilities_eia860` table
- [x] Document process for how to convert and test output tables
```

# PR Checklist

- [ ] Merge the most recent version of the branch you are merging into (probably `dev`).
- [ ] All CI checks are passing. [Run tests locally to debug failures](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#running-tests-with-tox)
- [ ] Make sure you've included good docstrings.
- [ ] For major data coverage & analysis changes, [run data validation tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#data-validation)
- [ ] Include unit tests for new functions and classes.
[...]

PR #2459, opened 2023-03-28
Use enforce_schema() and read-chunking in PudlSQLiteIOManager.

zaneselvans:
# PR Overview

This PR changes the `SQLiteIOManager` to run `Resource.enforce_schema()` before writing to SQLite, and after reading from SQLite **IF** a resource matching the asset `table_name` is defined in the pudl `Package`.

It applies `enforce_schema` to chunks of the dataframe as it's read out of the DB so that in the case of large tables that contain categorical values, peak and final memory usage is reduced.

# PR Checklist


src/pudl/metadata/constants.py
```diff
@@ -30,7 +31,10 @@
...
     "boolean": sa.Boolean,
     "date": sa.Date,
-    "datetime": sa.DateTime,
+    # Ensure SQLite's string representation of datetime uses only whole seconds:
+    "datetime": SQLITE_DATETIME(
+        storage_format="%(year)04d-%(month)02d-%(day)02d %(hour)02d:%(minute)02d:%(second)02d"
+    ),
     "integer": sa.Integer,
     "number": sa.Float,
     "string": sa.Text,
```
zaneselvans, line 40: This lets the datetime pseudo-type check work, and removes artificial precision in the datetime fields.

I think that we have enough SQLite specific stuff elsewhere in the system that our `to_sql()` is functionally `to_sqlite()` and so having an SQLite specific dtype in here (rather than the SQLAlchemy generics) isn't any more constrained than we already are.

If we want to be able to write into other DBs I think we'll probably want dialect-specific schema generation methods, like `to_sqlite()`, `to_bigquery()` and `to_postgres()`

src/pudl/metadata/classes.py
```diff
@@ -711,10 +711,8 @@ def to_sql(  # noqa: C901
                 )
             elif self.type == "date":
                 checks.append(f"{name} IS DATE({name})")
-            # Need to ensure that the string representation of the datetime only
-            # includes whole seconds or this check will fail.
-            # elif self.type == "datetime":
-            #    checks.append(f"{name} IS DATETIME({name})")
+            elif self.type == "datetime":
+                checks.append(f"{name} IS DATETIME({name})")
```
zaneselvans, line 715: This works now that I'm providing a formatting string in the SQLite specific datetime type.

src/pudl/io_managers.py
```diff
@@ -349,13 +356,26 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
                 name.
         """
         table_name = self._get_table_name(context)
+        pkg = Package.from_resource_ids()
+        all_resources = [resource.name for resource in pkg.resources]
+        res = None
+        if table_name in all_resources:
+            res = pkg.get_resource(table_name)
+
```
zaneselvans, line 364: @bendnorman I'm still a little unclear on what our expectations are about tables that would get written to SQLite with this IOManager.

Without the check to see whether the table had a resource, some unit tests failed, since they just had a temporary table with no metadata associated with it.

Do we only expect to write tables to SQLite that have `Resource` classes defined? Or do we want to be able to write interim tables that are so stringently described?  If so, should we differentiate between those cases somehow in the IO Manager?

zaneselvans (commented):
I didn't make any changes to the `PudlTabl` class, but there's definitely overlap in the functionality between that and the `SQLiteIOManager` Is that just the way it is for now, until we deprecate `PudlTabl`? Would it make sense to change `PudlTabl` to grab the assets using Dagster `AssetKey`s? That seems like it would be a significant change in how things work.

src/pudl/io_managers.py
bendnorman: I think we should expect all tables we are loading from the database to have a Resource class defined so we can be confident we are working with the correct dtype information.

`self._get_sqlalchemy_table(table_name)` throws an error if the table doesn't exist in the metadata.

src/pudl/io_managers.py
```diff
@@ -349,13 +356,26 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
         with engine.connect() as con:
             try:
-                df = pd.read_sql_table(table_name, con)
+                dfs = []
+                for chunk_df in pd.read_sql_table(table_name, con, chunksize=100_000):
+                    if res:
+                        chunk_df = res.enforce_schema(chunk_df)
+                    else:
+                        chunk_df = pudl.metadata.fields.apply_pudl_dtypes(chunk_df)
+                    dfs.append(chunk_df)
```
bendnorman, line 377: If we decide load_input should fail if the table_name doesn't show up in our metadata, this call to `apply_pudl_dtypes` would not be required.

src/pudl/io_managers.py
```diff
@@ -276,6 +276,13 @@ def _handle_pandas_output(self, context: OutputContext, df: pd.DataFrame):
         table_name = self._get_table_name(context)
 
         sa_table = self._get_sqlalchemy_table(table_name)
+        pkg = Package.from_resource_ids()
+        all_resources = [resource.name for resource in pkg.resources]
+        if table_name in all_resources:
+            res = pkg.get_resource(table_name)
+            df = res.enforce_schema(df)
+        else:
+            df = pudl.metadata.fields.apply_pudl_dtypes(df)
```
bendnorman, line 285: These changes intersect the work/discussion happening in #2445. It feels duplicative pass in a sqlalchemy metadata object created from the Package class to then create a package class inside of the IO Manager again.

I think it makes the most sense to:
1. Create a package object with the desired etl groups and pass it into a SQLiteIOManager object. `self.pkg.enforce_schema()` can be called in `handle_output` and `load_input`.
2. Create a new metadata field or etl_group that allows us to filter out SQL view tables so we can track the datatypes but avoid creating schemas for them in the db.

src/pudl/io_managers.py
```diff
@@ -276,6 +276,13 @@ def _handle_pandas_output(self, context: OutputContext, df: pd.DataFrame):
...
 
         column_difference = set(sa_table.columns.keys()) - set(df.columns)
```
bendnorman, line 287: This line contains duplicates some logic in enforce_schema.

zaneselvans: Which line, the `column_difference`?  Should that be inside the `else:` then?

zaneselvans: Okay, now it seems like this should be here, since this IO Manager doesn't use enforce schema.

src/pudl/io_managers.py
```diff
@@ -434,10 +476,10 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
                 )
             if df.empty:
                 raise AssertionError(
-                    f"The {table_name} table is empty. Materialize "
-                    "the {table_name} asset so it is available in the database."
+                    f"The {table_name} table is empty. Materialize the {table_name} "
+                    "asset so it is available in the database."
                 )
-            return pudl.metadata.fields.apply_pudl_dtypes(df)
```
zaneselvans, line 440: No longer required since enforce_schema is called in the loop above.

test/unit/io_managers_test.py
```diff
@@ -167,3 +171,43 @@ def test_missing_schema_error(sqlite_io_manager_fixture):
...
+        "track",
+        md,
+        Column("trackid", Integer, primary_key=True),
+        Column("trackname", String(16), nullable=False),
+        Column("trackartist", Integer, ForeignKey("artist.artistid")),
+    )
+    manager = PudlSQLiteIOManager(base_dir=tmp_path, db_name="pudl")
+    # Override the default PUDL metadata with this test metadata, so that they
+    # don't match, and we can check what happens.
+    manager.md = md
```
zaneselvans, line 196: This seems janky. What's the right way to do this?

src/pudl/io_managers.py
```diff
@@ -409,6 +407,37 @@ def __init__(
...
+        try:
+            res = self.package.get_resource(table_name)
+        except ValueError:
+            raise ValueError(
+                f"{table_name} does not appear in pudl.metadata.resources. "
+                "Check for typos, or add the table to the metadata and recreate the "
+                f"PUDL SQlite database. It's also possible that {table_name} is one of "
+                "the tables that does not get loaded into the PUDL SQLite DB because "
+                "it's a work in progress or is distributed in Apache Parquet format."
+            )
```
bendnorman, line 424: This check duplicates the `RuntimeError` raised in `sa_table = self._get_sqlalchemy_table(table_name)`. When we write dataframes to the database we need to make sure they exist in `self.md`. When we load a data frame from the database we want to make sure the table exists in has corresponding metadata in `self.package`.

src/pudl/io_managers.py
```diff
@@ -417,15 +446,28 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
                 name.
         """
         table_name = self._get_table_name(context)
-
         # Check if the table_name exists in the self.md object
         _ = self._get_sqlalchemy_table(table_name)
```
bendnorman, line 450: We shouldn't call `_ = self._get_sqlalchemy_table(table_name)` here because we'll be loading sql views as dataframes which won't have `Table()` objects in `self.md`.

test/unit/io_managers_test.py
bendnorman: I think the correct way to test this is to create a fresh Package object with test data.

test/unit/io_managers_test.py
```diff
@@ -167,3 +171,43 @@ def test_missing_schema_error(sqlite_io_manager_fixture):
...
+    return manager
+
+
+def test_missing_pudl_resource_error(pudl_sqlite_io_manager_fixture):
+    """Test that an error is raised when we try to write a non-existent table.
+
+    This is a bit contrived, as we have to somehow end up with a table that *does*
+    appear in the metadata object, but does *not* appear in the PUDL database schema,
+    which should be impossible, since the schema is generated from the PUDL Package,
+    unless we pass in a different Package. Maybe that's what we should be doing here?
```
bendnorman, line 206: Yeah this feels like an unlikely scenario.

bendnorman (commented):
Ah yay, the dtypes enforcement and `self.package` checks feel much better. I left a couple of comments about the tests and which checks are necessary.

src/pudl/io_managers.py
zaneselvans: Ohhh. Okay. If we reflected & inspected the DB, then we'd find them in **that** metadata, but that information about the table exists only the DB and in the Package, and not in the metadata generated by the Package (since we're creating those tables directly via CREATE VIEW)

bendnorman: Correct, though I'm unsure if `sa.Metadata().reflect()` would create Table objects for views. Worth experimenting.

src/pudl/io_managers.py
zaneselvans: Okay, so you're saying that we'll never see this error, because the thing it's checking has already been checked by attempting to get `sa_table`?

`self.package` is the more expansive definition of the data, right? Every table that appears in `self.md` will definitely appear in `self.package`, but the reverse isn't true -- there can be Resources (like `hourly_emissions_epacems` that appear in `self.package` but not `self.md` So `_get_sqlalchemy_table()` is the more restrictive test?

I feel like we might need to add a Venn diagram to the documentation explaining the difference between:

* Resources that appear in the Package (includes DB tables and non-SQL resources like EPA CEMS... as well as views?)
* Tables that appear in the SQLAlchemy metadata generated by the Package (only includes DB tables we write to directly)
* Tables that appear in the database itself (All tables we write to directly, plus all database views)

Do I understand that correctly?

bendnorman: Correct! Everything that appears in `self.md` will appear in `self.package` but not everything in `self.package` will appear in `self.md` (views, and things we don't want to the db like cems). I agree, we should clarify the difference between the metadata in `self.md` and `self.package` in the `PudlSQLiteIOManager` doc string.

bendnorman (approved):

PR #2466, opened 2023-03-28
Create `PudlSQLiteIOManager` to accept a `Package` object

bendnorman:
# PR Overview
This PR solves two problems we were running into in #2459 and #2445:
1. We need to keep track of SQL view datatypes in `pudl.metadata.resources` so we can read the views into dataframes and convert the columns to correct datatypes but we also don't want to create table schemas in the database for the views. My initial solution was to add a janky feeling `exclude_tables: list` init parameter to `SQLiteIOManager` which excludes a list of tables from being created in the database.
2. #2459 uses `Resource.enforce_schema()` in the `SQLiteIOManager` instead of the generic `pudl.metadata.fields.apply_pudl_dtypes()` function. The initial solution was to create a Package object in the io manager that loads all of the default PUDL resources. It felt strange to create a `sa.Metadata` object by filtering out certain etl groups in `pudl_sqlite_io_manager` to then create another `Package` object with the default Resources in the io manager.

This PR does two things to solves these problems:
1. Adds the `include_in_database` attribute to the `Resource` class. The `Package.to_sql()` method now only creates `sa.Table` objects for resources when `include_in_database == True`. This prevents us from creating table schemas for views and tables that are broken and we don't want to be included on datasette.
2. Creates a subclass of `SQLiteIOManager` called `PudlSQLiteIOManger` that accepts a `Package` object instead of a `sa.Metadata` object. This way we don't have to initialize a Package object with all of the default resources inside the io manager. Also, the `SQLiteIOManager` can stay de-coupled from pudl specific classes and metadata.

# PR Checklist


src/pudl/io_managers.py
```diff
@@ -341,6 +341,69 @@ def handle_output(self, context: OutputContext, obj: pd.DataFrame | str):
...
+                connection opens a transaction to modify the database, it will be locked
+                until that transaction is committed.
+        """
+        if not package:
+            package = Package.from_resource_ids()
+        self.package = package
+        md = self.package.to_sql()
+        super().__init__(base_dir, db_name, md, timeout)
+
     def load_input(self, context: InputContext) -> pd.DataFrame:
```
bendnorman, line 407: In #2459 @zaneselvans you'll probably rewrite the `PudlSQLiteIOManager.load_input()` method to:
1. Make sure the `table_name` exists as a resource in `self.package`.
2. Apply `enforce_schema()` to the df instead of `apply_pudl_dtypes()`.

You'll also need to overwrite the `_handle_pandas_output()` method to check the table exists in self.md and enforce schema.

src/pudl/io_managers.py
```diff
@@ -341,6 +341,69 @@ def handle_output(self, context: OutputContext, obj: pd.DataFrame | str):
...
+                    "the {table_name} asset so it is available in the database."
+                )
+            return df
+
+
+class PudlSQLiteIOManager(SQLiteIOManager):
+    """IO Manager that writes and retrieves dataframes from a SQLite database.
+
+    This class extends the SQLiteIOManager class to manage database metadata and dtypes
+    using the :func:`pudl.metadata.classes.Package` class.
```
zaneselvans, line 377: I think this needs to be `:class:` rather than `:func:` but I could be wrong.

src/pudl/io_managers.py
```diff
@@ -341,6 +341,69 @@ def handle_output(self, context: OutputContext, obj: pd.DataFrame | str):
...
 
+    def load_input(self, context: InputContext) -> pd.DataFrame:
+        """Load a dataframe from a sqlite database.
+
+        Args:
+            context: dagster keyword that provides access output information like asset
+                name.
+        """
+        table_name = self._get_table_name(context)
+        _ = self._get_sqlalchemy_table(table_name)
```
zaneselvans, line 352: Could you add a comment indicating whether this assignment to `_` is mutating something in the class or just checking whether an exception is raised? (I think the latter, right?)

src/pudl/io_managers.py
```diff
@@ -341,6 +341,69 @@ def handle_output(self, context: OutputContext, obj: pd.DataFrame | str):
...
+        package: Package = None,
+        timeout: float = 1_000.0,
+    ):
+        """Initialize PudlSQLiteIOManager.
+
+        Args:
+            base_dir: base directory where all the step outputs which use this object
+                manager will be stored in.
+            db_name: the name of sqlite database.
+            package: Tabular datapackage used to create sqlalchemy metadata and check
```
zaneselvans, line 393: It's not really a tabular datapackage -- our `Package` class echoes a lot of the parts of that standard but it doesn't actually implement it or allow us to output them / generate datapackage objects.

zaneselvans (approved):
This seems like a much cleaner solution overall!

Very minor docs / readability stuff. No substantive changes required so I'll approve (but if you want to make those little clarifications that would be great)

src/pudl/io_managers.py
zaneselvans: In `PudlSQLiteIOManager._handle_pandas_output()` does it make sense to use `enforce_schema()`?

If it calls the parent `_handle_pandas_output()` then `apply_pudl_dtypes()` will be called on the dataframe, and I think that will remove the categorical type information, and also any data source specific typing information (since the generic call to `apply_pudl_dtypes()` without `group="eia"` or something similar just uses the generic field definitions.  I guess maybe that means we can't call `super()._handle_pandas_output()` and we have to override the whole method, and should use `enforce_schema()` to get the data source specific types, even if the categorical information is lost when we write to SQLite.

PR #2491, opened 2023-04-04
Flesh-out the Sub Plant ID

cmgosnell:
# PR Overview

Integrating some of @grgmiller 's augmentation of the sub-plant id in the CEMS-EIA crosswalk. This is mostly a copy and paste. Plus some massaging to work over here in this context. Many remaining questions:

 - [x] what is time? I assumed intuitively that this should be a time-varied id... but what actually is varied in here? The `unit_id_pudl` does no vary over time. The `emissions_unit_id_epa` doesn't seem to vary over time. Maybe this isn't a time varying id 🤷🏻 
   -   It seems the `unit_id_pudl` definitely does not vary over time:
```
from pudl.etl import  defs
asset_key = "boiler_generator_assn_eia860"
bga = defs.load_asset_value(AssetKey(asset_key))
bga.groupby(["plant_id_eia", "generator_id", "boiler_id"])[["unit_id_pudl"]].nunique().unit_id_pudl.unique()

>> array([1])
```
   I think this makes sense and should probably not be altered. i assume it would be *extremely weird* if a physical unit was broken apart.
   
 - [x] Is this table an augmentation of the `epacamd_eia` table or a new table that looses information from `epacamd_eia`? i.e. should we have two db tables or is this new augmented table with the subplant id enough?
   -  these tables are indeed different. No boiler ID ect. ect.
 - [x] I thiiiink we should remove `add_operating_and_retirement_dates`. That seems like an OGE specific thing.
 - [x] what is dagster? I thiiiink this should be a second step to the main transform for the crosswalk. Which is to say `pudl.glue.epacamd_eia.transform` should result in an asset (on df not a {"name": df}) which is an input into this new asset...  but what I am the most ❓-ing about is how to handle cems itself. Right now I added `unique_plant_ids_epacems` which just grabs it from the parquet files so i could play with the code itself, but obviously that is not the ideal way. (it may be best for me to wait until Zach's CEMS edits have been merged?)
 - [x] Where should this live?
    - If this is glue.... is the stuff in `pudl.analysis.epacamd_eia` also live in glue? YES
    - The dagster assets act like import dependencies, so for now they are all in `pudl.etl.epacems`. I think I should also migrate the analysis networkx work into the cems module as well.
 - [x] I also think we should remove the `filter_crosswalk` step. I believe this will enable these subplant id's to be generated for all EIA records. 
    -  Making the table without the filter step works fine. it takes an extra 30 seconds w/ all the data, but that seems worthwhile
-  [x] How/where to handle the two manual ID correction steps?
   - @grgmiller : any advice on the `manually_update_subplant_id` function? The note in there indicates something that we could fix upstream? Update: For now we are calling `manually_update_subplant_id` in `pudl.etl.epacems.clean_epacamd_eia`. So I removed it from `subplant_ids_epacamd_eia`
 - [x] How do I tell dagster to actually load assets into the db?? @bendnorman ? 
 - [x] High level where this should live? what is glue?

### OOS
 - [ ] unit_id_pudl fleshing out
 - [ ] gregs manual update to the crosswalk
 - [ ] SPEEEEEED. Can we make `update_subplant_ids` faster? right now its an apply function and takes the longest to run (~2 minutes) outside of the cems id grabbing step (~5 minutes 👀).

# PR Checklist


src/pudl/glue/epacamd_eia.py
```diff
@@ -138,5 +143,400 @@ def transform(
...
     return {"epacamd_eia": crosswalk_clean}
+
+
+##################
+# Add Sub-Plant ID
+##################
+
+
+# @asset
+def identify_subplants_epacamd_eia(
```
bendnorman, line 156: If `identify_subplants_epacamd_eia` is a table we'd like to persist in the database I'd recommend turning it into an asset. If it just a cleaning function that augments to the `epacamd_eia` asset, I'd keep it as a python function.

Where is `identify_subplants_epacamd_eia` being used?

src/pudl/glue/epacamd_eia.py
```diff
@@ -138,5 +143,400 @@ def transform(
...
+        epacems_path,
+        use_nullable_dtypes=True,
+        engine="pyarrow",
+        index=False,
+        split_row_groups=True,
+        filters=year_state_filter(years, states),
+        columns=annual_id_cols,
+    )
+    epacems_ids = (
+        epacems_dd.compute().drop_duplicates().rename(columns={"year": "report_year"})
```
bendnorman, line 205: Instead of reading from the epacems path, you can depend on the cems asset, which will load the cems data as a dask dataframe using the `PandasParquetIOManager.load_input()`. However, to depend on cems asset, the function you're working on must also be an asset. Where does this work fit into the DAG? Is it an analysis table or a part of the ETL glue?

src/pudl/glue/epacamd_eia.py
cmgosnell: its not being used rn! I was planning on figuring out the dag of it all before slotting it in.  I was thinking we could have a clean_epacamd_eia asset (result of `pudl.glue.epacamd_eia.transform`) which feeds into this second stage of `epacamd_eia` asset. kind of in line with the clean_{table}_eia860 -> {table}_eia860 model? but also we could just append=sin this to the cleaning steps within `pudl.glue.epacamd_eia.transform`.

cmgosnell: hm! no i think we need to make this in two steps. bc CEMS relies on this epacamd crosswalk and now this sub-plant-id relies on both the cleaned crosswalk and CEMS.

src/pudl/glue/epacamd_eia.py
cmgosnell: Can it be a function that is called within an asset? Although I suppose I could make this little guy an asset on its own.

cmgosnell: also @bendnorman ... how would i actually call `PandasParquetIOManager.load_input()`? is there a good example of how I can pass around `context`?

src/pudl/glue/epacamd_eia.py
bendnorman: Correct! We can't have any circular dependencies. I think we just need to create convert `identify_subplants_epacamd_eia` to be a dagster asset that depends on `hourly_emissions_epacems`, `epacamd_eia`, and `generators_eia860`. 

To depend on `hourly_emissions_epacems` you'll need to specify [the io manager](https://docs.dagster.io/concepts/io-management/io-managers#asset-input-io-managers):

```python
@asset(
    ins={
        "first_asset": AssetIn(input_manager_key="epacems_io_manager"),
        # I'm not sure if we'll have to specify the io managers for the other upstream dependencies
    }
)
def identify_subplants_epacamd_eia(
    epacamd_eia: pd.DataFrame,
    generators_eia860: pd.DataFrame,
    hourly_emissions_epacems: dd.DataFrame # note this will be a dask dataframe
) -> pd.DataFrame:
  ...
```

src/pudl/glue/epacamd_eia.py
bendnorman: Sure! It can be it's own function called in an asset. The easiest way to load assets using io managers is to just specify the asset name in the new asset's signature:

```python
@asset(io_manager_key="pudl_sqlite_io_manager")
def a():
   ...

@asset(io_manager_key="pudl_sqlite_io_manager")
def b(a): # will load asset "a" using the pudl_sqlite_io_manager.
   ...
```

test/unit/transform/epacems_test.py
```diff
@@ -37,3 +41,85 @@ def test_harmonize_eia_epa_orispl():
...
+        StringIO(
+            """
+plant_id_eia,plant_id_epa,unit_id_pudl,emissions_unit_id_epa,generator_id,subplant_id
+1391,1391,1,1A,1A,0
+1391,1391,1,2A,2A,0
+1391,1391,1,3A,3A,0
+1391,1391,,4A,4A,1
+1391,1391,,5A,5A,2
+1,1,1,A,A,0
+"""
```
cmgosnell, line 111: hey @grgmiller @aesharpe ... when writing test for this subplant id , austen suggested I use the actual plant_id_eia 1391 as an example. But you both probably know better than I about what the desired behavior for this plant? I saw that greg you were manually assigning all of these as 0. Should we keep this manual assign? or is this behavior outlined in this little unit test expected?

src/pudl/etl/epacems_assets.py
```diff
@@ -148,8 +149,438 @@ def hourly_emissions_epacems(
...
+    # 1000 was chosen as an arbitrarily high number, since the largest unit_id_pudl is ~ 10.
+    subplant_crosswalk["numeric_generator_id"] = (
+        subplant_crosswalk["numeric_generator_id"] + 1000
+    )
+    # fill any missing unit_id_pudl with a number for each unique generator
+    subplant_crosswalk["unit_id_pudl_filled"] = (
+        subplant_crosswalk["unit_id_pudl_connected"]
+        .fillna(subplant_crosswalk["subplant_id_connected"] + 100)
+        .fillna(subplant_crosswalk["numeric_generator_id"])
+    )
```
bendnorman, line 535: I don't fully understand why `numeric_generator_id` could overlap with existing `unit_id` but I feel like there might be a more robust way to ensure they don't overlap. Does this function have access to the `unit_id` values? Could we create `numeric_generator_id`s that don't exist in `unit_id` without just adding an arbitrary number?

src/pudl/etl/epacems_assets.py
```diff
@@ -148,8 +149,438 @@ def hourly_emissions_epacems(
...
+
+def connect_ids(df, id_to_update, connecting_id):
+    """Corrects an id value if it is connected by an id value in another column.
+
+    If multiple subplant_id are connected by a single unit_id_pudl, this groups these
+    subplant_id together. If multiple unit_id_pudl are connected by a single subplant_id,
+    this groups these unit_id_pudl together.
+
+    Args:
+        df: dataframe containing columns with id_to_update and connecting_id columns
```
bendnorman, line 553: Might make the function purpose clearer if we rename `df` to something less generic.

src/pudl/etl/glue_assets.py
```diff
@@ -44,22 +46,22 @@ def create_glue_tables(
...
+    #         == dataset_settings.eia.eia860.data_source.working_partitions["years"]
+    #     )
+    #     glue_raw_dfs = pudl.glue.epacamd_eia.extract(ds)
+    #     glue_transformed_dfs = pudl.glue.epacamd_eia.transform(
+    #         glue_raw_dfs,
+    #         generators_entity_eia,
+    #         boilers_entity_eia,
+    #         processing_all_eia_years,
+    #     )
+    #     glue_dfs.update(glue_transformed_dfs)
```
bendnorman, line 64: Should this be deleted?

src/pudl/output/epacems.py
```diff
@@ -148,5 +148,5 @@ def epacems(
             years=epacems_settings.years,
         ),
     )
-
+    # TODO: correct_epa_eia_plant_id_mapping!
```
bendnorman, line 151: Could you elaborate on this TODO, please? Should we create an issue?

bendnorman (changes requested):
Looks good! Just a couple of minor requests.

src/pudl/etl/epacems_assets.py
```diff
@@ -148,8 +149,438 @@ def hourly_emissions_epacems(
...
+    values from various years. To keep the crosswalk in alignment with the available eia
+    data, we'll restrict it based on the generator entity table which has
+    ``plant_id_eia`` and ``generator_id`` so long as it's not using the full suite of
+    avilable years. If it is, we don't want to restrict the crosswalk so we can get
+    warnings and errors from any foreign key discrepancies. This isn't an ideal
+    solution, but it works for now.
+
+    Args:
+        context: dagster keyword that provides access to resources and config. This
+            determines whether
```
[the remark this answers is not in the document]
cmgosnell: doh!

src/pudl/etl/epacems_assets.py
```diff
@@ -148,8 +149,438 @@ def hourly_emissions_epacems(
...
+        dataframe with unique set of: "plant_id_eia", "report_year" and
+        "emissions_unit_id_epa"
+    """
+    epacems_ids = (
+        hourly_emissions_epacems[["plant_id_eia", "year", "emissions_unit_id_epa"]]
+        .drop_duplicates()
+        .compute()
+    )
+
+    return epacems_ids
```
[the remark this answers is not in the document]
cmgosnell: Oh I think this TODO is old! I should have scrubbed through them a little better b4 asking for another review. but this asset takes ~1.5 minute to make now with all of the cems data loaded which I think is pretty dang good considering how big the input is.

src/pudl/etl/epacems_assets.py
```diff
@@ -148,8 +149,438 @@ def hourly_emissions_epacems(
...
+            on=["plant_id_eia", "generator_id"],
+            how="inner",
+        )
+        crosswalk_clean = pd.merge(
+            crosswalk_clean,
+            boilers_entity_eia[["plant_id_eia", "boiler_id"]],
+            on=["plant_id_eia", "boiler_id"],
+            how="inner",
+        )
+    # TODO: Add manual crosswalk
```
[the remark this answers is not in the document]
cmgosnell: I wanted to spin it off, yes! I should make another issue for this and remove this line

src/pudl/etl/glue_assets.py
```diff
@@ -18,11 +17,14 @@
     outs={
         table_name: AssetOut(io_manager_key="pudl_sqlite_io_manager")
         for table_name in Package.get_etl_group_tables("glue")
+        # there may be a better way to restructure how we are making the glue post
+        # converting epacamd_eia -> an asset
+        if table_name != "epacamd_eia"
```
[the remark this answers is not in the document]
cmgosnell: yeah... I'd love your option about organization here. rn this new table and the old crosswalk is labeled as glue tables, but the code now all live over in the `etl.epacems_assets` module. but the metadata live in glue. Should I just reclassify all of these tables as cems tables? I feel swirly about it so I left it as is before getting more feedback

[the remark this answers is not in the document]
bendnorman: It still seems like glue. I forget, was there an issue with creating these new assets in the `etl.glue_assets` module? You'll still have to filter out `epacamd_eia` in this line but will keep the metadata and assets in glue world.

src/pudl/etl/epacems_assets.py
```diff
@@ -148,8 +149,438 @@ def hourly_emissions_epacems(
...
+    emissions_unit_ids_epacems: pd.DataFrame,
+    boiler_generator_assn_eia860: pd.DataFrame,
+) -> pd.DataFrame:
+    """Groups units and generators into unique subplant groups.
+
+    This function consists of three primary parts:
+
+    #.  Augement the EPA CADM:EIA crosswalk with all IDs from
+    #.  Use graph analysis to identify distinct groupings of EPA units and EIA
+        generators based on 1:1, 1:m, m:1, or m:m relationships.
```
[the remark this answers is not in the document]
cmgosnell: doh again! yes i'll add the third part

test/unit/transform/epacems_test.py
grgmiller: I had been manually assigning all generators at 1391 to the same subplant due to the associations in the BGA table. So I think that the expected behavior is that all subplant IDs are the same, but I think that this won't be the case until the BGA table is incorporated into the subplant ID code.

src/pudl/glue/epacamd_eia.py
```diff
@@ -7,136 +7,8 @@
...
-from pudl.metadata.fields import apply_pudl_dtypes
-from pudl.workspace.datastore import Datastore
 
 logger = pudl.logging_helpers.get_logger(__name__)
-
-
-def extract(ds: Datastore) -> pd.DataFrame:
-    """Extract the EPACAMD-EIA Crosswalk from the Datastore."""
-    logger.info("Extracting the EPACAMD-EIA crosswalk from Zenodo")
-    with ds.get_zipfile_resource("epacamd_eia", name="epacamd_eia.zip").open(
```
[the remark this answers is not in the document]
cmgosnell: oh lol this module is now empty i should delete it. I was toggling back and forth about where the code should live vs where the assets should be generated and wanted to keep this module around until that was settled.

src/pudl/etl/epacems_assets.py
```diff
@@ -148,8 +149,438 @@ def hourly_emissions_epacems(
...
+        ],
+        keep="last",
+    )
+    return subplant_crosswalk_complete
+
+
+def augement_crosswalk_with_eia860(
+    crosswalk_clean: pd.DataFrame, generators_eia860: pd.DataFrame
+) -> pd.DataFrame:
+    """Merge any plants that are missing from the EPA crosswalk but appear in EIA-860.
```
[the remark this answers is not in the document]
cmgosnell: Good eye! But from the docs in `clean_epacamd_eia` is looks like we only do this for the sake of the fast CI tests! It looks like the filtering only happens then because of FK constraints.

src/pudl/etl/epacems_assets.py
```diff
@@ -148,8 +149,438 @@ def hourly_emissions_epacems(
...
+        boiler_generator_assn_eia860[
+            ["plant_id_eia", "generator_id", "unit_id_pudl"]
+        ].drop_duplicates(),
+        how="outer",
+        on=["plant_id_eia", "generator_id"],
+        validate="m:1",
+    )
+
+
+def update_subplant_ids(subplant_crosswalk):
```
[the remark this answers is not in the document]
cmgosnell: yes good point! I'm honestly not super familiar with making pd.DataFrame.groupby().apply functions. From quick research, I believe the input is indeed a pd.DataFrame but maybe you or @aesharpe will know better.

src/pudl/etl/epacems_assets.py
cmgosnell: yeah this stuff was a direct copy/paste from greg that I didn't augment much or spend a lot of time fully understanding it tbqh. But I think we can change the +1000 to be `max(subplant_crosswalk["numeric_generator_id"]) + 1`. i'll double check this...

cmgosnell: Actually I think this is the way

```
numeric_generator_id_gb = subplant_crosswalk.groupby(
        ["plant_id_eia", "generator_id"], dropna=False
    )
    subplant_crosswalk["numeric_generator_id"] = (
        numeric_generator_id_gb.ngroup() + numeric_generator_id_gb["unit_id_pudl"].max()
    )
```

test/unit/transform/epacems_test.py
grgmiller: I think in this specific case, we actually assigned 4A and 5A to subplant 0 due to a reason not reflected in the BGA table. We noticed that the minimum generation reported in CEMS for subplant 0 was higher than the nameplate capacity for that subplant, which suggested that there was something strange with the subplant mapping (see https://github.com/singularity-energy/open-grid-emissions/pull/142#issuecomment-1186579260). However, we should probably remove this manual update in the pudl code since it is not verified. I think the mapping in this unit test is probably correct from the standpoint of what is reported in the EPA crosswalk and BGA table, although this mapping leads to strange results in the actual data (e.g. the generation for subplant 0 is greater than the nameplate capacity of subplant 0, and no generation is reported for subplant 1 or 2). I think this suggests an issue with the underlying data potentially.

src/pudl/etl/glue_assets.py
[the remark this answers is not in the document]
cmgosnell: There are I think three distinctions here: where the code lives, what the dagster asset group is and what the metadata/etl-group is. these all feels swirly to me rn.

* rn code lives: in `etl.epacems_assets` bc the `clean_epacamd_eia` asset is required for `hourly_emissions_epacems` THEN both `clean_epacamd_eia` and `hourly_emissions_epacems` are required to make this `epacamd_eia_subplant_ids` asset. bc the assets are dependencies keeping all of these epacamd_eia assets in a gluey place caused circular dependencies. I thiiiiink I could break it out into three modules instead of the primary attempted 2 and current 1:
  * a clean_epacamd_eia place (should this be `etl.glue_assets` or `pudl.glue.epacamd_eia`?)
  * epacems_assets
  * a epacamd_eia_subplant_ids place (ditto to `etl` vs `glue` directory question
* dagster asset group: rn the glue assets are being derived from the etl-group and the epacamd assets are being derived from the assets in the `epacems_assets` module I believe.
* metadata/etl-group: rn I left these epacamd_eia tables in the glue etl group. bc they do feel like glue to me!

lol this is a lot and I'd love your perspective on it!! haaalp

src/pudl/output/epacems.py
cmgosnell: ah no i moved this into the cems processing itself so this should be deleted

test/unit/transform/epacems_test.py
cmgosnell: okay i added this manually assign everything to 0 step back in. I'm curious if you have a sense of if this is an upstream cems problem or something else. if you know! i know tracking these little guys down to really understand (plus keeping it in your head/in docs) can be challenging.

src/pudl/etl/epacems_assets.py
cmgosnell: these columns are all interim values that are used to update the subplant_id but not actually published. I'll chew on helpful defensive checks to add in here though.

src/pudl/etl/glue_assets.py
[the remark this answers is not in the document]
bendnorman: Yes, those are the three distinctions we need to work out generally. Hopefully, we clarify these distinctions when we tackle #2517. 

I feel like these are the ideal categorizations:
- `clean_epacamd_eia`, `epacamd_eia_subplant_ids `
   - etl_group: glue
   - asset_group: glue
   - code location: `etl.glue_assets`
- `hourly_emissions_epacems`
   - etl_group: epacems
   - asset_group: epacems
   - code location: `etl.epacems_assets`

I'll spend 15 - 20 min moving these around. If I can't figure it out, maybe we just keep everything in `etl.epacems_assets` and the epacems etl_group? Not ideal but I expect the code structure and etl groups to change in the next few sprints.

src/pudl/etl/glue_assets.py
bendnorman: Also, I'm not seeing the `epacamd_eia` asset in the DAG anymore. Is it now called `clean_epacamd_eia`? If so, does the metadata need to be updated?

src/pudl/etl/glue_assets.py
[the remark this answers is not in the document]
bendnorman: I was able to move `raw_epacamd_eia`, `clean_epacamd_eia`, `emissions_unit_ids_epacems`, `epacamd_eia_subplant_ids` into `etl.glue_assets` without any circular imports. I think your little hack of filtering out `epacamd_eia` of the `create_glue_tables` multi_asset is fine for now.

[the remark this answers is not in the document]
cmgosnell: Okay cool! Wild. Now I'm confused as to why I was having circular import problems having both `clean_epacamd_eia` and `epacamd_eia_subplant_ids` in the same module. but alas this does seem to work all in the same place 🎉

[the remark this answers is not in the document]
cmgosnell: I've converted this one little line from `if table_name != "epacamd_eia"` to `if "epacamd_eia" not in table_name` so this now excludes the two epacamd_eia tables from this multi-asset.

src/pudl/glue/epacamd_eia.py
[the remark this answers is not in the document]
cmgosnell: it is GONE

src/pudl/analysis/epacamd_eia.py
```diff
@@ -1,46 +1,18 @@
-"""Use the epacamd_eia crosswalk to connect EPA units to EIA generators and other data.
+"""Helper functions for filtering the EPA CAMD crosswalk table.
```
cmgosnell, line 1: This module now only contains the filtering... not sure if we should just rip it out altogether but it could certainly be helpful so I left it here and rejiggered the docs. Most of this module doc string I migrated over into the subplant_id metadata

src/pudl/etl/epacems_assets.py
```diff
@@ -148,8 +149,30 @@ def hourly_emissions_epacems(
...
+    Returns:
+        dataframe with unique set of: "plant_id_eia", "year" and "emissions_unit_id_epa"
+    """
+    epacems_ids = (
+        hourly_emissions_epacems[["plant_id_eia", "year", "emissions_unit_id_epa"]]
+        .drop_duplicates()
+        .compute()
+    )
+
+    return epacems_ids
```
cmgosnell, line 178: even though right now this is only used in the glue_assets, I kept it in cems land bc it feels more cems-y than gluey.

src/pudl/etl/glue_assets.py
```diff
@@ -42,28 +46,609 @@ def create_glue_tables(
...
+@asset(
+    required_resource_keys={"dataset_settings"}, io_manager_key="pudl_sqlite_io_manager"
+)
+def clean_epacamd_eia(
+    context,
+    raw_epacamd_eia: pd.DataFrame,
+    generators_entity_eia: pd.DataFrame,
+    boilers_entity_eia: pd.DataFrame,
+) -> pd.DataFrame:
+    """Clean up the EPACAMD-EIA Crosswalk file.
```
cmgosnell, line 82: I am now second-guessing on this asset name. If this lil guy is loaded into the db - which I believe it should be bc it is distinct from the new derived subplant_id table.... should it just be `epacamd_eia`?

src/pudl/etl/epacems_assets.py
cmgosnell: okay i went back in there to add some logs and defensive checks in this function. I learned that there are actually no null `subplant_id` after the current version of `make_subplant_ids`. I believe this is because I condensed the step where IDs were being merged into the crosswalk to be before `make_subplant_ids` instead of sprinkled throughout the process. I did a quick check to make sure that I wasn't missing any records this way and it seems solid:

image

soooo I was able to delete some of the fillna steps in this function. Right now it is only doing step one and four of the original documented steps:
```
    High-level overview of method:
        1. Use the PUDL subplant_id if available. In the case where a unit_id_pudl groups several subplants,
        we overwrite these multiple existing subplant_id with a single subplant_id.
        2. Where there is no PUDL subplant_id, we use the unit_id_pudl to assign a unique subplant_id
        3. Where there is neither a pudl subplant_id nor unit_id_pudl, we use the generator ID to
        assign a unique subplant_id
        4. All of the new unique ids are renumbered in consecutive ascending order
```

cmgosnell: okay i went back in there to add some logs and defensive checks in this function. I learned that there are actually no null `subplant_id` after the current version of `make_subplant_ids`. I believe this is because I condensed the step where IDs were being merged into the crosswalk to be before `make_subplant_ids` instead of sprinkled throughout the process. I did a quick check to make sure that I wasn't missing any records this way and it seems solid:

image

soooo I was able to delete some of the fillna steps in this function. Right now it is only doing step one and four of the original documented steps:
```
    High-level overview of method:
        1. Use the PUDL subplant_id if available. In the case where a unit_id_pudl groups several subplants,
        we overwrite these multiple existing subplant_id with a single subplant_id.
        2. Where there is no PUDL subplant_id, we use the unit_id_pudl to assign a unique subplant_id
        3. Where there is neither a pudl subplant_id nor unit_id_pudl, we use the generator ID to
        assign a unique subplant_id
        4. All of the new unique ids are renumbered in consecutive ascending order
```

test/unit/transform/epacems_test.py
```diff
@@ -37,3 +40,86 @@ def test_harmonize_eia_epa_orispl():
     )
     actual_df = epacems.harmonize_eia_epa_orispl(cems_test_df, crosswalk_test_df)
     pd.testing.assert_frame_equal(expected_df, actual_df, check_dtype=False)
+
+
+def test_epacamd_eia_subplant_ids():
+    """Ensure subplant_id gets applied appropriately to example plants."""
```
cmgosnell, line 46: I think my one remaining question is about where this lil unit test should live! rn there isn't really a place to unit test the `pudl.etl` directory. there is a 👶🏻 glue unit test module but that is running one test from the `pudl.glue` directory.... I kept this her for rn but it feels wrong. @bendnorman  do you have a sense of if/when we should migrate some of these unit tests into a new dagster-ified directory structure? should i leave this here for now? make a new module or dir?

src/pudl/etl/glue_assets.py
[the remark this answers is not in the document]
bendnorman: Sounds good to me!

src/pudl/analysis/epacamd_eia.py
bendnorman: I think we can leave it here for now.

src/pudl/etl/epacems_assets.py
bendnorman: Do you feel like this needs to be an asset? Could it just be a line in `def epacamd_eia_subplant_ids`?

src/pudl/etl/glue_assets.py
bendnorman: Yeah, I think it should be named `epacamd_eia` because it will be in the db. For now, the `raw_` and `clean_` prefixes are used for interim assets that aren't included in the db.

test/unit/transform/epacems_test.py
bendnorman: Hmm I'm not sure how we want to structure the tests in the long run. For now I feel like a `test/unit/transform/glue_test.py` file would be a good place for this test.

src/pudl/etl/glue_assets.py
aesharpe: is that not what the OG crosswalk is called though?

src/pudl/etl/epacems_assets.py
cmgosnell: bc its reading the chonky `hourly_emissions_epacems` even with dask it still takes a little over a minute to make. which is not the end of the world but it definitely makes testing it easier

src/pudl/etl/glue_assets.py
cmgosnell: this is the og crosswalk!

cmgosnell: and okay ty this is helpful! will rename back to `epacamd_eia`

src/pudl/etl/epacems_assets.py
bendnorman: Oh ok, that way, you can rematerialize `epacamd_eia_subplant_ids` without doing the expensive dask computation.

src/pudl/metadata/resources/eia.py
```diff
@@ -495,7 +495,10 @@
                 "previously_canceled",
             ],
             "primary_key": ["plant_id_eia", "generator_id"],
-            "foreign_key_rules": {"fields": [["plant_id_eia", "generator_id"]]},
+            "foreign_key_rules": {
```
bendnorman, line 498: Why are we excluding epacamd_eia_subplant_ids here?

bendnorman (approved):
🎉

src/pudl/metadata/resources/eia.py
cmgosnell: ah good question! bc this table has some plant ids that don't actually come from eia so they are not in this ID table

PR #2251, opened 2023-01-31
add methods to PudlTabl so it can be serialized and de-serialized (v2)

arengel:
Apologies for the duplicate PR but was having issues pushing to PUDL. Now that it is sorted, this PR replaces #2228. It isn't super urgent but I'm guessing it's easier not to have to futz with GitHub Actions. I've also updated this text based on additional changes I needed to make so that `Datastore` can be serialized, an issue that didn't arise with local testing.

## Motivation

The purpose of this PR is to enable `PudlTabl` to be serialized using `pickle`. In RMI's work with PUDL, we've often found it useful to save a copy of a `PudlTabl` before testing or running our data processing. This can save substantial amounts of time without jeopardizing validity because essentially all of our code is downstream of PUDL. The proposed changes here are a refined and generalized version of the solution we have used for a while now. In addition to being a more elegant solution than us subclassing or wrapping `PudlTabl`, it bring this functionality to all users, and the proposed tests should ensure that it continues to work properly.

## The changes

1. Two new methods for `PudlTabl`. A description of the default versions of these methods can be found [here](https://docs.python.org/3/library/pickle.html#pickle-inst), an example of how they might be modified in special cases is discussed [here](https://docs.python.org/3/library/pickle.html#pickle-state).
   1. `__getstate__` this converts the `PudlTabl` state into one that can be serialized with `pickle`. This is required because `pickle` doesn't like the `lambda` in the `self._dfs` `defaultdict` or `sa.Engine` so they need to be converted into form that `pickle` can deal with.
   2. `__setstate__` this takes the state as it was saved above and puts the pieces back into `PudlTabl` so a restored `PudlTabl` behaves the same as it would have before it was serialized. One important detail is that ``__init__`` is not run when an  object is de-serialized, so under some circumstances, setup that happened there may also need to occur here.
2. The same new methods for `GoogleCloudStorageCache`. This is required because one of its attributes is a `google.cloud.storage.Client` which cannot be pickled because "Pickling client objects is explicitly not supported. Clients have non-trivial state that is local and unpickleable." My solution here is to add `__getstate__` / `__setstate__` methods to `GoogleCloudStorageCache` which allow a new instance of it to be recreated based on the arguments from which the original object was created. This now allows `Datastore`, or at least the one I test with, to be serialized using pickle.
3. A new test
    1. `test_pudltabl_pickle` - this tests that we can take a `PudlTabl` with at least one df and that it can be pickled and restored from a pickle.

## A few additional points

[...]

PR #2498, opened 2023-04-05
Create Dynamic PudlTabl methods

zschira:
## Scope
This PR adds the ability to dynamically generate `PudlTabl` methods for retrieving dagsterized output tables from the PUDL DB. It also adds the output table `denorm_balance_sheet_assets_ferc1`, which can be used for demonstration/testing.


src/pudl/output/pudltabl.py
```diff
@@ -129,6 +131,35 @@ def __init__(
...
+        }
+
+        for group, assets in denorm_assets.items():
+            for key in pudl.helpers.get_asset_keys(assets):
+                asset_name = key.path[0]
+                self.__dict__[asset_name] = partial(
+                    self.get_table_from_db, table_name=asset_name, group=group
+                )
+
+    def get_table_from_db(self, table_name: str, group: str) -> pd.DataFrame:
```
zschira, line 152: This method can be used like `pudltabl.get_table_from_db("denorm_balance_sheet_assets_ferc1", "ferc1")` .

src/pudl/output/pudltabl.py
```diff
@@ -129,6 +131,35 @@ def __init__(
         # Used to persist the output tables. Returns None if they don't exist.
         self._dfs = defaultdict(lambda: None)
 
+        self._register_output_methods()
+
+    def _register_output_methods(self):
```
zschira, line 136: This method gets called from the constructor and will dynamically add methods to get assets from `denorm_assets`. This makes it possible to do:

```py
df = pudltabl.denorm_balance_sheet_assets_ferc1()
```

Under the hood it's just using `get_table_from_db`

devtools/python-output-table-conversion-debug.ipynb
```diff
@@ -11,7 +11,7 @@
   },
```
zschira, line 11: nb-output-clear stuff

src/pudl/output/pudltabl.py
```diff
@@ -129,6 +131,35 @@ def __init__(
...
+        denorm_assets = {
+            "ferc1": load_assets_from_modules(
+                [pudl.output.denorm_ferc1], group_name="ferc1"
+            ),
+            "eia": load_assets_from_modules([pudl.output.denorm_eia], group_name="eia"),
+        }
+
+        for group, assets in denorm_assets.items():
+            for key in pudl.helpers.get_asset_keys(assets):
+                asset_name = key.path[0]
```
bendnorman, line 147: Instead of indexing the path attribute, we can use `key.to_python_identifier()` which just merges the elements in path together.

bendnorman (changes requested):
This is great!

src/pudl/output/pudltabl.py
```diff
@@ -129,6 +131,35 @@ def __init__(
...
 
+        self._register_output_methods()
+
+    def _register_output_methods(self):
+        """Load denorm assets and register methods for retrieving each asset."""
+        denorm_assets = {
+            "ferc1": load_assets_from_modules(
+                [pudl.output.denorm_ferc1], group_name="ferc1"
+            ),
+            "eia": load_assets_from_modules([pudl.output.denorm_eia], group_name="eia"),
```
bendnorman, line 142: We'll probably want to load the normalized 714 and 861 assets because there are a lot of `PudlTabl` methods for these tables.

src/pudl/output/pudltabl.py
```diff
@@ -129,6 +131,35 @@ def __init__(
...
+    def get_table_from_db(self, table_name: str, group: str) -> pd.DataFrame:
+        """Grab output table from PUDL DB.
+
+        Args:
+            table_name: Name of table to get.
+            group: The data group to use for overrides, if any. E.g. "eia", "ferc1".
+        """
+        return pd.read_sql_table(table_name, self.pudl_engine).pipe(
+            apply_pudl_dtypes, group=group
+        )
```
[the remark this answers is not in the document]
zschira: Fixed

src/pudl/output/pudltabl.py
zschira: Thanks, this is much better!

src/pudl/output/pudltabl.py
zaneselvans: Should we only limit the interface to the normalized FERC-714 & EIA-861 tables, which are few, or provide access to all of the tables that were previously available even though many of them are `clean` rather than `norm`?

For FERC-714 we have:
* `respondent_id_ferc714` (normalized)
* `demand_hourly_pa_ferc714` (normalized)

For EIA-861 we have:
* `balancing_authority_eia861` (normalized)
* `balancing_authority_assn_eia861` (normalized)
* `utility_assn_eia861` (normalized)
* `service_territory_eia861`(normalized)
* Everything else (including `sales_eia861` which we utilize in the state demand allocation) is still un-normalized.

zschira: I added both of these. Right now the `clean_balancing_authority_eia861` is causing to break because it doesn't have `Resource` metadata right now

zschira: Also, are there any assets in the 714/861 modules that we don't want to have `PudlTabl` methods? Write now it will create a method for every asset it loads, but we could make it more selective

bendnorman: Oh yeah `clean_balancing_authority_eia861` is still using the default io manager. @zaneselvans is this intentional?

src/pudl/output/pudltabl.py
```diff
@@ -129,6 +131,39 @@ def __init__(
...
+        package = Package.from_resource_ids()
+        for key in pudl.helpers.get_asset_keys(output_assets):
+            asset_name = key.to_python_identifier()
+
+            # Create method called asset_name that will read the asset from DB
+            self.__dict__[asset_name] = partial(
+                self._get_table_from_db,
+                table_name=asset_name,
+                resource=package.get_resource(asset_name),
+            )
```
bendnorman, line 154: Instead of creating the entire package, you could also call `Resource.from_id(asset_name)`.

bendnorman: What do you think is best @zaneselvans? I feel like there are other places in the code base where we create a Package with all of the default resources to get one resource object.

bendnorman (commented):
Thanks for making the changes! Once we sort out which 861 and 714 tables should be included, we should delete all of their PudlTabl methods?

src/pudl/output/pudltabl.py
zschira: I went ahead and changed it to `Resource.from_id`, seems better

zschira:
I definitely think we should remove the old methods once we know which to keep, or it's going to start getting confusing. If we want we could define a list of tables that should have `PudlTabl` methods if loading all of the assets is casting too broad a net.

src/pudl/output/pudltabl.py
zaneselvans: I always default to creating a Package because Pydantic is very fast (compared to everything else we do) and there are some kinds of metadata that only get created when the whole Package is instantiated -- in particular the foreign key relationships are only built by the Package and not by the Resource, and I think I ran into problems attempting use a Resource that wasn't part of a Package to enforce the schemas in the FERC 1 tables.

zaneselvans: The `Package` could be created upon instantiation of the `PudlTabl` object rather than inside the function, and then you could just use `self._package.get_resource()` rather than creating a new `Package` from scratch.

src/pudl/output/pudltabl.py
zaneselvans: Ah. Hmm. Probably `clean_balancing_authority_eia861` should be getting pickled.

zaneselvans: I think we want methods for all of the EIA-861 and FERC-714 tables that are written to the DB, with the exception of the `clean_balancing_authority_eia861` which is just used to create the finished, normalized `balancing_authority_eia861` table.

bendnorman:
I suppose there is the potential we'll have denormalized assets that use the default io manager and aren't written to the database. For example, we [might not want to load `pu_eia`](https://github.com/catalyst-cooperative/pudl/pull/2490/files#r1158946377) to the database.

We could filter out assets based on io_manager they use:
```python
for asset in output_assets:
    # Iterate through assets keys because multi_assets contain a key for each asset it produces
    for key in asset.keys:
        io_manager = asset.get_io_manager_key_for_asset_key(key)
        # We only want to create methods for assets that have been loaded to the DB
        if io_manager == "pudl_sqlite_io_manager":
            asset_name = key.to_python_identifier()

            # Create method called asset_name that will read the asset from DB
            self.__dict__[asset_name] = partial(
                self._get_table_from_db,
                table_name=asset_name,
                resource=Resource.from_id(asset_name),
            )
```

zaneselvans:
If a table isn't getting written to the database, should it have a `Resource` definition? I guess we have some of these right now like `plant_parts_eia`

So are the only criteria for accessing the table through PudlTabl that:
* the table has a Resource definition.
* the table is written into the database.

zaneselvans:
Also I think we can't get around needing to have a manually compiled dictionary that maps our old method names to the underlying tables which they represent, since we abbreviated many of them.

bendnorman:
Good point about the abbreviations. If we need a mapping of table names to method names I guess we don't need to mess with importing assets and filtering by io_manager_key? Just have something like this:

```python
def _register_output_methods(self):
        """Load output assets and register a class method for retrieving each one."""
        table_method_mapping = {
          "denorm_plants_utils_ferc1": "pu_ferc1",
          "advanced_metering_infrastructure_eia861": "advanced_metering_infrastructure_eia861",
          ...
        }
        for table_name, method_name in table_method_mapping.items():
            self.__dict__[method_name] = partial(
                self._get_table_from_db,
                table_name=asset_name,
                resource=Resource.from_id(table_name),
            )
```

src/pudl/output/pudltabl.py
```diff
@@ -129,6 +130,76 @@ def __init__(
...
+            "utility_assn_eia861": "utility_assn_eia861",
+            "balancing_authority_eia861": "balancing_authority_eia861",
+            "balancing_authority_assn_eia861": "balancing_authority_assn_eia861",
+            # ferc714
+            "respondent_id_ferc714": "respondent_id_ferc714",
+            "demand_hourly_pa_ferc714": "demand_hourly_pa_ferc714",
+        }
+
+        for table_name, method_name in table_method_map.items():
+            if hasattr(PudlTabl, method_name):
```
zschira, line 180: We can probably remove this at some point, but for now it will generate a warning if we leave in an old class method after it's added to the map

bendnorman:
Ohh I think the CI is failing because the runner ran out of memory when reading in the `demand_hourly_pa_ferc714` table. This wasn't happening before because @zaneselvans adjusted reading logic in `PudlTabl.demand_hourly_pa_ferc714()` to read in chunks then concat:

```python
# Concatenating a bunch of smaller chunks reduces peak memory usage drastically
      # and doesn't seem to take any longer.
      return pd.concat(
          [
              # enforce_schema() cuts memory use by ~70% b/c of categorical tzones
              dhpa_res.enforce_schema(df)
              for df in pd.read_sql(
                  "demand_hourly_pa_ferc714",
                  self.pudl_engine,
                  chunksize=100_000,
              )
          ]
        )
```
I don't think there is any harm in using this chunking logic for all of the tables.

devtools/python-output-table-conversion-debug.ipynb
```diff
@@ -66,53 +66,12 @@
...
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": null,
    "id": "48c79723-28b5-4091-9590-7dedfd23d7df",
    "metadata": {
     "tags": []
    },
-   "outputs": [
```
bendnorman, line 74: We should also update the last step in `devtools/python-output-table-conversion-debug.ipynb`to to reflect the new way to update the PudlTabl methods.

zschira: Updated

zschira:
Good call on reading in chunks, I just made that update

zaneselvans:
We're reading all the tables in chunks in the IO Managers now too and it seems to work well and doesn't add a noticeable amount of time.  The peak memory usage is reduce by a crazy amount. Like 10x.  I don't really understand why.

src/pudl/metadata/resources/ferc1.py
```diff
@@ -839,6 +839,27 @@
         "sources": ["ferc1"],
         "include_in_database": False,
     },
+    "denorm_balance_sheet_assets_ferc1": {
+        "description": "Denormalized table that contains FERC balance sheet asset information.",
+        "schema": {
```
zaneselvans, line 844: Do we want to include `primary_key` definitions in the denormalized table schemas if we think they should exist? They'll get checked for being non-null and unique automatically then, and it would probably be good to know if we expect them to be valid primary keys but they are not.

src/pudl/output/pudltabl.py
zaneselvans: This is a great check for now!

src/pudl/output/pudltabl.py
```diff
@@ -129,6 +130,79 @@ def __init__(
...
+        """Grab output table from PUDL DB.
+
+        Args:
+            table_name: Name of table to get.
+            resource: Resource metadata used to enforce schema on table.
+        """
+        return pd.concat(
+            [
+                resource.enforce_schema(df)
+                for df in pd.read_sql(table_name, self.pudl_engine, chunksize=100_000)
```
zaneselvans, line 202: We'll need to ensure that this query respects values of `self.start_date` and `self.end_date` or time and memory intensive unexpected behavior will result.  Ella already has a solution for this in the branch I took over from her, so I can add it in, unless this breaks the CI.

src/pudl/output/pudltabl.py
```diff
@@ -155,206 +229,6 @@ def pu_ferc1(self) -> pd.DataFrame:
...
-    def advanced_metering_infrastructure_eia861(self) -> pd.DataFrame:
-        """An interim EIA 861 output function."""
-        return pd.read_sql(
-            "advanced_metering_infrastructure_eia861", self.pudl_engine
-        ).pipe(apply_pudl_dtypes, group="eia")
-
-    def balancing_authority_eia861(self) -> pd.DataFrame:
-        """An interim EIA 861 output function."""
-        return pd.read_sql("balancing_authority_eia861", self.pudl_engine).pipe(
-            apply_pudl_dtypes, group="eia"
```
zaneselvans, line 167: SO MUCH GLORIOUS DELETION

src/pudl/output/pudltabl.py
zschira: That'd be great if you could add that in!

src/pudl/metadata/resources/ferc1.py
zschira: Yeah good catch, checking the primary keys will probably be a good sanity check

src/pudl/output/pudltabl.py
zschira: ✂️✂️✂️

zaneselvans (approved):

bendnorman (approved):

bendnorman:
Oh whoops I forgot about the pickling issue :/

zaneselvans:
I think @arengel said we should probably go ahead and remove the pickling infrastructure and they would stick with an older version until the transition is complete.

bendnorman:
Oh I see. I removed the changes in #2251.

zaneselvans:
I'm going to go watch the sunset and if it CI passes I'll merge it in when I get back and do some work on #2490 this evening / over the weekend.

Issue #2377, opened 2023-03-10
Warn user when a change to the PUDL DB schema is detected


jdangerx:
This is because we no longer `--clobber` and drop tables; the tables are created from the `Package.from_resource_list().to_sql()` one time when the DB is created.

```[tasklist]
### Scope
- [ ] #2331
```

#2378 is not in scope. I tried to do this, but because the `pudl_sqlite_io_manager` now gets created once per process, we run into a bunch of race conditions where the different processes are trying to drop/recreate tables (or, worse, delete/recreate files on disk).

The pysqlite transaction handling has "surprising behavior," so I'm trying to get the transactions to work following this: https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#pysqlite-serializable

bendnorman:
After adding a couple of output tables I found recreating the database every time a schema changes pretty frustrating. I wonder if instead of deleting the table data every time an asset is materialized, we drop the table, recreate the schema and write the data. I expect sqlalchemy will throw an error when the table that is a parent in a foreign key relationship is dropped. 

I think using database migrations is the best option for not having to recreate the database every time. I suspect managing the overhead of migrations is worth not having to rerun the full ETL whenever schemas change.

jdangerx:
I think that migrations are a good idea, though I'm not quite sure how we want to go about it with all the concurrency still going on. Maybe we have the IO managers each do this in their initialization:

1. begin transaction (this might implicitly happen when we run any DDL stuff?)
2. run a migration
3. end transaction

I think that in theory, if a different process were to try to run a migration in the middle, they would be met with some sort of lockout error. And when they tried again, the migration would be a no-op.

I'm sure there are edge cases that will break that, though.

jdangerx:
Another thing we could do is:

* still just have the ETL throw an error if the metadata is mismatched
* in nightly builds, have the script run a db migration before starting the ETL process
* locally, have a human say, "aw shoot!" and then run the db migration manually

bendnorman:
I was thinking we'd run migrations outside of dagster runs so we don't have to recreate the entire database whenever a schema changes. However, that doesn't solve the problem of notifying users when the metadata has changed and they need to recreate the database or apply migrations.

I have two potential solutions for notifying folks when the metadata has changed:

[...]

PR #2504, opened 2023-04-07
Use Alembic to set up database schema ahead of DAG run

jdangerx:
Setting up the pudl.sqlite DB & schema before the DAG run lets us stop worrying about concurrency & whether the schema is fully set up in any number of different race conditions. This should make our nightly builds less flaky & also allow us to do some more rigorous checks on our schema in the future.

This:
* configures `alembic` to point at our existing metadata object + our pudl.sqlite location
* generates one migration to get db synced with the metadata in this branch
* includes documentation on how to run the ETL now
* updates the nightly build scripts to use `alembic` to keep db schema synced with the code (though this is untested as of now).
* adds some checks to IOManager for "is the db initialized already", which breaks tests for now - we would need to run `alembic upgrade` in the test setup - which I think is OK.
See discussion in #2377 about whether we want to pull in alembic, or write our own script that does similar things.


bendnorman (commented):
This is great @jdangerx! Thanks for experimenting. It pretty much works as expected 👍 Going to leave more thoughts in #2377.

jdangerx:
Closing in favor of #2515 .

PR #2469, opened 2023-03-29
Solve `2023-03-29` nightly build failure issues

bendnorman:
# PR Overview

We got a few failures in the most recent nightly build. It looks like there are [two new PRs](https://github.com/catalyst-cooperative/pudl/pulls?q=is%3Apr+is%3Aclosed+merged%3A2023-03-28..2023-03-29+base%3Adev+sort%3Aupdated-desc+) that made it into the build:
1. A log file for the build is not being written to the output bucket or sent to the #pudl-deployments slack channel. I pulled the test errors below from [Cloud Logging](https://console.cloud.google.com/logs/query;query=resource.type%3D%22gce_instance%22%20resource.labels.instance_id%3D%222178497257830677871%22;timeRange=2023-03-29T09:12:49.926Z%2F2023-03-29T16:12:49.926935Z;cursorTimestamp=2023-03-29T14:15:38.299009740Z?project=catalyst-cooperative-pudl). 

2. `test/validate/eia_test.py::test_no_null_cols_eia` failed on the `frc_eia923` table:
```
E           ValueError: Null columns found in frc_eia923: ['bulk_agg_fuel_cost_per_mmbtu']
```
@zaneselvans @e-belfer did #2447 touch this table?

3. `test/integrations/etl_test.py:: test_pudl_engine` found a foreign key constraint failure:
```
E           pudl.io_managers.ForeignKeyErrors: Foreign key error for table: purchased_power_ferc1 -- power_purchase_types_ferc1 {self.foreign_key} -- on rows {self.rowids}
```
I'm a dingdong and forgot to make the second half of the error message an f-string. `purchased_power_ferc1` is the child table and `power_purchase_types_ferc1` is the parent table. I don't think either of the two PRs directly changed anything about FERC tables. I'm going to download `pudl.sqlite` and run the fk checks locally to debug this.

# PR Checklist

- [ ] Merge the most recent version of the branch you are merging into (probably `dev`).
- [ ] All CI checks are passing. [Run tests locally to debug failures](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#running-tests-with-tox)
- [ ] Make sure you've included good docstrings.
- [ ] For major data coverage & analysis changes, [run data validation tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#data-validation)
- [ ] Include unit tests for new functions and classes.
- [ ] Defensive data quality/sanity checks in analyses & data processing functions.
[...]

PR #2515, opened 2023-04-10
Add script to set DB schema before running ETL pipeline

### LINKED ISSUE #2377: Warn user when a change to the PUDL DB schema is detected

jdangerx:
Closes #2377 

This makes it so that we don't start with a completely empty DB, we start with one that has the same schema as whatever is written in our `pudl.metadata` package.

This sidesteps the various concurrency issues with trying to make sure the schema is correct *within* the DAG run.

In the future, we might want to use migrations, or have some sort of non-concurrent setup op in the DAG, but for now this will provide a decent UX and hopefully get around the flakiness we've been seeing with the nightly builds (see #2469).

We decided to not use `alembic` for the migrations because of the slightly more complicated UX, but I think using it for diffing schemas is fine and I trust them more than I trust myself, to catch the various edge cases of "does this database setup look right?"

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,7 @@ dynamic = ["version"]
 license = {file = "LICENSE.txt"}
 dependencies = [
     "addfips>=0.4,<0.5",
+    "alembic>=1.10.3,<1.11",
     "catalystcoop.dbfread>=3.0,<3.1",
     "catalystcoop.ferc-xbrl-extractor==0.8.2",
     "coloredlogs>=14.0,<15.1",  # Dagster requires 14.0
@@ -93,10 +94,11 @@ epacems_to_parquet = "pudl.convert.epacems_to_parquet:main"
 ferc_to_sqlite = "pudl.ferc_to_sqlite.cli:main"
 datasette_metadata_to_yml = "pudl.convert.datasette_metadata_to_yml:main"
 pudl_datastore = "pudl.workspace.datastore:main"
-pudl_etl = "pudl.cli:main"
+pudl_etl = "pudl.cli.etl:main"
 pudl_setup = "pudl.workspace.setup_cli:main"
 state_demand = "pudl.analysis.state_demand:main"
 pudl_check_fks = "pudl.etl.check_foreign_keys:main"
+pudl_reset_db = "pudl.cli.reset_db:main"
 # pudl_territories currently blows up memory usage to 100+ GB.
 # See https://github.com/catalyst-cooperative/pudl/issues/1174
 # pudl_territories = "pudl.analysis.service_territory:main"
--- a/src/pudl/cli/reset_db.py
+++ b/src/pudl/cli/reset_db.py
@@ -1,0 +1,54 @@
+"""Reset the PUDL output DB to match the schema in pudl.metadata."""
+import argparse
+import pathlib
+
+import sqlalchemy as sa
+
+from pudl.metadata.classes import Package
+from pudl.workspace.setup import get_defaults
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser(__doc__)
+    parser.add_argument(
+        "--sqlite-path", help="The path to the PUDL DB that you'd like to reset."
+    )
+    return parser.parse_args()
+
+
+def reset_db(
+    sqlite_path: str | None = None, metadata: sa.MetaData | None = None
+) -> None:
+    """Ensure empty SQLite DB at `sqlite_path` with schema `metadata`.
+
+    Args:
+        sqlite_path: the path to DB. Defaults to $PUDL_OUTPUT/pudl.sqlite
+        metadata: SQLAlchemy MetaData describing desired DB state. Defaults to
+            "all the tables defined in pudl.metadata".
+    """
+    if sqlite_path is None:
+        sqlite_path = pathlib.Path(get_defaults()["pudl_out"]) / "pudl.sqlite"
+
+    if metadata is None:
+        metadata = Package.from_resource_ids().to_sql()
+
+    # Check if a SQLite file already exists at the given path, and delete it if it does
+    if sqlite_path.exists():
+        sqlite_path.unlink()
+
+    # Use SQLAlchemy to create a new SQLite database at the given path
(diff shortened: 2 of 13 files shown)


pyproject.toml
```diff
@@ -93,10 +94,11 @@ epacems_to_parquet = "pudl.convert.epacems_to_parquet:main"
 ferc_to_sqlite = "pudl.ferc_to_sqlite.cli:main"
 datasette_metadata_to_yml = "pudl.convert.datasette_metadata_to_yml:main"
 pudl_datastore = "pudl.workspace.datastore:main"
-pudl_etl = "pudl.cli:main"
+pudl_etl = "pudl.cli.etl:main"
 pudl_setup = "pudl.workspace.setup_cli:main"
 state_demand = "pudl.analysis.state_demand:main"
 pudl_check_fks = "pudl.etl.check_foreign_keys:main"
```
bendnorman, line 100: Should we move `pudl.etl.check_foreign_keys` to `pudl.cli.check_foreign_keys`?

pyproject.toml
```diff
@@ -93,10 +94,11 @@ epacems_to_parquet = "pudl.convert.epacems_to_parquet:main"
...
+pudl_reset_db = "pudl.cli.reset_db:main"
```
bendnorman, line 101: Out of scope for this PR, but I wonder if we should think about a command tool reorg? We could group types of commands together:

- pudl db check_foreign_keys
- pudl db setup
- pudl etl run
- pudl ferc_to_sqlite run

IDK maybe not worth the time and confusion.

src/pudl/cli/reset_db.py
```diff
@@ -0,0 +1,54 @@
...
+    if sqlite_path.exists():
+        sqlite_path.unlink()
+
+    # Use SQLAlchemy to create a new SQLite database at the given path
+    engine = sa.create_engine(f"sqlite:///{sqlite_path}")
+    conn = engine.connect()
+
+    # Use the metadata to initialize the table schema in the new database
+    metadata.create_all(conn)
+    conn.close()
```
bendnorman, line 45: Could use:
```python
with engine.connect() as conn:
   metadata.create_all(conn)
```
So you don't have to call `conn.close()`.

src/pudl/cli/reset_db.py
```diff
@@ -0,0 +1,54 @@
...
+    Args:
+        sqlite_path: the path to DB. Defaults to $PUDL_OUTPUT/pudl.sqlite
+        metadata: SQLAlchemy MetaData describing desired DB state. Defaults to
+            "all the tables defined in pudl.metadata".
+    """
+    if sqlite_path is None:
+        sqlite_path = pathlib.Path(get_defaults()["pudl_out"]) / "pudl.sqlite"
+
+    if metadata is None:
+        metadata = Package.from_resource_ids().to_sql()
```
bendnorman, line 33: Should we add a couple of logging messages here to let folks know what is happening?

src/pudl/io_managers.py
```diff
@@ -412,12 +414,26 @@ def __init__(
...
 
+        existing_schema_context = MigrationContext.configure(self.engine.connect())
+        metadata_diff = compare_metadata(existing_schema_context, self.md)
+        if metadata_diff:
+            logger.info(f"Metadata diff:\n\n{metadata_diff}")
+            raise RuntimeError(
+                "Database schema has changed, try running `alembic revision "
+                "--autogenerate -m 'Your change message'` then `alembic "
+                "upgrade`."
+            )
```
bendnorman, line 435: This error message should tell users to run `pudl_reset_db` right?

bendnorman (changes requested):
This is great! Just a couple of small requests.

pyproject.toml
jdangerx: Yeah, I think that would be nice - though if we're pulling out `pudl_etl` and `ferc_to_sqlite` and getting rid of the pudl settings, we might find ourselves with a pretty small # of CLI tools soon.

src/pudl/cli/reset_db.py
jdangerx: Yeah, probably. I'll add some stuff.

src/pudl/cli/reset_db.py
jdangerx: Yeah, good thought - will add!

src/pudl/io_managers.py
jdangerx: Yes!

jdangerx (commented):
Thanks for the suggestions!

src/pudl/cli/reset_db.py
```diff
@@ -32,17 +35,18 @@ def reset_db(
...
     engine = sa.create_engine(f"sqlite:///{sqlite_path}")
-    conn = engine.connect()
 
     # Use the metadata to initialize the table schema in the new database
-    metadata.create_all(conn)
-    conn.close()
+    with engine.connect() as conn:
+        logger.info(f"Writing schema to DB at {sqlite_path}...")
+        metadata.create_all(conn)
+        conn.close()
```
bendnorman, line 49: `conn.close()` is called when using the context manager, so you can delete this line.

bendnorman (approved):

bendnorman:
A tiny comment that isn't a blocker but would be nice to address.

src/pudl/cli/reset_db.py
jdangerx: Oh, duh. Thanks for catching this!

PR #2521, opened 2023-04-13
Ferc output conversion

e-belfer:
Convert these output tables to be dagster python assets. These assets should live in the `pudl.output.denorm_ferc1` module.

Follow the output table conversion instructions laid out in the `devtools/python-output-table-conversion-debug.ipynb`.
```[tasklist]
- [x] `denorm_plants_utils_ferc1`
- [x] `denorm_plants_steam_ferc1`
- [x] `denorm_plants_small_ferc1`
- [x] `denorm_plants_hydro_ferc1`
- [x] `denorm_fuel_ferc1`
- [x] `denorm_plants_pumped_storage_ferc1`
- [x] `denorm_purchased_power_ferc1`
- [x] `denorm_plant_in_service_ferc1`
- [x] `denorm_plants_utilities_ferc1`
- [x] `denorm_plants_all_ferc1` (derived)
- [x] `denorm_fuel_by_plant_ferc1` (derived)
- [x] `denorm_balance_sheet_assets_ferc1`
- [x] `denorm_balance_sheet_liabilities_ferc1`
- [x] `denorm_cash_flow_ferc1`
- [x] `denorm_depreciation_amortization_summary_ferc1`
- [x] `denorm_electric_energy_dispositions_ferc1`
- [x] `denorm_electric_energy_sources_ferc1`
- [x] `denorm_electric_operating_expenses_ferc1`
- [x] `denorm_electric_operating_revenues_ferc1`
- [x] `denorm_electric_plant_depreciation_changes_ferc1`
- [x] `denorm_electric_plant_depreciation_functional_ferc1`
- [x] `denorm_electricity_sales_by_rate_schedule_ferc1`
- [x] `denorm_income_statement_ferc1`
- [x] `denorm_other_regulatory_liabilities_ferc1`
- [x] `denorm_retained_earnings_ferc1`
- [x] `denorm_transmission_statistics_ferc1`
- [x] `denorm_utility_plant_summary_ferc1`
```

Still to do:
- [x] Add new fields (those commented out) to `pudl.metadata.fields.py`, where relevant
- [x] Verify similarity between new and old outputs
- [x] Depracate functions in `pudl.output.ferc1.py`
- [x] Update `PudlTabl` functions
- [x] Figure out what to do with the `thresh` parameter in `denorm_plant_by_fuel_ferc1`

Out of scope
- Update notebooks to point to new pathways

# PR Checklist


zaneselvans:
For the `thresh` parameter in `fuel_by_plant_ferc1` I think you could make it a config so it could be tweaked externally if need be.  However in practice I don't think we've really treated it like a dynamic value.

e-belfer:
Remaining differences to resolve in PudlTabl deprecation:

`denorm_plants_all_ferc1` has a numeric ExtensionArray, while `plants_all_ferc1` has a float ExtensionArray.

`denorm_fuel_by_plant_ferc1` has more columns than `fbp_ferc1`. These are completely empty in the output table, and perhaps are getting dropped from the old df for this reason?: ['hydro_fraction_cost', 'hydro_fraction_mmbtu', 'other_fraction_cost', 'other_fraction_mmbtu', 'solar_fraction_cost', 'solar_fraction_mmbtu', 'wind_fraction_cost', 'wind_fraction_mmbtu']

zaneselvans:
I had a weirdness like this with an ExtensionArray in the net gen stuff too I think, or maybe even in the EIA output tables which I don't think I resolved. Not sure what was going on.

e-belfer:
I found the issue (edited above before I saw your comment) - I'm casting `ferc_license_id` to an integer here, and it's a float in the original output dataframe. Seems like an improvement. Finally found this by running `old_df_schema.compare(new_df)`

e-belfer:
There's an unexplained row count difference between the original and updated `fuel_by_plant_ferc1` output table. 
AssertionError: DataFrame are different

DataFrame shape mismatch
[left]:  (1465, 20)
[right]: (1920, 20)

These additional rows appear in the new `denorm_fuel_by_plant_ferc1` and are all rows with data. Outputs upstream look identical. Running the transform code manually on inputs also produces an output matching `denorm_fuel_by_plant_ferc1`. One further oddity: the columns noted above (e.g. 'wind_fraction_mmbtu') are getting dropped in the old output table, but not anywhere I'm able to see. Is it possible this table is getting transformed in a second location?

e-belfer:
Further exploration of differences. New dataframe:
image

Old dataframe:
image

Removing all records where there is neither primary fuel method removes 525 rows, while the difference between both dataframes is 437.

The upstream assets are confirmed to be identical.

src/pudl/output/new_ferc1.py
```diff
@@ -14,12 +18,249 @@ def denorm_plants_utilities_ferc1(
...
+
+    Args:
+        denorm_plants_utilities_ferc1: Denormalized dataframe of FERC Form 1 plants and
+            utilities data.
+        plants_steam_ferc1: The normalized FERC Form 1 steam table.
+
+    Returns:
+        A DataFrame containing useful fields from the FERC Form 1 steam table.
+    """
+    steam_df = (
```
jdangerx, line 41: It would be nice to share the code between this code and the old code, unless we decided that we are deleting the old code right now!

I think we can do this by just calling the asset in the old code:

```
@asset(...)
def denorm_plants_steam_ferc1(denorm_plants_utilities_ferc1, plants_steam_ferc1):
    # do cool stuff

def plants_steam_ferc1(...)
    "Deprecation warning, yada yada"
    # read input tables from SQL
    denorm_plants_steam_ferc1(... input tables that we just read above)
```

Might make it a little easier to have confidence in the code going forward - but if we're just gonna delete the old functions altogether then this whole point is moot.

jdangerx (changes requested):
This mostly looks good!

Changes requested:
* I'm curious about your thoughts on collapsing the old/new output code - if you think that's a bad idea, you don't have to do it, just want to know what you think :)
* Also want to make sure we can get away with not changing any of the PudlTabl calling code (i.e. in the tests)

If that all is sorted, feel free to Dismiss Review And Merge!

test/validate/plants_steam_ferc1_test.py
```diff
@@ -30,7 +31,9 @@ def test_vs_bounds(pudl_out_ferc1, live_dbs, cases):
     """Test distributions of reported plants_steam_ferc1 columns."""
     if not live_dbs:
         pytest.skip("Data validation only works with a live PUDL DB.")
-    validate_df = pudl_out_ferc1.plants_steam_ferc1().assign(
+    validate_df = pd.read_sql(
```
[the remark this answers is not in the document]
e-belfer: We could definitely do this, I'm just anticipating the eventual deprecation of the PudlTabl method. These methods should be identical, however.

src/pudl/metadata/fields.py
```diff
@@ -245,6 +245,41 @@
...
+    },
+    "capex_annual_per_kw": {
+        "type": "number",
+        "description": "Annual capital addition into `capex_total` per kw.",
+        "unit": "USD/kw",
+    },
+    "capex_annual_per_mw": {
+        "type": "number",
+        "description": "Annual capital addition into `capex_total` per mw.",
+        "unit": "USD/mw",
```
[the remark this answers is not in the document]
e-belfer: Great catch, updated.

test/validate/fuel_ferc1_test.py
```diff
@@ -15,7 +16,8 @@
 def test_fuel_ferc1_trivial(pudl_out_ferc1):
     """Test output routines for tables from FERC Form 1."""
     logger.info("Compiling FERC Form 1 fuel table...")
-    logger.info(f"{len(pudl_out_ferc1.fuel_ferc1())} fuel records found")
+    fuel_tab = pd.read_sql("denorm_fuel_ferc1", pudl_out_ferc1.pudl_engine)
+    logger.info(f"{len(fuel_tab)} fuel records found")
```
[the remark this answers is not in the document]
e-belfer: Agreed, added!

src/pudl/metadata/resources/ferc1.py
```diff
@@ -864,6 +864,807 @@
...
+                "energy_charges",
+                "non_coincident_peak_demand_mw",
+                "other_charges",
+                "purchase_type_code",
+                "purchased_mwh",
+                "received_mwh",
+                "tariff",
+                "total_settlement",
+            ],
+            # Flag for review - there are null values in this combination.
```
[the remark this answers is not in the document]
e-belfer: I've removed this set of values as the primary key.

src/pudl/output/new_ferc1.py
e-belfer: We've deleted these!

src/pudl/metadata/fields.py
```diff
@@ -245,6 +245,41 @@
...
+        "unit": "USD",
+    },
+    "capex_annual_addition_rolling": {
+        "type": "number",
+        "description": "Year-to-date capital addition into `capex_total`.",
+        "unit": "USD",
+    },
+    "capex_annual_per_kw": {
+        "type": "number",
+        "description": "Annual capital addition into `capex_total` per kw.",
```
aesharpe, line 260: It's a tad inconsistent to have `capex_annual_addition` and `capex_annual_XX` refer to the same type of thing. Maybe we could change `capex_annual_addition` to just `capex_annual`?

src/pudl/metadata/fields.py
```diff
@@ -736,6 +784,10 @@
         "type": "string",
         "description": "The docket number relating to the FERC qualifying facility exempt wholesale generator status.",
     },
+    "ferc_license_id": {
+        "type": "string",
+        "description": "The FERC license ID of a project.",
+    },
```
aesharpe, line 790: I think there is already a field called `license_id_ferc1`. I'm guessing they are referring to the same thing?

src/pudl/metadata/fields.py
```diff
@@ -818,14 +870,29 @@
...
     "fuel_consumed_units": {
         "type": "number",
         "description": "Consumption of the fuel type in physical unit. Note: this is the total quantity consumed for both electricity and, in the case of combined heat and power plants, process steam production.",
     },
+    "fuel_cost": {
+        "type": "number",
+        "description": "Total fuel cost for plant (in $USD).",
+        "unit": "USD",
+    },
```
aesharpe, line 886: Might be helpful to explain the difference between `fuel_consumed_total_cost` and `fuel_cost`. Also the fact that only one has the word `total` in it is a little confusing. Also if one of the descriptions says (in USD) then all the cost fields probably should. Unless the metadata has a way to importing the `unit` field as well which would be nice.

src/pudl/metadata/fields.py
```diff
@@ -818,14 +870,29 @@
...
     "fuel_cost_from_eiaapi": {
         "type": "boolean",
         "description": "Indicates whether the fuel cost was derived from the EIA API.",
     },
+    "fuel_mmbtu": {
```
aesharpe, line 891: Should this be `fuel_consumed_mmbtu`? I'm not super clear on the distinction.

aesharpe (commented):
Reviewed the contents of `fields.py`. Looks mostly good! Just a few comments on column name consistency and distinguishing between fuel consumed and fuel total. 

Another nit is that there are a few instances where the description says "mwh" when it should probably say "MWh". Similarly, there are instances of "mmBTU" that should be "MMBtu". 

I'm not sure what the formatting protocol is for the `unit` field, but we should make it consistent there too. It looks like it should probably be all lower case with _ separators, but it looks like there is some variation. Maybe @zaneselvans has thoughts here?

src/pudl/metadata/fields.py
e-belfer: Moved to #2533

src/pudl/metadata/fields.py
e-belfer: Moved to #2533

PR #2526, opened 2023-04-15
Run integration & validation tests in parallel in the background.

zaneselvans:
# PR Overview

I think we can shave an hour off of the nightly builds by running the unit+integration tests and the validation tests at the same time. Each of them takes about an hour on their own but they have no interdependency. I run them at the same time locally against a full DB whenever I'm trying to see if I broke something after a big change.

I'm not sure if just running them in the background is the right way to do this, but it seems like there should be *some* simple 2 line change that does it! If not these 2 lines, do you have other ideas @jdangerx?

# PR Checklist

### DIFF
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -32,16 +32,11 @@ function run_pudl_etl() {
         --loglevel DEBUG \
         --gcs-cache-path gs://internal-zenodo-cache.catalyst.coop \
         $PUDL_SETTINGS_YML
-    # Run unit+integration tests and data validations in parallel in the background
-    # They each take ~1 hour on their own and have no interdependencies.
-    pytest \
+    && tox parallel -e unit,integration,validation --parallel-live \
+        -- \
         --gcs-cache-path=gs://internal-zenodo-cache.catalyst.coop \
         --etl-settings=$PUDL_SETTINGS_YML \
-        --live-dbs test/unit test/integration &
-    pytest \
-        --gcs-cache-path=gs://internal-zenodo-cache.catalyst.coop \
-        --etl-settings=$PUDL_SETTINGS_YML \
-        --live-dbs test/validate &
+        --live-dbs
 }
 
 function shutdown_vm() {
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -31,11 +31,17 @@ function run_pudl_etl() {
     && pudl_etl \
         --loglevel DEBUG \
         --gcs-cache-path gs://internal-zenodo-cache.catalyst.coop \
-        $PUDL_SETTINGS_YML \
-    && pytest \
+        $PUDL_SETTINGS_YML
+    # Run unit+integration tests and data validations in parallel in the background
+    # They each take ~1 hour on their own and have no interdependencies.
+    pytest \
         --gcs-cache-path=gs://internal-zenodo-cache.catalyst.coop \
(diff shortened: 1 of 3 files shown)


jdangerx:
I think this *should* work! Let me kick off a build on this branch and see what happens :)

jdangerx:
Build succeeded, but as Zane mentioned in slack:

> This seems suspiciously fast. Did it just background the tests and then exit thinking that they were “finished”?

I think that's actually probably the case. 

One thing we could do is use [`tox run-parallel`](https://tox.wiki/en/latest/cli_interface.html#tox-run-parallel-(p)) to run these tests. Not quite sure if the `-o` flag actually captures stdout in the way we want, but worth a try.

docker/gcp_pudl_etl.sh
```diff
@@ -31,11 +31,17 @@ function run_pudl_etl() {
...
-        $PUDL_SETTINGS_YML \
-    && pytest \
+        $PUDL_SETTINGS_YML
+    # Run unit+integration tests and data validations in parallel in the background
+    # They each take ~1 hour on their own and have no interdependencies.
+    pytest \
+        --gcs-cache-path=gs://internal-zenodo-cache.catalyst.coop \
+        --etl-settings=$PUDL_SETTINGS_YML \
+        --live-dbs test/unit test/integration &
+    pytest \
```
rousik, line 41: Is there a specific reason for controlling the boundaries for parallel/serial execution in this manner over something like using `pytest-xdist` and parallelizing everything? https://www.tutorialspoint.com/pytest/pytest_run_tests_in_parallel.htm

jdangerx: I had thought there was *some* barrier to using `pytest-xdist` before (hence suggesting `tox run-parallel`), but @zaneselvans might know better.

When I experimented with splitting up tests across multiple workers using `pytest-shard`, I found that each test process still had to instantiate some very expensive test fixtures, so that didn't end up improving our times that much...

zaneselvans: A big majority of the time that the tests spend is setting up fixtures, and IIRC when we messed around with `pytest-shard` and `pytest-xdist` it resulted in many different copies of the same time-intensive fixtures getting set up, which didn't save any time.

The reason why this might work well for the `validate` and `full_integration` tests is because they mostly just read out of a pre-populated database (b/c `--live-dbs`) so I think any of these parallelization schemes would probably be okay.  I don't think we're going to get a ton of benefit from throwing more than 2 CPUs at this problem right now because both of these tests still depend on `PudlTabl` fixtures which are resource intensive. But as more of what was being done in `PudlTabl` is written directly into the DB, more CPUs will be helpful.

We could use `tox -p` which would split up the tests into 2 groups (integration + validate) well, but right now I don't think anything in the nightly builds are being run using Tox, because the Docker container already has the required Python environment in it. There are [a couple of different ways](https://pytest-xdist.readthedocs.io/en/stable/distribution.html) to distribute tests to workers with `pytest-xdist` by module, file, or marked test group.

rousik (approved):

jdangerx:
I don't remember what was stopping us from merging this, so I'm going to re-run the nightly builds on this branch and figure out what's going on.

zaneselvans:
I think the parallel tox environments might have been broken by our "re-use the same directory for all environments" setup, which I've tried and failed to untangle in the past

jdangerx:
After running nightly builds with this several times, and running into a new error each time, I think we're going to close this and try a different approach.

PR #2527, opened 2023-04-15
Convert EIA generation and fuel allocations to Dagster

zaneselvans:
# PR Overview

Migrate the net generation allocation process into software defined assets.

# PR Checklist


cmgosnell (commented):
bc of the re-ordering of functions, its a little hard to tell if you ended up needing to muck with the inside of these functions

src/pudl/metadata/resources/allocate_net_gen.py
```diff
@@ -0,0 +1,196 @@
...
+    },
+    # "generation_fuel_by_generator_energy_source_owner_monthly_eia923": {
+    #     "description": (
+    #         "Monthly estimated net generation and fuel consumption for each generator, "
+    #         "broken down by energy source, prime mover, and owner. Based on data "
+    #         "reported in the EIA-923 generation and generation_fuel tables."
+    #     ),
+    #     "schema": {
+    #         "fields": [
+    #             "report_date",
```
cmgosnell, line 132: should this section go away? why is only the monthly commented out?

src/pudl/metadata/resources/allocate_net_gen.py
```diff
@@ -0,0 +1,196 @@
+"""Resource metadata for the allocate_net_gen tables."""
+from typing import Any
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = {
```
cmgosnell, line 4: instead of full duplicating each of these table everything could we do something like:
```
{f"table_name_{freq}_eia": {"description": f"{freq} table stuff", ...}  for freq in ["monthly", "annually"]}
```

src/pudl/metadata/resources/allocate_net_gen.py
```diff
@@ -0,0 +1,196 @@
...
+                "fuel_consumed_mmbtu",
+                "fuel_consumed_for_electricity_mmbtu",
+            ],
+            "primary_key": [
+                "report_date",
+                "plant_id_eia",
+                "generator_id",
+                "prime_mover_code",
+                "energy_source_code",
+                "utility_id_eia",  # This is the OWNER not the operator.
```
cmgosnell, line 188: this feels like a note that is relevant to add into the table description or to overwrite in the table-specific field description overwriting place

cmgosnell: alternatively we could keep the id name as the owner name ( i think we are currently renaming the owner id ). I believe this rename was requested by someone at rmi bc it makes the a little more seamless to use in conjunction with other tables so I am on the fence about that.

cmgosnell (commented):
it looks good in general! bc of the re-ordering of functions in allocate_net_gen, its kinda hard to tell if you had to change anything inside of the function themselves... also this is a gonna be a hella messy merge now with the outstanding allocation pr... which i'm sure i can handle but the re-org in general is not idea.

src/pudl/analysis/allocate_net_gen.py
```diff
@@ -196,43 +197,169 @@
...
+                key=f"generation_fuel_by_generator_energy_source_{agg_freqs[freq]}_eia923"
+            ),
+            "pu": AssetIn(key="denorm_plants_utilities_eia"),
+            "bga": AssetIn(key="boiler_generator_assn_eia860"),
+        },
+        io_manager_key=io_manager_key,
+        compute_kind="Python",
+    )
+    def gen_fuel_by_gen(
+        net_gen_fuel_alloc: pd.DataFrame, pu: pd.DataFrame, bga: pd.DataFrame
```
cmgosnell, line 268: this is more of an idk how asset factory's work question than a suggestion/critique. Why make all of these assets in the same factory? would it by possible and maybe more de-tangled to have separate little factories for each of these assets that flow from the  the main `allocate_gen_fuel_by_generator_energy_source`?

src/pudl/analysis/allocate_net_gen.py
```diff
@@ -196,43 +197,169 @@
...
+    assets = [gen_fuel_by_gen_esc, gen_fuel_by_gen]
+    if freq == "AS":
+        # The monthly version is yuuuuge and we only use the annual data for now.
+        assets += [gen_fuel_by_gen_esc_owner]
+
+    return assets
+
+
+allocate_net_gen_assets = [
+    ass
```
cmgosnell, line 324: wow

src/pudl/analysis/allocate_net_gen.py
```diff
@@ -196,43 +197,169 @@
...
+                on=["plant_id_eia", "generator_id"],
+                date_on=["year"],
+                how="left",
+                report_at_start=True,
+                validate="m:1",
+            ),
+            own_eia860=own_eia860,
+            scale_cols=DATA_COLUMNS + ["capacity_mw"],
+            validate="m:m",  # m:m because there are multiple generators in gen_pm_fuel
+        )
```
cmgosnell, line 313: in a similar fashion to the other assets in here, does it make sense to have the actual work happening inside the factory insides of calling a version of `scale_allocated_net_gen_by_ownership`? if you are going to leave it in here i think you should at least migrate the more fleshed out documentation of that step into this function w/in a factory

cmgosnell: also now i'm noticing that you did not remove `scale_allocated_net_gen_by_ownership` so you should certainly do that or rely on it.

cmgosnell (changes requested):
mostly a few tiny docs things + some im a dagster baby questions. there are also a few old comments from last week that i'd still love a response to!

src/pudl/analysis/allocate_net_gen.py
```diff
@@ -196,43 +197,169 @@
...
-    gf, bf, bga, gens, gen = extract_input_tables(pudl_out)
-    bf, gens_at_freq, gen = standardize_input_frequency(
-        bf, gens, gen, freq=pudl_out.freq
-    )
-
-    # add any startup energy source codes to the list of energy source codes
-    # fix MSW codes
+    bf, gens_at_freq, gen = standardize_input_frequency(bf, gens, gen, freq)
+    # Add any startup energy source codes to the list of energy source codes
+    # Fix MSW codes
```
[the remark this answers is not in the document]
zaneselvans: Oops, deleted.

src/pudl/analysis/allocate_net_gen.py
zaneselvans: I could put them in their own separate factories, but they all need to iterate over the same frequencies, and there are dependencies between the assets that are defined inside the factory, meaning you can only define some of them if the others are also being defined, so it seemed like they were a package deal and it was simpler to bundle them all together in their own little subgraph with one set that's defined for `MS` and another for `AS` frequencies.

src/pudl/analysis/allocate_net_gen.py
zaneselvans: I think it was late.

src/pudl/analysis/allocate_net_gen.py
zaneselvans: Ah whoops you're right. It would be more analogous to use the outside function. I'll update it.

src/pudl/metadata/resources/allocate_net_gen.py
zaneselvans: We're going with just the annual table in the DB until we've got monthly updates to EIA-923 and EIA-860M

src/pudl/metadata/resources/allocate_net_gen.py
zaneselvans: I'm not sure I understand what you're saying. Maybe we can chat?

src/pudl/analysis/allocate_net_gen.py
zaneselvans: Oh right I started out using `asset` but then realized that shadowed the dagster `@asset` decorator and I was frustrated.   I can de-ass it.

zaneselvans: I made it rated-G in my MCOE PR.

src/pudl/metadata/resources/allocate_net_gen.py
zaneselvans: I've added note into the table description at least. Added to the [naming revamp issue](https://github.com/catalyst-cooperative/pudl/issues/2533)

src/pudl/metadata/resources/allocate_net_gen.py
zaneselvans: Okay I've applied this change to these resources and also the ones in the MCOE PR.

src/pudl/analysis/allocate_net_gen.py
```diff
@@ -196,43 +197,157 @@
...
-            ``frac`` column and  net generataion reported in the original
-            ``generation_eia923`` and ``generation_fuel_eia923`` tables) that are
-            useful for debugging. Default is False, which will drop the columns.
+    The allocation process (via `allocate_net_gen_by_gen_esc()`) entails
+    generating a fraction for each record within a ``IDX_PM_ESC`` group. We
+    have two data points for generating this ratio: the net generation in the
+    generation_eia923 table and the capacity from the generators_eia860 table.
+    The end result is a `frac` column which is unique for each
+    generator/prime_mover/fuel record and is used to allocate the associated
+    net generation from the `generation_fuel_eia923` table.
```
bendnorman, line 347: Some Arg descriptions would be helpful here.

src/pudl/metadata/resources/allocate_net_gen.py
```diff
@@ -0,0 +1,111 @@
+"""Resource metadata for the allocate_net_gen tables."""
+from typing import Any
+
+AGG_FREQS = ["yearly", "monthly"]
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = (
+    {
+        f"generation_fuel_by_generator_energy_source_{freq}_eia923": {
+            "description": (
+                f"{freq.title()} estimated net generation and fuel consumption "
```
bendnorman, line 10: Should we describe what is estimated and how?

src/pudl/output/pudltabl.py
```diff
@@ -241,12 +236,15 @@ def _agg_table_name(self, table_name: str) -> str:
                 table_name = table_name.replace("_AGG", "")
         return table_name
 
-    def _get_table_from_db(self, table_name: str, resource: Resource) -> pd.DataFrame:
+    def _get_table_from_db(
+        self, table_name: str, resource: Resource, update: bool = False
```
bendnorman, line 240: Is it worth logging a warning here?

src/pudl/output/pudltabl.py
```diff
@@ -276,8 +274,9 @@ def _select_between_dates(self, table: str) -> sa.sql.expression.Select:
             ``report_date`` or ``report_year``) to lie between ``self.start_date`` and
             ``self.end_date`` (inclusive).
         """
-        pt = pudl.output.pudltabl.get_table_meta(self.pudl_engine)
-        tbl = pt[f"{table}"]
+        md = sa.MetaData()
+        md.reflect(self.pudl_engine)
+        tbl = md.tables[f"{table}"]
```
bendnorman, line 279: Should we catch a `KeyError` if the table exists? I don't think this check happens prior to this method being called.

test/validate/eia_test.py
```diff
@@ -44,7 +44,7 @@ def test_no_null_cols_eia(pudl_out_eia, live_dbs, cols, df_name):
         ("bga_eia860", 130_326, 130_326, 130_326),
         ("boil_eia860", 74_086, 74_086, 74_086),
         ("frc_eia923", 597_000, 244_415, 24_065),
-        ("gen_eia923", None, 5_171_497, 432_570),
+        ("gen_eia923", None, 4_872_421, 407_653),
```
bendnorman, line 47: Why did the number of records change?

zaneselvans: IIRC this (substantial) change is because I got rid of the last year of records, which had only null data in it since the only 2022 data we have is from EIA-860M, and these outputs depend on data tables from EIA-923.

src/pudl/output/pudltabl.py
zaneselvans: A warning about what?

src/pudl/analysis/allocate_net_gen.py
zaneselvans: Okay, added the args.

src/pudl/metadata/resources/allocate_net_gen.py
zaneselvans: Probably yes, but I think we should get that detail from @cmgosnell who is more familiar with the actual guts of the process.

src/pudl/output/pudltabl.py
zaneselvans: Do you mean if the table doesn't exist?  Is there a useful action we can take if we get a bad table name, or is it just going to error out anyway? This is only used internal to the PudlTabl class and we're going to nuke it soon and I think it should probably just fail if we have a bad table name.

src/pudl/metadata/resources/allocate_net_gen.py
cmgosnell: There is a tremendous amount of docs in the module. i think a lot of the module-level docs would be helpful to move over here to make it more visible to users.

zaneselvans: Do you know if the module level docs are in sync with what we're currently doing in the module?

src/pudl/metadata/resources/eia860.py
```diff
@@ -232,6 +232,9 @@
                     "generation_eia923",
                     "denorm_generation_eia923",
                     "denorm_generation_monthly_eia923",
+                    "generation_fuel_by_generator_energy_source_monthly_eia923",
+                    "generation_fuel_by_generator_monthly_eia923",
+                    # "generation_fuel_by_generator_energy_source_owner_monthly_eia923",
```
cmgosnell, line 237: should you take out all of these #commented out monthly tables we probs aren't going to make?

src/pudl/output/pudltabl.py
```diff
@@ -241,12 +236,15 @@ def _agg_table_name(self, table_name: str) -> str:
...
-    def _get_table_from_db(self, table_name: str, resource: Resource) -> pd.DataFrame:
+    def _get_table_from_db(
+        self, table_name: str, resource: Resource, update: bool = False
+    ) -> pd.DataFrame:
         """Grab output table from PUDL DB.
 
         Args:
             table_name: Name of table to get.
             resource: Resource metadata used to enforce schema on table.
+            update: Ignored. Retained for backwards compatibility only.
```
cmgosnell, line 247: what does this mean? does `Ignored` mean that it doesn't do anything? but is here bc there may be some things that used `update` in the past?

cmgosnell (approved):

src/pudl/analysis/allocate_net_gen.py
```diff
@@ -269,81 +393,58 @@ def allocate_gen_fuel_by_generator_energy_source(
...
+    # of the primary key for this table, so we have to drop them.
+    len_before = net_gen_fuel_alloc.shape[0]
+    net_gen_fuel_alloc = net_gen_fuel_alloc.dropna(subset=["prime_mover_code"])
+    len_after = net_gen_fuel_alloc.shape[0]
+    fraction_dropped = (len_before - len_after) / len_before
+    if fraction_dropped > 5e-5:
+        raise ValueError(
+            "Too many records were found to have a NULL prime_mover_code and "
+            f"dropped. Expected less than 5e-5, but found {fraction_dropped:.1%}."
+        )
```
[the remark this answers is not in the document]
cmgosnell: bc generators only ever have one prime mover, we could fill in the prime mover codes instead of dropping these.

src/pudl/output/pudltabl.py
bendnorman: Ah sorry that was vague. I think I meant adding a warning about `update` being ignored but retained for backwards compatibility.

src/pudl/output/pudltabl.py
bendnorman: I figured we could catch the `KeyError` and raise a helpful message like `f"{table} not found in the metadata."`

src/pudl/output/pudltabl.py
katie-lamb: Added logging warning.

src/pudl/output/pudltabl.py
katie-lamb: Added try-except here

src/pudl/analysis/allocate_net_gen.py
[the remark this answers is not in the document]
katie-lamb: @cmgosnell I don't feel like I have enough context to know what change to make for this in this PR.

src/pudl/metadata/resources/eia860.py
katie-lamb: Took out the commented out table.

src/pudl/metadata/resources/allocate_net_gen.py
katie-lamb: @cmgosnell Does the updated description I added make any sense? I tried to paraphrase the known drawbacks section as well.

katie-lamb:
@cmgosnell I tried to address all the lingering comments in this PR. The ones that maybe still need a look are:

- [What to do about this ValueError and filling in prime mover code?](https://github.com/catalyst-cooperative/pudl/pull/2527#discussion_r1197058829)
- [Does the metadata description of the tables make any sense?](https://github.com/catalyst-cooperative/pudl/pull/2527#discussion_r1187951010)

src/pudl/validate.py
```diff
@@ -172,7 +174,12 @@ def check_max_rows(df, expected_rows=np.inf, margin=0.05, df_name=""):
     return df
 
 
-def check_min_rows(df, expected_rows=0, margin=0.05, df_name=""):
+def check_min_rows(
+    df: pd.DataFrame,
+    expected_rows: int | float = np.inf,
```
zaneselvans, line 179: The default value should probably be zero here, not infinity.

test/validate/eia_test.py
```diff
@@ -110,83 +110,63 @@ def test_minmax_rows(
...
         (
             "bf_eia923",
-            (
-                Package.from_resource_ids()
-                .get_resource("boiler_fuel_eia923")
-                .schema.primary_key
-            ),
+            Package.from_resource_ids()
+            .get_resource("boiler_fuel_eia923")
+            .schema.primary_key,
```
zaneselvans, line 115: It probably doesn't matter much in terms of speed, but I think that these could all be replaced with e.g.

```py
Resource.from_id("boiler_fuel_eia923").schema.primary_key
```

src/pudl/analysis/allocate_net_gen.py
zaneselvans, line 1: Given that this module is now allocating generation and fuel and not just net generation, and that the assets it produces are named `generation_fuel_by...` might it be appropriate to change all of the `allocate_net_gen` references to `allocate_generation_fuel` or `allocate_gen_fuel` or something similar for clarity?

src/pudl/metadata/resources/allocate_net_gen.py
```diff
@@ -0,0 +1,139 @@
...
+                f"{freq.title()} estimated net generation by generator. Based on net "
+                "generation reported in the EIA-923 generation and generation_fuel tables."
+                "The net electricity generation and fuel consumption reported in the EIA-923 "
+                "generation fuel are allocated to individual generators. This process does not "
+                "distinguish between primary and secondary energy_sources for generators. Net "
+                "generation is allocated equally between energy source codes, so if a "
+                "plant has multiple generators with the same prime_mover_code but different "
+                "energy source codes the generation_fuel_eia923 records will be associated "
+                "similarly between these two generators. Allocated net generation will still "
+                "be proportional to each generator's net generation or capacity."
```
zaneselvans, line 61: It's unclear to me from this description why the energy source code is involved or matters as far as the net generation allocation process. I think maybe we've forgotten to mention that the table reports fuel consumption as well as net generation?

src/pudl/metadata/resources/allocate_net_gen.py
```diff
@@ -0,0 +1,139 @@
...
+            "sources": ["eia923", "eia860"],
+            "etl_group": "outputs",
+        }
+        for freq in AGG_FREQS
+    }
+    | {
+        "generation_fuel_by_generator_energy_source_owner_yearly_eia923": {
+            "description": (
+                "Yearly estimated net generation and fuel consumption for each generator, "
+                "broken down by energy source, prime mover, and owner. Based on data "
```
zaneselvans, line 94: It's unclear from this description whether we are aggregating to individual prime movers, or within all prime movers of a given type within a plant.

It might make sense to pull out the text which is shared between all of these longer generation fuel descriptions into a module-level constant so that it can be re-used and edited without needing to be changed in several places.

src/pudl/analysis/allocate_net_gen.py
```diff
@@ -196,43 +197,166 @@
 """
 
 
-# Two top-level functions (allocate & aggregate)
+def allocate_net_gen_asset_factory(
+    freq: Literal["AS", "MS"],
+    io_manager_key: str | None = None,
+) -> list[AssetsDefinition]:
+    """Build net gen allocation assets at yearly and monthly frequencies."""
```
zaneselvans, line 204: Maybe include fuel consumption as well as net generation in the description here.  I think we might want to change all of the references to `allocate_net_gen` to `allocate_gen_fuel` or something similar, since I think we're always allocating both fuel and generation now.

src/pudl/analysis/allocate_net_gen.py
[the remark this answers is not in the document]
zaneselvans: There are a handful of records in the allocated dataframe that have NA values for their prime mover codes.  However, we can look up the prime mover code for the generator in that year in the `gens_eia860` dataframe and use that to fill in these few missing values, which means we don't have to lose a little bit of the generation / fuel consumption in the allocation/aggregation process (allowing us to conserve the overall amount of generation and fuel for later validation).

Where's the right place to fill in these missing values?  It seems like it might be earlier in the overall ETL process, like post-harvesting, at which point we have the entity / annual tables available for use in filling in the gen/fuel data table missing PM values.  That way the complete PM values would be available everywhere -- or is this only something that applies after the allocation/aggregation has been done since in the original tables there are no generator IDs to use for looking the PM code up?  In which case doing the PM code filling only after you've gotten the `generator_id` values would make sense.

src/pudl/analysis/allocate_net_gen.py
```diff
@@ -429,41 +528,46 @@ def standardize_input_frequency(
...
+            left=net_gen_fuel_alloc,
+            right=gens[IDX_GENS + ["utility_id_eia", "capacity_mw"]],
+            left_date_col="report_date",
+            right_date_col="report_date",
+            new_date_col="report_date",
+            on=["plant_id_eia", "generator_id"],
+            date_on=["year"],
             how="left",
+            report_at_start=True,
+            validate="m:1",
```
zaneselvans, line 565: This seems like a weird entanglement with downstream analysis / PPE tooling. I guess `scale_by_ownership()` is a static method? Should it be pulled out of `MakeMegaGensTbl()` for re-use in other contexts?  Maybe this is something that's dealt with in the MCOE / PPE dagsterization PRs?

src/pudl/validate.py
katie-lamb: Good catch, this seems like a mistake that never got changed back. Making it 0 now.

src/pudl/analysis/allocate_net_gen.py
katie-lamb: As far as I can tell, this is the only time `MakeMegaGensTbl()` is used outside of the context of the PPE. @cmgosnell is this still a useful function? I wasn't sure how to deal with this in the PPE PR since mega gens is downstream of this.

zaneselvans: IIRC the ownership scaling was just a convenient piece of code to re-use that happened to be embedded within a class that had been defined for another purpose.

But I think the generation-fuel scaled by ownership is (currently) only used in the context of the PPE.  Do we think it's a generally useful / interesting output that should be published alongside the other generation fuel outputs? Or is it really just an input into the PPE? @cmgosnell

src/pudl/metadata/resources/allocate_net_gen.py
katie-lamb: Took out the part about energy source codes. I think that doesn't apply here?

src/pudl/analysis/allocate_net_gen.py
katie-lamb: Seems like this generation and fuel table by owner table can just be taken out?

src/pudl/analysis/allocate_net_gen.py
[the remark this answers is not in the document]
katie-lamb: @zaneselvans The `denorm_generation_fuel_combined_eia923` table where the PM code for the allocations comes from doesn't have `generator_id` (it is in `denorm_generation_eia923`). So it makes sense to fill the PM code after you've joined together generation fuel and generation. 

Should I bring in `generators_eia860` in this allocation process to fill in the missing PMs after the generation fuel and generation has been allocated and merged together? Or a different table?

katie-lamb:
@zaneselvans I believe I addressed all of your comments except these two remaining questions:
- [fill in PM code in this allocation process or not](https://github.com/catalyst-cooperative/pudl/pull/2527#discussion_r1256446437)
- [get rid of the generation and fuel by ownership table](https://github.com/catalyst-cooperative/pudl/pull/2527#discussion_r1256436536)

zaneselvans:
Great!  I think both of the remaining questions you've highlighted are probably best directed to @cmgosnell. She felt strongly that we should keep the sums as invariants since it's an easy thing to check for internal consistency of the allocation process.  And I don't know what the potential (or existing?) additional applications of the ownership allocation are.

src/pudl/analysis/allocate_net_gen.py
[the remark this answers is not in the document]
cmgosnell: for starters, my initial concern doesn't feel like a blocking concern overall. nonetheless this does feel like a good thing to address. if we were to fill in the pm-codes i could see two decent options:
* **at harvesting** we could turn down how strict we are for this variable. right now we have a little function `pudl.transform.eia._manage_strictness` which... manages the strictness 😎  needed for each column we are harvesting. the default is that a record needs to be 70% consistent. There are a few names columns already in here that we turned the strictness of the harvesting to 0% - which is to say that the harvesting process grabs the **most consistent** option found even when it wasn't a plurality of the options found. For the annually-varying/not-widely-reported attributes, this consistency check can be a little weird bc there could be only two or three options in an entity/year. It's possible that turning down the strictness even by a little would result in no-nulls (which we could test somewhere/somehow).
* **some version of bfill in `denorm_generators_eia860`** we have several versions of back-filling/filling in gaps in the output layer mostly in `pudl.output.eia` (ex: `fill_in_missing_ba_codes`, `fill_generator_technology_description`). 

imo the harvesting option would be very easy to implement, with the maybe difficult part being determining where/when to add a "no-nulls in this column" check. the output layer version would be more... like human logic defensible something something if we thought hard about what we think the standard expectation should be. imo a lot of these types of error crop up in years of transition so deciding what is "right" might actually be challenging to do across the board (i.e. whether to backfill or forwardfill). I was initially more inclined to the output option but we could choose wrong so I think turning down the strictness and just getting the most frequently reported value is probably easier and just as likely to produce correct results.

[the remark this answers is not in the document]
cmgosnell: (this is all assuming the null pm codes are coming directly from the harvesting process not somewhere else...)

src/pudl/analysis/allocate_net_gen.py
cmgosnell: this isn't an input into the PPE. I don't think it feeds into anything else in PUDL rn. it uses a PPE method. which perhaps makes a case that this method is more broadly useful and shouldn't be in PPE land because I've also used this method in other rmi tasks. i don't think the use of this method in a different context should block moving forward on dagster stuff. it's been used over here for... idk years at this point. but moving it out into helpers or something would maybe be good as well.

this output was originally asked for by rmi folks and has subsequently been used by greg m as well iirc.

katie-lamb:
I did the following:

* Move scale_by_ownership out of the plant_parts_eia/GensMega analysis to remove ambiguity and stick it in helpers.
* At harvesting I turned down the strictness of `prime_mover_code` to 0% in `pudl.transform.eia._manage_strictness` and now there are no-nulls

PR #2536, opened 2023-04-19
Initial refactoring of dbf extraction process

rousik:
# PR Overview

This is the first step of a code refactoring that should generalize the process of pulling data from DBF (FoxPro Database) and into sqlite for further processing.

As part of this, the existing methods and functionality was wrapped in `FoxProDatastore` class which generalizes the early-stage handling of fox pro databases, including:
1. listing of available tables and their schemas
2. mapping table schema to SQLAlchemy table schemas
3. reading of dataframes from series of years worth of data

This datastore is then used by `FoxProExtractor` class which will:
1. initialize database schema
2. load data from dbf files

Additionally, this extractor offers several methods that can be overriden in order to inject dataset specific logic. At this point, there are three options:
1. finalize_schema for manipulating the schema (e.g. adding primary/foreign key constraints to tables)
2. transform_table for manipulating dataframe before it's loaded into sqlite
3. postprocess for database-level manipulation of data after it's been written to sqlite (this can be used e.g. to identify missing respondents)

This is still an early stage concept PR and the following is missing:
- [ ] appropriate documentation for all methods and the overall structure
- [ ] unit tests (unsure how to best unit test this b.c. it relies on actual data)
- [ ] appropriate organization of concept into files (e.g. `dbf.py` and `ferc1.py` may need to be reshuffled)

# PR Checklist


src/pudl/extract/dbf.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def get_dataset(self) -> str:
+        """Returns name of the dataset that this datastore provides access to."""
+        ...
+
+    def get_table_names(self) -> list[str]:
+        """Returns list of all available table names."""
+        ...
+
+    def get_table_schema(self, table_name: str, year: int) -> TableSchema:
```
zaneselvans, line 71: Curious why `get_table_schema` is part of the interface, and `get_db_schema` isn't.

src/pudl/extract/ferc1.py
```diff
@@ -213,429 +208,272 @@
...
-            sa.ForeignKeyConstraint(
-                columns=[
-                    "respondent_id",
-                ],
-                refcolumns=["f1_respondent_id.respondent_id"],
-            )
+        This wraps the generic base_datastore and constructs instance of
+        FercFoxProDatastore.
+        """
+        # TODO(rousik): mapping from dataset name to dbc_filename may be moved to constants or kept somerwhere with the rest
```
zaneselvans, line 359: It seems like `F1_PUB.DBC` and its equivalents for the other data sources should definitely be in the `package_data` that maps out the files & directories, as we've done with the Excel spreadsheets. If the root directory that we need to look for all the DBF files in is always defined by wherever the DBC file is, then maybe we just store the path to the DBC path, and drop the filename to get the directory path (which I think is being stored independently now).

src/pudl/extract/dbf.py
```diff
@@ -0,0 +1,301 @@
...
+"""
+
+# TODO(rousik): circular dependency between FercFoxProDatastore and FoxProTable,
+# could be addressed through interfaces/ABCs
+
+
+# TODO(rousik): instead of using class-level constants, we could pass the params in the constructor, which should
+# allow us to instantiate these dataset-specific datastores in the extractor code.
+# That may make the manipulations little easier.
+class FercFoxProDatastore:
```
zaneselvans, line 139: This class **has** a Datastore, but it is not itself a Datastore, which I think is different from how we've managed the other data source specific Datastores, and might be confusing. What are the trade-offs / reasons for doing it this way rather than inheriting?

src/pudl/extract/ferc1.py
```diff
@@ -213,429 +208,272 @@
...
-    "O": "XXX",  # Double, 8 bytes
-    "G": "XXX",  # OLE 10 digit/byte number of a .DBT block, stored as string
-}
-"""dict: A mapping of DBF field types to SQLAlchemy Column types.
 
-This dictionary maps the strings which are used to denote field types in the DBF objects
-to the corresponding generic SQLAlchemy Column types: These definitions come from a
-combination of the dbfread example program dbf2sqlite and this DBF file format
-documentation page: http://www.dbase.com/KnowledgeBase/int/db7_file_fmt.htm
+class FoxProExtractor:
```
zaneselvans, line 212: Seems like `FoxProExtractor` should probably go in the `dbf.py` module?

src/pudl/extract/ferc1.py
```diff
@@ -213,429 +208,272 @@
...
-}
-"""Missing FERC 1 Respondent IDs for which we have identified the respondent."""
+    Dataset specific logic and transformations can be injected by overriding:
+    1. finalize_schema() in order to modify sqlite schema. This is called just before the
+       schema is written into the sqlite database. This is good place for adding primary and/or
+       foreign key constraints to tables.
+    2. transform_table(table_name, df) will be invoked after dataframe is loaded from the foxpro
+       database and before it's written to sqlite. This is good place for table-specific
+       preprocessing and/or cleanup.
+    3. postprocess() is called after data is written to sqlite. This can be used for database
```
zaneselvans, line 228: `postprocess()` would be database-level in the case of the missing `respondent_id` values because we have to actually scan the whole DB to know which IDs are missing, right?

src/pudl/extract/ferc1.py
```diff
@@ -213,429 +208,272 @@
...
-                )
+        for tn in self.datastore.get_table_names():
+            self.datastore.get_table_schema(tn, refyear).to_sqlite_table(
+                self.sqlite_meta
             )
-    return observed
+        self.finalize_schema(self.sqlite_meta)
+        self.sqlite_meta.create_all(self.sqlite_engine)
 
+    def transform_table(self, table_name: str, in_df: pd.DataFrame) -> pd.DataFrame:
```
zaneselvans, line 301: As this is set up right now, would we need to have a big if/elsif/else block to do the right thing for each table, if the operations weren't uniform across all tables?

src/pudl/extract/ferc1.py
```diff
@@ -655,112 +493,20 @@ def get_raw_df(
...
-        logger.info(f"SQLite: loading {n_recs} rows into {table}.")
-        new_df.to_sql(
-            table,
-            sqlite_engine,
-            if_exists="append",
-            chunksize=100000,
-            dtype=coltypes,
-            index=False,
-        )
+    # TODO(rousik): this thin wrapper seems to be somewhat quirky. Maybe there's a way to make the integration
```
zaneselvans, line 496: Dagster seems to be mostly organized around functions, so I don't know if there's a way to avoid having this class-to-function wrapper. The way the FERC DBF to SQLite extraction happens in the DAG is also a bit different -- with the entire extraction showing up as a standalone job that runs before and separate from all of the downstream operations in the main PUDL ETL. If anyone has thoughts on how to organize this appropriately I suspect it's @bendnorman.

src/pudl/extract/ferc1.py
```diff
@@ -213,429 +208,272 @@
 }
 """A mapping of PUDL DB table names to their XBRL and DBF source table names."""
```
zaneselvans, line 209: `TABLE_NAME_MAP` seems like another thing that we might want to relegate to `package_data`

zaneselvans (commented):
Hey sorry I didn't get to this until today. I was out all last week with visitors.

I think this looks conceptually simpler than what we had before, and seems like it should work for the `ferc2`, `ferc6` and `ferc60` data. I left some questions / thoughts.

zaneselvans:
As far as adding "unit tests" in the context of building something that's pretty bespoke to the particular data that it's supposed to be consuming, we've definitely fallen back to small integration tests, that use just a little bit of the real data. Since it's easy to pull that data in, storing some expected outcomes for the extraction of a couple of `ferc2` / `ferc6` tables and checking that that's what we get is probably good enough?

src/pudl/extract/dbf.py
```diff
@@ -0,0 +1,301 @@
...
+from typing import Any, Protocol
+
+import pandas as pd
+import sqlalchemy as sa
+from dbfread import DBF, FieldParser
+
+from pudl.workspace.datastore import Datastore
+
+
+class TableSchema:
```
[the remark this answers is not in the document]
rousik: Renamed this to `FoxProTableSchema` for clarity. It seems to me that there's a bit of vagueness in the use of dbf vs foxpro and I did not make it much better here. Do we have preference for one or the other?

src/pudl/extract/dbf.py
```diff
@@ -0,0 +1,301 @@
...
+    def get_columns(self) -> Iterator[tuple[str, sa.types.TypeEngine]]:
+        """Itereates over the (column_name, column_type) pairs."""
+        for col_name in self._columns:
+            yield (col_name, self._column_types[col_name])
+
+    def get_column_rename_map(self) -> dict[str, str]:
+        """Returns dictionary that maps from short to long column names."""
+        return dict(self._short_name_map)
+
+    def to_sqlite_table(self, sqlite_meta: sa.MetaData) -> sa.Table:
```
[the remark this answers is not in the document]
rousik: This seems generic enough. Renamed accordingly.

src/pudl/extract/dbf.py
rousik: `get_table_schema` is used by the `FoxProExtractor`, while `get_db_schema` is only used internally as part of dataframe loading.

src/pudl/extract/ferc1.py
rousik: 🤔 yeah, this sounds like the good way to go about that.

src/pudl/extract/dbf.py
rousik: As it is now, `FercFoxProDatastore` effectively wraps generic datastore and offers very different higher level interface for reading data and table schemas. As such, it doesn't really make sense to inherit, because we're redefining the offered functionality. If the naming is confusing, we could perhaps consider not naming it `Datastore`. How about `FercFoxProReader` or something like that?

src/pudl/extract/ferc1.py
rousik: Yes, indeed.

src/pudl/extract/ferc1.py
rousik: Correct.

src/pudl/extract/ferc1.py
rousik: Right now yes. I haven't really gotten to it yet, but I think this should be replaced with a map of table transformation `dict[str,Callable[[pd.DataFrame],pd.DataFrame]]`, where for each of these extractor, you can simply register all the transformers for each table, or possibly use reflections and find all methods with names such as `transform_foo` as a transformation for the table `foo`.

I was planning on refactoring this when it becomes more necessary/relevant (i.e. when there are more transformations needed)

src/pudl/extract/ferc1.py
rousik: Yep. This might require some rework of both dbf and xbrl extractors.

src/pudl/extract/ferc1.py
rousik: It might not be easy to eliminate all of boiler-plate. It seems a bit unwieldy, but it's a minor issue. I can remove this todo and leave as-is.

src/pudl/extract/ferc1.py
rousik: I think this is now done. Marking as resolved.

src/pudl/extract/dbf.py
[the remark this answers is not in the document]
zaneselvans: Thus far we've mostly organized the extraction tools by file format (XBRL, Excel, DBF) and I think that the [DBF / dbase format](https://en.wikipedia.org/wiki/.dbf) is more generic than the [Microsoft VisualFoxPro](https://en.wikipedia.org/wiki/Visual_FoxPro) ecosystem (though that is the one that these files came from) so I think probably going with DBF is likely the best option.

src/pudl/extract/dbf.py
zaneselvans: I think that non-Datastore naming would be less confusing, given the number of other Datastore subclasses we have flying around. Maybe `FercDbfReader`?

src/pudl/extract/ferc1.py
zaneselvans: Okay that sounds totally reasonable. We're trying to minimize the changes that happen in this step anyway (and deal with all the real transformations when we go from the cloned FERC DB into the main PUDL DB) so hopefully there won't actually very many cases to manage in here. Maybe even just one, for the `respondent_id` table.

src/pudl/extract/ferc1.py
zaneselvans: Okay, not a high priority for the moment I think. Just something to add to the cleanup list.

zaneselvans (changes requested):
I think just some minor class renames to make it clear that these are FERC DBF specific structures, and that we're not inheriting from the Datastore class. Excited to see this applied to FERC 2 and FERC 6!

src/pudl/extract/ferc1.py
rousik: I wonder if this rework could be done independently of this initial effort. WDYT?

src/pudl/extract/dbf.py
```diff
@@ -0,0 +1,450 @@
...
+        Args:
+            sa_meta: new table will be written to this MetaData object.
+        """
+        table = sa.Table(self.name, sa_meta)
+        for col_name, col_type in self.get_columns():
+            table.append_column(sa.Column(col_name, col_type))
+        return table
+
+
+class AbstractFoxProDatastore(Protocol):
```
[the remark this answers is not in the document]
rousik: Yeah, this sounds reasonable. I will make it so.

src/pudl/extract/dbf.py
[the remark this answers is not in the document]
rousik: Done & done.

src/pudl/extract/dbf.py
```diff
@@ -0,0 +1,452 @@
...
+class FercDbfExtractor:
+    """Generalized class for loading data from foxpro databases into SQLAlchemy.
+
+    When subclassing from this generic extractor, one should implement dataset specific
+    logic in the following manner:
+
+    1. set DATABASE_NAME. This is going to be used as the file for the resulting sqlite
+    database.
+    2. Override get_datastore() method to return the right kind of dataset specific
+    datastore.
```
zaneselvans, line 324: It seems like there's not default implementation of `get_datastore()` to override. Is there supposed to be? Or is this a method that always has to be defined by child classes? It gets called in `__init__()`  By default should it just set  `self.datastore` to whatever datastore was passed in as an argument?

test/conftest.py
```diff
@@ -404,8 +405,9 @@ def pudl_settings_dict(request, pudl_input_output_dirs):  # noqa: C901
 
 @pytest.fixture(scope="session")  # noqa: C901
 def ferc1_dbf_datastore_fixture(pudl_datastore_fixture):
```
zaneselvans, line 407: Is this fixture still required, now that we're constructing the `Ferc1DbfReader` inside of `test_ferc1_schema`?

zaneselvans (commented):
I edited some docstrings to make them consistent with the new class names.

* There didn't seem to be a default implementation of `get_datastore()` which seemed like maybe it was wrong.
* Do we still need the dbf datastore fixture in `conftest.py`?

src/pudl/extract/dbf.py
rousik: I suppose we could make default implementation that throws NotImplemented exception. Given the renamings, this also needs to be renamed to `get_dbf_reader()`.

zaneselvans: Yeah that rename sounds good. So this is kind of an abstract class that lays out some infrastructure but has stuff that needs implementation.

test/conftest.py
rousik: I thought I have removed this, but apparently it was left behind. Now it's gone again.

rousik:
Ok, all comments should be addressed now and this should be good for another round of reviews.

test/conftest.py
zaneselvans: It looks like it's still here. Is there a commit that still needs to be pushed?

zaneselvans (approved):
Looks good assuming the CI all passes!

rousik:
Ran `ferc_to_sqlite` with one year, few years and full set of years and sqlite databases are the same. Merging.

Issue #2437, opened 2023-03-22
Convert EIA-861 + FERC-714 service territory and state electricity demand to Dagster assets


zaneselvans:
Notes on how to do this:
- Interim table transforms in `pudl.output.ferc714` should use the standard io manager
- `compile_geoms()` should have a factory function, iterating over entity type. The save to geoparquet function has been flaky, if we are able to get it working we could use geoparquet and write a geoparquet io manager. Otherwise, we can strip geometry and  save as a table. Extensive additional work on the geometrical component is a low priority in this iteration of the transition.
- Whatever is implemented in `state_demand.py` should be used as a default for the `ba_ids` and `util_ids`, but these should be parameters passed in by dagster's `context` to enable manual changes.
- Best order to implement change in: service_territory -> FERC 714 -> state_demand
- Rip out the `Respondents` class in `pudl.output.ferc714`
- In `pudl.output.ferc714`, the `summarize_demand` and `fipsify` tables are the best candidates for inclusion into PUDL. The other two tables are just upstream intermediate outputs.

```[tasklist]
- [x] `pudl.output.ferc714.Respondents.annualize`
- [x] `pudl.output.ferc714.Respondents.categorize`
- [x] `pudl.output.ferc714.Respondents.summarize_demand`
- [x] `pudl.output.ferc714.Respondents.fipsify`
- [x] `pudl.output.ferc714.Respondents.georef_counties`
- [x] `pudl.output.ferc714.Respondents.georef_respondents`
- [x] `pudl.analysis.service_territory.compile_geoms(entity_type="util")`
- [x] `pudl.analysis.service_territory.compile_geoms(entity_type="ba")`
- [x] `pudl.analysis.state_demand.predict_state_hourly_demand`
```

e-belfer:
Merged in #2550

Issue #2427, opened 2023-03-21
Add data validations for finished EIA-861 and FERC-714 tables


zaneselvans:
We've got a bunch of new tables in the DB! We should figure out some criteria for validating them. We should restrict this to only the close-to-finished tables that we've actually done a lot of work on. The other `raw_` and `clean_` tables aren't expected to be valid yet.

Validations to include:
* expected number of records
* ???

```[tasklist]
- [ ] `balancing_authority_assn_eia861`
- [ ] `balancing_authority_eia861`
- [ ] `demand_hourly_pa_ferc714`
- [ ] `respondent_id_ferc714`
- [ ] `sales_eia861`
- [ ] `service_territory_eia861`
- [ ] `utility_assn_eia861`
```

Wilson-Energy:
Not sure if this is the right place for this, but here are some errors in FERC 714 data, table planning_area_hourly_demand_and_forecast_summer_and_winter_peak_demand_and_annual_net_energy_for_load_table_03_2_duration

1. Oglethorpe (C003561) entered "23" rather than "2023" (etc.) for planning_area_hourly_demand_and_forecast_year
2. Gulf (C001554) has a stray entry in row 288
3. PJM (C000030) and NorthWestern (C001789) each has one extra 2033 entry for the 2022 report - this may be ok to retain but it is nonstandard
4. Avista (C000379) has extra 2033-2045 entries for the 2022 report - overachiever!
5. Southern (C001556) files "0" entries on this form - its data are filed by subsidiaries GPC, APC and MPC

zaneselvans:
Thank you for compiling these wierdnesses @Wilson-Energy!

zaneselvans:
Closing this since we've long since added row-counts and other basic data validations for these tables. Though of course there's always room for more!

Issue #2412, opened 2023-03-17
Bring Census DP1 database into Dagster


zaneselvans:
We use the Census DP1 geodatabase for spatial analyses (e.g. turning FIPS codes into county geometries) and as a source for demographic information (e.g. population by census tract).

It seems like it may have slipped through the cracks in the Dagster migration? The ad-hoc "ETL" is almost a one-liner, converting the published geodatabase into SQLite it's in [pudl.convert.censusdp1tract_to_sqlite.py](https://github.com/catalyst-cooperative/pudl/blob/dagster-asset-etl/src/pudl/convert/censusdp1tract_to_sqlite.py)

The database only has 3-4 tables in it (state, county, and tract-level demographic data). Should it be an asset? A multi-asset? A standalone resource like the FERC DBs?

I'm a little surprised that this didn't break anything in the tests or nightly builds. We currently distributed this converted DB along with the other SQLite DBs if the nightly builds succeed.

```[tasklist]
### Scope
- [ ] census1dbtract.sqlite db is explicitly generated by dagster, as opposed to by ferc714_out fixture
```

```[tasklist]
### Next steps
- [ ] wrap pudl.convert.censusdp1tract_to_sqlite.censusdp1tract_to_sqlite in a @multi_asset
- [ ] new IOManager for censusdp1tract DB
```

bendnorman:
The [most recent `dagster-asset-etl` nightly](https://console.cloud.google.com/storage/browser/nightly-build-outputs.catalyst.coop/87c27052dee0fa8404d4f4a8b577090c624c61d5-dagster-asset-etl;tab=objects?prefix=&forceOnObjectsSortingFiltering=false) build included the census1dbtract.sqlite db. The `gcp_pudl_et.sh` on a dev and dagster-asset-etl doesn't create the db using the `censusdp1tract_to_sqlite` command. Instead, the database is created by the `ferc714_out` fixture.

We should probably dagsterify the census ETL soon but given the db is still being created by the nightly builds on the dagster-asset-etl branch I don't think this is blocking #2104.

bendnorman:
As for how we would dagsterify the ETL, I think we could wrap `pudl.convert.censusdp1tract_to_sqlite.censusdp1tract_to_sqlite` in a @multi_asset and use and create a new io manager the points to the census db:

```python
@multi_asset(
    outs={
        table_name: AssetOut(io_manager_key="census_sqlite_io_manager")
[...]

Issue #2517, opened 2023-04-10
Create design doc for naming convention of PUDL tables


bendnorman:
It sounds like users favor ripping out `PudlTabl` ASAP and accessing the table directly from the database #2503. Given we won't have `PudlTabl` method to act as aliases for the actual table names, we need to decide on a naming convention for our denormalized tables. We'll likely have to update the normalized table names because the ideal denormalized table names might conflict with the existing normalized table names. This is a scary change!

```[tasklist]
### Scope
- [x] what the naming conventions are
- [x] why we decided on them
- [x] how we will implement/roll out the changes
```

```[tasklist]
### Tasks
- [x] draft a proposal document that covers: "what is the problem we're trying to solve? how are we solving it? why did we choose this way? what are some other approaches we considered and didn't do?"
- [x] circulate for feedback: what worries you about this approach? what excites you?
- [x] incorporate feedback / adapt plan
- [x] get sign-off from whole team
- [ ] Create a github discussion that summarizes the design doc
```

## Questions
- Does our existing data modeling mirror a common pattern? Can we adopt the naming conventions associated with the common data modeling pattern?
- If we aren't following a common pattern, do we want to make modeling changes to follow a common modeling pattern? This  is potentially a V big change but potentially worth doing. I think if we use a common data modeling pattern, we'll be able to rely on existing concepts, discussions and resources when answering future modeling questions.

e-belfer:
Some of the discussion in #2275 should be relevant here. Right now IIRC the normalized tables are the only tables which don't have some kind of qualifying suffix (e.g. the referent `boiler_fuel_eia923` table is the normalized one). 

In terms of updating naming conventions, the name should convey the following four pieces of information:
1. Table name: should be obvious to users familiar with FERC/EIA data what its analogue in the raw data is, if there is one. If not, theme of content should be relatively clear from the name (e.g. `plant_parts_eia`).
2. Data source: making it clear what source the data originates from (e.g. EIA 860, or EIA if combining data from multiple forms)
3. Stage of processing: raw, clean, normalized, denormalized
4. Any other modifications, if we decide to save multiple variations of a table (e.g. imputed vs. not, annual vs monthly).

While doing this, we should aim for the following:
- Removing or standardizing any abbreviations (e.g. gens, utils)
[...]

PR #2550, opened 2023-04-27
Convert EIA-861 and FERC 714 service territory outputs to Dagster assets

e-belfer:
This PR migrates the `pudl.analysis.service_territory`, `pudl.output.ferc714` and `pudl.analysis.state_demand` modules into dagster, producing 5 new assets in PUDL (see issue #2437). 

Some key design decisions:
- Drop the geometry column of the `compiled_geometry` assets to avoid loading 15GB of spatial data into the dataframe.
- Update CLI functions to read directly from dagster caches and PUDL tables, or remove CLI functionality where this is not possible.
- Group some functions in `state_demand` to reduce the number of interim assets.
- In #2649, enable FERC 714 data to be filtered by year in order to enable this PR to run without memory issues
- In #2412, update census data inputs. This PR is branched off of the census PR in order to be able to run correctly.

Still to do for both:
- [x] Update release notes
- [x] Remove all unnecessary intermediate outputs

Out of scope:
- [ ] Fix existing `save_geoparquet()` method in `pudl.analysis.service_territory`
- [ ] Write a geoparquet io manager and save outputs as geoparquet files

Particular questions for the reviewer:
- Do asset names make sense?
- Are field descriptions accurate for new fields?
- Is the level of testing adequate? Should we write additional tests or consider this a part of #2427?
- We're adding 15-20min of processing time to the ETL (time subsetting isn't applied to 861 ETL processes currently). Is this acceptable? Should we do more substantial code revisions here to chip away at this?

# PR Checklist


jdangerx:
@e-belfer your PR description was very helpful!

running/disorganized notes from trying to orient myself to this:

What is this thing?

* basically changes the analysis from reading/writing from the pudl_out object to reading/writing to SQLite. Sweet.
* also some docs changes while she's at it.
* adds the service territories schema to the database.
* includes a small addition to the 714 stuff, maybe an accident since it's not mentioned in the PR description?

Questions:

* tests? Do we have tests for this? how would we write tests for this?
* wow, the annualized 714 stuff just needs to get piped to add_dates? cool - should look at how that works
* do I need to write the geoparquet IO manager? why or why not?
* dig into the whole "dropping geometry columns" thing.
* how to fix?
  ```
  FAILED test/integration/etl_test.py::test_pudl_engine - pudl.io_managers.ForeignKeyErrors: Foreign key error for table: compiled_geometry_utility_eia861 -- utilities_eia860 (utility_id_eia, report_date) -- on rows [     1      2      3 ... 237870 237871 237872]
  ```

zaneselvans:
* The FK failure is because there are a bajillion utilities included in the EIA-861 data that do not show up anywhere in EIA-860 or EIA-923, and we are not yet harvesting all those IDs for the utility entity table, so that relationship should be excluded for now -- this is being done for many of the EIA-861 tables already, if you take a look at the resource definitions.
* IIRC the spatial outputs (geodataframes, goeparquet files, or hypothetical SpatiaLite DB outputs) are not currently working, and we don't know why -- generating them blows up memory usage. It would be great to get them working again, and output geoparquet, but I *think* that's beyond the scope of this transition to using Dagster.  Thankfully all of the "spatial" outputs we're generating can be done without using the actual geometries, since states, counties, and census tracts all have unique FIPS IDs, and those are the only spatial entities we are referencing.
* There are unit tests for some of the analytical subcomponents, but I think we were primarily doing "Can this thing even run?" integrating testing previously. Maybe also some expected row counts.

e-belfer:
At present, there are three major roadblocks remaining to merging this PR.

1. #2412 is likely a blocking issue for this PR, as this PR removes the `ferc714_out` fixture that was running the census ETL.
2. Memory issues cause regular failures locally for me, and reducing peak memory usage might be a blocking issue here.
3. Adequate minimal testing infrastructure for these new outputs (see below for more details). What else should be added here in this PR?

The current state of testing on this branch is quite minimal, and is as follows:

Integration:
- 861 and 714 outputs currently get checked for no null columns, and not being an empty dataframe
- State demand output gets checked for not being an empty dataframe

Validation:
- Expected row checks for all new assets

src/pudl/analysis/state_demand.py
```diff
@@ -266,13 +265,20 @@ def load_ventyx_hourly_state_demand(path: str) -> pd.DataFrame:
...
+        "raw_hourly_demand_matrix_ferc714": AssetOut(),
+        "utc_offset_ferc714": AssetOut(),
+    },
+)
+def load_hourly_demand_matrix_ferc714(
+    demand_hourly_pa_ferc714: pd.DataFrame,
 ) -> tuple[pd.DataFrame, pd.DataFrame]:
     """Read and format FERC 714 hourly demand into matrix form.
 
     Args:
```
e-belfer, line 280: Outdated and should get removed

src/pudl/metadata/fields.py
```diff
@@ -1779,7 +1794,7 @@
     },
     "particulate_control_id_eia": {
         "type": "string",
-        "description": "Particulate matter control identification number. This ID is not a unique identifier.",
+        "description": "Alphanumeric particulate matter control ID.",
```
e-belfer, line 1797: Check why this is in this PR!

src/pudl/analysis/state_demand.py
```diff
@@ -266,15 +265,18 @@ def load_ventyx_hourly_state_demand(path: str) -> pd.DataFrame:
 # --- Datasets: FERC 714 hourly demand --- #
 
 
-def load_ferc714_hourly_demand_matrix(
-    pudl_out: pudl.output.pudltabl.PudlTabl,
+@multi_asset(
+    compute_kind="Python",
+    outs={
+        "raw_hourly_demand_matrix_ferc714": AssetOut(),
```
e-belfer, line 271: Might be more optimal to make `utc_offset_ferc714` an asset that takes in `raw_hourly_demand_matrix_ferc714`, rather than use a multi-asset. Thoughts?

src/pudl/output/ferc714.py
```diff
@@ -196,455 +194,437 @@ def categorize_eia_code(eia_codes, ba_ids, util_ids, priority="balancing_authori
...
+                mask |= is_child
+    return pd.concat([df[~mask], pd.concat(tables)]).drop_duplicates()
+
+
+@asset(compute_kind="Python")
+def filled_service_territory_eia861(
+    filled_balancing_authority_assn_eia861: pd.DataFrame,
+    service_territory_eia861: pd.DataFrame,
+) -> pd.DataFrame:
+    """Modified service_territory_eia861 table."""
```
bendnorman, line 304: The doc strings for some of these intermediate norm_ferc714 tables aren't super verbose. What is this function doing?

src/pudl/etl/__init__.py
```diff
@@ -39,6 +39,7 @@
     *load_assets_from_modules([pudl.transform.ferc1], group_name="norm_ferc1"),
     *load_assets_from_modules([pudl.extract.ferc714], group_name="raw_ferc714"),
     *load_assets_from_modules([pudl.transform.ferc714], group_name="clean_ferc714"),
+    *load_assets_from_modules([pudl.output.ferc714], group_name="norm_ferc714"),
```
bendnorman, line 42: I'm a little confused by what assets live in `norm_ferc714`. It looks like a mix of intermediate, 861, and 714 and "normalized" assets. Ideally, we'd have separate groups for intermediate 714 assets, intermediate 861 assets, normalized 714 assets and normalized 861 assets. I can't think of a great place to place the logic for these 4 groups given our current structure. Once we iron out our naming conventions and code structure #2517, we'll probably have a module and asset group for each of the four groups listed above.

I think leaving this collection of assets in `pudl.output.ferc714` is ok for now. Can you think of another way to structure these asset groups?

bendnorman (changes requested):
Looking good! I mostly left some questions about what pieces of logic should be made an intermediate asset.

test/validate/service_territory_test.py
```diff
@@ -0,0 +1,42 @@
...
+    """Verify that output DataFrames don't have too many or too few rows.
+
+    Args:
+        pudl_out_orig: A PudlTabl output object.
+        live_dbs: Boolean (wether we're using a live or testing DB).
+        expected_rows (int): Expected number of rows that the dataframe should
+            contain when all data is loaded and is output without aggregation.
+        df_name (str): Shorthand name identifying the dataframe, corresponding
+            to the name of the function used to pull it from the PudlTabl
+            output object.
```
bendnorman, line 30: We could use type hints instead.

src/pudl/metadata/resources/ferc714.py
```diff
@@ -37,6 +37,63 @@
...
+                "balancing_authority_name_eia",
+                "utility_id_eia",
+                "utility_name_eia",
+                "state",
+                "county",
+                "state_id_fips",
+                "county_id_fips",
+            ]
+            # No primary key here because the state and county FIPS columns
+            # which are part of the natural primary key can be null.
```
bendnorman, line 62: Can you include the names of the fields that would be a part of the natural key in a comment here so folks can understand what a single record in the table represents?

src/pudl/metadata/resources/eia861.py
```diff
@@ -656,4 +656,49 @@
...
+                "state_id_fips",
+            ],
+            "primary_key": ["utility_id_eia", "report_date", "county_id_fips"],
+        },
+        "sources": ["eia861", "censusdp1"],
+        "field_namespace": "eia",
+        "etl_group": "service_territories",
+    },
+    "compiled_geometry_balancing_authority_eia861": {
+        "description": "County-level spatial data for EIA861 utilities.",
```
bendnorman, line 680: Are these descriptions suppose to be the same?

src/pudl/analysis/state_demand.py
```diff
@@ -266,15 +265,18 @@ def load_ventyx_hourly_state_demand(path: str) -> pd.DataFrame:
...
+)
+def load_hourly_demand_matrix_ferc714(
+    demand_hourly_pa_ferc714: pd.DataFrame,
 ) -> tuple[pd.DataFrame, pd.DataFrame]:
     """Read and format FERC 714 hourly demand into matrix form.
 
-    Args:
-        pudl_out: Used to access
-          :meth:`pudl.output.pudltabl.PudlTabl.demand_hourly_pa_ferc714`.
-
```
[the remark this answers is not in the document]
e-belfer: Oops yes, I should do this more consistently.

src/pudl/analysis/state_demand.py
```diff
@@ -499,45 +548,61 @@ def load_eia861_state_total_sales(
...
+        ),
+    },
+)
+def predicted_state_hourly_demand(
+    context,
+    imputed_hourly_demand_ferc714: pd.DataFrame,
+    census_counties: pd.DataFrame,
+    county_assignments_ferc714: pd.DataFrame,
+    total_state_sales_eia861: pd.DataFrame = None,
 ) -> pd.DataFrame:
```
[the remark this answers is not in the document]
e-belfer: In this PR the `clean` and `impute` stages are both long. I grouped the tiny stages (`filter` and `melt`) with their corresponding pair to whittle down to 2 assets, rather than 4. The `load` step currently produces 2 separate assets, rather than 1. I could whittle this down further, but I do think not by much.

The other assets (e.g. `county_assignments`, `census_counties` etc.) perform minor modifications to existing tables and probably could just be called as functions in the state demand asset itself, rather than being defined as intermediate assets. I can double check they don't get called elsewhere and consolidate.

src/pudl/analysis/service_territory.py
```diff
@@ -232,35 +254,37 @@ def get_territory_geometries(
...
+    balancing_authority_assn_eia861: pd.DataFrame,
+    utility_ids_all_eia: pd.DataFrame,
+    service_territory_eia861: pd.DataFrame,
+    utility_assn_eia861: pd.DataFrame,
+    census_counties: pd.DataFrame,
+    entity_type: Literal["ba", "util"],
+    save_format: Literal["geoparquet", "geodataframe", "dataframe"],
+    dissolve: bool = False,
+    limit_by_state: bool = True,
+):
```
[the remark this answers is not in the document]
e-belfer: It depends on the `save_format` selected. I could choose `pd.DataFrame` which also encompasses geodataframes, but perhaps this is not specific enough.

src/pudl/analysis/service_territory.py
```diff
@@ -491,9 +583,22 @@ def main():
 
     for kwargs in kwargs_dicts:
         _ = compile_geoms(
-            pudl_out,
+            balancing_authority_eia861=pd.read_sql(
```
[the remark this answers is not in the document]
e-belfer: This CLI command doesn't add anything additional to the dag, so perhaps it is not necessary to keep?

src/pudl/etl/__init__.py
e-belfer: I agree, this isn't actually normalized but its own analysis layer building on 861/714 tables. Perhaps I could call it `respondents` to differentiate it from the ETL assets? Or leave it as is until the reorganization?

migrations/versions/1ec25c296a6d_add_service_terrtory_ferc714_state_.py
```diff
@@ -0,0 +1,99 @@
+"""Add service terrtory, FERC714, state demand assets
```
e-belfer, line 1: @bendnorman This will be incorrect since we rebased dev, but I'll fix this all with one migration when this gets merged into the Census PR.

src/pudl/analysis/state_demand.py
```diff
@@ -740,28 +807,10 @@ def main():
...
-        assignments=assignments,
-        state_totals=state_totals,
-        mean_overlaps=False,
+
+    # --- Read in inputs from PUDL + dagster cache --- #
+    prediction = pudl.etl.defs.load_asset_value(
+        AssetKey("predicted_state_hourly_demand")
     )
 
     # --- Export results --- #
```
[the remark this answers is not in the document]
bendnorman: Sounds good! I think it makes sense to remove them once we deprecate `PudlTabl`.

src/pudl/analysis/service_territory.py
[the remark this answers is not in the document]
bendnorman: Ah I see. I think it would be helpful to flesh out the docstring so folks know what to expect for each value of `save_format`.

src/pudl/etl/__init__.py
bendnorman: Yeah maybe leave everything in `pudl.output.ferc714` for now and rename the group to `respondents`? Does it make sense to include the data source in the asset group for now? `respondents_ferc714`, `service_territory_eia861`, `state_demand_ferc714`.

src/pudl/analysis/state_demand.py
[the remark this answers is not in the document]
bendnorman: Thanks for pruning the intermediate assets! The DAG feels much cleaner now.

src/pudl/etl/__init__.py
e-belfer: Done!

PR #2552, opened 2023-04-28
Add oil well gas analysis bucket

jdangerx:
Quick terraform change to make a bucket for the well gas analysis work 🏀 

In theory we could split the terraform across a bunch of repositories but I think we are a long way from needing to even split out beyond a single `main.tf` file, so I'm putting the config in our main repo for now.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -36,6 +36,16 @@ resource "google_storage_bucket" "tfstate" {
   }
 }
 
+resource "google_storage_bucket" "well_gas_pdfs" {
+  name          = "${random_id.bucket_prefix.hex}-catalyst-coop-well-gas-pdfs"
+  force_destroy = false
+  location      = "US"
+  storage_class = "STANDARD"
+  versioning {
+    enabled = true
+  }
+}
+
 module "gh_oidc" {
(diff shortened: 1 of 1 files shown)


jdangerx:
I tagged @catalyst-cooperative/inframundo which then auto-requested review from @zschira - who may or may not have bandwidth to review as he's traveling. So I added @bendnorman as another reviewer

terraform/main.tf
```diff
@@ -36,6 +36,16 @@ resource "google_storage_bucket" "tfstate" {
   }
 }
 
+resource "google_storage_bucket" "well_gas_pdfs" {
+  name          = "${random_id.bucket_prefix.hex}-catalyst-coop-well-gas-pdfs"
```
jdangerx, line 40: To make this `well-gas-pdfs.catalyst.coop` I would have to verify ownership of that domain, which seemed like a pain. So I just stuck with this bucket prefix.

zaneselvans: You personally would need to verify ownership? We've verified ownership of `catalyst.coop` already. I thought this was a project-level thing...

jdangerx: I think we'd need to verify ownership of the subdomain too - at least that's what the instructions seemed to imply.

Issue #2535, opened 2023-04-19
Fix expansion of `unit_id_pudl` to be compatible with annually varying `prime_mover_code`


cmgosnell:
### Describe the bug
A while back @zaneselvans developed another step in our pursuit of complete coverage of `unit_id_pudl` #1037. We didn't actually turn it on in our tests so we didn't notice it is not compatible with the change we made to how we are normalizing the `prime_mover_code` - we determined that the PM code should actually be annually varying so it now gets harvested into the annually varying `generators_eia860` table instead of the static `generators_entity_eia` table.

This change is incompatible with the way the current `pudl.output.eia860.fill_unit_ids` works during the `assign_single_gen_unit_ids` step.  This could very well effects other stages of the unit id assignment as well!

### To Reproduce
Steps to reproduce the behavior -- ideally including a code snippet that causes the error to appear.
 - Is the bug related to the software / database? If so, please attach the `settings.yml` file you're using to specify which data to load, and make a note of where in the ETL process the error is happening.
 - Have you found an error or inconsistency in the data that PUDL brings together? If so, what is the data source, year, plant_id, etc -- how can we find the data you're looking at, and what is the nature of the error or inconsistency.
 - Does the problem happen when you're starting up the database (importing data) or does it happen later when you're trying to use the database?

With an all-year pudl.sqlite:
```
pudl_out = pudl.output.pudltabl.PudlTabl(pudl_engine,freq="AS",unit_ids=True)
gens = pudl_out.gens_eia860()
```
Error:
```
2023-04-19 11:51:16 [    INFO] catalystcoop.pudl.transform.eia861:582 Started with 81265 missing BA Codes out of 185553 records (43.80%)
2023-04-19 11:51:18 [    INFO] catalystcoop.pudl.transform.eia861:606 Ended with 14923 missing BA Codes out of 185553 records (8.04%)
2023-04-19 11:51:18 [    INFO] catalystcoop.pudl.output.eia860:182 91.6% of plant records have consistently reported BA Codes
2023-04-19 11:51:18 [    INFO] catalystcoop.pudl.output.eia860:232 Before any filling treatment has been applied. 43.8% of records have no BA codes
2023-04-19 11:51:19 [    INFO] catalystcoop.pudl.output.eia860:232 Backfilling and consistent value is the same. Filled w/ most consistent BA code. 10.8% of records have no BA codes
[...]

Issue #2438, opened 2023-03-22
Convert MCOE analysis to Dagster assets


zaneselvans:
```[tasklist]
- [x] Create `heat_rate_by_unit` asset for both `AS` and `MS` frequency
- [x] Create `heat_rate_by_generator` asset for both `AS` and `MS` frequency
- [x] Create `fuel_cost_by_generator` asset for both `AS` and `MS` frequency
- [x] Create `capacity_factor_by_generator` asset for both `AS` and `MS` frequency
- [x] Create `mcoe` asset for both `AS` and `MS` frequency
- [x] Check whether the MCOE outputs look reasonable
- [x] Add `mcoe` to DB
- [x] Add `heat_rate_by_unit` to DB
- [x] Add `heat_rate_by_gen` to DB
- [x] Add `capacity_factor` to DB
- [x] Add `fuel_cost` to DB
- [x] Update `PudlTabl` to read all MCOE outputs from the DB
- [x] Run integration & validation tests on the MCOE outputs
- [x] Decide if/how to filter MCOE outputs for valid capacity factor, heat rate, and fuel prices
- [x] Decide which of the pre-MCOE tables should actually live in the DB (hr, cf, fc)
- [x] Decide which wide / filled MCOE tables are required with @cmgosnell
```

zaneselvans:
# Notes from Sanity Checks:

Extremely low (11.5%) coverage of `unit_id_pudl` and partial (~70%) coverage of fuel prices means that only a small fraction of all heat rates (10.1%) and fuel costs (6.5%) can be estimated at all. Has it always been this bad?

In the MCOE output we have far better coverage of `net_generation_mwh` than for `total_mmbtu` which I think is the equivalent value on the fuel side. Now that we have allocated estimates for both net generation and fuel consumption, should we be using those estimates in tandem in this analysis to estimate heat rates by generator-month?  What's the difference between these two estimates of total fuel consumption at the generator level?

```python
data_cols = [
    "unit_id_pudl",
    "capacity_factor",
    "fuel_cost_per_mmbtu",
    "fuel_cost_per_mwh",
    "heat_rate_mmbtu_mwh",
    "net_generation_mwh",
    "total_fuel_cost",
    "total_mmbtu",
]

dude = (
    mcoe_monthly[["report_date"] + data_cols]
    .groupby("report_date")
    .apply(lambda x: x.isna().sum() / len(x))
)
for col in data_cols:
    plt.plot(dude[col], label=col)
[...]

PR #2553, opened 2023-04-29
Dagsterize MCOE output tables

zaneselvans:
# PR Overview

Integrate the MCOE outputs into the PUDL DB at both monthly and yearly frequencies.

# PR Checklist


katie-lamb:
Picking up this PR, I think this is what needs to be done (in this order):
- [ ] Move `all_gens` and `gens_cols` functionality that's necessary for creating the mega gens and PPE downstream into the PPE infrastructure (subtasks are notes from Zane, still relevant?)
  - [ ] Create a generators datamart table, using the last bits of the MCOE function.
  - [ ] Redirect PPE / MegaGens to point at the generators datamart table
- [ ] Run integration/validation tests on MCOE outputs
- [ ] Run through comments in #2438 one more time and make sure everything is addressed in terms of which tables should live in the DB and keeping the MCOE table skinny.
- [ ] merge into `dagsterize-net-gen-alloc` and then address the comments on #2527

zaneselvans:
If you're planning to tackle the cleanup issues in the gen-fuel dagsterization PR, I think that could be done first (I don't *think* there's anything in the MCOE PR that needs to happen before the gen-fuel PR is ready to go) and that would bring in new assets more incrementally.

I'm not sure where you saw the reference to the gens datamart table, but I think it's a very wide table of information that @cmgosnell wanted in the context of the PPE / GensMega outputs? Maybe she remembers more specifics.  I think it was a bunch of information that was getting merged into the MCOE outputs, but which I thought was really beyond the scope of those outputs -- but that it needed to be created for other reasons as its own separate thing.

katie-lamb:
@zaneselvans @cmgosnell  I left some questions and notes in my to do list in [this comment](https://github.com/catalyst-cooperative/pudl/pull/2553#issuecomment-1603314101) but consolidating the lingering questions below:

-  In the wide MCOE table (`mcoe_generators`), by default, should only `DEFAULT_GENS_COLS` be included (the previous default for the MCOE table), or should all generator columns be included since it's a data mart table?
- Filtering for valid heat rate, capacity factor, and fuel price is now done at the end of `pudl.analysis.mcoe.mcoe()`. Would one ever need the MCOE table unfiltered for valid CF, HR, and FP or is it fine to leave this filtering in the analysis function?
- Currently the heat rate, capacity factor, and fuel price intermediary tables are loaded into the DB. Wasn't sure where to land on this.
- Take out xfail for MCOE in the integration test `test_eia_outputs`? I think the time coverage issues were addressed in this PR but maybe that message refers to something else?
- Merge this branch into `dagsterize-net-gen-alloc` or merge `dagsterize-net-gen-alloc` first?

docs/release_notes.rst
```diff
@@ -193,6 +226,14 @@ Analysis
   the :ref:`generators_eia860` table so associating those gf and bf records are more
   cleanly associated with generators. Thanks to :user:`grgmiller` for his
   contribution, which was integrated by :user:`cmgosnell`! See PRs :pr:`2235,2446`.
+* The :mod:`pudl.analysis.mcoe` table now uses the allocated estimates for per-generator
```
katie-lamb, line 229: Do these bullets about the new MCOE calculations and tables belong under this analysis section? 

What else am I missing to add to the release notes?

src/pudl/analysis/mcoe.py
```diff
@@ -499,48 +590,88 @@ def mcoe(
         )
         .pipe(pudl.validate.no_null_cols, df_name="fuel_cost + capacity_factor")
     )
+    mcoe_out = mcoe_out.sort_values(
```
katie-lamb, line 593: Should this MCOE function have these `pudl.helpers.oob_to_nan` calls? Or do they belong somewhere else?

src/pudl/metadata/resources/mcoe.py
```diff
@@ -0,0 +1,224 @@
...
+            "etl_group": "outputs",
+        }
+        for freq in AGG_FREQS
+    }
+    | {
+        f"mcoe_{freq}": {
+            "description": (
+                f"{freq.title()} generator capacity factor, heat rate, fuel cost per MMBTU and fuel cost "
+                "per MWh. These calculations are based on the allocation of net generation reported on "
+                "the basis of plant, prime mover and energy source to individual generators. Heat rates "
```
katie-lamb, line 140: I updated this description to reflect the new way for estimating heat rates using allocated estimates for net generation and fuel consumption. Does this seem right?

src/pudl/output/pudltabl.py
```diff
@@ -339,173 +386,11 @@ def gen_eia923(self, update: bool = False) -> pd.DataFrame:
...
-        if update or self._dfs["mcoe"] is None:
-            self._dfs["mcoe"] = pudl.analysis.mcoe.mcoe(
-                self,
-                min_heat_rate=min_heat_rate,
-                min_fuel_cost_per_mwh=min_fuel_cost_per_mwh,
-                min_cap_fact=min_cap_fact,
-                max_cap_fact=max_cap_fact,
-                all_gens=all_gens,
-                gens_cols=gens_cols,
-            )
```
katie-lamb, line 501: I saw that at some point the filtering for valid heat rate, fuel price, and capacity values was being done in this explicitly defined MCOE PudlTabl function. I took out this function, but maybe it should happen here?

src/pudl/analysis/mcoe.py
```diff
@@ -38,7 +39,136 @@
...
+            "gen": AssetIn(
+                key=f"generation_fuel_by_generator_{agg_freqs[freq]}_eia923"
+            ),
+            "bf": AssetIn(key=f"denorm_boiler_fuel_{agg_freqs[freq]}_eia923"),
+        },
+        io_manager_key=io_manager_key,
+        compute_kind="Python",
+    )
+    def hr_by_unit_asset(gen: pd.DataFrame, bf: pd.DataFrame) -> pd.DataFrame:
+        return heat_rate_by_unit(gen=gen, bf=bf)
```
zaneselvans, line 63: Previously we were **not** using the allocated generation fuel inputs to the MCOE calculation -- just the directly reported per-generator generation from the `generation_eia923` table, so this new output will potentially have **very different outputs** (and hopefully also much higher coverage). What kind of sanity checking have we done to make sure that the new heat rates look correct?

We should at least make some histograms of the heat rates so we can see whether the combined cycle natural gas plants, the gas combustion turbines, and the coal-fired steam turbines have the expected distribution of heat rates.

We also previously filtered out the obviously ridiculously wrong values -- e.g. cases where net generation was negative due to low capacity factors + parasitic loads, resulting in negative fuel costs, etc.

What outputs do we want to be writing into the database now?  Should we be filtering out the junk results and if so how?

src/pudl/analysis/mcoe.py
```diff
@@ -38,7 +39,136 @@
...
+    mcoe_asset
+    for freq in ["AS", "MS"]
+    for mcoe_asset in mcoe_asset_factory(
+        freq=freq,
+        io_manager_key="pudl_sqlite_io_manager",
+    )
+]
+
+
+def heat_rate_by_unit(gen: pd.DataFrame, bf: pd.DataFrame) -> pd.DataFrame:
```
zaneselvans, line 171: Some of the docstring for this function seems obsolete, e.g. the parts talking about the BGA dataframe, which no longer appears to be involved?

docs/release_notes.rst
zaneselvans: IIRC when I put down the MCOE PR I had just started playing with using the new allocated generation + fuel outputs to calculate heat rates, and **had not done any validation of the outputs** so this might be dangerous.  What kind of sanity checking have you done?  Does it still pass the data validation tests?  How do the old vs. new distributions compare when broken down by the type of unit (combustion turbine, combined cycle, and steam turbine)?

We should compare the outputs from the old and new methods, including the overall coverage, the restricted outputs (where we only include reasonable values) and the unrestricted outputs (where crazy outputs are still present) to see what the effects of this change are.

src/pudl/analysis/mcoe.py
zaneselvans: I don't think we want to be writing known-to-be-crazy MCOE values into the DB if we can help it.  As this is currently set up, it seems like we'll put unfiltered per-generator capacity factors, fuel costs, and heat rates into the DB in their individual skinny tables, but then filter for realistic values in the combined MCOE table, which is where folks are going to actually access the data I think, which seems like an okay middle ground -- the raw results of our calculations are available, but not highlighted.  We should make it very clear in the descriptions of these assets that they are not intended for use without filtering, and why we expect some of the numbers in them to be totally unreasonable.

Should we parametrize the MCOE constraints in a `context` input to the creation of that asset?

I think @cmgosnell also has thoughts on what we should do here.

src/pudl/metadata/resources/mcoe.py
```diff
@@ -0,0 +1,224 @@
...
+
+AGG_FREQS = ["yearly", "monthly"]
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = (
+    {
+        f"heat_rate_by_unit_{freq}": {
+            "description": (
+                f"{freq.title()} heat rate estimates by generation unit. Generation "
+                "units are identified by ``unit_id_pudl`` and are composed of a set of "
+                "interconnected boilers and generators."
```
zaneselvans, line 12: Need to investigate the unfiltered distribution of these heat rates and add caveats here explaining why some of them are insane.

src/pudl/metadata/resources/mcoe.py
```diff
@@ -0,0 +1,224 @@
...
+        }
+        for freq in AGG_FREQS
+    }
+    | {
+        f"capacity_factor_by_generator_{freq}": {
+            "description": (
+                f"{freq.title()} estimates of generator capacity factor. Capacity "
+                "factor is calculated based on reported generator capacity and the "
+                "allocated net generation reported in the generation and generation "
+                "fuel tables."
```
zaneselvans, line 71: Similarly need to investigate, caveat and explain capacity factors that are < 0.0 and > 1.0.

src/pudl/metadata/resources/mcoe.py
```diff
@@ -0,0 +1,224 @@
...
+    | {
+        f"fuel_cost_by_generator_{freq}": {
+            "description": (
+                f"{freq.title()} estimate of per-generator fuel costs both per MMBTU "
+                "and per MWh. These calculations are based on the allocation of net "
+                "generation and plant-level delivered fuel prices reported in the fuel "
+                "receipts and cost table, along with the un-allocated fuel consumption "
+                "from the boiler fuel table. This means fuel cost coverage is low."
+                "The fuel costs are also currently aggregated to coarse fuel "
+                "categories rather than using the more detailed energy source codes."
```
zaneselvans, line 103: If the heat rates are based on allocated fuel consumption, why should the fuel costs not also be based on the allocated fuel consumption?  Is this documentation from when we had done the net generation allocation, but hadn't yet allocated the fuel consumption?

What fuel cost coverage do we expect here, in terms of overall records, and in terms of fraction of total generation? Why are the fuel costs aggregated to coarse categories rather than the energy source codes?

Should investigate, explain, and caveat negative or other ridiculous fuel costs per MWh.

src/pudl/metadata/resources/mcoe.py
zaneselvans: I think we should probably also note that there are good reasons why one might get ridiculous numbers for some of these outputs, and that those outputs have been filtered out of these results.

However, doing this filtering only at the final step means that there are probably some accidentally reasonable numbers that get left in the outputs.

src/pudl/analysis/mcoe.py
zaneselvans: The fact that we're waiting until the final step to do this filtering, and then only nulling out the individually ridiculous values in each column seems a little bit suspect to me. Since the fuel cost per MWh depends on all of the other values in a row, if any of those values is ridiculous, we should probably null the entire row -- otherwise we can get cases where two numbers that are ridiculous on their own multiply to yield a "reasonable" heat rate / capacity factor / fuel cost value, which we then retain, even though it's based on obviously bogus inputs.

zaneselvans (commented):
* I think there may be some additional sanity checking required given our use of the newly allocated generation and fuel data
* We should probably explain why some of the MCOE inputs have ridiculous values.
* When we do finally filter the MCOE outputs, we should probably null out more values in a row, once we've identified that some of the inputs are invalid, lest we leave in "valid" values that are composed of combinations of invalid values. Either that or we should be filtering along the way so the invalid values and their dependent outputs are nulled as we go.

src/pudl/analysis/mcoe.py
cmgosnell: okay i think i understand what is all going on here.

I agree that keeping in the baddies in our lil skinny tables and nulling the oobs in this "hey look at me, i'm mcoe" table is a good idea. agreed on the noting the known baddies in the resource defs of the skinny guys.

on the only null-ing the bad field, not the downstream related fields that are maybe deceptively normal looking: what if we included a `related_fields` arg into `oob_to_nan` and pointedly nulled based on known related things? It would feel bad to just null the whole record if one thing was oob.

a question that this all would result in... how do we deal with the validation tests in this new arrangement? rn all of the mcoe validation tests use the mcoe `pudl_out` table with all the min/maxes set to `None` so it keeps all the baddies in there. I **think** a simple enough solution to that would be to set these min/max oob default in this asset, but in the validation tests we could use the skinny all the stinky stuff in there assets as inputs to this function with all `None`s for the min/max.

src/pudl/analysis/mcoe.py
cmgosnell: one thing to note is that i think the inputs to the mcoe as *slightly* updated from the net gen/fuel allocation stuff. the net generation side is definitely incorporated into this equation but the allocated fuel *is not* incorporated. The allocated generation was easy to insert into mcoe as an option because we had a pudl table that was the same "shape" (i.e. it had date/plant id/gen id as primary keys that we could swap into the same methodology). There was no date/plant id/gen id level table on the fuel side. so we kept this `heat_rate_by_unit` step which takes the `bf` table which is not the allocated fuel. so the fuel side of the whole mcoe dag is still the reported values.

src/pudl/analysis/mcoe.py
zaneselvans: Rather than overloading `oob_to_nan` with more responsibilities maybe we should use it in a new function? Really it should be operating on a single column at a time rather than a whole dataframe but it's an old function.  So the idea would be to say "If you find an out of bounds value in column X, you should null it as well as columns Y and Z (which we know a priori depend on X)?

I agree that we should add validations which expect all of the mess for the skinny and unfiltered MCOE inputs, and then a separate validation which looks at the filtered / cleaned up values in the MCOE table.

With the new outputs is it possible to validate particular prime movers rather than just gross fuel type?  Like can we separately check the heat rates on gas combustion turbines, combined cycle natural gas generators, and coal fired steam turbines?  They should each have pretty distinctive distributions in the clean MCOE outputs.

src/pudl/analysis/mcoe.py
zaneselvans: OHHH, you're right.  So we're not really updating the calculation much if at all here, but it would be easy to do so if we wanted to, and then this asset would only depend on the `gen` dataframe, right? And we'd be using both the generation and the fuel consumed within each record of that dataframe. Do we have reason to believe that this would be a lower quality output?  Would it deal with the multi-fuel units worse?

I guess this means that we still aren't getting a boost in the overall MCOE output coverage.

src/pudl/output/pudltabl.py
```diff
@@ -554,6 +436,9 @@ def gens_mega_eia(
                     "The frequency of the pudl_out object must be `AS` for the "
                     f"plant-parts table and we got {self.freq}"
                 )
+            # TODO: gens_cols doesn't do anything right now, make the default behavior
+            # for mcoe_generators to load all generator attributes and then filter
+            # gens_mega by gens_cols?
             if gens_cols is None:
                 gens_cols = []
             if gens_cols != "all":
```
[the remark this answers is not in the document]
katie-lamb: Handling this in the PPE/gens mega dagsterization PR.

Expected behavior is that all columns are included in the MCOE table and only a subset of columns is included in the gens mega and PPE tables.

src/pudl/analysis/mcoe.py
katie-lamb: I added a helper function `oob_to_nan_with_dependent_cols` that calls `oob_to_nan` to nullify columns that depend on other columns that have been nulled.

The three columns in the MCOE calculation where `oob_to_nan` is called are:
- `heat_rate_mmbtu_mwh`, used to derive `total_mmbtu`, `fuel_cost_per_mwh`
- `fuel_cost_per_mwh`, no downstream derived columns
- `capacity_factor`, no downstream derived columns

src/pudl/metadata/resources/mcoe.py
katie-lamb: I updated the descriptions for the intermediate tables where ridiculous numbers are expected and additionally added a helper function (`oob_to_nan_with_dependent_cols`) to filter out accidentally reasonable calculations.

src/pudl/analysis/mcoe.py
```diff
@@ -38,7 +39,196 @@
...
+                    "aren't being used enough to be valid."
+                ),
+            ),
+            "max_cap_fact": Field(
+                float,
+                default_value=1.5,
+                description=(
+                    "The maximum generator capacity factor. Generator records "
+                    "with a higher capacity factor will be filtered out."
+                ),
```
katie-lamb, line 155: I added a context input to parameterize the bounds for these values. The default value that was in there for `max_cap_fact` was 1.5. Why not 1.0? Should I leave as 1.5?

katie-lamb:
Subbing in the fuel allocations to get new heat rate per generator estimations looks pretty good. Coverage increases and the distribution looks a little better.

**Coverage change over report dates with raw heat rates and heat rates filtered from 5.5 to 25 mmbtu/mwh (15% efficiency):**
heat_rate_coverage

**Change in distribution by type of generator**:
heat_rate_distribution

zaneselvans:
Would you make all those histograms with at least 100 bins?

katie-lamb:
@zaneselvans 
**Heat rate distribution**
heat_rate_distribution

**Fuel cost weighted by the heat content of fuel consumed**
fuel_cost_by_pm

**Fuel cost per MMBTU**
fuel_cost_per_mmbtu_by_pm

katie-lamb:
Since fuel consumption doesn't get allocated to the generator level, I merged on fuel consumption at the plant + fuel type level, filtered out null fuel costs, and dropped duplicates to get the fuel consumption for plants with non-null fuel costs. Then summed by report date and divided by the total fuel consumption. Is this along the lines of what you were thinking?

```python
def get_fuel_consumption_coverage(df, frc):
    df = df[df.fuel_cost_per_mmbtu.notnull()]
    df = df.drop_duplicates(subset=["report_date", "plant_id_eia", "fuel_type_code_pudl"], keep="first")
    return (
        df.groupby("report_date")["fuel_consumed_mmbtu"].sum()
        / frc.groupby("report_date")["fuel_consumed_mmbtu"].sum()
    )

fc_cov = get_fuel_consumption_coverage(df=fuel_cost, frc=frc)
fc_alloc_cov = get_fuel_consumption_coverage(df=fuel_cost_alloc, frc=frc)

plt.plot(fc_alloc_cov, label="fuel allocations")
plt.plot(fc_cov, label="old estimation")
plt.title("Fuel consumption coverage")
plt.ylabel("Fraction of fuel consumed covered by fuel costs")
plt.legend()
```
fraction_fuel_consumption_coverage

zaneselvans:
@katie-lamb Are the heat rate histograms weighted by MMBTU consumed?

zaneselvans:
On the coverage I was thinking of the heat rate coverage graphs you made -- heat rate coverage raw & heat rate coverage filtered. Like of all the fuel consumed, what fraction of it do we have associated with a heat rate?

Or is the main reason that we have missing heat rates that we having missing fuel consumption?

zaneselvans:
It's interesting that the fuel cost coverage is trending down over time. Can you separate it out by fuel?  Is it trending down overall because the proportion of fuel that's gas is trending up, and they redact more of the gas prices?

katie-lamb:
> Or is the main reason that we have missing heat rates that we having missing fuel consumption?
Yep, the main reason for missing heat rates I think is missing fuel consumption - does it make sense that pretty much 100% of fuel consumption is covered with unfiltered heat rates?  

```python
def get_coverage_of_fuel_consumed(df, filtered=False):
    if filtered:
        hr_col = "heat_rate_mmbtu_mwh_filtered"
    else:
        hr_col = "heat_rate_mmbtu_mwh"
    return (
        df[["report_date"] + ["heat_rate_mmbtu_mwh", "heat_rate_mmbtu_mwh_filtered", "fuel_consumed_fraction"]]
        .groupby("report_date")
        .apply(lambda x: x.dropna(subset=hr_col).sum(numeric_only=True))
    )
fig, axs = plt.subplots(1, 2, sharey=True, sharex=True, figsize=(10, 6), tight_layout=True)

hr_by_gen_cov = get_coverage_of_fuel_consumed(hr_by_gen)
hr_by_gen_alloc_cov = get_coverage_of_fuel_consumed(hr_by_gen_alloc)

axs[0].plot(hr_by_gen_cov["fuel_consumed_fraction"], label="old estimation")
axs[0].plot(hr_by_gen_alloc_cov["fuel_consumed_fraction"], label="fuel allocations", alpha=.5)
axs[0].set_title("Heat rate coverage raw")
axs[0].set_ylabel("Fraction of fuel consumption covered")
axs[0].legend()

hr_by_gen_cov_filtered = get_coverage_of_fuel_consumed(hr_by_gen, filtered=True)
hr_by_gen_alloc_cov_filtered = get_coverage_of_fuel_consumed(hr_by_gen_alloc, filtered=True)
axs[1].plot(hr_by_gen_cov_filtered["fuel_consumed_fraction"], label="old estimation")
axs[1].plot(hr_by_gen_alloc_cov_filtered["fuel_consumed_fraction"], label="fuel allocations", alpha=.5)
axs[1].set_title("Heat rate coverage filtered")
axs[1].set_ylabel("Fraction of fuel consumption covered")
axs[1].legend()
```
fraction_of_fuel_coverage

katie-lamb:
> Are the heat rate histograms weighted by MMBTU consumed? I think we're probably only interested in seeing the weighted histograms -- otherwise we're counting records, which means lots of tiny records will overwhelm a few big ones.
Is this what you were thinking?

```python
fig, axs = plt.subplots(2, 2, figsize=(10, 10), tight_layout=True)
codes = iter(thermal_ftcp)
for row in range(2):
    for col in range(2):
        ax = axs[row][col]
        code = next(codes)
        alloc_df = hr_by_gen_alloc.loc[
            hr_by_gen_alloc.fuel_type_code_pudl == code,
            ["heat_rate_mmbtu_mwh", "fuel_consumed_for_electricity_mmbtu"]
        ]
        old_df = hr_by_gen.loc[
            hr_by_gen.fuel_type_code_pudl == code,
            ["heat_rate_mmbtu_mwh", "fuel_consumed_for_electricity_mmbtu"]
        ]
        ax.hist(
            alloc_df["heat_rate_mmbtu_mwh"],
            range=(0, 50),
            bins=150,
            label="fuel allocations",
            weights=alloc_df["fuel_consumed_for_electricity_mmbtu"],
        )
        ax.hist(
            old_df["heat_rate_mmbtu_mwh"],
            range=(0, 50),
            bins=150,
            label="old estimation",
            weights=old_df["fuel_consumed_for_electricity_mmbtu"],
            alpha=0.5,
        )
        ax.set_title(f"{code}")
        ax.set_xlabel("heat_rate_mmbtu_mwh")
```
heat_rate_distribution

katie-lamb:
> It's interesting that the fuel cost coverage is trending down over time. Can you separate it out by fuel? Is it trending down overall because the proportion of fuel that's gas is trending up, and they redact more of the gas prices?
fuel_cost_coverage

zaneselvans:
Yes, the heat content weighted histograms are more what I was thinking, but they need to be normalized so we can compare the distributions since the total quantity of fuel reported in the two cases are so different. We particularly care about the coal and gas since they're the overwhelming bulk of the energy / emissions, so it would be nice to see them from like 0-20 mmbtu/MWh.

It looks like there are some weird gaps showing up in the old MCOE distributes for coal and gas and waste.

Generally the new coal heat-rate distribution looks right, but I am suspicious of the new gas heat rate distribution.

It's got 2 peaks at about the right places for combined cycle (~7) and open-cycle gas turbines (~11) but way more heat content in the high peak, which is probably wrong. IIRC a large majority of gas consumed goes to the much higher efficiency CC plants, and the GTs are used infrequently as peakers.

My guess is that the distribution looks like this because the data is bad in earlier years, maybe with only the gas turbine portion of the CC unit reporting generation, and getting ~100% of the fuel consumption attributed to them, which would result in a heat rate that looks like a gas turbine.  If you plot the same histogram for every year of data, I think you'll find that some time around 2015 the distribution shifts radically -- at the point when EIA started collecting generation data for both the gas turbines and the heat recovering steam generators. In the data validations, we only look at gas heat rates for 2015+.

The little lump in the gas heat rate distribution that's close to zero might be associated with the HRSGs, which consume little or no fuel directly, but still generate electricity.  If so, then if you weight by MWh generated rather than fuel heat content consumed, I would expect that lump to get much bigger (i.e. how many MWh were generated at a given heat rate, rather than how much fuel was consumed at a given heat rate).

I wouldn't have guessed that coal fuel price coverage would be dropping over time and gas coverage has been rising! It's also interesting that there's a big drop in coal coverage at about the same time as there's a big jump in gas coverage in like 2014-2015 in the older estimates.

katie-lamb:
As theorized above, around 2014 is the shift in distribution, when combined cycle plants are reported correctly. Unfortunately, the high peak remains too large. 
gas_heat_rate_timeseries

zaneselvans:
Well, the new plots are informative at least!

I think my next step would probably be to try and break down the new allocated gen-fuel data somehow, into different allocation cases, and see if some of them get the right answer, and others get the wrong answer, so we can understand which part of the allocation processes might be the source of the problem -- looking just at the 2015+ years where we expect to be able to get a good result based on the available data.

### Questions
* @cmgosnell do you have any intuition as to what the relevant difference might be between the old and new processes?
* Are the heat rates implied by the gen-fuel allocations aware of the generation unit ID level aggregations required to get meaningful heat rates in the case of multi-generator units (like all the combined cycle units)?
* Does this affect how we feel about publishing the allocated gen-fuel data? Are there additional caveats that we want to add?
* What do we want to do as far as the MCOE outputs go? Try and figure out what the issue is and fix it in pursuit of new and better MCOE estimates / coverage? Or just keep using the older lower coverage outputs?
* Is it possible / would it be informative to merge in Prime Mover information in association with the generator IDs in the gen-fuel allocated data, and look just at the heat rate distribution for the combined cycle vs. open cycle prime movers? (CC+CT vs. GT+ST)

katie-lamb:
@zaneselvans @cmgosnell I couldn't figure out how to best aggregate the gen-fuel allocated generator level heat rates up to the unit level (the old estimations use hr by unit as an intermediary step to get to hr by gen). So breaking it down by prime mover instead:

> Are the heat rates implied by the gen-fuel allocations aware of the generation unit ID level aggregations required to get meaningful heat rates in the case of multi-generator units (like all the combined cycle units)?

The gen-fuel allocations result in `CA` generators being allocated very little fuel while the `CT` generator gets all the fuel for the combined cycle unit, which makes sense at a generator level but makes for incorrect heat rates when considering the unit as a whole. In the old estimations, every generator in the unit is given the same heat rate as the entire unit (so ~7 for the `CA` and `CT` gen in CC units). 

The open cycle `ST` generators look okay. Note that there were only a handful of `GT` generators with valid net generation and heat rates, so I didn't include that distribution.

> Does this affect how we feel about publishing the allocated gen-fuel data? Are there additional caveats that we want to add?

I feel as if the fuel allocation is theoretically correct right now, but the docs should probably say that the steam turbine part of CC units aren't allocated fuel. Or maybe I'm wrong in thinking that's what's intuitive.



hr_by_pm


> What do we want to do as far as the MCOE outputs go? Try and figure out what the issue is and fix it in pursuit of new and better MCOE estimates / coverage? Or just keep using the older lower coverage outputs?

I think if it's easy to re-allocate the combined cycle fuel allocations (maybe in the MCOE analysis itself and keeping the fuel allocations as is) so that the heat rate by unit makes more sense then it's worth fixing so that we can boost the coverage of the MCOE table. This would be helpful for me in the FERC to EIA match context. That being said, I'm not sure how difficult this will be and don't feel that familiar with how this process might go down.

zaneselvans:
The heat rates are only really meaningful at the generation unit level, since that's the aggregation of infrastructure that bundles together fuel inputs and their corresponding electricity outputs.  The combined cycle units are the main place that this makes a mess of things.

If we merge in `unit_id_pudl` from `boiler_generator_assn_eia860` (the unit IDs should probably be in all of the denormalized generator tables and maybe built into the gen-fuel tables for just this purpose) we should be able to calculate the overall heat-rate by unit, and see if it looks reasonable at all:

```py
gen_fuel = pd.read_sql("generation_fuel_by_generator_energy_source_monthly_eia923", pudl_engine)
bga = pd.read_sql("boiler_generator_assn_eia860", pudl_engine)

gen_fuel_by_unit = pudl.helpers.date_merge(
    left=gen_fuel,
    right=bga[["report_date", "plant_id_eia", "generator_id", "unit_id_pudl"]],
    on=["plant_id_eia", "generator_id"],
    how="left",
)

heat_rate_by_unit = (
    gen_fuel_by_unit.dropna(subset="unit_id_pudl")
    .groupby(["report_date", "plant_id_eia", "unit_id_pudl"], as_index=False)[
        ["net_generation_mwh", "fuel_consumed_for_electricity_mmbtu"]
    ].sum()
    .convert_dtypes()
    .assign(hr_by_unit=lambda x: x.fuel_consumed_for_electricity_mmbtu / x.net_generation_mwh)
)

hr_recent = heat_rate_by_unit[heat_rate_by_unit.report_date >= "2015-01-01"]
plt.hist(hr_recent.hr_by_unit, weights=hr_recent.fuel_consumed_for_electricity_mmbtu, range=(1,20), bins=100)
plt.xlabel("Heat Rate [MMBTU/MWh]")
plt.ylabel("Total Heat Content [MMBTU]")
plt.title("Generation Unit Heat Rates Weighted by MMBTU Consumed.");
```
image

This distribution looks really good!

Unfortunately, it has also aggregated away all of the information about the prime movers involved and the primary fuel being consumed, but it suggests that -- at least for for the generators where we've inferred `unit_id_pudl` -- the distribution of heat rates that are implied by the new fuel allocation is very good.

I think this table of generation unit level heat rates (the only level that makes physical sense) can be merged back onto the allocated generation-fuel table, as we did in the old MCOE process, to end up with a heat rate that's associated with all the more specific generator-level information like prime mover and the energy source code, so that we can look at the heat rate distributions associated with coal fired steam turbines, gas fired combined cycle units, or stand-alone gas turbines, and we have a good idea of what each of those curves should look like.

I don't know how much better the overall coverage will be -- but you could look at the proportion of allocated fuel consumed that ends up in the aggregated `unit_id_pudl` level dataframe above and comapre it to the existing MCOE outputs, which would give us some idea.

There's also a much more involved (and unvalidated) set of heuristics for assigning `unit_id_pudl` values to many more boilers and generators laying around, which would improve the overall coverage.

* Using the allocated generation-fuel table should improve coverage because it includes data from boilers and generators that only appear in the original `generation_fuel_eia923` table.
* Increasing the number of boilers & generators that have `unit_id_pudl` values will increase overall coverage since we'll be able to associate more boilers and generators into generation units, which is necessary for calculating real heat rates.
* Increasing the number of fuel delivery records that have (estimated) fuel prices associated with them will then allow us to use these heat rates to estimate the fuel cost per MWh for more units.

katie-lamb:
I used @zaneselvans `unit_id_pudl` aggregated heat rates from above and then merged them onto the allocated generation fuel table (`generation_fuel_by_generator_monthly_eia923`). The distribution of heat rates by prime mover now looks pretty good (weighting by fuel consumed) when looking at the 2015 and later data:

unit_hr_by_pm

However, as expected, we get barely any increase in coverage. Here's the fraction of generation covered by the `unit_id_pudl` aggregated heat rates compared to the old process:

unit_hr_frac_of_fuel_coverage

So where does this leave us? 
- We could plug in the fuel allocations aggregated to the `unit_id_pudl` level into the MCOE process
- We could leave the MCOE process as is
- Make an issue to try to improve the coverage by assigning more `unit_id_pudl` to boilers and generators. Are there any other ideas for improving coverage?

cmgosnell:
hey @katie-lamb ! I'm vv glad the unit aggregations look reasonable!

I certainly agree that heat rates in particular are only informative at the unit level. We do have an open issue about the unit id expansion #2535. @zaneselvans  did a ton of work a while ago to try to get complete unit coverage that a) was never tested in an ongoing way or deeply validated and b) is no longer working bc we've changed some of the expectations of the inputs.

In terms of moving forward I could see a few modes:
* Use what we've got way: We could incorporate the allocated fuel consumption into the mcoe tables & only calculate heat rates for records when we have a `unit_id_pudl`. _If_ we wanted to enable the ability to turn the allocation on or off i think we'd need to make a new asset or table which is the fuel consumption at the unit level: which could be either made from the `boiler_fuel_eia923` table or the allocated generator level fuel table.
* Longer/complete way: make the `unit_id_pud` expansion work with current data. Validate the new unit ids somehow/someway. Then do everything above.
 
Updating/integrating the more-complete `unit_id_pudl` would be helpful for other things! The `subplant_id` uses `unit_id_pudl`.
But/and i don't think theses two ways forward are conflicting at all. I assume there would still be unique instances of null `unit_id_pudl` no matter what. To me, the main risk of integrating what we currently have is that users could easily calculate heat rates for everything that would be... wrong/unhelpful. If we do this we should at a minimum add a little warning into the user-facing docs.

katie-lamb:
In the interest of dagsterizing the PPL so that the FERC to EIA modeling effort can be continued, I think we should do the "use what we've got" route while perhaps working in parallel on the "longer more complete" route of filling in `unit_id_pudl`. 

> * _If_ we wanted to enable the ability to turn the allocation on or off i think we'd need to make a new asset or table which is the fuel consumption at the unit level: which could be either made from the `boiler_fuel_eia923` table or the allocated generator level fuel table. This would give us barely more heat rates, but we'd get the complete coverage of fuel consumption!

@cmgosnell I feel a little confused about the need for this additional table of fuel consumption at the unit level. What contexts would we want the ability to turn the allocation off? Is it useful/necessary to be able to go back to the old estimates? Is this unit level fuel consumption just necessary so that, as you say, users can't calculate heat rates that are wrong/bad, or is that a separate issue?

katie-lamb:
I compared the old MCOE to the new MCOE with fuel allocations aggregated at the unit level. Breaking `total_mmbtu` per generator out by fuel type and report year, the comparison is looking pretty good. Seems like it makes sense to proceed with both generation and fuel allocations.
coal_total_mmbtu

gas_total_mmbtu

zaneselvans:
I'm not sure if these plots are in response to my request on the call, but I was wanting to compare the heat rate distributions by fuel and year, since we know what they should look like and also that we think they should be about the same in the old and new versions of the data.
 
The features I would be looking for to know that things are still working as expected (and that I think are present in the old data):
* In gas (weighted by heat content) we should see a big peak around 7 mmbtu/MWh (CCGT) and a small peak around 10-11 mmbtu/MWh (standalone gas turbines, which don't generate much).
* The relative size of the CCGT peak should increase over time, as combined cycle plants become more prevalent, both compared to the open cycle gas turbines, and compared to coal fired generation.
* We should see a coal peak around 10-11 mmbtu/MWh (typical steam turbine heat rate)
* The coal peak should shrink over time relative to the gas peak as generation shifts from coal to gas.

katie-lamb:
@zaneselvans my understanding on the call was that you wanted to see fuel consumption in the final MCOE tables broken out by fuel and time. Here's heat rate broken out by fuel and time weighted by total MMBTU

hr_fuel_time

zaneselvans:
I can't really tell what's going on with 4 different types of very skinny bars in those histograms.

katie-lamb:
@zaneselvans what would you prefer?

coal_hr_fuel_time
gas_hr_fuel_time

zaneselvans:
Separating them out into coal and gas series separately helps a lot!

The only big change that I'm curious about is the 2014 gas, which with the newly allocated fuel looks much more like the subsequent years with a big peak in the CC heat rate range around 7.5 MMBTU/MWh. This is an improvement, but do we understand what happened to change it? Does 2013 look like 2012 or 2014?

katie-lamb:
@zaneselvans hopefully there's an actual explanation for the 2014 vs 2015 data reporting difference, not just an off by one.

gas_hr_fuel_time

jdangerx:
OK, took a look at this again since Katie raised a flag about “this is taking a long time." 

My proposal for next steps is:

* [ ] @zaneselvans / @katie-lamb / me pair on beefing up the validation tests to assert things about the actual shape of the data: "there should be peaks in the gas heat rate at ~7 and ~10" etc.
* [ ] We get the validation tests passing and merge with as little effort as possible - likely using allocated generation, but not allocated fuel
* [ ] We make a new ticket for using cool new data in MCOE, which can be worked on separately from Plant Parts.

------

**notes/context/thought process, so you can see all the things I forgot to consider**

Sounds like we have a couple intertwined things to consider:

1. what is the overall goal of this PR?

2. are we going to keep trying to use the allocated generation data, or just use the same data sources we were using before?

3. if we use the same data sources we were using before, do we need to check the outputs for "actually connected to reality"?

4. if we want to keep trying with the newer/better data, how can we make the feedback loop shorter? right now it appears to be "katie makes some graphs, zane asks for new graphs, repeat"

My own understanding as someone parachuting into this project is:

1. we want MCOE outputs in the database (mcoe and mcoe_generators in particular) so that end users can access it; secondarily we want to move these calculations into dagster so that PPL can depend on that

2. I think that we can decouple "get data into the database" and "make the data better," and just focus on the old data sources. We can take all this analysis we've done and put it towards another PR, that can be worked on in parallel with the PPL stuff.

3. I think we should write some tests that express [Zane's acceptance criteria for heat rates](https://github.com/catalyst-cooperative/pudl/pull/2553#issuecomment-1665796434). I'm imagining some sort of clever usage of `scipy.signal.find_peaks` or similar... This basically caches all the thinking y'all have done about the correctness of the data, so that when we work on improving things later we can do so with confidence. Are there other acceptance criteria, and do we need these for CF/FC/MCOE/etc.? 

4. I think the tests would help the feedback loop be shorter, but also I could imagine a good Zane/Katie pairing session would be much faster and more pleasant than the async back-and-forth we have going on here. That might be a good way to discover the right assertions we want to make in our tests!

katie-lamb:
Agree that we should build the "success criteria" into the validation tests to make improvements to the MCOE that we do later down the line easier to validate (when we get better coverage for `unit_id_pudl` for example). This would be somewhat new to me, so it might be useful to pair with @jdangerx on how this should be best implemented (`find_peaks` seems like a good place to start). I think it's likely that the MCOE passes these validation tests with both allocated generation and fuel, in which case I think we should merge with both. Otherwise, I can rewind a commit and we can just use generation allocations.

zaneselvans:
We know the Gregs use this data downstream, so I've been more concerned than normal about wanting to see confirmation that we haven't broken anything with the new methodology.

I haven't seen any problems with using both the allocated fuel and allocated generation, it has just a bit of a communications struggle to see the relevant metrics to know that. IIRC EIA started reporting the full fuel consumption for the CCGTs mid-2014, which I suspect is why the 2014 data looks different -- not sure why the new and old data would have yielded different results there, but the new results look more like what we would expect from CCGTs in general, so I don't think it's bad.

We have basic quantile value checks for the heat rate distribution coded in the validations right now, and given the plots above, I think that the MCOE using both allocated fuel and allocated generation will pass them. Getting a new type of data validation check set up sounds like a good idea in general -- we need to overhaul and expand our current data validations... given that we've basically never expanded them to cover the many, many new tables we've been creating, and there is much better tooling for this kind of validation than what we've got now -- but to me it feels beyond the scope of this PR. It would be great to do in conjunction with revisiting the expansion of coverage by `unit_id_pudl` which is the limiting factor in our heat rate coverage.

But I don't think that work is going to happen within the scope of any work we currently have funded, and I would really like to get the work that GregM contributed and we integrated into the outputs within the scope of our current funding.

zaneselvans:
@katie-lamb I created this PR originally so I can't approve it, but I'm good with it getting merged in and using the allocated fuel and generation for the MCOE.

jdangerx:
> @katie-lamb I created this PR originally so I can't approve it, but I'm good with it getting merged in and using the allocated fuel and generation for the MCOE.

rubber stamp incoming

jdangerx (approved):
Let's merge this so we can get on with our lives! And also, separatelyt, we can try to codify a lot of the verification work that was done by @zaneselvans and @katie-lamb so we don't lose all that juicy context.

cmgosnell:
I honestly don't have strong feelings about the dropping of null fuel_cost_per_mwh. BUT if we did want that then you could make a non-null non-db asset that feeds into a db skinny asset & the wide db asset.

Also I assume that the xfail situation is because of the expanded time series? which the skinny table doesn't have right?

All these other options sound good to me!

zaneselvans:
Any idea what's up with the integration tests? Looks like there's a `unit_id_pudl` column missing somewhere.

PR #2558, opened 2023-05-01
incorporate boiler_id into epacamd_eia table

cmgosnell:
# PR Overview
note: this commit does not also update the alembic schema or remove the original epacamd_eia table from the db.

# PR Checklist

### DIFF
--- a/src/pudl/etl/glue_assets.py
+++ b/src/pudl/etl/glue_assets.py
@@ -287,6 +287,7 @@ def epacamd_eia_subplant_ids(
                 "plant_id_eia",
                 "generator_id",
                 "subplant_id",
+                "boiler_id",
             ]
         )
     ).any():
--- a/src/pudl/metadata/resources/glue.py
+++ b/src/pudl/metadata/resources/glue.py
@@ -80,6 +80,7 @@ successfully mapped to EIA generators.
                 "unit_id_pudl",
                 "emissions_unit_id_epa",
                 "generator_id",
+                "boiler_id",
             ],
         },
         "field_namespace": "glue",
--- a/test/unit/transform/glue_test.py
+++ b/test/unit/transform/glue_test.py
@@ -40,11 +40,11 @@ plant_id_eia,emissions_unit_id_epa
     boiler_generator_assn_eia860_test = pd.read_csv(
         StringIO(
             """
-plant_id_eia,generator_id,unit_id_pudl
-1392,1A,1
-1392,2A,1
-1392,3A,1
-1,A,1
+plant_id_eia,generator_id,unit_id_pudl,boiler_id
+1392,1A,1,1A
+1392,2A,1,2A
+1392,3A,1,3A
+1,A,1,
 """
         )
     )
@@ -65,13 +65,13 @@ plant_id_eia,generator_id
     epacamd_eia_subplant_ids_expected = pd.read_csv(
         StringIO(
             """
-plant_id_eia,plant_id_epa,unit_id_pudl,emissions_unit_id_epa,generator_id,subplant_id
-1392,1392,1,1A,1A,0
-1392,1392,1,2A,2A,0
-1392,1392,1,3A,3A,0
-1392,1392,,4A,4A,1
-1392,1392,,5A,5A,2
-1,1,1,A,A,0
+plant_id_eia,plant_id_epa,unit_id_pudl,emissions_unit_id_epa,generator_id,boiler_id,subplant_id
+1392,1392,1,1A,1A,1A,0
+1392,1392,1,2A,2A,2A,0
+1392,1392,1,3A,3A,3A,0
+1392,1392,,4A,4A,,1
+1392,1392,,5A,5A,,2
+1,1,1,A,A,,0
 """
         )
     )


aesharpe:
@cmgosnell some thougts:

Taking a closer look at the `epacamd_eia_subplant_ids` table and seeing this part of the `glue_assets.py` module:

```
    epacamd_eia_complete = (
        augement_crosswalk_with_generators_eia860(epacamd_eia, generators_eia860)
        .pipe(augement_crosswalk_with_epacamd_ids, emissions_unit_ids_epacems)
        .pipe(augement_crosswalk_with_bga_eia860, boiler_generator_assn_eia860)
    )
```

When we augment the crosswalk with the generators_eia860 table we fill empty `emissions_unit_id_epa` values with existing `generator_id` values:

```
emissions_unit_id_epa=lambda x: x.emissions_unit_id_epa.fillna(x.generator_id)
```

I'm not sure this is the best idea, because many of these generators are renewables and thus wouldn't have an associated smokestack as is implied by the existence of an `emissions_unit_id_pudl`. For example:

```
epacamd_eia_subplant_ids[(epacamd_eia_subplant_ids["plant_id_eia"]==141) & (epacamd_eia_subplant_ids["generator_id"]=="PV-3")]
```
|     |   plant_id_eia |   plant_id_epa |   subplant_id |   unit_id_pudl | emissions_unit_id_epa   | generator_id   | boiler_id   |
|----:|---------------:|---------------:|--------------:|---------------:|:------------------------|:---------------|:------------|
| 392 |            141 |            141 |             3 |            nan | PV-3                    | PV-3           |             |

```
generators_eia860[(generators_eia860["plant_id_eia"]==141) & (generators_eia860["generator_id"]=="PV-3")].drop_duplicates(["plant_id_eia"])[["plant_id_eia", "generator_id", "technology_description"]]
```
|       |   plant_id_eia | generator_id   | technology_description   |
|------:|---------------:|:---------------|:-------------------------|
| 31903 |            141 | PV-3           | Solar Photovoltaic       |

aesharpe:
To address the above comment, instead of filling the NA gaps in `emissions_unit_id_epa` with `generator_id`, I created a temporary column called `network_x_key` that does the same thing. This column is then fed into the sub-plant id creation method instead of `emissions_unit_id_epa` and then gets dropped due to it's exclusion from the metadata, and the `emissions_unit_id_epa` column remains intact.

aesharpe:
## BGA-to-subplant_id bug

I noticed that we merge the sub-plant id table with the bga table, but the m:m bga boiler-generator relationship is not retained after the sub-plant id creation. I'm not exactly sure why, but I think it has to do with how the network x is currently calculating the sub-plant ids. 

for example: 
```
bga_eia860[bga_eia860["plant_id_eia"]==52151]
```

|      |   plant_id_epa |   emissions_unit_id_epa |   plant_id_eia | boiler_id   | generator_id   |
|-----:|---------------:|------------------------:|---------------:|:------------|:---------------|
| 4302 |          52151 |                       1 |          52151 | PB1         | GEN1           |
| 4303 |          52151 |                       1 |          52151 | RF1         | GEN1           |
| 4304 |          52151 |                       1 |          52151 | PB2         | GEN2           |
| 4305 |          52151 |                       1 |          52151 | RF2         | GEN2           |

Vs. 

```
subplant_id[subplant_id["plant_id_epa"]==52151]
```
|       |   plant_id_eia |   plant_id_epa |   subplant_id |   unit_id_pudl |   emissions_unit_id_epa | generator_id   | boiler_id   |
|------:|---------------:|---------------:|--------------:|---------------:|------------------------:|:---------------|:------------|
| 24733 |          52151 |          52151 |             0 |              1 |                       1 | GEN1           | RF1         |
| 24734 |          52151 |          52151 |             0 |              2 |                       1 | GEN2           | RF2         |

cmgosnell:
>  I noticed that we merge the sub-plant id table with the bga table, but the m:m bga boiler-generator relationship is not retained after the sub-plant id creation. I'm not exactly sure why, but I think it has to do with how the network x is currently calculating the sub-plant ids. I have a feeling there are some (hopefully) easy tweaks to fix this, but I'm not super familiar with network x.

hm good catch! i was wondering about this. it makes a degree of sense honestly that the boiler_id could be ignored/condensed. Conceptually it makes sense that the subplant id doesn't need to retain the detail about the boilers. from the generation side of things it should mostly care about the generators because that is largely what the epa linked to the smokestacks. So my first inclination would not be to go in and change the network_x process. (it might be good to drop the `boiler_id` column before/right after the subplant_id making so this doesn't confuse anyone in the future.

I thiiiink this means that if we want the boilers we need a 1:m merge of the bga table after the fact. since rn we have two version of the epacamd_eia glue table it would be good to ponder whether or not we should make a specific `denorm_{whatever the table name is}` table to receive these boiler ids. and somewhat related but not necessary to make this new table is whether the `epacamd_eia` should be dropped as a db table - like does it provide any unique information from the subplant_id & this new potential denorm_ table with the boiler_ids

aesharpe:
Restarted the migrations because I was getting the error `raise ValueError("Constraint must have a name")` when I tried to create a new migration for this data off of the current `dev` branch. Conversations on slack in the technical questions channel indicated that this was the best course of action (for now).

src/pudl/etl/glue_assets.py
```diff
@@ -343,6 +344,28 @@ def epacamd_eia_subplant_ids(
...
+                "plant_id_eia",
+                "generator_id",
+                "boiler_id",
+                "unit_id_pudl",
+            ]
+        ].drop_duplicates(),
+        on=["plant_id_eia", "generator_id", "unit_id_pudl"],
+        how="outer",
+    ).drop_duplicates()
+
```
aesharpe, line 368: The reason that I added this chunk of code was because the network x process was dropping some of the boiler_ids (see comments above). This isn't pretty, but it remerges the subplant id table with the bga table so that we retain all the boiler-generator relationships. 

There are two main reasons it's not ideal: 

1) we already have a function called `augement_crosswalk_with_bga_eia860()` so it's weird / duplicative that we are merging the bga table with the subplant id table for a second time. (Can't find a way around this for now without changing how the network X is working which we agreed wasn't a good idea for right now). 

2) The `pytest test/unit/transform/glue_test.py::test_epacamd_eia_subplant_ids` was failing without the line that converts the `unit_id_pudl` field to a float. This step makes the tests pass, causes the asset build to fail -.- grr. I don't know how to get around it other than adding this column dtype conversion step into the tests rather than here. But that also feels janky. I would prefer to avoid this dtype issue if possible! It wasn't an issue before this merge was added which I don't really understand.

cmgosnell: on your second point, my first instinct is to add a `convert_dtypes()` step in the unit tests for all input dfs. hard agree that this convert to float is not ideal.

aesharpe:
> denorm_{whatever the table name is}

I think it would be really confusing to have a version of the subplant table with all the boilers and a version with just some or none. My vote is for a subplant table with all the boilers. 

> whether the `epacamd_eia` should be dropped as a db table - like does it provide any unique information from the subplant_id & this new potential denorm_ table with the boiler_ids

Technically it shouldn't provide anything new. If it does, we probably did something wrong. I think now that erroneous filling of `emissions_unit_id_pudl` has stopped it makes sense to just keep this subplant id table. 

It probably makes sense then to change the names to something like `epacamd_eia_annual` for the crosswalk we are removing from the database and keep `epacamd_eia_subplant_ids`.

src/pudl/etl/glue_assets.py
cmgosnell: and on #1... maybe we should slightly generalize the `augment_crosswalk_with_bga_eia860`. it looks like i? or you? idk added `boiler_id` into that merge. but if the nextwork_x step doesn't care about boilers/effectively drops boiler detail then they shouldn't be in the table pre-network_x anyway.

We could give `augment_crosswalk_with_bga_eia860` an `idx: list[str]` arg that is either just `["plant_id_eia", "generator_id"]` or `["plant_id_eia", "generator_id", "boiler_id"]` and always `boiler_generator_assn_eia860[idx + ["unit_id_pudl"]].drop_duplicates()`

aesharpe: I don't think this function is used more than once so we could also just call it `augment_crosswalk_with_unit_id_pudl` and remove `boiler_id` from the merge cols

aesharpe:
I just removed the `epacamd_eia` table from the db. The only novel information it contained that is _not_ encapsulated in the `subplant_id` table is the year that the crosswalk data was created with. The `epacamd_eia` table had a date column but the `subplant_id` table just takes the unique values form both years and calls it a day.

src/pudl/etl/glue_assets.py
```diff
@@ -389,11 +404,16 @@ def augement_crosswalk_with_bga_eia860(
...
+                "unit_id_pudl",
+                "boiler_id",
+            ]
         ].drop_duplicates(),
         how="outer",
-        on=["plant_id_eia", "generator_id"],
-        validate="m:1",
+        on=["plant_id_eia", "generator_id", "boiler_id"],
+        validate="m:m",
     )
```
cmgosnell, line 417: since this merge is happening pre-network x, it seems like we should probably revert these changes no? so there is no boiler_id pre-network_x.

cmgosnell (commented):
this all looks good! two hopefully small requests for tweaks. one here and one in comment land.

if we are removing the old `epacamd_eia` table (which i think is the right call!) can we rename the `epacamd_eia_subplant_id` -> `epacamd_eia`? its just such a nicer name.

src/pudl/etl/glue_assets.py
aesharpe: I don't think it really matters? Because we need to merge with the bga table for the `unit_id` which is used in the network x process. We could drop the boiler_id more explicitly before the process, but it wouldn't make a huge difference.

cmgosnell:
Im going to close this because its so old and we have no direct plans to work on it T_T

Issue #2338, opened 2023-02-27
Transform EIA 860 environmental equipment summary and association data


e-belfer:
## Short summary:
From 2014-2021, important summary information on each plant's associated control equipment is compiled in the Emissions Control Equipment table. From 2009-2021, the same information on IDs, cost, equipment type and operational status for SO2 and PM control is available in the table associated with each type of equipment in the EnvrEquipment dataset. More minimal information is also available on NOx and mercury controls. To effectively incorporate information on control equipment investments and operational status throughout time, we should pull the data from both the EnvrEquip tables (6.2 D-G) and the Emissions Control Equipment table, where available. When there is more than one source of information available for a given piece of control equipment, we should compare these values and harmonize them. 

The output of this result will be an Emissions Control Equipment table for 2009-2021, with a unique ID for each piece of control equipment, containing the following information:
- Control ID
- Control type (boolean for particulate type)
- Equipment type
- Operating month and year of the control equipment
- Operational status
- Acid gas control (boolean)
- Capital cost of equipment

```[tasklist]
### Tasks
- [ ] #2561 
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2586
```

## Preliminary Task Detail
(more in individual PRs)

**Metadata preparation**
- [x] Rename operating and retirement dates in pudl.package_data.eia860 to distinguish from generator operating/retirement dates

**Emissions Control Equipment Table (2014-2021)** 
- [x] Add all code columns to fields.py and define code metadata
- [x] Convert inservice month and year columns to a date
- [x] Minor QC on total cost column
- [x] Normalize “Acid Gas Control?” column into boolean 
- [x] Define a unique PUDL ID for each piece of equipment, and make a shared control_id_pudl column that acts as a primary key for this dataset. Propagate this column to other incorporated datasets.

**Association tables (2009-2021): Boiler Cooling, Boiler Particulate Matter, Boiler SO2, Boiler NOx, Boiler Mercury**
- [x] Add all code columns to fields.py and define code metadata
- [x] Minor cleaning (e.g., make sure no plant or control IDs are missing)
- [ ] Propagate unique PUDL ID from Emissions Control Equipment tab, or generate ID following same format for data prior to 2014 -- **need to do as part of cooling and stack flue equipment tables**

**EnvrEquipment tables (FGD, FGP, Cooling, Stack Flue)**
[...]

PR #2561, opened 2023-05-02
Emission control table

aesharpe:
# Add Emission Control Equipment & Boiler <> Emissions Control Equipment Tables

Based off issue #2338

## Changes
- Adds **`clean_emissions_control_equipment_eia860`** to the `/transform/eia860.py` module. Transforms include:
    - `fix_eia_na`
    - spot-fixing bad month values (depending on what we decide for month columns, I might not need the convert to NA one).
    - convert the `acid_gas_control` column to a boolean
    - spot fix a cost outlier in `emissions_control_equipment_cost`
    - utilize `pudl.metadata.classes.Package.from_resource_ids`.

- Adds **`clean_boiler_emissions_control_equipment_assn_eia860`** to the `/transform/eia860.py` module. Transforms include:
     - combine raw boiler emissions control association tables for nox, so2, particulate, and mercury
     - drop duplicate rows from different utilities (double counting)
     - drop rows with NA values for emissions control id

- Adds both of these tables to the harvesting process.

- Adds a new function to `pudl.helpers` called `convert_col_to_bool`. This function allows you to turn any column into a boolean by specifying what should be true and false. It preserves NA values. Also adds a unit test for this function.

- Adds relevant fields to `fields.py`

- Adds metadata to `/resources/eia860.py` for the new tables.

- Adds table **`denorm_emissions_control_equipment_eia860`** to `outputs/eia860`

## TO-DO
See more detailed notes about these to-dos in the code comments.

- [x] Finish up metadata
- [x] Decide which columns to include in the final version (ex: month/year vs. date and how to handle lost data)
- [x] Create an output table that gets persisted to the PUDL db
- [x] Decide how to handle `operational_status_code` column
- [x] Write some tests
- [x] Do the alembic change thing!!

# PR Checklist


src/pudl/metadata/resources/eia860.py
```diff
@@ -476,6 +476,36 @@
         "sources": ["eia860", "eia923"],
         "etl_group": "eia860",
     },
+    #     "emissions_control_equipment_eia860": {
+    #         "description": (
+    #             """
+    # The cost, type, operating status, retirement date, and install year of emissions control
+    # equipment reported to EIA. Includes control ids for sulfur dioxide (SO2), particulate
+    # matter, mercury, nitrogen oxide (NOX), and acid (HCl) gas monitoring.
+    # """
```
aesharpe, line 485: This whole section is commented out because it's not currently working (I think due to foreign key/harvesting stuff)

src/pudl/metadata/resources/eia860.py
```diff
@@ -476,6 +476,36 @@
...
+    #                 "emission_control_id_pudl",
+    #                 "data_maturity",
+    #                 "emission_control_equipment_type",
+    #                 "operational_status_code",
+    #                 "mercury_control_id_eia",
+    #                 "nox_control_id_eia",
+    #                 "particulate_control_id_eia",
+    #                 "so2_control_id_eia",
+    #                 "acid_gas_control",
+    #                 "emission_control_equipment_cost",
```
aesharpe, line 500: I excluded fields I know are foreign but not part of the primary key including: `plant_name_eia`, `utility_id_eia`, `utility_name_eia`, and `steam_plant_type_code`. I'm assuming that not including them here means they are not included in the final version of the table. This assumption means that I don't alter or drop these columns in the transform module. Does that make sense?

src/pudl/metadata/resources/eia860.py
```diff
@@ -476,6 +476,36 @@
...
+    #             ],
+    #             "primary_key": ["report_year", "plant_id_eia", "emission_control_id_pudl"],
+    #             "foreign_key_rules": {},
```
aesharpe, line 503: I don't really understand how this works @cmgosnell maybe you can help me here?

src/pudl/transform/eia860.py
```diff
@@ -748,3 +748,51 @@ def clean_boilers_eia860(
...
+    raw_emissions_control_equipment_eia860: pd.DataFrame,
+) -> pd.DataFrame:
+    """Pull and transform the emissions control equipment table."""
+    # Replace empty strings, whitespace, and '.' fields with real NA values
+    emce_df = pudl.helpers.fix_eia_na(raw_emissions_control_equipment_eia860)
+
+    # Fix bad months
+    emce_df["operating_month"] = emce_df["operating_month"].replace({"88": "8"})
+    emce_df["retirement_month"] = emce_df["retirement_month"].replace({"True": np.nan})
+
```
aesharpe, line 764: I'm not sure what to do about the date columns pertaining to operation and retirement. There's a month and a year column for each of them. In the past we've mostly converted these types of columns into a single Datetime column using the helper `convert_to_date` function. The issue with this is that A) it adds a "day" to the date that just ends up being the first of the month -- not a huge deal and relatively standard practice, just worth nothing. More importantly, B) it will null out rows that have a `year` value but not a `month` value. Even if we decide to change that, Datetime still requires M/D/Y and I feel less good about defaulting to January as I do the first day of the month. My question here is whether we want to keep these columns separate (month/year) or create a new Datetime column and if so, what we should do about rows with no month data.

src/pudl/transform/eia860.py
```diff
@@ -748,3 +748,51 @@ def clean_boilers_eia860(
...
+    # Convert column to boolean
+    emce_df = pudl.helpers.convert_col_to_bool(
+        df=emce_df, col_name="acid_gas_control", true_values=["Y"], false_values=[]
+    )
+    # Add a emission_control_id_pudl as a primary key. This is not unique over years.
+    # We could maybe try and do this, but not doing it now.
+    emce_df["emission_control_id_pudl"] = (
+        emce_df.groupby(["report_year", "plant_id_eia"]).cumcount() + 1
+    )
```
aesharpe, line 773: Decided not to waste any effort trying to make this value identifiable over time.

aesharpe:
I want to convert the `operating_status_code` column into values vs. codes, but before doing that I have some questions:

1) Generally speaking, where it is appropriate to convert data using `/metadata/codes.py`? Should this happen as part of the transform step? This makes the most sense to me (and is how it works for the `generators_eia860 table`), but I wanted to make sure there wasn't another way to enforce codes akin to dtype conversions or enums. 

2) We currently have `CODE_METADATA` for EIA under `operational_status_eia`. These codes mostly line up with the operational status codes for the emissions control equipment with one exception: `OZ = Operated only during the ozone season (May through September)`. I'm inclined to add this to the list of codes, but I wanted to check that it was ok. 

3) I compared the operational status code definitions used throughout EIA for boilers, generators, cooling systems, stacks, and emissions control equipment. They are almost all the same (yay). However, the language is slightly different for generators as for the other types of equipment. Right now, the `operational_status_eia` code descriptions align with the generator descriptions. I was wondering if we should expand them to include the other, adjacent definitions too:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/austensharpe/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/austensharpe/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">



  | OTHER DESCRIPTIONS | GENERATOR DESCRIPTION
-- | -- | --
OP | Operating (in commercial service   or out of service within 365 days) | Operating - in service   (commercial operation) and producing some electricity. Includes peaking units   that are run on an as needed (intermittent or seasonal) basis.
OS | Out of service (365 days or   longer) | Out of service – was not used   for some or all of the reporting period and is NOT expected to be returned to   service in the next calendar year.
OA | NA | Out of service – was not used   for some or all of the reporting period but is expected to be returned to   service in the next calendar year.
PL | Planned (on order or expected to   go into commercial service within 10 years) | NA
RE | Retired (no longer in service   and not expected to be returned to service) | Retired - no longer in service   and not expected to be returned to service.
SB | Standby (or inactive reserve,   i.e., not normally used, but available for service) | Standby/Backup - available for   service but not normally used (has little or no generation during the year)   for this reporting period.



</body>

</html>

src/pudl/transform/eia860.py
```diff
@@ -748,3 +748,51 @@ def clean_boilers_eia860(
...
+    # Fix outlier value in emission_control_equipment_cost. We know this is an
+    # outlier because it is the highest value reported in the dataset and it
+    # the other years from the same plant show that it likely contains three
+    # extra zeros. We use the primary keys to spot fix the value.
```
aesharpe, line 777: I could also do some analysis to see whether there are different units being reported here (dollars/thousands of dollars) but I haven't done that yet.

src/pudl/metadata/resources/eia860.py
cmgosnell: that does make sense. if you don't want these columns in the published table, the enforce_schema step - which is always applied - will drop them. all of transformed assets made in transform.eia860 are fed into the harvesting process automatically so the values from this table are being slurped up then. which is all good.

src/pudl/metadata/resources/eia860.py
cmgosnell: I don't think you want `foreign_key_rules` on this table. The fk rules say: if you find this tables PKs in *other* tables, you should create a FK relationship in that table. You can also `exclude` tables if need be. You can look at the `boilers_eia860` tables metadata as an example of both.

src/pudl/transform/eia860.py
cmgosnell: a) we default to first day of the month all over the place so i am pro keeping that convention and keeping it standard.
b) yeesh this is worse for sure.... I think these should either be null of January. We want a date column as this is consistent throughout our data. I'm personally pro January of year so there is some info in there but idk

src/pudl/transform/eia860.py
```diff
@@ -748,3 +748,51 @@ def clean_boilers_eia860(
...
+    outlier_primary_keys = (
+        (emce_df["report_year"] == 2017)
+        & (emce_df["plant_id_eia"] == 57794)
+        & (emce_df["emission_control_id_pudl"].isin([1, 2]))
```
cmgosnell, line 781: if these are not stable you probably shouldn't do this

src/pudl/metadata/resources/eia860.py
aesharpe: Hmmm ok. The primary keys are `report_year`, `plant_id_eia`, and `emission_control_id_pudl`. Are there any circumstances were you would report part of the primary key (i.e., the first two cols) because the last one is made specifically for this table? Or would you only list them if all three were found in another table that you specifically wanted to link to?

src/pudl/transform/eia860.py
aesharpe: Ok that's what I'll do. I'll make sure the metadata for this field explains this caveat.

src/pudl/transform/eia860.py
aesharpe: That's true...but index also doesn't feel stable, and those are the primary keys so I don't know of another way to identify the correct row. the `emission_control_id_pudl ` is not table over time, but within a given year/plant group it shouldn't change unless we sort the data in a new way or new data gets added to this specific year/plant group retroactively.

src/pudl/transform/eia860.py
aesharpe: I took a closer look at the instances where there was an `operating_year` but not `operating_month` and was able to spot fix all of them based on future years. So I feel cozy doing the date thing!

src/pudl/transform/eia860.py
zschira: It feels to me like this will probably be fine as is, but maybe you could check that `emission_control_equipment_cost` equals the expected outlier value, just to make sure that nothing strange happens.

src/pudl/metadata/resources/eia860.py
```diff
@@ -476,6 +476,42 @@
         "sources": ["eia860", "eia923"],
         "etl_group": "eia860",
     },
+    "denorm_emissions_control_equipment_eia860": {
```
zschira, line 479: Is there a reason you've only added the `denorm` table and not the normalized one?

zschira (changes requested):
Looks good, just a couple of small questions!

zschira:
I don't have the best understanding of the `Encoder` stuff, but I don't *think* you need to call `encode` here.

src/pudl/metadata/resources/eia860.py
aesharpe: I guess I wasn't sure what the protocol was for our table storage yet. But I could easily add it.

aesharpe: I added this table to the harvesting process!

aesharpe:
Question for @jrea-rmi or @arengel or @cmgosnell: 

The 860 EnviroAssoc table includes tabs for:
- boiler <> generator associations
- boiler <> nox emissions equipment association
- boiler <> so2 emissions equipment association
- boiler <> particulate emissions equipment association
- boiler <> mercury emissions equipment association
- boiler <> cooler association
- boiler <> stack / flue associations

For the association table included in this PR I combined all the IDs for the **emissions control equipment**

Do you have any interest in tables for the boiler <> cooler associations or boiler <> stack / flue associations while I'm in here (we obviously already have the boiler <> generator table).

test/validate/eia_test.py
```diff
@@ -42,15 +45,18 @@ def test_no_null_cols_eia(pudl_out_eia, live_dbs, cols, df_name):
...
-        ("boil_eia860", 74_086, 74_086, 74_086),
+        ("boil_eia860", 74_169, 74_169, 74_169),
         ("frc_eia923", 597_000, 244_415, 24_065),
         ("gen_eia923", None, 5_171_497, 432_570),
         ("gens_eia860", 523_563, 523_563, 523_563),
         ("gf_eia923", 2_687_345, 2_687_345, 230_149),
         ("own_eia860", 84_440, 84_440, 84_440),
-        ("plants_eia860", 185_553, 185_553, 185_553),
-        ("pu_eia860", 184_743, 184_743, 184_743),
-        ("utils_eia860", 119_366, 119_366, 119_366),
```
aesharpe, line 53: The difference in rows here comes from the fact that some utilities aren't showing up in the utils_eia860 or pu_eia860 tables in later years. None of these new rows are from new utilities, just utilities that didn't report certain years. This is only a little annoying because it means that the utils_eia860 table will have a certain years of utilities that don't have all the standard information you'd find in the utils_eia860 table like address etc. We could potentially backfill. Not doing anything about it now, just wanted to flag!

arengel:
> Question for @jrea-rmi or @arengel or @cmgosnell:
> 
> The 860 EnviroAssoc table includes tabs for:
> 
> * boiler <> generator associations
> * boiler <> nox emissions equipment association
> * boiler <> so2 emissions equipment association
> * boiler <> particulate emissions equipment association
> * boiler <> mercury emissions equipment association
> * boiler <> cooler association
> * boiler <> stack / flue associations
> 
> For the association table included in this PR I combined all the IDs for the **emissions control equipment**
> 
> Do you have any interest in tables for the boiler <> cooler associations or boiler <> stack / flue associations while I'm in here (we obviously already have the boiler <> generator table).
> 
> **For now, just keeping it to the emissions equipment.**

I'd say do them now if it's not too much incremental effort. Stack / flue data is relevant for pollution -> health impact modeling and cooling could help us understand water availability for alternative uses of the site, among other things of course.

src/pudl/transform/eia860.py
```diff
@@ -748,3 +748,187 @@ def clean_boilers_eia860(
...
+        table = table.drop_duplicates(
+            subset=[
+                x
+                for x in table.columns
+                if x not in ["utility_id_eia", "utility_name_eia"]
+            ]
+        )
+        value_col = [col for col in table if "control_id_eia" in col]
+        id_cols = [col for col in table if "control_id_eia" not in col]
+        assert len(value_col) == 1
```
zschira, line 913: Could you add a little comment explaining this check? My understanding is that each table should have exactly one column with `control_id_eia` in it's name somewhere, but I find it helpful to have explicit documentation of these inline checks

src/pudl/metadata/fields.py
```diff
@@ -436,6 +440,22 @@
         "description": "Purchase type under which receipts occurred in the reporting month. C: Contract, NC: New Contract, S: Spot Purchase, T: Tolling Agreement.",
         "constraints": {"enum": ["S", "C", "NC", "T"]},
     },
+    "emission_control_id_eia": {
+        "type": "string",
+        "description": (
+            "The emission control ID used to collect so2, nox, particulate, "
+            "and mercury emissions data. This column shoudl be used in conjunction "
```
zschira, line 447: *should

src/pudl/metadata/fields.py
```diff
@@ -1493,15 +1537,15 @@
     },
     "operational_status": {
         "type": "string",
-        "description": "The operating status of the generator. This is based on which tab the generator was listed in in EIA 860.",
+        "description": "The operating status of the asset. For generators rhis is based on which tab the generator was listed in in EIA 860.",
```
zschira, line 1540: this

zschira (changes requested):
Overall looks good to me! A couple little typos and things.

As for the code coverage, I'm not personally too worried about it given that it's a decrease of 0.1%, and many of the changes are to metadata.

zschira (approved):
Approved!

PR #2564, opened 2023-05-04
Integrate FERC Form 2 dbf formats into ferc_to_sqlite

rousik:
This adds the capability to extract FERC Form 2 years 1996-2020 from dbf archives.


zaneselvans:
@rousik is the comment on this PR still accurate? It looks more like it allows data from 1996-2021 to be converted from DBF to SQLite. Does it do the older years too?

rousik:
> @rousik is the comment on this PR still accurate? It looks more like it allows data from 1996-2021 to be converted from DBF to SQLite. Does it do the older years too?

Updated the comment. Prior to 1996, the split respondent archives is just a collection of DAT files, doesn't seem to be DBF at all.

src/pudl/extract/ferc.py
```diff
@@ -0,0 +1,33 @@
...
+) -> sa.MetaData:
+    """Adds primary and foreign key to tables present in meta.
+
+    Args:
+        meta: constraints will be applied to this metadata instance
+        pk_table: name of the table that contains primary-key
+        column: foreign key column name. Tables that contain this column will
+            have foreign-key constraint added.
+        pk_column: (optional) if specified, this is the primary key column name in
+            the table. If not specified, it is assumed that this is the same as pk_column.
```
zaneselvans, line 17: It is assumed that this is the same as `column` not `pk_column` right?

src/pudl/extract/ferc.py
```diff
@@ -0,0 +1,33 @@
...
+
+def add_key_constraints(
+    meta: sa.MetaData, pk_table: str, column: str, pk_column: str | None = None
+) -> sa.MetaData:
+    """Adds primary and foreign key to tables present in meta.
+
+    Args:
+        meta: constraints will be applied to this metadata instance
+        pk_table: name of the table that contains primary-key
+        column: foreign key column name. Tables that contain this column will
```
zaneselvans, line 14: In general these could be composite keys composed of multiple columns (e.g. `report_year` and `respondent_id`). Thus far the only keys we've been able to make work are the simple `respondent_id` and even that has required some repair of its home table.

If it would be easy to let these be composite keys rather than single columns that would be good. If it's not easy, then we should at least note that only simple keys are possible and why that's been okay for the FERC DBF data so far.

src/pudl/ferc_to_sqlite/__init__.py
```diff
@@ -14,11 +14,23 @@
...
+    """Clone the FERC FoxPro databases into SQLite."""
+    dbf2sqlite()
+
+
+@graph
+def ferc_to_sqlite_xbrl_only():
+    """Clone the FERC XBRL databases into SQLite."""
+    xbrl2sqlite()
+
+
```
zaneselvans, line 33: Would it be easy to break these out both by data format and by individual forms?  I think in development we'd typically be working with a single form and format at a time, and being able to just regenerate that one SQLite DB in isolation would be convenient.

I guess you've created this possibility with the "disabled" option.

src/pudl/metadata/sources.py
```diff
@@ -400,7 +400,10 @@
         ),
         "field_namespace": "ferc2",
         "working_partitions": {
-            "years": sorted(set(range(2021, 2022))),  # XBRL only
+            # Years 1991-1995 use strange formats that need to be investigated further.
+            # Years 1996-1999 come in split archives and full archives and we are going
+            # to be using the aggregated archives (part=None).
+            "years": sorted(set(range(1996, 2022))),
```
zaneselvans, line 406: 🎉

src/pudl/package_data/ferc2/table_file_map.csv
```diff
@@ -0,0 +1,118 @@
+table,filename
```
zaneselvans, line 1: Much more uniform naming in Form 2 than Form 1. Maybe they learned things!

src/pudl/settings.py
```diff
@@ -476,6 +483,23 @@ class Ferc2XbrlToSqliteSettings(FercGenericXbrlToSqliteSettings):
     taxonomy: AnyHttpUrl = "https://eCollection.ferc.gov/taxonomy/form2/2022-01-01/form/form2/form-2_2022-01-01.xsd"
 
 
+class Ferc2DbfToSqliteSettings(GenericDatasetSettings):
```
zaneselvans, line 486: It seems like we should have a validator in this and the other DBF and XBRL to SQLite settings classes that behaves like the one in `Ferc1Settings` such that if someone requests an invalid year of data we can either fix it or raise an error.

test/conftest.py
```diff
@@ -193,6 +193,46 @@ def pudl_out_orig(live_dbs, pudl_engine):
     return PudlTabl(pudl_engine=pudl_engine)
 
 
+@pytest.fixture(scope="session")
+def ferc_to_sqlite_dbf_only(live_dbs, pudl_datastore_config, etl_settings, pudl_env):
```
zaneselvans, line 197: Would it be easy & make sense to further break these down by both data format and individual form? (which would require breaking down the underlying jobs as well)

test/integration/ferc_dbf_extract_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_ferc1_dbf2sqlite(ferc1_engine_dbf):
+    """Attempt to access the DBF based FERC 1 SQLite DB fixture."""
+    assert isinstance(ferc1_engine_dbf, sa.engine.Engine)  # nosec: B101
+    assert (  # nosec: B101
+        "f1_respondent_id" in sa.inspect(ferc1_engine_dbf).get_table_names()
+    )
+
+
+def test_ferc1_schema(ferc_to_sqlite_settings, pudl_datastore_fixture):
```
zaneselvans, line 20: Can we / should we create similar integration tests for the extraction of the other forms?  Could they be easily parameterized to just do the same thing for each of them?

src/pudl/ferc_to_sqlite/cli.py
```diff
@@ -87,23 +87,42 @@ def parse_command_line(argv):
...
+    loglevel: str = "INFO",
+    enable_xbrl: bool = True,
+    enable_dbf: bool = True,
 ) -> Callable[[], JobDefinition]:
     """Factory for parameterizing a reconstructable ferc_to_sqlite job.
 
     Args:
         loglevel: The log level for the job's execution.
         logfile: Path to a log file for the job's execution.
+        enable_xblr: if True, include XBRL data processing in the job.
```
zaneselvans, line 100: `enable_xblr` -> `enable_xbrl`

src/pudl/ferc_to_sqlite/cli.py
```diff
@@ -87,23 +87,42 @@ def parse_command_line(argv):
...
+        enable_dbf: if True, include DBF data processing in the job.
 
     Returns:
         The job definition to be executed.
     """
+    if not (enable_xbrl or enable_dbf):
```
zaneselvans, line 106: The way we're splitting up or specifying which data to translate into SQLite feels messy to me in general. Not sure if there's a better way to do it though.

Right now in the ETL settings we ask the user for datasets and years, and then figure out whether that means translating from XBRL or DBF internally, which seems correct -- ideally the user wouldn't need to know which source the data is coming from.  But with 5 FERC forms (1, 2, 6, 60, 714) involved and 2 data formats for each of them (though the older 714 data isn't distributed as DBF for some reason) it feels like having some clear enumeration of the chunks of work that are going to get done internally would be more readable than having this high level job switch that says do DBF or do XBRL, and then another lower layer that's only visible in the settings that says which datasets to translate, and then yet another layer within the settings beneath that which indicates which partitions (years) to translate.

I guess what I'm imagining is that instead of just pulling the DBF vs. XBRL processing up to this high level, it might also make sense to pull both the dataset and data format up to this level and treat each dataset + data format as a separate job, since they'll each result in a separate SQLite output, and in a development context we'll often just want to re-run one of them in isolation. What do you think?

src/pudl/extract/ferc2.py
```diff
@@ -0,0 +1,49 @@
...
+
+import pudl
+from pudl.extract.dbf import FercDbfExtractor
+from pudl.extract.ferc import add_key_constraints
+from pudl.settings import FercToSqliteSettings, GenericDatasetSettings
+
+logger = pudl.logging_helpers.get_logger(__name__)
+
+
+class Ferc2DbfExtractor(FercDbfExtractor):
```
zaneselvans, line 25: So concise!

src/pudl/extract/ferc1.py
```diff
@@ -350,13 +347,19 @@ def dbf2sqlite(context) -> None:
     """Clone the FERC Form 1 Visual FoxPro databases into SQLite."""
```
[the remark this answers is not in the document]
zaneselvans: Yeah I totally agree. I think this is an artifact of only have 1 dataset to work with early on. Especially as the amount of XBRL data grows over time being able to separate them out will be very helpful in development, since the XBRL processing is quite slow compared to DBF (though also much more parallelizable currently)

zaneselvans (commented):
I love how concise the per-form extractors ultimately are! Just a little metadata compilation and some optional fixes.

I left a bunch of comments, but it seems like maybe one higher level design question we both have is whether it would be better to split out the individual FERC forms as separately addressible jobs / ops / processes rather than lumping all the DBF and all the XBRL together.

src/pudl/extract/dbf.py
```diff
@@ -70,6 +72,108 @@ def create_sa_table(self, sa_meta: sa.MetaData) -> sa.Table:
...
+        """
+        sch = self.get_table_schema(table_name)
+        df = pd.DataFrame(iter(self.get_table_dbf(table_name)))
+        df = df.drop("_NullFlags", axis=1, errors="ignore").rename(
+            sch.get_column_rename_map(), axis=1
+        )
+        return df
+
+
 class AbstractFercDbfReader(Protocol):
```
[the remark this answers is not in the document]
rousik: I think the idea here is that whenever something has these methods, it will be of type `AbstractFercDbfReader` even if it's not explicitly its subclass. In essence, I think this allows us to create "interface definitions" that can then be passed as types to methods/objects that consume this and it can ensure that it's called in the right way.

src/pudl/extract/dbf.py
```diff
@@ -70,6 +72,108 @@ def create_sa_table(self, sa_meta: sa.MetaData) -> sa.Table:
...
 
+class FercDbfArchive:
+    """Represents API for accessing files within a single DBF archive."""
+
+    def __init__(
+        self,
+        zipfile: zipfile.ZipFile,
+        dbc_path: Path,
+        table_file_map: dict[str, str],
+        partition: dict[str, Any],
```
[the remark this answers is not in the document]
rousik: Additionally, this `partition` is only used for logging purposes, so we could do away with it completely.

src/pudl/extract/dbf.py
```diff
@@ -70,6 +72,108 @@ def create_sa_table(self, sa_meta: sa.MetaData) -> sa.Table:
...
+        field_parser: FieldParser,
+    ):
+        """Construct new instance of FercDbfArchive."""
+        self.zipfile = zipfile
+        self.partition = (dict(partition),)
```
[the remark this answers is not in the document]
rousik: Uh, this actually seems incorrect! Fixed.

src/pudl/extract/dbf.py
```diff
@@ -70,6 +72,108 @@ def create_sa_table(self, sa_meta: sa.MetaData) -> sa.Table:
...
+            fname,
+            encoding="latin1",
+            parserclass=self.field_parser,
+            ignore_missing_memofile=True,
+            filedata=self.get_file(fname),
+        )
+
+    @lru_cache
+    def get_table_schema(self, table_name: str) -> DbfTableSchema:
+        """Returns TableSchema for a given table and a given year."""
```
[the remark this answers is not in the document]
rousik: Yep. More precisely, `FercDbfArchive` is associated with single zipfile archive, and each year should map to one of those (although some years could map to several archives, but even then, only one of those is useful)

src/pudl/extract/dbf.py
```diff
@@ -193,122 +306,61 @@ def _open_csv_resource(self, base_filename: str) -> csv.DictReader:
...
-        fname = self._table_file_map[table_name]
-        fd = self._get_file(year, fname)
-        return DBF(
-            fname,
-            encoding="latin1",
-            parserclass=self.field_parser,
-            ignore_missing_memofile=True,
-            filedata=fd,
+    @lru_cache
+    def get_archive(self, year: int, **filters) -> FercDbfArchive:
```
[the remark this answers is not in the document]
rousik: Ah, this is little confusing. We need to know the `year` here to pick the right path from `self._dbc_path` but note that when we call get_zipfile_resource, we pass both `**nfilters` as well as `year` so technically, we are selecting by both of these. We are also assuming that any partitioning scheme should have `year` as one of its components, so when you call `get_archive(**partition)`, it should map the year dimension to `year: int` here.

src/pudl/extract/dbf.py
```diff
@@ -374,6 +436,13 @@ def get_db_path(self) -> str:
...
     def execute(self):
         """Runs the extraction of the data from dbf to sqlite."""
+        logger.info(
+            f"Running dbf extraction for {self.DATASET} with settings: {self.settings}"
+        )
+        if self.settings.is_disabled:
+            logger.warning(f"Dataset {self.DATASET} extraction is disabled, skipping")
+            return
+
         # TODO(rousik): perhaps we should check clobber here before starting anything.
```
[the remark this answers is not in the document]
rousik: Yep. This is checked in `pudl.helpers.drop_tables` which is invoked by `delete_schema`. We could move this check here and simply test for existence and if `clobber` is true, delete that file. That way we may even get away with avoiding `delete_schema` operation altogether. AFAIK we should always operate on a file-level granularity so there should not be need for fine-grained table/schema deletion. What do you think?

src/pudl/ferc_to_sqlite/cli.py
rousik: Yeah. I agree that having one job per dataset/format combination would be the most natural structure. Let me refactor this accordingly.

test/integration/ferc_dbf_extract_test.py
rousik: I have parametrized this so the same test is run for both `ferc1` and `ferc2`.

src/pudl/ferc_to_sqlite/cli.py
rousik: After some early tinkering with this, I feel like I will need to dig deeper into dagster and how it works and how to properly structure this. I think that it should be possible to auto generate some of these dagster ops/graphs programmatically which should reduce the boilerplate, but it might take a little extra effort and care to do so.

My suggestion would be to proceed with this PR as-is and follow up with cleanup/refactoring independently of this. What do you think about that? I can file an issue to track this.

rousik:
When attempting to diff the output I found out that form 2 requires use of `--clobber` because the current production dataset does not contain `part=N` partition annotation which causes non-uniqueness violations when opening the archive. So, perhaps we should bump the production versions before merging this.

zaneselvans:
New FERC 2 archive up here: https://zenodo.org/record/8006881

src/pudl/extract/dbf.py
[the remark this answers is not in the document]
rousik: Yeah, exactly.

src/pudl/ferc_to_sqlite/cli.py
zaneselvans: Sure, I think that's beyond the scope of this PR for right now, but it'd be good to make it more dynamic / granular at some point.  And we don't extract the FERC DBs very frequently, so this won't come up too often.

zaneselvans (approved):
Exciting!

zaneselvans:
Oh, probably want to add the new FERC 2 SQLite DB to the list of DBs which are published to Datasette in `devtools/datasette/publish.sh`

zaneselvans (changes requested):
I ran the ETL  locally including the new FERC 2 DBF to SQLite conversion and I inspected 10 tables and found that none of them contained data from before 2004.  Looking at the input file sizes, it seems like the archives become sizeable (~100MB) starting in 2002 and are only a few MB each from 1996-2001, but that doesn't seem to line up with the lack of data in 2002-2003 so I'm wondering if something is wrong with the extraction of earlier tables? Is there a reason why we might expect to see this behavior?

```py
ferc2_db_path = "sqlite:////Users/zane/code/catalyst/pudl-work/output/ferc2.sqlite"
ferc2_engine = sa.create_engine(ferc2_db_path)
md = sa.MetaData()
md.reflect(bind=ferc2_engine)
ferc2_dfs = {}
for t in md.tables:
    ferc2_dfs[t] = pd.read_sql(t, ferc2_engine)
{t: ferc2_dfs[t].report_yr.min() for t in ferc2_dfs if "report_yr" in ferc2_dfs[t].columns}
```

Issue #2586, opened 2023-05-19
Transform boiler-cooling and boiler-stack-flue association tables


aesharpe:
The Environmental association table from EIA860 also has boiler stack flue and boiler cooler associations. This issue is to integrating those into the PUDL db.

Addressed in #2587

PR #2587, opened 2023-05-19
Boiler cooler stack flue association tables

aesharpe:
# Overview

This PR adds the boiler <> cooling id and boiler <> stack flue association tables to the database. Addresses #2586

Still a WIP:
- [ ] Figure out how to handle stack flue primary key
- [ ] Add both tables to the harvesting process
- [ ] Update any tests that are altered by these tables getting harvested

# PR Checklist


aesharpe:
**QUESTION:** 
@zschira @zaneselvans @cmgosnell 

The stack flue association table contains the following columns: `flue_id_eia`, `stack_id_eia`, and `stack_flue_id_eia`. Sometimes the stack id and the flue id are difference and a record will have a value for `stack_id_eia` and `flue_id_eia`. Sometimes the stack and flue are inextricable and the record has a `stack_flue_id_eia`. There are no records that have a `stack_flue_id_eia` and also a `stack_id_eia` or `flue_id_eia`. 

This makes things complicated for creating a primary key, which can't be NA. `stack_flue_id_eia` will always be NA when there are values for `stack_id_eia` and `flue_id_eia`. `stack_id_eia` and `flue_id_eia` can't be combined because they represent distinct entities that have a many-to-many relationship with one another. 

Options I've considered:
- Create a `stack_flue_id_pudl` column that would create unique values for each of the rows. This feels kind of weird because it's combining entities that are also linked to other things.
- Break this table down into boiler <> stack id and boiler <> flue id tables with the `stack_flue_id_eia` column values used in both of the tables under either `stack_id` or `flue_id`.

If anyone else has other ideas I would love to know

e-belfer:
I'm not 100% sure if I'm following this comment:

> This feels kind of weird because it's combining entities that are also linked to other things.

But, I would lean towards making a `stack_flue_id_pudl` column that concatenates the stack and flue IDs (e.g. 1A-1-2) where they are separate and takes the `stack_flue_id_eia` column where it has a value. This seems consistent with how we combine other columns. It also makes it relatively trivial to identify the components.

aesharpe:
> If the goal of the table is to conclusively link each boiler to 1 stack and 1 flue, and the dataset pivots towards having a combined stack-flue ID in later years, then I would aim to reproduce this logic retroactively.

Unfortunately, it's not something that changes in later years. Rather, some plants have a combined stack and flue system and others are separate and interconnected with other stacks or flues.

src/pudl/metadata/fields.py
```diff
@@ -2241,6 +2249,31 @@
...
+        "type": "string",
+        "description": (
+            "The value EIA uses to identify hybrid stack and flue units. Plants with "
+            "individual stack and flue IDs do not have a stack_flue_id value."
+        ),
+    },
+    "stack_flue_id_pudl": {
+        "type": "string",
+        "description": (
+            "The stack_id_eia value and the flue_id_eia value seperated by an "
```
e-belfer, line 2269: I would suggest something like this: 
"A unique identifier of the stack and/or flue unit, produced by PUDL. If there is an ID for a combined stack-flue unit (stack_flue_id_eia), this value is used. Otherwise, this value is a combination of the EIA stack ID (stack_id_eia) and the flue ID (flue_id_eia), separated by an underscore. If there is no flue ID, this is equivalent to the stack ID provided by EIA. This value is used as the primary key for the boiler_stack_flue_assn_eia860 table and shouldn't necessarily be used to connect the values in this table to those of another."

With regards to the last line, do we not want to add this column to the other cooling and stack flue tables to enable joining by this column?

src/pudl/metadata/resources/eia860.py
```diff
@@ -546,7 +546,7 @@
     },
     "boiler_emissions_control_equipment_assn_eia860": {
         "description": (
-            """A table that links EIA boiler id to emissions control IDs for nox, so2,
+            """A table that links EIA boiler ID to emissions control IDs for nox, so2,
```
e-belfer, line 549: Should be NOx and SO2, and "EIA boiler IDs".
Re: the 1:many relationship, is this really 1:many or is this actually an m:m relationship?

src/pudl/metadata/resources/eia860.py
```diff
@@ -571,6 +571,54 @@
         "sources": ["eia860"],
         "etl_group": "eia860",
     },
+    "boiler_cooler_assn_eia860": {
+        "description": "A table that links EIA boiler ID to EIA cooling system ID.",
```
e-belfer, line 575: Same here and throughout, think this should be "IDs". Otherwise descriptions look clear.

migrations/versions/6e9328b6b275_.py
```diff
@@ -0,0 +1,24 @@
+"""empty message
```
e-belfer, line 1: What does this revision do? It appears to point to two heads. Perhaps this interfaces with the PR @zschira wrote up about revising the two-headed nightly build?

src/pudl/metadata/fields.py
```diff
@@ -871,6 +875,10 @@
         "type": "string",
         "description": "EIA short code indicating the type of firing used by this boiler.",
     },
+    "flue_id_eia": {
+        "type": "string",
+        "description": "test",
```
e-belfer, line 880: This seems to be a leftover draft. Should likely be "The flue identification number provided by EIA" or something similar.

src/pudl/metadata/fields.py
```diff
@@ -2241,6 +2249,31 @@
         "type": "boolean",
         "description": "Indicates whether the generator is part of a solid fuel gasification system",
     },
+    "stack_id_eia": {
+        "type": "string",
+        "description": (
+            "The value EIA uses to identify a stack or chimney on a power plant where "
```
e-belfer, line 2255: One non-blocking question: Is this the value used by EIA, or the value that an individual plant owner uses to identify this unit, as reported by the EIA? I wonder if the wording of these descriptions is slightly misleading, as we're intending to point to the data source rather than the original creator of these IDs.

src/pudl/transform/eia860.py
```diff
@@ -941,3 +941,70 @@ def clean_boiler_emissions_control_equipment_assn_eia860(
...
+
+
+@asset
+def clean_boiler_stack_flue_assn_eia860(
+    raw_boiler_stack_flue_eia860: pd.DataFrame,
+) -> pd.DataFrame:
+    """Pull and transform the EIA 860 boiler to stack flue ID table.
+
+    Args:
+        raw_boiler_booling_eia860: Raw EIA 860 boiler to stack flue ID association
```
e-belfer, line 978: Arg name should match the input name. Description looks good though!

src/pudl/transform/eia860.py
```diff
@@ -941,3 +941,70 @@ def clean_boiler_emissions_control_equipment_assn_eia860(
...
+
+
+@asset
+def clean_boiler_cooler_assn_eia860(
+    raw_boiler_cooling_eia860: pd.DataFrame,
+) -> pd.DataFrame:
+    """Pull and transform the EIA 860 boiler to cooler ID table.
+
+    Args:
+        raw_boiler_booling_eia860: Raw EIA 860 boiler to cooler ID association table.
```
e-belfer, line 953: Arg name should match the input name. Description looks good though!

src/pudl/transform/eia860.py
```diff
@@ -941,3 +941,70 @@ def clean_boiler_emissions_control_equipment_assn_eia860(
...
+    # have unique IDs for each stack & flue unit that may have m:m relationships with
+    # one another. These IDs are part of the primary key for this table, and because we
+    # can't have NA values in the primary key we'll need to create a hybrid column that
+    # turns the seperate stack_id_eia and flue_id_eia values into one stack_flue_id_pudl
+    # value. Anytime there's a stack_flue_id_eia we'll use that.
+    bsf_assn = bsf_assn.assign(
+        stack_flue_id_pudl=lambda x: (
+            x.stack_id_eia.astype("string") + "_" + x.flue_id_eia.astype("string")
+        )
+        .fillna(x.stack_flue_id_eia)
```
e-belfer, line 1004: There aren't any overlaps in the two sets of columns, correct? Otherwise I would suggest prioritizing filling with the combined ID over the two separate ID columns by reordering the order of assignment.

src/pudl/transform/eia860.py
```diff
@@ -941,3 +941,70 @@ def clean_boiler_emissions_control_equipment_assn_eia860(
...
+        .fillna(
+            x.stack_id_eia.astype("string")
+        )  # sometimes there's a stack id and no flue id
+    )
```
e-belfer, line 1008: Do you want to drop records without a PUDL stack/flue ID here, as you do with the other table?

e-belfer (changes requested):
No blocking comments here, other than fixing that "test" description. I haven't yet re-run the full ETL or the tests, and it seems like the last of the kinks are still being worked out here. The vast majority of my comments have to do with minor doc changes, and a leftover in-progress comment that should get updated. All assets appear in dagster as expected and materialize without issue.

A flag for future integration: we should be sure to add the PUDL stack/flue ID column into the `stack_flue_equipment` table when we integrate it.

src/pudl/metadata/fields.py
aesharpe: We certainly could, I just want to preserve/emphasize the fact that the relationship between stacks and flues is not necessarily 1:1.

aesharpe: Also I don't think it's necessarily a unique identifier either--just when paired with the `report_date` and `plant_id_eia`

src/pudl/metadata/resources/eia860.py
aesharpe: It's m:m, you're right!

migrations/versions/6e9328b6b275_.py
aesharpe: This was the result of running `alembic merge heads` with Zach's PR. I tried to consolidate them, but it didn't work.

src/pudl/metadata/fields.py
aesharpe: This is a great point! These values are reported by the utilities and are thus used by EIA.

src/pudl/transform/eia860.py
aesharpe: Yup! there is no overlap between the two.

src/pudl/transform/eia860.py
aesharpe: That's probably a good idea! Will do.

src/pudl/metadata/fields.py
```diff
@@ -397,7 +397,10 @@
     },
     "cooling_id_eia": {
         "type": "string",
-        "description": "The cooling system identification number used by EIA.",
+        "description": (
+            "The cooling system identification number reported to EIA. This value may "
+            "not be a unique identifier"
```
e-belfer, line 402: I actually find this addition confusing, as it is presumably unique within a given plant? I don't think it's a necessary addition personally.

src/pudl/metadata/fields.py
```diff
@@ -2252,26 +2258,29 @@
...
-            "as the primary key for the boiler_stack_flue_assn_eia860 table and "
-            "shouldn't necessarily be used to connect the values in this table to "
-            "those of another."
+            "A stack/flue identification value produced by PUDL and used as the "
+            "primary key for the boiler_stack_flue_assn_eia860 table. If there is an "
+            "ID for a combined stack-flue unit `stack_flue_id_eia`, this value is "
+            "used. Otherwise, this value is a combination of the EIA stack ID "
+            "`stack_id_eia` and the flue ID `flue_id_eia`, separated by an underscore. "
+            "If there is no flue ID, this is equivalent to the stack ID provided by "
+            "EIA. This value may not be a unique identifier."
```
e-belfer, line 2283: Here is where it makes sense to add this caveat, definitely!

e-belfer (approved):
One extremely minor kibbutz you can choose to ignore if you feel strongly otherwise. Otherwise looks great! You might just want to quickly edit the PR description to remove the WIP stuff before you merge.

src/pudl/metadata/fields.py
zaneselvans: I also don't understand what this means.  I would have guessed that cooling system ID was a unique ID within a plant in the same way that generator ID or boiler ID is unique within a plant, so it would be part of a composite primary key that includes the plant ID.  Is that not true?  Or if the emissions control equipment is primarily associated with boilers, then is the unique composite primary key `(plant_id_eia, boiler_id, cooling_id_eia)`?

I would have assumed the same thing across all of the other environmental equipment IDs below which you've included the same disclaimer for.  Is that wrong?

src/pudl/transform/eia860.py
zaneselvans: If there's never any overlap between these two ways of identifying the stacks and flues -- each record has either a stack ID and a flue ID *OR* a stack-and-flue ID, is there any reason why we wouldn't want to assign null values of both the stack ID and flue ID the stack-and-flue ID?  That would avoid creating a new composite `stack_flue_id_pudl` column and also ensure that both `stack_id_eia` and `flue_id_eia` are always non-null.  And it seems like it would preserve the general idea of the stack ID and flue ID without losing any information.

zaneselvans (commented):
Maybe there's a reason this wouldn't make sense that I'm not seeing, but it seems like we could fill in NA values in both `stack_id_eia` and `flue_id_eia` with whatever value is reported `stack_flue_id_eia` without losing information, and have a simpler overall structure.  Is there any good reason why they've broken these IDs into 3 different fields?

src/pudl/transform/eia860.py
aesharpe: This only works insofar as there are always values for `stack_id_eia` and `flue_id_eia`, but it looks like there are 15 rows with no `flue_id_eia`, only a `stack_id_eia`.

zaneselvans: Wow that's annoying. 15 records out of how many?  Do we think that's accurate or bad data? It would be much simpler / more similar to all of our other data if we just had these IDs in two nice clean columns. @e-belfer what do you think about this?

aesharpe:
In 2013 EIA stopped breaking out `stack_id_eia` and `flue_id_eia` in favor of a single `stack_flue_id_eia`. This value generally reflects whatever value was reported for `stack_id_eia` for a given `plant_id_eia` and `boiler_id` combo in prior years.

These early (seperate) stack and flue do have some 1:m or m:1 relationships, so we can't just use the `stack_id_eia` column to fill in `stack_flue_id_eia` for older years and call it a day.

src/pudl/metadata/fields.py
aesharpe: Ok I agree, I will remove.

src/pudl/transform/eia860.py
aesharpe: Also, they don't report `stack_id_eia` and `flue_id_eia` anymore, so we could just take `stack_flue_id_eia` and split it into both `flue_id_eia` and `stack_id_eia`, but A) that means conserving old columns that aren't used anymore, and B) potentially recording false `flue_id_eia` values.

zaneselvans:
But within a given `(plant_id_eia, boiler_id)` the combination of `stack_id_eia` and `flue_id_eia` is always unique, so we **could** use them as part of a natural composite primary key, except that in some cases one of them is `NA`?

In which case we can either concatenate them to create a new element of a composite PK that won't show up in any other tables (unless we also create it there) which sounds like a bit of special case logic that might be nice to avoid.  Alternatively we could add a null-ish string value like "NULL" so SQLite doesn't yell at us about having missing values in a PK field.

If there are many-to-1 or 1-to-many combinations of these stack/flue IDs in older years, it seems like there's some loss of information in the 2013 and later data where only a single ID is ever reported?  Are we going to end up with a bunch of IDs that change spuriously across years?  Ideally we'd like to have continuity across years.

aesharpe:
> we can either concatenate them to create a new element of a composite PK that won't show up in any other tables (unless we also create it there) which sounds like a bit of special case logic that might be nice to avoid. Alternatively we could add a null-ish string value like "NULL" so SQLite doesn't yell at us about having missing values in a PK field.

I took a look at the NA `flue_id_eia` values and I can spot fix them all easily!

aesharpe:
@e-belfer and I decided the best option is to revert back to `stack_flue_id_pudl`

aesharpe:
Annoying addendum here #2629 b/c I forgot to push some commits when I merged.

aesharpe:
And final annoying addendum to fix the build fails #2634

PR #2593, opened 2023-05-23
Clean epacamd_eia mismatches before epacems creation

zschira:
This PR cleans up mismatches between years of the `epacamd_eia` asset. These occur when there are multiple `plant_id_eia` values for the same `plant_id_epa` and `emissions_unit_id_epa`. The `hourly_emissions_epacems` asset checks that there are none of these one-to-many type matches, but adding the 2021 version of the crosswalk introduced a handful of these due to different matches being found for different years. To resolve these cases there's now an intermediate asset called `epacamd_eia_unique` that will always use the row from 2021 when one of these mismatches occurs.

### DIFF
--- a/src/pudl/etl/glue_assets.py
+++ b/src/pudl/etl/glue_assets.py
@@ -267,6 +267,10 @@ def epacamd_eia_subplant_ids(
 ) -> pd.DataFrame:
     """Groups units and generators into unique subplant groups.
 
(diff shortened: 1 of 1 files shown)


aesharpe (approved):
This looks good to me! The one thing I would suggest is including a description of why we are using the `epacamd_eia_unique` table for the subplant ids instead of the `epacamd_eia` as is in the function doc string. @cmgosnell might have the most specific insight on this.

PR #2595, opened 2023-05-23
Integrate FERC Form 6 from dbf

rousik:
This adds support for loading FERC Form 6 data from dbf archives for years 2000-2020. Additionally fixes few correctness issues in the new dbf extractor code. 

This builds on top of the ferc-form2 PR which is in flight, which is why the base branch is set to `rousik-ferc2` for now (to make it easy to understand cumulative changes).

This should be merged into `dev` once `rousik-ferc2` lands.


src/pudl/settings.py
```diff
@@ -508,7 +525,9 @@ class Ferc6XbrlToSqliteSettings(FercGenericXbrlToSqliteSettings):
     """
 
     data_source: ClassVar[DataSource] = DataSource.from_id("ferc6")
-    years: list[int] = data_source.working_partitions["years"]
+    years: list[int] = [
+        year for year in data_source.working_partitions["years"] if year >= 2021
+    ]
```
zaneselvans, line 530: As I mentioned with the FERC 2, I think it probably makes sense to use a validator in this class to ensure that only appropriate years can be requested.

zaneselvans (commented):
I merged some fixed from the `ferc2` branch into this branch to try and get the CI to pass.

Great that so little needs to be done.

I think we should do some settings validation, and am curious why there's some FERC 2 & 6 stuff happening in the FERC 1 module.

rousik:
This is close to being ready. The minor shuffling of the code to move the generic ferc stuff out of `ferc1.py` remains, but the rest should be good to go. I'm testing with full extraction now.

zaneselvans:
I suspect we should add a `table_transform()` method that deduplicates whatever the FERC Form 6 respondent ID table is, as happens now in FERC Forms 1 & 2.

zaneselvans:
If the CI tests pass on the FERC 2 DBF PR #2661 I’ll merge it in, and then I think the same respondent ID deduplication should get integrated into this PR, and assuming it all passes then I think this PR is ready to go too.

zaneselvans (changes requested):
I think the only thing we need to do here is get #2661 merged in, and then bring the respondent ID deduplication code into the FERC 6 process.

rousik:
Code to deduplicate respondent table has been added and sync'ed with the most recent `dev` branch also.

zaneselvans (approved):
Ran it locally and didn't see anything crazy!

I did create a followup issue to check all the FK constraints that we're adding to these DBs in #2674.

I'll add a release note referencing this PR.

We can add the link to the FERC 6 SQLite DB for download after we get a good nightly build.

PR #2598, opened 2023-05-24
Minor Documentation changes

AppTrain:
# PR Checklist

### DIFF
--- a/docs/dev/dev_setup.rst
+++ b/docs/dev/dev_setup.rst
@@ -272,6 +272,13 @@ The path stored in ``PUDL_OUTPUT`` contains all ETL outputs like
 It is recommended you create these directories outside of the pudl repository
 directory so the inputs and outputs are not tracked in git.**
 
+Also, activate profile changes above in the current session.
+
+.. code-block:: console
+
+    $ export PUDL_OUTPUT=/absolute/path/to/pudl_output 
+    $ export PUDL_INPUT=/absolute/path/to/pudl_input
+
 .. _Legacy PUDL Setup:
 
 PUDL Workspace Setup (legacy method)
--- a/docs/dev/run_the_etl.rst
+++ b/docs/dev/run_the_etl.rst
@@ -163,7 +163,7 @@ Running the ETL with Dagit
 Dagster needs a directory to store run logs and some interim assets. We don't
 distribute these outputs, so we want to store them separately from
 ``PUDL_OUTPUT``. Create a new directory outside of the pudl respository
-directory called ``dagster_home/``. Then set the ``DAGSTER_HOME`` environment
+directory called ``dagster_home/``, and create a ``dagster_home/dir`` subdirectoy. Then set the ``DAGSTER_HOME`` environment
 variable to the path of the new directory:
 
 .. code-block:: console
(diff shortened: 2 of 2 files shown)


docs/dev/run_the_etl.rst
```diff
@@ -163,7 +163,7 @@ Running the ETL with Dagit
 Dagster needs a directory to store run logs and some interim assets. We don't
 distribute these outputs, so we want to store them separately from
 ``PUDL_OUTPUT``. Create a new directory outside of the pudl respository
-directory called ``dagster_home/``. Then set the ``DAGSTER_HOME`` environment
+directory called ``dagster_home/``, and create a ``dagster_home/dir`` subdirectoy. Then set the ``DAGSTER_HOME`` environment
```
jdangerx, line 166: What was the issue that made you do this? It's working for me without this subdir, but I want to know what weirdness popped up in your setup!

docs/dev/dev_setup.rst
```diff
@@ -272,6 +272,13 @@ The path stored in ``PUDL_OUTPUT`` contains all ETL outputs like
 It is recommended you create these directories outside of the pudl repository
 directory so the inputs and outputs are not tracked in git.**
 
+Also, activate profile changes above in the current session.
```
jdangerx, line 275: Yeah, good point :)

jdangerx (changes requested):
Welcome, thanks for contributing to our documentation! I hope the rest of the setup process wasn't too confusing 😅 

I want to make sure I understand when the `dagster_home/dir` subdirectory is required before telling all our users to create it at setup, but otherwise this looks good!

docs/dev/run_the_etl.rst
AppTrain: For me , if I remove this subir, when I run
`dagit -m pudl.etl -m pudl.ferc_to_sqlite`
i get
` File "/home/apptrain/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/instance/__init__.py", line 508, in get
    raise DagsterInvariantViolationError(
dagster._core.errors.DagsterInvariantViolationError: $DAGSTER_HOME "/home/apptrain/dagster_home/dir" is not a directory or does not exist. Dagster requires this environment variable to be set to an existing directory in your filesystem`

`(pudl-dev) apptrain@greenmachine:~/pudl$ dagit --version
dagit, version 1.3.5
(pudl-dev) apptrain@greenmachine:~/pudl$ python --version
Python 3.11.3
(pudl-dev) apptrain@greenmachine:~/pudl$`

jdangerx: Ah, I think the root cause here is actually an earlier typo - If you set `DAGSTER_HOME` to the path to the `dagster_home` directory, not `dagster_home/dir`, it should work without adding the `dir` subdirectory.

Here's what I get:

```fish
> echo $DAGSTER_HOME                                              (pudl-dev) 
/Users/dazhong-catalyst/work/dagster_home
> ls $DAGSTER_HOME                                                (pudl-dev) 
dagster.yaml  history/      schedules/    storage/
```

And I suspect you'd get

```fish
> echo $DAGSTER_HOME                                              (pudl-dev) 
.../dagster_home/dir
> ls $DAGSTER_HOME                                                (pudl-dev) 
dagster.yaml  history/      schedules/    storage/
```

If that's true, I think we should just clarify the env var value instead of adding a new directory!

jdangerx: @AppTrain - just checking in - if you confirm that the above works for you, we can update the docs changes accordingly!

AppTrain: @jdangerx , yes, thanks for the reminder. I just confirmed, your assumptions are correct.  My DAGSTER_HOME included that 'dir' subdir.  I fix my ENV variable with "export DAGSTER_HOME=/home/apptrain/dagster_home" and  I'm able to run dagster without that 'dir' directory now. So I agree , we should fix the EXPORT line.

PR #2602, opened 2023-05-25
Fix rouge operational status codes causing foreign key failure

aesharpe:
Little fix for an operational status code value in the `emissions_control_equipment_eia860` table that was causing the nightly builds to fail. There was a value called "CS" for `plant_id_eia`: 2682, `particulate_control_id_eia` values 9 and 10 in 2016 that wasn't specified in the EIA's list of possible operational status codes. I looked at the operational status codes for the same plant and ids in 2015 and 2017 and both were "RE" so I changed it to RE. 

I made this change in the transform module rather than in the code_fixes part of the CODE_METADATA dictionary because it felt more like a one-off fix then something that should be applied across the board. I.e., I couldn't be sure that future instances of "CS" in the data should be automatically converted to RE.

### DIFF
--- a/src/pudl/transform/eia860.py
+++ b/src/pudl/transform/eia860.py
@@ -860,15 +860,6 @@ def clean_emissions_control_equipment_eia860(
         "emission_control_equipment_cost",
     ] = 3200
 
-    # Fix bad operational status code. Decided to fix it here instead of in the
-    # code_fixes portion of the CODE_METADATA because it felt more like a one-off
-    # rather than something that should be applied across the board. I determined that
-    # this was the right fix by looking at the operational_status_code for this
-    # emissions control unit in the surrounding yeras.
-    emce_df.loc[
-        emce_df["operational_status_code"] == "CS", "operational_status_code"
-    ] = "RE"
-
     return emce_df
 
 
--- a/src/pudl/transform/eia860.py
+++ b/src/pudl/transform/eia860.py
@@ -860,6 +860,15 @@ def clean_emissions_control_equipment_eia860(
         "emission_control_equipment_cost",
     ] = 3200
 
+    # Fix bad operational status code. Decided to fix it here instead of in the
+    # code_fixes portion of the CODE_METADATA because it felt more like a one-off
+    # rather than something that should be applied across the board. I determined that
+    # this was the right fix by looking at the operational_status_code for this
+    # emissions control unit in the surrounding yeras.
+    emce_df.loc[
+        emce_df["operational_status_code"] == "CS", "operational_status_code"
+    ] = "RE"
+
     return emce_df
 
 
--- a/src/pudl/metadata/codes.py
+++ b/src/pudl/metadata/codes.py
@@ -943,6 +943,7 @@ CODE_METADATA: dict[str, dict[str, Any]] = {
             "(V) Under construction, more than 50 percent complete": "V",
             "BU": "SB",
         },
+        "ignored_codes": ["CS"],
     },
     "energy_sources_eia": {
         "df": pd.DataFrame(


src/pudl/transform/eia860.py
```diff
@@ -860,6 +860,15 @@ def clean_emissions_control_equipment_eia860(
         "emission_control_equipment_cost",
     ] = 3200
 
+    # Fix bad operational status code. Decided to fix it here instead of in the
+    # code_fixes portion of the CODE_METADATA because it felt more like a one-off
+    # rather than something that should be applied across the board. I determined that
+    # this was the right fix by looking at the operational_status_code for this
+    # emissions control unit in the surrounding yeras.
```
zaneselvans, line 867: What's the argument for not integrating this into the general encoding system we've already got? `CS` is an invalid code, and the only reason this wouldn't work in general is if it we discovered that it showed up somewhere else with a different meaning.  Do you know what `CS` stands for here or why it showed up rather than one of the expected codes?

zaneselvans (commented):
I'm not sure I understand the rationale for doing this fix outside of the existing encoding system.

src/pudl/transform/eia860.py
zschira: I think I agree with @zaneselvans that using the existing encoding system would be preferable unless there's a specific reason that might be a problem. However, if we decide not to do that I think it would be better to have a more targeted spot-check here. As it is now, if `CS` were to show up again in the future we would quietly replace it when that might not be what we want.

zaneselvans: I think @aesharpe's idea here is that the information about what code to replace `CS` with in this case actually isn't contained in the `CS` code itself -- unlike most of the fixable bad codes which are misspellings, or changes in the codes that are used over time, and so you can tell which code it ought to be based just on the observed code.  Whereas here, the information about what code ought to be used instead of `CS` is contextual, and depends on the adjacent records, and that kind of information wouldn't necessarily be applicable to other instances of `CS` that crop up elsewhere.  It's more of a value interpolation / fillna than it is a value substitution.

The more I think about it the more I think not using the encoder is probably the right thing to do.  Ideally we would have another more general process for filling in missing values based on context, but we don't have that yet, so I think @aesharpe's solution is probably good.

zaneselvans (approved):

src/pudl/transform/eia860.py
aesharpe: @zaneselvans yes, that was exactly my thinking.

aesharpe: @zaneselvans @zschira  We could still certainly NULL out the `CS` value and then create an interpolation function to fill it in if that feels better.

e-belfer: The logic for the interpolation completely makes sense to me, but I'd lean towards nulling it to be consistent with the way we've treated other non-standard codes, and creating an issue to write an interpolation function separately. I've certainly seen other one-off codes that aren't in the historical versions of a form and have nulled them, and it'd be helpful to spell out the interpolation logic a bit more clearly / get a sense of how many values might be affected if we start filling in gaps this way, rather than doing it in a one-off way. Just my two cents!

aesharpe: Ok @e-belfer I will null it and link an issue for interpolation.

Do we have a standard way of mapping all values in a column to a set of enum values? I want all instances of `operational_status_code` to adhere to the the codes in `CODE_METADATA["operational_status_eia"]`

zaneselvans: I think @e-belfer is right that the "correct" thing to do is treat all these code / attribute interpolations the same way. If we have a number of examples of this kind of cleaning needing to be done already and we can easily standardize the process, it would be great to include a list of the known cases in the issue.

zaneselvans: If you add `CS` to the list of codes which cannot be fixed, it will be NA after encoding step.

zaneselvans: We don't want to both have an ENUM constraint on the field and a FK relationships, since that would be specifying the same information (about allowable values) in two different ways.

aesharpe: When / where does the encoding step happen? Just curious.

aesharpe: NM found it.

aesharpe:
Interpolation of values to be addressed in this issue: #2616

aesharpe:
Had to continue this in #2617 because it didn't actually add the encoder to the transform process which is needed to convert the `ignored_codes` to NA.

Issue #2584, opened 2023-05-18
when there is a second dimension (namely `utility_type`) add a second layer of calcuations to check that `utility_type=="total"` is calculable


e-belfer:
A sub-task of #2016. Within each FERC1 table to be exploded, `utility_type` contains both specific types of utilities and a "total" category. We should add a set of calculations within each table that checks to see whether the components sum to the total.

- [x] Within `check_table_calcs` add a sum to total calculation check for the relevant calculated field
- [x] If the calculations don't sum to total, log the total # of calcs which don't sum to total as an error

PR #2615, opened 2023-05-29
Compare sub-total calculations to total calculations for XBRL explosion

e-belfer:
Addresses issue #2584. When there is a second dimension within a table (e.g. `utility_type`), this PR adds a second set of calculations to the `check_table_calcs` function of the FERC1 transform to log any calculated values that aren't equal to the sum of their components. 

Tables and their sub-components:
- `plant_in_service_ferc1`: None
- `utility_plant_summary_ferc1`: `utility_type`
- `balance_sheet_assets_ferc1`: None
- `balance_sheet_liabilities_ferc1`: None
- `depreciation_amortization_summary_ferc1`: `plant_function`
- `electric_operating_revenues_ferc1`: None
- `retained_earnings_ferc1`: None
- `income_statement_ferc1`: `utility_type`
- `electric_operating_expenses_ferc1`: None

Remaining tasks:
- [x] Finish adding sub-components for remaining tables
- [x] Test when #2618 integrates `check_table_calcs` into the transform

# PR Checklist


src/pudl/transform/ferc1.py
```diff
@@ -781,6 +785,38 @@ def check_table_calcuations(
             f"Calculations in {table_name} are off by {off_ratio}. Expected tolerance "
             f"of {params.calculation_tolerance}."
         )
+
+    # Check that sub-total calculations sum to total.
+    if params.subtotal_column is not None:
+        sub_group_col = params.subtotal_column
+        pks_wo_subgroup = [col for col in pks if col != sub_group_col]
+        calculated_df["sub_total_sum"] = (
+            calculated_df.pipe(lambda df: df[df[sub_group_col] != "total"])
```
cmgosnell, line 794: wow i didn't know you could do this! neat

src/pudl/transform/ferc1.py
```diff
@@ -781,6 +785,38 @@ def check_table_calcuations(
...
+            .transform("sum")
+        )
+        calculated_df["sub_total_sum"] = calculated_df["sub_total_sum"].fillna(
+            calculated_df.calculated_amount
+        )
+        sub_total_errors = (
+            calculated_df.groupby(pks_wo_subgroup)
+            .filter(lambda x: x["sub_total_sum"].nunique() > 1)
+            .groupby(pks_wo_subgroup)
+        )
```
cmgosnell, line 805: i don't doubt that this is doing the thing but this feels opaque to me. can you explain it? i can guess that if a group has more than one value than its a baddie. But why not just find the instances where `calculated_df.sub_total_sum.notnull() & (calculated_df.sub_total_sum != calculated_df.calculated_amount)`? i think this is just something i am not understanding.

src/pudl/transform/ferc1.py
```diff
@@ -781,6 +785,38 @@ def check_table_calcuations(
...
+        off_ratio_sub = (
+            sub_total_errors.ngroups / calculated_df.groupby(pks_wo_subgroup).ngroups
+        )
+        if sub_total_errors.ngroups > 0:
+            logger.warning(
+                f"{table_name}: has {sub_total_errors.ngroups} ({off_ratio_sub:.02%}) sub-total calculations that don't "
+                "sum to the equivalent total column."
+            )
+        if off_ratio_sub > params.calculation_tolerance:
```
cmgosnell, line 814: i could imaging wanting different tolerances for the metadata-based calcs vs the other-dimension calcs but if it works across both that'd be ideal!

src/pudl/transform/ferc1.py
```diff
@@ -781,6 +785,38 @@ def check_table_calcuations(
...
             f"of {params.calculation_tolerance}."
         )
+
+    # Check that sub-total calculations sum to total.
+    if params.subtotal_column is not None:
+        sub_group_col = params.subtotal_column
+        pks_wo_subgroup = [col for col in pks if col != sub_group_col]
+        calculated_df["sub_total_sum"] = (
+            calculated_df.pipe(lambda df: df[df[sub_group_col] != "total"])
+            .groupby(pks_wo_subgroup)["calculated_amount"]
```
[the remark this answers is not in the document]
e-belfer: There are two possibilities here: 1) check that the reported calculated value is the sum of the reported sub-components and 2) check that the calculated total is the sum of the reported (and maybe also calculated) sub-components. Do we actually care about both here? Or is the column_to_check (aka reported value) more interesting to us here?

src/pudl/transform/ferc1.py
e-belfer: Basically I fill every value in this column with the sum of the sub-components (i.e. the total as calculated from the sub-components), except for the "total" row which is filled in directly with the calculated total value. If there is more than one kind of value in the group, it means that the total from the sub-components and the calculated total aren't equivalent. I group and filter like this because I'm trying to compare values both across rows and across groups. It's a bit confusing, I can document it a bit better! But directly comparing the `sub_total_sum` to `calculated_amount` within the same row won't provide the needed information.

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
e-belfer: Updated to use reported values rather than calculated.

cmgosnell (approved):
these look like good updates!! (as long as all the tests pass of course :-)

Issue #2604, opened 2023-05-25
xbrl metadata cleaning and management for 💥 tables


cmgosnell:
```[tasklist]
- [x] create generic process to rename the calculations found in the xbrl metadata using known transform params that could alter the names of record types.
- [x] enable adding edits to the calculations
- [ ] move calculation fixes into the table transformers
- [x] check if names involved in calcs actually show up in table
- [ ] add `source_table` to upstream xbrl json-based metadata when a calculation component isn't in the existing table
- [ ] add `name_original` to `xbrl_factoid`'s in upstream xbrl json-based metadata to enable downstream/post-table-based-renamed id-ing of same factoid showing up in multiple tables.
- [ ] clean the calculation metadata (renames & fixes) in the table transformers
- [ ] use the table-based metadata in the 💥 calculation checks
```

PR #2618, opened 2023-05-31
migrate calcuation checks into the ferc1 table transformers

### LINKED ISSUE #2605: Use tabular calculation fixes in FERC 1 table transformers
```[tasklist]
- [x] convert giant dictionary into CSV
- [x] make calc cleaning to use calc-comp df & fixes CSV
- [x] deal with flow in `process_xbrl_metadata`
- [x] integrate any straggler calc additions in table-specific `process_xbrl_metadata`
```

notes about the existing calc fixes
```[python]
        # typing for calculated_fields_to_fix. for clarity/in anticipation of a pydantic
        # calc defintion
        xbrl_factoid_name: str
        calc_component: dict[Literal["name", "weight"], str | int]
        calc_component_fix: dict[
             Literal["calc_component_to_replace", "calc_component_new"],
             None | calc_component,
        ]
```

## Design Questions
- to CSV or to parameterize? Or CSV->params? The parameterization of the calc fixes felt really necessary when the calculations are stored in json, but if we store them in a CSV it'll be cleaner and easier to integration into the calc component table.
- is `row_type_xbrl` really necessary? there is a _very solvable_ problem of this column needing to be edited after the calcs are fixed/cleaned. if we do need this we are going to need a flow like: 
    pre-calc-cleaned meta table -> calc comp table -> cleaned calc comp table -> cleaned metadata table
instead of:
    pre-calc-cleaned meta table -> calc comp table -> cleaned calc comp table

cmgosnell:
# PR Overview
Closes #2605, #2604

This PR migrates the `check_table_calculations` into the table transformers and does some cleaning/standardization that came up in that process. It became clear that despite the dollar column name standardization, it wasn't crystal clear how we could extract the `dollar_value` or `ending_balance` column from the parameters like we are now pulling out the `xbrl_factoid` from the params. So I made a new param + transform function + transform wrapper for this stage.

I added a `calc_contains_components_from_other_tables` column into the table of metadata so its easy to identify downstream if a calc is inter- or intra-table calc. I'd love some work-shopping on that name. it long.

The one weird thing (imo) to note is that this transform step really cares about/needs to know about more than just the `df` and `params`. I used the table transformer method to access those bits that the table transformer is aware of and pass them in as args to the transform function.

I also made some `@property`'s of the `Ferc1TableTransformParams` to have a simpler access point to some of the info we're now pulling out of the params to access during the metadata processing and/or the calc checking.

I also also removed some of the cruft in `process_xbrl_metadata` that is no longer needed bc of the standard `rename_column_xbrl_to_pudl`

# PR Checklist


src/pudl/transform/ferc1.py
```diff
@@ -1084,13 +1208,33 @@ def process_xbrl_metadata(self, xbrl_metadata_json) -> pd.DataFrame:
...
+                    k: self.rename_column_xbrl_to_pudl(v) if k == "name" else v
+                    for (k, v) in calc_component.items()
+                }
+                for calc_component in literal_eval(calc)
+            ]
+            return str(renamed_calc)
+
+        tbl_meta.calculations = tbl_meta.calculations.apply(
+            rename_calculation_components
+        )
```
cmgosnell, line 1237: babies first apply function 👶🏻  (they seemed so mysterious)

src/pudl/transform/ferc1.py
```diff
@@ -694,6 +694,96 @@ def combine_axis_columns_xbrl(
...
+        intra_tbl_calcs.xbrl_factoid, intra_tbl_calcs.calculations
+    ):
+        calc_df = (
+            pd.merge(
+                df,
+                pd.DataFrame(literal_eval(calculation)),
+                left_on=xbrl_factoid_name,
+                right_on="name",
+            )
+            .assign(calculated_amount=lambda x: x[params.column_to_check] * x.weight)
```
e-belfer, line 753: Just to confirm, is this the +/- sign from the calculation?

src/pudl/transform/ferc1.py
```diff
@@ -694,6 +694,96 @@ def combine_axis_columns_xbrl(
...
+            .groupby(pks_wo_factoid, as_index=False)["calculated_amount"]
+            .sum(min_count=1)
+            .assign(**{xbrl_factoid_name: calcuated_factoid})
+        )
+        calculated_dfs.append(calc_df)
+
+    calculated_df = pd.merge(
+        df, pd.concat(calculated_dfs), on=pks, how="left", validate="m:1"
```
e-belfer, line 761: Why is this an m:1 validation and not a 1:1?

src/pudl/transform/ferc1.py
```diff
@@ -1900,6 +2025,27 @@ def assign_utility_id_ferc1(
...
+            logger.info(
+                f"{self.table_id.value}: Checking the XBRL metadata-based calcuations."
+            )
+            _ = check_table_calcuations(
+                df=df,
+                tbl_meta=self.xbrl_metadata,
+                xbrl_factoid_name=self.params.xbrl_factoid_name,
+                table_name=self.table_id.value,
+                params=params,
+            )
```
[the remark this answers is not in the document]
e-belfer: We'll probably want this column when we are doing the corrections / balancing column, so if we don't need it now we'll probably want it in the next several days?

src/pudl/transform/ferc1.py
```diff
@@ -3848,15 +3994,6 @@ def process_xbrl_metadata(self, xbrl_metadata_json) -> pd.DataFrame:
         meta = (
             super()
             .process_xbrl_metadata(xbrl_metadata_json)
-            .assign(
```
e-belfer, line 3851: To confirm, these are all now reflected in the `rename_columns_ferc1` parameters? If so, they aren't being added in this PR so I'd just like to confirm!

src/pudl/transform/ferc1.py
```diff
@@ -1084,13 +1208,33 @@ def process_xbrl_metadata(self, xbrl_metadata_json) -> pd.DataFrame:
                     "row_type_xbrl": pd.StringDtype(),
                 }
             )
-            .assign(xbrl_factoid_name_original=lambda x: x.xbrl_factoid)
+            .assign(
+                xbrl_factoid_name_original=lambda x: x.xbrl_factoid,
+                calc_contains_components_from_other_tables=lambda x: x.calculations.str.contains(
```
e-belfer, line 1213: I would just call this "inter_table_calc_flag"

src/pudl/transform/ferc1.py
```diff
@@ -694,6 +694,96 @@ def combine_axis_columns_xbrl(
...
+
+    off_df = calculated_df[
+        ~np.isclose(
+            calculated_df.calculated_amount, calculated_df[params.column_to_check]
+        )
+        & (calculated_df["abs_diff"].notnull())
+    ]
+    calced_values = calculated_df[(calculated_df.abs_diff.notnull())]
+    off_ratio = len(off_df) / len(calced_values)
+    logger.info(
```
e-belfer, line 775: I would suggest changing this to a warning, otherwise it's almost impossible to find in the logging statements.

e-belfer (changes requested):
Overall looks good, mostly have spelling changes and a few questions to suggest. A few more major issues:

When I run this, I get a ZeroDivisionError for `utility_plant_summary_ferc1` from the relative difference calculation:
`rel_diff=lambda x: abs(x.abs_diff / x[params.column_to_check])`
When the reported calculated value is 0, we should add an exception here to avoid dividing by zero!

I also get `ValueError: No objects to concatenate` on the `retained_earnings_ferc1` table - if we haven't integrated this yet enough for it to work, perhaps we shouldn't run it through the explosion just yet.

The values calculated for the `balance_sheet_assets_ferc1` is >20%, which is far higher than I remember it being in previous iterations. Have we diagnosed why that is?

src/pudl/transform/ferc1.py
cmgosnell: sorry i should have done this separately.. these little deletions in the bespoke-table `process_xbrl_metadata` should have happened in the base branch once i integrated the `rename_column_xbrl_to_pudl` into the standard `process_xbrl_metadata`

src/pudl/transform/ferc1.py
```diff
@@ -718,6 +808,39 @@ class Ferc1TableTransformParams(TableTransformParams):
         UnstackBalancesToReportYearInstantXbrl()
     )
     combine_axis_columns_xbrl: CombineAxisColumnsXbrl = CombineAxisColumnsXbrl()
+    check_table_calculations: CheckTableCalculations = CheckTableCalculations()
+
+    @property
+    def xbrl_factoid_name(self) -> str:
+        """Access the column name of the ``xbrl_factiod``."""
```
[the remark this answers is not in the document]
cmgosnell: ty! my fingers can absolutely not remember how to spell this.

src/pudl/transform/ferc1.py
```diff
@@ -694,6 +694,96 @@ def combine_axis_columns_xbrl(
     return df
 
 
+class CheckTableCalculations(TransformParams):
+    """Parameters for checking xbrl-metadata based calculations within a table."""
+
+    column_to_check: str | None = None
+    """Name of data column to check.
+
+    This will typically be ``dollar_amount`` or ``ending_balance`` in hte
```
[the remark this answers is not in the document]
cmgosnell: yes this is definitely me just trailing off. i'll make this a full sentence!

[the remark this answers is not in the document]
cmgosnell: yes this is definitely me just trailing off. i'll make this a full sentence!

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
cmgosnell: yes this is true. I'll try running the tests with returning while returning the df. if it doesn't break anything we might as well keep it in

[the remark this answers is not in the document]
e-belfer: Although actually we could use 'abs_diff` to give us the same info, obviously...

src/pudl/transform/ferc1.py
```diff
@@ -694,6 +694,96 @@ def combine_axis_columns_xbrl(
...
+    inter_table_calculated_values = list(
+        tbl_meta[
+            tbl_meta.calc_contains_components_from_other_tables
+        ].xbrl_factoid.unique()
+    )
+    if inter_table_calculated_values:
+        logger.info(
+            "Skipping calcuated values because they are inter-table calucations: "
+            f"{inter_table_calculated_values}"
+        )
```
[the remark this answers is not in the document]
cmgosnell: this makes sense. i'll convert it to a warning & make a note that maybe it should be moved when we deal with these.

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
cmgosnell: i ended up returning the df and passing it along. it didn't junk up my tests so i think we are doing okay with it.

src/pudl/transform/ferc1.py
```diff
@@ -694,6 +694,106 @@ def combine_axis_columns_xbrl(
...
+class CheckTableCalculations(TransformParams):
+    """Parameters for checking xbrl-metadata based calculations within a table."""
+
+    column_to_check: str | None = None
+    """Name of data column to check.
+
+    This will typically be ``dollar_amount`` or ``ending_balance`` in hte
+    """
+    calculation_tolerance: float = 0.05
+    """The tolerance ratio of the off calcuations and the possible calcuations."""
```
zaneselvans, line 706: It's common for these kinds of tolerance checks to include both "absolute" (difference) and "relative" (ratio) tolerances (e.g. see the `atol` and `rtol` parameters to `np.isclose()`.  Even if we aren't using both types of tolerance here I think it would be helpful if the parameter name indicated which type we're talking about (relative in this case)

cmgosnell: i hear you but this is not the tolerance of the calced-value vs the reported-value, but rather the # of calculations that are `~np.isclose` w/ default `atol` and `rtol` vs the total number of calculated values.

although another name for this is encouraged!

src/pudl/transform/ferc1.py
```diff
@@ -694,6 +694,106 @@ def combine_axis_columns_xbrl(
...
+    pks = pudl.metadata.classes.Resource.from_id(table_name).schema.primary_key
+    pks_wo_factoid = [col for col in pks if col != xbrl_factoid_name]
+    calculated_dfs = []
+    for calcuated_factoid, calculation in zip(
+        intra_tbl_calcs.xbrl_factoid, intra_tbl_calcs.calculations
+    ):
+        calc_df = (
+            pd.merge(
+                df,
+                pd.DataFrame(literal_eval(calculation)),
```
zaneselvans, line 749: Why are we using `ast.literal_eval()` (which can evaluate any string as if it were python code) rather than the built-in JSON serialization / deserialization functions `json.load()` / `json.dump()`?  It seems like all of the interaction with the calculations and other metadata is about working with JSON rather than lower level Python code parsing.

cmgosnell: bc of the " / ' problem

cmgosnell: which you've mentioned before is an upstream problem that we should probably deal with. it'd be great to use the `json` tools directly! and this should be very straightforwad to swap out when there isn't mixed quotes.

e-belfer (approved):
Still conceptually confused about the m:1 validation, but otherwise this branch now runs as expected for me, with all calc checks passing. Duplicated components are now removed, and the div by 0 issue has been fixed. Presuming it passes CI should be ready to merge into the main reshape branch.

src/pudl/transform/ferc1.py
zaneselvans: Okay that's what I thought. We should never be writing invalid JSON into these fields in the first place.  Do we know where that's happening? The JSON we're getting from the original extractor seems to be fine (it parses as good JSON in `json_normalize()`) so I would guess this is happening in `process_xbrl_metadata()` somewhere, Probably because we assign some dataframe cells to contain a python object rather than a JSON string representation of that object which would be generated by e.g. `json.dumps(my_object)` (the `s` is for string).

zaneselvans: A [JSON & Python tutorial](https://realpython.com/python-json/#python-supports-json-natively) from RealPython.

cmgosnell: converted!

src/pudl/transform/ferc1.py
cmgosnell: ah this! I was also confused. it can't be a 1:1 before two bespoke drop dupes that happen for "income_statement_ferc1" & "electric_operating_revenues_ferc1" post the standard `transform_main`. So I moved the check calc transform to the very beginning of `transform_end` & converted the validation to be 1:1

PR #2620, opened 2023-06-02
Add records correcting FERC 1 calculations that are off

zaneselvans:
# PR Overview

* Fix a bad reference to old XBRL Metadata asset names in our FERC 1 ETL debugging notebook.
* Add a generic `{xbrl_factoid}_correction` component to all calculations in `process_xbrl_metadata()` method of the FERC 1 table transformers.
* Add a step to the `check_table_calculations` method that calculates the difference between the reported totals and our calculated totals and adds corrective records so that these totals match.

## Questions:
* It seemed wrong to have any value of `record_id` in the correction records since they have no source record, so I set it to NA.  Are there any downstream consequences for this?
* It seemed wrong to say that the correction records had `xbrl_row_type` value of `reported_value` or `calculated_value` so I set them to `correction` and updated the ENUM constraint on that field.
* What set of calculations we want to create correction records for? Right now they're only being generated for the ones  that fail the `np.isclose()` test (and end up in `off_df`), but we say "match exactly" in some places, and there are lots of records which are off by just one dollar -- none of which will get corrections in the current setup, which seems fine, even though there will technically still be some difference between the calculated and reported values.
* Do we want to change the name of `check_table_calculations` to `correct_table_calculations` or something else that indicates we're actually altering the database tables?

# PR Checklist


src/pudl/transform/ferc1.py
```diff
@@ -790,11 +789,27 @@ def check_table_calculations(
             f"{table_name}: has {len(off_df)} ({off_ratio:.02%}) records that don't "
             "calculate exactly."
         )
+
+    if off_ratio > 0 and off_ratio <= params.calculation_tolerance:
```
cmgosnell, line 793: I think we don't want the second half of this. a table can be off by less than the `calculation_tolerance` and i think we still want to add corrections to the off guys.

cmgosnell: or are you saying we don't add corrections if the table isn't so bad as to trip the assertion right under this section? okay maybe that is fine but it also seems a lil duplicative. if you take the second chunk away you could slip the corrections all under the `if off_ratio > 0:` above

src/pudl/transform/ferc1.py
```diff
@@ -790,11 +789,27 @@ def check_table_calculations(
...
+        logger.info(
+            "Creating correction records to make calculations match reported values."
+        )
+        corrections = off_df.copy()
+        corrections[params.column_to_check] = (
+            corrections[params.column_to_check] - corrections["calculated_amount"]
+        )
```
cmgosnell, line 800:
```suggestion
        corrections[params.column_to_check] = (
            corrections[params.column_to_check].fillna(0) - corrections["calculated_amount"]
        )
```

we should probs fillna so we get the full cacled amount if the value is null

cmgosnell (changes requested):
one suggestion for fillna-ing & a nit/musing about a slight simplification. but overall this looks great

zaneselvans:
> I am of two minds about this. I like the way you did it. But also having a correction be applied across the board no matter what seems reasonable and maybe more informative to users (i.e. getting the $1 or mostly $0 corrections).

If we correct these apparent rounding errors, there will be a *lot* of corrections, and they'll be correcting values that are off by only 1 part in a million. But they won't always show up, because there will be many cases in which nothing was off by enough to trip up `np.isclose()` -- and in those cases no corrections of any kind would be generated.

Unless we store these dollar values as decimal numbers (rather than floating point -- decimal numbers are just an integer number of cents with a baked-in assumption that you have 2 sigfigs after the decimal point) the only reasonable way to compare them is with something like `np.isclose()` and I think correcting only those values that `np.arentclose()` will be clearer -- they'll be real corrections of real reporting issues, rather than a ton of fixes most of which aren't important.

> I'm personally not opposed to the idea of using the `record_id` of the calculated value because this is truly the source.

I can see how this isn't crazy, but it could also end up creating additional tables with non-unique `record_id` values for reasons that have nothing to do with reshaping the data.  It doesn't appear to right now though.  I really want to change this to `source_record_id` across the board so folks don't ever try and use it as a unique ID for the tables you find it in but that's another story.  I just don't want anyone to think that these correction values come from any original data source, which is kinda what having a `record_id` indicates to me.

Should I go ahead and rename `check_table_calculations` to something that indicates both checking and correction is happening?

And do we truly always want corrections to be added?

cmgosnell:
using np.arentclose() as the threshold for whether to make a correction sounds grand to me

> Should I go ahead and rename check_table_calculations to something that indicates both checking and correction is happening?

Sure!

> And do we truly always want corrections to be added?

If we end up not wanting to we can always add another param into `CheckTableCalculations`

src/pudl/transform/ferc1.py
```diff
@@ -793,17 +808,33 @@ def check_table_calculations(
...
+            corrections[params.column_to_check].fillna(0.0)
+            - corrections["calculated_amount"]
+        )
+        corrections[xbrl_factoid_name] = corrections[xbrl_factoid_name] + "_correction"
+        corrections["row_type_xbrl"] = "correction"
+        corrections["record_id"] = pd.NA
+
+        calculated_df = pd.concat(
+            [calculated_df, corrections], axis="index"
+        ).reset_index()
```
cmgosnell, line 836: are you sure the index will match perfectly here? this seems more risky than merging on the `pks_wo_factoid` with some merge validation

cmgosnell: hm nvm!

cmgosnell (approved):

Issue #2470, opened 2023-03-30
Use test-only $DAGSTER_HOME and default IOManager in tests


zaneselvans:
Attempting to test some of our ETL debugging notebooks wit pytest / Tox in CI in #2450 I found that:
* The `execute_in_process()` executor used by the tests stores all assets using the Default IO Manager in memory, rather than on disk in `$DAGSTER_HOME`.
* The test environment doesn't have its own isolated `$DAGSTER_HOME` environment variable, and inherits that value from the user environment.

This combination meant that when I ran the tests locally, the notebooks were able to access outputs that came from my personal environment rather than the test environment, and passed, but when the same tests ran in CI, they failed (since there were no outputs in the test runner filesystem).

* To ensure better test isolation and avoid this kind of confusion / false negative (or false positives!) we should ensure that the `DAGSTER_HOME` envvar is set to a temporary directory created for/by pytest.
* When running with `--live-dbs` should `DAGSTER_HOME` still be a temporary directory? Or should it point at the user's existing dagster asset outputs?
* Rather than storing all of the assets that use the default IO Manager in memory, I think we might want to make sure they're written out to disk as they will be in production. This will make the tests more similar to production, limit the memory footprint of the tests (which need to be able to run with only 7GB of memory in CI), and allow out-of-process tests to access the outputs (like the Jupyter notebooks, or any CLI tools that are being tested)

PR #2621, opened 2023-06-02
Bring the Census DP1 to SQLite ETL into dagster

e-belfer:
Addresses issue #2412.

This PR brings `pudl.convert.censusdp1tract_to_sqlite` and `pudl.output.censusdp1` into dagster, producing 3 pickled outputs (1 for each layer of the Census DP1 database that is of interest). As of now, these outputs are not written into the PUDL database. They currently have 100+ columns, and are mostly used for their geometries. We pickle these outputs using the current default IO manager and feed them into downstream outputs that are written into PUDL (e.g. the state demand outputs in #2550). This is what is currently in this PR.

Remaining tasks:
- [x] Convert `pudl.output.censusdpt1tract.py` into dagster assets

# PR Checklist


src/pudl/etl/__init__.py
```diff
@@ -39,6 +39,9 @@
     *load_assets_from_modules([pudl.transform.ferc1], group_name="norm_ferc1"),
     *load_assets_from_modules([pudl.extract.ferc714], group_name="raw_ferc714"),
     *load_assets_from_modules([pudl.transform.ferc714], group_name="clean_ferc714"),
+    *load_assets_from_modules(
```
e-belfer, line 42: This shouldn't actually be here - instead, we should add a new extraction module for Census data that is loaded here!

zaneselvans:
I think we want to treat the Census DP1 like we do the `ferc1.sqlite` database -- where the whole `censusdp1tract.sqlite` DB is created as a "source asset" rather than as individual tables that are written into PUDL or dataframes that are pickled. This would preserve the current set of distributed outputs, which includes several SQLite DBs of which the Census DP1 is one.

jdangerx:
> I think we want to treat the Census DP1 like we do the `ferc1.sqlite` database -- where the whole `censusdp1tract.sqlite` DB is created as a "source asset" rather than as individual tables that are written into PUDL or dataframes that are pickled. This would preserve the current set of distributed outputs, which includes several SQLite DBs of which the Census DP1 is one.

Still persisting the data in a SQLite file that we distribute makes sense to me!

I would add that we can do this without making a whole new DAG - we could add a new subclass of a `SQLiteIOManager` (a la `PudlSQLiteIOManager` or `FercSQLiteIOManager` that knows to read/write to `censusdp1tract.sqlite`, then use that IOManager to write out the three tables instead of just pickling them.

Now that we have a bunch of different instances of "thing that reads/writes from SQLite" though, it might make sense to refactor the code to better fit the use-cases/patterns we've discovered 🤷 .

src/pudl/convert/censusdp1tract_to_sqlite.py
```diff
@@ -152,7 +175,9 @@ def main():
 
     pudl_settings["sandbox"] = args.sandbox
 
-    censusdp1tract_to_sqlite(pudl_settings=pudl_settings, ds=ds, clobber=args.clobber)
+    _ = censusdp1tract_to_sqlite(
```
jdangerx, line 178: Did you want to use this assignment somewhere else?

src/pudl/output/censusdp1tract.py
```diff
@@ -1,82 +1,84 @@
...
-from pudl.workspace.datastore import Datastore
 
 logger = pudl.logging_helpers.get_logger(__name__)
 
 
-def get_layer(
-    layer: Literal["state", "county", "tract"], pudl_settings=None, ds=None
-) -> gpd.GeoDataFrame:
-    """Select one layer from the Census DP1 database.
+@multi_asset(
```
jdangerx, line 15: My understanding is that multi-assets are hard to parallelize - that's probably not a huge problem here, since there's only 3 assets anyways, but you might want to try making a bunch of separate assets.

Looks like you already have a bit of an asset factory idea going on, based on the name of the function, but this function is returning a list of Outputs as opposed to an Asset. Did you want to try something like the `finished_eia_asset_factory` in `transform/eia.py`?


```python
def finished_eia_asset_factory(
    table_name: str, io_manager_key: str | None = None
) -> AssetsDefinition:
    """An asset factory for finished EIA tables."""
    clean_table_name = "clean_" + table_name

    @asset(
        ins={clean_table_name: AssetIn()},
        name=table_name,
        io_manager_key=io_manager_key,
    )
    def finished_eia_asset(**kwargs) -> pd.DataFrame:
        """Enforce PUDL DB schema on a cleaned EIA dataframe."""
        df = convert_cols_dtypes(kwargs[clean_table_name], data_source="eia")
        res = Package.from_resource_ids().get_resource(table_name)
        return res.enforce_schema(df)

    return finished_eia_asset


finished_eia_assets = [
    finished_eia_asset_factory(table_name, io_manager_key="pudl_sqlite_io_manager")
    for table_name in [
        "boiler_fuel_eia923",
        "coalmine_eia923",
        "fuel_receipts_costs_eia923",
        "generation_eia923",
        "generation_fuel_eia923",
        "generation_fuel_nuclear_eia923",
        "ownership_eia860",
        "emissions_control_equipment_eia860",
        "boiler_emissions_control_equipment_assn_eia860",
        "boiler_cooling_assn_eia860",
        "boiler_stack_flue_assn_eia860",
    ]
]
```

If you follow that pattern it would look like

```python
def census_asset_factory(
    layer: str
) -> AssetsDefinition:
    """An asset factory for finished EIA tables."""
    @asset(
        ins={"censusdp1tract_to_sqlite": AssetIn()},
        name=f"{layer}_censusdp1",
    )
    def census_layer(**kwargs) -> pd.DataFrame:
        """yada yada"""
        dp1_engine = ...
        def get_layer(...):
            ...
        # could be nice to pass in the sql engine as opposed to closing over it while defining the get_layer function
        layer_gdf = get_layer(layer, dp1_engine) # or, just in-line the body of get_layer since you're not calling it in a loop anymore?
        return layer_gdf
    return census_layer


census_dp1_layers = [
    census_asset_factory(layer)
    for layer in ["state", "county", "tract"]
]
```

jdangerx (changes requested):
I'm glad this diff turned out so small, nice work! 

A couple questions that I'd like you to respond to, though your answer to both could just be "no" 😅

zaneselvans:
Maybe I am confused (I am often confused) but @jdangerx your first comment still seems to be true -- that the DP1 SQLite DB is getting written to the output directory as a side effect of `censusdp1tract_to_sqlite()` being run, rather than creating explicit tabular assets that Dagster is aware of, either as normal assets or as "source assets" like in the case of the FERC DBs. But if this is true.

I was confused in a different way before, and thought that the pickled geodataframes were the only artifacts that contained the DP1 data. But given that we are producing the DP1 DB and can distribute it alongside everything else like we have been, then I'm confused about your newer comment about needing to create a census SQLite IO Manager. Would that be to generate additional tables that aren't the ones in the converted SQLite DB?

If we did have a separate Census SQLite IO Manager, and we run all the Census DP1 data through geopandas rather than just having the `ogr2ogr` to the whole DB conversion, do we risk getting different data type behaviors than we've previously had? (E.g. losing the important leading zeroes on the FIPS codes if they are interpreted as integers rather than strings). We could avoid this if we defined full metadata (fields, resources, etc.) for the entire DP1 database, but that seems like a extra work. But I agree it would be nice to have these tables be real assets somehow if it's easy.

jdangerx:
It seems like you might be reading the chronology of my comment wrong! The top part is the new content and the stuff below the "edit" line is the old content. I'll re-edit to make that clear!

I don't think we need to create a new IOManager until we want to distribute the three layers separately. The asset that is created by `censusdp1tract_to_sqlite` is *the path to the SQLite db* and then later assets consume that to generate the pickled geodataframes.

I think you're right about risking getting weird data type behaviors, which sounds like a pain but a manageable pain.

src/pudl/convert/censusdp1tract_to_sqlite.py
jdangerx, line 1: I guess we could get rid of the CLI interface here completely? Or refactor it to just wrap around however `dagster asset materialize` works? 🤷

src/pudl/convert/censusdp1tract_to_sqlite.py
e-belfer: This just returns the path to the saved DB, which we don't need to call anywhere else. The loggers for this function already print the output path, which would be the only possible application I could think of for this variable.

jdangerx: Hmm, at this point could you just avoid the assignment altogether?

src/pudl/convert/censusdp1tract_to_sqlite.py
jdangerx: 🔥 🔥 🔥

jdangerx (approved):

src/pudl/convert/censusdp1tract_to_sqlite.py
e-belfer: Indeed, I deleted it (and the whole CLI interface)! :)

jdangerx (approved):
🎉 love to see a diff that makes the codebase smaller

jdangerx:
CI looks like it's failing because we're still testing for the `census_dp1tract` CLI command in console_scripts_test and because `dagster_home` appears to not be a directory on the runner (??? how did anything work before). You should be able to make those directories (and set the $DAGSTER_HOME env var) in `.github/workflows/tox-pytest.yml` if you follow the pattern set by `$PUDL_OUTPUT` and `$PUDL_INPUT`. I'm also happy to help out with the GH actions stuff if you want - let me know!

zaneselvans:
If you're removing the CLI for the census DB conversion (which seems good and fine) you'll also need to remove it from the list of scripts enumerated in `pyproject.toml` -- that's what determines which CLI scripts are tested, and what indicates which modules & functions provide a CLI.

jdangerx:
CI is still failing, with these messages:

```
     FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/pudl-work/dagster_home/storage/county_censusdp1'
     FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/pudl-work/dagster_home/storage/county_censusdp1'
```

I think we need to tell the test to materialize this asset before we actually access it. Maybe you can run some code copied from `devtools/materialize_asset.py` to programmatically materialize those assets?

zaneselvans:
@e-belfer Could the file not found failure here be related to the lack of test environment isolation outlined in #2470?  Locally the tests may be finding your personal copy of the output if they are using your `$DAGSTER_HOME` rather than the one that's associated with the tests.

e-belfer:
> @e-belfer Could the file not found failure here be related to the lack of test environment isolation outlined in #2470? Locally the tests may be finding your personal copy of the output if they are using your `$DAGSTER_HOME` rather than the one that's associated with the tests.
I also thought this would be the case, so I'd manually deleted the relevant file out of `$DAGSTER_HOME` locally. Tests were still passing.

e-belfer:
@zaneselvans In the long-term we'll want to figure out how to access interim dagster assets for tests, I'm sure. But I don't think this is one of those cases, since this issue gets effectively resolved in #2550 and the issue here is only to do with managing an interim workaround. I propose we merge that PR into this one and then merge both into dev if they pass the tests.

zaneselvans:
Go for it, if that fixes the issue, great!

e-belfer:
Great! Will do after #2550 gets a proper review.

PR #2627, opened 2023-06-04
Prototype of output differ

rousik:
This is an early prototype of the output differ tool. This can be used to compare the two PUDL_OUTPUT directories in order to check for differences. At the moment, this tool is quite limited, and can do the following:
1. compare two local directories (support for gcs is not yet there)
2. compare for file presence and sqlite databases
3. for sqlite, this compares presence of tables, table schemas (columns and their types) and row counts and individual rows

Testing for this tool has been rather limited so far, but this is expected to evolve over time.


src/pudl/output_diff/cli.py
```diff
@@ -191,11 +191,15 @@ def __init__(self, root_path: str, gcs_project=None):
...
-        }
+        # TODO(rousik): check if root_path is an actual directory.
+        if self.root_path.endswith(".sqlite"):
+            return {"sqlite": self.root_path}
+        else:
+            return {
+                os.path.basename(fpath): fpath
+                for fpath in self.fs.glob(self.root_path + "/*")
+                if self.match_filetype(fpath)
+            }
```
zaneselvans, line 202: We've been trying to stick to using the newer `pathlib` interface for paths across the project. Are there reasons why this depends on `os.path`?

src/pudl/output_diff/cli.py
```diff
@@ -0,0 +1,403 @@
...
+        Returns dictionary where keys are diff categories and values are GenericDiff
+        representing the differences at the given category.
+
+        The following categories of diffs are generated:
+        - filename/table_name, values are table names
+        - filename/table_name/schema, values are (col_name, col_type) tuples
+        - filename/table_name/rowcounts, values are [row_count]
+        - filename/table_name/rows, values are [distinct_row_count]
+        """
+        # TODO(rousik): using GenericDiff for rowcounts and rows are not ideal,
```
jdangerx, line 270: Seems like you might have added some more type-specific diff classes already? GenericDiff seems to be gone.

src/pudl/output_diff/cli.py
```diff
@@ -0,0 +1,403 @@
...
+        self.experiment = experiment
+        self.args = args
+
+    def compare_files(self) -> ListDiff:
+        """Produces GenericDiff that contains file basenames."""
+        baseline_files = set(self.baseline.list_files())
+        experiment_files = set(self.experiment.list_files())
+        return ListDiff.from_sets("File", baseline_files, experiment_files)
+
+    def compare_sqlite_databases(self, file_diff: ListDiff) -> list[CategoryDiff]:
```
jdangerx, line 258: You mentioned wanting to rework the depth flags above. Was your plan to configure which categories to diff and pass that into this method?

src/pudl/output_diff/cli.py
```diff
@@ -0,0 +1,403 @@
...
+        - filename/table_name/rowcounts, values are [row_count]
+        - filename/table_name/rows, values are [distinct_row_count]
+        """
+        # TODO(rousik): using GenericDiff for rowcounts and rows are not ideal,
+        # perhaps better approach would be to have type-specific diffs that allow
+        # either numbers for left/right (for rowcount) or panda dataframes for rows.
+        # But let's work with the hacky thing for now.
+        output_diffs = []
+
+        def maybe_append_diff(diff: CategoryDiff):
```
jdangerx, line 276: not that important: If we're already collecting & filtering the diffs at the end of `main()`, can we just do a normal append instead of a maybe append? We can shove the IO into `main()` like

```python
    all_diffs = [d for d in [file_diff] + sql_diffs if not d.sides_equal()]
    print("\n".join(str(d) for d in all_diffs))
```

src/pudl/output_diff/cli.py
```diff
@@ -0,0 +1,403 @@
...
+            """Appends diff to output_diffs if there are actual differences."""
+            if not diff.sides_equal():
+                output_diffs.append(diff)
+                diff.print_diff()
+
+        for fname in file_diff.both:
```
jdangerx, line 282: This seems like a reasonable way to encode "you can only diff table names between sqlite files that actually match", "you can only diff table structures between tables that are named the same", etc. 

If we were to add more types of diffs in the future, these dependencies might get a bit harder to parse out - maybe at that point we could switch to a pattern like

```python
        categories = ["file", "tablename", "schema", "row_count", "row"]
        category_funcs = {"file": ...}

        category_diffs = {}
        for category in categories:
            diff = category_funcs[category](category_diffs) # in each function, we'd look at the previous outputs to determine what to run diff calcs on
            category_diffs[category] = diff

        return [category_diffs[category] for category in categories]
```

We'd then have to make sure the `categories` list is in a specific order (bleh), but for each category func it would be clear which other categories it depends on.

If we wanted to go farther we could explicitly keep track of the deps in a dictionary, and then build a DAG, but that seems likely overkill :)

src/pudl/output_diff/cli.py
```diff
@@ -0,0 +1,403 @@
...
+            maybe_append_diff(table_diff)
+            for table_name in table_diff.both:
+                schema_diff = self.compare_table_schemas(fname, table_name)
+                maybe_append_diff(schema_diff)
+                # We could consider comparing tables that have divergent schemas,
+                # but that would be trickier.
+                if schema_diff.sides_equal():
+                    rc_diff = self.compare_table_rowcounts(fname, table_name)
+                    maybe_append_diff(rc_diff)
+                    # if rc_diff.sides_equal():
```
jdangerx, line 298: It might be worth diffing rows even if the row counts are different - I could imagine a situation where we're both missing data in the experiment and also have rows with completely different data.

src/pudl/output_diff/cli.py
```diff
@@ -0,0 +1,403 @@
...
+        erc = self.experiment.get_row_count(fname, table_name)
+        return RowCountDiff(
+            category=f"RowCount({fname}/{table_name})", left=brc, right=erc
+        )
+
+    def compare_table_rows(self, fname: str, table_name: str) -> RowSampleDiff:
+        """Compares individual rows within given file and table."""
+        # TODO(rousik): this may be very memory/resource expensive for tables
+        # with many rows. We might want to switch this off for very large tables
+        # or switch to some cheaper approach, e.g. statistical sampling or
```
jdangerx, line 342: Sampling above a configurable threshold seems like a pretty good approach to me!

jdangerx (commented):
I like the direction this is all going! I added one suggestion to (maybe) make the sql diff stuff easier to add to in the future, but I trust your judgment there.

Not approving/requesting changes yet since this seems like it's still a WIP - but let me know if you want a hard yes/no decision!

jdangerx:
Summary of call:

Merge blockers:
* [ ] write some high level regression tests

That's it - once we get this in active use, we will have a much better sense of what needs to change.

zaneselvans:
It'd be great to get this in and start using it and be able to iterate. I think it'll be a hugely useful tool!

src/pudl/output_diff/cli.py
rousik: I've decided to drop row-level comparisons entirely for now. I'm quite aware that this is going to be extra complicated and we might need to come up with some system to compare these without blowing up the memory for extremely large tables, but I think this is beyond the scope of the initial prototype. Perhaps we can compare row-counts and schemas and assume that most changes should impact those high-level cheap metrics.

Loading both tables into dataframes and comparing those directly is likely not feasible for very large tables with lots of rows, so I'm thinking that we could employ some kind of either statistical approach (sample bunch of records from either side and see if they are present on the other side), or full hashing approach where we could scan the two tables and calculate hash for each row (hash of hashes of columns) and compare those, or even do the above with sharded approach where we take subset of rows based on some hash-key. This might be necessary to support very large tables, but right now I think it's a bit too early to optimize in this space. Let's see how this goes first before embarking on algorithmically complex solutions.

rousik:
Work on this has been shifted to a separate repository.

PR #2630, opened 2023-06-05
Boostrap the new run-etl action

rousik:
This is the placeholder action that will allow us to run the ETL on google batch service. Right now it builds the CI version of the etl docker image (using sha as tag) but does nothing else so far.


zaneselvans:
@rousik Did you mean for this to have all of the intervening commits? Or did you just want to make a PR that creates the new action, so it can be dispatched manually from other branches?

rousik:
> @rousik Did you mean for this to have all of the intervening commits? Or did you just want to make a PR that creates the new action, so it can be dispatched manually from other branches?

Ugh, this was based off of `dev` which is why all the unrelated changes. Let me try to make another PR to fix this.

PR #2634, opened 2023-06-06
Update max rows for boils_eia860, plants_eia860, and pu_eia860 tables…

aesharpe:
Fix build fails from `test_minmax_rows` likely linked to #2587


aesharpe:
> Have you looked at the new records to see what changed? Is this new `plant_id_eia` + `boiler_id` values that we've never seen before? Are we already stripping leading zeroes on the boiler IDs in the new table? There are only 3 additional rows in the plants, but a few hundred in the boilers, which seems like kind of a lot.

There are 157 new year-plant-boiler combos. Of those, 20 are new boiler-plant combos (i.e., not just the same plant-boiler combo showing up in different years): 

|       |   plant_id_eia | boiler_id   |
|------:|---------------:|:------------|
|  3948 |           4040 | 3           |
|  3951 |           4040 | 2           |
|  3954 |           4040 | 1           |
|  4105 |           3492 | 4           |
|  4106 |           3492 | 3           |
|  4107 |           3492 | 2           |
|  4668 |           2384 | 6           |
|  4669 |           2384 | 4           |
| 47433 |          57664 | 3           |
| 50695 |           3138 | 2           |
| 50696 |           3138 | 1           |
| 61977 |          55642 | HRSG 3      |
| 61978 |          55642 | HRSG 2      |
| 61979 |          55642 | HRSG 1      |
| 66782 |          55341 | 1B          |
| 66783 |          55341 | 1A          |
| 66791 |          55330 | HRSG 4      |
| 66792 |          55330 | HRSG 3      |
| 66793 |          55330 | HRSG 2      |
| 66794 |          55330 | HRSG 1      |

I investigated some of them:
- The records from `plant_id_eia` 4040 could be a typo. The `boiler_eia860` table contains the `boiler_id` values 2CT2, 2CT1, 1CT1, 1CT2 which could coorespond to the new 1 and 2 `boiler_id` values, but it's not clear. Also there is no cooresponding record for the 3 `boiler_id` value.
- Similarly `plant_id_eia` 57664 new `boiler_id` 3 could correspond to old `boiler_id` CT3, but there's also a CT4
- The rest are just new

zaneselvans:
Hmm. If there are 20 truly new combinations of `(plant_id_eia, boiler_id)` then why wouldn't they have shown up as extra rows beyond the expected number of rows in the `minmax_rows` tests for `boilers_entity_eia`?

Ah. I see. Because we don't test for the number of rows in the entity tables. Ugh.  I can't wait for this output differ thing to get integrated.

Okay I guess there's nothing nefarious in here except some possible data entry errors... but I bet they aren't the only ones and we don't know how to identify them conclusively so. 👍🏼

zaneselvans (approved):

Issue #2622, opened 2023-06-02
enable proper treatment of inter-table xbrl meta calculations @cmgosnell


cmgosnell:
there seeming two types of inter-table calculations: 

1.  those that have calculation components that are entirely sourced from other tables
2. those that have *some* but not all of their calculation components from other tables.

In the exploded table context, the # 1 option seems easy to deal with. When

For but the # 2 option... what should we do? @jrea-rmi  We could:
* remove all of the sub-components in the  table. This would leave the reported total so the data would stay intact but we'd loose some detail from the sub-components. (easiest, less  detailed)
* track down the data in all of the source tables, grab that level of detail, run the calc checks/corrections. (most difficult, probably most accurate, most level of detail)
* run the calc checks w/ out the other table detail to generate a correction line (medium difficult, lil risky, preserves most detail)

An example to keep it tanglible:
in the `electric_operating_expenses_ferc1` table there are two calculated fields ('power_production_expenses_steam_power' & `power_production_expenses_hydraulic_power`) that each have one calculation component that is source from that plant tables:

```python
[{'name': 'steam_power_generation_operations_expense', 'weight': 1.0},
 {'name': 'steam_power_generation_maintenance_expense', 'weight': 1.0},
 {'name': 'operation_supervision_and_engineering_expense',
  'weight': 1.0,
  'source_table': ['plants_steam_ferc1',
   'plants_hydro_ferc1',
   'plants_pumped_storage_ferc1']}]
``` 

All examples of calcs with any inter-table calculation components
```python
income_statement_inter_table_calcs = [
    {
        'xbrl_factoid': 'power_production_expenses_steam_power',
        'table_name': 'electric_operating_expenses_ferc1',
        'calculations': [
            {'name': 'steam_power_generation_operations_expense', 'weight': 1.0},
            {'name': 'steam_power_generation_maintenance_expense', 'weight': 1.0},
            {'name': 'operation_supervision_and_engineering_expense',
             'weight': 1.0,
             'source_table': [
                 'plants_steam_ferc1',
                 'plants_hydro_ferc1',
                 'plants_pumped_storage_ferc1']}]},
    {
        'xbrl_factoid': 'power_production_expenses_hydraulic_power',
        'table_name': 'electric_operating_expenses_ferc1',
        'calculations': [
            {'name': 'hydraulic_power_generation_operations_expense',
             'weight': 1.0},
            {'name': 'hydraulic_power_generation_maintenance_expense',
[...]

PR #2635, opened 2023-06-06
add minor inter table calc fixes

cmgosnell:
# PR Overview

small pr to address some suggestions from https://github.com/catalyst-cooperative/pudl/issues/2622

additional changes this provoked:
- renaming `source_table` -> `source_tables` (bc it is a list of tables  - usually a list of 1 but still)
- migrating the de-duplication of the metadata based on the `xbrl_factoid` out of `process_xbrl_metadata` into it own method. this is by default just a check/a no-op, but now three table transformers have bespoke version. This was necessary because we added a calc to fix to a table that had duplicative metadata, which broke the calc fixer bc it expects to be able to extract **a** calculation from the metadata based on an `xbrl_factoid` name. So now the deduping of the metadata happens within `process_xbrl_metadata` but before `apply_xbrl_calculation_fixes`

# PR Checklist

### DIFF
--- a/docs/dev/run_the_etl.rst
+++ b/docs/dev/run_the_etl.rst
@@ -276,7 +276,7 @@ in the ``pudl.etl`` definition. Subsets of the ``pudl.etl`` asset graph
 are organized by asset groups. These groups are helfpul for visualizing and
 executing subsets of the asset graph.
 
-To execute the job, select ``fast_etl`` or ``full_etl`` and click "Materialize all".
+To execute the job, select ``etl_fast`` or ``etl_full`` and click "Materialize all".
 You can congifure which years to process by shift+clicking "Materialize all".
 Read the :ref:`resource_config` section to learn more.
 To view the status of the run, click the date next to "Latest run:".


src/pudl/transform/ferc1.py
```diff
@@ -1697,6 +1721,33 @@ def apply_xbrl_calc_fixes(self: Self, tbl_meta: pd.DataFrame) -> pd.DataFrame:
                         },
                     },
                 ],
+                "sales_to_ultimate_consumers": [
+                    # these next two fixes are related. we're replacing
+                    # commercial_and_industrial_sales which is from a different table
+                    # with two swappable components from this table. There is no way rn
+                    # to say swap thing A for thing B & C for two so we swap A for B and
+                    # then add C.
```
zaneselvans, line 1729: I don't understand what this comment is saying. What are "two swappable components?"

We're deleting one component and adding two components, right?

It seems like it might be simpler to have the more atomic "addition" and "deletion" operations rather than always defaulting to "replace" which is a composite of a deletion and an addition.  We could define a replacement method in terms of deletion and addition methods.

What happens when you replace a component that has a list of source tables with a component that doesn't have a list of source tables?

src/pudl/transform/ferc1.py
```diff
@@ -1489,6 +1489,18 @@ def apply_xbrl_calc_fixes(self: Self, tbl_meta: pd.DataFrame) -> pd.DataFrame:
...
                     },
+                    {
+                        "calc_component_to_replace": {
+                            "name": "operation_supervision_and_engineering_expense",
+                            "weight": 1.0,
+                            "source_table": [
+                                "plants_steam_ferc1",
+                                "plants_hydro_ferc1",
+                                "plants_pumped_storage_ferc1",
+                            ],
```
zaneselvans, line 1500: How does the list of source tables affect the behavior of this replacement operation?

I think it would be more intuitive if we used `source_tables` since this element is a list of tables.

src/pudl/transform/ferc1.py
zaneselvans: I think what's happening is

```py
# Replace commercial_and_industrial_sales which is reported in
# electricity_sales_by_rate_schedule_ferc1 with two components
# (small_or_commercial_sales... and large_or_industrial_sales...)
# from this table which should sum to the same amount as the value
# being replaced.
```

cmgosnell: code wise i hard agree that having "additions" and "deletions" would make way more sense. As a human conceptually it makes way more sense to say "replace A for B" (and A or B can be a null element which results in "add B" or "delete A" respective).

i only added that comment bc rn the fixes are setup as "replace A for B" not "replace A for B & C" which is what needed to happen here. but that needed to be added as "replace A for B" & "add C".

i'd love a deeper structural change for these calcs but that is for  #2605.

src/pudl/transform/ferc1.py
cmgosnell: the fixer looks for the full calculation component. and these calculation components contain this source table info so it needs to be here for the fixer to actually find the component.

and hard agree on `source_tables` i'll go change that now.

zaneselvans (approved):

PR #2637, opened 2023-06-06
Add business_model, service_type to sales_eia861 PK

### LINKED ISSUE #2636: Missing delivery utilities in sales_eia861
### Describe the bug
**Hi all! I've been pulling data from the sales to ultimate customer sheet of 861, and I realized that, at least for the 2021 data, there is only a single delivery utility (service_type == delivery) in the data, which is Rockland Electric Co (utility_id_eia == 16213). Whereas in the actual EIA 861 form, there are over 70 utilities when I filter for delivery under service type. Is this data missing, or should I be pulling from another data source?**

### Bug Severity
 - **High:** This bug is preventing me from using PUDL.
**Because I need the delivery utilities in 861, I am required to manually download the 861 excel forms and use that for now.**

### To Reproduce
Steps to reproduce the behavior -- ideally including a code snippet that causes the error to appear.
**(This is in R, but I'm basically just pulling the sales_eia861 sheet, filtering for 2021, and then filtering for delivery service_type utilities)**

```
file = "pudl.sqlite"
pudl <- dbConnect(SQLite(), file)

sales_ult_cust_861 <- dbReadTable(pudl, "sales_eia861") %>%
  filter(year(report_date) == 2021)

delivery_utilities <- sales_ult_cust_861 %>% 
  filter(service_type == "delivery")
```

 - Is the bug related to the software / database? If so, please attach the `settings.yml` file you're using to specify which data to load, and make a note of where in the ETL process the error is happening.
**N/A**

 - Have you found an error or inconsistency in the data that PUDL brings together? If so, what is the data source, year, plant_id, etc
**I haven't thoroughly checked all years of the data, but it seems like a few delivery utilities appear in other years, but not the full amount of the data**

 -- how can we find the data you're looking at, and what is the nature of the error or inconsistency.
**See above**

 - Does the problem happen when you're starting up the database (importing data) or does it happen later when you're trying to use the database?
**This is after the data is imported when I am trying to use it**

### Expected behavior
A clear and concise description of what you expected to happen, or what you expected the data to look like.
**All the utilities to show up in sales_eia861, including the utilities that are "delivery" by "service_type"**

### Software Environment?
 - Operating System. (e.g. MacOS 14.5, Ubuntu 22.04, Windows Subsystem for Linux v2)
**MacOS 13.3.1**

 - Python version and distribution (e.g. Anaconda Python 3.10.6)
**R 4.2.2**

 - How did you install PUDL?
    - If you installed using `git clone` what branch are you using (probably `main` or `dev`)
    - If you installed using `pip`, `conda` or `mamba` what version did you install?
**I downloaded the SQLite with the following link: https://s3.us-west-2.amazonaws.com/intake.catalyst.coop/dev/pudl.sqlite**

### Additional context
**N/A**

zaneselvans:
# PR Overview

We were using an incomplete set of columns as the primary key in the `sales_eia861` table, but didn't end up with any duplicates because those additional values in those columns were being lost in a reshaping operation. This problem was reported in #2636 by @christiantfong.

I've added `business_model` and `service_type` to the PK for this table, resulting in nearly 20,000 additional rows in the table.

We should add some kind of check into the `pudl.transform.eia861._tidy_class_dfs()` function to catch this type of error, since it could be present in other EIA-861 tables, and this function is used 20 times... See #2638 

I clobbered all the alembic migrations because for some reason just changing the PK fields was not resulting in a change to the primary key definition for the table -- instead all it did was make the new PK fields non-nullable -- so I kept getting duplicate primary key errors (since the new PK fields weren't being used).

Closes #2636 

# PR Checklist


jdangerx:
I think this will be OK! We're using the migrations primarily to speed up development, not because we have a production database that needs in-place schema changes - so the worst case scenario is that people have to re-generate their databases from scratch again.

e-belfer (approved):
Re-ran the fast ETL from scratch with the new migration. All looks as expected and the PK is unique. Along with the PK assertion discussed in #2638, would probably be good to add the EIA 861 expected rows into our validations at some point to make it easier to catch changes in the outputs here. This is definitely out of scope of this issue, though!

Issue #2293, opened 2023-02-08
Remove `pudl_settings` and `pudl_setup` in favor of PUDL_INPUT and PUDL_OUTPUT


jdangerx:
```[tasklist]
### Scope
- [x] pudl_setup command is gone
- [x] pudl_settings object is not used anymore
- [x] "Creating a workspace" portion of development docs are updated
- [x] `pudl-examples` notebooks work
```

```[tasklist]
### Next steps
- [x] go through list that @bendnorman put in the comments and start removing references to pudl_settings.
```

bendnorman:
To deprecate pudl_settings and pudl_setup we'll need to remove all references to `pudl_settings` and `pudl.workspace.setup.get_defaults()`. Locations I could find:

- [ ] `pudl.analysis.service_territory`:gets default pudl_settings and creates a SQL engine to pass to `PudlTabl`. It also passes pudl_settings to `pudl.output.censusdp1tract.get_layer()`.
- [ ] `pudl.analysis.state_demand`: loads default and creates SQL engine.
- [ ] `pudl.convert.censusdb1tract_to_sqlite`: gets sqlite output path from pudl_settings. Also gets the defaults to initializing a datastore.
- [ ] `pudl.convert.datasette_metadata_to_yml`: passes pudl_settings to `DatasetteMetadata.from_data_source_ids(pudl_settings=pudl_settings)`.
- [ ] `pudl.convert.epacems_to_parquet`: will mostly be handled by #2292
- [ ] `pudl.extract.ferc1`: Mostly references will be removed and functionality has already been converted to dagster.
- [ ] `pudl.extract.ferc714`: References pudl_settings but I don't actually think it uses it?
- [ ] `pudl.glue.ferc1_eia`: `get_plants_ferc1_raw` uses pudl_settings but I'm not sure what the function is actually used for. Looks like it's just used for an integration test @zaneselvans?
- [ ] `pudl.metadata.classes`: DatasetteMetadata.from_data_source_ids() and Datasource.add_datastore_metadata() need pudl_settings to construct a datastore.
- [ ] `pudl.output.censusdptract`: gets defaults and passes to `censusdp1tract_to_sqlite`
- [ ] `pudl.output.epacems`: gets cems path from defaults
- [ ] `pudl.output.ferc714`: Just passes pudl_settings to pudl.output.censusdp1tract.get_layer()
[...]

PR #2640, opened 2023-06-07
Simplify config reading and path configurations

rousik:
We are now fully committed to using PUDL_INPUT and PUDL_OUTPUT env variables.

This refactoring removes the legacy support for reading yml config files with path settings and removes the ubiquitous warning that plagued our ETL runs.

This also replaces the type-less dict `pudl_settings` with `PudlPaths` pydantic settings object. In addition, `pudl_output_path` and other path-like configuration fields are also removed from the dagster `@op` config and is pulled from the `pudl_paths` resource (in fact, ops can directly read from `PudlPaths` and we can assume that the paths are correctly configured via `pudl_paths`)

Some file paths that were contained in `pudl_settings` (mainly databases and metadata file names for specific datasets) were removed and it is now up to the each module that accesses such file to construct the right paths. Ideally, we would move this metadata (files used by given dataset) somewhere dataset-specific rather than aggregating this in rather generic `PudlPaths`. This part is still WIP and will likely need to be resolved before we consider this ready to be merged.

Overall, there seems to be a lot of ways how these paths could be setup. While this reduces some complexity it still seems like it should be possible to simplify further.


src/pudl/workspace/setup.py
```diff
@@ -58,139 +46,38 @@ def get_defaults(
         os.environ["PUDL_OUTPUT"] = str(Path("~/pudl-work/output").expanduser())
         os.environ["PUDL_INPUT"] = str(Path("~/pudl-work/data").expanduser())
 
-    yaml_settings = _munge_legacy_yaml(
-        yaml_file=yaml_file, default_pudl_yaml=default_pudl_yaml
-    )
+    if input_dir:
```
jdangerx, line 49: Overriding the env vars here makes `input_dir` and `output_dir` arguments persist in subprocesses... that seems like less surprising behavior than before!

src/pudl/workspace/setup.py
```diff
@@ -58,139 +46,38 @@ def get_defaults(
...
     Args:
-        pudl_in (os.PathLike): Path to the directory containing the PUDL input
-            files, most notably the ``data`` directory which houses the raw
-            data downloaded from public agencies by the
-            :mod:`pudl.workspace.datastore` tools. ``pudl_in`` may be the same
-            directory as ``pudl_out``.
-        pudl_out (os.PathLike): Path to the directory where PUDL should write
-            the outputs it generates. These will be organized into directories
-            according to the output format (sqlite, parquet, etc.).
+        pudl_in (Path): directory containing PUDL input files, most notably
```
jdangerx, line 71: I think this is now the `data` directory itself, not the parent folder.

src/pudl/workspace/setup.py
```diff
@@ -58,139 +46,38 @@ def get_defaults(
...
-        )
-
-    settings = derive_paths(settings["pudl_in"], settings["pudl_out"])
-
-    # override env vars so Dagster can see the most up-to-date configs
-    if output_dir or "PUDL_OUTPUT" not in os.environ:
-        os.environ["PUDL_OUTPUT"] = settings["pudl_out"]
-    if input_dir or "PUDL_INPUT" not in os.environ:
-        os.environ["PUDL_INPUT"] = settings["data_dir"]
+    settings = derive_paths(
```
jdangerx, line 58: It sort of seems like we could inline `derive_paths` here too - then we could override other things beyond `input_dir` and `output_dir`.

I guess if everything is derived from the input/output dirs then that just allows us to accidentally override ourselves into weird states, though, so maybe this is better.

jdangerx (changes requested):
I'm so glad you went through and did this! Yes, there's definitely more streamlining we can do with the setup process, but this is already a breath of fresh air :)

A couple blocking comments:
* a docstring update
* I think it's worth testing the override behavior a bit more thoroughly

But overall this looks great and also cleans up a lot of my previous mistakes, so thanks!

test/unit/pudl_environment_test.py
```diff
@@ -18,173 +15,14 @@ def setup():
...
-            yaml.dump(
-                {
-                    "pudl_in": "/test/workspace",
-                    "pudl_out": "/test/workspace",
-                }
-            )
-        ),
-    ],
-)
-def test_get_defaults_in_test_environment_use_env_vars_tmpdir(
```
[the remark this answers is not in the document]
rousik: I wonder if we could consider a decoupling such that we have one method `set_path_overrides(**kwargs)` to potentially set overrides (which would assign to env variables), and  `get_defaults()` would not have any override functionality at all.

It seems that situations where we need to actually override these are very infrequent and perhaps we could completely get rid of this, assuming that we can set env variables ahead of time. I think the current complexity is probably residue of the merging of these various sources of truth and now there's only one left anyways.

src/pudl/workspace/setup.py
rousik: Yeah, the non-consistent semantics for input/output env variables is somewhat strange and counter-intuitive. I was caught off guard by this few times.

src/pudl/workspace/setup.py
rousik: I have inlined all of this and refactored somewhat. Generally I think the good idea is to reduce the number of tuning knobs so I think that having simply just input/output should be enough. There should be no need to have too much flexibility around this IMO.

test/unit/pudl_environment_test.py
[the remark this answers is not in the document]
zaneselvans: IIRC in #2293 we were envisioning using the [Pydantic BaseSettings](https://docs.pydantic.dev/latest/usage/settings/) model instead of the ubiquitous `pudl_settings` dictionary that's passed around everywhere now. Pydantic settings provide a well defined system for overriding defaults including using envvars.

I think the main cases in which we were overriding these paths is in the tests, where we want to separate the test impacts from the user's working environment.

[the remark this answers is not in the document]
rousik: Hmm, yeah, Pydantic `BaseSettings` look promising. I will see if I could switch to that as part of this.

rousik:
The last commit replaces the ad-hoc `pudl_settings` and `get_defaults()` accessors with `PudlPaths()` pydantic object which is configured exclusively through env variables. This has been tested very briefly w/ `ferc_to_sqlite` with fast settings.

The following may still need to be evaluated:
- [ ] full ETL execution
- [ ] doc builds 
- [ ] unit tests/data tests

rousik:
Woohoo, looks like this PR is net-negative LOCs which is always a thing to strive for :D

src/pudl/workspace/setup_cli.py
```diff
@@ -102,20 +103,17 @@ def main():
...
     )
 
-    if args.pudl_input:
+    if args.pudl_in:
         pudl_in = pathlib.Path(args.pudl_in).expanduser().resolve()
         if not pathlib.Path.is_dir(pudl_in):
             raise FileNotFoundError(f"Directory not found: {pudl_in}")
-
-    if args.pudl_output:
+        set_path_overrides(input_dir=pudl_in)
```
zaneselvans, line 110: Does using this global envvar override in the CLI have the side effect of changing the value of the environment variable in the user's shell? Is that the desired outcome here?

src/pudl/workspace/setup.py
```diff
@@ -4,310 +4,120 @@
...
-
-    Input/output workspace roots can be the same directories.
+    Args:
+        input_dir: if set, overrides PUDL_INPUT env variable.
+        output_dir: if set, overrides PUDL_OUTPUT env variable.
+    """
+    if input_dir:
+        os.environ["PUDL_INPUT"] = input_dir
+    if output_dir:
+        os.environ["PUDL_OUTPUT"] = input_dir
```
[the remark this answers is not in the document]
zaneselvans: In general what's the idea behind using this global override of the environment variables (will it persist in the user's environment?) rather than creating a new `PudlPaths` object with arguments that override whatever the environment variables are within the scope of the use of that object?

It seems like the `BaseSettings` class is already designed to allow overriding the defaults via arguments when initializing the class, the `.env` file, or environment variables.

For the CLIs would it be better to just tell users "Set your environment variables or use .env" rather than having the arguments?  And then in the tests, since we want them to be isolatable from the user's environment, use a `PudlPaths` fixture that's been created with the appropriate test-only paths?

test/conftest.py
```diff
@@ -87,15 +87,6 @@ def pytest_addoption(parser):
     )
 
 
-@pytest.fixture(scope="session")
-def pudl_env(pudl_input_output_dirs):
```
zaneselvans, line 91: Looks like there are several remaining references to this fixture in `conftest.py`

test/conftest.py
```diff
@@ -392,54 +381,36 @@ def pudl_tmpdir(tmp_path_factory):
...
+            in_tmp.mkdir()
+            pudl.workspace.setup.set_path_overrides(
+                input_dir=str(Path(in_tmp).resolve()),
+            )
+        if request.config.getoption("--live-dbs"):
+            out_tmp = pudl_tmpdir / "output"
+            out_tmp.mkdir()
+            pudl.workspace.setup.set_path_overrides(
+                output_dir=str(Path(out_tmp).resolve()),
+            )
```
zaneselvans, line 413: This logic seems a little funny to me -- if we've gotten `--live-dbs` then currently we're using the user's already-existing databases rather than building temporary databases from scratch, so the output directory should be the user's default output directory shouldn't it? And not a temporary location? Or are the paths to the user's PUDL DB and FERC DBs being set independently somewhere else?

test/conftest.py
```diff
@@ -392,54 +381,36 @@ def pudl_tmpdir(tmp_path_factory):
...
-    tmpdir.mkdir()
-    return tmpdir
+def pytest_sessionstart(session):
+    """Configures input/output paths for the tests."""
+    # TODO(rousik): Should we be using fixed paths instead
+    # of using tmpdir capabilities here?
+    pudl.workspace.setup.set_path_overrides(
+        input_dir="~/pudl-work/data",
+        output_dir="~/pudl-work/output",
+    )
```
zaneselvans, line 391: Why are hard-coded Unix paths a good idea here?  Don't we want the test outputs to go to a pytest temporary directory for isolation from the user's environment?

src/pudl/workspace/datastore.py
```diff
@@ -609,7 +593,18 @@ def main():
...
+
+    cache_path = None
+    if not args.bypass_local_cache:
+        cache_path = PudlPaths().input_dir
+
+    dstore = Datastore(
+        gcs_cache_path=args.gcs_cache_path,
+        sandbox=args.sandbox,
+        local_cache_path=cache_path,
+    )
```
zaneselvans, line 607: Does it make sense to integrate the remove cache URLs / remote output URLs into the settings object?

zaneselvans (changes requested):
Left a bunch of questions. Mostly I'm confused about the need for or desirability of the path overrides function, which seems like it will have unexpected side-effects.

src/pudl/workspace/setup.py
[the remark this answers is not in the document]
rousik: If we can drop the support for command-line flags that would simplify things a bit.

re/ overrides - it seems to me that in many cases, the path configuration was accessed via `get_defaults()` which implies that it should have globally consistent value. I would further also assume that when you set the overrides, it should be okay to have it set to this value until the end of the process or until someone else sets up another override. 

Now that I think about it, it might be possible that some unit/integration tests assume that they run in a hermetic environment where inputs/outputs are isolated in between tests that run in the same process. But, given the above widespread use of `get_defaults()` I'm not sure if this behavior can be safe.

I generally like the idea of being able to get the paths simply by accessing `PudlPaths()` w/o having to pass this config around. For something that is stable/unchanging and globally consistent, this seems like a straightforward choice. Even for unit tests, we could consider simply setting the ENV variables in the `tox.ini` or via some other mechanism at the startup time.

src/pudl/workspace/setup.py
```diff
@@ -4,310 +4,120 @@
...
+
+    def sqlite_db(self, name: str) -> str:
+        """Returns url of locally stored pudl slqlite database with given name.
+
+        The name is expected to be the name of the database without the .sqlite
+        suffix. E.g. pudl, ferc1 and so on.
+        """
+        db_path = PudlPaths().output_dir / f"{name}.sqlite"
+        return f"sqlite:///{db_path}"
+        return self.output_dir / f"{name}.sqlite"
```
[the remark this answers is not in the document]
rousik: Right, I wanted to have this conversation. I think that dumping all the dataset-specific filenames in this global config is a bit of an anti-pattern. It breaks the isolation in that every time you add new dataset, the associated metadata (incl. filenames) will be encoded in the dataset specific objects (e.g. configs, dataset descriptors and so on) and then you will have some of the metadata dumped in this grab-bag config here.

It would be IMO preferable if we could move the generation of filenames into dataset-specific objects/metadata. This object/function could access `PudlPaths` to construct absolute pathnames. I'm not quite sure where this would naturally fit so I haven't done it yet, but this would be the ideal direction to take this in.

WDYT?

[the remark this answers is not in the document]
rousik: For example (unsure if `DatasetSettings` is the right place to put this), we could consider something along the lines of:

```python
class Ferc1Settings(GenericDatasetSettings):
  ...
  def get_output_database(self) -> Path:
    return PudlPaths().output_dir() / "ferc1.sqlite"
```

src/pudl/workspace/setup.py
```diff
@@ -4,310 +4,120 @@
...
+        return self.output_dir / filename
+
+    class Config:
+        """Pydantic configuration. Loads from .env file."""
+
+        env_file = ".env"
+        fields = {
+            "pudl_input": {"env": "PUDL_INPUT"},
+            "pudl_output": {"env": "PUDL_OUTPUT"},
+        }
```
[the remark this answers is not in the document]
rousik: Is it not case sensitive then?

src/pudl/workspace/setup_cli.py
```diff
@@ -102,20 +103,17 @@ def main():
...
         pudl_out = pathlib.Path(args.pudl_out).expanduser().resolve()
         if not pathlib.Path.is_dir(pudl_out):
             raise FileNotFoundError(f"Directory not found: {pudl_out}")
-
-    settings = pudl.workspace.setup.get_defaults(
-        input_dir=args.pudl_input, output_dir=args.pudl_output
-    )
-    pudl.workspace.setup.init(settings, clobber=args.clobber)
+        set_path_overrides(output_dir=pudl_out)
+    pudl.workspace.setup.init(colbber=args.clobber)
```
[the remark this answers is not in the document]
rousik: Steven Clobbert? 😆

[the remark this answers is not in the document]
rousik: Also, wow, this seems like a dangerous typo.

rousik:
> Left a bunch of questions. Mostly I'm confused about the need for or desirability of the path overrides function, which seems like it will have unexpected side-effects.

btw, overrides function essentially replaces the previously existing `set_defaults()` method that now no longer exists.

test/unit/pudl_environment_test.py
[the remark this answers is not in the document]
rousik: Conversion to pydantic settings is now done.

src/pudl/workspace/setup_cli.py
rousik: AFAIK the env variables are bound to process. You can inherit from your parent but once new process is started, any changes that would be done to the env would not be propagated back to the parent. Python (and also shell subprocesses) are new processes, hence they should be isolated from the shell that starts them and they should have no impact on the parent's env.

If this was not the case (i.e. it would be easy to accidentally modify shell envs), things would go horribly wrong everywhere all the time IMO.

rousik: Perhaps this also explains why there was so much emphasis on reverting env variables to former state? I was wondering why are we doing it at all.

test/conftest.py
rousik: Thanks, dropped those.

test/conftest.py
rousik: Um, you're right, this doesn't sound right at all.

test/conftest.py
rousik: Uh, indeed. I must have conflated this with the `READTHEDOCS` case. Will also fix.

src/pudl/workspace/datastore.py
rousik: Yep, this is indeed bit of a gray area IMO. What is the overall ETL configuration (inputs/outputs certainly), and what is the configuration of the individual pipeline run (e.g. where we'd configure datasets, years, perhaps output destinations). In essence, I'm assuming that PUDL_INPUT/PUDL_OUTPUT remain fixed within an environment (whether this is local or docker image) and etl configuration varies. Output destination seems like part of the etl config, perhaps also gcs cache configuration, although there we might expect more fixed setup as well.

src/pudl/workspace/setup.py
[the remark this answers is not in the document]
zaneselvans: I think it's case-insensitive by default (so UPPER_CASE for envvars) but you can also [configure this](https://docs.pydantic.dev/latest/usage/pydantic_settings/#case-sensitivity).

zaneselvans:
It looks like there's some issue with the path to the CEMS parquet files in the tests. It might also be that the tests that depend on those files having been produced no longer have a fixture dependency that forces them to wait until after the ETL has happened to run? Or that for some reason the order in which pytest is running the tests has changed and they just coincidentally happened to be running after the ETL before.

src/pudl/workspace/setup.py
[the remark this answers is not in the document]
zaneselvans: @jdangerx do you have thoughts here?  I can see generating these paths based on metadata that's stored within the dataset being a good idea, but I'm nervous that if we don't set the paths up to be correct now (even if they're hard coded inside the PudlPaths for now) we may not get around to fixing the paths later.

Also given that we have the JSON version of the XBRL metadata, the SQLite DB outputs, the datapackage.json files, and in some cases the Parquet outputs, it seems like a substantial amount of information to potentially add to the dataset metadata -- and it's information about where the data ends up, rather than where it comes from, which is a bit different.

If we specify the paths or rules for generating them inside `PudlPaths` now but access them via a method, we can change where that information comes from later if/when we add it to the dataset metadata.

src/pudl/workspace/setup_cli.py
zaneselvans: I think it's the very strange locale / envvar behavior we had with Arelle early on that is making me uneasy here. I don't remember exactly what was going on but it resulted in some kind of persistent change that broke other stuff. Maybe @zschira remembers more...

src/pudl/workspace/datastore.py
zaneselvans: @jdangerx do you have thoughts on this?

src/pudl/workspace/setup.py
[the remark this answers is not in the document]
rousik: The problem imo is that PudlPath becomes an object that needs to know a little bit about every dataset and that goes against modularity in a big way. The output paths are output_dir + base_filename. PudlPaths should be responsible for the first part but it should be dataset specific method/object that generates the latter (and it could easily combine its stuff w/ PudlPaths(). output_dir). 

I'm happy to do this refactor as part of this PR, we just have to decide on where to put it.

In theory, we could even make it Ferc1ToSqliteSettings.output_db. This may not be entirely clean but perhaps acceptable? Settings should be available where you need to connect to the output files and it could reach into PudlPaths for directory where this lives.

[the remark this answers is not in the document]
jdangerx: Jan's vision of 
> PudlPaths should be responsible for the first part but it should be dataset specific method/object that generates the latter (and it could easily combine its stuff w/ PudlPaths(). output_dir).
makes a lot of sense to me - I know that when working with the previous settings object, generating all these paths in one place got pretty confusing.

Zane, is your point here:
> Also given that we have the JSON version of the XBRL metadata, the SQLite DB outputs, the datapackage.json files, and in some cases the Parquet outputs, it seems like a substantial amount of information to potentially add to the dataset metadata -- and it's information about where the data ends up, rather than where it comes from, which is a bit different.

that basically, each DatasetSettings would end up with a different set of output paths? Like the `Ferc1Settings` would have a `.dbf_db`, a `.xbrl_db`, a `.datapackage`, and a `.taxonomy_metadata`... and then `CensusDP1Settings` would just have a `.sqlite_db`? That seems fine, since each different dataset settings is a different class.

My vote would be for each subclass of `GenericDatasetSettings` to then include the dataset-specific output paths, and each different dataset has its own set of output paths which makes sense.

I understand the worry about not following up on the refactor, but I think this PR has lingered long enough that we should probably merge this and then do a follow-up.

src/pudl/workspace/setup.py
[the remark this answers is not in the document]
jdangerx: Pydantic BaseSettings does read env vars in a case-INsensitive way, by default.

I would probably just set the field names to `input` and `output` and then set `env_prefix = "PUDL_"` in the Config, personally. But if you want to keep the field names `PudlPaths().pudl_input`/`PudlPaths().pudl_output` then I think you don't need to set anything in `PudlPaths.Config` except for `env_file`.

src/pudl/workspace/setup_cli.py
jdangerx: We did a bunch of reverting env vars to former state because they're not automatically cleaned up between tests. Setting them here in the CLI seems fine to me!

src/pudl/workspace/setup.py
```diff
@@ -4,310 +4,120 @@
...
     settings_pkg = "pudl.package_data.settings"
     deploy(settings_pkg, settings_dir, ignore_files, clobber=clobber)
 
     # Make output directory:
-    pudl_out = pathlib.Path(pudl_settings["pudl_out"])
-    pudl_out.mkdir(parents=True, exist_ok=True)
+    PudlPaths().output_dir.mkdir(parents=True, exist_ok=True)
+    # TODO(rousik): it might make sense to turn this into a method of
+    # PudlPaths object and to move this to settings.py from this module.
+    # Unclear whether deployment of settings files makes much sense.
```
jdangerx, line 120: I personally would like to avoid deploying these settings files, but that definitely seems out of scope for this PR.

test/conftest.py
```diff
@@ -392,54 +381,39 @@ def pudl_tmpdir(tmp_path_factory):
...
-
-
-@pytest.fixture(scope="session")
-def pudl_input_tmpdir(pudl_tmpdir):
-    tmpdir = pudl_tmpdir / "data"
-    tmpdir.mkdir()
-    return tmpdir
+def pytest_sessionstart(session):
+    """Configures input/output paths for the tests."""
+    # TODO(rousik): Should we be using fixed paths instead
```
jdangerx, line 386: It seems like we're setting up the `PudlPaths` here and also in `pudl_path_setup`.

In theory, we want the *inputs* to default to whatever our env var points to, because we don't want to have to redownload a bunch of data. And then we want the *outputs* to go to a tmpdir so we don't pollute our dev outputs - which is what the `pudl_path_setup` function does better than `pytest_sessionstart`.

But that only works when `pudl_path_setup` is explicitly requested - should we just pull that logic into `pytest_sessionstart`?

src/pudl/workspace/setup.py
```diff
@@ -4,310 +4,120 @@
...
 
-    Determines input/output directory locations from YAML, then overrides with
-    env vars, then overrides with keywords passed in.
-
-    Input/output workspace roots can be the same directories.
+    Args:
+        input_dir: if set, overrides PUDL_INPUT env variable.
+        output_dir: if set, overrides PUDL_OUTPUT env variable.
+    """
+    if input_dir:
```
jdangerx, line 25: The reason we want to use env vars here instead of just initializing the `PudlPaths()` with different parameters is so that child processes get the right settings automatically, right?

src/pudl/workspace/setup.py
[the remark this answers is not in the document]
jdangerx: I can also see that the configs are easier to _access_ when they're all in one place, but easier to _generate_ when they're set up separately. What do you think of some sort of container object that holds *all* the configs @rousik @zaneselvans ?

jdangerx (changes requested):
This looks pretty good! I've a few small suggestions about `BaseSettings` features that we could use, and also a vote for "separate the dataset output paths from `PudlPaths` in a follow-up PR to this one."

src/pudl/workspace/setup.py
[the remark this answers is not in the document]
rousik: Can you clarify what you mean by all the configs? Are we talking about the etl settinings, PudlPaths and other such configurations?

As it stands now, I'm expecting that `PudlPaths()` can simply be instantiated wherever it is needed (it reads env variables and so it should always give you the right answer) so there's no need to pass that through.

AFAIK there's really one root etl settings from which you can pull out sub-components (e.g. `FercToSqliteSettings`) so there is an enclosing configuration.

There are few stray configuration fields that are standalone (e.g. datastore configuration fields such as `gcs_cache_path` and `sandbox`; or `clobber` field that seems to be set in various parts of the system).

It would be good idea to consolidate all of these into some enclosing structures, possibly even embed those in the etl settings object, but this can be done independently of this PR.

Let me know if I understand you point here or whether you were thinking about something else here.

src/pudl/workspace/datastore.py
rousik: I'd suggest adding TODO here with subsequent cleanup to keep the scope of this PR from growing?

src/pudl/workspace/setup.py
rousik: Seems in-line with the effort of moving to data-only releases. Even with those, it's easy to point pudl at the settings files from the source repo IMO.

src/pudl/workspace/setup.py
rousik: Yup, that's the point. We call `set_path_overrides` early on and then assume that `PudlPaths` always returns the right things from then on. I think we may have passed pudl_settings, but this was not actually hermetic and many call-sites still got their own instance via `get_defaults()` which would break the consistency.

src/pudl/workspace/setup.py
```diff
@@ -4,310 +4,120 @@
...
-    Args:
-        pudl_in (os.PathLike): Path to be used as the default input directory
-            for PUDL -- this is where :mod:`pudl.workspace.datastore` will look
-            to find the ``data`` directory, full of data from public agencies.
-        pudl_out (os.PathLike): Path to the default output directory for PUDL,
-            where results of data processing will be organized.
-        clobber (bool): If True and a user settings file exists, overwrite it.
-            If False, do not alter the existing file. Defaults to False.
+    pudl_input: str
+    pudl_output: str
```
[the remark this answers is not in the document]
rousik: Actually, turns out that `input` is an internal python builtin that will get shadowed so we probably better keep `pudl_input` and `pudl_output` here.

test/conftest.py
rousik: Okay. I have switched this to session scoped `configure_paths_for_tests` fixture that has `autouse=True` which essentially executes that before any tests regardless of whether it is explicitly requested or not. That should cover all the cases.

src/pudl/workspace/setup.py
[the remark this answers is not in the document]
rousik: I still think that the best approach might be having per-dataset configuration object (akin to `PudlPaths`) that would implicitly construct absolute paths by combining `PudlPaths` base paths with the dataset specific relative paths/filenames. It might make sense to have few global paths embedded in `PudlPaths` (e.g. pudl database).

I think that single object that aggregates everything from across all datasets is going to be tricky to scale with number of supported datasets. That said, it might make sense to use _registry pattern_ here and have a way to retrieve dataset specific config via dataset name.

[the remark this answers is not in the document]
rousik: The above approach maintains modularity, but allows central access to all of these via the registry in situations where appropriate.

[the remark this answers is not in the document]
jdangerx: Yeah, a registry would be a nice way to not have to keep aggregating settings into some central SuperConfig object!

What do you think of having the `DatasetSettings` construct the dataset paths, vs. having a new `DatasetPaths` object that just handles those paths?

zaneselvans:
It looks like the docs builds are failing because of some PUDL Path related issues.
```
Extension error:
Handler <function data_sources_metadata_to_rst at 0x7f01624d1f80> for event 'builder-inited' threw an exception (exception: 2 validation errors for PudlPaths
pudl_input
  file or directory at path "/home/runner/pudl-work/data" does not exist (type=value_error.path.not_exists; path=/home/runner/pudl-work/data)
pudl_output
  file or directory at path "/home/runner/pudl-work/output" does not exist (type=value_error.path.not_exists; path=/home/runner/pudl-work/output))
```

rousik:
> It looks like the docs builds are failing because of some PUDL Path related issues.
> ```
> Extension error:
> Handler <function data_sources_metadata_to_rst at 0x7f01624d1f80> for event 'builder-inited' threw an exception (exception: 2 validation errors for PudlPaths
> pudl_input
>   file or directory at path "/home/runner/pudl-work/data" does not exist (type=value_error.path.not_exists; path=/home/runner/pudl-work/data)
> pudl_output
>   file or directory at path "/home/runner/pudl-work/output" does not exist (type=value_error.path.not_exists; path=/home/runner/pudl-work/output))
> ```

Aye, turns out that DirectoryPath validates existence. This is now fixed and will be simplified once we switch to pydantic V2+

zaneselvans:
Cool! We'll have to wait for [Dagster to update to Pydantic 2](https://github.com/dagster-io/dagster/issues/15162), hopefully it won't take too long.

CI is occasionally failing now because we're running out of memory, which I think is what happened here. I've got it trying again now.

zaneselvans (approved):

jdangerx (approved):
Thanks for all the work overhauling our settings system! There's still more to do but this is much nicer as far as I'm concerned.

Issue #2016, opened 2022-10-24
Enable aggregation of reshaped FERC DBF/XBRL data


zaneselvans:
Organize the per-row/col IDs compiled in #2015 such that common / useful aggregations of the reported financial data can be recreated programmatically.

```[tasklist]
# Phase 1
- [x] #2604 @cmgosnell
- [x] spot check xbrl metadata calculations within `income_statement_ferc1`: usable or not? @zaneselvans
- [x] spot check xbrl metadata calculations within `depreciation_amortization_summary_ferc1`: usable or not? @cmgosnell
- [x] spot check xbrl metadata calculations within `electric_operating_expenses_ferc1` @cmgosnell
- [x] Check metadata/calculations for: `balance_sheet_assets_ferc1` @e-belfer
- [x] Check metadata/calculations for: `utility_plant_summary_ferc1` @e-belfer
- [x] Check metadata/calculations for: `plant_in_service_ferc1`
- [x] https://github.com/catalyst-cooperative/pudl/issues/2579 @e-belfer
- [x] check metadata/calculations in `balance_sheet_liabilities_ferc1` @e-belfer
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2581
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2584
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2597
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2619
```

```[tasklist]
# Phase 2 - Draft Outputs!
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2646
- [ ] https://github.com/catalyst-cooperative/pudl/issues/1811
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2599
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2683
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2623
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2625
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2624
```
```[tasklist]
### Phase 3 - Refinements/Validations
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2700
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2738
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2605
- [ ] https://github.com/catalyst-cooperative/pudl/issues/2736
[...]

Issue #1811, opened 2022-08-02
Fix transform for `f1_retained_erng` xbrl + dbf


aesharpe:
## Updated tasklist
- [ ] Add a method to `process_xbrl_metadata()` that adds a completely new calculation to a field where one is missing
- [ ] Use this method to add calculations for `retained_earnings_ferc1` to the transform

## Table Notes
* The DBF table is almost entirely unstructured, with at least 6 freeform subsections, each of which has a subtotal -- the subtotal rows have well defined meanings, but only the row number and overall category is standardized for the rows making up the sections.
* XBRL has again broken out these unstructured tables into their own unstructured, sequenced tables containing a description + amount, with (presumably) the total from those tables matching the totals reported in the coordinating structured table. But we need to confirm that.
* The main table also has a start + end of year balance in the instant table `retained_earnings_118_instant`.
* There's also an unstructured instant table: `retained_earnings_appropriated_118_instant` with end of last year, end of this year dates.
* Good table for testing whether unstructured tables sum to reported totals, since everybody reports unstructured data.

## Transformation Plan
* Row numbers don't change over the years, but do need to be labeled for DBF + XBRL concatenation.
* Extreme case of unstructured data with only aggregate values showing up in the structured table.

zaneselvans:
Whoops, forgot there are a bunch of unstructured tables yet to go here.

e-belfer:
In issue #2016 and #2581 we ran into some errors with this table. Here are some compiled notes:

- Has `starting_balance` and `ending_balance` and also `amount`. Unclear looking at the form what `amount` corresponds to. Mostly NA start/end balances. Looking at form it seems like it might be unstructured data? Something is wrong here. The primary affected FERC account is getting lost, the freeform `item` description is also being lost. I think we need to revisit this transform.
- In 2020 and earlier data, they seem to have put some ending_balance values in the amount column.

e-belfer:
First-pass assessment of explosion-specific problems with this table:
- Fields sum to one another, but there are no calculations at all defined for this table
- Unclear distinction between starting_balance, ending_balance and amount for some line items
- Item and primary account only found in processed dbf data, not in processed xbrl.

PR #2645, opened 2023-06-08
Fix `retained_earnings_ferc1` transform

e-belfer:
Addresses #1811 and #2016. 

Issues addressed in this PR:
- **Drop `amount` column from `retained_earnings_ferc1`**: At present, there is an `amount`, `starting_balance` and `ending_balance` column. `amount` and `starting_balance` originate from the source data, while `ending_balance` is produced and filled with amount when reconciling duplicate years' data in the `condense_double_year_earnings_types_dbf` method. At present, this is only filled when the data has overlapping years. I've updated this method to fill all NA values for `ending_balance` with `amount` where available. This adds approximately another 6k values into the `ending_balance` column and resolves the ambiguity surrounding the relationship between the 3 columns.
- As discussed below, change `condense_double_year_earnings_types_dbf` into `reconcile_double_year_earnings_types_dbf`, to retain the `_previous_year` versions of existing factoids that require starting balances for their calculations. 
- Adds calculations into the `retained_earnings_ferc1` table based on the form itself, and checks them to be correct.
- Manually spot fixes the account numbers associated with fields.

Questions for the reviewer:
- I could probably fill in the `balance` column, but I don't have 100% certainty about what's a debit and what's a credit and some factoids could be either. Thoughts here?

--------------------------------------------------------------------------

**Notes about table weirdnesses:**
- Right now our tables either have `starting_balance` and `ending_balance` as columns or factoids. When starting balance values are included in calculations in other tables, they are included as factoids and summed in the `dollar_value` column. 
- Where tables have `ending_balance` as a value to be calculated (e.g., `utility_plant_summary_ferc1`), they do not directly calculate it using the `starting_balance`, even if starting balance is reported (e.g. `balance_sheet_assets_ferc1`).
- Post 2004, the `retained_earnings_ferc1` is a weirdo because it has `starting_balance` in 2 places: as a factoid for the table and as a column for each variable. We collapse these two values into one in the `condense_double_year_earnings_types_dbf` method but this means that we're losing the ability to add this value as a component in the calculations in the `ending_balance` column. In general, tables mirror the vertical addition structure used by FERC, and we're deviating from this format in this instance by reshaping the data.
- Our calculation checks right now are only set up to calculate values across one column (either `dollar_value` or `ending_balance`). The retained_earnings table has `starting_balance` and `ending_balance` columns for all variables, but there are calculations which require the addition of a starting balance and the ending balance of other components to get to the ending balance.

**Design decisions:**
To deal with the above, we could do one of the following:
- Add back `unappropriated_retained_earnings_start_balance` and `unappropriated_retained_earnings_ending_balance` (etc.) as factoids, reassigning the `ending_value` column to be a `dollar_value` column and dropping the `starting_balance` data. We'd lose half the data in this table this way.
- Add back `unappropriated_retained_earnings_start_balance` and `unappropriated_retained_earnings_ending_balance` (etc.) as factoids, keeping `starting_balance` and `ending_balance` and just filling in the values for this column as they presently are in the table (in effect, undo the work from `condense_double_year_earnings_types_dbf`). This table is frustratingly less clean than it was before, but we can do the calculation checks. Even so, we'll still have to deal with the same issue being present in the xbrl data, where there is only one factoid with a starting and ending balance, possibly by following a similar workaround to that used for the `electric_plant_depreciation_changes_ferc1` table.
  - We're going with this option, which most closely mirrors FERC's format.
- Reorganize the calculations to be able to take a value from a different column. Probably the neatest path forward but will require serious workarounds all the way down and strays from FERC's existing format.

Remaining tasks:
- [x] Change the behavior of `condense_double_year_earnings_type_dbf` to keep the filling behavior without dropping the previous year factoids
- [x] Change xbrl to include current and previous year factoids, mirroring dbf structure
- [x] Add these as components into calcs
- [x] Add table into `reconcile_table_calculations()` method
- [x] Debug any remaining issues with the calculations
- [x] Fix missing account numbers?

# PR Checklist


e-belfer:
Plan of action based on discussion with @zaneselvans 
- Change the behavior of `condense_double_year_earnings_type_dbf` to keep the filling behavior without dropping the previous year factoids
- Change xbrl to include current and previous year factoids, mirroring dbf structure
- Add these as components into calcs

e-belfer:
With the updates above, the table fields are able to be calculated and can be integrated into the explosion process. For the full ETL, the table as-is has 12.5% of calculations not summing correctly. This requires further diagnosis.

zaneselvans:
### Notes from messing around with the data:

I apologize if I sound like I'm being dense here... I'm just running through the FERC 1 ETL debugging notebook and finally actually getting to see what in the data. These are just my running notes from that process. Maybe we can sit down with the notebook and the custom transform methods briefly? 😵‍💫 

* What's up with there being two different kinds of `starting_balance` and `ending_balance` columns in the instant XBRL dataframe? They two sets of columns are the same but one has a `_reserve_federal` suffix.
* In the DBF output the `prmry_account` column looks like it contains FERC account numbers for ~75 rows (though it's usually blank) Is there any analogous information contained in the XBRL data? Is it okay that we're losing this information?  By the time the DBF and XBRL data have been combined in the output from `transform_start()` there are only 11 of these values left. Oh, I see these are the only ones which weren't associated with an `_unstructured` `earnings_type`, so I guess that the FERC Account numbers we get from the XBRL metadata take the place of this information for all the structured entries.
* There are 2 `earnings_type` values which only appear in the DBF data and not the XBRL data: `unappropriated_undistributed_subsidiary_earnings_previous_year` and `unappropriated_retained_earnings_previous_year`. Is that expected or concerning? It looks like they're the only `_previous_year` values.  I notice that they also never end up with an `xbrl_factoid_name_original` value. Is that okay? The same is true of the `calculations` field. Is this because they only exist in the DBF data?
* The `_unstructured` values only show up in the DBF data and get dumped when we combine XBRL & DBF because that information comes from separate tables in XBRL-land, and we're not dealing with them now, right?
* I notice that in the finished transformation we still have the `amount` column, and that the `starting_balance` and `ending_balance` values are always `NaN` when we have a non-null `amount`.  I recall that you're filling in the `ending_balance` based on `amount` later on somewhere... is there a reason why it doesn't make sense to do that here in the initial transformation and then drop the `amount` column?  Is there some conditionality that's not clear how to apply without more context from other tables? Ah. I see that we **are** dropping the `amount` column in `reconcile_double_year_earnings_types` but... then it's still around in the finished table.
* Reading through the reconciliation function, it sounds like we're trying to use the duplicated information available in the `_current_year` and `_previous_year` factoids to fill in missing values and check for internal consistency.  Once we've done that it **seems** like it would be desirable to drop the records that don't refer to the current year. What keeps us from doing that? If they represent duplicated information what do we need these values for later on?

e-belfer:
> What's up with there being two different kinds of starting_balance and ending_balance columns in the instant XBRL dataframe? They two sets of columns are the same but one has a _reserve_federal suffix.

I'm not sure what you mean about two sets of starting/ending balances. I see two components in the raw XBRL instant data, These are two different components that sum to `appropriated_retained_earnings_including_reserve_amortization` and are reflected in existing calculations. On my local they're getting correctly processed in the first XBRL processing stage.

> There are 2 earnings_type values which only appear in the DBF data and not the XBRL data: unappropriated_undistributed_subsidiary_earnings_previous_year and unappropriated_retained_earnings_previous_year. Is that expected or concerning? It looks like they're the only _previous_year values. I notice that they also never end up with an xbrl_factoid_name_original value. Is that okay? The same is true of the calculations field. Is this because they only exist in the DBF data?

DBF has factoids in `current_year` and `previous_year`, and XBRL only has the factoid itself. We were previously collapsing the DBF data into one factoid, but we need both for the calculations, so we are now splitting the XBRL data to match based on the timestamps in the original data. I've gone ahead and dropped the `current_year` suffix from the DBF data so we now have `unappropriated_undistributed_subsidiary_earnings` and `unappropriated_undistributed_subsidiary_earnings_previous_year`, e.g.

Re: the metadata - technically these values can be calculated using the same formula, but with one year's prior data. I think for simplicity though, we should relate to them as reported values. We are in practice checking that they are the same as the ending balance for the prior year's data. Lines 4457-59 attempt to add existing metadata associated with DBF `previous_year` factoids to the converted XBRL data, but this data was originally null. Perhaps we need to assign these values or troubleshoot here.

> The _unstructured values only show up in the DBF data and get dumped when we combine XBRL & DBF because that information comes from separate tables in XBRL-land, and we're not dealing with them now, right?

This table only deals with the structured data currently, correct.

> I notice that in the finished transformation we still have the amount column, and that the starting_balance and ending_balance values are always NaN when we have a non-null amount. I recall that you're filling in the ending_balance based on amount later on somewhere... is there a reason why it doesn't make sense to do that here in the initial transformation and then drop the amount column? Is there some conditionality that's not clear how to apply without more context from other tables? Ah. I see that we are dropping the amount column in reconcile_double_year_earnings_types but... then it's still around in the finished table.

I'm not sure we are looking at the same table. At the end of `reconcile_double_year_earnings_dbf()` the amount column gets dropped (line 4396). It also isn't in the table schema and should get dropped by `transform_end()`, if somehow it's still in the table. I don't see it in my `retained_earnings_ferc1()` table at the end of the full transform process.

> Reading through the reconciliation function, it sounds like we're trying to use the duplicated information available in the _current_year and _previous_year factoids to fill in missing values and check for internal consistency. Once we've done that it seems like it would be desirable to drop the records that don't refer to the current year. What keeps us from doing that? If they represent duplicated information what do we need these values for later on?

We can't drop the factoid because we need it as a `starting_balance` factoid in order to do the calculation, and we can't reconcile calculations across multiple columns. I've gone ahead and changed the transform that previously purged this value in order to retain it, while using the information in it to fill some potentially missing gaps. Duplicative but necessary.

zaneselvans (approved):

Issue #2599, opened 2023-05-24
Integrate minor cleaning fixes from spot-checks of XBRL explosion


e-belfer:
This issue compiles a few small fixes identified in discussions in #2016.

```[tasklist]
- [x] 1.) `utility_plant_summary_ferc1`: Where `depreciation_utility_plant_in_service` is negative and exactly opposite the sign of its calculation, flip it. See details here: https://github.com/catalyst-cooperative/pudl/issues/2016#issuecomment-1553191877
- [x] 2.) `plant_in_service_ferc1`: Fix negatives in the `electric_plant_sold` column. See details here: https://github.com/catalyst-cooperative/pudl/issues/2016#issuecomment-1559672614
- [x] 3.) `electric_operating_revenues_ferc1`: figure out why `small_or_commercial`and `large_or_industrial` values are getting dropped from the cleaned table @e-belfer
```

e-belfer:
With `electric_operating_revenues`, it looks like the column names are getting split by the `wide_to_tidy` function, with revenue attached to `small_or_commercial` and the other two fields attached to `small_or_commercial_sales`.
![image](https://github.com/catalyst-cooperative/pudl/assets/37471869/c78f1332-7eeb-4980-b614-c6dfefcd16e2)

I fixed this in the FERC transform parameters.

aesharpe:
@e-belfer and I decided that the first task makes sense to do this as a one-off spot fix so that we can run the `reconcile_table_calculations` functions after the records are fixed

aesharpe:
Values pertaining to Task 1 spot fix: 

|        |   report_year |   utility_id_ferc1 |   ending_balance |   calculated_amount |        abs_diff |
|-------:|--------------:|-------------------:|-----------------:|--------------------:|----------------:|
|  52661 |          2002 |                170 |  -8732           |      8732           | 17464           |
|  75784 |          2006 |                211 |      2.0784e+09  |        -2.0784e+09  |     4.1568e+09  |
|  75797 |          2006 |                211 |      2.0784e+09  |        -2.0784e+09  |     4.1568e+09  |
|  85062 |          2007 |                393 |      3.67392e+08 |        -3.67392e+08 |     7.34784e+08 |
|  85073 |          2007 |                393 |      3.67392e+08 |        -3.67392e+08 |     7.34784e+08 |
|  93853 |          2009 |                211 |      2.46878e+09 |        -2.46878e+09 |     4.93756e+09 |
[...]

PR #2647, opened 2023-06-08
Spot fix ferc exploder

aesharpe:
This PR addresses the spot fixes outlined in [#2599](https://github.com/catalyst-cooperative/pudl/issues/2599)

1.) Fix certain negative values in `utility_plant_summary_ferc1`: https://github.com/catalyst-cooperative/pudl/pull/2647#issuecomment-1583461155
2.) Fix negatives in the `plant_in_service_ferc1` table, `electric_plant_sold` column: https://github.com/catalyst-cooperative/pudl/pull/2647#issuecomment-1591754266


aesharpe:
# The Problem
There are some records in the `utility_plant_summary_ferc1` table that appear to contain obvious errors. 

Specifically, there are records where the value reported for `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility`, a calculated value, is a positive value and the records that comprise the subcomponents are all reported as negative values. This causes a discrepancy between _our_ calculated values (`calculated_amount`) and the _reported_ calculated values (`ending_balance`). In this case, the reported calculated values are positive, but our calculated values are negative. 

This is one example of 29 (see https://github.com/catalyst-cooperative/pudl/issues/2599#issuecomment-1579224702):

|       |   report_year |   utility_id_ferc1 | utility_type   | utility_plant_asset_type                                                           |   ending_balance |   calculated_amount |
|------:|--------------:|-------------------:|:---------------|:-----------------------------------------------------------------------------------|-----------------:|--------------------:|
| 93866 |          2009 |                211 | electric       | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility |      2.46878e+09 |        -2.46878e+09 |
| 93868 |          2009 |                211 | electric       | amortization_of_other_utility_plant_utility_plant_in_service                       |     -4.14332e+07 |       nan           |
| 93871 |          2009 |                211 | electric       | depreciation_utility_plant_in_service                                              |     -2.42735e+09 |       nan           |

In total, **53 records are spot fixed** (this is higher than 29 b/c we are spot fixing the components that add up to the calculated value, not the calculated value itself). This represents just 0.03% of all records!

# The Solution

This PR spot-fixes the subcomponents so that they are positive instead of negative. This causes our `calculated_amount` values to match the values in `ending_balance` for these calculated values.

I employ a spot fix rather than a programmatic fix here. This is because the way to programmatically identify errors is to use the outputs from the `reconcile_table_calculations` function. This function is also what we use to check that the reported calculations are correct. To fix the errors programmatically, I would have to run `reconcile_table_calculations`, identify the errors, fix the errors, and then run `reconcile_table_calculations` again to get rid of any adjustments made by running it the first time with the erroneous records (the function assumes that the programmatically calculated values are correct and will add an adjustment record to make the values add up).

By flipping the sign value of the subcomponents used in our calculation of the `calculated_amount`, we can run `reconcile_table_calculations` and get a calculation that matches the value in `ending_balance`.

|       |   report_year |   utility_id_ferc1 | utility_type   | utility_plant_asset_type                                                           |   ending_balance |   calculated_amount |
|------:|--------------:|-------------------:|:---------------|:-----------------------------------------------------------------------------------|-----------------:|--------------------:|
| 93866 |          2009 |                211 | electric       | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility |      2.46878e+09 |        2.46878e+09 |
| 93868 |          2009 |                211 | electric       | amortization_of_other_utility_plant_utility_plant_in_service                       |     4.14332e+07 |       nan           |
| 93871 |          2009 |                211 | electric       | depreciation_utility_plant_in_service                                              |     2.42735e+09 |       nan           |

# The Problem with the Problem
The 29 problem utility-years all have another field called `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail` (the same as the calculated field above but with the suffix **_detail**. This is a relic of the DBF form. In all 29 cases, the `detail` field is negative! (i.e., the opposite of the non-detail field). According to the form (below), these values should be the same.

(this is just 3 of the 29)
|    |   report_year |   utility_id_ferc1 | utility_type   | utility_plant_asset_type_y                                                                |   ending_balance_y |   calculated_amount_y |
|---:|--------------:|-------------------:|:---------------|:------------------------------------------------------------------------------------------|-------------------:|----------------------:|
|  0 |          2002 |                170 | other1         | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility        |     -8732          |         8732          |
|  1 |          2002 |                170 | other1         | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail |      8732          |          nan          |
|  6 |          2006 |                211 | total          | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility        |         2.0784e+09 |           -2.0784e+09 |
|  7 |          2006 |                211 | total          | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail |        -2.0784e+09 |          nan          |
| 20 |          2006 |                211 | electric       | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility        |         2.0784e+09 |           -2.0784e+09 |
| 21 |          2006 |                211 | electric       | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail |        -2.0784e+09 |          nan          |

We decided to spot fix these values because we felt that if the calculated value was reported as a positive value then the components were probably miss-reported as negative values. However, now there are _two_ reported calculated values (the latter isn't listed as a calculated value because it comes from DBF, pre-XBRL metadata including calculated values etc.), one of which is negative and one of which is positive...grr. **This erodes my confidence that these values should be spot fixed at all.** 

The reason why I've kept the spot fixes (for now) is because negative values for depreciation, amortization, and depletion (DAD) are almost always positive. It seems especially rare/bad for a utility report repeated negative values for DAD which is the case for `utility_id_ferc1` 211.

In this form: 
- line no. 14 = `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility`
- line no. 33 = `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail`

Screen Shot 2023-06-08 at 5 52 02 PM

jrea-rmi:
Spot fixes to change component values to positive sounds like the right move to me. Negative depreciation means the asset gains value over time, but I don't think that should happen for power plants - as they get used, they deteriorate. But there could be exceptions...

to determine which record carries more weight, we want this to match the value of UtilityPlantNet in `balance_sheet_assets_ferc1`. Checking the FERC taxonomy, I got turned around a couple times but it seemed like both the _detail and non-_detail sections were referenced...can you look into that also @aesharpe?

Another thought: does `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail` start getting reported to a separate table, as opposed to "stops getting reported in XBRL"? I wonder if that's true, then `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail` being different than `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility` would help identify sign errors in that new detail table? And it would also make sense to me that the non-detail table would include calculations that reference the detail table, because the bottom section of the form should be detail that adds up to values in the top section of the form.

aesharpe:
> does `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail` start getting reported to a separate table, as opposed to "stops getting reported in XBRL"?

It appears that `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail` gets subsumed into `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility`. The XBRL metadata for the `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility` column points to rows 14 and 33 in the Form which are the same rows highlighted above.

See `package_data/settings/ferc1_xbrl_taxonomy_metadata.json` and search for `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility`

aesharpe:
@jrea-rmi 

> to determine which record carries more weight, we want this to match the value of UtilityPlantNet in `balance_sheet_assets_ferc1`. Checking the FERC taxonomy, I got turned around a couple times but it seemed like both the _detail and non-_detail sections were referenced...can you look into that also @aesharpe?

Great idea. A preliminary check shows that the positive value (`accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility`) is correct! **This serves as validation for our spot fixing.**

According to the FERC taxonomy, `utility_plant_net` = `utility_plant_and_construction_work_in_progress` - `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility` (`_detail??`). As you can see below, the positive value 2.46878e+09 is the correct value for this equation! 

5.09278e+09 - 2.46878e+09 = 2.624e+09

|       |   report_year |   utility_id_ferc1 | utility_type   | utility_plant_asset_type                                                                  |   ending_balance |   calculated_amount |
|------:|--------------:|-------------------:|:---------------|:------------------------------------------------------------------------------------------|-----------------:|--------------------:|
| 93866 |          2009 |                211 | electric       | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility        |      2.46878e+09 |        -2.46878e+09 |
| 93867 |          2009 |                211 | electric       | accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail |     -2.46878e+09 |       nan           |
| 93872 |          2009 |                211 | electric       | utility_plant_and_construction_work_in_progress                                           |      5.09278e+09 |       nan           |
| 93878 |          2009 |                211 | electric       | utility_plant_net                                                                         |      2.624e+09   |         2.624e+09   |

aesharpe:
After all this investigation, I would feel pretty comfortable removing the `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_detail` records. What do @jrea-rmi, @cmgosnell think? I could do a little more investigation into the records that don't match up perfectly. I could also just ignore them and keep it all in there b/c it's such a small amount of problem records that differ.

aesharpe:
> What's the purpose of removing, and for what output?

Because it's technically a duplicate row with no new information that also isn't flagged as a total or calculated value which isn't great because we don't want folks to accidentally concatenate it. I would propose removing it from the `utility_plant_summary_ferc1` table pre-explosion so it's gone from the table if anyone wants to use just that table and also doesn't junk up the explosion.

jrea-rmi:
I figured it would have been flagged as a calculated value!

Would it still be useful for doing checks that ensure signs and values of components are correct? If not, then I'm in agreement with removing it pre-explosion.

aesharpe:
**Updates:** 

We did not get rid of the `_detail` row (for now). Instead, we changed the PUDL column mapping reference to point row 33 (see screen grab from https://github.com/catalyst-cooperative/pudl/pull/2647#issuecomment-1583461155) to `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility` (was previously the same but with the `_detail` suffix). Because this is the value that should be marked as `calculated_value` within the taxonomy.

We then turned the previous value for `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility` (row 14) into `accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_reported`.

aesharpe:
# 2.) Fix certain negatives in the `plant_in_service_ferc1` table, `electric_plant_sold` column

## The Problem

There are some calculation values within the `plant_in_service_ferc1` table for `ferc_account_label` == `plant_in_service` that are wrong. One reason that they are wrong is a negative sign that's incorrectly added to the `electric_plant_sold` values. This crops up because the calculation for `plant_in_service` subtracts the value for `electric_plant_sold` so it's often reported with a negative sign in front of it when it shouldn't be. This causes the calculations to add `electric_plant_sold` instead of subtract (think 10 - -5 = 15 instead of 10 - 5  = 5). Thus, the `calculated_field` will not match the value in `ending_balance`.

## The Solution

Create a spot fix by adding to the `transform_main` function for this table class: 
- Identify `plant_in_service` `calculated_amount` records that don't match their `ending_balance` values.
- Use the `report_year` and `utility_id_ferc1` from those records to identify `electric_plant_sold` records that are the subcomponents of the faulty calculations.
- Replace the `starting_balance` and `ending_balance` values for these records with their absolute value.

This flips about 78 records so that the calculation matches the ending balance!

e-belfer:
> I did not flip all negatives to positives because sometimes the negative value in electric_plant_sold leads to the correct calculation!

@aesharpe How often is this the case? Doing an across the board flip seems more justifiable to me and neater.

src/pudl/transform/ferc1.py
```diff
@@ -3042,8 +3042,49 @@ def transform_main(self, df: pd.DataFrame) -> pd.DataFrame:
...
+        # the original df and run the calculations later (in transform_end)
+        df_calc = df.pipe(self.reconcile_table_calculations).copy()
+        index_group_1 = ["report_year", "utility_id_ferc1"]
+        index_group_2 = ["report_year", "utility_id_ferc1", "ferc_account_label"]
+        # Get the index values for electric_plant_in_service fields that don't match
+        # their calculations
+        bad_calcs_pis_index = (
+            df_calc[
+                (df_calc["ferc_account_label"] == "electric_plant_in_service")
+                & (df_calc["ending_balance"] != df_calc["calculated_amount"])
```
e-belfer, line 3063: We've been using `~np.isclose(abs_diff, 0)` elsewhere, so I think this would be more consistent.

src/pudl/transform/ferc1.py
```diff
@@ -4003,6 +4044,189 @@ class UtilityPlantSummaryFerc1TableTransformer(Ferc1AbstractTableTransformer):
...
+
+        # We need to track which years are getting transformed so the spot fixes
+        # don't fail on the fast tests.
+        df_years = df.report_year.unique().tolist()
+
+        # The utility_id_ferc1 211 follows the same pattern for several years
+        # instead of writing them all out in spot_fix_pks, we'll create a loop that
+        # generates all of them and then append them to spot_fix_pks later
+        spot_fix_211 = []
+        for year in np.append(2006, range(2009, 2021)):
```
e-belfer, line 4066: Just confirming this should exclude 2021?

src/pudl/transform/ferc1.py
```diff
@@ -4003,6 +4044,189 @@ class UtilityPlantSummaryFerc1TableTransformer(Ferc1AbstractTableTransformer):
...
+            # Measure to account for fast tests where not all years are used
+            if year in df_years:
```
e-belfer, line 4068: Is there a reason not to do this using the spot_fixer(), rather than inside of `transform_main()`?

src/pudl/transform/ferc1.py
aesharpe: yes! No 2021. Thank you for double checking.

src/pudl/transform/ferc1.py
```diff
@@ -3042,8 +3042,49 @@ def transform_main(self, df: pd.DataFrame) -> pd.DataFrame:
...
+                bad_calc_groups_df["ferc_account_label"] == "electric_plant_sold"
+            ]
+            .reset_index()
+            .set_index(index_group_2)
+            .index
+        )
+        df = df.reset_index().set_index(index_group_2)
+        # Replace the electric_plant_sold ending_balance values identified above with
+        # their absolute value. Also enforce the schema so we can run the calcs again.
+        df.loc[bad_calc_eps_index, "ending_balance"] = abs(df["ending_balance"])
```
e-belfer, line 3084: Would be nice to have a logger statement here counting the # of values changes so we can tell if things go off the rails.

e-belfer (changes requested):
One minor change and a bigger question - is there a reason why you aren't using the spot fixer to make the changes for the `utility_plant_summary_ferc1` table? It's already unit tested and is set up to take PKs and values to change, so it seems like a good fit for this issue unless there is something I'm not seeing?

aesharpe:
> > I did not flip all negatives to positives because sometimes the negative value in electric_plant_sold leads to the correct calculation!
> 
> @aesharpe How often is this the case? Doing an across the board flip seems more justifiable to me and neater.

I just checked, and idk what I saw before, but you're right. Which is a bummer because that code was kind of labor intensive. But yay to things being simpler than I thought.

e-belfer (changes requested):
Oh no! I have one tiny suggestion on the wording of the logger statement, and a bigger question about the `utility_plant_summary` spot fix.

With this spot fix:
- `plant_in_service_ferc1` goes from 156 to 96 records with incorrect calculations. Nice!
-  `utility_plant_summary_ferc1` goes from 179 to 270 records with incorrect calculations. Uh oh! This spot fix should probably not get applied if it is in aggregate making things worse.

src/pudl/transform/ferc1.py
aesharpe: It makes more sense to keep them like this because there are over one hundred of these spot fixes and it would be very annoying to type all of those out in the current spot fixer (which doesn't accomadate any programatic creation of spot fixes)

e-belfer:
@aesharpe One last thought here other than the alembic rebase - can you add a note to the release notes explaining the `detail` column rename with reference to the PR?

e-belfer (approved):
Looking good!

PR #2648, opened 2023-06-08
Manage duplicate PKs in EIA-861 table transforms

### LINKED ISSUE #2638: Check for incorrect primary key columns in _tidy_class_dfs
The function `pudl.transform.eia861._tidy_class_dfs()` will silently drop records if given an incomplete list of primary key columns **and** the resulting reshaped dataframe **will** have unique values for the (incorrect) primary keys that were specified. As pointed out by @christiantfong in #2636 and fixed (for that one case) in #2637.

We should implement a check in that function which prevents this behavior, since it's used 20 times all across the EIA-861 transforms, and it would be easy for another incorrect set of PK columns to have slipped by us.

## Tables affected:
- EIA-861 Demand Response: 40/19120 (0.21%) records with duplicated PKs
- EIA-861 Sales: 402/417180 (0.10%) records with duplicated PKs

```[tasklist]
## Fixes
- [x] Verify that input dataframes have very limited (<0.5%) duplicate primary key values.
- [x] Consolidate reported data values across the small number of duplicate records.
- [x] Drop duplicates from the denormalized (non-data) columns as well.
- [x] Explicitly validate the merge which brings together the denormalized columns and reshaped columns.
```

zaneselvans:
# PR Overview

The `sales_eia861` and `demand_response_eia861` tables each have a handful of duplicate primary keys due to NA values in the `balancing_authority_code_eia` column. Quantify and log the extent of this problem, and consolidate the data in the duplicated records if they constitute less than 0.5% of all records in the table.

This check would also have caught the incorrect primary key columns reported in #2636 and fixed in #2637.

Because there were so few duplicate records, I decided to just consolidate them all (with a hard limit on the fraction of records that could be consolidated) rather that requiring that the only duplication be due to the BA Code column.

Closes #2638

# PR Checklist


src/pudl/transform/eia861.py
```diff
@@ -723,13 +717,39 @@ def _tidy_class_dfs(
     data_cols = data_cols.stack(level=0, dropna=False).reset_index()
     denorm_cols = _filter_non_class_cols(raw_df, class_list).reset_index()
 
+    # Check to make sure that the idx_cols are actually valid primary key cols:
+    # This is tricky, because NA values in the BA Code column creates actual duplicate
+    # values. These NA values are filled with UNK and the duplicates are later dropped,
+    # which means we are losing data. But it's not obvious how we could actually fill
+    # in real BA codes, so we consolidate the indistinguisbale records.
```
e-belfer, line 724: indistinguishable?

src/pudl/transform/eia861.py
```diff
@@ -723,13 +717,39 @@ def _tidy_class_dfs(
...
+    # These are only a fraction of a percent of all records, and only affect a couple
+    # of tables. In sales_eia861 there are 67 duplicate records.
```
e-belfer, line 726: We can probably drop the reference to this specific issue in the docstring?

src/pudl/transform/eia861.py
```diff
@@ -723,13 +717,39 @@ def _tidy_class_dfs(
...
+    # See: https://github.com/catalyst-cooperative/pudl/issues/2638
+    dupe_mask = data_cols.duplicated(subset=idx_cols + [class_type], keep=False)
+    dupes = data_cols[dupe_mask]
+    fraction_dupes = len(dupes) / len(data_cols)
+    err_msg = (
+        f"{df_name} table: Found {len(dupes)}/{len(data_cols)} "
+        f"({fraction_dupes:0.2%}) records with duplicated PKs. "
+    )
+    if fraction_dupes <= 0.005:
```
e-belfer, line 735: Do you want this to be a configurable percentage that we call in through `context`? Either way seems ok to me, we just often do so rather than hardcoding this value. The value itself seems reasonable though!

src/pudl/transform/eia861.py
```diff
@@ -723,13 +717,39 @@ def _tidy_class_dfs(
...
+        err_msg += "Consolidating duplicated records."
+        logger.info(err_msg)
+        deduped = dupes.groupby(idx_cols + [class_type], as_index=False).sum()
+        data_cols = pd.concat([data_cols[~dupe_mask], deduped], axis="index")
+        # PK dupes also affect denorm_cols. Drop dupes to ensure clean 1 to many merge:
+        denorm_cols = denorm_cols.drop_duplicates(subset=idx_cols)
```
e-belfer, line 741: I would similarly want to have some kind of check to make sure we aren't dropping huge amounts of data here.

e-belfer (approved):
Looks good, some minor non-blocking suggestions below. Ran the fast ETL with the debugging log statements and everything printed as expected and ran without error.

src/pudl/transform/eia861.py
zaneselvans: Hmm, I don't think I've really used the dagster context this way before. Is there an example you're thinking of?  I know magic numbers aren't great. I just wanted something that would alert us if the current very small number of duplicates were to change.

e-belfer: I'm thinking of `thresh` in `denorm_fuel_by_plant_ferc1` but I think in this case either way is alright and this seems like a reasonable value to hard-code in.

src/pudl/transform/eia861.py
zaneselvans: Okay, also checking the non-data columns now...

PR #2651, opened 2023-06-09
Use PUDL_INPUT not hard-coded data dir in datastore CLI

zaneselvans:
# PR Overview

While debugging FERC 2 DBF extraction build failure I found a lingering hard-coded PUDL input directory in the datastore CLI and fixed it.

# PR Checklist

### DIFF
--- a/src/pudl/workspace/datastore.py
+++ b/src/pudl/workspace/datastore.py
@@ -534,7 +534,7 @@ def _get_pudl_in(args: dict) -> Path:
     if args.pudl_in:
         return Path(args.pudl_in)
     else:
-        return Path(pudl.workspace.setup.get_defaults()["pudl_in"])
+        return Path(pudl.workspace.setup.get_defaults()["PUDL_INPUT"])
 
 
 def _create_datastore(args: argparse.Namespace) -> Datastore:
@@ -542,7 +542,7 @@ def _create_datastore(args: argparse.Namespace) -> Datastore:
     # Configure how we want to obtain raw input data:
     ds_kwargs = dict(gcs_cache_path=args.gcs_cache_path, sandbox=args.sandbox)
     if not args.bypass_local_cache:
-        ds_kwargs["local_cache_path"] = _get_pudl_in(args) / "data"
+        ds_kwargs["local_cache_path"] = _get_pudl_in(args)
     return Datastore(**ds_kwargs)
 
 


src/pudl/workspace/datastore.py
```diff
@@ -534,15 +534,15 @@ def _get_pudl_in(args: dict) -> Path:
     if args.pudl_in:
         return Path(args.pudl_in)
     else:
-        return Path(pudl.workspace.setup.get_defaults()["pudl_in"])
+        return Path(pudl.workspace.setup.get_defaults()["PUDL_INPUT"])
```
rousik, line 537: The fact that `pudl_in` and `PUDL_INPUT` both exist in the settings dict and have different semantics seems like a dangerous thing. Additionally, it seems that PUDL_INPUT is set into `data_dir`. Should we then use `data_dir` here instead of `PUDL_INPUIT`?

src/pudl/workspace/datastore.py
```diff
@@ -534,15 +534,15 @@ def _get_pudl_in(args: dict) -> Path:
     if args.pudl_in:
         return Path(args.pudl_in)
```
rousik, line 535: If we could drop support for `--pudl_in`, we could purely rely on `get_defaults()` and drop this function altogether. I think that could make sense.

src/pudl/workspace/datastore.py
```diff
@@ -534,15 +534,15 @@ def _get_pudl_in(args: dict) -> Path:
...
+        return Path(pudl.workspace.setup.get_defaults()["PUDL_INPUT"])
 
 
 def _create_datastore(args: argparse.Namespace) -> Datastore:
     """Constructs datastore instance."""
     # Configure how we want to obtain raw input data:
     ds_kwargs = dict(gcs_cache_path=args.gcs_cache_path, sandbox=args.sandbox)
     if not args.bypass_local_cache:
-        ds_kwargs["local_cache_path"] = _get_pudl_in(args) / "data"
+        ds_kwargs["local_cache_path"] = _get_pudl_in(args)
```
rousik, line 545: It would be nice if we could simply replace this with `get_defaults()["data_dir"]`

rousik (approved):
Approving for tactical fix, but I think there's a deeper issue that might be worth cleaning up (e.g. `pudl_in`, `PUDL_INPUT` and `data_dir` all being valid options, causing confusion)

src/pudl/workspace/datastore.py
zaneselvans: IIRC the use-case for setting a different input directory is/was being able to test the downloads of archives from Zenodo independent of the user's working cache directory.

As you've already noticed... there's a ton of room for simplification in the settings system.

src/pudl/workspace/datastore.py
zaneselvans: It's definitely not ideal!  I went with `PUDL_INPUT` rather than `data_dir` because I think we're eventually just going to depend on a real settings class & have it take cues from the environment variables.

There's a ton of vestigial mess in this settings system and I would ❤️❤️❤️ to blow it all away and replace it with a rational Pydantic settings class, but that'll mean chasing down and removing all the `pudl_settings` dictionaries that are floating around right now which will be a bit of work (but totally worth it in the medium term...). I'm inclined toward that option rather than trying to rationalize the current mess in-place. But we haven't been able to prioritize it. 😭

Issue #2623, opened 2023-06-02
find/add inter-table relationships when they include a second dimension


cmgosnell:
Ex: in the `income_statement_ferc1` table, there is an `operating_revenues` factoid. In the `electric_operating_revenues_ferc1` there is a `electric_operating_revenues` factoid. They are conceptually related but not the same and do not have the same factoid name. The `income_statement_ferc1` table & thus the `operating_revenues` has another dimension of the `utility_type`. So there is a line of `operating_revenues` in the income statement table which is `electric`.

This one is  really easy to see bc it is `{other_dimension}_{xbrl_factoid_name}` but i expect it won't always be that straightforward.

If we add in the implied dimension in the tables where these dimensions are implied (which would probably be helpful anyway) and renamed the factoids to match we could convert these conceptual connections into more tangible ones that can be id-ed programmatically, checked and scrubbed for the 💥  

```[tasklist]
- [x] HUNT DOWN all of these inter-table connections that don't link up by name (@jrea-rmi  if you have any pointers on this that would be vv helpful)
- [x] when required, add the other dimension into the implicitly-dimension-ed table (i.e. `assign(utility_type="electric")` to `electric_operating_revenues_ferc1`.
- [ ] possibly add check into the explosion (e.g. if dimension and factoid name the same, remove duplicate)
```

e-belfer:
Will continue to update as I go through the tables

For the electric-specific columns of `income_statement_ferc1` table:
- `operating_revenues` comes from Schedule 400 -> `electric_operating_revenues` (`electric_operating_revenues_ferc1`)
- `operating_expense` comes from Schedule 401 -> `operation_supervision` (`electric_operating_expenses_ferc1`)
- `maintenance_expense` comes from Schedule 402 -> `maintenance_supervision_and_engineering` (`electric_operating_expenses_ferc1`)
- `depreciation_expense` comes from Schedule 403 -> `depreciation_expense` in `depreciation_amortization_summary_ferc1` where `plant_function == common_plant_electric`
- `depreciation_expense_for_asset_retirement_costs` comes from Schedule 403.1 -> `depreciation_expense_asset_retirement` in `depreciation_amortization_summary_ferc1` (where `plant_function == common_plant_electric` only??)

- `amortization_and_depletion_of_utility_plant` comes from Schedule 404-5 -> ??
- `amortization_of_electric_plant_acquisition_adjustments` comes from Schedule 406 -> ??
[...]

PR #2653, opened 2023-06-10
Organize FERC 1 XBRL metadata with a calculation forest

zaneselvans:
# PR Overview

* Add correction records to the processed XBRL metadata dataframe, so that it matches what's found in the calculations and the post-corrected dataframes.
* Introduces a tree structure for management of the relationships between nested calculations across different FERC tables.
* The `XbrlCalculationTreeFerc1` and `XbrlCalculationForestFerc1` can be used to modify metadata and propagate it to child calculations so it can be applied to all the XBRL facts that make up the leaves of the tree.

# PR Checklist


zaneselvans:
The addition of metadata for the calculation correction records seems to have broken something in the table transforms:

```
dagster._core.errors.DagsterExecutionStepExecutionError: Error occurred while executing op "electric_operating_revenues_ferc1":

  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/execute_plan.py", line 273, in dagster_event_sequence_for_step
    for step_event in check.generator(step_events):
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/execute_step.py", line 369, in core_dagster_event_sequence_for_step
    for user_event in check.generator(
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/execute_step.py", line 90, in _step_output_error_checked_user_event_sequence
    for user_event in user_event_sequence:
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/compute.py", line 192, in execute_core_compute
    for step_output in _yield_compute_results(step_context, inputs, compute_fn):
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/compute.py", line 161, in _yield_compute_results
    for event in iterate_with_context(
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_utils/__init__.py", line 443, in iterate_with_context
    with context_fn():
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/contextlib.py", line 155, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/utils.py", line 84, in op_execution_error_boundary
    raise error_cls(

KeyError: "None of [Int64Index([-2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -1, -1, -1, -1, -1, -1, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2], dtype='int64')] are in the [columns]"

  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/utils.py", line 54, in op_execution_error_boundary
    yield
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_utils/__init__.py", line 445, in iterate_with_context
    next_output = next(iterator)
                  ^^^^^^^^^^^^^^
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/compute_generator.py", line 124, in _coerce_op_compute_fn_to_iterator
    result = invoke_compute_fn(
             ^^^^^^^^^^^^^^^^^^
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/compute_generator.py", line 118, in invoke_compute_fn
                                                                    ^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/transform/ferc1.py", line 4855, in ferc1_transform_asset
    df = transformer.transform(
         ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/transform/classes.py", line 1190, in transform
    .pipe(self.transform_end)
     ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/generic.py", line 5839, in pipe
    return com.pipe(self, func, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/common.py", line 513, in pipe
    return func(obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/transform/classes.py", line 1022, in _wrapper
    df = func(self, *args, **kwargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/transform/ferc1.py", line 1232, in transform_end
    df = self.reconcile_table_calculations(df).pipe(self.enforce_schema)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/transform/ferc1.py", line 2511, in reconcile_table_calculations
    df = reconcile_table_calculations(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/transform/ferc1.py", line 776, in reconcile_table_calculations
    tbl_meta[~tbl_meta.intra_table_calc_flag].xbrl_factoid.unique()
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/frame.py", line 3813, in __getitem__
    indexer = self.columns._get_indexer_strict(key, "columns")[1]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/indexes/base.py", line 6070, in _get_indexer_strict
    self._raise_if_missing(keyarr, indexer, axis_name)
  File "/Users/zane/mambaforge/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/indexes/base.py", line 6130, in _raise_if_missing
    raise KeyError(f"None of [{key}] are in the [{axis_name}]")
```

zaneselvans:
The above error was due to the fact that when creating the metadata records for the corrections, we were producing a dataframe with a mix of `bool` and `boolean` types -- which looks like it's fine, but is actually an `object` column.  It can be used to select records in a dataframe just fine as it is, but if you negate it with the `~` operator, it **doesn't** behave like a boolean column.  Instead the objects get converted into random integers, which act like an index not a boolean mask.  Ugh.

zaneselvans:
Okay, an example of a closer to functional calculation tree.

```py
import json
from pudl.output.ferc1 import MetadataExploder, resolve_calculation_components

balance_sheet_asset_tables = [
    "balance_sheet_assets_ferc1",
    "plant_in_service_ferc1",
]

exploded_meta = (
    MetadataExploder(balance_sheet_asset_tables).boom(xbrl_meta)
)[[
    "table_name",
    "xbrl_factoid",
    "calculations",
    "intra_table_calc_flag",
    "xbrl_factoid_original",
]]

calculation_tree = resolve_calculation_components(
    exploded_meta=exploded_meta.set_index(["table_name", "xbrl_factoid"]),
    xbrl_factoid="utility_plant_net",
)

print(json.dumps(calculation_tree.dict(), indent=4))
```

```json
{
    "components": [
        {
            "name": "utility_plant_and_construction_work_in_progress",
            "weight": 1.0,
            "source_tables": [
                "balance_sheet_assets_ferc1"
            ],
            "calculation": {
                "components": [
                    {
                        "name": "utility_plant",
                        "weight": 1.0,
                        "source_tables": [
                            "balance_sheet_assets_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "balance_sheet_assets_ferc1",
                            "xbrl_factoid": "utility_plant",
                            "xbrl_factoid_original": "utility_plant"
                        }
                    },
                    {
                        "name": "utility_plant_in_service_classified_and_unclassified",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [
                                {
                                    "name": "utility_plant_in_service_classified",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "utility_plant_in_service_classified",
                                        "xbrl_factoid_original": "utility_plant_in_service_classified"
                                    }
                                },
                                {
                                    "name": "utility_plant_in_service_property_under_capital_leases",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "utility_plant_in_service_property_under_capital_leases",
                                        "xbrl_factoid_original": "utility_plant_in_service_property_under_capital_leases"
                                    }
                                },
                                {
                                    "name": "utility_plant_in_service_plant_purchased_or_sold",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "utility_plant_in_service_plant_purchased_or_sold",
                                        "xbrl_factoid_original": "utility_plant_in_service_plant_purchased_or_sold"
                                    }
                                },
                                {
                                    "name": "utility_plant_in_service_completed_construction_not_classified",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "utility_plant_in_service_completed_construction_not_classified",
                                        "xbrl_factoid_original": "utility_plant_in_service_completed_construction_not_classified"
                                    }
                                },
                                {
                                    "name": "utility_plant_in_service_experimental_plant_unclassified",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "utility_plant_in_service_experimental_plant_unclassified",
                                        "xbrl_factoid_original": "utility_plant_in_service_experimental_plant_unclassified"
                                    }
                                },
                                {
                                    "name": "utility_plant_in_service_classified_and_unclassified_correction",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "utility_plant_in_service_classified_and_unclassified_correction",
                                        "xbrl_factoid_original": "utility_plant_in_service_classified_and_unclassified"
                                    }
                                }
                            ],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "utility_plant_in_service_classified_and_unclassified",
                            "xbrl_factoid_original": "utility_plant_in_service_classified_and_unclassified"
                        }
                    },
                    {
                        "name": "utility_plant_leased_to_others",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "utility_plant_leased_to_others",
                            "xbrl_factoid_original": "utility_plant_leased_to_others"
                        }
                    },
                    {
                        "name": "utility_plant_held_for_future_use",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "utility_plant_held_for_future_use",
                            "xbrl_factoid_original": "utility_plant_held_for_future_use"
                        }
                    },
                    {
                        "name": "construction_work_in_progress",
                        "weight": 1.0,
                        "source_tables": [
                            "balance_sheet_assets_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "balance_sheet_assets_ferc1",
                            "xbrl_factoid": "construction_work_in_progress",
                            "xbrl_factoid_original": "construction_work_in_progress"
                        }
                    },
                    {
                        "name": "utility_plant_acquisition_adjustment",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "utility_plant_acquisition_adjustment",
                            "xbrl_factoid_original": "utility_plant_acquisition_adjustment"
                        }
                    },
                    {
                        "name": "utility_plant_and_construction_work_in_progress_correction",
                        "weight": 1.0,
                        "source_tables": [
                            "balance_sheet_assets_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "balance_sheet_assets_ferc1",
                            "xbrl_factoid": "utility_plant_and_construction_work_in_progress_correction",
                            "xbrl_factoid_original": "utility_plant_and_construction_work_in_progress"
                        }
                    }
                ],
                "source_table": "balance_sheet_assets_ferc1",
                "xbrl_factoid": "utility_plant_and_construction_work_in_progress",
                "xbrl_factoid_original": "utility_plant_and_construction_work_in_progress"
            }
        },
        {
            "name": "accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility",
            "weight": -1.0,
            "source_tables": [
                "balance_sheet_assets_ferc1"
            ],
            "calculation": {
                "components": [
                    {
                        "name": "depreciation_utility_plant_in_service",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "depreciation_utility_plant_in_service",
                            "xbrl_factoid_original": "depreciation_utility_plant_in_service"
                        }
                    },
                    {
                        "name": "amortization_and_depletion_of_producing_natural_gas_land_and_land_rightsutility_plant_in_service",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "amortization_and_depletion_of_producing_natural_gas_land_and_land_rightsutility_plant_in_service",
                            "xbrl_factoid_original": "amortization_and_depletion_of_producing_natural_gas_land_and_land_rightsutility_plant_in_service"
                        }
                    },
                    {
                        "name": "amortization_of_underground_storage_land_and_land_rightsutility_plant_in_service",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "amortization_of_underground_storage_land_and_land_rightsutility_plant_in_service",
                            "xbrl_factoid_original": "amortization_of_underground_storage_land_and_land_rightsutility_plant_in_service"
                        }
                    },
                    {
                        "name": "amortization_of_other_utility_plant_utility_plant_in_service",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "amortization_of_other_utility_plant_utility_plant_in_service",
                            "xbrl_factoid_original": "amortization_of_other_utility_plant_utility_plant_in_service"
                        }
                    },
                    {
                        "name": "depreciation_amortization_and_depletion_utility_plant_in_service",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "depreciation_amortization_and_depletion_utility_plant_in_service",
                            "xbrl_factoid_original": "depreciation_amortization_and_depletion_utility_plant_in_service"
                        }
                    },
                    {
                        "name": "depreciation_amortization_and_depletion_utility_plant_leased_to_others",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [
                                {
                                    "name": "depreciation_utility_plant_leased_to_others",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "depreciation_utility_plant_leased_to_others",
                                        "xbrl_factoid_original": "depreciation_utility_plant_leased_to_others"
                                    }
                                },
                                {
                                    "name": "amortization_and_depletion_utility_plant_leased_to_others",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "amortization_and_depletion_utility_plant_leased_to_others",
                                        "xbrl_factoid_original": "amortization_and_depletion_utility_plant_leased_to_others"
                                    }
                                },
                                {
                                    "name": "depreciation_amortization_and_depletion_utility_plant_leased_to_others_correction",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "depreciation_amortization_and_depletion_utility_plant_leased_to_others_correction",
                                        "xbrl_factoid_original": "depreciation_amortization_and_depletion_utility_plant_leased_to_others"
                                    }
                                }
                            ],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "depreciation_amortization_and_depletion_utility_plant_leased_to_others",
                            "xbrl_factoid_original": "depreciation_amortization_and_depletion_utility_plant_leased_to_others"
                        }
                    },
                    {
                        "name": "depreciation_and_amortization_utility_plant_held_for_future_use",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [
                                {
                                    "name": "depreciation_utility_plant_held_for_future_use",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "depreciation_utility_plant_held_for_future_use",
                                        "xbrl_factoid_original": "depreciation_utility_plant_held_for_future_use"
                                    }
                                },
                                {
                                    "name": "amortization_utility_plant_held_for_future_use",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "amortization_utility_plant_held_for_future_use",
                                        "xbrl_factoid_original": "amortization_utility_plant_held_for_future_use"
                                    }
                                },
                                {
                                    "name": "depreciation_and_amortization_utility_plant_held_for_future_use_correction",
                                    "weight": 1.0,
                                    "source_tables": [
                                        "utility_plant_summary_ferc1"
                                    ],
                                    "calculation": {
                                        "components": [],
                                        "source_table": "utility_plant_summary_ferc1",
                                        "xbrl_factoid": "depreciation_and_amortization_utility_plant_held_for_future_use_correction",
                                        "xbrl_factoid_original": "depreciation_and_amortization_utility_plant_held_for_future_use"
                                    }
                                }
                            ],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "depreciation_and_amortization_utility_plant_held_for_future_use",
                            "xbrl_factoid_original": "depreciation_and_amortization_utility_plant_held_for_future_use"
                        }
                    },
                    {
                        "name": "abandonment_of_leases",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "abandonment_of_leases",
                            "xbrl_factoid_original": "abandonment_of_leases"
                        }
                    },
                    {
                        "name": "amortization_of_plant_acquisition_adjustment",
                        "weight": 1.0,
                        "source_tables": [
                            "utility_plant_summary_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "utility_plant_summary_ferc1",
                            "xbrl_factoid": "amortization_of_plant_acquisition_adjustment",
                            "xbrl_factoid_original": "amortization_of_plant_acquisition_adjustment"
                        }
                    },
                    {
                        "name": "accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_correction",
                        "weight": 1.0,
                        "source_tables": [
                            "balance_sheet_assets_ferc1"
                        ],
                        "calculation": {
                            "components": [],
                            "source_table": "balance_sheet_assets_ferc1",
                            "xbrl_factoid": "accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility_correction",
                            "xbrl_factoid_original": "accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility"
                        }
                    }
                ],
                "source_table": "balance_sheet_assets_ferc1",
                "xbrl_factoid": "accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility",
                "xbrl_factoid_original": "accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility"
            }
        },
        {
            "name": "utility_plant_net_correction",
            "weight": 1.0,
            "source_tables": [
                "balance_sheet_assets_ferc1"
            ],
            "calculation": {
                "components": [],
                "source_table": "balance_sheet_assets_ferc1",
                "xbrl_factoid": "utility_plant_net_correction",
                "xbrl_factoid_original": "utility_plant_net"
            }
        }
    ],
    "source_table": "balance_sheet_assets_ferc1",
    "xbrl_factoid": "utility_plant_net",
    "xbrl_factoid_original": "utility_plant_net"
}
```

e-belfer:
I _think_ this actually has to do with issue #2623 rather than with anything incorrect in the metadata. 

The `utility_plant_summary` table:
image

The `plant_in_service` table:
image

Where there is an electric utility type, these values should come from the `plant_in_service` table. These definitions don't seem to be explicitly defined in the metadata, however, since the values only originate out of table if they refer to an electric utility specifically.

src/pudl/output/ferc1.py
```diff
@@ -1337,3 +1339,93 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+        if source_table is not None:
+            all_tables = all_tables.union({source_table})
+        return True if len(all_tables) > 1 else False
+
+
+Ferc1XbrlCalculationComponent.update_forward_refs()
+Ferc1XbrlCalculation.update_forward_refs()
+
+
+def resolve_calculation_components(
```
jdangerx, line 1395: I'm having a little bit of trouble figuring out what the expectations of this function are, mostly because I don't know much about what the exploded metadata should look like, or what we expect these calculation trees to look like!

I started writing some tests to try to nail down some of that behavior, they look like this so far - it might make sense for someone who knows more about the expected behavior to actually flesh these out.

From my perspective having these tests will give a lot of confidence that this calculation tree will have the shape we expect, but if you feel like the tests would be redundant I'm happy to be convinced to have less code :)

```python
import json
import pandas as pd

from pudl.output.ferc1 import resolve_calculation_components


def test_resolve_calculation_components_empty():
    # maybe there's a better way to get an exploded test metadata than just hand-coding
    # this structure...
    test_table = "test_table"
    test_factoid = "test_factoid"
    calculations = json.dumps([])
    exploded_meta = pd.DataFrame({
        "table_name": [test_table],
        "xbrl_factoid": [test_factoid],
        "xbrl_factoid_original": [f"{test_factoid}_original"],
        "calculations": [calculations],
    })
    exploded_meta = exploded_meta.set_index(["table_name", "xbrl_factoid"])

    observed = resolve_calculation_components(exploded_meta, test_table, test_factoid)

    assert observed.source_table == test_table
    assert observed.xbrl_factoid == test_factoid
    assert observed.xbrl_factoid_original == f"{test_factoid}_original"
    assert observed.components == []


def test_resolve_calculation_components_single():
    pass


def test_resolve_calculation_components_one_level():
    pass


def test_resolve_calculation_components_multi_level():
    pass


def test_resolve_calculation_components_multi_level_non_unique_source_tables():
    pass
```

src/pudl/output/ferc1.py
```diff
@@ -1337,3 +1339,93 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+    components = [Ferc1XbrlCalculationComponent(**c) for c in components]
+    new_components = []
+    for c in components:
+        new_comp = Ferc1XbrlCalculationComponent(
+            name=c.name,
+            weight=c.weight,
+            source_tables=c.source_tables,
+            calculation=resolve_calculation_components(
+                exploded_meta=exploded_meta,
+                table_name=c.source_tables[0],
```
jdangerx, line 1420: If the JSON object representing the calculation has multiple source tables, is it safe to just ignore the tail of the array?

src/pudl/output/ferc1.py
```diff
@@ -1337,3 +1339,93 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+Ferc1XbrlCalculationComponent.update_forward_refs()
+Ferc1XbrlCalculation.update_forward_refs()
+
+
+def resolve_calculation_components(
+    exploded_meta: pd.DataFrame,
+    table_name: str,
+    xbrl_factoid: str,
+) -> Ferc1XbrlCalculation:
+    """Build a calculation given exploded metadata, xbrl_factoid, and table_name."""
```
jdangerx, line 1400: Broadly, this recursive function looks fine, if I'm reading it right:

* base case is "If the exploded_meta for this table_name/xbrl_factoid has an empty list for "calculations", just return a `Ferc1XbrlCalculationComponent` with `components=[]`"
* otherwise, call this on whatever components this factoid points at

Though I suppose this assumes that walking the tree will eventually get you to nodes that have empty calculations.

I don't suppose there's a risk of hitting the recursion limit here before we get to the bottom of the tree?

jdangerx (changes requested):
The recursive structure looks fine to me, but I think getting some behavior under test would be very useful. Maybe that's my own lack of context 🤷

src/pudl/output/ferc1.py
zaneselvans: Yeah... it definitely needs some tests.

The high level idea is that we want to expand all of the individual calculation components which are actually references to other values and expose the dependencies between them so that we can see how they nest within each other, and apply metadata from one node to all of its children. E.g. if the `utility_plant_net` fact has been tagged as being "in rate base" then all of its child components should also get tagged with "in rate base" -- then this information can be merged onto the exploded output table, without needing to manually compile and update a huge table that applies every possible tag to every possible fact within one of these big explosions (that involve hundreds of facts).

src/pudl/output/ferc1.py
zaneselvans: The only reason this is okay is because we're requiring `source_tables` to be of length 1 elsewhere in the process.

src/pudl/output/ferc1.py
zaneselvans: I don't think there's any very deep nesting -- maybe like 5-6 layers max. I think the tree I created in the comments on this issue is probably one of the biggest ones there is. And it's definitely our expectation that **eventually** all the components are reported values.  If we end up with cycles then either there's a problem with the specified calculations, or we've fundamentally misunderstood something.

There's another kind of termination condition that I don't think we've really conveyed here though, which is what happens if there's a reference to a table that is NOT in the set of tables that make up the explosion? In that case we also don't want continue recursing, and instead just refer to the reported value.  I think there was another edge-of-the-calculation-space case that I'm not remembering now though...

zaneselvans:
@e-belfer Thank you for checking on the `plant_in_service` references!  I think I confused myself by doing:

```py
exploded_meta[exploded_meta.calculations.str.contains("plant_in_service")]
```
which of course DOES find a bunch of metadata records but... that's just because other facts contain that substring. 

If I look more specifically for just the `"plant_in_service"` string as a standalone reference to a source table:
```py
exploded_meta[exploded_meta.calculations.str.contains('"plant_in_service"')]
```

I find no matching metadata records.

src/pudl/output/ferc1.py
```diff
@@ -1337,3 +1339,93 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
         )
     ]
     return exploded
+
+
+################################################################################
+# XBRL Calculation Tree
+################################################################################
+Ferc1XbrlCalculationComponent = ForwardRef("Ferc1XbrlCalculationComponent")
```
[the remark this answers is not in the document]
zaneselvans: If I comment out the `ForwardRef` stuff, then I get this error in the definition of `Ferc1XbrlCalculationComponent`:
```
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

I got the suggestion to use `ForwardRef` from the [Pydantic docs](https://docs.pydantic.dev/latest/usage/postponed_annotations/). Is there a better way to resolve this?

zaneselvans:
Another version with a single `Ferc1XbrlCalculationNode` class for building the tree.  It mostly works, but I'm messing up the termination of the recursion somehow.  Every node that should be a leaf node ends up having another single node hanging off of it, that is a copy of itself but with no children.  For example:

```json
            {
              "xbrl_factoid": "utility_plant",
              "source_table": "balance_sheet_assets_ferc1",
              "xbrl_factoid_original": "utility_plant",
              "weight": 1.0,
              "children": [
                {
                  "xbrl_factoid": "utility_plant",
                  "source_table": "balance_sheet_assets_ferc1",
                  "xbrl_factoid_original": "utility_plant",
                  "weight": 1.0,
                  "children": []
                }
              ]
            }
```

test/unit/output/ferc1_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+TEST_CALC_2 = [
+    {"name": "reported_1", "weight": 1.0, "source_tables": ["table_1", "table_2"]},
+    {"name": "reported_2", "weight": -1.0, "source_tables": ["table_1"]},
+]
+
+TEST_CALC_3 = [
+    {"name": "reported_1", "weight": 1.0, "source_tables": ["table_1"]},
+    {"name": "reported_3", "weight": 1.0, "source_tables": ["table_3"]},
+]
```
zaneselvans, line 36: @cmgosnell @e-belfer 

I'm not sure how we want to handle this case, where the calculation refers to a table that is not in the set of tables being "exploded." This means the calculation subcomponents can't be looked up in the `exploded_meta` dataframe, which currently results in a `KeyError` while trying to build the tree.

Maybe we should catch that `KeyError` and verify that it's failing because the table is missing (rather than e.g. a typo in the `xbrl_factoid`) and return an empty calculation? Will that result in just looking up the "reported" calculated value rather than attempting to calculate the value that's out-of-bounds?

test/unit/output/ferc1_test.py
```diff
@@ -0,0 +1,119 @@
...
+    [
+        pytest.param("table_1", "reported_1", 1.0, LEAF_NODE_1, does_not_raise()),
+        pytest.param("table_1", "reported_2", -1.0, LEAF_NODE_2, does_not_raise()),
+        pytest.param("table_1", "calc_1", 1.0, CALC_TREE_1, does_not_raise()),
+        pytest.param(
+            "table_2", "calc_2", 1.0, CALC_TREE_2, pytest.raises(AssertionError)
+        ),
+        pytest.param("table_1", "calc_3", 1.0, CALC_TREE_1, pytest.raises(KeyError)),
+    ],
+)
```
zaneselvans, line 102: @jdangerx Are there other obvious test cases you think we should have in here?

Is there a less verbose way that I could be specifying these?

test/unit/output/ferc1_test.py
e-belfer: To clarify, in this case one but not all of the components comes from outside of the set of exploded tables? 

Re: your second question - if this process is upstream of the reconciliation process and it returns an empty calculation, I think it should just return the reported value. The metadata will still mark it as `calculated_value` though, so we should think about whether we want to flag this in any way. 

I think the question of having some components sourced from in-table and some from out-of-table is discussed in #2622. For now, it seems like this isn't behavior that we're expecting to happen, so I think catching the `KeyError`, verifying it's due to an out-of-bounds table and returning either a warning ("unexpected components, component will not be calculated") or an error would make sense as the behavior here. I'd actually lean towards an error, though perhaps this will make things overly brittle.

cmgosnell: I don't have a lot of clarify around how the metadata explosion/calculation tree treatment is going to interplay with the data explosion. but i do feel clear-ish about how this type of thing needs to be treated in the output.

If any component of a calculated factoid is outside of the explosion tables that we care about we remove all of the components of the calculation *that are within the explosion* so we can leave only the calculated value. And then that calculated factoid is now the leaf.

So in the context of a factoid's calculation tree... idk. Removing the calculation itself and redefining the factoid as a non-calculated value would effectively remove the information about how to remove it's remaining sub-components. UNLESS we use the calc trees to compile the set of leaves we actually care about?

src/pudl/output/ferc1.py
```diff
@@ -1337,3 +1340,205 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
         )
     ]
     return exploded
+
+
+################################################################################
+# XBRL Calculation Tree
+################################################################################
+class Ferc1XbrlCalculationTree(BaseModel):
```
cmgosnell, line 1348: i find this name a lil too ambiguous. it took me a minute to realize this was a tree for a *specific* `xbrl_factoid`. Also the putting `Ferc1` first feels the opposite of our general convention (mostly, there are a few instances of this in transform.ferc1)

test/unit/output/ferc1_test.py
zaneselvans: Based on our conversation this morning, I think what we decided to do here is pre-emptively any calculation that mentions a table outside of the set of tables being exploded in the `ExplodeMeta` step, replacing it with `"[]"` and changing `xbrl_row_type` to `reported_value` -- for the purposes of the exploded table analysis.

Then, the calculation tree which is generated from the exploded metadata will automatically consider those values leaves, and any other values within the exploded tables that were referenced in mixed calculations (those with values from more than one table) will automatically get pruned from the overall calculation tree, making it very straightforward to use the calculation tree to identify the facts which need to be retained in the exploded dataframe (i.e. only those facts which are associated with leaves of the tree).

test/unit/output/ferc1_test.py
```diff
@@ -0,0 +1,119 @@
+"""Tests for the FERC Form 1 output functions.
+
+Calculation tree generation:
+- [x] Test base recusion case (leaf node w/ empty calculations)
+- [x] Test simple 1-level recursion w/ 1-2 child nodes.
+- [x] Test calculation with non-unique source table (should fail)
+- Test calculation that references an outside table (???)
```
jdangerx, line 7: Is this the `KeyError` test case down below now?

jdangerx (approved):
Re: testing strategy + the tree class structure, both look good now! Left a couple suggestions to make the tests more "readable" but obviously that's somewhat subjective.

test/unit/output/ferc1_test.py
zaneselvans: Ah yeah -- but I think @cmgosnell and I worked out what should actually be happening, which is that we should never encounter this situation because `ExplodeMeta` has already revised the calculation metadata to remain within the boundaries of the Exploded Metadata.  So I'll change this to an `xfail`.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+
+    def to_leafy_meta(self: Self) -> pd.DataFrame:
+        """Convert an XbrlCalculationTree back into a pandas dataframe.
+
+        - Indexed by (source table, xbrl_factoid)
+        - Each family of tags gets its own column
+        - Weights reflect updated / propagated values
+        - FOREST ONLY: Include column indicating the ID of the root node
+        """
+        G = self.to_networkx()  # noqa: N806
```
cmgosnell, line 1637: why the single letter variables? is this a networkx convention or something? it's... ya know hard to read.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+
+        leaves = [n for n, d in G.out_degree() if d == 0]
+        H = nx.DiGraph()  # noqa: N806
+        H.add_nodes_from(leaves)
+        nx.set_node_attributes(H, dict(G.nodes))
+
+        idx_cols = ["source_table", "xbrl_factoid"]
```
cmgosnell, line 1644: 1000% a nit but if you aren't gonna use these until the very last line i think it'd be more readable to move it down there.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+            nx.set_node_attributes(digraph, dict(child_graph.nodes))
+            digraph.add_edges_from(child_graph.edges)
+        return digraph
+
+    def to_leafy_meta(self: Self) -> pd.DataFrame:
+        """Convert an XbrlCalculationTree back into a pandas dataframe.
+
+        - Indexed by (source table, xbrl_factoid)
+        - Each family of tags gets its own column
+        - Weights reflect updated / propagated values
```
cmgosnell, line 1634: these two are only true if these steps have definitely been run right? it doesn't seem like there is anything in here that directly does that?

cmgosnell (changes requested):
this all looks really cool! i'm so glad that it seems like networkx had a lot of the functionality that we needed and you were able to implement it (bc omigosh i've had trouble navigating their docs just bc of pure breadth of content/lingo).

I sprinkled a ton of "how does this work" kind of questions. some of these are just from my own ignorance of the tool so don't feel the need to answer everything if you don't think it's super relevant for using the setup.

My high level question or suggestion is just about flow of the methods. I think there is some implied "this gets run first" kind of things sprinkled throughout. I tried to comment on those when i saw them but that was my biggest source of concern tbh.

It would also be helpful for me to see the forest generated inside of `Exploder` so i could get a sense of how we would build and maybe use the forest. even if it was a lil make the forest/get the leaves but not use them yet.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
         )
     ]
     return exploded
+
+
+################################################################################
+# XBRL Calculation Tree
+################################################################################
+class XbrlCalculationTreeFerc1(BaseModel):
+    """A FERC Form 1 XBRL calculation tree.
```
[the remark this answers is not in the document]
zaneselvans: Can you say more about what information you feel like is missing / incorrect / opaque / surprising here?  The overall data structure is built around the calculations, which refer to `(table, fact)` tuples as their elements -- and those tuples are the unique identifiers for the nodes in the tree. Because we want to uniquely identify a root node and the associated leaf nodes, we have to use a directed graph and we have to identify which node we're using as the root of the tree -- otherwise you can select any node that is only connected to a single other node, and declare it the root, and still have a valid tree (structurally).

Is there another thing other than the factoids that you are imagining could potentially have been the basis of the tree? What does including factoid help disambiguate?

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+            source_table: The table within which we are searching for the XBRL fact,
+                required as many appear in multiple tables.
+            xbrl_factoid: Equivalent to the "name" field in the calculation metadata.
+                Potentially renamed from the originally reported XBRL value.
+            exploded_meta: A dataframe containing metadata for all of the facts
+                referenced by the exploded tables. Must be indexed by the columns
+                (table_name, xbrl_factoid) and the index must be unique.
+            weight: The weight associated with this node in its calculation. Must be
+                passed in because it is only available when we have access to the
+                calculation metadata. Defaults to 1.0 so that we can create the root
```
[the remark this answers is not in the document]
zaneselvans: The meaning of the weight is different in the root node vs. all the other nodes. The root node has been plucked out of whatever its original calculation context was -- it effectively has no weight... and the only weight that really makes sense in that context is 1.0 (since presumably we're interested in the actual value that would be calculated for this node, based on all of the nested child calculations that feed into it).

In practice, for all nodes except the root node, this value is filled in with the weight which is read in from `exploded_meta` but because we need to use the same constructor for the root node and all of those child nodes which are part of the tree.  We are passing `exploded_meta` down through the recursive construction of the tree as a place to look up the weights, lists of children, etc.  I forgot to edit the docstring when I changed the default value to `np.nan`.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+class XbrlCalculationForestFerc1(BaseModel):
+    """A class representing a collection of :class:`XbrlCalculationTreeFerc1`."""
+
+    trees: list[XbrlCalculationTreeFerc1] = []
+
+    @classmethod
+    def from_exploded_meta(
+        cls,
+        source_tables: list[str],
+        xbrl_factoids: list[str],
```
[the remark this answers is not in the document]
zaneselvans: I zip the lists together and build `(table, fact)` tuples out of them, so they do end up being uniquely associated.  I thought about making the input a list of tuples, or defining a `XbrlCalculationSeedFerc1` class which is just a tuple.  Not a big fan of having these lists be separate either.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+        - Convert these Graphs back into Calculation Trees.
+        """
+        trees = [
+            XbrlCalculationTreeFerc1.from_exploded_meta(
+                source_table=source_table,
+                xbrl_factoid=xbrl_factoid,
+                exploded_meta=exploded_meta,
+            )
+            for source_table, xbrl_factoid in zip(source_tables, xbrl_factoids)
+        ]
```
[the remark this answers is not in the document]
zaneselvans: By zipping together the two lists, I'm creating a set of tuples that map the elements of the lists together 1:1.  These lists contain all the (probably root) facts that we want to ensure are included in the output tree. They could be a hand-picked set of facts that we care about from, say, the top-level income statement table, or they could be the list of all facts which show up in the income statement table. If not all of them are actually root facts, that's fine -- we use the in-degree and out-degree to figure out which is what later on.

If we give it a non-existent (table, fact) then I think it will interpret that as having gone out of bounds (i.e. it won't find that element in the `exploded_meta` table) -- I was thinking of adding a check that can ensure when this happens the `table` is always not found at all in the index of `exploded_meta` because if the table **is** found, and it's the **fact** that makes it missing... that's probably a typo or something.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+        forest = cls.to_networkx(trees)
+
+        # Identify the roots of each tree and verify it's one of the input seeds
+        roots = [n for n, in_deg in forest.in_degree() if in_deg == 0]
+        logger.debug(f"{roots=}")
+        leaves = [n for n, out_deg in forest.out_degree() if out_deg == 0]
+        logger.debug(f"{leaves=}")
```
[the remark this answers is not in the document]
zaneselvans: I guess I'm not verifying that it's one of the input seeds. But also I'm not 100% sure why I thought I needed to do that.

No, I'm not doing anything with the leaves here. I think I just cut-and-pasted this over from my notebook where I was experimenting with these things.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+                source_table=source_table,
+                xbrl_factoid=xbrl_factoid,
+                exploded_meta=exploded_meta,
+                weight=1.0,
+                propagate_weights=propagate_weights,
+                tags_df=tags_df,
+            )
+            for source_table, xbrl_factoid in roots
+        ]
+        return XbrlCalculationForestFerc1(trees=minimal_trees)
```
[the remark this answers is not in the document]
zaneselvans: Graphs behave kind of like sets -- if you add the same node more than once, it doesn't actually show up again. It's just ignored.  So it's impossible to end up with duplicate roots.  In the `to_networkx()` method we add all of the nodes from all of the trees to the same graph, which lets us identify what the distinct trees making up the forest are, and the minimal set of root nodes required to get that same forest, and all of the leaf nodes which are part of the forest.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+
+    @classmethod
+    def to_networkx(cls, trees: list["XbrlCalculationTreeFerc1"]) -> nx.DiGraph:
+        """Convert a forest into a :class:`networkx.DiGraph`."""
+        forest = nx.DiGraph()
+        for tree in trees:
+            nx_tree = tree.to_networkx()
+            assert nx.is_tree(nx_tree)
```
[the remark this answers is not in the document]
zaneselvans: It sure better be a tree!  This is just defensive.  If there's something screwy in the calculation metadata where we end up with a relationship that's not structured like a tree then everything will be broken. It's an easy thing to check so I thought I would just go ahead and check.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+            forest.add_nodes_from(nx_tree.nodes)
+            nx.set_node_attributes(nx_tree, dict(nx_tree.nodes))
+            forest.add_edges_from(nx_tree.edges)
```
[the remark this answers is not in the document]
zaneselvans: I'll expand the docstring. Let me know if it's clear enough and if not I'll add some inline comments too.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+
+    def to_leafy_meta(self: Self) -> pd.DataFrame:
+        """Generate a metadataframe from the leaves of an XBRL Calculation Forest."""
+        dfs = []
+        for tree in self.trees:
+            new_df = tree.to_leafy_meta()
+            new_df["root_table"] = tree.source_table
+            new_df["root_xbrl_factoid"] = tree.xbrl_factoid
+            dfs.append(new_df)
+        return pd.concat(dfs).sort_index()
```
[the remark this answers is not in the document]
zaneselvans: At this point the forest has been constructed -- it's `self`. The forest is just a collection of trees which (in our case) have been deduplicated / simplified by adding them all to the same graph.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+                    f"{calc=}"
+                )
+            children.append(
+                XbrlCalculationTreeFerc1.from_exploded_meta(
+                    weight=calc["weight"],
+                    xbrl_factoid=calc["name"],
+                    source_table=calc["source_tables"][0],
+                    exploded_meta=exploded_meta,
+                )
+            )
```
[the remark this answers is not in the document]
zaneselvans: Yes exactly! Each layer is only responsible for creating its own children, and then those children create their children, etc. -- this is recursion.  The "termination condition" is when you get to a node which has nothing in its list of calculations, in which case it just returns an empty list.

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+
+        tree = XbrlCalculationTreeFerc1(
+            source_table=source_table,
+            xbrl_factoid=xbrl_factoid,
+            weight=weight,
+            xbrl_factoid_original=exploded_meta.at[idx, "xbrl_factoid_original"],
+            children=children,
+        )
```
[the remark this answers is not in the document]
zaneselvans: I think this **is** the "primary object" -- above we've constructed its children, but this is the root node of the tree that we were actually asked to construct.  Because this is a recursive function, it can't tell what level in the overall construction its operating at.  This object is the current node being constructed.  The children are each also constructed by this -- but in a deeper call to `from_exploded_meta()`

src/pudl/output/ferc1.py
```diff
@@ -1399,3 +1402,341 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+        if propagate_weights:
+            tree = tree.propagate_weights()
+        if tags_df is not None:
+            tree = tree.add_tags_from_df(tags_df).propagate_tags()
+
+        return tree
+
+    def propagate_weights(
```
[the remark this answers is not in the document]
zaneselvans: The new weight isn't directly stored within the children, but it's used to modify the weights of the children.

src/pudl/output/ferc1.py
zaneselvans: As it is right now, there's definitely some time when things are internally inconsistent. I chain together a bunch of stuff in the `to_leafy_meta()` method that ensures the output reflects all of the necessary stuff having been run, but I don't feel great about how it's all arranged.

src/pudl/output/ferc1.py
zaneselvans: I think there were some other similar contexts where I needed it earlier and so it just became habitual. I've added it into the `set_index()` call directly.

zaneselvans:
I agree about the flow of the methods not being great. I'm not sure how to structure the class / methods such that the methods are only being called / relied upon when appropriate -- like having propagated the tags / weights before trying to generate a leafy metadataframe, or construct a calculation with the weights.

There are several things we need to do with the leafy metadata:
* Need the list of leaf facts, and the root facts that they are associated with, along with all the metadata that is associated with each leaf. This is used to remove irrelevant records from the exploded dataframe and to merge useful metadata onto the exploded dataframe.
* Need to be able to calculate the root values based only on the most granular reported leaf values using our existing infrastructure, so that we know we didn't mess anything.
* Need to use the propagated / adjusted weights to adjust the values in exploded dataframe.
* Will be feeding information into `reconcile_table_calcs()` to verify the calculations.

zaneselvans:
Okay, I have the exploded data, pruned to include only the most granular values (leaves) for the balance sheet assets tables. The tables I'm including:
```py
balance_sheet_asset_tables = [
    "balance_sheet_assets_ferc1",
    "plant_in_service_ferc1",
]
```

## Caveats:

* The `plant_in_service_ferc1` table still isn't connected to the rest of the tree by a calculation. I'm including it by adding its root node explicitly.
* The calculations and tree relationships are not yet aware of additional dimensions like `utility_type`
* We need more robust internal consistency checks on our edits to the calculations. They're hand compiled and likely to contain some errors.
* There's one instance in which the same facts really do show up in two different calculations, which means the tree isn't really a tree:

The facts being calculated:
```py
[
    NodeId(source_table='balance_sheet_assets_ferc1', xbrl_factoid='current_and_accrued_assets'),
    NodeId(source_table='balance_sheet_assets_ferc1', xbrl_factoid='other_property_and_investments'),
]
```

The duplicated components:
```py
[
    NodeId(source_table='balance_sheet_assets_ferc1', xbrl_factoid='noncurrent_portion_of_allowances'),
    NodeId(source_table='balance_sheet_assets_ferc1', xbrl_factoid='derivative_instrument_assets_long_term'),
    NodeId(source_table='balance_sheet_assets_ferc1', xbrl_factoid='derivative_instrument_assets_hedges_long_term')
]
```

* There's one real node that is being orphaned -- it shows up in the metadata, but is never referenced in any calculation -- is this expected and desired? `NodeId('utility_plant_summary_ferc1', 'depreciation_amortization_and_depletion_utility_plant_in_service')`
* There's one node `NodeId(source_table='utility_plant_summary_ferc1', xbrl_factoid='accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility')` that is part of a redirect, and it's also involved in a calculation with a negative sign, so it shows up twice in the construction of the tree, but with different signs, which creates a conflict. Somehow we need to identify the passthrough nodes preemptively and just not add them during construction, rather than removing them after the fact.

zaneselvans:
> `NodeId('utility_plant_summary_ferc1', 'depreciation_amortization_and_depletion_utility_plant_in_service')` is currently being orphaned -- it shows up in the metadata but not in any calculation. Is this expected / desired?

**This is the desired outcome -- it is being replaced by its constituent parts.

> There's one node `NodeId(source_table='utility_plant_summary_ferc1', xbrl_factoid='accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility')` that is part of a redirect, and it's also involved in a calculation with a negative sign, so it shows up twice in the construction of the tree, but with different signs, which creates a conflict. Somehow we need to identify the passthrough nodes preemptively and just not add them during construction, rather than removing them after the fact.

**I'm setting the weight to -1.0 if the weights are not the same, and they multiply to equal -1.0.***

src/pudl/output/ferc1.py
```diff
@@ -1677,3 +1790,508 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+            .groupby(["root_table", "root_xbrl_factoid"], as_index=False)
+            .apply(leafy_meta_to_calculations)
+        )
+        root_calcs.columns = ["root_table", "root_xbrl_factoid", "calculations"]
+        return root_calcs
+
+    @property
+    def table_names(self: Self) -> list[str]:
+        """Produce the list of tables involved in this explosion."""
+        return list(self.exploded_meta.reset_index()["table_name"].unique())
```
zaneselvans, line 2223: I think this is correct, but it's also kind of duplicative of information in the `Exploder` that exists outside of the tree context.

src/pudl/output/ferc1.py
```diff
@@ -1677,3 +1790,508 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+        self.plot(self.forest, table_names=self.table_names)
+
+    def leafy_data(
+        self: Self, exploded_data: pd.DataFrame, value_col: str
+    ) -> pd.DataFrame:
+        """Use the calculation forest to prune the exploded dataframe.
+
+        - Drop all rows that don't correspond to either root or leaf facts.
+        - Verify that the reported root values can still be generated by calculations
+          that only refer to leaf values.
```
zaneselvans, line 2263: This is not currently happening.

src/pudl/output/ferc1.py
```diff
@@ -1677,3 +1790,508 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+        - Merge the leafy metadata onto the exploded data, keeping only those rows
+          which refer to the leaf facts.
+        - Use the leaf weights to adjust the reported data values, and then drop the
+          leaf weights.
```
zaneselvans, line 2267: We aren't dropping the columns here -- eventually it'll be done by `enforce_schema()`

src/pudl/transform/ferc1.py
```diff
@@ -1353,6 +1353,23 @@ def rename_calculation_components(calc: str) -> str:
             )
         ).astype({"row_type_xbrl": pd.StringDtype(), "calculations": pd.StringDtype()})
 
+        # Create metadata records for the calculation correction factoids
+        correction_meta = (
```
zaneselvans, line 1357: Blanket addition of correction factoids for all calculated values, which may or may not end up getting used -- it'll depend on whether the calculations add up correctly.

src/pudl/transform/ferc1.py
```diff
@@ -1353,6 +1353,23 @@ def rename_calculation_components(calc: str) -> str:
...
+            tbl_meta[tbl_meta.row_type_xbrl == "calculated_value"]
+            .copy()
+            .assign(
+                calculations="[]",
+                intra_table_calc_flag=True,
```
zaneselvans, line 1362: Are they really always intra-table calculations?

src/pudl/transform/ferc1.py
```diff
@@ -2138,20 +2155,58 @@ def apply_xbrl_calculation_fixes(
                 ],
             },
             "balance_sheet_assets_ferc1": {
+                "current_and_accrued_assets": [
```
zaneselvans, line 2158: These are the values that truly showed up in two different calculations, and so had to be duplicated.  However they **also** seem to have had the wrong sign associated with them in the calculation metadata.

src/pudl/transform/ferc1.py
```diff
@@ -2138,20 +2155,58 @@ def apply_xbrl_calculation_fixes(
...
+                        },
+                    },
+                ],
                 "nuclear_fuel_net": [
                     {
                         "calc_component_to_replace": {
                             "name": "nuclear_fuel_materials_and_assemblies",
                             "weight": 1.0,
-                            "source_tables": ["balance_sheet_assets_ferc1"],
+                            "source_tables": [],
```
zaneselvans, line 2201: I think these might show up in another dedicated nuclear fuel table actually.

And I think there may be some naming confusing where in one table they end with `_major_only` and in another they don't?

src/pudl/transform/ferc1.py
```diff
@@ -2243,32 +2298,87 @@ def apply_xbrl_calculation_fixes(
                         },
                     }
                 ],
+                "utility_plant_net": [
```
zaneselvans, line 2301: I think this is a redirect / passthrough.

src/pudl/transform/ferc1.py
```diff
@@ -2656,30 +2774,6 @@ def apply_xbrl_calculation_fixes(
...
-                        },
-                    },
-                    {
-                        "calc_component_to_replace": {},
-                        "calc_component_new": {
-                            "name": "steam_production_plant",
-                            "weight": 1.0,
-                            "source_tables": ["plant_in_service_ferc1"],
-                        },
-                    },
```
zaneselvans, line 2682: These were part of what seemed to be an incomplete and unnecessary redirect that went around the node gathering all `production_plant` values into a single calculated value.  However it left out some of the elements of that calculated value, and I'm not sure why it seemed to be necessary. It was breaking the tree structure and seemed incorrect so I removed it.

src/pudl/output/ferc1.py
```diff
@@ -1677,3 +1790,508 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+    def seeded_digraph(self: Self) -> nx.DiGraph:
+        """A digraph of all calculations that contribute to the seed values."""
+        seeded_nodes = set()
+        for seed in self.seeds:
+            seeded_nodes = seeded_nodes.union([seed])
+            seeded_nodes = seeded_nodes.union(nx.descendants(self.full_digraph, seed))
+        seeded_digraph: nx.DiGraph = self.exploded_meta_to_digraph(
+            exploded_meta=self.exploded_meta.loc[list(seeded_nodes)],
+            tags=self.tags,
+        )
```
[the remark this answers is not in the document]
cmgosnell: i don't have enough context/knowledge here to help you on this one. but it looks like you need this `exploded_meta_to_digraph` in other contexts so if it's useful for more than just here and it isn't causing a problem here than idk maybe just stick with it

src/pudl/output/ferc1.py
```diff
@@ -1677,3 +1790,530 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+
+        In practice this method will probably return a single tree rather than a forest,
+        but a forest with several root nodes might also be appropriate, since the root
+        table may or may not have a top level summary value that includes all underlying
+        calculated values of interest.
+        """
+        forest = self.seeded_digraph
+        # Remove any node that has only one parent and one child, and add an edge
+        # between its parent and child.
+        for node in self.passthroughs:
```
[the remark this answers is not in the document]
cmgosnell: wait now that i say this i am seeing that really the `self.seeded_digraph` is a forest once these passthroughs are removed?

cmgosnell (commented):
i am not through reviewing this yet but here are a smattering of comments thus far 🌳

src/pudl/output/ferc1.py
```diff
@@ -1677,3 +1790,508 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+    tags: pd.DataFrame = pd.DataFrame()
+
+    class Config:
+        """Allow the class to store a dataframe."""
+
+        arbitrary_types_allowed = True
+
+    @validator("exploded_meta", "tags")
+    def ensure_correct_dataframe_index(cls, v):
+        """Ensure that dataframe is indexed by table_name and xbrl_factoid."""
```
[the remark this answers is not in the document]
zaneselvans: Yes, this just says "use this validator on both of these inputs" rather than needing to specify it multiple times for each input individually.

src/pudl/output/ferc1.py
```diff
@@ -1677,3 +1790,508 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+        idx_cols = ["table_name", "xbrl_factoid"]
```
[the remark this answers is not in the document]
zaneselvans: Yeah that's probably true.

src/pudl/output/ferc1.py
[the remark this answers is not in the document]
zaneselvans: For the Balance Sheet Assets table, this is true. Right now we've got 3 different sources of non-tree structures:
* errors in the calculations
* passthrough calculations that can be pruned
* the one instance where the same values were involved in multiple calculations but with opposite signs, where we duplicated the factoids, and they cancel each other out in higher level calculations.

For the Balance Sheet Liabilities and Income Statement tables, we still have some calculation cleanup to do, so after removing the passthroughs they still aren't trees.

I think if we end up with more than one pruning step (to turn the DAG into a tree) then we should break the different steps out into their own methods and chain them together here, but it felt premature to do that when there's only one step.

src/pudl/output/ferc1.py
```diff
@@ -1117,9 +1206,8 @@ def boom(
         #. Prep all of the individual tables for explosion.
         #. Concatenate all of the tabels together.
         #. Remove duplication in the concatenated exploded table.
-        #. Add in metadata for the remaining ``xbrl_factoid`` s regarding linage (not
-           implemented yet).
-        #. Validate (not implemented yet).
+        #. Annotate the fine-grained data with additional metadata.
+        #. Validate that calculated top-level values are correct.
```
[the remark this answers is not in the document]
zaneselvans: Noted!

src/pudl/output/ferc1.py
```diff
@@ -1034,19 +1088,54 @@ class Exploder:
...
+            tags: Additional metadata to merge onto the exploded dataframe.
         """
-        self.table_names = table_names
-        self.root_table = root_table
-        self.meta_exploder = MetadataExploder(self.table_names)
-        self.metadata_exploded = self.meta_exploder.boom(clean_xbrl_metadata_json)
+        self.table_names: list[str] = table_names
+        self.root_table: str = root_table
+        # I don't think we actually need the metadata exploder to stick around?
+        # self.meta_exploder: MetadataExploder = MetadataExploder(self.table_names)
```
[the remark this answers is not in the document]
zaneselvans: Okay, removed!

src/pudl/output/ferc1.py
cmgosnell: can you note that it is not currently implemented? also do you think that the verification would end up living in this method? it seems like it would probably be a separate step no?

src/pudl/output/ferc1.py
cmgosnell: can you update the docs to be more reflective of whats going on in here then?

src/pudl/output/ferc1.py
```diff
@@ -912,54 +937,83 @@ def exploded_tables_asset(
...
+    # Also, these tags are only applicable to the balance_sheet_assets_ferc1 table, but
+    # we need to pass in a dataframe with the right structure to all of the exploders,
+    # so we're just re-using this one for the moment.
+    pkg_source = importlib.resources.files("pudl.package_data.ferc1").joinpath(
+        "xbrl_factoid_rate_base_tags.csv"
+    )
+    with importlib.resources.as_file(pkg_source) as tags_csv:
+        tags = pd.read_csv(
+            tags_csv, usecols=["xbrl_factoid", "table_name", "in_rate_base"]
+        ).drop_duplicates(subset=["table_name", "xbrl_factoid"])
```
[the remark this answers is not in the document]
zaneselvans: You may have, but I don't know whether it should be in or not in rate base, and I couldn't tell from looking at the CSV.  The `xbrl_factoid` shows up 3 times -- twice as "no" (in 2 different tables) and once as "yes":

`accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility`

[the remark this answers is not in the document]
zaneselvans: I talked to @jrea-rmi and he said that it's possible this fact contains some components that are in rate base, and others that aren't, so we might need to label some of its subcomponents, but more generally we need to get these labels reviewed by e.g. David Posner or someone who actually understands the accounting when we've got a result to share.  Much of it is supposed to be in rate base so I'm going with that for now.

src/pudl/output/ferc1.py
```diff
@@ -1034,19 +1088,54 @@ class Exploder:
...
+            )
+            logger.info(f"Identified {seeds=}")
+
+        self.calculation_forest: XbrlCalculationForestFerc1 = (
+            XbrlCalculationForestFerc1(
+                exploded_meta=self.metadata_exploded,
+                seeds=seeds,
+                tags=tags,
+            )
+        )
```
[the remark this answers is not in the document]
zaneselvans: I've turned this into a property, so it's only constructed when asked for.

src/pudl/output/ferc1.py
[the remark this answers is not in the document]
zaneselvans: I'm being more careful about dropping duplicates in the tagging metadata now, such that if another inconsistency like this appears it'll fail.

src/pudl/output/ferc1.py
```diff
@@ -1677,3 +1790,508 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+    @validator("exploded_meta", "tags")
+    def dataframe_has_unique_index(cls, v):
+        """Ensure that exploded_meta has a unique index."""
+        if not v.index.is_unique:
+            raise ValueError("DataFrame has non-unique index values.")
+        return v
+
+    @validator("seeds", always=True)
+    def seeds_not_empty(cls, v, values):
+        """If no seeds are provided, use all nodes in the index of exploded_meta."""
```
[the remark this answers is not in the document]
zaneselvans: The normal use-case is the provide a short list of seed nodes, whose descendants will make up the seeded digraph -- this is a first coarse round of pruning. Those short lists of seed nodes are inputs into the exploder / exploded assets.

The difference in behavior here is what happens when no seed nodes are provided.  In the Exploder, there's currently a notion of a "root table" -- though I think we are going to remove that -- and so the Exploder chooses its seeds based on the set of nodes in that root table.  Here in the `XbrlCalculationForest` there's no such thing as a root table.

I think the right thing to do here is probably have both of them do the same thing, which is add all the nodes if no seeds are provided, given that the `root_table` is going to go away.

src/pudl/output/ferc1.py
```diff
@@ -1677,3 +1790,508 @@ def remove_intra_table_calculated_values(exploded: pd.DataFrame) -> pd.DataFrame
...
+            v = list(values["exploded_meta"].index)
+        return v
+
+    @validator("seeds")
+    def seeds_within_bounds(cls, v, values):
+        """Ensure that all seeds are present within exploded_meta index."""
+        bad_seeds = [seed for seed in v if seed not in values["exploded_meta"].index]
+        if bad_seeds:
+            raise ValueError(f"Seeds missing from exploded_meta index: {bad_seeds=}")
+        return v
```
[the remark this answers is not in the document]
zaneselvans: I'm not sure what you're saying about needing the exploded data to do this check. I'm just suggesting that we look at all the listed calculation components (after they've been re-written to remove references to tables outside the explosion) to make sure that all of them are present within the metadata.

I've gone ahead and added this validation step in the metadata exploder.

src/pudl/output/ferc1.py
zaneselvans: Okay, updated.

 Given the NA values that are showing up in the leafy data, should we debug the missing `_correction` records? Or do an inner merge rather than a left merge onto the metdataframe?

src/pudl/output/ferc1.py
zaneselvans: Like I say below in the docstring I think it could live either place. It'll need access to both the exploded data and the leafy metadata. This class contains the leafy data.  The Exploder contains the exploded data.  Either we have to pass the data into this class to do something, or we have to pull the leafy metadata out of this class into a method that's part of the exploder.

It seems like it might be nice to keep the root calculation validation and the merge in the same place because we temporarily need to have the root node data & metadata in the dataframe to do the validation, but then we want to drop those records.  Where do you think it makes the most sense to do these things?

src/pudl/output/ferc1.py
```diff
@@ -1130,21 +1218,45 @@ def boom(
...
-            # REMOVE THE DUPLICATION
-            # .pipe(self.remove_factoids_from_mutliple_tables, tables_to_explode)
-            # .pipe(self.remove_totals_from_other_dimensions)
-            # .pipe(self.remove_inter_table_calc_duplication)
-            # .pipe(remove_intra_table_calculated_values)
+            .pipe(self.calculation_forest.leafy_data, value_col=self.value_col)
         )
-        return exploded
+        # Identify which columns should be kept in the output...
+        # TODO: Define schema for the tables explicitly.
```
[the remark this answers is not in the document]
zaneselvans: The tag columns are included here with the filter regex.

Would you like me to go ahead and define the database tables for the 3 exploded tables and start trying to load them in Dagster?

zaneselvans:
I think I have responded to all of your comments and made some minor changes.  Bigger questions and higher level responses:

Other than breaking out the visualization, I'm not a fan of breaking the class up into smaller pieces, because those pieces will still be very tightly coupled to each other. The difference between the full, seeded, and pruned (forest) graphs isn't just what nodes and edges they contain -- there are relationships between them and they get used together in several methods, and having all those methods in the same class, which knows how to produce those different graphs based on the relatively simple top level inputs seems simpler to me than having a bunch of that logic outside of the class, and working with multiple simpler instances of graphs.

## TODO:
I think it makes sense to pull the plotting / visualization out into a standalone function that takes one of the output digraphs as an argument, so I'll do that.

## QUESTIONS
What should we do about the fact that the ratio of valid calculations is apparently quite variable by year, and the most recent years (including the XBRL data that the calculations are derived from...) do worse than the historical average, to the point where the CI fails?

Do you want me to go ahead and define database table schemas for the new `exploded_*` assets in this PR?

I looked at making the index columns an attribute of the `XbrlCalculationForest` class, so that as we move that indexing / definition further upstream we can pass it in. However, at the moment refactoring the class to use a generic index attribute looks kind of painful for 2 reasons:
* We're still dealing with two different sets of names for these columns `(source_tables, name)` and `(table_name, xbrl_factoid)`
* In a variety of contexts we use just one element of the index for other purposes, and those would still be hard coded, or we would need to add other attributes that indicate what value to look at for what's currently the table name or the fact columns.
Should I just go ahead and store the index columns in the class for now and not worry about those other hard-codings?

zaneselvans:
@cmgosnell: 

* I've simplified the plotting so there's just one method
* I've stored the exploded metadata index columns in the calculation forest class -- we'll need to revisit this and migrate it up into other higher level classes as we deal with the dimensions and refactor.
* I've bumped the calculation error tolerances for the income statement and balance sheet liability tables so that they pass CI -- I think, anyway.  I'm getting the same divide by zero error that @e-belfer was getting now, since I merged in her changes from upstream.  So I don't know if the thresholds are actually high enough.
* Holding off on defining the DB tables until we're ready to merge into `dev`
* Still need to merge the multi-parent fixes from @e-belfer in the balance sheet liabilities table.

cmgosnell (approved):
i still think this giant class out be nicer to work with if it was decomposed a little bit but in the same vein of ella's inter-dim pr, we will keep working on this and we are not merging into dev so lets goooo

PR #2654, opened 2023-06-12
Throw different exception when dbc file is missing

rousik:
Earlier, generic `KeyError` was thrown here. This resulted in the code silently bypassing archives that did not have correct dbc path as the `KeyError` was also caught when tables were missing.

Perhaps there could be further schema tests along the lines of - each year/archive should have at least 90% of the known tables present. This could catch situation where we are, for whatever reason, not reading the tables from some partitions.

### DIFF
--- a/src/pudl/extract/dbf.py
+++ b/src/pudl/extract/dbf.py
@@ -21,6 +21,12 @@ from pudl.workspace.datastore import Datastore
 logger = pudl.logging_helpers.get_logger(__name__)
 
 
+class DbcFileMissing(Exception):
+    """This is raised when the DBC index file is missing."""
+
+    pass
+
+
 class DbfTableSchema:
     """Simple data-wrapper for the fox-pro table schema."""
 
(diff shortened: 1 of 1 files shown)


zaneselvans (approved):
Should we be creating more custom error classes rather than just raising `AssertionError` or `ValueError` everywhere? I've never really understood what all extra functionality we could be adding to the errors.

rousik:
> Should we be creating more custom error classes rather than just raising `AssertionError` or `ValueError` everywhere? I've never really understood what all extra functionality we could be adding to the errors.

It depends. In theory, you could add a lot of structured information and/or context to errors that could make them more usable, but I'm not sure there's a clear need for that here. In this more general situation, you could attach things like dataset/partition/filename/table to the exception so that the stacktrace becomes more informative about where the problem originated, but I suppose that this would be quite a bit of work to plumb through and I'm unsure about the benefits.

In this particular case, the problem was that one public API function could return `KeyError` in two very different failure modes so new exception allows us to distinguish these cases. In general, I think that would be useful in some scenarios where the generic exceptions are too broad.

PR #2659, opened 2023-06-12
Add publish_destinations field to the ETL config

rousik:
`publish_destinations` feature will optionally allow writing contents of the `PUDL_OUTPUT` directory to remote paths, e.g. to gcs or s3 buckets. This can be used to further reduce the complexity of the runner shell script that we currently run in the docker container.


src/pudl/settings.py
```diff
@@ -611,17 +611,20 @@ class EtlSettings(BaseSettings):
     pudl_in: str = pudl.workspace.setup.get_defaults()["pudl_in"]
     pudl_out: str = pudl.workspace.setup.get_defaults()["pudl_out"]
 
+    # This is list of fsspec compatible paths to publish the output datasets to.
+    publish_destinations: list[str] = []
```
jdangerx, line 615: Do we need set some destinations in the settings files so that the nightly build script still publishes?

jdangerx (dismissed):
Seems useful :) Just one blocking question - if the answer is "no, this is unnecessary" then feel free to dismiss & merge!

src/pudl/settings.py
rousik: No, the current functionality (publishing via wrapper shell script) should be unaffected. Eventually we'll want to switch to the "new" mode and drop the shell script stuff, but for now, we keep relying on the old method.

PR #2661, opened 2023-06-12
Deduplicate FERC 2 respondent IDs

zaneselvans:
# PR Overview

Add a `table_transform()` method to the `Ferc2DbfExtractor` that deduplicates the `f2_s0_respondent_id` table, similar to what we do in the FERC 1.

# PR Checklist

### DIFF
--- a/src/pudl/extract/ferc2.py
+++ b/src/pudl/extract/ferc2.py
@@ -12,6 +12,7 @@ million dekatherms.
 
 from typing import Any
 
+import pandas as pd
 import sqlalchemy as sa
 
 import pudl
@@ -39,6 +40,18 @@ class Ferc2DbfExtractor(FercDbfExtractor):
         return add_key_constraints(
             meta, pk_table="f2_s0_respondent_id", column="respondent_id"
         )
+
+    def transform_table(self, table_name: str, in_df: pd.DataFrame) -> pd.DataFrame:
+        """FERC Form 2 specific table transformations.
+
+        Remove duplicate IDs from the table enumerating all respondents, retaining the
+        most recently reported version of the record. Assumes that records have been
+        added to the DB in chronological order.
+        """
+        if table_name == "f2_s0_respondent_id":
+            return in_df.drop_duplicates(subset="respondent_id", keep="last")
+        else:
+            return in_df
 
     @staticmethod
     def is_valid_partition(fl: dict[str, Any]):


src/pudl/extract/ferc2.py
```diff
@@ -40,6 +41,18 @@ def finalize_schema(self, meta: sa.MetaData) -> sa.MetaData:
             meta, pk_table="f2_s0_respondent_id", column="respondent_id"
         )
 
+    def transform_table(self, table_name: str, in_df: pd.DataFrame) -> pd.DataFrame:
+        """FERC Form 2 specific table transformations.
+
+        Remove duplicate IDs from the table enumerating all respondents, retaining the
+        most recently reported version of the record. Assumes that records have been
+        added to the DB in chronological order.
```
zaneselvans, line 49: Is this a valid assumption?  Would it be more robust and also easy for us to inject a `report_yr` column into this table as it's built up, sort by `report_yr` and `respondent_id`, drop duplicates, and then drop the `report_yr` column?

zaneselvans:
I ran the full `ferc_to_sqlite` script on all the data sources and years locally and it seemed to work as expected -- it ends up with 265 records rather than ~5000 in the `f2_s0_respondent_id` table.

README.rst
```diff
@@ -185,6 +185,7 @@ Datasette:
 
 * Raw FERC Form 2:
 
+  * `FERC-1 SQLite derived from DBF (1996-2020) <https://s3.us-west-2.amazonaws.com/intake.catalyst.coop/dev/ferc2.sqlite>`__
```
rousik, line 188:
```suggestion
  * `FERC-2 SQLite derived from DBF (1996-2020) <https://s3.us-west-2.amazonaws.com/intake.catalyst.coop/dev/ferc2.sqlite>`__
```

rousik (approved):
Minor typos to fix. I'm happy to either merge as-is with the subsequent refactor to make it use the `report_yr` outlined in the comment, or wait for the refactor to happen here. Whatever you prefer.

src/pudl/extract/ferc2.py
zaneselvans: I think I've implemented something like what you're suggesting here, if you'd take a look!

Miraculously it seems to have worked on the first try and not broken anything? Very suspicious.

zaneselvans:
I think these `transform_table_part()` methods are going to be very similar across all of the FERC Forms, and it seems like the add PK methods are also likely to be similar, with the main (only?) difference being the name of the table that contains the respondent IDs, so maybe we'd want to consolidate them somehow.

But I think this seems robust and good enough for now.

src/pudl/extract/dbf.py
```diff
@@ -181,28 +181,34 @@ def load_table(self, table_name: str) -> pd.DataFrame:
...
         ...
 
     def load_table_dfs(
-        self, table_name: str, partitions: list[dict[str, Any]]
+        self: Self, table_name: str, partitions: list[dict[str, Any]]
     ) -> pd.DataFrame | None:
         """Returns dataframe that contains data for a given table across given years."""
         ...
 
+    def transform_table_part(
```
rousik, line 206: Hm. This shows the limitation of the current architecture. I hoped that we can contain all dataset-specific logic in the FercDbfExtractor class and then rely on a single `FercDbfReader`. Now, we have `transform_table` in the Extractor but `transform_table_part` in the `FercDbfReader` which feels less than ideal.

I can think of two alternatives from the top of my head:
1. pass `df_transformer: Callable[...]` to `load_table_dfs` method and pass `FercDbfExtractor.transform_table_part` to it. This feels a bit off, because the dependencies go both ways, so the layering is not quite right.
2. move the logic for concatenating dfs out into `FercDbfExtractor`. e.g. instead of `load_table_dfs` returning single df, it could return list of tuples `(partition, df)` and that would be aggregated in extactor.

(2) might be the most straightforward approach for this IMO.

zaneselvans: I agree (2) sounds better. Would you be willing to take a stab at implementing this? I think I need to focus on the XBRL calculations/metadata stuff.

zaneselvans:
If possible I would like to get the bad respondent ID thing fixed though since this DB is out in the wild now. Actually I’m not sure why it isn’t barfing on the FK constraints, since the respondent ID column isn’t unique.  I suspect this is just SQLite being typically lazy about constraints. We should probably have SQLite explicitly validate the FK constraints at the end of these extractions, since it doesn’t do it automatically, along the lines of what the `pudl_check_fks` script is currently doing for `pudl.sqlite`.

src/pudl/extract/dbf.py
rousik: Sure, let me try to take a stab at this.

rousik (approved):
Cleaned up the code structure to use `aggregate_table_frames` method from `FercDbfExtractor`. This seems like a cleaner approach overall.

src/pudl/extract/ferc1.py
```diff
@@ -327,6 +306,15 @@ def get_observed_respondents(self) -> set[int]:
                 )
         return observed
 
+    def aggregate_table_frames(
+        self, table_name: str, dfs: list[PartitionedDataFrame]
+    ) -> pd.DataFrame | None:
+        """Deduplicates records in f_respondent_year table."""
+        if table_name == "f1_respondent_year":
```
zaneselvans, line 313: I think this should be `f1_respondent_id` table. Will update.

src/pudl/extract/ferc2.py
```diff
@@ -29,21 +30,31 @@ class Ferc2DbfExtractor(FercDbfExtractor):
...
         """Add primary and foreign keys for respondent_id."""
         return add_key_constraints(
             meta, pk_table="f2_s0_respondent_id", column="respondent_id"
         )
 
     @staticmethod
     def is_valid_partition(fl: dict[str, Any]):
-        """Returns False if part key has value other than None.
+        """Drops partition with non-empty `part` fields."""
+        return fl.get("part", None) is None
```
zaneselvans, line 47: I'm kind of confused by this method and how it works.

zaneselvans (commented):
Need to change table name from `f1_respondent_year` to `f1_respondent_id` in deduplicator for FERC 1.

Issue #2663, opened 2023-06-12
Convert 1000s of dollars to dollars & encode emissions control equipment type


zaneselvans:
The [`emissions_control_equipment_eia860`](https://data.catalyst.coop/pudl/emissions_control_equipment_eia860) table has a column `emissions_control_equipment_cost` which seems to contain suspiciously small numbers. Referring to the original EIA-860 spreadsheets, it appears that what's reported is thousands of dollars, not dollars, and so we need to multiply by 1000 to get the real values like we do in other tables.

The `emissions_control_equipment_type` column also appears to be a coded column, and is pretty inscrutable without a coding table to explain what the codes correspond to. Is there a reason we can't encode it like we do, say, the `operational_status_code` and call it `emissions_control_equipment_type_code` or `emissions_control_equipment_code`?

The meanings for these codes are reported in the Layout spreadsheet, in the tabs entitled Reference Tables 19-21, 22-24, and 25-28

Issue #2616, opened 2023-05-30
Add function to interpolate gaps in encoded columns


aesharpe:
## Issue Context

This issue was born out of the discussion in #2602

The `emissions_control_equipment_eia860` table had a values (`CS`) for `operational_status_code` in 2016 that were not outlined in the EIA code documentation for that year (or others).

The solution is two-fold:

1) Null out the bad codes by specifying them in `ignored_codes` as part of the `CODE_METADATA` in `codes.py` (this task is accomplished by #2602)

2) Create a function to interpolate missing `operational_status_code` values and other `CODE_METADATA` that is missing.

## Solution Breakdown

I was able to figure out that the rogue `CS` in `operational_status_code` was supposed to be `RE` based on the value for the same `particulate_control_id_eia` in the previous and following years. 

We might be able to: 
- Identify null values in the encoded columns
- Use the primary key minus `report_date` to see if the encoded values are consistent in other years
- Develop methodology to determine when we feel confident enough to interpolate the null value given the contextual information
- Fill in the null values with interpolated values.

zaneselvans:
I think this type of fix can probably be generalized beyond the encoded columns to any column of categorical values which we expect to change rarely if at all.  The encoding process will null out bad values, but there are other ways that we do already or could ensure that unencoded categorical values are also set to NA (e.g. with the [`pd.Series.map()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.map.html) method).

Retaining `report_date` and using it to order the records before filling is probably useful, since we probably want to prioritize continuity -- e.g. a generator technology type might actually change, and if there's an NA value that lies between the old technology type and the new technology type, it's much more ambiguous as to what value should be filled than if you've got one missing technology type value in the middle of a long series of identical values.

If this kind of data repair happens after the harvesting step then it won't influence our evaluation of whether a value is consistent enough to be retained.  It might be reasonable for the repair to happen before the harvesting step in all the tables where the value is originally reported, but that could have downstream effects.

PR #2319, opened 2023-02-20
Add boiler-associated attributes from EIA 860 6.2 EnvrEquip tables to ETL

e-belfer:
This issue scopes out a subset of work identified in the issues https://github.com/catalyst-cooperative/pudl/issues/1189, https://github.com/catalyst-cooperative/pudl/issues/1162, https://github.com/catalyst-cooperative/pudl/issues/889.

EIA 860 Schedule 6 (Environmental Equipment and Environmental Association) reports data on the environmental quality control equipment, investments, regulatory standards and control strategies associated with particular boilers. Much of this information will need to be organized outside of our existing EIA860 data structure, as multiple pieces of control equipment may be associated with multiple boilers. More immediately, however, data associated with a boiler ID can be integrated into the existing harvesting process. To bring boiler-associated attributes from these two datasets into PUDL, this PR incorporates two new tables into `boilers_eia860` and `boilers_entity_eia`: `emissions_control_strategies` and `boiler_info` (from 6.2 Environmental Equipment). Since all columns in both tables match 1:1 with boilers, they are entirely harvested into these two tables and do not separately appear in the final ETL output.

Notable decisions:
- Dropped `mercury_emissions_control` boolean (old boolean column whose information is contained in other columns).
- Converted the many boolean mercury equipment categories (2009-2012) from `emissions_control_strategies` into codes in the `mercury_existing_strategy` columns. Newer data has three of these columns but as many as six pieces of equipment are found in older columns, so there are now 6 of these columns.
- Where columns get less specific over time, I've for now merged the older columns to match the latest set of codes. The strategy codes (e.g. `pm_control_out_of_compliance_strategy_1`) get much less granular over time. This is most notable for the SO2 codes, with the addition of an "Use flue gas desulfurization unit or other SO2 control process (specify the specific type of equipment in Schedule 6A)" category for all equipment-based emission control strategies that likely links to one or more fields in the environmental equipment dataset. For now I've included all the codes that aren't easily mapped to the present set of codes so as to not lose this data. Eventually we'll want to more thoroughly reconfigure these codes to be consistent across time when the equipment data is integrated.
- Where columns get more specific over time, I've for now added the older more generic columns into the encoding tables. 
- Renamed `operating_date`, `retirement_date` and related columns to refer specifically to generators and make clear the distinction between generator dates and boiler dates. This needed to be updated in many places, including tests, which is why this PR has so many file changes!
[...]

PR #2664, opened 2023-06-12
Tweaks to emissions control equipment table

zaneselvans:
# PR Overview

Addresses the issues identified in #2663:
* converts emission control equipment cost to dollars from thousands of dollars
* adds a coding table to decipher the emissions control equipment type column.

# PR Checklist


src/pudl/metadata/codes.py
```diff
@@ -352,6 +352,95 @@
...
+                ("TR", "wet_scrubber_tray", "Tray type (wet) scrubber"),
+                ("VE", "wet_scrubber_venturi", "Venturi type (wet) scrubber."),
+            ],
+        ).convert_dtypes(),
+        "code_fixes": {
+            "SR-2": "SR",
+            "sn": "SN",
+            "LN": "LNB",
+        },
+        "ignored_codes": ["HRSG1", "HRSG2", "FGD", "OV"],
```
aesharpe, line 442: Why are these ones ignored? Are they not specified in the EIA documentation?

aesharpe (approved):
This looks good to me! Some of those equipment names are pretty funny.

aesharpe:
@zaneselvans I'm having some alembic issues when I try and run this locally. Not 100% sure if that is a me problem or related to this PR.

src/pudl/metadata/codes.py
e-belfer: FGD is a pre-2013 code for "flue gas desulfurization" (see the `mercury_compliance_strategies_eia` coding table). 
OV is "overfire air" (see the `nox_compliance_strategies_eia` coding table).

e-belfer: I think there's a whole bunch of notes about this in the original PR for those encoders #2319. Some codes get more or less specific over time, and sometimes where the details get reported changes over time. It'd be wise for us to do some bigger reconciling across all of these columns and I think this is even identified as a potential step in #2338.

You've double-checked that the layout codes haven't changed over time (esp pre-2013), correct?

^ If so, I think leaving this as-is is fine but we should note explicitly in #2672 that we should revisit codes that have gotten dropped.

zaneselvans: Sigh.. no I was lazy and just looked at the 2021 data. I'll check out the older ones and see how they map to each other.

zaneselvans: Okay, I've searched back through several different Layout spreadsheets (including pre-2013) to see if FGD or OV were ever valid codes for the environmental control equipment, and they never seem to have been.  In addition, FGD is showing up in 2018 and OV is showing up in 2013, and those codes definitely weren't valid equipment types in those years, so I think they are data entry errors and can't be cleanly mapped to any particular equipment type, so they should probably be dropped.

Unfortunately I don't think we'll be able to fill in either of these codes via #2616 either, since the OV codes show up only in the first year of data for their emissions unit, and the FGD code shows up sandwiched between two other valid codes which are different from each other in earlier / later years.

I've updated #2672 to remind us to look at the dropped codes.

e-belfer: Great, I'm ok to drop them then. Thanks for looking into it.

PR #2669, opened 2023-06-14
Identify and link inter-table relationships when they occur within a dimension (e.g. `utility_type`)

e-belfer:
Addresses #2623. When tables have multiple dimensions (e.g. a column that breaks out rows by `utility_type`), sometimes the factoids in these dimensions should correspond to other factoids in other tables. However, these aren't linked in the metadata provided by FERC. Let's link them!

Here's my first attempt at compiling all of these factoids into a table:

| **factoid**                                                    | **table_name**              | **dimension** | **matching_factoid1**                                                                                                                                       | **table_name_factoid1**                   | **dimension_factoid1** | **matching_factoid2** | **table_name_factoid2** | **dimension_factoid2** |
|----------------------------------------------------------------|-----------------------------|---------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------|------------------------|-----------------------|-------------------------|------------------------|
| depreciation_utility_plant_in_service                          | utility_plant_summary_ferc1 | electric      | **depreciation_expense_excluding_adjustments (confirm)**                                                                                                    | electric_plant_depreciation_changes_ferc1 | N/A                    |                       |                         |                        |
| utility_plant_in_service_classified                            | utility_plant_summary_ferc1 | electric      | _??? This is broken out differently in the plant_in_service table, this and utility_plant_in_service_completed_construction_not_classified sum to 1 field._ | plant_in_service_ferc1                    | N/A                    |                       |                         |                        |
| utility_plant_in_service_classified_and_unclassified           | utility_plant_summary_ferc1 | electric      | electric_plant_in_service                                                                                                                                   | plant_in_service_ferc1                    | N/A                    |                       |                         |                        |
| utility_plant_in_service_completed_construction_not_classified | utility_plant_summary_ferc1 | electric      | _??? This is broken out differently in the plant_in_service table, this and utility_plant_in_service_classified sum to 1 field._                            | plant_in_service_ferc1                    | N/A                    |                       |                         |                        |
| utility_plant_in_service_experimental_plant_unclassified       | utility_plant_summary_ferc1 | electric      | experimental_electric_plant_unclassified                                                                                                                    | plant_in_service_ferc1                    | N/A                    |                       |                         |                        |
| utility_plant_in_service_plant_purchased_or_sold               | utility_plant_summary_ferc1 | electric      | electric_plant_purchased                                                                                                                                    | plant_in_service_ferc1                    | N/A                    | electric_plant_sold   | plant_in_service_ferc1  |                        |
| depreciation_expense                                           | income_statement_ferc1      | electric      | depreciation_expense                                                                                                                                        | depreciation_amortization_summary_ferc1   | total                  |                       |                         |                        |
| depreciation_expense_for_asset_retirement_costs                | income_statement_ferc1      | electric      | depreciation_expense_asset_retirement                                                                                                                       | depreciation_amortization_summary_ferc1   | total                  |                       |                         |                        |
| maintenance_expense                                            | income_statement_ferc1      | electric      | maintenance_supervision_and_engineering                                                                                                                     | electric_operating_expenses_ferc1         | N/A                    |                       |                         |                        |
| operating_revenues                                             | income_statement_ferc1      | electric      | electric_operating_revenues                                                                                                                                 | electric_operating_revenues_ferc1         | N/A                    |                       |                         |                        |
| utility_operating_expenses                                     | income_statement_ferc1      | electric      | operations_and_maintenance_expenses_electric                                                                                                                | electric_operating_expenses_ferc1         | N/A                    |                       |                         |                        |
|                                                                |                             |               |                                                                                                                                                             |                                           |                        |                       |                         |                        |

This PR should do the following:
- [ ] assigns `utility_type` to tables where it is implied by not explicitly named
- [ ] renames factoids to be more closely similar to their matches
- [ ] add these factoids into the explosion process, so that they're not getting duplicated.

Complications:
- Some of these relationships are 1:1. These can be renamed to be identified as identical, and should be compared during the explosion, with the duplicate field dropped.
- Some of these relationships are 1:2. E.g. `utility_plant_in_service_plant_purchased_or_sold` should correspond to the sum of `electric_plant_purchased` and `electric_plant_sold` in the `plant_in_service_ferc1` table. Here, perhaps adding a calculation rather than a rename makes most sense.
- Some of these relationships are 2:1. E.g. the `utility_plant_in_service_classified` and `utility_plant_in_service_completed_construction_not_classified` are probably equivalent to the `electric_plant_in_service_and_completed_construction_not_classified_electric` field in the `plant_in_service_ferc1` table, but these two lines provide detail not available in the nested table. Here, I'm not sure what's the best course of action so as not to lose detail.

# PR Checklist


e-belfer:
Made some updates to the table above, comparing all 1:1 values in the data directly.

`utility_operating_expenses` doesn't match to any category in `electric_operating_expenses_ferc1`, so I've removed it.
`maintenance_expense` values don't align with one consistent category from `electric_operating_expenses_ferc1`, so I've gone ahead and removed this match.

| factoid name | match count |
|---|---|
| transmission_maintenance_expense_electric | 32 |
| maintenance_of_general_plant | 7 |
| maintenance_of_miscellaneous_transmission_plant | 4 |
| maintenance_of_generating_and_electric_plant | 2 |
| other_power_generation_maintenance_expense | 2 |
| hydraulic_power_generation_maintenance_expense | 1 |
| distribution_maintenance_expense_electric | 1 |
| maintenance_supervision_and_engineering_electric_transmission_expenses | 1 |
| maintenance_of_station_equipment_transmission | 1 |

e-belfer:
@jrea-rmi As promised, I've compiled the table above. Everything that's in the first table is a factoid that is clearly identical between the two tables noted and can be de-duplicated during the explosion process. These factoids are all confirmed to be correctly matched. The second table documents some more complex relationships between factoids that we could explicitly check without replacing any of these factoids, or could just note, depending on what you all would find most useful.

Note that we're actually currently missing a 1:1 link between the `income_statement_ferc1` and `electric_operating_expenses_ferc1`: maintenance and operation costs are split out in the `income_statement_ferc1` table, but `electric_operating_expenses_ferc1` only has a total maintenance+operations column, rather than a total maintenance and a total operations column. This shouldn't have a direct impact on the columns that appear in the final exploded table, but it does mean these values won't get checked against one another unless we write another factoid & calculation into the metadata.

Let me know if there are any 1:1 matches that are obviously missing for the `electric` columns and the corresponding electric tables.

jrea-rmi:
Great compilation @e-belfer!

A couple of comments on your complications:

- We are comfortable with different factoids for different utility_types, so would choose to explode the 1:2 relationships. But the specific differentiation of plant purchased vs. plant sold is probably not critical, so we'd consider it acceptable to lose that detail if you really don't want to explode it.
- Have you found the calculated field `electric_plant_in_service_and_completed_construction_not_classified_electric` to be correct generally? In the dbf form, I remember it being common that the value of "completed construction not classified" from the summary of utility plant table was included in the electric plant in service table, which was confusing. Related, it's odd that the `plant_in_service_ferc1` table total field is `electric_plant_in_service_and_completed_construction_not_classified_electric` rather than just `electric_plant_in_service`, because there's no line that distinguishes the completed construction not classified, so that means that it's been included in one of the classified categories. But if it's classified within a technology category, that's useful additional information that we'd like to include. So I think the `utility_plant_in_service_classified` and `utility_plant_in_service_completed_construction_not_classified` fields should both get dropped and the detailed components from `plant_in_service_ferc1` should be kept in the exploded table.
- I think corrections would be useful to include, after comparisons, sign changes, & calculation fixes have the data to a state without systematic errors.
- Agree that the `depreciation_amortization_summary_ferc1` table is all electric, not total.
- So there's no 1:1 link for `electric_operating_expenses_ferc1`, and also no 1:m calculation to check that the components of `electric_operating_expenses_ferc1` add up to what's in `income_statement_ferc1`? An added calculation of `OperationExpense =sum of all operation expense lines from electric_operating_expenses_ferc1`, with corrections for mismatches, and the same for maintenance expenses, sounds like a necessary addition.

jrea-rmi:
An added consideration for the complicatios where you're worried that exploding loses detail with `electric_plant_in_service_and_completed_construction_not_classified_electric`, I think it's important that there's another level of explosion from that total that gives us the detail of electric plant in service by ferc class / technology type, so ultimately including that explosion adds rather than loses detail.

e-belfer:
Attempted plan of action:
- Drop both `totals` factoids and only keep sub-components.
- We assume the root/parent table factoid is the 'correct' one
- With correction fields, if there is already a correction field from within a table for a 1:1 factoid, we will modify it to make the correction refer to the parent table, not the intra-table total. This should happen before the intra-table total factoid gets dropped. Otherwise we'll add a separate correction line.
- We'll treat 1:1 relationships as calculations w/ one component (but where?).
- Add calculation metadata upstream in the same place we already are (table transforms), as they'll get ignored during the first round of intra-table calc checks. We'll add to the parent table.

e-belfer:
A rough draft for the income statement table works up until the stage of making corrections. Further problems with the balance_sheet_assets_ferc1 table:

- `utility_plant_in_service_classified_and_unclassified` is already a calculated field. Adding a component will merely make the calculations incorrect. This table should actually have two separate calculations, but we don't currently have a way to do this.
- Need to replace the `depreciation_changes` table with components from the `depreciation_functional` table.

e-belfer:
Another oddity found while checking the merges for errrors. It seems like there may be utilities that are filling out some tables but not others, particularly the `plant_in_service_ferc1` table and not the `utility_plant_summary_ferc1` table. When we aggregate components for the electric plants purchased and sold there are a large number of records that do not have a corresponding reported field to compare to. See screenshots below for a sample. These mismatches are present both in our cleaned data and in the exploded table, so if they originate from our ETL and not from the raw data they are at least not introduced in the explosion process.

E.g. 1994 electric records from the `utility_plant_summary_ferc1` table:
image

Compared to a subset of the calculated fields from the `plant_in_service_ferc1` table for the same year:
image

src/pudl/transform/ferc1.py
```diff
@@ -1868,6 +2064,62 @@ def apply_xbrl_calculation_fixes(
...
+                            "weight": 1.0,
+                            "source_tables": [
+                                "electric_plant_depreciation_functional_ferc1"
+                            ],
+                            "utility_type": "electric",
+                            "subdimension": "in_service",
+                        },
+                    }
+                ],
+                # "utility_plant_in_service_classified_and_unclassified": [
```
e-belfer, line 2082: This is actually incorrect, delete it.

src/pudl/transform/ferc1.py
```diff
@@ -2136,6 +2136,62 @@ def apply_xbrl_calculation_fixes(
...
+                        "calc_component_to_replace": {
+                            "name": "utility_plant",
+                            "weight": 1.0,
+                            "source_tables": ["balance_sheet_assets_ferc1"],
+                        },
+                        "calc_component_new": {
+                            # Only used in pre-2004 calculations, aggregate of later sub-components.
+                            "name": "utility_plant_construction_work_in_progress",
+                            "weight": 1.0,
+                            "source_tables": ["utility_plant_summary_ferc1"],
```
e-belfer, line 2175: Possibly this should be "utility_type: total"?

src/pudl/transform/ferc1.py
```diff
@@ -2192,6 +2192,78 @@ def apply_xbrl_calculation_fixes(
...
+                "accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility": [
+                    {
+                        "calc_component_to_replace": {
+                            "name": "depreciation_utility_plant_in_service",
+                            "weight": 1.0,
+                            "source_tables": ["utility_plant_summary_ferc1"],
+                        },
+                        "calc_component_new": {
+                            # Only used in pre-2004 calculations, aggregate of later sub-components.
+                            "name": "accumulated_provision_for_depreciation_amortization_and_depletion_of_plant_utility",
```
e-belfer, line 2229: Possibly this should be "utility_type: total"?

e-belfer:
Updated notes from discussion with @jrea-rmi:

For the plant in service and utility plant summary link:
- If there is more than one calculation for a dimension, pick the one that leads to the leafier set of nodes.
- For the plant in service table, try to see whether Line 100 (Account 101+106) is equal to the sum of all the subtotals in the table. If not, see if adding the `utility_plant_in_service_completed_construction_not_classified` field from the utility plant summary table fixes the issue. Most likely, combine the utility plant summary fields into one factoid which is equivalent to the 101 + 106 field in the plant in service table, which itself is equivalent to a sum of the sub-totals.
-- One complication here: it seems like the sub-totals are generally equivalent to or larger than the 101+106 line.

jrea-rmi:
I finally got my old notebook open, and found that in previous work, it was common for utilities to include the value of `property_under_capital_leases` from the summary of utility plant table in the values in the electric plant in service table. My list of `utility_id_ferc1_dbf` that did this is: 
[2,9,10,12,30,43,51,57,72,77,96,99,101,112,126,128,136,137,141,142,143,147,159,175,177,182,188,194,255,258,268,274,297,403,443,516]. That may be the majority of the 1/3 of cases that we expect to have a correction of the same magnitude.

Almost all utilities also included `completed_construction_unclassified` in the electric plant in service table. I had only one exception, `utility_id_ferc1_dbf 213`.

There was one case of a utility including `held_for_future_use` in the electric plant in service table, `utility_id_ferc1_dbf 60`.

on the depreciation side, I also had two cases of including depreciation of held_for_future_use from the summary of utility plant table in the accumulated provision for depreciation table, `utility_id_ferc1_dbf 129 and 150`. And a couple cases of including amortization_of_other_utility_plant, `utility_id_ferc1_dbf 2, 227, 274`.

e-belfer:
The failure in the balance sheet liabilities table comes from the XBRL data, where retained earnings is always reported either in the balance sheet liabilities table or in the retained earnings table, but never in both. This results in `abs_diff` values that are always NaN, producing a 'division by 0' error. For now I'll aim to fix this by adding another assertion in the reconciliation function, but we should do a fix upstream to make sure we aren't losing values here.

cmgosnell (approved):
this is a huge improvement and because we are still actively editing some of these funcs/methods/calc fixes & this is being merged into a feature branch - not dev - i think this is ready to be merged. at the very least to have us all stop jumping off a gazzillion lil branches.

PR #2681, opened 2023-06-19
Fix Alembic migration diversions in dev

e-belfer:
Looks like we simultaneously pushed two different alembic migration histories into dev and broke the nightly build. Let's fix it!

# PR Checklist

### DIFF
--- a/migrations/versions/e608f95a3b78_drop_amount_from_retained_earnings.py
+++ b/migrations/versions/e608f95a3b78_drop_amount_from_retained_earnings.py
@@ -1,8 +1,8 @@
-"""Remove amount column from retained earnings
+"""drop amount from retained earnings
 
-Revision ID: cd70aa9ce4e9
-Revises: 88d9201ae4c4
-Create Date: 2023-06-13 09:18:41.174833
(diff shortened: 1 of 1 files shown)


zaneselvans:
I'm confused about why we don't catch these things. I think the last branch that got merged into `dev` on Friday was my update to the dependencies of `ferc-xbrl-extractor` and [the `tox-pytest` action on that PR passed](https://github.com/catalyst-cooperative/pudl/actions/runs/5295345971). If there's a problem with the Alembic migration history, why would the CI pass but the nightly build fail? Is the nightly build doing something fundamentally different than the CI with regards to the DB schema?

e-belfer:
One possibility is that you and I both merged branches that passed in reference to the same `dev` but yours changed the schema entirely and mine was based on the existing one.

jdangerx:
The nightly build is running alembic upgrade while I think CI doesn't do that. @rousik and I have talked about not running alembic in the nightly build, also - I think it should be pretty easy to do that. Though, if alembic is going to be part of our dev flow, maybe we want alembic to run in CI as opposed to not running it at all.

PR #2697, opened 2023-06-26
ferc1 :boom: rate base tags

cmgosnell:
# PR Overview
- [x] match as many of rmi's existing tags as possible semi-automatically (w/ rename dicts and row-literals)
- [x] make draft CSV of tags w/ matched and unmatched tags. Draft has two temporary columns (`xbrl_factoid_rmi`  and `xbrl_factoid_rmi_renamed`) that are necessary for manual mapping.
- [ ] test tags? (or maybe this is a @zaneselvans task over in the forest branch?)
- [x] match renaming tags manually
- [ ] remove two temporary RMI columns
- [ ] reduce the factoids to their root-y factoids (i think these factoids are the "leafy" factoids. if a set of leaves that have a common root all have the same tag, just hat root factoid can be labeled in this set of saved tags
- [ ] are there any factoids from the 💥 tables that we need to add?

## oos
- once we settle on a set of tags, we need to be able to quickly know if something in the transforms or the data itself has changed which results in different names for the `xbrl_fatoid`s. so we'll need to add tests to make sure this stored set of tags is compatible with the data.

# PR Checklist


zaneselvans (commented):
Since this isn't yet being referenced anywhere, and the place where it would be nice to have it to work with is the `xbrl-calculation-tree` branch, is there any harm in going ahead and merging it in as a working draft that we'll probably update later?

zaneselvans (approved):

PR #2701, opened 2023-06-28
fix `source_tables` definition

cmgosnell:
# PR Overview
I was previously assuming that if a calculation component showed up in the source_table of the calculation, than i didn't need to check if that component also showed up in other tables.

That was wrong! There are instances like for factoid `utility_plant_and_construction_work_in_progress` which shows up in multiple tables and has calculations components like `construction_work_in_progress` that show up in multiple tables.

# PR Checklist

### DIFF
--- a/src/pudl/transform/ferc1.py
+++ b/src/pudl/transform/ferc1.py
@@ -54,6 +54,37 @@ def clean_xbrl_metadata_json(
     return add_source_tables_to_xbrl_metadata(raw_xbrl_metadata_json)
 
 
+def all_fields_in_table(table_meta) -> list:
+    """Compile a list of all of the fields reported in a table."""
+    return [field["name"] for meta_list in table_meta.values() for field in meta_list]
+
+
+def extract_tables_to_fields(raw_xbrl_metadata_json: dict) -> dict[str, list[str]]:
+    """Compile a dictionary of table names (keys) to list of fields.
+
+    Args:
+        raw_xbrl_metadata_json: dictionary of xbrl metadata.
+    """
+    return {
+        table_name: all_fields_in_table(table_meta)
+        for table_name, table_meta in raw_xbrl_metadata_json.items()
+    }
+
+
+def label_source_tables(calc_component: dict, tables_to_fields: str) -> dict:
+    """Add a ``source_tables`` element to the calculation component."""
+    calc_component["source_tables"] = [
+        table_name
+        for table_name, fields in tables_to_fields.items()
+        if calc_component["name"] in fields
+    ]
+    # weirdly there are a number of nuclear xbrl_factoid calc components that seem
+    # to have no source_tables.
+    if not calc_component["source_tables"]:
+        logger.debug(f"Found no source table for {calc_component['name']}.")
+    return calc_component
+
+
 def add_source_tables_to_xbrl_metadata(
     raw_xbrl_metadata_json: dict[str, dict[str, list[dict[str, Any]]]]
 ) -> dict[str, dict[str, list[dict[str, Any]]]]:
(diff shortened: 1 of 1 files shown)


src/pudl/transform/ferc1.py
```diff
@@ -1851,15 +1851,15 @@ def apply_xbrl_calculation_fixes(
                         "calc_component_to_replace": {
                             "name": "nuclear_fuel_materials_and_assemblies",
                             "weight": 1.0,
-                            "source_tables": ["balance_sheet_assets_ferc1"],
+                            "source_tables": [],
```
zaneselvans, line 1854: What does it mean to have gone from having 1 source table to zero source tables? I thought they all had to have at least one.

src/pudl/transform/ferc1.py
```diff
@@ -54,6 +54,37 @@ def clean_xbrl_metadata_json(
...
+    calc_component["source_tables"] = [
+        table_name
+        for table_name, fields in tables_to_fields.items()
+        if calc_component["name"] in fields
+    ]
+    # weirdly there are a number of nuclear xbrl_factoid calc components that seem
+    # to have no source_tables.
+    if not calc_component["source_tables"]:
+        logger.debug(f"Found no source table for {calc_component['name']}.")
+    return calc_component
```
zaneselvans, line 85: This seems fishy. Is there a rename issue somewhere maybe? Do they not show up in the XBRL taxonomy anywhere?

src/pudl/transform/ferc1.py
```diff
@@ -54,6 +54,37 @@ def clean_xbrl_metadata_json(
     return add_source_tables_to_xbrl_metadata(raw_xbrl_metadata_json)
 
 
+def all_fields_in_table(table_meta) -> list:
+    """Compile a list of all of the fields reported in a table."""
+    return [field["name"] for meta_list in table_meta.values() for field in meta_list]
+
+
+def extract_tables_to_fields(raw_xbrl_metadata_json: dict) -> dict[str, list[str]]:
+    """Compile a dictionary of table names (keys) to list of fields.
```
zaneselvans, line 63: What is a field in this context? Is it a fact / factoid?

zaneselvans (commented):
The no source table thing seems... wrong? Is there some way we could be missing them?

cmgosnell:
I'm a lil confused bc I'm getting this error:

`AssertionError: Calculations in electric_plant_depreciation_changes_ferc1 are off by 0.48723232685243256. Expected tolerance of 0.4.`

but i couldn't immediately find out why. i'm sure it'll avail itself to me.

src/pudl/transform/ferc1.py
cmgosnell: to me it seems like the xbrl metadata includes a field that does not actually exist. The way i was doing the source_table before (unintentionally) defaulted to just assigning the source_table as the native table.

src/pudl/transform/ferc1.py
cmgosnell: these have not been renamed at all at this point. and yes to the latter that definitely happens sometimes. as in there are references in the calcs to fields that do not exist i just moved these little internal functions out of the main label function, i didn't edit them at all.

src/pudl/transform/ferc1.py
```diff
@@ -1851,15 +1851,15 @@ def apply_xbrl_calculation_fixes(
...
                             "name": "nuclear_fuel_materials_and_assemblies",
                             "weight": 1.0,
-                            "source_tables": ["balance_sheet_assets_ferc1"],
+                            "source_tables": [],
                         },
                         "calc_component_new": {},
                     },
                     {
                         "calc_component_to_replace": {
                             "name": "spent_nuclear_fuel",
```
e-belfer, line 1860: Is this actually meant to be `spent_nuclear_fuel_major_only` which also comes from `f1_comp_balance_db`? Possibly OOS for this issue.

e-belfer, line 1860: Is this not `spent_nuclear_fuel_major_only` from the `f1_comp_balance_db`?

cmgosnell:
i'm closing this bc we decided that the current behavior is actually okay. if a calculation component shows up in its native table the `source_tables` will be the native table and it will not keep searching for additional instances of the factoid. we decided not to move forward on this now bc a lot of the explosion tooling assumes or desires there to be one source table for calc components bc if there are two instances then.... well which one do we choose?!? they are often instances of the same factoid which i think over time we will link directly. but for now we are going to keep the setup the way it is.

obviously if we want to revisit this we can come back and reopen this lil guy.

Issue #1393, opened 2022-01-07
utility "owners" are not harvested into normalized EIA utility table


cmgosnell:
There are about 1000 utilities that are currently not captured in the `utilities_eia` table through the harvesting process because they only show up as "owners" in the `ownership_eia860` table.

```python
# with a pudl_out object
own = pudl_out.own_eia860()
utils = pudl_out.utils_eia860()

len(own[~own.owner_utility_id_eia.isin(utils.utility_id_eia.unique())].owner_utility_id_eia.unique())
1009
```

This is causing me trouble because I am using the ownership table in the EIA plant-part list to generate plant-part records for each owner. This results in ~3% of the plant-part records not having `utility_id_pudl` because I was assuming that all of the utilities were in `utilities_eia`. *I think* these owner records do have `utility_id_pudl` because we are capturing them in the PUDL id mapping process - but they are not actually linked to them because this link happens through the `utilities_eia` table.

zaneselvans:
I feel like we've come across something like this before.

This seems like a specific case of a more general issue that I suspect we'll run into other places -- that it's possible to have multiple columns in a table which have foreign keys referring to the same other table -- because the same entities can have multiple different kinds of relationships with each other. So somehow in the normalization process we need to be able to indicate which columns are associated with what entities, even when they don't have the canonical column names. But then we'll also have to deal with the other attribute columns in those tables, and identifying which of the multiple PK columns they're supposed to be associated with.

Also, doesn't the original `utilities_eia860` table come from a spreadsheet tab that lists all of the utilities and indicate whether they are reporting as owners or operators? I would have thought that would catch all of these.  Maybe that's a recent development? Looking at the distribution of missing IDs across years, it seems like it starts being a problem around 2011, and ramps up from there. I wonder if these utilities are mostly part owners of renewable plants?

cmgosnell:
In this case I actually don't think it is that messy of a PK/FK relationship

```
utilities_pudl
  * utility_id_pudl
      |
      V
  utilities_eia
    * utility_id_eia
    * utility_id_pudl (FK from utilities_pudl)
       |
       V
    ownership_eia860
      * utility_id_eia (FK from utilities_eia.utility_id_eia)
      * owner_utility_id_eia (FK from utilities_eia.utility_id_eia)
```

Codifing the need to harvest both the `utility_id_eia` and the `owner_utility_id_eia` seems potentially challenging.

[...]

Issue #509, opened 2020-01-18
Do not remove harvested fields during transform


zaneselvans:
Currently we are in many cases dropping columns from the EIA 923 data in the transform step that *would* feed into the entity harvesting if they were left in place... which means we're missing that data. In the course of moving to the object oriented transformation process, need to make sure that we stop doing that!

zaneselvans:
@cmgosnell and I are going to help get Kurt Nordback working on this issue as a way to become more familiar with the harvesting process, working with our code, Jupyter, etc. But first he needs a personal GitHub account!

zaneselvans:
@cmgosnell while talking over some of these fields with @knordback yesterday, I noticed that the `associated_combined_heat_power` field is part of the `generators_entity_eia` table, but there's another `combined_heat_power` field being reported in e.g. the `generation_fuel_eia923` table, and looking at the spreadsheets, it seems like that field pertains to the plant (which makes some sense given that `generation_fuel_eia923` is reported on a date, plant, prime-mover, fuel basis).

Are these different attributes? Should there be a CHP field at both the generator and the plant level? Should this really be a permanent attribute, or is it another one that changes slowly? Does the generator field really just indicate that the generator is part of a plant that does CHP?  Or that it's part of a generation unit that does CHP? Could the plant or plant-prime-fuel level CHP status be inferred from the generator-level CHP attributes?

Right now we're discarding the CHP column reported in `generation_fuel_eia923`.

@grgmiller or @gschivley do either of you have more context on the relationship between these two different CHP fields?

cmgosnell:
I don't know exactly. `associated_combined_heat_power` originates in the generator table. I would not be surprised if there were plants that had some units contributing to a CHP and some that just generated power. I don't think it's generally a good idea to base any logic about the workings of a plant based off of the reporting structure of the `generation_fuel_eia923` table. I personally would check whether this value is actually consistent across all generators within a plant before thinking about moving it. But also i could definitely imagine this changing over time (albeit very rarely!).

zaneselvans:
It seems like we should probably do an exhaustive check of all the currently "permanent" generator attributes on the pre-harvested dataframes... and see how permanent they actually are.

grgmiller:
I do not have any context on these two fields.

knordback:
I'll hold off on this one for now.

knordback:
[...]

PR #2714, opened 2023-07-05
Dagsterize `mega_generators` and `plant_parts_eia`

katie-lamb:
# PR Overview

# PR Checklist


katie-lamb:
@zaneselvans @cmgosnell as briefly discussed on Slack, the primary issue with getting these tables in the DB is that `mega_gens` has null values in `utility_id_eia` which is part of its primary key. The null values arise because the utility owner IDs are not properly harvested (see #1393 ). I see 3 potential solutions, ordered from easiest and least satisfactory to hardest and most satisfactory:

- Don't write the mega gens table to the DB. Just put `plant_parts_eia` in there. How useful is the mega gens table as a standalone output? Or is the PPL really the more useful output?
- The "janky" solution that Zane proposes [here](https://github.com/catalyst-cooperative/pudl/issues/1393#issuecomment-1007618245), where we rename `owner_utility_id_eia` to `utility_id_eia` and `utility_id_eia` to `operator_utility_id_eia` (then the ownership relationship would be captured / harvested, and we could drop the `operator_utility_id_eia` column from the table altogether). I think we're talking about the original `utilities_eia860` table here.
-  Rework the harvesting process slightly: "it should be possible to harvest selectively based on name and the foreign key relationships. Like any column that is already identified as referring to an entity table column (like `utility_id_eia`) even if it doesn't have exactly the same name as the column it refers to (e.g. `owner_utility_id_eia`) should still be considered a source for harvested values."

I'm not entirely sure where we stand on the harvesting process (#509  seems relevant). I guess my vote would be option 1 or 2 and then do option 3 when a more thorough harvesting revamp is conducted?

katie-lamb:
I tried a quick fix to rework the harvesting process to harvest the owner utilities and results seem pretty good. I was able to harvest the missing ~1300 utilities but mysteriously lost ~400 utilities. Investigating where they went.

zaneselvans:
#509 just retains columns that we can/should be harvesting using the current process.

I could be wrong, but I don't think that the current harvesting process is capable of harvesting the same entity ID / value (like `utility_id_eia`) from multiple columns in the same input dataframe, even though that's clearly a kind of data that can exist, since an entity can have a relationship with another entity of the same kind, as we see here with the owners and operators showing up in the same records.

We realized in retrospect that leaving the (operator) `utility_id_eia` in this table is not well normalized, since you can look up the operator based on the date and the plant ID.  So one simple solution would be to just drop the operator ID column prior to harvesting, rename the owner IDs to be `utility_id_eia` and then all those owner IDs would get harvested, but we wouldn't be capturing any unique utility IDs that appeared only in the operator ID column in this table (of which there should be none but... you never know).

I think the "automatically detect columns to harvest based on FKs" solution suggested in #1393 probably isn't the right approach, since we harvest IDs from columns that don't end up in the final table, and so would not have any FK relationships defined -- as **should** be the case for this table (since to normalize the ownership table, we should be removing the operator `utility_id_eia` column and only keeping the owner ID column... which should probably be renamed to `utility_id_eia` as per our convention more broadly).

So I think the simple solution here is just to preemptive drop the operator ID and hope we don't lose any actually novel utility IDs, and the more involved solution is to allow the specification of multiple columns that should be harvested, mapping them to a particular destination column.

How did your attempted rework approach it? Any idea what happened with the lost utilities?

katie-lamb:
@zaneselvans I had similar thoughts to what you said above. The column mapping solution seemed to work pretty well.

> How did your attempted rework approach it? Any idea what happened with the lost utilities?
Basically what you said above, where I added a dictionary to the EIA entities mapping from column names to the "standard" name of the column, i.e. `owner_utility_id_eia` to `utility_id_eia`. (Will push this after fixing a CI issue with the gens mega table) 

This added 1228 new utilities. The missing 400 IDs that I previously mentioned were just because I had forgotten to recreate the intermediate pickled assets with the new 2022 data, but all the IDs are showing up now.

There are some slight changes to the non-ID columns with the owner utilities included. 206/15301 records in the `utility_entity_eia` table have a different `utility_name_eia`, but for the most part it seems like superficial changes, i.e. "Banning, City of" vs. "City of Banning". This is a histogram of the string similarity/edit distance between these new and old names. Most of the records with 0 similarity are because one of the records is null. Most records with a score of > .4 still share a token or two. I've also included this for `street_address` in the `utilities_eia860` table of which there were 151/125280 rows changed. All the rest of the columns are unchanged except for 1 or 2 records with a changed state, city, zip code.

jaro_winkler_utility_name_sim
levenshtein_utility_name

jaro_winkler_street_address

levenshtein_street_address

src/pudl/metadata/resources/__init__.py
```diff
@@ -288,11 +288,30 @@
...
+
+Mapped columns allow for harvesting the same entity ID / value relationship
+from multiple columns in the same input dataframe. This is useful if an entity has a
+relationship with another entity of the same kind, for example owner and operator utilities
+showing up in the same ownerhip table records. The mapped column dictionary maps from
+column names of the second group of entity ID / value columns to the standard names
+for the entity ID / values columns.
+
+Mapped column dictionaries must include all columns that are to be harvested
+for that relationship, even if a column name maps to itself.
```
katie-lamb, line 314: I tried to explain the change in this docstring, but maybe this doesn't make sense. Not sure what the best explanation of this behavior is that would generalize to potential future scenarios.

zaneselvans: I'm not understanding the last bit:
> Mapped column dictionaries must include all columns that are to be harvested
for that relationship, even if a column name maps to itself.

katie-lamb: I think the example I was thinking of wouldn't actually come up, because it would require a table input that's not so well normalized/clean. I think I can take this part out, but this is what I was originally thinking:

- There are the columns `owner_utility_id_eia`, `utility_id_eia`, but the owner and operator utility share a column, maybe `state` or something. Maybe this example won't ever come up, but perhaps we're only looking at CA utilities so there's one state column shared between owner and operator
- In this case, you'd want to harvest `state` for both the owner and operator utility, but with the owner utility you don't need to rename the `state` column, so you map it to itself `state:state`.

katie-lamb:
> except to make sure that we get all of the `utility_id_eia` values that show up anywhere? I.e. can we just look at harvesting multiple ID columns and not need to worry about potentially finding multiple copies of entity attribute columns in a single table (although I can see that being more general, in cases where we have multiple entities being referenced in the same table, potentially each with their own associated attributes).

In my current implementation, all the multiple ID column from the same table relationships are defined, and there is no automatic searching for multiple copies of entity attribute columns in a single table. Rather than dropping `utility_id_eia` and potentially missing some utility IDs, I thought this was a more general solution that could be easily applied to different table. Christina also mentioned the goal of not dropping columns pre-harvesting.



> Is there any reason not to redefine the final ownership table here to get rid of the unusual names and the `utility_id_eia` column that refers to the operator (post-harvesting).

The post-harvesting table won't have a separate column name for owner utilities, just the `utility_id_eia` that has both owner and operator tables in it

src/pudl/analysis/ferc1_eia.py
```diff
@@ -107,7 +107,7 @@ def __init__(
             start_date: Start date that cooresponds to the tables passed in.
             end_date: End date that cooresponds to the tables passed in.
         """
-        self.plant_parts_eia = plant_parts_eia
+        self.plant_parts_eia = plant_parts_eia.set_index("record_id_eia")
```
katie-lamb, line 110: Unless I'm missing something, the table needs to be loaded into the DB with `record_id_eia` as a column, not the index, but the match expects `record_id_eia` as the index.

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -365,6 +366,55 @@
...
+) -> list[AssetsDefinition]:
+    """Build EIA plant parts table related assets."""
+
+    @asset(
+        name="mega_generators_eia",
+        ins={
+            "mcoe": AssetIn(key="mcoe_generators_yearly"),
+            "own_eia860": AssetIn(key="denorm_ownership_eia860"),
+        },
+        io_manager_key=None,
```
katie-lamb, line 380: Is this all that needs to happen so that this table isn't written to the DB?

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -787,20 +826,18 @@ def add_additonal_cols(self, plant_parts_eia):
...
-                report_year=lambda x: x.report_date.dt.year,
-                plant_id_report_year=lambda x: x.plant_id_pudl.astype(str)
-                + "_"
-                + x.report_year.astype(str),
-            )
-            .pipe(
-                pudl.helpers.cleanstrings_snake,
-                ["record_id_eia", "appro_record_id_eia"],
-            )
-            .set_index("record_id_eia")
```
katie-lamb, line 801: You can't write a table to the DB where the index is a non-numeric column, right? I ended up just making `record_id_eia` a column and then setting it to be the index in the FERC to EIA match

src/pudl/metadata/resources/__init__.py
```diff
@@ -288,11 +288,32 @@
...
+from multiple columns in the same input dataframe. This is useful if an entity has a
+relationship with another entity of the same kind, for example owner and operator utilities
+showing up in the same ownership table records. The mapped column dictionary maps from
+column names of the second group of entity ID / value columns to the standard names
+for the entity ID / values columns.
+
+Mapped column dictionaries must include key value pairs for all the column names
+that should be harvested for that relationship, even if the column doesn't need to be renamed.
+If a column doesn't need to be renamed, it should map to itself,
+i.e. `utility_id_eia: utility_id_eia`.
```
katie-lamb, line 316: I added this functionality because the EIA ownership table has both owner and operator utilities in the same table, and we want to harvest both the owners and operators into the post-harvesting utilities table. This docstring could maybe be improved to be more clear.

src/pudl/analysis/plant_parts_eia.py
zaneselvans: There's a default IO manager which will be used if this is None, which could be set explicitly instead.  But unless it's set to `pudl_sqlite_io_manager` it won't be written to the DB.

src/pudl/analysis/plant_parts_eia.py
zaneselvans: I could be wrong but I think the index column is ignored entirely when writing to the DB.  The schema for the table which is defined in the metadata needs to indicate what set of columns comprise the primary key if there is one.  If there's no natural primary key then I think it'll it'll use an autoincrementing integer key, which is what happens with the `fuel_receipts_costs` table(s).

src/pudl/metadata/resources/__init__.py
zaneselvans: Did we also want to remove the `utility_id_eia` column from `ownership_eia860` and rename `owner_utility_id_eia` with `utility_id_eia` since it's poorly normalized as it is, and the operator ID can/should be merged in based on the operator associated with the plant ID instead?  This would probably require an update to the denormalized `ownership_eia860` asset.

src/pudl/transform/eia860.py
```diff
@@ -158,9 +158,24 @@ def _core_eia860__ownership(raw_eia860__ownership: pd.DataFrame) -> pd.DataFrame
     state_to_country = {
         x.subdivision_code: x.country_code for x in POLITICAL_SUBDIVISIONS.itertuples()
     } | {"CN": "CAN"}
-    own_df["owner_country"] = own_df["owner_state"].map(state_to_country)
+    own_df["country"] = own_df["owner_state"].map(state_to_country)
```
zschira, line 161: Not a big deal at all, but I might leave this as `owner_country`, then rename below with all the other columns just for consistency.

src/pudl/metadata/resources/__init__.py
```diff
@@ -288,11 +288,28 @@
             "phone_number_2",
             "data_maturity",
         ],
+        "map_cols_dict": {
```
zschira, line 291: It seems like maybe `map_cols_dict` is no longer used?

zschira (changes requested):
Looks good! My biggest comments are that I'm not sure the assets need to be in a factory function, and I'm not sure how/if the `map_cols_dict` is being used.

src/pudl/metadata/resources/eia860.py
```diff
@@ -258,21 +258,20 @@
...
+                "zip_code",
                 "fraction_owned",
                 "data_maturity",
             ],
             "primary_key": [
                 "report_date",
                 "plant_id_eia",
                 "generator_id",
-                "owner_utility_id_eia",
+                "utility_id_eia",
```
zaneselvans, line 274: Since `utility_id_eia` in this table refers to the **owner** rather than the **operator** we may want to give it a resource-specific metadata override at the bottom of `pudl/metadata/fields.py` so the description highlights this difference.

src/pudl/transform/eia.py
```diff
@@ -299,6 +299,13 @@ def _compile_all_entity_records(
     id_cols = ENTITIES[entity.value]["id_cols"]
     static_cols = ENTITIES[entity.value]["static_cols"]
     annual_cols = ENTITIES[entity.value]["annual_cols"]
+    # A dictionary of columns representing additional data to be harvested,
+    # whose names should map to an ID, static, or annual column name.
+    (
+        ENTITIES[entity.value]["map_cols_dict"]
```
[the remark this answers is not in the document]
katie-lamb: Ha oops that's supposed to be assigned to something, must have messed with it for testing and not fixed it.

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -365,6 +366,54 @@
 ]
 
 
+def plant_parts_eia_asset_factory(
```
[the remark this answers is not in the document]
katie-lamb: Changed to stand alone assets.

katie-lamb:
@zschira For some reason the harvesting of the owner utilities got lost somewhere and weren't in the initial commits I pushed, so I just added that back in. Basically it uses the `map_cols_dict` to rename columns and harvest additional relationships from the tables.

I'm not sure why `test_nuclear_fraction` is failing. Somehow more nuclear gens were added? I'll poke into that later.

In addition to fixing that test, I need to update the release notes and merge dev.

Also @bendnorman I'm not sure if you want to get in here in the context of naming conventions. But tagging you just in case.

bendnorman:
I have a few questions about the mega_gens and ppl assets:
- Are there suitable natural keys for these tables?
- Does every generator in `mcoe_generators_monthly` exist in `mega_gens`?
- Eventually, should the ppl asset not live in the database if it's only used to link EIA and FERC plant parts? Is it a reasonable table for users to access?
- Is the `mega_gens` table the `mcoe_generators_monthly` table but split out by ownership fractions?

katie-lamb:
@zaneselvans [Here](https://github.com/catalyst-cooperative/pudl/actions/runs/6241340079/job/16943345900?pr=2714#step:11:5267) is the `test_nuclear_fraction` failure in case it seems familiar to you. Otherwise I'll compare the tables to `dev` and see what's changed.

zaneselvans:
It looks like the fraction of nuclear generation has just declined a little bit? Previously we were checking that it was 20% +/- 2% but now it seems to be at 17.75% which isn't sooo far off, and it looks like there was 1 big retirement in 2022, and another big retirement in 2021, so maybe that's all it is.  Here's the [NEI stats through 2021](https://nei.org/resources/statistics/us-nuclear-generating-statistics). And the failure here is just looking at 1 year of data, so it's going to be more volatile.  My intuition is that either reducing the expected value to 19 +/- 2 or expanding the allowed range to 20 +/- 3 would be reasonable and fix the issue.

katie-lamb:
@zaneselvans Ok great, that's along the lines of what I thought was happening, I was just suspicious of why this PR (adding ownership utilities) would change anything with the nuclear fraction.

I'll make try playing with range of the expected value.

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -365,6 +366,38 @@
...
+    mcoe_generators_yearly: pd.DataFrame, denorm_ownership_eia860: pd.DataFrame
+) -> pd.DataFrame:
+    """Create mega generators table asset."""
+    return MakeMegaGenTbl().execute(
+        mcoe=mcoe_generators_yearly,
+        own_eia860=denorm_ownership_eia860,
+    )
+
+
+@asset(
```
katie-lamb, line 383: It would be nice to be able to generate the PPE for only a subset of years, how have we been doing that type of thing? Add `context` with a start and end year? Ideally you'd be able to run the fast ETL and only get those 2 years of PPE.

zaneselvans: What happens if you try and generate the PPE in the fast ETL right now? It'll only have the last couple of years of data available in the database to work with won't it?

katie-lamb: I guess I mean if you have full data in the database (with all the years) but you're making changes to the PPE and want to generate a new PPE with a couple years of data but don't want to wait the full 45 min

zschira (approved):
This looks good to me once the tests are passing, and @bendnorman's questions are addressed

src/pudl/analysis/plant_parts_eia.py
zaneselvans: Ah, okay. Hmm. A 45 minute asset that runs at the end of the DAG is going to be a bottleneck too.  I wonder if there's anything we can do to split it up or speed it up to avoid extending the whole ETL by 45 minutes and also make it quick and easy to regenerate the whole thing on the fly in development.  I'm not sure how to twiddle input parameters like this on a per-asset per-run basis but I assume it's possible.

zaneselvans:
Unfortunately @bendnorman I think these are probably @cmgosnell questions...

> Are there suitable natural keys for these tables?
> Does every generator in mcoe_generators_monthly exist in mega_gens?
> Is the mega_gens table the mcoe_generators_monthly table but split out by ownership fractions?

I don't know the answer to any of these.

> Eventually, should the PPE asset not live in the database if it's only used to link EIA and FERC plant parts? Is it a reasonable table for users to access?

I suspect the answer here is NO -- this table is incredibly difficult to work with and explain to users, and I think once we've created a much simpler join table that allows the FERC and EIA data to be merged, that's the output that should end up in the DB and be used to pull the FERC financial info into our huge generator attributes table.

src/pudl/analysis/plant_parts_eia.py
katie-lamb: It probably wouldn't be hard to parallelize. Currently it loops through each plant part and aggregates records to that part level, then concatenates into one dataframe. This loop could be parallelized. But probably worth making that a separate issue/PR at this point.

src/pudl/transform/eia.py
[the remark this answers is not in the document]
katie-lamb: Ah right! Good call!

src/pudl/analysis/plant_parts_eia.py
zaneselvans: Agree, sounds like a good follow-up issue. Too many 45-minute assets will definitely get frustrating!

PR #2715, opened 2023-07-06
`utility_plant_summary_ferc1` to `plant_in_service_ferc1` link

cmgosnell:
# PR Overview
I added an aggregated factoid into the `utility_plant_summary_ferc1` table for two factoids that *when combined* link up to a record in the `plant_in_service_ferc1` table.

RN I added a calculation that is testing this linkage within the table. But what we what is to point this new factoid at the line in the pis table as a new pass through calc.

We can *either* do that by:
*  just having one calculation that links the electric portion of this record up to the PIS table - effectively stranding the non-electric break out of `utility_plant_in_service_classified` and `utility_plant_in_service_property_under_capital_leases`. So we'd get gas/other/etc "leaves" (most detailed factoids) as this new aggregated record and the electric leaves would keep going up along the chain into the pis table.
* OR we can make *all* calculations aware of *all* of the `utility_type`/other dimensions that they are relevant to. this would result in effectively having two separate calculations: one for electric which points the pis table and one for gas/other/etc that points to the two components within the summary table

# PR Checklist

### DIFF
--- a/src/pudl/transform/ferc1.py
+++ b/src/pudl/transform/ferc1.py
@@ -13,7 +13,6 @@ import json
 import re
 from collections import namedtuple
 from collections.abc import Mapping
-from io import StringIO
 from typing import Any, Literal, Self
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/pudl/transform/ferc1.py
```diff
@@ -4840,10 +4841,101 @@ class UtilityPlantSummaryFerc1TableTransformer(Ferc1AbstractTableTransformer):
...
+        new_fact = pd.read_csv(
+            StringIO(
+                f"""xbrl_factoid,calculations,balance,ferc_account,xbrl_factoid_original,intra_table_calc_flag,row_type_xbrl
+{new_factoid_name},[],debit,,{new_factoid_name},True,calculated_value
+"""
+            )
+        ).convert_dtypes()
+        # I couldn't figure out how to format the csv version above to get it in the
+        # first time so i'm adding it down here.
+        new_fact.loc[0, "calculations"] = json.dumps(calc)
```
zaneselvans, line 4873: This section feels hard to read to me. This seems to yield equivalent results:

```py
calc = [
    {
        "name": factoid,
        "weight": 1.0,
        "source_tables": [self.table_id.value],
    }
    for factoid in factoids_to_agg
]

new_fact = pd.DataFrame({
    "xbrl_factoid": [new_factoid_name],
    "calculations": [json.dumps(calc2)],
    "balance": ["debit"],
    "ferc_account": [pd.NA],
    "xbrl_factoid_original": [new_factoid_name],
    "intra_table_calc_flag": [True],
    "row_type_xbrl": ["calculated_value"],
}).convert_dtypes()
```

zaneselvans (changes requested):
I found the CSV-based dataframe compilation kinda hard to read so I offered an alternative.

With the two options for connecting UPS and PIS, if I understand correctly we can't do option 2 until we integrate other dimensions into what other thing? The UPS table must already understand the multiple dimensions, and we can add (or have added?) the electric utility type to the PIS table already right?

src/pudl/transform/ferc1.py
cmgosnell: yeah this looks way nicer ty for the suggestion

cmgosnell:
> With the two options for connecting UPS and PIS, do we have to integrate the other dimensions into all calculations, or just the ones for these two tables, in order to get the calculation validation stuff working? The UPS table must already understand the multiple dimensions, and we can add (or have added?) the electric utility type to the PIS table already, right?

Option 1 is definitely the way. The calcs/exploders are now cognizant of multiple dimensions. But we don't yet have a case where we have a calculation for one set of dimensions and a different one for another set of dimensions. So choosing option 1 enables us to avoid building in this optionality.

the tests aren't passing rn but this is bc of a problem on the underlying branch. I've tested these updates locally i would suggest for the sake of you @zaneselvans  getting this pulled into the 🌳  would that we merge it before the tests actually pass (its a key error in the income statement table I plan an fixing it straight away).

zaneselvans (approved):

Issue #2709, opened 2023-07-03
Extend calculation trees to work with additional dimensions


zaneselvans:

zaneselvans:
Dupe of #2736

Issue #2688, opened 2023-06-22
Standardize corrections and treatment of sub-totals into the ferc1 table transforms


cmgosnell:
we are checking the subtotal's in the inter-dimensional 👾  totals within a table inside of `reconcile_table_calculations` when any `subtotal_column` is provided.  but we are not currently adding any corrections.

Fix some infrastructural issues:

```[tasklist]
- [x] Move sub-total to total calculation component generation into transform step and update to check using same method as for everything else.
- [x] Possibly streamline calc in transform and in explosion to use the same method (e.g. generate_intertable_calcs and calculate_values_from_components)
- [x] add correction for inter-dimensional subtotals with `subtotal_column` == "correction" or some such other way to identify the correction.
- [x] Check whether we renamed any `totals` sub-total names, as this will break our existing method.
- [ ] integrate removal of `totals` rows into pruning of exploded tables (this got done in some other pr!
```

PR #2721, opened 2023-07-11
Make a calculation component table and use in the inter-table/inter-dimension checking

cmgosnell:
# PR Overview
There are two main things going on in this PR:
* Make a calculation component table as a dagster asset
* Use the calculation component table

The calculation component table is made from a small method that is inside of the default table transformer which build a table-specific df of calculation components that get cobbled together in a big calc component table. Right now this is not used anywhere inside of the transform process, but the plan is to use this within the calculation fixing process in #2605 .

This calculation component table is then used within the explosion step. It is currently being used within `generate_intertable_calculations` but not within the tree tooling. There is a little bit of leftover functions in here that enable the tree tooling to still work using the old calcs from the json embedded within the table level metadata. That will all be swapped to use the new calc component table with #2709 .

```[tasklist]
- [x] draft tbl!
- [x] rename subdimensions in calc fixes to their og names
- [x] add in implicit dimensions
- [x] integrate this calc component table into inter-table-dim checks over in the explosions land
```

# PR Checklist


src/pudl/transform/ferc1.py
```diff
@@ -2731,6 +2731,24 @@ def remove_nones_in_list(list_of_things):
...
+            return pd.DataFrame()
+
+        calc_dfs = []
+        for calc, factoid in zip(metadata.calculations, metadata.xbrl_factoid):
+            calc_dfs.append(
+                pd.DataFrame(json.loads(calc)).assign(
+                    table_name=self.table_id.value, xbrl_factoid=factoid
+                )
+            )
+        return pd.concat(calc_dfs).reset_index(drop=True).explode("source_tables")
```
zaneselvans, line 2750: I've also found myself using this anti-pattern. I feel like there should be a simple way to do this that doesn't involve iterating over the rows. Like using `df.apply()` and/or `pd.json_normalize()`?

If you do need to iterate over a dataframe, rather than zipping columns together there's also the built-in [`df.itertuples()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.itertuples.html) which lets you refer to named elements in a single row like:

```py
for row in metadata.itertuples():
    calc_dfs.append(
        pd.DataFrame(json.loads(row.calculations)).assign(
            table_name=self.table_id.value, xbrl_factoid=row.xbrl_factoid
        )
    )
```

src/pudl/transform/ferc1.py
```diff
@@ -6195,3 +6213,15 @@ def plants_steam_ferc1(
...
+@asset
+def calculation_components_xbrl_ferc1(clean_xbrl_metadata_json):
+    """Create calculation-compnent table from table-level metadata."""
+    calc_metas = []
+    for table_name, tranformer in FERC1_TFR_CLASSES.items():
+        calc_meta = tranformer(
+            xbrl_metadata_json=clean_xbrl_metadata_json[table_name]
+        ).process_xbrl_metadata_calculations()
+        calc_metas.append(calc_meta)
+    return pd.concat(calc_metas)
```
zaneselvans, line 6227: Do we want a `.reset_index(drop=True)` on this?

src/pudl/output/ferc1.py
```diff
@@ -1582,7 +1624,10 @@ def convert_calculations_into_calculation_component_table(
             pd.DataFrame(json.loads(calc)).assign(table_name=tbl, xbrl_factoid=factoid)
         )
     calcs = pd.concat(calc_dfs).reset_index(drop=True)
-
+    if unpack_source_tables:
```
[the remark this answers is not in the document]
cmgosnell: oh i didn't realize i kept this function in here at all! it's no longer being used. i migrated it over into transform.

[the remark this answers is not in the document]
cmgosnell: that was a relic of a test. good catch

src/pudl/transform/params/ferc1.py
```diff
@@ -3980,7 +3980,6 @@
             # Note: this table does not currently get exploded. It will require
             # additional debugging at a later date.
             "calculation_tolerance": 0.4,
-            "subtotal_column": "depreciation_type",
```
cmgosnell, line 3983: this subtotal column is also the `xbrl_factoid` column. i think it was here bc there are `"total"`s in this column? @e-belfer i removed it bc it was somewhat duplicative to that the factoid and the subtotal be the same - it was junking up the idea of the primary key for the `table_dimensions_ferc1`

bc this table isn't being slurped into the explosion it felt okay to just ✂️  it

e-belfer: Yes that makes sense to me.

src/pudl/transform/ferc1.py
cmgosnell: vv good point! i made a big revision here

src/pudl/output/ferc1.py
```diff
@@ -970,11 +978,41 @@ def create_exploded_table_assets() -> list[AssetsDefinition]:
...
+                ~calc_explode.table_name_calculation_component.isin(self.table_names),
+                "xbrl_factoid",
+            ].unique()
+        )
+        calc_explode = calc_explode[
+            ~calc_explode.xbrl_factoid.isin(not_in_explosion_xbrl_factoids)
+        ].copy()
+        return calc_explode
+
+    def boom(self):
```
cmgosnell, line 1015: should this just be:
```
@property
def table_metadata_explosion(self)
```

or some other such name like `xbrl_factiod_metadata`. now that this class has two care outputs: the table-factoid level metadata and the `calculation_components_explosion` this no longer seems like the right name

src/pudl/output/ferc1.py
```diff
@@ -1007,11 +1045,16 @@ def boom(self, clean_xbrl_metadata_json: dict):
...
-        calc_explode = convert_calculations_into_calculation_component_table(
-            meta_explode
-        )
-        calc_explode["in_explosion"] = calc_explode.source_tables.apply(
+        """Overwrite the calculations with calculation components not in explosion.
+
+        TODO: remove this method and the one call for it within boom once the
+        Tree/Forest situation directly uses the calculation components instead of the
+        embedded calculations.
+        """
```
cmgosnell, line 1053: this step will no longer be necessary once the tree setup uses the calc components directly bc `calculation_components_explosion` now does this work

src/pudl/output/ferc1.py
```diff
@@ -1082,11 +1137,10 @@ def exploded_pks(self) -> list[str]:
...
         # table_name here.
-        pks = (
-            pudl.helpers.dedupe_n_flatten_list_of_lists(pks)
-            + ["xbrl_factoid"]
-            + ["table_name"]
-        )
+        pks = [
+            "table_name",
+            "xbrl_factoid",
+        ] + pudl.helpers.dedupe_n_flatten_list_of_lists(pks)
```
cmgosnell, line 1143: i put these two first bc it looks prettier in the dfs when you `set_index(self.exploded_pks)` 🎨

src/pudl/output/ferc1.py
```diff
@@ -1533,38 +1522,16 @@ def remove_inter_table_calc_duplication(
         return exploded
 
 
-def in_explosion_tables(
-    source_tables: list[str], in_explosion_table_names: list[str]
-) -> bool:
+def in_explosion_tables(table_name: str, in_explosion_table_names: list[str]) -> bool:
```
cmgosnell, line 1525: we can delete this function once we delete `redefine_calculations_with_components_out_of_explosion`

src/pudl/transform/ferc1.py
```diff
@@ -2672,6 +2672,37 @@ def remove_nones_in_list(list_of_things):
...
+                columns={
+                    "name": "xbrl_factoid_calculation_component",
+                    "source_tables": "table_name_calculation_component",
+                }
+            )
+            .merge(
+                metadata.drop(columns=["calculations"]),
+                left_index=True,
+                right_index=True,
+            )
```
cmgosnell, line 2700: i didn't know how to keep any of the other `metadata` columns in the `pd.json_normalize` step. so my reaction to that was just to make sure the index was fresssshh above and merge here. it feels okay but if anyone knows how to make `json_normalize` just do the thing lmk

src/pudl/transform/ferc1.py
```diff
@@ -6141,3 +6172,135 @@ def plants_steam_ferc1(
...
+    # reconcile_table_calculations
+    other_dimensions = []
+    for transformer in FERC1_TFR_CLASSES.values():
+        other_dimensions.append(
+            transformer().params.reconcile_table_calculations.subtotal_column
+        )
+    # remove nulls and dedupe
+    other_dimensions = [sub for sub in other_dimensions if sub]
+    other_dimensions = list(set(other_dimensions))
+    return other_dimensions
```
cmgosnell, line 6189: this is now kiiiind of a dupe of `other_dimensions` over int the outputs.. i could add a `table_names=None` which defaults to all of the tables to reuse this in both places.

src/pudl/transform/ferc1.py
```diff
@@ -2976,6 +2976,37 @@ def remove_nones_in_list(list_of_things):
...
+        if all(metadata.calculations == "[]"):
+            logger.info(f"{self.table_id.value}: No calculations.")
+            return pd.DataFrame()
+
+        metadata.calculations = metadata.calculations.apply(json.loads)
+        # reset the index post calc explosion so we can merge on index later
+        metadata = metadata.explode("calculations").reset_index(drop=True)
+        calc_comps = (
+            pd.json_normalize(metadata.calculations)
+            .explode("source_tables")
```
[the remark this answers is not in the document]
zaneselvans: I think this is the new expected behavior and will let us work with the calculation components much more like we work with the other data now.

src/pudl/transform/ferc1.py
```diff
@@ -2976,6 +2976,37 @@ def remove_nones_in_list(list_of_things):
...
+            .rename(
+                columns={
+                    "name": "xbrl_factoid_calculation_component",
```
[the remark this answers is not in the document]
zaneselvans: Those names read better, but I think these are the copies of the PK columns that are associated with the calculation components in the association table, so just having a suffix to differentiate them from the "main" versions of the same columns makes sense.  But I think it could just be `_calc` and that would be fine.

cmgosnell:
Ahh @e-belfer great point about the off ratios! I thought i noted this somewhere but i definitely didn't. Rn it's testing all of the calcs - not just the inter-table calcs. I think I got it in my head that this was a good idea bc it's possible/easier to test everything but i will revert this.

src/pudl/output/ferc1.py
```diff
@@ -1094,15 +1130,19 @@ def redefine_calculations_with_components_out_of_explosion(
...
+        Tree/Forest situation directly uses the calculation components instead of the
+        embedded calculations.
         """
-        calc_explode = convert_calculations_into_calculation_component_table(
-            meta_explode
-        )
-        calc_explode["in_explosion"] = calc_explode.source_tables.apply(
-            in_explosion_tables, in_explosion_table_names=self.table_names
-        )
+        calc_explode = self.calculation_components_xbrl_ferc1
```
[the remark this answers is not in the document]
cmgosnell: it is the same process. I almost make a little function that reduced the duplication here but this whole method is temporary. I need the full set of calculation components in order to ID the calculations that have any calculation components that are o/s of the explosion (which is happening in both methods). In `calculation_components_explosion` i remove all of those. but here i need to do that removal from the metadata bc rn the tree setup currently uses the metadata table. i can't use the processed exploded calc components bc all of the info about which calcs are o/s of the explosion already removed. i figured bc this is temporary it was okay but it wouldn't be too hard to clean it up.

src/pudl/transform/ferc1.py
```diff
@@ -6509,3 +6540,135 @@ def plants_steam_ferc1(
...
+        table_name: transformer().params.xbrl_factoid_name
+        for (table_name, transformer) in FERC1_TFR_CLASSES.items()
+    }
+
+
+@asset(ins={table_name: AssetIn(table_name) for table_name in FERC1_TFR_CLASSES})
+def table_dimensions_ferc1(**kwargs) -> pd.DataFrame:
+    """Build a table of dimensions found in the transformed data tables.
+
+    The dimensions that show up within the transformed tables differ across table and
```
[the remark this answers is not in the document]
zaneselvans: I am also not really understanding.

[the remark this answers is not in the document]
cmgosnell: yea i had trouble describing this so your not off base for not following. it builds a table with columns of: table_name + xbrl_factoid + all the other dimension columns ( i think its just utility_type, plant_function rn). the data in the table represent all of the other dimensions that are present in the tables.

(as an aside I sometimes want to type hints with column name for dataframes bc i think that would be generally helpful. I think pandera could be used to do this and i kinda wanna figure that out)

src/pudl/transform/ferc1.py
```diff
@@ -6509,3 +6540,135 @@ def plants_steam_ferc1(
...
+    """Make all of the implied dimensions in the calculation components explict.
+
+    We are assuming that if a dimension was associted with a calculation component, then
+    that is the only dimension that applies whereas if no dimension was associated with
+    a calculation component, that calculation component should be applied across all
+    relevant dimensions.
+
+    Relate all of the calculation components with all found dimensions from
+    :func:`table_dimensions_ferc1`, unless the calculation components already
+    contained information regarding dimensions.
```
zaneselvans, line 6639: I'm not understanding what this function is doing from the docstring or the code.

What do we expect the structure of the input calculation component df to look like, and how do we change it? What does it mean to "relate all of the calculation components with all found dimensions?"

src/pudl/transform/ferc1.py
```diff
@@ -6509,3 +6540,135 @@ def plants_steam_ferc1(
...
+            .dropna(),
+            on=[
+                "table_name_calculation_component",
+                "xbrl_factoid_calculation_component",
+            ],
+            how="left",
+        )
+        calc_comps_w_explicit_dims = calc_comps_w_dims[~null_dim_mask]
+        calc_comps_w_dims = pd.concat(
+            [calc_comps_w_implied_dims, calc_comps_w_explicit_dims]
```
zaneselvans, line 6672: Is there a reason to incrementally concatenate with every iteration of the loop rather than doing one concatenation at the end?

zaneselvans (changes requested):
Definitely feeling confused about a bunch of stuff in here.

src/pudl/output/ferc1.py
cmgosnell: i like that much better! because of the class name could it be `MetadataExploder.metadata` (or `MetadataExploder.xbrl_factoids`) & `MetadataExploder.calculations`? or something i think having exploded many times in the chain is a lil duplicative.

zaneselvans: I think `.metadata` and `.calculations` in context seems good.  We can assign them to some other variable name elsewhere when we read them if we need to transmit the 💥 context.

src/pudl/transform/ferc1.py
```diff
@@ -1707,7 +1707,7 @@ def apply_xbrl_calculation_fixes(
                                 "depreciation_amortization_summary_ferc1"
                             ],
                             "utility_type": "electric",
-                            "subdimension": "electric",
+                            "plant_function": "electric",
```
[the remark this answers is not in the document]
cmgosnell: ah yeah this was a rename that [happened outside of the standard categorization](https://github.com/catalyst-cooperative/pudl/blob/09ee58540446e8b1dff5e47cc122b8586f69a30a/src/pudl/transform/ferc1.py#L5947). hm well we do need this thing that is a total to show up as `electric` in this table. i'll have to investigate whether that would be appropriate in any other case.

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
cmgosnell: yeah agreed! we can add checks down the line if there are multiple source tables in the output table for applications where that matters.

src/pudl/transform/ferc1.py
cmgosnell: we are not making new little dfs each iteration, we are split/apply/combining on the same df each iteration. So for each dimension column (`utility_type` or `plant_function` rn) we split the records w/&w/o values in the dimension column, for the records w/o dimension values, we merge in the dimensions found in `table_dimensions` then combine.

src/pudl/output/ferc1.py
```diff
@@ -1328,106 +1375,41 @@ def generate_intertable_calculations(self, exploded: pd.DataFrame) -> pd.DataFra
...
-            # Fill NAs with 0 to add corrections
-
-            calculated_dfs.append(calc_df)
+            # apply the weight from the calc to convey the sign before summing.
+            .assign(calculated_amount=lambda x: x[self.value_col] * x.weight)
+            .groupby(self.exploded_pks, as_index=False, dropna=False)[
+                ["calculated_amount"]
+            ]
+            .sum(min_count=1)
+        )
```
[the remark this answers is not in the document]
cmgosnell: from:
# effectively convert the data in the exploded table into calc components with
# a rename/merge & groupby the xbrl_factoid column from the calc components tbl

to:
# grab all of the calculation components from the exploded data by renaming the
# columns in the exploded data to the `_calc` columns from the calculation component
# table & merging that with the calculation components. then we have a table that
# includes the data of every single calculation component. Bc the `xbrl_factoid` and 
# `table_name` that each calculation component is associated with is in the calc 
# component table, you can then just groupby/sum by the `self.explode_pks`

... i think i need help? that seems really verbose and probably still doesn't make sense?

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
e-belfer: Oops this is my muck up. I'm going to undo this in #2688.

src/pudl/transform/ferc1.py
cmgosnell: okay i added a lot of backstory to the docs.. does that make it more clear?

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
zaneselvans: I'm still not getting your explanation. What does each record in the table being compiled represent?  When you say "The data in the table represent all of the other dimensions that are present in the tables" do you mean there's something other than the dimension columns in the table that's being compiled?

Or rather than talking about something outside of the dimensions, are you looking at the **values** which appear in each dimension, for each table?  Maybe...

"Compile a dataframe indicating what distinct values are observed in the data for each dimension column in association with each unique combination of `table_name` and `xbrl_factoid`. E.g. for all factoids found in the `plant_in_service_ferc1` table, the only value observed for `utility_type` is `electric` and the only value of `plant_status` observed is `in_service`." (Though this might be a bad example because I think both `utility_type` and `plant_status` are implicit in the `plant_in_service_ferc1` table so maybe they don't actually end up in this compilation? Or have they already been added into the data at this point?)

(This does seem to be exactly what Pandera is for and I agree it would be super helpful. Just saying "it's a dataframe" is not a whole lotta info.)

src/pudl/transform/ferc1.py
zaneselvans: So the intention is that the new records which are added in each iteration of the loop are going to be included as inputs into subsequent iterations?  If that's true, does it matter what order we loop over the dimensions in? Could we get different outputs if we did `plant_status` before `utility_type` vs. the opposite ordering?

src/pudl/transform/ferc1.py
```diff
@@ -6509,3 +6540,152 @@ def plants_steam_ferc1(
...
+            calc_comps_w_dims[null_dim_mask].drop(columns=[dim_col]),
+            table_dimensions[["table_name", "xbrl_factoid", dim_col]]
+            .rename(
+                columns={
+                    "xbrl_factoid": "xbrl_factoid_calc",
+                    "table_name": "table_name_calc",
+                }
+            )
+            .drop_duplicates()
+            .dropna(),
```
zaneselvans, line 6680: I'm finding it difficult to read this statement with the construction of the right dataframe inside the merge. Is this attempting to fill NA values found in the dimension columns?

Rather than using a merge here and iterating over all of the dimension columns, would it make sense to use `df1.fillna(df2)`?  If we give both dataframes the same index, we can fill NA values in `df1` using whatever values are found at the same location (row index + column name) in `df2`. Is this a `many-to-one` merge? I think this `fillna()` would broadcast values in the event that the row index didn't identify unique records in `df1`?

But maybe I'm still not understanding the structure of this table, or what it means to "relate all calculation components with all found dimensions..."

In the case of the `plant_in_service_ferc1` table, what happens?  That table has implied `utility_type=="electric"` and `plant_status=="in_service"` and there's a reference to it from a calculation in the `utility_plant_summary_ferc1` table that knows about the `utility_type` dimension. But not the `plant_status` dimension? Does it end up only being associated with `electric` but associated with all values of `plant_status`?

src/pudl/transform/ferc1.py
```diff
@@ -6509,3 +6540,152 @@ def plants_steam_ferc1(
...
+        )
+        calc_metas.append(calc_meta)
+    # squish all of the calc comp tables then add in the implicit table dimensions
+    calc_components = pd.concat(calc_metas).pipe(
+        make_implict_dimensions_explict_in_calculaiton_components, table_dimensions
+    )
+    return calc_components
+
+
+def make_implict_dimensions_explict_in_calculaiton_components(
```
zaneselvans, line 6630: Typo in function name. Should be `_calculation_` not `_calculaiton_`. Maybe just call it `add_explicit_dimensions()` or `make_dimensions_explicit()`?

cmgosnell: ty for the typo :eyes:. I also like the simplification but i'd like to go with `make_calculation_dimensions_explicit()` because this is very calculation focused.

src/pudl/transform/ferc1.py
cmgosnell: awe i love the idea of using `fillna` but i don't think it supports one-to-many broadcasts

```
df1 = pd.DataFrame(
    [1, 2, pd.NA, 11, pd.NA, 14],
    index=["a"]*3 + ["b"]*3,
    columns=["data"]
)

df2 = pd.DataFrame(
    [3,4, 11, 12,13],
    index=["a"]*2 + ["b"]*3,
    columns=["data"]
)
df1.data.fillna(df2.data)
>>> ValueError: cannot reindex on an axis with duplicate labels
```

i did move the manipluation of `table_dimensions` out of the merge bc yes that was too confusing.

re your `plant_in_service_ferc1` example: for the `utlity_type` dimension, because it `electric` was added in the calc fixes, no additional `utility_type`'s get added. For `plant_status`, we are effectively checking for all of the `plant_status`'s that are observed in that `plant_in_service_ferc1` table and merging those on in order to broadcast the observed dimension. In this case, there are no `plant_status`'s found in `plant_in_service_ferc1` so nothing gets broadcast merged. But if two `plant_status`'s were observed in `plant_in_service_ferc1` it would be a one-to-many merge.

src/pudl/transform/ferc1.py
cmgosnell: this is a good question but I don't believe the order will effect the outpcome. for each dimension column when the values are null, I am dropping the full column & one-to-many merging on a null dimension column.

cmgosnell: i'll make a unit test for this!

src/pudl/output/ferc1.py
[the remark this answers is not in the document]
zaneselvans: Merge the reported data and the calculation component metadata to enable validation of calculated values.  Here the data table `exploded` is supplying the values associated with individual calculation components, and the `table_name` and `xbrl_factoid` to which we aggregate are coming from the calculation components table. After merging we use the weights to adjust the reported values so they can be summed directly. This gives us aggregated calculated values that can later be compared to the higher level reported values.

src/pudl/transform/ferc1.py
```diff
@@ -6636,28 +6636,42 @@ def make_calculation_dimensions_explicit(
...
-    ``xbrl_factoid``.
+    In the raw XBRL metadata's calculations, there is an implicit assumption that
+    calculated values are aggregated within categorical columns called Axes or
+    dimensions, in addition to being grouped by date, utility, table, and fact. However,
+    the relevant dimensions and values are not specified in the calculation components,
+    and the same calculation is assumed to apply to all groups within a dimension.
+
+    We have extended this calculation system to allow independent calculations to be
+    specified for different values within a given dimension, and for these calculations
+    to include values reported in multiple tables.  For example, the
```
cmgosnell, line 6716: `and for these calculations to include values reported in multiple tables.`

This is not new or super duper relevant imo.

zaneselvans (approved):

src/pudl/transform/ferc1.py
zaneselvans: I was adapting a sentence that already existed talking about calculating spanning multiple tables. I can remove it.

src/pudl/transform/ferc1.py
```diff
@@ -6636,28 +6636,42 @@ def make_calculation_dimensions_explicit(
...
-    contain many values within these dimension columns.
-
-    Any dimension that was specified will not be touched. Any remaining nulls are a
-    result of never having observed values for that ``table_name`` and
-    ``xbrl_factoid``.
+    In the raw XBRL metadata's calculations, there is an implicit assumption that
+    calculated values are aggregated within categorical columns called Axes or
+    dimensions, in addition to being grouped by date, utility, table, and fact. However,
+    the relevant dimensions and values are not specified in the calculation components,
+    and the same calculation is assumed to apply to all groups within a dimension.
```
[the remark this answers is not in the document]
zaneselvans: I'm trying to get at different things in the two parts. I've tried to clarify the difference.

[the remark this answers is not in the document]
cmgosnell: i don't actually see the difference in the new version of the two instances of what your saying but both still seem true to me. so cooool!

PR #2741, opened 2023-07-20
Add EIA860, EIA860m, & EIA923 2022 early release data

aesharpe:
# PR Overview

This includes: adding new zenodo doi archive values, adding the new year of data to fast and full etl yml files, add the new year to the working_partitions in the metadata, update the package data, fix encoding errors, update validation tests for row count.

# PR Checklist

# Lingering things / findings that will be made into new issues once this is merged:
- The need to lower case things
- A check that upper cases encoding stuff


src/pudl/package_data/settings/etl_full.yml
```diff
@@ -215,6 +215,7 @@ datasets:
           2019,
           2020,
           2021,
+          2022,
         ]
       eia860m: true
```
zaneselvans, line 220: We need to disable EIA-860M here too don't we?

src/pudl/settings.py
```diff
@@ -201,7 +201,7 @@ class Eia860Settings(GenericDatasetSettings):
     eia860m_date: ClassVar[str] = eia860m_data_source.working_partitions["year_month"]
 
     years: list[int] = data_source.working_partitions["years"]
-    eia860m: bool = True
+    eia860m: bool = False  # True # temp change! need to add 2023 monthly data
```
zaneselvans, line 204: Do we need to change the default value for the class, which we ought to be overriding in the ETL settings already, temporarily? It seems like in the general case we do still want to default to True -- and override it as necessary (as we're doing now).

zaneselvans (changes requested):
Minor issues w/ the exclusion of EIA-860M.

Given how intertwined they are, and how fast that was, and how few tables there are in the EIA-923 compared to EIA-860, might we want to go ahead and update the EIA-923 for 2022 in this PR as well?

src/pudl/transform/eia860.py
```diff
@@ -161,6 +161,16 @@ def clean_ownership_eia860(raw_ownership_eia860: pd.DataFrame) -> pd.DataFrame:
...
+    # Spot fix NA generator_id. Might want to change this once we have the official 2022
+    # data not just early release.
+    constraints = (own_df["plant_id_eia"] == 62844) & (
+        own_df["report_date"].dt.year == 2022
+    )
+    if len(own_df[constraints]) > 1:
+        raise AssertionError("Too many records getting spot fixed.")
+
+    own_df.loc[constraints, "generator_id"] = "1"
+
```
aesharpe, line 173: This is a bit of a hack-y way to get around an error that was throwing due to their being a null value in the primary key (i.e., `generator_id`) for just one row in this table. The plant doesn't show up for any other years in this table or the generators table. It does show up in the plants table, however, so I'm hoping it's just a quirk of the early_release. If you have a better suggestion of how to handle this I'd love to know. If, for some reason, this plant shows up in the generators table later on with a different generator id that would obviously not be good.

zaneselvans: Hmm, that's unfortunate! Does it show up in the EIA-860M data by chance?

It might be good to apply the spot-fix contingent on seeing an actual NA in the generator ID, so if it does get filled in later we're not clobbering it.  Or maybe we just have an assertion that fails if the ID shows up later, and we can then remove the hack.

src/pudl/package_data/settings/etl_full.yml
aesharpe: I talked to @bendnorman about this and apparently not. But given that we don't have any eia860 monthly data right now we could probably set this to false anyways. It won't affect the ETL because we use the jobs defined in dagster of which only the fast tests is based on the yml file. I do think it might impact the tox tests however as those are the only files that sill use this etl_full.yml file I think.

src/pudl/settings.py
aesharpe: Yeah I agree. I wasn't sure the best way to make this a temporary fix that would have to be changed later once we add in monthly data. I suppose we could write something that would check for monthly data that was greater than the annual data available and if that was not true then eia860m bool would be set to false?

src/pudl/transform/eia860.py
aesharpe: That's a good suggestion, I'll do that!

src/pudl/settings.py
aesharpe: Alternatively, I could just go ahead and integrate the 2023 monthly EIA 860 data

src/pudl/package_data/settings/etl_full.yml
zaneselvans: I think we should set it to `false` otherwise we're specifying settings which we expect to fail here.

If the builds aren't running based on these YML files, where do we need to update those settings?

src/pudl/settings.py
zaneselvans: I don't understand. Can't you just leave the fault True? The temporary changes you've made to the settings should override that default.

src/pudl/package_data/settings/etl_full.yml
aesharpe: In the dagster launchpad for the etl_full job

aesharpe: but I agree it should be false for the tox tests

src/pudl/settings.py
aesharpe: I think this is a @bendnorman question.

src/pudl/package_data/settings/etl_full.yml
zaneselvans: @bendnorman I'm definitely confused. Are the `etl_fast.yml` and `etl_full.yml` files not the source of the fast/full job configurations in Dagster?

src/pudl/settings.py
bendnorman: The `etl_fast` job pulls from `src/pudl/package_data/settings/elt_fast.yaml`, and `etl_full` uses the default values of the settings objects. We should probably create `etl_full` from the settings file instead so it's less confusing and we don't run into the issue where people are temporarily changing the default values to work with `etl_full`.

Want me to make a PR off of this branch?

src/pudl/metadata/codes.py
```diff
@@ -1564,6 +1564,8 @@
             "WT": "WND",
             "H2": "OG",
             "OOG": "OG",
+            "Sun": "SUN",
+            "sub": "SUB",
```
aesharpe, line 1568: We could probably write a function that would just turn this column uppercase and avoid this problem in the future. I don't think we have a helper function that does that right now. Maybe it's too simple for a helper function? Thoughts on where this type of generalized solution to this spot fix should go? @zaneselvans

src/pudl/glue/ferc1_eia.py
```diff
@@ -451,20 +461,21 @@ def get_util_ids_eia_unmapped(
...
+    utilities_eia_db = utilities_eia_db.set_index(["utility_id_eia"])
+    unmapped_utils_eia = utilities_eia_db.loc[unmapped_utils_eia_index]
+    util_recent_cap = get_utility_most_recent_capacity(pudl_engine)
+
+    unmapped_utils_eia = pd.merge(
+        unmapped_utils_eia,
+        util_recent_cap,
+        left_index=True,
+        right_index=True,
+        how="left",
```
aesharpe, line 478: This change was because there were some new utilities that don't show up in the `generators_eia860` table which is where the capacity is coming from and the original `.loc[unmapped_utils_eia] was failing because there were util values not in the capacity table.

src/pudl/glue/ferc1_eia.py
```diff
@@ -541,9 +552,7 @@ def glue(ferc1=False, eia=False):
     # or trailing white space... since this field is being used as a key in
     # many cases. This also needs to be done any time plant_name is pulled in
     # from other tables.
-    plant_map = get_plant_map().pipe(
-        pudl.helpers.simplify_strings, ["plant_name_ferc1"]
-    )
+    plant_map = get_plant_map()
```
aesharpe, line 555: Added the simplify strings to all the string fields in get_plant_map() instead

src/pudl/metadata/classes.py
```diff
@@ -518,10 +518,10 @@ def encode(
         """Apply the stored code mapping to an input Series."""
         # Every value in the Series should appear in the map. If that's not the
         # case we want to hear about it so we don't wipe out data unknowingly.
+        logger.info(f"Encoding {col.name}")
         unknown_codes = set(col.dropna()).difference(self.code_map)
         if unknown_codes:
             raise ValueError(f"Found unknown codes while encoding: {unknown_codes=}")
-        logger.debug(f"Encoding {col.name}")
         col = col.map(self.code_map)
```
aesharpe, line 525: Moved so it was easier to see what column failed

src/pudl/metadata/codes.py
cmgosnell: Encoder

test/integration/ferc1_eia_train_test.py
```diff
@@ -70,7 +70,7 @@ def ferc1_eia_training_data():
         ),
         pytest.param(
             ["True"],
-            [2019],
+            [2021],
             ["2_2020_plant_owned_195"],
             ["f1_steam_2020_12_2_0_1"],
             [18],
```
cmgosnell, line 76: is it a problem that the year here is different than the year in the two test ids?

test/integration/glue_test.py
```diff
@@ -86,6 +86,12 @@ def glue_test_dfs(
             ),
         }
     )
+    # Make everything lowercase
+    glue_test_dfs = {
+        df_name: df.applymap(lambda x: x.lower() if type(x) == str else x)
+        for (df_name, df) in glue_test_dfs.items()
+    }
+
```
cmgosnell, line 94: is this necessary now that you moved where the lower casing is happening in the glue module (which i think is the more correct place to have any of that happen)?

cmgosnell (commented):
this mostly looks real good to me! the simplifications of when/where the glue names are getting lower cased make a lot of sense. we chatted about this offline but agree with you that it would be good to make a lil pr right after this to lowercase everything in the saved CSVs.

I haven't fully wrapped my head around the eia860m flag issue/threads tbh. At a very basic level, was it not possible to just add the most recent 2023 month for 860m? it sounds like more work upfront but probably less work in the long wrong bc we don't have to deal with it not being in the etl (which i'm not sure why that would be a problem either - even though i don't doubt that it did pose one).

Is the overall eia860m question just about where we need to tell the etl "do or don't load 860m"? rn its being set in the settings object? so the settings classes are not actually talking to the settings files? is that the basic gist here?

also do you know why the tests failed? it looks like they were stopped but pretty late in the process?

aesharpe:
> also do you know why the tests failed? it looks like they were stopped but pretty late in the process?

I think they failed because now the fast tests include three years of data instead of two to accommodate the lack of FERC data and that was too heavy a lift for the ci.

src/pudl/package_data/settings/etl_fast.yml
```diff
@@ -37,9 +37,9 @@ datasets:
     years: [2019, 2020]
   eia:
     eia923:
-      years: [2020, 2021]
+      years: [2020, 2021, 2022]
     eia860:
-      years: [2020, 2021]
+      years: [2020, 2021, 2022]
```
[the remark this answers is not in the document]
aesharpe: yeah. But now having three years in the fast tests is causing the ci to fail I think

test/integration/ferc1_eia_train_test.py
aesharpe: No, that's the point of the test - to fail when there is a discrepancy in years. I just wanted to update the year.

test/integration/glue_test.py
aesharpe: The tables added by `glue_test_dfs.update(glue(eia=True, ferc1=True))` will be lowercase because the `glue` function calls `get_plant_map` where I added the lower-casing function. 
However, the tables added before:
```
    glue_test_dfs = {
        "plants_eia_pudl_db": pudl_out.plants_eia860(),
        "util_ids_ferc1_raw_xbrl": get_util_ids_ferc1_raw_xbrl(ferc1_engine_xbrl),
        "util_ids_ferc1_raw_dbf": get_util_ids_ferc1_raw_dbf(ferc1_engine_dbf),
        "plants_ferc1_raw": plants_ferc1_raw(dataset_settings_config),
        "plants_eia_labeled": label_plants_eia(pudl_out),
    }
```
and after:
```
    glue_test_dfs.update(
        {
            "utilities_ferc1_dbf_labeled": label_utilities_ferc1_dbf(
                glue_test_dfs["utilities_ferc1_dbf"],
                glue_test_dfs["util_ids_ferc1_raw_dbf"],
            ),
            "utilities_ferc1_xbrl_labeled": label_utilities_ferc1_xbrl(
                glue_test_dfs["utilities_ferc1_xbrl"],
                glue_test_dfs["util_ids_ferc1_raw_xbrl"],
            ),
        }
    )
```
Are not part of that function and won't get lower-cased. It looks like the later bit says: "more glue test input tables that are compiled from tables already in glue_test_dfs" so I wonder A) are they already lower case because they're derived from tables that are? and B) could I move this lower casing function up to just affect the first few tables added to the `glue_test_dfs` that aren't lower cased already?

src/pudl/metadata/codes.py
aesharpe: Is there an instance when we wouldn't want to apply it that you can think of?

cmgosnell:
hmmm if the ci will fail with three years of data then we need a new solution. can we duplicate some ferc-eia training data for the new year? it doesn't have to be a full duplication of the last year of training data. even a few records would suffice i think.

test/integration/glue_test.py
cmgosnell: ah okay! well then yea i think this is a fine fail safe. perhaps we can make a more systematic LOWER EVERYTHING issue after this is done so we can do that more uniformly but i think this is a good solution for now!

src/pudl/metadata/codes.py
aesharpe: I think I'm gonna make another PR for this

zaneselvans (commented):
Have you been able to run the full ETL and data validations locally? Always hard to evaluate these big updates to the CSVs.

src/pudl/package_data/settings/etl_fast.yml
```diff
@@ -37,10 +37,10 @@ datasets:
     years: [2019, 2020]
   eia:
     eia923:
-      years: [2021, 2022]
+      years: [2020, 2021, 2022]
     eia860:
-      years: [2021, 2022]
-      eia860m: false # true; change back when add eia860m
+      years: [2020, 2021, 2022]
```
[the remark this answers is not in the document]
aesharpe: This is a good question. I don't really know the answer tbh. I think the idea is that we don't need to add more training data because it's supposed to train the model to fill it in. But we could certainly add a few rows of training data for a new year if we needed to.It was a a big undertaking to do all the top utilities.

aesharpe:
> Have you been able to run the full ETL and data validations locally? Always hard to evaluate these big updates to the CSVs.

Yes!

src/pudl/transform/eia860.py
```diff
@@ -161,6 +161,19 @@ def clean_ownership_eia860(raw_ownership_eia860: pd.DataFrame) -> pd.DataFrame:
...
+        own_df["report_date"].dt.year == 2022
+    )
+
+    if len(own_df[constraints]) > 1:
+        raise AssertionError("Too many records getting spot fixed.")
+    if own_df[constraints].generator_id.notna().all():
+        raise AssertionError("Generator ID filled in, you can remove this hack!")
+    else:
+        own_df.loc[constraints, "generator_id"] = "1"
+
```
cmgosnell, line 176: i think if you added `elif 2022 not in own_df.report_date.dt.year.unique(): pass` into this mix this would enable a non-2022 run.

perhaps you could just add it into the if own[constraints] line?

aesharpe: Ah right, I figured it wasn't necessary because we are adding 2022 in, but there will be a time in the future when 2022 isn't in the fast tests.

src/pudl/analysis/allocate_gen_fuel.py
```diff
@@ -389,9 +390,10 @@ def allocate_gen_fuel_by_generator_energy_source(
         validate="1:1",
         suffixes=("_net_gen_alloc", "_fuel_alloc"),
     ).sort_values(IDX_GENS_PM_ESC)
-    _ = test_original_gf_vs_the_allocated_by_gens_gf(
-        gf=gf, gf_allocated=net_gen_fuel_alloc
-    )
+    if not os.environ.get("GITHUB_ACTIONS", False):
+        _ = test_original_gf_vs_the_allocated_by_gens_gf(
+            gf=gf, gf_allocated=net_gen_fuel_alloc
+        )
```
cmgosnell, line 396: this is such a nice solution to this problem! i do think we should add a little comment re: this is a defensive data check that takes much memory so we're not running it in our memory constrained CI

cmgosnell (approved):
this looks good! ty for the fiddly memory issue stuff needed to get it over the finish line. I suggested one comment but it is not a show stopper/shouldn't effect any of the code so i think this all looks very ready to me!

src/pudl/analysis/allocate_gen_fuel.py
aesharpe: great idea!

zaneselvans (approved):

PR #2753, opened 2023-07-24
Add parent dimensions into calculation component table

cmgosnell:
# PR Overview

This PR does three main things:
* converts the calculation component table to have `_parent` suffix for the parent fact/table columns instead of `_calc` suffix for all of the calculation components
* adds the parent dimensions into the calc component table by:
   * adding total == rest of dimension value records
   * assuming non-total dimensions should have the same dimensions for parent & child.

I recommend starting from the bottom of the files changes: test -> transform -> output.

# PR Checklist


src/pudl/transform/ferc1.py
```diff
@@ -6745,31 +6752,80 @@ def make_calculation_dimensions_explicit(
...
             [calc_comps_w_implied_dims, calc_comps_w_explicit_dims]
         )
     return calc_comps_w_dims
+
+
+def add_parent_dimensions(calc_comps, dimensions):
+    """Add in the dimensions of the parents in the calculation components.
+
+    First, we treat the totals in these dimension columns by assuming that any total we
+    observe is the sum of all non-total dimensions. We implement this by creating new
```
zaneselvans, line 6783: I've often been confused by whether we're talking about multiple dimensions, or multiple values within a single dimension. I think here you mean:

"...is the sum of all non-total values within the same dimension..."

src/pudl/transform/ferc1.py
```diff
@@ -6745,31 +6752,80 @@ def make_calculation_dimensions_explicit(
...
+    calculation component records which have a parent of dimension total and calculation
+    components of the non-total dimensinos.
+
+    Then, we assume that every parent factoid should have the same dimensions as its
+    calculation component dimensions.
```
zaneselvans, line 6788: I think you mean that every parent factoid should have the same dimension **values** as its calculation components? E.g. if a parent has `utility_type=="electric"` then all of its components are also electric? Not that if a parent has a `utility_type` dimension then the child should also have a `utility_type` dimension.

Aren't parent nodes with the value `total` in any dimension an exception to this?

Maybe we mean "With the exception of parents having a value of `total` in some dimension, we assume that all parents should have the same dimension values as the calculation components that they are composed of."

src/pudl/transform/ferc1.py
```diff
@@ -6745,31 +6752,80 @@ def make_calculation_dimensions_explicit(
...
+        "table_name",
+        "xbrl_factoid",
+    ]
+    # for each dimension col, add a _parent column while
+    for dim in dimensions:
+        total_mask = calc_comps[dim] == "total"
+        total_w_subdim_components = (
+            pd.merge(
+                # the totals will become the parent record
+                calc_comps.loc[total_mask, calc_comp_idx + dimensions],
```
zaneselvans, line 6807: I think it's helpful to have `left=` and `right=` in these complex merges.

src/pudl/transform/ferc1.py
```diff
@@ -6745,31 +6752,80 @@ def make_calculation_dimensions_explicit(
...
+            calculation fixes applied.
+        dimensions: list of dimension columns to check.
+    """
+    calc_comp_idx = [
+        "table_name_parent",
+        "xbrl_factoid_parent",
+        "table_name",
+        "xbrl_factoid",
+    ]
+    # for each dimension col, add a _parent column while
```
zaneselvans, line 6801: "while..." ???

src/pudl/transform/ferc1.py
```diff
@@ -6745,31 +6752,80 @@ def make_calculation_dimensions_explicit(
...
+                how="left",
+                validate="1:m",
+                suffixes=("_parent", ""),
+            )
+            # these new total -> sub-dimension records should have the same
+            # parent and component fact/table so overwrite the og
+            .assign(
+                table_name_parent=lambda x: x.table_name,
+                xbrl_factoid_parent=lambda x: x.xbrl_factoid,
+            )
```
zaneselvans, line 6820: What's the scenario where this isn't already the case?

src/pudl/transform/ferc1.py
```diff
@@ -6745,31 +6752,80 @@ def make_calculation_dimensions_explicit(
...
+            # drop the other non-total parent dim cols (they will be merged back on later)
+            .drop(columns=[f"{d}_parent" for d in dimensions if d != dim])
+        )
+        calc_comps = pd.concat([calc_comps, total_w_subdim_components])
+        # we shouldn't be adding any duplicates in this process!
```
zaneselvans, line 6825: I think this comment got separated from its line of code below. where you assert duplicates are empty.

src/pudl/transform/ferc1.py
```diff
@@ -6745,31 +6752,80 @@ def make_calculation_dimensions_explicit(
...
+        # now fill in everything with the non-totals
+        calc_comps = calc_comps.assign(
+            **{f"{dim}_parent": lambda x: x[f"{dim}_parent"].fillna(x[dim])}
+        )
```
zaneselvans, line 6829: This line is confusing. Can we have a longer explanatory comment? What's being filled in by what?

src/pudl/transform/ferc1.py
zaneselvans: Could you add a description of what we expect to be true of the dataframe that's reported at the end of this process?  Or maybe you can explain it to me on the phone and I could write it?

zaneselvans (changes requested):
Maybe we can collaborate on some docstrings on the phone? You explain and I try to translate?

src/pudl/transform/ferc1.py
cmgosnell: fair! i decided to move this up before the concat so we wouldn't have to fillna at all & added more explanation. better?

src/pudl/transform/ferc1.py
cmgosnell: this whole step 1 in the treatment is generating *fully new* calculation records which specificly link the totals to their sub-components. so the parent values of the table_name/xbrl_factoid of these new records need to be the same as their child/components.

there are already valid calculation component records with total as a child dimension value that is say a part of a different calculation. Those records are not being overwritten. these records in `total_w_subdim_components` are just new and different records which communicate this total -> non-total calculation that was not previously encapsulated within the calculation components.

src/pudl/transform/ferc1.py
cmgosnell: i added some context of what these tables looked like coming into this function as well as what actually happens/what is expected on the way out. very happy to chat on the phone about it. i also recommend going to looking at the unit test.

src/pudl/transform/ferc1.py
```diff
@@ -6745,31 +6752,100 @@ def make_calculation_dimensions_explicit(
...
+        calc_comps: a table of calculation component records which have had some manual
+            calculation fixes applied.
+        dimensions: list of dimension columns to check.
+    """
+    calc_comp_idx = [
+        "table_name_parent",
+        "xbrl_factoid_parent",
+        "table_name",
+        "xbrl_factoid",
+    ]
```
zaneselvans, line 6815: Do the PKs necessarily not include the dimension columns? E.g. couldn't there be cases where we define different calculations depending `utility_type`? This almost happens where we have one calculation with lots of details from the `plant_in_service_ferc1` table, and a different simpler calculation for the gas and other utility types. In that case I think it happens to be the case that the calculation components are being reported in different tables, but does that necessarily have to be true?

Or is it the case that at this point in the processing of the calculation components table we're still adding values to the dimension columns, and at the end of this step they're going to be part of the PKs in a way that they weren't before? That seems true with respect to the addition of the `total` calculations, but is it true more generally?

zaneselvans (approved):

PR #2759, opened 2023-07-28
Modernize importlib usage

zaneselvans:
# PR Overview

* Updates our usage of `importlib.resources` to use the current API in a uniform and relatively simple way across the repo.  Previously we had a mix of [old deprecated functions](https://docs.python.org/3/library/importlib.resources.html#deprecated-functions) and an unnecessarily complex syntax from the new API. In almost all cases we can simply use the `Traversable` class and its `pathlib.Path` like interface.
* Also avoids use of the deprecated `pkg_resources` module which is going to be removed from the standard library in Python 3.12, replacing it with functionality from [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html)

# PR Checklist


docs/dev/packaging.rst
zaneselvans, line 1: Note that nearly everything in this documentation page is outdated, since we've moved to using `pyproject.toml` exclusively to define our packaging.

tox.ini
```diff
@@ -300,7 +300,6 @@ doctest_optionflags = NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL ELLIPSIS
 filterwarnings =
     ignore:distutils Version classes are deprecated:DeprecationWarning
```
zaneselvans, line 301: Unsure of why this warning is still coming up. Maybe from an imported package? I don't think we're directly using any of the `pkg_resources` versions now.

PR #2763, opened 2023-07-31
Use all dimensions in `XbrlCalculationForestFerc1` and exploded tables

zaneselvans:
# PR Overview

This PR updates the `MetadataExploder` and `Exploder` and `XbrlCalculationForestFerc1` classes to  use the additional dimensions that are required to uniquely identify all reported facts so they can be independently annotated and used to filter the output data.

# PR Checklist


src/pudl/output/ferc1.py
```diff
@@ -1063,6 +1077,88 @@ def calculations(self):
...
+        - Only parents with table_name in the explosion tables should be retained.
+        - All calculations in which *any* components were outside of the tables in
+          the explosion should be turned into leaves -- i.e. they should be replaced
+          with a single calculation component filled with NA values.
+        - Every remaining calculation component must also appear as a parent (if it
+          is a leaf, then it will have a single null calculation component)
+        - There should be no records where only one of table_name or xbrl_factoid
+          are null. They should either both or neither be null.
+        - table_name_parent and xbrl_factoid_parent should be non-null.
+        """
```
zaneselvans, line 1095: @cmgosnell this is a new version of the exploded calculation components that's trying to work with the new association table format, but several of the sanity checks are failing and I could use some external 👀

cmgosnell: i think i found at least one of your errors re my comment below

src/pudl/output/ferc1.py
```diff
@@ -1063,6 +1077,88 @@ def calculations(self):
...
+        # An index of all non-null child nodes / calculation components:
+        calc_idx = [
+            NodeId(*x)
+            for x in calc_explode[calc_cols].dropna(how="all").itertuples(index=False)
+        ]
+        # An index of all parent nodes:
+        parent_idx = [
+            NodeId(*x)
+            for x in calc_explode[parent_cols].dropna(how="all").itertuples(index=False)
+        ]
```
cmgosnell, line 1147: you can't drop all nulls because it is very expected that the dimension columns will have nulls in them

cmgosnell: higher level q: is there a benefit to converting these into nodes as opposed to just setting/using the index of the df?

cmgosnell: oh sorry i get the "all" now! i was assuming this was how="any"

src/pudl/output/ferc1.py
```diff
@@ -1063,6 +1077,88 @@ def calculations(self):
...
+
+        # Nodes that *only* appear as children / calculation components:
+        calc_only = list(set(calc_idx).difference(parent_idx))
+        if len(calc_only) > 0:
+            logger.error(
+                f"Found {len(calc_only)} nodes which *only* appear as calc components."
+            )
```
cmgosnell, line 1154: spot checking the calc-only nodes it looks like many of them are the total -> subdimension records.

src/pudl/output/ferc1.py
zaneselvans: Doesn't `how="all"` mean it'll only drop records where *all* the values are null?

src/pudl/output/ferc1.py
```diff
@@ -1063,6 +1077,88 @@ def calculations(self):
...
+            lambda x: x.isin(self.table_names + [pd.NA]).all()
+        )
+        # Convert any calculation componets from outside the explosion into reported
+        # values / leaf nodes:
+        calc_explode.loc[
+            ~calc_explode.is_within_explosion, list(NodeId._fields) + ["weight"]
+        ] = pd.NA
+        calc_explode = calc_explode.drop(columns=["is_within_explosion"]).reset_index(
+            drop=True
+        )
```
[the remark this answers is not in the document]
zaneselvans: The intention with:
```py
        calc_explode.loc[
            ~calc_explode.is_within_explosion, list(NodeId._fields) + ["weight"]
        ] = pd.NA
```
is to set all of the calculation component fields to NA, which turns their parents into leaves. But I guess then there are many identical duplicate null records, since there's one for every former component, and we just want a single one for the parent (with no other values in the record).

src/pudl/output/ferc1.py
zaneselvans: I mentioned this in the comment, but there are 2 reasons for using the node list here:
* It has `pd.NA` rather than `np.nan` which is required to use it to select values (this is a bit of a hack and I'm sure I can make the index work this way too. It's an index data type thing).
* The list of nodes doesn't know what set of columns it applies to, so the same list can be used to select either parent or calculation component records, without needing to rename anything.

cmgosnell: i did this bc simpler/seems easier to grab the outpts but i don't doubt the nulls are a problem
```
calc_idx = calc_explode.dropna(subset=["table_name", "xbrl_factoid"]).set_index(calc_cols).index
# An index of all parent nodes:
parent_idx = calc_explode.dropna(subset=["table_name_parent", "xbrl_factoid_parent"]).set_index(parent_cols).index

# Nodes that *only* appear as children / calculation components:
calc_only = calc_idx.difference(parent_idx)
```

zaneselvans:
I've added the `drop_duplicates()` and weirdly it seems like now I have more records than before (1044 total).

I tried selecting all of the "leaf" nodes (which have parent columns, but all NA calc columns) and it looks all of the leaves also currently lack any additional dimensions in their parents, which doesn't seem like what we would expect.

```py
new_calcs = MetadataExploder(
    table_names=table_names,
    clean_xbrl_metadata_json=clean_xbrl_metadata_json,
    calculation_components_xbrl_ferc1=calculation_components_xbrl_ferc1,
).calculations

new_calcs[new_calcs[calc_cols].isna().all(axis="columns")][parent_cols + calc_cols + ["weight"]].info()
```
```
<class 'pandas.core.frame.DataFrame'>
Int64Index: 239 entries, 15 to 782
Data columns (total 11 columns):
 #   Column                 Non-Null Count  Dtype 
---  ------                 --------------  ----- 
 0   table_name_parent      239 non-null    string
 1   xbrl_factoid_parent    239 non-null    string
 2   utility_type_parent    0 non-null      string
 3   plant_status_parent    0 non-null      string
 4   plant_function_parent  0 non-null      string
 5   table_name             0 non-null      string
 6   xbrl_factoid           0 non-null      string
 7   utility_type           0 non-null      string
 8   plant_status           0 non-null      string
 9   plant_function         0 non-null      string
 10  weight                 0 non-null      Int64 
dtypes: Int64(1), string(10)
memory usage: 22.6 KB
```

cmgosnell:
I'm having trouble sifting through all of the cruft i've now piled up in fixing the edge cases for these calculations.... so i'm going to list them all here to see if we can take a step back and do some simplification:
- **`reconcile_table_calculations` vs `calculation_components_xbrl_ferc1`**: we need to apply many of the same dimension treatments while reconciling the calcs.... which is fine it just means that there are two contexts in which the dimensions need to be managed. Mostly we are running the same methods in both places, but `reconcile_table_calculations` needs a little more prep/post cleaning bc it is only one table.
- **the corrections**. They don't always actually show up in the data, which is how we infer the dimensions. So we have to just add all of the possible dimensions during `reconcile_table_calculations`... not idea but it seems to be working. in `calculation_components_xbrl_ferc1` we have the full set of corrections in the data so we can just use the correction dims found in `table_dimensions_ferc1`
- **broke out `assign_parent_dimensions` from `add_dimension_total_calculations`**: these were previously done in one step but that was a little bit confusing.
- **differing dimensions of parent vs child factoids**: Previously during `assign_parent_dimensions` I was just saying child dimensions == parent dimensions. That was so simple & also wrong in some edge cases.
   - Here are the actual options:
       - _almost always_ a child fact has the same dimensions as the parent fact. 
       - _sometimes_ the child or parent fact has no dimensions (null only) while the other has dimension values
       - _sometimes_ the parent or the child fact's dimensions are a superset of the other fact.
    - This all means that we can't just do the simple child == parent.
    - When the child fact has non-null dimensions, we don't want to broadcast merge *all* of the available parent dimensions on every single child dimension, so I decided to merge with `table_dimensions_ferc1` using the dim column as a merge key. When the parental dimension is the superset of dimensions we need to outer merge because we want all of the parental factoids to have all of their dimensions. Ex: parent `fact_a` has a child `fact_b` that has `utility_type =="other"` while `fact_a` has both `"other"` and `"other2"`. Result should be *two* records linking parent `fact_a` w/ child `fact_b` on `utility_type =="other"` as well as the solo parent `fact_a` with `utility_type_parent =="other2"`
    - When the child fact as null dimensions, I decided to reuse `make_calculation_dimensions_explicit`. I think this side of this solution was more straightforward and hasn't been causing me headaches.

is a part of the fixes within `transform.ferc1` but i don't think is currently causing a headache:
- renaming factoids in the transform when it comes from a different table! (conceptually this was very simple to deploy... the hard part came in how i hadn't been preserving the order of `value_types` when there were more than one `wide_to_tidy` step being employed)
- 
-

zaneselvans:
Oooookay, I've got the whole explosion + filtering based on the calculation forest working again, but there are lots of issues involving the new dimensions that we'll need to untangle. Hopefully many of them can be fixed systematically rather than manually.

Not everything in this PR is pretty. I'm sure it can be improved. But do we want to get this stuff merged in and work on the fixes to the calculations with all the dimensions separate from these changes?

The setup I'm using right now to identify issues looks like this...

# Setup Inputs
```python
import importlib

from dagster import AssetKey

from pudl.etl import defs
from pudl.output.ferc1 import (
    Exploder,
    MetadataExploder,
    NodeId,
    XbrlCalculationForestFerc1,
)

tags_csv = (
    importlib.resources.files("pudl.package_data.ferc1")
    / "xbrl_factoid_rate_base_tags.csv"
)
tags_df = (
    pd.read_csv(tags_csv, usecols=["table_name", "xbrl_factoid", "in_rate_base"])
    .drop_duplicates()
    .dropna(subset=["table_name", "xbrl_factoid"], how="any")
    .astype(pd.StringDtype())
)

clean_xbrl_metadata_json = defs.load_asset_value(AssetKey("clean_xbrl_metadata_json"))
metadata_xbrl_ferc1 = defs.load_asset_value(AssetKey("metadata_xbrl_ferc1"))
calculation_components_xbrl_ferc1 = defs.load_asset_value(
    AssetKey("calculation_components_xbrl_ferc1")
)

explosion_args = {
    "income_statement_ferc1": {
        "root_table": "income_statement_ferc1",
        "table_names_to_explode": [
            "income_statement_ferc1",
            "depreciation_amortization_summary_ferc1",
            "electric_operating_expenses_ferc1",
            "electric_operating_revenues_ferc1",
        ],
        "calculation_tolerance": 0.27,
        "seeds": [
            NodeId(
                table_name="income_statement_ferc1",
                xbrl_factoid="net_income_loss",
                utility_type="total",
                plant_status=pd.NA,
                plant_function=pd.NA,
            ),
        ],
        "tags": tags_df,
    },
        "root_table": "balance_sheet_assets_ferc1",
        "table_names_to_explode": [
            "balance_sheet_assets_ferc1",
            "plant_in_service_ferc1",
        ],
        "calculation_tolerance": 0.81,
        "seeds": [
            NodeId(
                xbrl_factoid="assets_and_other_debits",
                utility_type=pd.NA,
                plant_status=pd.NA,
                plant_function=pd.NA,
            )
        ],
        "tags": tags_df,
    },
    "balance_sheet_liabilities_ferc1": {
        "root_table": "balance_sheet_liabilities_ferc1",
        "table_names_to_explode": [
            "balance_sheet_liabilities_ferc1",
            "retained_earnings_ferc1",
        ],
        "calculation_tolerance": 0.075,
        "seeds": [
            NodeId(
                table_name="balance_sheet_liabilities_ferc1",
                xbrl_factoid="liabilities_and_other_credits",
                utility_type=pd.NA,
                plant_status=pd.NA,
                plant_function=pd.NA,
            )
        ],
        "tags": tags_df,
    },
}
```

# Coordinating Function
```python
def exploded_table(
    root_table: str,
    table_names_to_explode: list[str],
    calculation_tolerance: float,
    seeds: list[NodeId],
    tags: pd.DataFrame,
):
    metadata_xbrl_ferc1 = defs.load_asset_value(
        AssetKey("metadata_xbrl_ferc1")
    )
    calculation_components_xbrl_ferc1 = defs.load_asset_value(
        AssetKey("calculation_components_xbrl_ferc1")
    )

    dfs_to_explode = {
        table: pd.read_sql(table, pudl_engine) for table in table_names_to_explode
    }

    exploder = Exploder(
        root_table=root_table,
        table_names=table_names_to_explode,
        metadata_xbrl_ferc1=metadata_xbrl_ferc1,
        calculation_components_xbrl_ferc1=calculation_components_xbrl_ferc1,
        seed_nodes=seeds,
        tags=tags,
    )
    return {
        "exploder": exploder,
        "exploded_meta": exploder.metadata_exploded,
        "exploded_calcs": exploder.calculations_exploded,
        "forest": exploder.calculation_forest,
        "leafy_meta": exploder.calculation_forest.leafy_meta,
        "root_calcs": exploder.calculation_forest.root_calculations,
        "exploded_data": exploder.boom(
            tables_to_explode=dfs_to_explode,
            calculation_tolerance=calculation_tolerance,
        ),
    }
```

# Run the Explosions
```python
test_explode = {}
for root_table in explosion_args:
    print(f"Exploding: {root_table}")
    test_explode[root_table] = exploded_table(**explosion_args[root_table])
```

# Display Results
```python
for root_table in test_explode:
    test_explode[root_table]["forest"].plot("full_digraph")
    test_explode[root_table]["forest"].plot("forest")
    
    print(f"\n ======== ORPHAN NODES: ========\n")
    display(pd.DataFrame(test_explode[root_table]['forest'].orphans))
    
    print(f"\n ======== PRUNED NODES: ========\n")
    display(pd.DataFrame(test_explode[root_table]['forest'].pruned))
```

# Outputs
## Income Statement
image
image

## Balance Sheet Assetse
image
image

## Balance Sheet Liabilities
image
image

Issue #2394, opened 2023-03-13
Update PUDL to be compatible with pandas 2.0


zaneselvans:
Work out dependency issues and update PUDL to work with [pandas 2.0](https://pandas.pydata.org/pandas-docs/version/2.0/whatsnew/index.html#version-2-0). Major improvements include the option to [back dataframes with Arrow arrays](https://datapythonista.me/blog/pandas-20-and-the-arrow-revolution-part-i), providing much richer data types and perfect compatibility with PyArrow generated Parquet file outputs.

```[tasklist]
- [x] Update `ferc-xbrl-extractor` to work with / allow pandas 2.0
- [x] Update `recordlinkage` to a version that works with pandas 2.0.
- [x] Extracting `eia860` (but **not** `eia923` or `eia861`) we get `AttributeError: 'XlsxWriter' object has no attribute 'save'` Seems to be related to streaming the DBF data as an Excel "file" in 2001-2003.
- [x] Failure in `df.replace()` within `fix_eia_na()` with `raise ValueError('cannot call `vectorize` on size 0 inputs '`
- [x] Census DP1 seems to have SQLAlchemy syntax issues
- [x] Replace deprecated `df.append()` with `pd.concat()` in `boiler_emissions_control_equipment_assn`
- [x] Convert `None` to numeic NA in `utility_plant_summary_ferc1.calculated_amount` when reconciling calcs
- [x] `clean_hourly_demand_matrix_ferc714` `pd.option_context("display.max_colwidth", -1)` need positive integer?
- [x] `fipsified_respondents_ferc714` failed merge on `object` and `datetime64[s]` columns
- [x] `income_statement_ferc1` divide by zero: had None rather than numeric NA in calculated amounts.
- [x] gen fuel by gen esc yearly: can't cast dtype('<M8[ns]') to dtype('<M8[s]')
- [x] gen fuel by gen esc monthly: cannot reindex on axis with duplicate labels. `plant_year` index in `distribute_annually_reported_data_to_months_if_annual()` is apparently non-unique.
- [x] `denorm_plants_steam_ferc1` `generate_rolling_avg` applies `mean()` to non-numeric columns.
- [x] `denorm_fuel_receipts_costs_eia923` `generate_rolling_avg` applies `mean()` to Categorical column.
- [x] in `plants_small_ferc1` transform: `ValueError: 'utility_id_ferc1' is both an index level and a column label, which is ambiguous.`
[...]

PR #2320, opened 2023-02-21
Update PUDL to pandas 2.0

zaneselvans:
# PR Overview

# PR Checklist
- [x] Merge the most recent version of the branch you are merging into (probably `dev`).
- [x] All CI checks are passing. [Run tests locally to debug failures](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#running-tests-with-tox)
- [ ] Make sure you've included good docstrings.
- [ ] For major data coverage & analysis changes, [run data validation tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#data-validation)
- [ ] Include unit tests for new functions and classes.
- [ ] Defensive data quality/sanity checks in analyses & data processing functions.
- [ ] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/latest/release_notes.html) and reference reference the PR and related issues.
- [ ] Do your own explanatory review of the PR to help the reviewer understand what's going on and identify issues preemptively.

PR #2769, opened 2023-08-03
Upgrade PUDL to Pandas 2.0

nelsonauner:
# PR Overview

Addresses https://github.com/catalyst-cooperative/pudl/issues/2394 and I will use that issue as a task tracker

# PR Checklist

- [x] Merge in `pandas-2` branch


src/pudl/helpers.py
```diff
@@ -995,14 +995,18 @@ def simplify_columns(df):
...
     """
-    df.columns = (
-        df.columns.str.replace(r"[^0-9a-zA-Z]+", " ", regex=True)
-        .str.strip()
-        .str.lower()
-        .str.replace(r"\s+", " ", regex=True)
-        .str.replace(" ", "_")
-    )
-    return df
+    # Do nothing, if empty dataframe (e.g. mocked for tests)
```
nelsonauner, line 998: Otherwise pandas 2.0 will fail when you attempt to call `.str` on a non-string

test/unit/io_managers_test.py
```diff
@@ -234,7 +234,12 @@ def test_handling_view_with_metadata(pudl_sqlite_io_manager_fixture):
 
     # read the view data as a dataframe
     input_context = build_input_context(asset_key=AssetKey(asset_key))
-    pudl_sqlite_io_manager_fixture.load_input(input_context)
```
nelsonauner, line 237: 🆘  I need some help here - I'm not a SQLAlchemy wizard 🧙 so I'm having a hard time understanding these mocked/fixtured SQLAlchemy objects

test/unit/transform/classes_test.py
```diff
@@ -347,19 +347,21 @@
         (5, pd.NA, pd.NA, -5.0, "barry"),
         (6, 2000, "2000-01-01", 231.1, "replace me"),
         (7, pd.NA, pd.NA, 101.10, pd.NA),
-        (8, 2012, "01-01-2020", 899.98, "another plant name"),
```
nelsonauner, line 350: pandas 2.0 really does not like this syntax. We can support it but would want to call out the specific row that needs to be cast in this format. Are we expected to support this for any of the input data?

zaneselvans: I don't know if this scenario (wildly varying date formats in the same column) comes up anywhere in our actual data -- and if it does, we'll need to do the `pd.to_datetime(format="mixed")` wherever it happens, so I suspect the test data should probably be altered to avoid this problem.

test/unit/io_managers_test.py
zaneselvans: I think this is probably a @bendnorman question.

zaneselvans (commented):
Is there a clear underlying reason why all of the `df.convert_dtypes()` are required in pandas 2 whereas they weren't in pandas 1? It'd just be nice to understand so we can work with it going forward.

I think the SQLAlchemy mocking question is probably in an area that @bendnorman worked on.

zaneselvans (commented):
It looks like there's an integration test failure that has to do with SQLAlchemy too.  If the only remaining issues are with SQLAlchemy, it *could* make sense to bump the SQLAlchemy version to `>=2.0,<2.1` and try to get everything working with the new version, rather than debugging it under the old version and then immediately thereafter switching to the new one and needing to debug many changes again.

test/unit/io_managers_test.py
bendnorman: `load_input()` is calling `pd.read_sql_table(table_name, con)`, which I think pre-pandas 2.0 worked for reading views but is now failing. We could do `pd.read_sql_query(f"select * from {table_name}", con)` to handle tables and views but I think `read_sql_table` does a better job inferring types. However, we might not run into datatype conversion issues given we are correcting dtypes using `Resource.enforce_schema()` in `load_input()`.

All that to say, I think we should xfail it for now. We aren't actually creating any views right now and we should probably do more testing to make sure the types are conserved.

test/unit/transform/classes_test.py
nelsonauner: Thanks Zane. Since I did indeed change the test data to avoid this problem, I'm interpreting your comment to mean that you approve of the way I handled this problem by changing the test data.

zaneselvans:
I'm pretty sure it doesn't have a green build -- for some reason it seems to have been waiting for approval to run the CI, even on the original pandas-2.0 branch. I think there are still many remaining issues. If you run `tox` locally you can see what goes wrong -- mostly I suspect it'll be in the integration tests (which run the full ETL).

zaneselvans:
It's getting a divide-by-zero error in `pudl.transform.ferc1.reconcile_table_calculations()` when it's being used in the transformation of the `income_statement_ferc1` table.

To replicate this error outside of the test environment for debugging and without running the full ETL I think you should be able to just extract the FERC databases and then attempt to materialize the `income_statement_ferc1` asset (and any of its upstream dependencies) in the Dagster UI. I suspect the error will come up in many tables, but that's the one it's tripping up on first in the integration tests.  You can also just try and materialize everything in the `norm_ferc1` asset group and see which tables encounter this problem.

Maybe somehow a value that was previously NA is now showing up as a zero, and that's breaking a division?

zaneselvans:
Thank you for all your work on this @nelsonauner! I dragged it over the finish line with many commits from your branch in #2320

nelsonauner:
Yay!

PR #2773, opened 2023-08-07
Enable `multi_asset` subsetting

bendnorman:
# PR Overview

# PR Checklist

### DIFF
--- a/src/pudl/etl/glue_assets.py
+++ b/src/pudl/etl/glue_assets.py
@@ -9,19 +9,14 @@ from pudl.metadata.classes import Package
 logger = pudl.logging_helpers.get_logger(__name__)
 
 
-# TODO (bendnorman): Currently loading all glue tables. Could potentially allow users
-# to load subsets of the glue tables, see: https://docs.dagster.io/concepts/assets/multi-assets#subsetting-multi-assets
-# Could split out different types of glue tables into different assets. For example the cross walk table could be a separate asset
-# that way dagster doesn't think all glue tables depend on generators_entity_eia, boilers_entity_eia.
-
-
 @multi_asset(
     outs={
-        table_name: AssetOut(io_manager_key="pudl_sqlite_io_manager")
+        table_name: AssetOut(io_manager_key="pudl_sqlite_io_manager", is_required=False)
         for table_name in Package.get_etl_group_tables("glue")
         #  do not load epacamd_eia glue assets bc they are stand-alone assets below.
         if "epacamd_eia" not in table_name
     },
+    can_subset=True,
     required_resource_keys={"datastore", "dataset_settings"},
 )
 def create_glue_tables(context):
@@ -46,9 +41,9 @@ def create_glue_tables(context):
     # Ensure they are sorted so they match up with the asset outs
     glue_dfs = dict(sorted(glue_dfs.items()))
 
-    return (
-        Output(output_name=table_name, value=df) for table_name, df in glue_dfs.items()
-    )
+    for table_name, df in glue_dfs.items():
+        if table_name in context.selected_output_names:
+            yield Output(output_name=table_name, value=df)
 
 
 #####################
--- a/src/pudl/extract/eia923.py
+++ b/src/pudl/extract/eia923.py
@@ -94,7 +94,7 @@ class Extractor(excel.GenericExtractor):
 
 
 # TODO (bendnorman): Add this information to the metadata
-eia_raw_table_names = (
+raw_table_names = (
     "raw_boiler_fuel_eia923",
     "raw_fuel_receipts_costs_eia923",
     "raw_generation_fuel_eia923",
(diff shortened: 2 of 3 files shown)


devtools/materialize_asset.py
```diff
@@ -19,33 +27,23 @@ def _parse():
...
-                config={
-                    "resources": {
-                        "dataset_settings": {
-                            "config": etl_fast_settings.dict(),
-                        },
-                    },
-                },
-            ),
-        ],
+    materialize(
```
bendnorman, line 35: @jdangerx @zschira is there a reason why y'all didn't use `materialize()` when you initially created this script?

src/pudl/extract/eia923.py
```diff
@@ -109,7 +109,7 @@ def get_dtypes(page, **partition):
     # from being extracted currently. When we update to a new DOI this problem will
     # probably fix itself. See comments on this issue:
     # https://github.com/catalyst-cooperative/pudl/issues/2448
-    # "raw_emissions_control_eia923",
+    # "raw_eia923__emissions_control",
```
zaneselvans, line 112: I think we've moved on to a new DOI by now if we want to turn this back on.

devtools/materialize_asset.py
jdangerx: Nope! This is better.

jdangerx: Though I think this whole script could probably be replaced with `dagster asset materialize`, like:

```bash
$ dagster asset materialize -m pudl.etl --select "raw_eia860__fgd_equipment"
```

Which works on this branch but fails with the `This AssetsDefinition does not support subsetting.` error on `dev`.

environments/conda-lock.yml
jdangerx, line 1: We can exclude these conda-lock files from rendering by default by marking them as 'generated files': https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

src/pudl/etl/glue_assets.py
```diff
@@ -46,9 +41,9 @@ def create_glue_tables(context):
     # Ensure they are sorted so they match up with the asset outs
     glue_dfs = dict(sorted(glue_dfs.items()))
 
-    return (
-        Output(output_name=table_name, value=df) for table_name, df in glue_dfs.items()
-    )
+    for table_name, df in glue_dfs.items():
```
jdangerx, line 44: Non-blocking - would a generator expression work here instead? 

```python
return (
    Output(output_name=table_name, value=df)
    for table_name, df in glue_dfs.items()
    if table_name in context.selected_output_names
)
```
? I guess that's not really much better, is it.

jdangerx (changes requested):
Whee, I'm glad this is finally getting into `dev`!

Questions I'd like to see addressed before merging (and once you address them, feel free to dismiss this review and merge!)
I think we can just get rid of `materialize_asset.py` completely. What do you think?

There are still lots of `multi_asset`s that don't support subsetting on this branch - how did you choose what to change and what to not change?

Non-blocking comment: 

It would also be good to follow up on Zane's comment, about the new DOI maybe fixing things.

devtools/materialize_asset.py
bendnorman: True but I'm not sure you can feed the CLI command run config like you can with the `materialize()` function which is helpful to running subsets of years:

```
Usage: dagster asset materialize [OPTIONS]

  Execute a run to materialize a selection of assets

Options:
  -a, --attribute TEXT          Attribute that is either a 1) repository or job or 2) a function that returns a
                                repository or job
  --package-name TEXT           Specify Python package where repository or job function lives
  -m, --module-name TEXT        Specify module where dagster definitions reside as top-level symbols/variables and
                                load the module as a code location in the current python environment.
  -f, --python-file PATH        Specify python file where dagster definitions reside as top-level symbols/variables
                                and load the file as a code location in the current python environment.
  -d, --working-directory TEXT  Specify working directory to use when loading the repository or job
  --select TEXT                 Asset selection to target  [required]
  --partition TEXT              Asset partition to target
  -h, --help                    Show this message and exit.
```

Also, can we use the vs code debugger with the CLI command?

src/pudl/etl/glue_assets.py
bendnorman: I'm not sure. I am just working off of the [dagster docs examples](https://docs.dagster.io/concepts/assets/multi-assets#subsetting-multi-assets). IDK if dagster cares if its a return or yield statement.

jdangerx:
I think these are all things we have fixed somehow somewhere in the last two years. Thanks for everything @bendnorman :smiling_face_with_tear:

zaneselvans:
The error went away because Dagster now gives you the option of also materializing all of the upstream assets that are part of a multi-asset, even if they aren't necessary, and we do end up doing a fair bit of that in development, rather than being able to split them out.  But mostly it's the small static tables that are bundled together, and they're fast, so it's not too annoying.

PR #2782, opened 2023-08-10
Eia861 2022

aesharpe:
This PR adds in EIA861 early release data.
The only quirks were that the directory name was different from the zipfile name, so the file_map includes a prefix of `f8612022er/` for each of the files, and a change in skip rows for the reliability table in 2021. It also adds an encoding step for the reliability table to avoid foreign key failures.

The former might change for the regular release data. See issue #2780 (linked)

There are no tests for 861 so that's all we've got.


e-belfer:
I think you'll have to add `reliability_eia861` to the exceptions list for the `momentary_interruptions_eia` table to fix this last error.

`FAILED test/integration/etl_test.py::test_pudl_engine - pudl.io_managers.ForeignKeyErrors: Foreign key error for table: reliability_eia861 -- momentary_interruptions_eia (code) -- on rows [18738]`

zaneselvans (approved):
Looks good assuming the CI passes successfully -- we should look at some of the stuff that's downstream too, like the service territory tables.

PR #2789, opened 2023-08-14
Apply new naming convention to raw and core intermediate assets

bendnorman:
# PR Overview

# PR Checklist


src/pudl/extract/eia860.py
```diff
@@ -69,29 +69,29 @@ def get_dtypes(page, **partition):
...
-    "raw_generator_existing_eia860",
-    "raw_generator_proposed_eia860",
-    "raw_generator_retired_eia860",
-    "raw_multifuel_existing_eia860",
-    "raw_multifuel_retired_eia860",
-    "raw_ownership_eia860",
-    "raw_plant_eia860",
-    "raw_stack_flue_equipment_eia860",
-    "raw_utility_eia860",
+    "raw_eia860__boiler_cooling",
```
jdangerx, line 72: Looking at a whole stack of these, I'm shocked at how much nicer it is to have the double underscore as a sort of "column separator!"

src/pudl/extract/eia923.py
```diff
@@ -129,7 +129,7 @@ def extract_eia923(context):
 
     # create descriptive table_names
     eia923_raw_dfs = {
-        "raw_" + table_name + "_eia923": df for table_name, df in eia923_raw_dfs.items()
+        "raw_eia923__" + table_name: df for table_name, df in eia923_raw_dfs.items()
```
jdangerx, line 132: Nice to not have the table name in the middle, too :)

src/pudl/io_managers.py
```diff
@@ -658,6 +658,8 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
         ferc1_settings = context.resources.dataset_settings.ferc1
 
         table_name = self._get_table_name(context)
+        # Remove preceeding asset name metadata
+        table_name = table_name.replace("raw_ferc1_dbf__", "")
```
jdangerx, line 662: I wonder if we want to write a utility function that gets the base table name - do you think that would be helpful in the future? 

```python
def get_base_table_name(s): # or... strip_metadata(s)?
    return s.split('__', 1)[-1]
```

I guess at least it would get rid of the need for this comment 😅

jdangerx (approved):
I think this looks good! The new consistent format is pretty nice, and these are all non-user-facing, so I think it's pretty low risk. If we run into something that fell through the cracks somehow, it should be an easy fix. So long as the ETL runs I think we're good to merge!

However! I have this inkling of a thought, that a table name should be a class with layer/source/type/intermediate attrs and a nice `__str__` implementation that serializes along the lines of our naming convention.

Benefits:
* encodes our design decisions in the code! very clear to code readers what the different pieces of a name are
* we don't have to write annoying component extractors, etc.
* less code duplication!
* unlikely to accidentally typo stuff

Downsides:
* extra work now
* `str(table_name)` calls wherever we interface with Dagster - though I suppose we could write some wrapper code if we really want to do this.
  * how does this interact with the "asset keys default to function names" thing? we'd have to add explicit keys to each asset, I think.
* is this an extra set of concepts to understand? I guess you have to understand the concepts anyways, it's just whether those concepts are implicit in the way we interact with these strings vs. whether the concepts are explicitly in the code.

I think I'm talking myself into the position of "because Dagster really cares a lot that the asset names are strings, and because of the asset ID <> function and parameter name magic that Dagster does, we should probably keep them as strings." But maybe it's worth making some utility functions that build and disassemble these strings for us - what do you think?

bendnorman:
I like the idea of creating a class for asset names that can perform some validation and serialization. I think we'd have to use the asset `name` parameter:

```python
@asset(name=str(AssetName(layer="core", source="eia860", type="entity", name="generators")))
def placeholder_asset_func_name():
    ...

@asset(
     ins = {"generators": AssetIn(key=str(AssetName(layer="core", source="eia860", type="entity", name="generators"))},
     name = str(AssetName(layer="core", source="eia860", type="downstream", name="asset"))
)
def downstream_asset(generators):
   ...
```

We could also use [multi-component asset keys](https://docs.dagster.io/concepts/assets/software-defined-assets#multi-component-asset-keys) to construct the asset name. Then we could have an `AssetName` class validate and construct the full asset name within the IO Managers.

```python
from dagster import asset, AssetIn

@asset(key_prefix=["core", "eia860"])
def entity_generators():
    ...

@asset(
   ins={"entity_generators": AssetIn(key_prefix=["core", "eia860"])},
   key_prefix=["output", "eia860"]
)
def monthly_generators(entity_generators):
    ...

entity_generators.key # AssetKey(['core', 'eia860', 'entity_generators'])

# The `AssetKey`s values could then be passed into an AssetName class in the IO managers for validation.
```

I think both options might add too much complexity. In option 1, we'd have to figure out what to name the function name given the actual asset name will be specified in the `@asset` decorator. Both option 1 and 2 would require contributors to construct the `ins` dictionary. I find typing out the name in the function decorator a bit easier.

Another option would be to use the full asset names as the function names, then in the IO managers, use regexes to parse out the layer, source, and asset type for validation. This is maybe more ergonomic for contributors creating assets but might be a brittle method for validating asset names. Like, why do regex parsing when we could just create data structure like you suggested.

Maybe there's a simpler solution I'm missing. What do you think @jdangerx ?

jdangerx:
My sense is that, to hook into the Dagster machinery, we really want the asset names to be things we can also use as variable names. Otherwise things get pretty hairy pretty quickly - we sort of have to write the opposite of all of Dagster's function/variable/asset matching logic.

What if we subclassed `str` directly...? Is that a terrible idea? We could add some validation in `__new__`...

```python
>>> class TableName(str):
...     def __new__(cls, string):
...         if "__" not in string:
...             raise ValueError()
...         instance = super().__new__(cls, string)
...         return instance
... 
>>> TableName("abc")
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 4, in __new__
ValueError
>>> TableName("a_b__c")
'a_b__c'
>>> isinstance(TableName("a_b__c"), str)
True
```

We could make a factory function that stitches the various fields together into a `TableName`... 🤷 . Seems convoluted.

All this speculation is fun, but I think this PR is mergeable as is. The most straightforward path towards "some" validation is adding a few utility functions, I think, but even that isn't necessary for this PR IMO.

bendnorman:
I agree the easiest solution would be write a utility function to do some basic validation: is there a double underscore, are the layer and data source valid? I think this could live in the IO managers. I'm comfortable merging this in as is though.

PR #2792, opened 2023-08-16
Add support for running ETL on google batch

rousik:
This is set of changes that will allow us to run the full (or partial) ETL using batch service instead of running on dedicated VMs.


bendnorman:
Should we do a more thorough comparison of [Dagster Cloud](https://dagster.io/pricing) and Batch before committing to Batch? We'll have to double-check how the pricing differs, but a pro of Dagster Cloud is the dagster infrastructure is fully managed, and we'll be able to debug runs using the UI.

PR #2795, opened 2023-08-17
Fix anonymous constraints

jdangerx:
We often run into issues with creating new foreign keys in existing tables. [Alembic seems to complain frequently - this is because SQLite allows anonymous constraints by default.](https://alembic.sqlalchemy.org/en/latest/batch.html#dropping-unnamed-or-named-foreign-key-constraints)

On this branch, I can make a new "fake" table that has some foreign key rules, and the migration will name the new constraints - which means you can actually run `alembic upgrade`/`alembic downgrade` without issues. On `dev`, when I make this table and try to run `alembic revision --autogenerate` and then `alembic upgrade head`, I get 

```
ValueError: Constraint must have a name
```

Here's the table I tested with:
```diff
diff --git a/src/pudl/metadata/resources/eia.py b/src/pudl/metadata/resources/eia.py
index 0812ad54..287e727d 100644
--- a/src/pudl/metadata/resources/eia.py
+++ b/src/pudl/metadata/resources/eia.py
@@ -4,6 +4,17 @@ from typing import Any
 from pudl.metadata.codes import CODE_METADATA
 
 RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "fake": {
+        "description": "fake table",
+        "schema": {
+            "fields": ["report_date"],
+            "primary_key": ["report_date"],
+            "foreign_key_rules": {"fields": [["report_date"]]},
+        },
+        "sources": ["eia860"],
+        "etl_group": "static_eia",
+        "field_namespace": "eia",
+    },
     "balancing_authorities_eia": {
         "description": "A coding table describing balancing authorities in EIA-860 and EIA-923.",
         "schema": {
```

And [here's a comparison](https://gist.github.com/jdangerx/541db1c90c0ec22084737df583366d9d) of the migrations generated on `dev` vs. `fix-anonymous-constraints` - you can see that the `dev` version passes `None` for constraint names.

The actual changes are just two:
* add a naming convention as in the docs - I *think* adding them to this `MetaData()` call should be enough, but there's also others in e.g. `Resource.to_sql()` that I think we don't need to change?
* Update the check constraints generator to make *some* sort of name for the check constraint. I couldn't find a single call to `sa.CheckConstraint()` in our migrations, though, so maybe that's a moot point so far.


src/pudl/metadata/classes.py
```diff
@@ -735,7 +735,7 @@ def to_sql(  # noqa: C901
         return sa.Column(
             self.name,
             self.to_sql_dtype(),
-            *[sa.CheckConstraint(check) for check in checks],
+            *[sa.CheckConstraint(check, hash(check)) for check in checks],
```
zaneselvans, line 738: What's going on here? I can't tell what role `hash(check)` is supposed to be playing based on the SQLAlchemy docs.

zaneselvans (commented):
It seems like the named constraints will only show up if someone is looking at the SQL used to create the DB, and will avoid an annoyance in the migrations which is great, with a very small change, to use basically the default naming convetions.

I'm confused about what's going on with the `hash(check)` though. It didn't make sense to me based on looking at the SQLAlchemy docs.

zaneselvans (approved):

zaneselvans:
Argh! The dreaded code 143.

PR #2798, opened 2023-08-19
Allow a mix of Zenodo sandbox & production DOIs

zaneselvans:
# PR Overview

Okay I did this off the clock since it has been driving me a little bit nuts and I wanted to do something technical that felt easy and satisfying as a break from the never-ending saga of #2016.

Historically we've required that all Zenodo DOIs in the datastore come either from the Sandbox or the Production server, which makes testing a single new archive on its own a hassle, and adds complexity across the whole application with switches for sandbox vs. not-sandbox data sources.

This commit removes this requirement, and allows a mix of sandbox and production DOIs to be used in development.

I also removed some very sparse documentation about how to create an archive in the Datastore by hand, which I think was very old and probably no longer supported and certainly not being tested, since it seemed likely to confuse and frustrate anyone who actually tried to do it.

There's a unit test which checks that all DOIs are production, rather than sandbox to make it difficult to accidentally check in code that refers to unofficial input data.

# PR Checklist


docs/dev/datastore.rst
```diff
@@ -64,28 +66,13 @@ archival and versioning of datasets. See the `documentation
...
-   * ``remote_url`` with the same url as the ``path``
-   * ``name`` of the file
-   * ``hash`` with the md5 hash of the file
-   * ``parts`` a set of key / value pairs defining additional attributes that
-     can be used to select a subset of the whole datapackage. For example, the
-     ``epacems`` dataset is partitioned by year and state, and
-     ``"parts": {"year": 2010, "state": "ca"}`` would indicate that the
-     resource contains data for the state of California in the year 2010.
-     Unpartitioned datasets like the ``ferc714`` which includes all years in
-     a single file, would have an empty ``"parts": {}``
```
zaneselvans, line 91: I removed this because I think it's quite old and I don't think it would actually work anymore -- we really expect all the input data to come from `pudl-archiver` outputs. Anyone trying to create an archive by hand will probably just get frustrated.

README.rst
```diff
@@ -124,7 +124,7 @@ Want access to all the published data in bulk? If you're familiar with Python
 and `Jupyter Notebooks <https://jupyter.org/>`__ and are willing to install Docker you
 can:
 
-* `Download a PUDL data release <https://sandbox.zenodo.org/record/764696>`__ from
```
zaneselvans, line 127: Not sure why the sandbox was being referenced in the README... but I pointed this at the real data release concept DOI on the production server.

test/unit/workspace/datastore_test.py
```diff
@@ -235,35 +235,30 @@ def setUp(self):
...
+        identified.
+        """
+        ds = datastore.ZenodoFetcher()
         self.assertTrue(ds.get_known_datasets())
         for dataset in ds.get_known_datasets():
+            doi = ds.get_doi(dataset)
+            self.assertFalse(
+                re.fullmatch(r"10\.5072/zenodo\.[0-9]{5,10}", doi),
+                msg=f"Zenodo sandbox DOI found for {dataset}: {doi}",
+            )
```
zaneselvans, line 252: This ensures we don't accidentally leave any sandbox DOIs in the codebase.

src/pudl/workspace/datastore.py
```diff
@@ -167,54 +167,46 @@ class ZenodoFetcher:
...
+        "ferc1": "10.5281/zenodo.7314437",
+        # "ferc1": "10.5072/zenodo.1070868",
+        "ferc2": "10.5281/zenodo.8006881",
+        # "ferc2": "10.5072/zenodo.1188447",
+        "ferc6": "10.5281/zenodo.7130141",
+        # "ferc6": "10.5072/zenodo.1098088",
+        "ferc60": "10.5281/zenodo.7130146",
+        # "ferc60": "10.5072/zenodo.1098089",
+        "ferc714": "10.5281/zenodo.7139875",
+        # "ferc714": "10.5072/zenodo.1098302",
```
zaneselvans, line 196: At some point I think we agree the DOIs should come out of the codebase and go into a settings file, but I'm not trying to do that in this PR.  I left the sandbox DOIs here and commented out for easy reference if someone wants to test out one of them, or look up which Zenodo archive is referenced in the sandbox.

src/pudl/workspace/datastore.py
```diff
@@ -240,16 +236,24 @@ def _fetch_from_url(self, url: str) -> requests.Response:
...
+
+        doi_prefix = match.groups()[0]
+        zenodo_id = match.groups()[1]
+        if doi_prefix == "10.5072":
+            api_root = self.API_ROOT["sandbox"]
+        elif doi_prefix == "10.5281":
+            api_root = self.API_ROOT["production"]
+        else:
+            raise ValueError(f"Invalid Zenodo DOI: {doi}")
+        return f"{api_root}/deposit/depositions/{zenodo_id}"
```
zaneselvans, line 252: I'm sure there's a more eloquent way of switching between production and sandbox on a per-dataset basis (rather than the whole instance of the class being tied to one or the other) but this seems relatively self-contained and not terrible for the moment.

src/pudl/workspace/datastore.py
```diff
@@ -463,25 +457,19 @@ def parse_command_line():
...
         help="Download the specified dataset only. See below for available options. "
-        "The default is to download all, which may take an hour or more."
-        "speed.",
+        "The default is to download all datasets, which may take hours depending on "
+        "network speed.",
     )
     parser.add_argument(
         "--pudl_in",
-        help="Override pudl_in directory, defaults to setting in ~/.pudl.yml",
+        help="Input directory to use, overridng the $PUDL_INPUT environment variable.",
```
zaneselvans, line 465: I think there are some other lingering references to `.pudl.yml` floating around that we should chase down now that we've switched over to using `$PUDL_INPUT` and `$PUDL_OUTPUT` entirely, but that's for another PR.

src/pudl/workspace/datastore.py
zaneselvans: I messed around with creating a DOI class:

```py
class ZenodoDoi(BaseModel):
    """A class defining useful validations and methods for working with Zenodo DOIs."""

    doi: constr(regex=r"^10\.(5072|5281)/zenodo\.[\d]+$")  # noqa: F722

    def __str__(self: Self) -> str:
        """String representation of the DOI"""
        return self.doi

    @property
    def is_prod(self: Self) -> bool:
        """Return True if DOI is from Zenodo production server, False otherwise."""
        if self.doi.startswith("10.5281/zenodo"):
            return True
        else:
            assert self.doi.startswith("10.5072/zenodo")
            return False

    @property
    def token(self: Self) -> str:
        """Zenodo read-only personal access token corresponding to this DOI.

        Zenodo tokens recorded here should have read-only access to our archives.
        Including them here is correct in order to allow public use of this tool, so
        long as we stick to read-only keys.
        """
        # Read-only personal access tokens for pudl@catalyst.coop:
        if self.is_prod:
            return "KXcG5s9TqeuPh1Ukt5QYbzhCElp9LxuqAuiwdqHP0WS4qGIQiydHn6FBtdJ5"
        else:
            return "qyPC29wGPaflUUVAv1oGw99ytwBqwEEdwi4NuUrpwc3xUcEwbmuB4emwysco"

    @property
    def zenodo_id(self: Self) -> str:
        """The Zenodo deposition ID, extracted from the DOI."""
        match = re.search(r"(10\.5072|10\.5281)/zenodo.([\d]+)", self.doi)
        return match.groups()[1]

    @property
    def api_root(self: Self) -> HttpUrl:
        """Return appropriate production or sandbox Zenodo API root URL."""
        if self.is_prod:
            return "https://zenodo.org/api"
        else:
            return "https://sandbox.zenodo.org/api"

    @property
    def url(self: Self) -> HttpUrl:
        """Zenodo URL corresponding to this DOI."""
        return f"{self.api_root}/deposit/depositions/{self.zenodo_id}"
```

src/pudl/workspace/datastore.py
rousik: I wonder if turning this into pydantic settings object that could read from env variables (e.g. `PUDL_FERC1_DOI`) could be a good way to pass sandbox values here, with production defaults as... well, defaults :)

src/pudl/workspace/datastore.py
```diff
@@ -240,16 +236,24 @@ def _fetch_from_url(self, url: str) -> requests.Response:
 
     def _doi_to_url(self, doi: str) -> str:
```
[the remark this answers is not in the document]
rousik: Hm, looks like you've already attempted this in the other PR that comes my way?

test/unit/workspace/datastore_test.py
```diff
@@ -235,35 +235,30 @@ def setUp(self):
...
+                msg=f"Zenodo production DOI for {dataset} is {doi}",
             )
 
     def test_get_known_datasets(self):
         """Call to get_known_datasets() produces the expected results."""
         self.assertEqual(
-            sorted(datastore.ZenodoFetcher.DOI["production"]),
+            sorted(datastore.ZenodoFetcher.DOI),
             self.fetcher.get_known_datasets(),
         )
```
rousik, line 263: Huh, I'm not sure if the test does what it says it should.

src/pudl/workspace/datastore.py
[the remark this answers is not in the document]
zaneselvans: I tried something like this in the other PR but it wasn't very satisfying. I think there's a simpler way to integrate some of those mechanics directly into the `ZenodoFetcher` class here.

src/pudl/workspace/datastore.py
zaneselvans: I think we need to store the DOIs in a file in the repo (which could be used to populate env vars) so we can look them up for cache invalidation, and easily edit them eventually with PRs when new archives become available.

But for this PR I just want to get to where we can have mixed sandbox/production DOIs to make integrating new archives by hand this fall easy.

src/pudl/workspace/datastore.py
zaneselvans: I don't know. If anywhere I think it would be in the tests. I grepped a bit and didn't see it anywhere.

src/pudl/workspace/datastore.py
```diff
@@ -29,6 +30,7 @@
 # long as we stick to read-only keys.
 
 PUDL_YML = Path.home() / ".pudl.yml"
+ZenodoDOI = constr(regex=r"(10\.5072|10\.5281)/zenodo.([\d]+)")
```
zaneselvans, line 33: This defines the type, but doesn't actually do any validation. I use the `ZenodoDOI.validate()` method in the `ZenodoFetcher.__init__()` to check.

zaneselvans:
Not sure what's going on with the unit test failure here, it works fine for me locally...

zaneselvans:
@rousik I am stumped why the datastore unit tests are failing in CI but working fine locally. Is there anything that looks obviously fishy to you in the test? Does it work locally for you? The huge diff that it's reporting between the mocked and expected datapackage descriptor makes me wonder if it's getting the real datapackage descriptor rather than the mocked one somehow. But I don't know why that would happen in CI but not locally.

```bash
pytest test/unit/workspace/datastore_test.py
```

src/pudl/workspace/datastore.py
zaneselvans: Using a `BaseSettings` model was easy!  Now the DOIs all get validated automatically by Pydanic, and they can optionally be set using environment variables too.

src/pudl/workspace/datastore.py
```diff
@@ -154,106 +152,127 @@ def get_json_string(self) -> str:
         return json.dumps(self.datapackage_json, sort_keys=True, indent=4)
 
 
+class ZenodoDoiSettings(BaseSettings):
+    """Digital Object Identifiers pointing to currently used Zenodo archives."""
+
+    # Sandbox DOIs are provided for reference
```
jdangerx, line 158: This works, and it makes sense to me to keep the sandbox DOIs around somewhere. It'd be nice if we could keep them somewhere that's actually encoded, as opposed to in a comment. But, we still want to be able to swap around "which datasets come from which environment" without accidentally committing a change that sends people to the sandbox randomly.

I think we sort of have these moving pieces at play, conceptually:

* there are multiple environments which use different DOIs for each dataset
* each dataset is drawn from one environment at a time, defaulting to prod
* the environment affects the HTTP access token as well as the base URL

Right now, the environment is encoded as "does the DOI include `10.5072` or `10.5281`?" or "does the URL include `sandbox`?". We could easily turn it into an explicit part of the data model, which will help us avoid the issue of committing sandbox changes completely. I'm thinking something like this:

```python
class ZenodoEnv(enum.Enum):
  SANDBOX: "sandbox"
  PRODUCTION: "production" 
  
class ZenodoEnvSettings(BaseModel):
  dois: ZenodoDoiSettings
  token: str
  base_url: HttpUrl
  
  def get_doi(self, dataset):
    return self.dois.__getattribute__(dataset)

class ZenodoFetcher:
  target_envs = ZenodoTargetEnvs.from_yaml("some_config.yaml")
  envs = {ZenodoEnv.SANDBOX: ZenodoEnvSettings(...), ZenodoEnv.PRODUCTION: ZenodoEnvSettings(...)}
  
  def _fetch_from_url(self, url, env):
    token = self.envs[env].token
    ...

  def get_resource(self, res: PudlResourceKey) -> bytes:
    env = self.target_envs[res.dataset]
    doi = self.envs[env].get_doi(res.dataset)
    ...
```

Do you think the benefits of the refactor (ZenodoFetcher not having to think about inter-env differences, lower chance of messing up the env config) are worth the time of actually doing it?

src/pudl/workspace/datastore.py
```diff
@@ -267,26 +286,18 @@ def get_descriptor(self, dataset: str) -> DatapackageDescriptor:
...
         return self._descriptor_cache[doi]
 
-    def get_resource_key(self, dataset: str, name: str) -> PudlResourceKey:
-        """Returns PudlResourceKey for given resource."""
-        return PudlResourceKey(dataset, self._dataset_to_doi[dataset], name)
-
-    def get_doi(self, dataset: str) -> str:
-        """Returns DOI for given dataset."""
-        return self._dataset_to_doi[dataset]
+    def get_resource_key(self: Self, dataset: str, name: str) -> PudlResourceKey:
```
jdangerx, line 289: This isn't used anywhere but the unit test - should we make this a private method?

test/integration/zenodo_datapackage_test.py
```diff
@@ -8,24 +8,19 @@
...
-        """All datasets point to valid descriptors with 1 or more resources."""
-        ds = Datastore(sandbox=True)
-        for dataset in ds.get_known_datasets():
-            desc = ds.get_datapackage_descriptor(dataset)
-            assert list(desc.get_resources())
-
-    @pytest.mark.xfail(
-        raises=(MaxRetryError, ConnectionError, RetryError, ResponseError)
+        raises=(
+            MaxRetryError,
```
jdangerx, line 15: We could mock out the Zenodo interaction, though I guess a lot of the funky logic is "are we using the right access token for the URL for the dataset" so maybe we just keep this around. How flaky does this turn out to be?

test/unit/workspace/datastore_test.py
```diff
@@ -302,7 +300,7 @@ def test_get_resource_key(self):
     def test_get_resource_key_for_unknown_dataset_fails(self):
         """When get_resource_key() is called for unknown dataset it throws KeyError."""
         self.assertRaises(
-            KeyError, self.fetcher.get_resource_key, "unknown", "blob.zip"
+            AttributeError, self.fetcher.get_resource_key, "unknown", "blob.zip"
```
jdangerx, line 303: I *think* we usually use the `pytest` specific code when catching errors in test:

```
with pytest.raises(AttributeError, match=r"some regex"):
    self.fetcher.get_resource_key("unknown", "blob.zip")
```

Though, as noted above, we seem to only use `get_resource_key()` in test, so maybe we should just privatize it and make sure we test its observed behavior through the public interface.

src/pudl/workspace/datastore.py
jdangerx: I took a look around and also only see "manually run this script to cache datastore locally." So my vote is to use the env vars completely.

test/integration/zenodo_datapackage_test.py
```diff
@@ -8,24 +8,19 @@
 
 
 class TestZenodoDatapackages:
-    """Ensure production & sandbox Datastores point to valid datapackages."""
+    """Ensure all DOIs in Datastore point to valid datapackages."""
```
jdangerx, line 11: It might be nice to also test that we can download a couple of the resources that are actually pointed at by the `datapackage.json`.

zaneselvans: It would, and we used to do this, but all of the "Actually download something from Zenodo" tests have ended up being so flaky that we've marked them `XFAIL` since they were routinely breaking the tests for no real reason.

src/pudl/workspace/datastore.py
zaneselvans: Sure, that makes sense.

src/pudl/workspace/datastore.py
zaneselvans: I agree that this would be nice, but I'd prefer not to go all the way down the path of configuring the DOIs with environment variables and a config file right now. I'm doing this as an open source contribution because it's a low priority but it's annoying me. We change DOIs rarely, and just being able to cut-and-paste by hand a mix of sandbox and production DOIs will be a big ergonomic improvement. The sandbox DOIs also aren't particularly stable -- sometimes an update to the archivers necessitates starting a new archive lineage. So formally encoding them is of somewhat less utility than the production DOIs. And also every time we want to mess around with a new archive we're going to have to manually update DOIs anyway.

I think the right time to do this work might be when we have someone willing to pay for the (semi-) automation of creating fresh PRs from newly minted archives to attempt running through the ETL, since parsing / editing a YAML file that contains the DOIs will be cleaner and easier to automate.

src/pudl/workspace/datastore.py
zaneselvans: Okay, I'll remove this and see if everything keeps working!

src/pudl/workspace/datastore.py
jdangerx: Sure, if we're mostly just In A Hurry this totally works :)

src/pudl/workspace/datastore.py
zaneselvans: But more generally I think @rousik agreed to take on the "scour the repo for remaining references to `.pudl.yml` and friends" task.

test/integration/zenodo_datapackage_test.py
jdangerx: Yeah, I can definitely see how that would happen. Maybe someday we can use a "flaky test re-runner" like https://github.com/box/flaky.

test/unit/workspace/datastore_test.py
zaneselvans: We've tried to standardize on using `pytest` across the board but @rousik created these test classes originally using `unittest` and I was hoping not to refactor more than necessary to get this minor ergonomic improvement in.

Given that this method is literally used... nowhere but the tests, is there any reason not to just delete it entirely? Seems odd.

jdangerx: Let's trash it!

zaneselvans: I've removed `get_resource_key()` and the related tests, and added another test which tries to get a DOI for a dataset that doesn't exist to verify that it fails as expected.

src/pudl/workspace/datastore.py
zaneselvans: DELETED!

test/integration/zenodo_datapackage_test.py
zaneselvans: Something like that pytest flaky plugin seems like it would be good! But that one looks a bit stale.

IIRC much of the flakiness wasn't "re-run immediately and things are okay" it was more like "network is out for an hour so you'll just keep failing until it comes back" which seems weird given that GitHub and CERN should both be online kind of always but... 🤷🏼

jdangerx (approved):

PR #2805, opened 2023-08-24
Make leafy balance sheet assets & liabilities data

cmgosnell:
# PR Overview

Some final fixes that are specific to the balance sheet asset calculation tree, allowing it to be used for annotating and filtering the exploded balance sheet asset & liability data tables.

```[tasklist]
- [x] remove all stepparents nodes even when they are not dim = total
- [x] add a hacky way to remove specific node to get rid of stepchildren.
- [x] resolve conflicting node weights
- [x] apply weights and tags to the tree
- [x] make leafy data for balance sheet assets
- [x] make leafy data for balance sheet liabilities
- [x] add a check ensuring we haven't dropped any nodes with negative weights.
- [x] add a check that the fraction of conflicting weights is below some threshold.
- [x] move code specific to assigning attributes out of `forest()` into `set_tree_attributes()`
- [x] add an output method that allows us to inspect the full calculation tree as a table / dataframe.
- [x] rename `pruned_forest` to `annotated_forest` for internal consistency
- [x] add a check ensuring we haven't dropped any tagged nodes
- [x] add a check that we don't have conflicting tags that are clobbering each other
- [x] consolidate maximum calculation error thresholds into a single data structure / parameter
- [x] add "row literals" and/or Form 1 page + row numbers into `metadata_xbrl_ferc1` to aid in review
- [x] Model calculation weights as edge attributes rather than child node attributes.
- [x] Evaluate need for `passthroughs` and the infrastructure around them, given additional dimensions
- [ ] Add inter-layer weights into `forest_as_table` (read from edges, not nodes)
```

# PR Checklist


zaneselvans:
For reasons I don't yet understand, I am getting non-deterministic behavior in the graph pruning process. For both `balance_sheet_assets_ferc1` and `income_statement_ferc1` whether or not we get a tree at the end depends on something random, and both of them sometimes come out as trees with no multi-parent nodes.

Running the `balance_sheet_asset_ferc1` explosion process 10 times, I got the following sets of stepchildren:
```python
[
    [
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_acquisition_adjustment', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_other_utility_plant_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='depreciation_utility_plant_in_service', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='construction_work_in_progress', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='construction_work_in_progress', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='depreciation_utility_plant_in_service', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_acquisition_adjustment', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='depreciation_utility_plant_in_service', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_held_for_future_use', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_held_for_future_use', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_held_for_future_use', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_other_utility_plant_utility_plant_in_service', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='abandonment_of_leases', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_other_utility_plant_utility_plant_in_service', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_other_utility_plant_utility_plant_in_service', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='depreciation_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='abandonment_of_leases', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='abandonment_of_leases', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='construction_work_in_progress', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_acquisition_adjustment', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='abandonment_of_leases', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA)
    ],
    [
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='other3', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_other_utility_plant_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='other2', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_plant_acquisition_adjustment', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='depreciation_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='abandonment_of_leases', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA)
    ],
    [],
    [
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_plant_purchased_or_sold', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_other_utility_plant_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='abandonment_of_leases', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_plant_purchased_or_sold', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_completed_construction_not_classified', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='depreciation_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_property_under_capital_leases', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_property_under_capital_leases', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_plant_acquisition_adjustment', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_completed_construction_not_classified', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA)
    ],
    [
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_plant_purchased_or_sold', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_property_under_capital_leases', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_plant_purchased_or_sold', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_completed_construction_not_classified', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_completed_construction_not_classified', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_property_under_capital_leases', utility_type='other1', plant_status=pd.NA, plant_function=pd.NA)
    ],
    [],
    [
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_other_utility_plant_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='abandonment_of_leases', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_plant_acquisition_adjustment', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='depreciation_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA)
    ],
    [
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='abandonment_of_leases', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_plant_acquisition_adjustment', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='depreciation_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='amortization_of_other_utility_plant_utility_plant_in_service', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA)
    ],
    [
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_in_service_classified_and_unclassified', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_held_for_future_use', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='construction_work_in_progress', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='construction_work_in_progress', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_acquisition_adjustment', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_acquisition_adjustment', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_held_for_future_use', utility_type='common', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_acquisition_adjustment', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='utility_plant_held_for_future_use', utility_type='gas', plant_status=pd.NA, plant_function=pd.NA),
        NodeId(table_name='utility_plant_summary_ferc1', xbrl_factoid='construction_work_in_progress', utility_type='electric', plant_status=pd.NA, plant_function=pd.NA)
    ],
    [],
]
```

Interestingly, the result remains consistent so long as I **don't restart the notebook kernel** -- I can re-run all the cells and get the same result over and over again, as long as I don't restart the kernel.

cmgosnell:
I think the former way to do this it removed one stepparent at a time. So every time during the `for node in self.stepparents(forest):` loop the forest, it was checking against a new forest. which is good. because as we've discussed there are multiple valid paths and we want *one* of them. if you remove all of these stepparents then your removing all paths.

I think a better way to get a deterministic outcome would be to sort the stepparents going into the loop so it happens in the same order every time. I'm gonna check doing something like that.

zaneselvans:
Sorting the list of nodes to iterate over and allowing the list to mutate while we iterate may make the process deterministic, but will it necessarily be correct? What's our criteria for choosing which path to keep?

This is why originally I only removed the stepparents that also had `utility_type=="total"` as a way to choose which path to keep and which to remove.

cmgosnell:
well turns out it didn't make the process deterministic! sooo there is that.

but in terms of what is "right", i don't think it matters which path we choose if there are two paths to the same set of outcomes. your right that choosing the totals as the ones to remove will help narrow the choice down, but it looks like there are still a lot of leftover stepparents with that method.

cmgosnell:
we employed sorted lists and it locked in the orderrrrr 🎉

src/pudl/output/ferc1.py
```diff
@@ -1944,12 +1944,25 @@ def forest(self: Self) -> nx.DiGraph:
...
+        # than one weight (-1 vs. 1) and in those cases, we always want to keep the
+        # weight of -1, since it affects the overall root->leaf calculation outcome.
+        # Maybe this should actually happen in set_forest_attributes?
+        multi_valued_weights = (
+            self.exploded_calcs.groupby(self.calc_cols, dropna=False)["weight"]
+            .transform("nunique")
+            .gt(1)
+        )
+        calcs_to_drop = multi_valued_weights & (self.exploded_calcs.weight == 1)
+        deduplicated_calcs = self.exploded_calcs.drop(calcs_to_drop.index)
```
cmgosnell, line 1958: this works fine and i think is a good solution for now, but i think the more holistic way to do this would be to get all of the nodes direct predecessor nodes in order to generate both the parent_cols and calc_cols in order to select the individual records to select on within `exploded_calcs` during the merge in `set_forest_attributes`

zaneselvans: We should talk about this. I'm not sure I understand what you're saying. I agree this probably belongs over in `set_forest_attributes()` or maybe it its own `deduplicate_calcs()` method that's called by `set_forest_attributes()`

zaneselvans:
I think we need to check whether the nodes we're dropping when enforcing the tree structure:
- are NOT tagged
- do NOT appear in any calculations with a weight of -1

If either of those statements are untrue, then by removing them we will be altering the outcome of the final root -> leaf calculations and what tags end up being applied to the leaf nodes.

src/pudl/output/ferc1.py
```diff
@@ -1912,12 +1919,71 @@ def forest(self: Self) -> nx.DiGraph:
...
+        # than one weight (-1 vs. 1) and in those cases, we always want to keep the
+        # weight of -1, since it affects the overall root->leaf calculation outcome.
+        # Maybe this should actually happen in set_forest_attributes?
+        multi_valued_weights = (
+            self.exploded_calcs.groupby(self.calc_cols, dropna=False)["weight"]
+            .transform("nunique")
+            .gt(1)
+        )
+        calcs_to_drop = multi_valued_weights & (self.exploded_calcs.weight == 1)
+        deduplicated_calcs = self.exploded_calcs.loc[~calcs_to_drop]
```
cmgosnell, line 1979: how many true `multi_valued_weights` do we expect? it might be good to add a little assertion with some expected value to make sure we don't end up unexpectedly dropping a ton.

zaneselvans: Good idea. It should be very small proportion compared to the number of overall calculations.

src/pudl/output/ferc1.py
```diff
@@ -1752,36 +1753,57 @@ def set_forest_attributes(
...
+        for table in self.table_names:
+            if table in MAX_MULTIVALUE_WEIGHT_FRAC:
+                max_mv_wt_frac = MAX_MULTIVALUE_WEIGHT_FRAC[table]
+
+        logger.info(
+            f"Found {sum(multi_valued_weights)}/{len(self.exploded_calcs)} "
+            f"({mv_wt_frac:.2%}; max allowed: {max_mv_wt_frac:.2%}) "
+            "nodes having conflicting weights. Dropping "
+            f"{sum(calcs_to_drop)/len(self.exploded_calcs):.2%} where weight == 1"
         )
```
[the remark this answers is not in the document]
zaneselvans: I think it needs to be down here, and may or may not be useful earlier in the process, because in the end what we care about is whether there are *conflicting* weights, and only in the nodes that remain in the tree after we've done all the pruning, which is just a subset of the records in the original `exploded_calcs`.  I left a quieter, less extensive logging message up in `single_valued_weights()` just for informational purposes.

My intuition is this shouldn't be validated using a pydantic validator because it's quite a complicated derived value, rather than one of the inputs, which we need to check for  validity.

src/pudl/output/ferc1.py
```diff
@@ -906,10 +912,44 @@ class NodeId(NamedTuple):
     plant_function: str | pandas_NAType
 
 
+@asset
+def _out_tags_ferc1(table_dimensions_ferc1) -> pd.DataFrame:
```
zaneselvans, line 916: I'm still getting oriented to them but I think we might call this `_out_ferc1__explosion_tags()` under the new naming conventions? Does that look right to you @bendnorman?

bendnorman: `_out_ferc1__explosion_tags` looks good to me! Just to double check, this is not intended to be written to the database correct? Also, I included some criteria for when to create an intermediate asset as oppose to just keeping it as a function that can be called in a user facing asset:

>Reasons to create intermediate assets:
> - Long-running processing steps. It is convenient to separate the long and short-running processing portions of a dataset into separate assets so debugging the short-running process doesn’t take forever.
> - Steps that frequently need to be inspected when developing. For example, the _core_eia860 assets don’t take a long time to run, but it’s helpful to have them as separate assets so we can inspect the data before it is fed into the harvesting process.

bendnorman: Oh also, do you think this asset deserves as `asset_type` portion? For example, do you think it's important to include the report frequency or a word that describes how the data is modeled?

zaneselvans: Yeah, this isn't meant to get written to the DB. It's constructed based on a manually compiled CSV of metadata that categorizes various values reported in the FERC 1, e.g. whether they should contribute to the utility's rate base or not. This is then programmatically filled in (so we don't have to enumerate every single combination of all of the dimensions). It's fast to build, but because it's manually compiled and then programmatically filled... probably a bit flaky and might need inspection as we go -- though maybe with better validation criteria that could be reduced.

There's nothing like report frequency, but the metadata is manually compiled on the basis of `table_name`, `xbrl_factoid`, `utility_type`, `plant_status`, and `plant_function` (the XBRL dimensions we're working with so far). Not sure what if any of that information would be useful to try and convey in the asset name.

zaneselvans:
@e-belfer I fleshed out the Tasklist based on the comments @cmgosnell left before she went on vacation and notes from our handoff discussion.  The tasks that sound significant include the following (which hopefully aren't just gobbledygook...)

## Re-evaluate conflicting weights check and passthrough calculations

Previously we were excising passthrough calculations, that had a single parent and a single child (plus a correction node) to redirect calculations that involved duplicate values, but doing this we had to be careful never to remove a node with a weight that would alter the full root-to-leaf calculation weight (i.e. any weight other than 1.0).

Since adding all of the other dimensions (`utility_type`, `plant_status` and `plant_function`) it seems that there are no longer any passthrough calculations... so all of that complication may be unnecessary, and if that's true it would be nice to remove it.

However, another kind of issue with "conflicting" weights has come up, in which there are multiple paths to a few nodes in the forest, because sometimes calculations are decomposed in two different ways in parallel -- one breaking things down by xbrl fact, and another breaking things down by the additional dimensions.  We have to choose one or the other in order to avoid having duplicate value getting aggregated by the tree, but at some level it doesn't matter which one we choose (unless there's another metadata dimension like `is_in_rate_base` that applies to one decomposition and not the other which we need to capture).

Right now we are using a heuristic and some random chance to decide which calculations to drop.  Any time that **all** of a node's children have multiple parents, we know that we can safely remove that node without dropping any of the downstream (leafward) values.  This gets rid of almost all of the duplicate paths, but there are a couple of cases where some of the children don't **quite** cover one of the decompositions completely, so we have to purposefully select the decomposition that **does** include all of the children and drop the node that doesn't.

I think that Christina is suggesting that as long as removing a node doesn't result in the loss of any leaves, and we do end up with forest structure in the end, then the calculations should be valid.

## Debugging tag conflicts & tag propagation

Some similar issues come up with the tagging -- which nodes get manually tagged is a bit random, and based on what facts are most salient to the tagger. We need to propagate those tags all the way down to the leaf nodes, so that when we do the root-to-leaf calculation, we have all of the tagging information available.  But this creates the possibility of conflicts between separate tags.  If a node and one of its descendants have different tags, then whichever one is propagated last is going to end up having precedence, which may or may not be correct.  I think these conflicts are probably user input errors and should be flagged and resolved manually.

We also need to check and make sure that none of the nodes we are pruning contain tags that are then lost from the tree.

It also looks like there's some kind of issue with the way the tags are being propagated to the leaves, since there are at leaset a few cases in which the tags of parent nodes aren't showing up in their children.  But it could also be that I'm generating the `forest_as_table` before I've propagated the tags (need to double-check).

src/pudl/output/ferc1.py
```diff
@@ -2035,7 +2130,7 @@ def leafy_meta(self: Self) -> pd.DataFrame:
         - The weight associated with the leaf, in relation to its root.
         """
         # Construct a dataframe that links the leaf node IDs to their root nodes:
-        pruned_forest = self.forest
+        pruned_forest = self.annotated_forest
```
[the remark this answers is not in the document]
zaneselvans: Done.

zaneselvans:
After adding all of the dimensions and merging in the intra-dimensional fixes:

### No more passthroughs
* There are no longer any nodes with a a single parent and a single child + a `_correction`.
* Should there be something analogous but with a different definition because the totals contains subdimensions now so having a "single" child previously means having only within-fact subcomponents?
* Or should this concept be retired?

### We have actual forests
* All 3 of the calculation forests are free of multi-parent nodes, as required... though this is kind of by construction / brute force on our part.

### Some calculations aren't matching exactly
* `balance_sheet_liabilities_ferc1` and `income_statement_ferc1 have single-digit percentage calculation errors.
* More than half of the `balance_sheet_assets_ferc1` don't exactly match reported values.
* Programmatically things seem like they're probably pretty good, but something needs to be fixed in the balance sheet assets table(s).
* The `balance_sheet_assets` table is also the only one where we have to manually identify "almost pure stepparents" for removal so maybe we're messing something up there.

### Pruned and conflicting tags
* The `balance_sheet_assets_ferc1` table has 2 tags that conflict (`is_in_rate_base` is `partial` vs. `yes`). This seems like it's probably just a manual tagging error that we can resolve.
* `balance_sheet_assets_ferc1` also has a handful of nodes which were explicitly tagged, but that don't appear in the final calculation forest, either because they were orphans from the start or they got pruned. It might be that the tagged nodes which are getting removed are from inferred calculations, and it's not important that they're being removed.
* It might also be that we are choosing to remove nodes that have been manually identified as important in the process of enforcing a hierarchical calculation tree, in which case we might need to choose other nodes to remove.

### Conflicting weights & multiple possible calculation paths
* The `income_statement_ferc1` table has a handful of nodes with weights of -1.0 that get removed due to conflict.
* This may not actually be an issue. We're not losing any leaf nodes in the process, and those leaves still have a valid calculation path to the root nodes that's being retained.
* Right now we are changing the structure of the tree and the weights in independent processes.
* The cases where a node shows up more than once with different weights are **probably** also cases in which they show up in more than one calculation with different parents.
* How do we ensure that the calculation path we choose (mostly via the choice of which stepparents to prune) is consistent with which calculation weight we keep when there's more than one option?

# Debugging via calculation consistency
* There are lots of different reasons that calculation might not match up.
* Sometimes the formulae that come from FERC are wrong!
* Sometimes the data entered by utilities are internally inconsistent (e.g. due to a revision)
* Sometimes utilities use the wrong version of the form, so row numbers + factoids don't align correctly.
* We may have incorrectly inferred/corrected/constructed the calculation formula.
* We may have introduced inconsistency between the calculation path and the retained weights.
* How can we differentiate between these cases and use calculation mismatches to identify problems and fix them?

src/pudl/output/ferc1.py
```diff
@@ -1912,12 +1941,78 @@ def forest(self: Self) -> nx.DiGraph:
...
+            ),
+        ]
+        forest.remove_nodes_from(almost_pure_stepparents)
+
+        # Ensure that we haven't removed any calculation components that *would* have
+        # altered the final root-to-leaf calculations.
+        # * Weights pertain to child nodes.
+        # * But the nodes we're removing above are guaranteed to be parents (Though
+        #   many them will also be children)
+        # * We want to check that any of them that *are* children have a weight of 1.0
```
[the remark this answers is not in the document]
zaneselvans: I realized that a lot of the complexity in all this was because the weights were being associated with nodes, rather than edges -- they really are an attribute of a calculation which links two nodes... and if we put the weight on the edge then there's no problem with having lots of different weights associated with different edges that point at the same node.  And by virtue of enforcing a tree structure, we necessarily end up with a single remaining incoming edge for each node, with a single weight, removing this potential conflict / ambiguity.

zaneselvans:
@e-belfer here's what I've been doing to generate the outputs.

## Imports & pull some tables
```py
import pandas as pd

import pudl

from pudl.etl import defs
from pudl.output.ferc1 import (
    Exploder,
    MetadataExploder,
    NodeId,
    XbrlCalculationForestFerc1,
    EXPLOSION_CALCULATION_TOLERANCES,
)

clean_xbrl_metadata_json = defs.load_asset_value(AssetKey("clean_xbrl_metadata_json"))
metadata_xbrl_ferc1 = defs.load_asset_value(AssetKey("metadata_xbrl_ferc1"))
calculation_components_xbrl_ferc1 = defs.load_asset_value(
    AssetKey("calculation_components_xbrl_ferc1")
)
tags_df = defs.load_asset_value(AssetKey("_out_ferc1__explosion_tags"))
```

## Define the input arguments:
```py
explosion_args = {
    "income_statement_ferc1": {
        "root_table": "income_statement_ferc1",
        "table_names_to_explode": [
            "income_statement_ferc1",
            "depreciation_amortization_summary_ferc1",
            "electric_operating_expenses_ferc1",
            "electric_operating_revenues_ferc1",
        ],
        "calculation_tolerance": EXPLOSION_CALCULATION_TOLERANCES["income_statement_ferc1"],
        "seeds": [
            NodeId(
                table_name="income_statement_ferc1",
                xbrl_factoid="net_income_loss",
                utility_type="total",
                plant_status=pd.NA,
                plant_function=pd.NA,
            ),
        ],
        "tags": tags_df,
    },
        "root_table": "balance_sheet_assets_ferc1",
        "table_names_to_explode": [
            "balance_sheet_assets_ferc1",
            "plant_in_service_ferc1",
        ],
        "calculation_tolerance": EXPLOSION_CALCULATION_TOLERANCES["balance_sheet_assets_ferc1"],
        "seeds": [
            NodeId(
                xbrl_factoid="assets_and_other_debits",
                utility_type=pd.NA,
                plant_status=pd.NA,
                plant_function=pd.NA,
            )
        ],
        "tags": tags_df,
    },
    "balance_sheet_liabilities_ferc1": {
        "root_table": "balance_sheet_liabilities_ferc1",
        "table_names_to_explode": [
            "balance_sheet_liabilities_ferc1",
            "retained_earnings_ferc1",
        ],
        "calculation_tolerance": EXPLOSION_CALCULATION_TOLERANCES["balance_sheet_liabilities_ferc1"],
        "seeds": [
            NodeId(
                table_name="balance_sheet_liabilities_ferc1",
                xbrl_factoid="liabilities_and_other_credits",
                utility_type=pd.NA,
                plant_status=pd.NA,
                plant_function=pd.NA,
            )
        ],
        "tags": tags_df,
    },
}
```

## Function to create exploded assets
```py
def exploded_table(
    root_table: str,
    table_names_to_explode: list[str],
    calculation_tolerance: float,
    seeds: list[NodeId],
    tags: pd.DataFrame,
):
    metadata_xbrl_ferc1 = defs.load_asset_value(AssetKey("metadata_xbrl_ferc1"))
    calculation_components_xbrl_ferc1 = defs.load_asset_value(
        AssetKey("calculation_components_xbrl_ferc1")
    )

    dfs_to_explode = {
        table: pd.read_sql(table, pudl_engine) for table in table_names_to_explode
    }

    exploder = Exploder(
        root_table=root_table,
        table_names=table_names_to_explode,
        metadata_xbrl_ferc1=metadata_xbrl_ferc1,
        calculation_components_xbrl_ferc1=calculation_components_xbrl_ferc1,
        seed_nodes=seeds,
        calculation_tolerance=calculation_tolerance,
        tags=tags,
    )
    return {
        "exploder": exploder,
        "exploded_data": exploder.boom(tables_to_explode=dfs_to_explode),
    }

test_explode = {}
for root_table in explosion_args:
    print(f"Exploding: {root_table}")
    test_explode[root_table] = exploded_table(**explosion_args[root_table])
```

## Inspect one of the explosions
```py
root_table = "balance_sheet_assets_ferc1"
exploder = test_explode[root_table]["exploder"]
exploded_data = test_explode[root_table]["exploded_data"]
calc_forest = exploder.calculation_forest
exploded_meta = calc_forest.exploded_meta
forest_as_table = calc_forest.forest_as_table
leafy_meta = calc_forest.leafy_meta
```

src/pudl/output/ferc1.py
```diff
@@ -1731,68 +1842,135 @@ def exploded_calcs_to_digraph(
...
+            .to_dict(orient="index")
         )
+        return node_attrs
 
-        attr_cols = ["weight"]
-        deduped_calcs = exploded_calcs[~calcs_to_drop][
-            self.calc_cols + attr_cols
-        ].drop_duplicates()
+    @cached_property
+    def edge_attrs(self: Self) -> dict[Any, Any]:
```
jdangerx, line 1879: Love this split between edges and nodes - way better than the old world!

src/pudl/output/ferc1.py
```diff
@@ -1731,68 +1842,135 @@ def exploded_calcs_to_digraph(
...
+        that ancestor nodes (closer to the roots) might have tags associated with them
+        that are in conflict with descendant nodes (closer to the leaves). If that's
+        the case then when we propagate the tags to the leaves, whichever tag is
+        propagated last will end up taking precedence.
+
+        These kinds of conflicts are probably due to errors in the tagging metadata, and
+        should be investigated.
+        """
+        nodes = annotated_forest.nodes
+        for ancestor in nodes:
```
jdangerx, line 1957: This is some deep nesting, ha. I don't really see an obvious way to deal with it beyond moving the nesting elsewhere... So this seems fine. I guess you could define a few little helper functions inline here.

src/pudl/output/ferc1.py
```diff
@@ -1811,28 +1989,50 @@ def full_digraph(self: Self) -> nx.DiGraph:
...
             .loc[seeded_parents]
             .reset_index()
         )
-        seeded_digraph: nx.DiGraph = self.exploded_calcs_to_digraph(
-            exploded_calcs=seeded_calcs
+        # All child nodes in the seeded calculations that are part of the input graph.
+        seeded_child_nodes = list(
+            set(
+                seeded_calcs[self.calc_cols].itertuples(index=False, name="NodeId")
+            ).intersection(graph.nodes)
```
jdangerx, line 2026: Just making sure - we want `graph` here, not `seeded_nodes`?

src/pudl/output/ferc1.py
```diff
@@ -1862,87 +2062,84 @@ def forest(self: Self) -> nx.DiGraph:
...
         if pruned_nodes := set(nodes_before_pruning).difference(nodes_after_pruning):
             raise AssertionError(f"Unexpectedly pruned stepchildren: {pruned_nodes=}")
 
-        forest = self.set_forest_attributes(
-            forest,
-            exploded_meta=self.exploded_meta,
-            exploded_calcs=self.exploded_calcs,
-            tags=self.tags,
-        )
+        # HACK alter.
```
jdangerx, line 2087: What does this mean?

src/pudl/output/ferc1.py
```diff
@@ -1862,87 +2062,84 @@ def forest(self: Self) -> nx.DiGraph:
...
             stepchildren = set(self.stepchildren(forest)).intersection(children)
-            if (
-                (children == stepchildren)
-                & (len(children) > 0)
-                & (node.utility_type == "total")
-            ):
+            if (children == stepchildren) & (len(children) > 0):
+                pure_stepparents.append(node)
                 forest.remove_node(node)
+        logger.info(f"Removed {len(pure_stepparents)} redundant/stepparent nodes.")
```
jdangerx, line 2076: Probably good to define stepparents/stepchildren somewhere 😅

src/pudl/output/ferc1.py
```diff
@@ -1862,87 +2062,84 @@ def forest(self: Self) -> nx.DiGraph:
...
-            forest,
-            exploded_meta=self.exploded_meta,
-            exploded_calcs=self.exploded_calcs,
-            tags=self.tags,
-        )
+        # HACK alter.
+        # two different parents. those parents have different sets of dimensions.
+        # sharing some but not all of their children so they weren't caught from in the
+        # only stepchildren node removal from above. a generalization here would be good
+        almost_pure_stepparents = [
```
jdangerx, line 2091: This is a manual spot fix sort of situation?

jdangerx (changes requested):
A couple abstraction-level questions that we should probably discuss, though largely I think we can merge this. It would probably also be good to discuss my thoughts on [how to clean up the graph more upstream](https://gist.github.com/jdangerx/e29729cac5ef34f58ca25b8c3e77b857) and whether they're totally off base or not.

src/pudl/output/ferc1.py
```diff
@@ -906,12 +947,46 @@ class NodeId(NamedTuple):
     plant_function: str | pandas_NAType
 
 
+@asset
+def _out_ferc1__explosion_tags(table_dimensions_ferc1) -> pd.DataFrame:
+    """Grab the stored table of tags and add infered dimension."""
+    # NOTE: there are a bunch of duplicate records in xbrl_factoid_rate_base_tags.csv
+    # Also, these tags are only applicable to the balance_sheet_assets_ferc1 table, but
+    # we need to pass in a dataframe with the right structure to all of the exploders,
```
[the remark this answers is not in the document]
jdangerx: We talked about this, `income_statement` expects `tags` but we just don't have the tags yet. So they are all the same thing, so we should keep using the same interface for them.

jdangerx (approved):

jdangerx:
Most of my comments were non-blocking and we addressed the blocking one on a synchronous call - see [note above. ](https://github.com/catalyst-cooperative/pudl/pull/2805#discussion_r1330786635)

There are things we might want to change, but those can happen as a different PR onto `explode_ferc1` - no need to scope creep this PR.

.pre-commit-config.yaml
```diff
@@ -27,7 +26,7 @@ repos:
   # Formatters: hooks that re-write Python & documentation files
   ####################################################################################
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.0.288
+    rev: v0.0.290
```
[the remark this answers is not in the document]
zaneselvans: It's been great so far! pre-commit only updates once a week (I accidentally did it locally) and I don't have an upper bound on the version in `pyproject.toml` right now.

src/pudl/output/ferc1.py
```diff
@@ -10,13 +12,52 @@
...
 from pandas._libs.missing import NAType as pandas_NAType
-from pydantic import BaseModel, validator
+from pydantic import BaseModel, confloat, validator
 
 import pudl
 
 logger = pudl.logging_helpers.get_logger(__name__)
 
 
+class CalculationToleranceFerc1(BaseModel):
```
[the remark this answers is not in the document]
zaneselvans: One of the things I was tracking in here originally went away, but I still think we are going to end up with a variety of metrics we want to check in the intra- and inter-table calculations, and the root-to-leaf calculations and having it be codified in a structured way seemed like it would be helpful.

src/pudl/output/ferc1.py
```diff
@@ -1160,6 +1226,60 @@ def metadata(self):
...
+        )
+
+        def get_dbf_row_metadata(pudl_table: str, year: int = 2020):
+            dbf_tables = pudl.transform.ferc1.FERC1_TFR_CLASSES[
+                pudl_table
+            ]().params.aligned_dbf_table_names
+            dbf_metadata = (
+                pudl.transform.ferc1.read_dbf_to_xbrl_map(dbf_table_names=dbf_tables)
+                .pipe(pudl.transform.ferc1.fill_dbf_to_xbrl_map)
+                .query("report_year==@year")
```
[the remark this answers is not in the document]
zaneselvans: It was just to keep it in the pipeline.  I wish there were a `df.select()` method.

src/pudl/output/ferc1.py
zaneselvans: @cmgosnell made this note and I've also been confused by it.

src/pudl/output/ferc1.py
zaneselvans: Yes, this is just a weird spot fix and will hopefully be obviated by the only-infer-leafy-totals that we talked about.

src/pudl/output/ferc1.py
zaneselvans: There are `calc_forest.stepparents` and `calc_forest.stepchildren` methods / properties that define them but maybe I should reference those methods directly here.

src/pudl/output/ferc1.py
zaneselvans: Yeah I don't know what I was thinking before.

src/pudl/output/ferc1.py
zaneselvans: It's not great... but it checks the thing and it's isolated off here by itself.

src/pudl/output/ferc1.py
zaneselvans: Yes, taking the intersection with `graph.nodes` means its ignoring any other nodes that might have existed in the full list of calculations in `seeded_calcs` even if they didn't get included in `graph`.

src/pudl/output/ferc1.py
```diff
@@ -1133,21 +1202,18 @@ def calculations(self: Self):
         return calc_explode
 
     @property
-    def metadata(self):
+    def metadata(self: Self) -> pd.DataFrame:
```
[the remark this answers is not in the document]
zaneselvans: @cmgosnell and I talked a couple of weeks ago about how re-organizing the relationships between `Exploder`, `MetadataExploder` and `CalculationForest`, and also about factoring the metadata transformations out of the data transformation classes since it's a whole separate parallel system now, but we put it off and I don't remember right now how we were thinking about organizing it.  It does seem like we could just turn these into `Exploder.exploded_meta` and `Exploder.exploded_calcs` properties.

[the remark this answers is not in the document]
zaneselvans: I went ahead and consolidated these classes.

PR #2815, opened 2023-08-30
Update description for small plants table

aesharpe:
I'm doing some spring (because it's spring in Argentina!) cleaning of my crufty local branches, and I found one that updated the description for the small plants table. It's a little change, but it's useful, and I figured I'd go ahead and get it into dev.


src/pudl/metadata/resources/ferc1.py
```diff
@@ -537,14 +537,21 @@
...
-            "less than 10 MW installed nameplate capacity. As reported on FERC Form 1 "
-            "Schedule 410 (pages 410-411), and extracted from the FERC Visual FoxPro "
-            "database table f1_gnrt_plant."
-        ),
+        "description": """The generating plant statistics for steam plants with less
+than 25 MW installed nameplate capacity as well as internal combustion
+plants, gas turbine-plants, conventional hydro plants, and pumped storage plants with
+less than 10 MW installed nameplate capacity. As reported on FERC Form 1 Schedule 410
+(pages 410-411) and extracted from the FERC Visual FoxPro and XBRL. See our
+:py:const:`pudl.extract.ferc1.TABLE_NAME_MAP_FERC1` for linkages to the raw tables.
```
zaneselvans, line 545: Note that RST in these strings will get rendered in our documentation (Read The Docs), but will show up as-is in Datasette.

src/pudl/metadata/resources/ferc1.py
```diff
@@ -537,14 +537,21 @@
...
+
+The raw version of this table is more like a digitized PDF than an actual data table.
+The rows contain lots of information in addition to what the columns might suggest.
+For instance, a single column may contain header rows, note rows, and total rows. This
+extraneous information is useful, but it prevent proper analysis when mixed in with the
```
zaneselvans, line 550: prevent -> prevents

src/pudl/metadata/resources/ferc1.py
```diff
@@ -537,14 +537,21 @@
...
+rest of the values data in the column. We employ a couple of data transformations to
+extract these rows from the data and preserve some of the information they contain
+(fuel type, plant type, FERC license, or general notes about the plant) in seperate
```
zaneselvans, line 553: seperate -> separate

zaneselvans (approved):
A couple of typos but looks fine.

src/pudl/metadata/resources/ferc1.py
aesharpe: mmmm annoying, I will remove the :py:const: part then

aesharpe:
@zaneselvans The beginning of this says ``The generating plant statistics for steam plants...`` could we just say``the generation statistics for steam plants...`` or is "generating plant statistics" a specific thing?

zaneselvans:
Ah sorry I missed this Q.

I think we should probably change this description in a different way -- it's not steam plants, I think that's probably a cut-and-paste error that came over from the large steam plants description.

Across the FERC data (and in discussion of corporate assets more generally) ["plant" means a fixed capital asset](https://financial-dictionary.thefreedictionary.com/plant), and there are a bunch of different subcategories... In FERC there's "transmission plant" and "distribution plant" and "production plant" which includes "generation plant" which includes "steam plant" and "hydraulic plant" and "nuclear plant" etc. "plant in service" means "all the big fixed capital assets that are actively engaged in our business operations"

I think a good phrase here would probably be "Generating plant statistics for small plants" or "Small generating plant statistics" -- In Form 1 itself it's entitled "Generating Plant Statistics (Small Plants)" which is analogous to the titles for all of the other pages describing electric plant like "Steam Electric Generating Plant Statistics" and "Hydroelectric Generating Plant Statistics"

aesharpe:
> In FERC there's "transmission plant" and "distribution plant" and "production plant" which includes "generation plant"

Ahh ok that makes sense / answers my question as to whether "generating plant" was a technical term or not (which it is!)

aesharpe:
> I think we should probably change this description in a different way -- it's not steam plants, I think that's probably a cut-and-paste error that came over from the large steam plants description.

This is because I only copied part of the sentence. The full description is as follows: 

```
The generating plant statistics for steam plants with less
than 25 MW installed nameplate capacity as well as internal combustion
plants, gas turbine-plants, conventional hydro plants, and pumped storage plants with
less than 10 MW installed nameplate capacity.
```

zaneselvans:
Ahh, okay. I think we might as well include the whole description.

aesharpe:
> Ahh, okay. I think we might as well include the whole description.

meaning leave it as is?

Issue #2812, opened 2023-08-28
Apply naming convention to core assets


bendnorman:
Apply naming convention to non-intermediate core tables. Reference the [Proposed Asset Name](https://docs.google.com/spreadsheets/d/1RBuKl_xKzRSLgRM7GIZbc5zUYieWFE20cXumWuv5njo/edit#gid=1126117325) column in the Naming Conventions sheet.

#2788 and this issue should be merged into dev at the same time so user-facing tables all use the new naming convention. See #2765 for release plan.

```[tasklist]
- [x] hourly_emissions_epacems
- [x] plants_entity_eia
- [x] generators_entity_eia
- [x] boilers_entity_eia
- [x] utilities_entity_eia
- [x] coalmine_eia923
- [x] boiler_emissions_control_equipment_assn_eia860
- [x] boiler_stack_flue_assn_eia860
- [x] boiler_cooling_assn_eia860
- [x] boiler_generator_assn_eia860
- [x] generation_eia923
- [x] plants_eia860
- [x] emissions_control_equipment_eia860
- [x] utilities_eia860
- [x] generators_eia860
- [x] fuel_receipts_costs_eia923
- [x] generation_fuel_nuclear_eia923
- [x] boilers_eia860
- [x] boiler_fuel_eia923
- [x] generation_fuel_eia923
- [x] ownership_eia860
- [x] purchased_power_ferc1
- [x] depreciation_amortization_summary_ferc1
- [x] plants_small_ferc1
- [x] fuel_ferc1
- [x] retained_earnings_ferc1
- [x] transmission_statistics_ferc1
- [x] balance_sheet_liabilities_ferc1
- [x] clean_xbrl_metadata_json
- [x] plants_pumped_storage_ferc1
- [x] electric_plant_depreciation_functional_ferc1
- [x] electric_operating_expenses_ferc1
- [x] plants_steam_ferc1
- [x] plants_hydro_ferc1
- [x] utility_plant_summary_ferc1
- [x] cash_flow_ferc1
- [x] electric_plant_depreciation_changes_ferc1
- [x] electric_energy_dispositions_ferc1
- [x] electricity_sales_by_rate_schedule_ferc1
- [x] income_statement_ferc1
- [x] other_regulatory_liabilities_ferc1
- [x] balance_sheet_assets_ferc1
- [x] electric_energy_sources_ferc1
- [x] electric_operating_revenues_ferc1
- [x] plant_in_service_ferc1
[...]

PR #2818, opened 2023-08-30
Feature branch: Rename core + output assets to match new naming protocols

e-belfer:
See #2812 for issue description and progress tracking. This will collect naming PRs, which will all need to be merged in together to avoid major headaches for downstream users.

The PR includes renamed assets for:
- core EIA 923, EIA 860 and harvested EIA tables
- core FERC Form 1 tables
- core EIA 861 tables
- core FERC Form 714 tables
- coding tables
- glue/association tables
- EPA CEMS parquet files and core assets

It also does a few other things:
- fixes mixed usage of annual and yearly in core assets
- reorganizes modules to match _core and core structure.

Not converted:
- `entity_types_eia` , which is currently not loaded into the PUDL DB.
- assets from the XBRL explosion (currently just `clean_xbrl_metadata_json`)

# Steps for merging

1. Rename all of the tables and merge docs PR into this branch
2. Run a full build with validation for this branch
3. Once the build passes and this branch is ready to be merged into `dev`, create a PR for `dev -> main`.
4. Create a tag and push a tag called `pre-name-change`. **Question: Should we do a final full software release or just let people install from git tab?**
5. Squash merge this branch into `dev`
6. Once the schedule dev nightly build passes, merge `dev -> main` and tag a release using our desired release naming convention: `vYYYY.MM.DD`
7. Help known users migrate!

# PR Checklist


bendnorman:
Thanks @e-belfer! I feel pretty good about all of the proposed asset names in the [naming convention sheet](https://docs.google.com/spreadsheets/d/1RBuKl_xKzRSLgRM7GIZbc5zUYieWFE20cXumWuv5njo/edit#gid=1126117325) except for the assets currently in the `glue` group:

- We decided to use `assn` for the asset type for all glue/association tables. 
- I'm not sure what the `datasource` portion of the asset name should be. EIA EPA crosswalk assets come from EPA. Most of the other current glue assets are hand-crafted excel sheets we create so maybe `pudl` is a reasonable datasource for these assets? Would it be awkward to have a module named `pudl.core.pudl` to contain these assets?
- `plants_pudl` and `utilities_pudl` feel more like entity tables assets because they only contain pudl ids and names. If we follow the naming convention the new names would be `core_pudl__entity_plants_pudl` and `core_pudl__entity_utilities_pudl`. These names are pretty similar to `core_eia__entity_utilities` and `core_eia__entity_plants`. Do you think this will confuse users? Is there another way we can name these?

e-belfer:
>     * I'm not sure what the `datasource` portion of the asset name should be. EIA EPA crosswalk assets come from EPA. Most of the other current glue assets are hand-crafted excel sheets we create so maybe `pudl` is a reasonable datasource for these assets? Would it be awkward to have a module named `pudl.core.pudl` to contain these assets?

I think the alternative to `pudl.core.pudl` would be using a name like "manual" or "hand" but it wouldn't well characterize all the machine learning outputs that will likely come to live here. Alternately we could use "cat" to point to Catalyst as the origin of this data, in the event that we start integrating glue tables from other non-agency sources.

>     * `plants_pudl` and `utilities_pudl` feel more like entity tables assets because they only contain pudl ids and names. If we follow the naming convention the new names would be `core_pudl__entity_plants_pudl` and `core_pudl__entity_utilities_pudl`. These names are pretty similar to `core_eia__entity_utilities` and `core_eia__entity_plants`. Do you think this will confuse users? Is there another way we can name these?

Is there a reason to keep the `_pudl` other than legacy compatibility? I think they'd be clear mirrors of the EIA tables but containing PUDL-only info in this case.

e-belfer:
Updated decisions: keep "pudl" as datasource, and leave duplicated PUDL as is.

bendnorman (approved):
Looks good to me! Just a couple of assets in notebooks that might need to be renamed.

I'm thinking we should use this as the main feature branch for the core/output rename. What do you think @e-belfer?

notebooks/work-in-progress/better-heatrates.ipynb
```diff
@@ -648,7 +648,7 @@
     "* `generation_fuel_eia923`\n",
     "* `boiler_generator_assn_eia860`\n",
     "* `generators_eia860`\n",
-    "* `generators_entity_eia`\n",
+    "* `core_eia__entity_generators`\n",
     "* `boiler_fuel_eia923`"
```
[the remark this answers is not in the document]
e-belfer: Generally I was trying to avoiding dealing with renames for work in progress notebooks, with the idea that they'll need other revamps to stay up to date anyways. This is probably just a case of getting carried away with Ctrl+F and I'll revert it to make things consistent.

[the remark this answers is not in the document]
bendnorman: Ah I see. That seems reasonable.

zaneselvans:
The unit test failure isn't real. It's a rare stochastic failure in the timeseries interpolation.

zaneselvans:
Maybe this already got fixed in some of the stuff that hasn't been able to merge in yet, but looking at the [PUDL Data Dictionary](https://catalystcoop-pudl--2818.org.readthedocs.build/en/2818/data_dictionaries/pudl_db.html#pudl-data-dictionary) in the last docs build it looks like `plant_parts_eia` and `retained_earnings_appropriations_ferc1` still have entries, despite having been superseded by new table names (they're right at the top)

bendnorman:
Hmm I'll take a look and update this branch with the changes from dev again.

bendnorman:
It looks like `retained_earnings_appropriations_ferc1` is a "disabled" FERC table. Should we rename it or just remove it from the metadata?

devtools/data-release.sh
zaneselvans, line 1: This file has been deleted from `dev`. Do we know why it's showing up here now?

src/pudl/convert/epacems_to_parquet.py
zaneselvans, line 1: This file has been deleted on `dev`

src/pudl/etl/analysis_assets.py
zaneselvans, line 1: This file has been deleted on `dev`

src/pudl/metadata/resources/mcoe.py
zaneselvans, line 1: If nothing inside this module is going to be called MCOE maybe we should change the name of the module?

devtools/data-release.sh
bendnorman: Yeah that is strange. I'll look into it.

src/pudl/metadata/resources/mcoe.py
bendnorman: I agree! I decided to punt updating references to MCOE everwhere because it's a large change. #2910

bendnorman:
The CI and full build passed... should we merge this thing in?

zaneselvans:
I think merging it in and then using it ourselves for a while to iron out the kinks before we cut a new release with the new names is probably a good idea.  I'm sure there will be issues and inconsistencies that we only discover when using it day-to-day.

bendnorman:
Should we point datasette at `v2023.12.01` data so datasette continues to have the old names until we do a full release with the new names?

zaneselvans:
You mean just stop updating it?

zaneselvans:
I dunno. The docs say that Datasette reflects the nightly builds, so we would need to change that. And the `latest` tag on Read The Docs points at `dev` so by default the documentation that anyone sees after the merge will reflect the new names.

If we can get on an actual monthly release schedule, then having the public Kaggle and Datasette outputs track the `stable` release rather than nightly doesn't sound like a bad idea.

bendnorman:
Having datasette and docs point at `stable` seems like the ideal setup. I guess we should leave it on `dev` for now.

zaneselvans (approved):
I'm sure we will find things to fix and improve as we use it, but I think that's only going to happen through actual usage.

PR #2824, opened 2023-09-02
Simplify & enhance linting by using ruff.

zaneselvans:
ruff is a Python linter written in Rust. It's literally 1000x as fast as flake8, and re-implements the functionality of a bajillion flake8 plugins and more. Using ruff means our linting setup is both much simpler and much faster.

Changes include:
- Moved all remaining tool configuration from tox.ini to pyproject.toml
- Replaced bandit with ruff
- Replaced pyupgrade with ruff
- Replaced all flake8 plugins with ruff
- Updated tox testenvs to reflect use of ruff
- Update .pre-commit-config.yaml to reflect use of ruff
- Added NumPy and pandas specific checks
- Removed other linter dependencies from dev & test environments
- Added ruff as a dependency in the test & dev environments
- Fixed a few new issues identified by ruff
- Removed the rstcheck linter, since there's no ruff support for it yet and it's
  been causing dependency issues in our Docker builds.
- Removed the --sign option from twine since PyPI no longer supports it.
- Updated to build v1.0.0

I tried to keep the linting rules applied in the initial commit as close to the old ones that we had specified with flake8 and friends, and then added a few more incremental fixes (which uncovered some minor bugs)

I added but comment out some rule codes in `pyproject.toml` for checks I think we might want to enable in the future, but which require more effort to deal with.


src/pudl/metadata/codes.py
```diff
@@ -209,15 +209,16 @@
                     "duquense",
                     "Duquense Light Company & Energy Systems Associates",
                 ),
-                ("DS", "doosan" "Doosan"),
+                ("DS", "doosan", "Doosan"),
```
jdangerx, line 212: Woah, did ruff somehow catch this too?

jdangerx (approved):
Sweet, this makes the linting setup much less piecemeal :) The new numpy/pandas checks are a nice win too!

src/pudl/metadata/codes.py
zaneselvans: Yeah, it's got a "oops did you accidentally concatenate two strings implicitly?" check!

src/pudl/convert/epacems_to_parquet.py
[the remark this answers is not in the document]
zaneselvans: I don't think it's an empty file... I think this might just mean that the diff is empty -- it's still a script in the working tree.

PR #2825, opened 2023-09-03
Run ferc_to_sqlite and pudl_etl independent of integration tests

rousik:
This reworks how tests (unit, integration) are done by github actions. 

In particular, it modifies how ci-integration is executed by running `ferc_to_sqlite` and `pudl_etl` commands independently, making use of dagster parallelism and larger runners.


.github/workflows/tox-pytest.yml
```diff
@@ -160,63 +163,139 @@ jobs:
           workload_identity_provider: "projects/345950277072/locations/global/workloadIdentityPools/gh-actions-pool/providers/gh-actions-provider"
           service_account: "tox-pytest-github-action@catalyst-cooperative-pudl.iam.gserviceaccount.com"
 
-      - name: Run integration tests, trying to use GCS cache if possible
+      - name: Run Fast ETL
         run: |
-          tox -e integration -- --gcs-cache-path=gs://zenodo-cache.catalyst.coop --durations 0
+          pudl_setup
+          ferc_to_sqlite src/pudl/package_data/settings/etl_fast.yml
```
zaneselvans, line 169: I suspect `ferc_to_sqlite` should also use the `--gcs-cache-path` argument so it doesn't attempt to download anything from Zenodo unnecessarily.

zaneselvans: To use the `conda` environment that you have create with `setup-micromamba` I think you may need to use the [`micromamba-shell` command](https://github.com/mamba-org/setup-micromamba#custom-micromamba-shell-wrapper) -- otherwise I think everything is just going to be running in the system python environment.  I suspect this is why you've had to manually install libsnappy-dev.  I suspect that you've `pip install`ed the PUDL package in the system python environment.

rousik: The new setup step claims to hack the default shell rc files to ensure that the new environment is run automatically and that indeed seems to be the case here. Win win.

rousik: It seems that for the full ETL run, having both ferc_to_sqlite and pudl_etl part of a single command would be AFAIK nice simplification. That is not critical, but would be nice.

It does seem that in order to run ETL here, we effectively need to replicate the runtime environment we have in docker quite a bit (installing various dependencies, running alembic upgrade head), which might be another reason why running from docker image built once might be preferrable.

.github/workflows/tox-pytest.yml
```diff
@@ -137,6 +137,7 @@ jobs:
       - name: Install PUDL and dependencies
         run: |
           ls -l
+          sudo apt-get install libsnappy-dev
```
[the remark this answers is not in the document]
rousik: Wrong shell, this should now be good

.github/workflows/tox-pytest.yml
```diff
@@ -3,6 +3,10 @@ name: tox-pytest
 
 on: [pull_request, workflow_dispatch]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
zaneselvans: Ooooh, this is a good idea!

.github/workflows/tox-pytest.yml
```diff
@@ -160,63 +168,140 @@ jobs:
...
+          pudl_etl --gcs-cache-path=gs://zenodo-cache.catalyst.coop \
+            src/pudl/package_data/settings/etl_fast.yml
 
-      - name: Upload coverage
-        uses: actions/upload-artifact@v3
-        with:
-          name: coverage-integration
-          path: coverage.xml
+      - name: Run integration tests on live data
+        run: pytest --live-dbs src/pudl/test/integration
```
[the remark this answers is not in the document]
rousik: Uh. Looks like I have un-fixed it rather than fixing it.

zaneselvans:
I think the call to `pytest` is failing because you need to install some or all of the extra dependencies when you install the `catalystcoop.pudl` package -- right now you're just installing the bare minimum requirements.  so...

```
pip install ./[doc,test,dev,datasette]
```

rather than just 
```
pip install .
```

.github/workflows/tox-pytest.yml
```diff
@@ -97,44 +107,46 @@ jobs:
...
-          conda config --show-sources
-          conda config --show
-          printenv | sort
-
-      - name: Log SQLite3 version
-        run: |
-          which sqlite3
-          sqlite3 --version
+      - name: Install PUDL and dependencies
+        run: pip install "./[test]"
```
zaneselvans, line 149: I added the test "extras" to the install so that `tox` can run.

.github/workflows/tox-pytest.yml
```diff
@@ -97,44 +107,46 @@ jobs:
           name: coverage-unit
           path: coverage.xml
 
+  # TODO(rousik): the following could simply run docker
+  # container mounted to specific directory.
+  # TODO(rousik): run this on larger runners, see dev for
+  # how to do that.
   ci-integration:
+    needs:
+      - ci-unit
```
zaneselvans, line 116: Is this dependency required to ensure that the CI shuts down if anything fails, before kicking off the integration tests?

zaneselvans:
Now the tests are failing because they're looking for the databases in the wrong location -- in the temporary directory that they would have been written to if they'd been built with the fixtures, rather than the directories specified by `$PUDL_INPUT` and `$PUDL_OUTPUT`.  Those envvars are being passed into the Tox environment though, so it seems as if there may be a problem in how the correct output directory is identified by the settings.

zaneselvans:
A separate issue here is that we're breaking the python environment isolation that Tox provides by running `pudl_etl` outside of it and installing from the local directory, rather than from the package that Tox builds, in the virtual environment that Tox creates, which means it might have access to files that appear in the repository that don't appear in the package. Instead we could run the ETL inside Tox like is done with `tox -e nuke`.

zaneselvans:
Hmm. I tried running the following locallly:
```
tox -e integration -- --live-dbs
```
And it didn't have any trouble finding the correct pre-existing DBs in my `$PUDL_OUTPUT` directory.

rousik:
> t

I'm still contemplating what would be the best approach here. I would like to reuse tox, but I would also like to be able to run the three steps (ferc_to_sqlite, pudl_etl, integration tests) independently as that AFAIK allows better timing insights (how long each stage took) and gives you faster root cause/logs (i.e. you know that ETL or ferc_to_sqlite failed w/o having this intermixed with integration test outputs/logs).

I suppose the missing bit here is telling tox that all these three operations should run with the same tox environment and that we'd like to cache this environment between github action steps. Let me investigate some more.

.github/workflows/tox-pytest.yml
rousik: This was unfinished attempt to see if we'd like to persist outputs of these as artifacts which could aid analysis, but let's defer this until later.

rousik:
I have switched this back to use `tox` to run the commands && perhaps made it such that coverage is exported both for `ferc_to_sqlite` and `pudl_etl` (I still need to figure out how to verify that).

I have also installed tox plugins that should allow us to reuse the same environment for all of the three actions. The only remaining question is whether tox and GHAs are interoperable for caching/reuse of envs across separate runs. My initial assessment seems to reveal a lot of people struggling with this and if we can't cache this, we may be paying recurrent cost for rebuilding the same env over and over.

tox.ini
```diff
@@ -82,6 +86,35 @@ commands =
...
+commands =
+    bash -c 'coverage run --append src/pudl/ferc_to_sqlite/cli.py {gcs_cache} src/pudl/package_data/settings/etl_fast.yml'
+
+[testenv:ci_pudl_etl]
+description = Run the ETL script in the CI environment.
+envdir = {toxworkdir}/.ci_env
+runner = ignore_env_name_mismatch
+commands =
+    bash -c 'alembic upgrade head'
+    bash -c 'coverage run --append src/pudl/etl/cli.py {gcs_cache} src/pudl/package_data/settings/etl_fast.yml'
```
zaneselvans, line 108: When I tried running the ETL with coverage like this, it fully utilized my CPUs, but I only got ~35% coverage, which seemed to reflect what you'd get from just importing PUDL, with all of the code that was actually run by the Dagster processes not showing up.

.github/workflows/tox-pytest.yml
```diff
@@ -1,8 +1,14 @@
 ---
 name: tox-pytest
 
-on: [pull_request, workflow_dispatch]
-
+on:
+  workflow_dispatch:
+  pull_reqyest:
```
zaneselvans, line 6: Typo? `pull_request`?

tox.ini
rousik: Perhaps we need to run codecov in a multi process friendly way.

zaneselvans: I tried it with `--parallel-mode` and it didn't change the behavior. I don't really understand the difference between subprocesses that the parent process can keep track of, or entirely separate processes that have been spawned some other way?

rousik:
I've fixed minor typos and tried to make sure that our `coverage` collection works. AFAICT, to make this work, coverage needs to be run with `--concurrency=multiprocessing` option which generates several `.coverage-${stuff}` files for each process, followed by `coverage combine --append` which will merge these fragments into single `.coverage` report.

Right now, I'm kind of assuming that the three operations (ferc_to_sqlite, pudl_etl and integration tests) will be run in a sequence and all append into single coverage file which will then be converted to `xml` report as part of the integration tests.

I will see if the most recent run will show promise.

It still feels that using `tox` under the hood here just makes things more complicated by adding yet another indirection layer that doesn't really do us much good here, i.e. we don't really benefit much from the tox env management (and we need to specifically force tox to reuse environments across the run) and the dependency between the steps and coverage files is just a bit painful here.

tox.ini
```diff
@@ -82,6 +85,41 @@ commands =
     bash -c 'coverage run --source={envsitepackagesdir}/pudl -- $(which sphinx-build) -W -b html docs docs/_build/html'
     coverage xml
 
+###########################################################################
+# Continuous integration tests.
+# 
+# The following steps are run in sequence as part of the CI workflows on
+# github to run the ETL and integration tests in a modular fashion.
+###########################################################################
+[testenv:ci_ferc_to_sqlite]
```
zaneselvans, line 94: Let's make these environment names specific to github since that's the only place they are supposed to be run, and they can have destructive consequences for the user's environment.  "CI" is ambiguous within our vocabulary -- the main Tox environment has always been named `ci` and we don't necessarily think of CI as *only* running remotely.

```
[testenv:github_ferc_to_sqlite]
```

zaneselvans: For all of the `github_*` environments, can we make sure that they only run if the `$GITHUB_ACTIONS` environment variable is set, to avoid unexpected destructive side-effects within the user's `$DAGSTER_HOME`? `$GITHUB_ACTIONS` is already automatically set on the runners. I.e. if someone is fooling around locally trying to run these tox test environments they won't run, unless they go out of their way to set a `$GITHUB_ACTIONS` variably locally.

zaneselvans (changes requested):

tox.ini
rousik: These requests are now addressed.

zaneselvans:
@jdangerx it looks like this PR is working now, in that it runs the ETL before the tests and collects coverage from all of that, but it also doesn't seem to have resulted in any speedup whatsoever, even though it's running on the bigger runner. Any ideas as to why that might be?  At least looking at the `ferc-to-sqlite` output, it seems like the XBRL extraction is happening in groups of 4 parallel chunks at a time.

rousik:
> It looks like the coverage isn't completely working since it dropped from 88% to 59%. This is more coverage than we get from just doing the imports (which IIRC is like 30%) so _some_ coverage is being collected, but apparently a bunch is still getting lost by virtue of the way the ETL is getting run.
> 
> Hmm, looking at [the coverage report for the PR](https://app.codecov.io/gh/catalyst-cooperative/pudl/pull/2825/indirect-changes) it looks like huge swaths of the transform functions have lost coverage, so I assume the `pudl_etl` isn't actually collecting any coverage. The `ferc_to_sqlite` CLI module also doesn't appear to be getting covered, so I don't think the coverage from either script is working.

Hmph. That is unfortunate. I'm not sure what else to try to make `coverage` interop well with dagster then :-/

rousik:
> @jdangerx it looks like this PR is working now, in that it runs the ETL before the tests and collects coverage from all of that, but it also doesn't seem to have resulted in any speedup whatsoever, even though it's running on the bigger runner. Any ideas as to why that might be? At least looking at the `ferc-to-sqlite` output, it seems like the XBRL extraction is happening in groups of 4 parallel chunks at a time.

If you look at top of this run https://github.com/catalyst-cooperative/pudl/actions/runs/6576790075/job/17867156644#step:1:2, you can see runner name `ubuntu-22.04-4core_674e4143c3a6` which would imply 4 core runner.

.github/workflows/tox-pytest.yml
```diff
@@ -138,20 +142,6 @@ jobs:
...
-          conda list
-          conda config --show-sources
-          conda config --show
-          printenv | sort
-
-      - name: Log SQLite3 version
-        run: |
-          which sqlite3
-          sqlite3 --version
-
```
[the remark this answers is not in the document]
rousik: I may not know how useful this diagnostic signal is. Was this put in place in response to a problem caused by a wrong version? Has that helped diagnosing subsequent problems? I'm generally in favor of having diagnostic information, but I also think it's perhaps good to keep the output succint and focused, and generating large swaths of outputs might drown the useful signal.

I would be happy to hear more about the value of these diagnostic signals.

.github/workflows/tox-pytest.yml
```diff
@@ -166,13 +168,20 @@ jobs:
           service_account: "tox-pytest-github-action@catalyst-cooperative-pudl.iam.gserviceaccount.com"
 
       - name: Run ferc_to_sqlite
-        run: tox -e github_ferc_to_sqlite
-
+        run: | 
+          coverage --concurrency=multiprocessing run \
+            src/pudl/ferc_to_sqlite/cli.py --clobber ${{ env.ETL_COMMANDLINE_OPTIONS }} ${{ env.ETL_CONFIG }}
```
zaneselvans: Previously, Tox was responsible for installing all of the software dependencies, so everything that's not part of the `test-environment.yml` (which is almost everything) will need to be installed separately somehow if Tox isn't being used.  On the `conda-lockfile` branch that's done using `micromamba` and the locked dependencies.

rousik: Yeah, this is now installed explicitly via `pip install .[test,datasette]` command above.

rousik:
We have determined that parallelism works as expected (the ETL just got a lot slower), so the only remaining piece is coverage, which still seems to be broken. I will try to dig into that some more to see if there's anything that can be done about it.

rousik:
I do not know what changed, but it seems that coverage works well now :shrug: so this should be good to go

zaneselvans:
Somehow it seems like the integration tests are still taking the same amount of time to run. Do we expect to get a speedup from parallelism? Or is that still not enabled here?

zaneselvans:
@rousik It doesn't seem like the coverage results are showing up at all now.

zaneselvans:
Hmm, so the coverage is sadly not working 😢

.coveragerc
```diff
@@ -1,4 +1,7 @@
 [run]
+source =
+    src
+    test
```
zaneselvans, line 4: We haven't historically included `test` in the test coverage, so if it works, this may result in a decrease in overall coverage.  Looking at our `Makefile` the coverage arguments we have defined are:

For use with `coverage run`:
```
--append --source=src/pudl
```

For use with `pytest`:
```
--cov-append --cov=src/pudl --cov-report=xml
```

The command we use to generate the coverage report:
```
coverage report --sort=cover
```

zaneselvans:
@rousik I merged `dev` in to get the `pre-commit` version of the lockfile updates, which should involve much less churn.

It seems like it's *mostly* getting the coverage, with 90.1%. Are the tests themselves not being covered now maybe?

zaneselvans:
@rousik Can you update the main description for this PR to reflect the current situation? It's been through so many changes and the code on `main` has changed so much that I'm no longer clear what the net effect is.

The integration test run time seems to be almost identical to the current tests (~1.5 hours).

Is the primary change that this PR is running the ETL outside the context of `pytest`?  Where were we going with that? Is the next step to kick off the FERC to SQLite conversions as their own jobs on smaller runners which can run in parallel rather than serially, and have the main PUDL ETL start on a larger runner as soon as the FERC 1 DBs are available, hopefully shortening the overall integration test runtime and also reducing the running time for the paid runner?

rousik:
> @rousik Can you update the main description for this PR to reflect the current situation? It's been through so many changes and the code on `main` has changed so much that I'm no longer clear what the net effect is.
> 
> The integration test run time seems to be almost identical to the current tests (~1.5 hours).
> 
> Is the primary change that this PR is running the ETL outside the context of `pytest`? Where were we going with that? Is the next step to kick off the FERC to SQLite conversions as their own jobs on smaller runners which can run in parallel rather than serially, and have the main PUDL ETL start on a larger runner as soon as the FERC 1 DBs are available, hopefully shortening the overall integration test runtime and also reducing the running time for the paid runner?

I suppose I need to rerun some of these tests again, because we should be getting speed-ups from this even if nothing more is done. Isolating `ferc_to_sqlite`, `pudl_etl` and integration tests in separate steps should make it a little more easy to diagnose which parts failed, and should also (as pointed out) allow us to further speed-up/improve the individual steps as needed. We could beef-up the `pudl_etl` part, hopefully getting faster there, and then sharding `ferc_to_sqlite` which should be a good time improvement.

rousik:
> @rousik Can you update the main description for this PR to reflect the current situation? It's been through so many changes and the code on `main` has changed so much that I'm no longer clear what the net effect is.
> 
> The integration test run time seems to be almost identical to the current tests (~1.5 hours).
> 
> Is the primary change that this PR is running the ETL outside the context of `pytest`? Where were we going with that? Is the next step to kick off the FERC to SQLite conversions as their own jobs on smaller runners which can run in parallel rather than serially, and have the main PUDL ETL start on a larger runner as soon as the FERC 1 DBs are available, hopefully shortening the overall integration test runtime and also reducing the running time for the paid runner?

I have re-read contents of this PR and description and I think they are still in sync. This refactors ci-integration into smaller individual tasks.

.github/workflows/pytest.yml
```diff
@@ -8,11 +8,16 @@ on:
       - opened
       - synchronize
       - ready_for_review
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
+  cancel-in-progress: true
```
zaneselvans, line 13: Does `cancel-in-progress` mean that another push to the same PR will (say) kill an already running `ci-integration` job, and replace it with a new one?

.github/workflows/pytest.yml
```diff
@@ -105,31 +109,31 @@ jobs:
           path: coverage.xml
 
   ci-integration:
-    runs-on:
-      group: large-runner-group
-      labels: ubuntu-22.04-4core
+    needs:
+      - ci-unit
```
zaneselvans, line 113: What's your thought on including this dependency between jobs? Is it just to avoid running the long, expensive `ci-integration` tests if we already know something is wrong?

.github/workflows/pytest.yml
```diff
@@ -105,31 +109,31 @@ jobs:
...
+    runs-on: ubuntu-22.04-4core
     if: github.event.pull_request.draft == false
     permissions:
       contents: read
       id-token: write
-    strategy:
-      fail-fast: false
```
zaneselvans, line 116: Why do we want `fail-fast` enabled, given that we've already the dependency above specified? I think this would mean that if the docs build fails, the unit and integration tests would be canceled, right?

.github/workflows/pytest.yml
```diff
@@ -169,22 +173,41 @@ jobs:
...
+          coverage run --concurrency=multiprocessing \
+            src/pudl/etl/cli.py ${{ env.ETL_COMMANDLINE_OPTIONS }} ${{ env.ETL_CONFIG }}
+      - name: Run integration tests
+        env:
+          COVERAGE_FILE: .coverage.pytest
+        run: |
+          coverage run --concurrency=multiprocessing \
+           -m pytest -n auto --live-dbs test/integration
+      - name: Checksum coverage files
+        run: ls .coverage* | xargs md5sum | sort
```
zaneselvans, line 196: Why is it useful to checksum the coverage files?

.github/workflows/pytest.yml
```diff
@@ -169,22 +173,41 @@ jobs:
...
+      - name: Generate coverage
+        run: |
+          coverage --version
+          coverage combine
```
zaneselvans, line 200: Why do we need to do `coverage combine` here? Is it because we've got a bunch of coverage files from various different processes?

.github/workflows/pytest.yml
```diff
@@ -198,37 +221,40 @@ jobs:
         with:
           path: coverage
```
zaneselvans, line 222: Why do we use `path: coverage` here but when we're uploading the artifact above we use `path: coverage.xml`?

pyproject.toml
```diff
@@ -310,11 +310,31 @@ curl = ">=8.4.0"
...
 [tool.coverage.run]
 # See note above on need to specify separate sources for pytest-coverage and coverage.
-source = ["src/pudl/", "test/integration/", "test/unit/"]
+# source = ["src/pudl/", "test/integration/", "test/unit/"]
+include = [
+  "src/pudl/**",
+  "test/integration/**",
+  "test/unit/**",
+  "*/site-packages/pudl/**",
+]
```
zaneselvans, line 319: What's going on with the change from source to include here?  Should these be identical to the `pudl_sources` below? Should the commented out `source` line be removed? Is the comment referring above still relevant? Does switching from source to include mean we don't need the `--cov=...` arguments above?

.github/workflows/pytest.yml
```diff
@@ -204,30 +222,33 @@ jobs:
...
-                color: '${{ needs.ci-test.result }}' === 'success' ? 'good' : '${{ needs.ci-test.result }}' === 'failure' ? 'danger' : 'warning',
-                text: `${process.env.AS_REPO}@${process.env.AS_REF}\n ${process.env.AS_WORKFLOW} (${process.env.AS_COMMIT})\n by ${process.env.AS_AUTHOR}\n Status: ${{ needs.ci-test.result }}`,
-              }]
-            }
-        env:
-          GITHUB_TOKEN: ${{ github.token }} # required
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }} # required
-          MATRIX_CONTEXT: ${{ toJson(matrix) }} # required
+  # TODO(rousik): The following slack notification gives no value and
+  # needs to be fixed. Until then, it might be better to do nothing
```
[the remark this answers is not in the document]
zaneselvans: I think it's fine to remove this Slack notification entirely. We have so many CI tests running these days that it's really just a source of noise.

zaneselvans (changes requested):
Mostly requests for clarification on why some things are changing.

rousik (commented):
Huh, lots of my comments were left in limbo. Posting.

.github/workflows/pytest.yml
zaneselvans: I think this is probably better. But we should make sure we tell everyone about the change since I've also definitely used the fact that they don't get killed to test a couple of things in parallel.

pyproject.toml
```diff
@@ -310,11 +310,30 @@ curl = ">=8.4.0"
...
+[tool.coverage.paths]
+# When running pudl tools installed with pip, the sources are imported
+# from package-data/pudl directory. The following maps this to raw
+# source files.
+pudl_sources = [
+  "src/pudl/",
+  "*/site-packages/pudl/",
+  "test/unit",
+  "test/integration",
+]
```
zaneselvans, line 336: I'm re-running the CodeCov tests now, but it seems like something has changed that's reducing the test coverage by a few hundred lines / a couple of percent, and I don't understand what it is.

Looking at the [indirect coverage changes](https://app.codecov.io/gh/catalyst-cooperative/pudl/pull/2825/indirect-changes#) some differences make sense, like increased coverage for the `ferc_to_sqlite` and `pudl_etl` CLIs, but there's also substantial apparent losses of coverage in modules related to the FERC-EIA record linkage, and the `pudltabl` output module which don't make sense.

If I run
```console
make pytest-coverage
```
locally I get overall coverage of 92.8% which is comparable to the coverage on `main`
```
Name                                                           Stmts   Miss  Cover
----------------------------------------------------------------------------------
src/pudl/__init__.py                                              14      0 100.0%
src/pudl/analysis/__init__.py                                      1      0 100.0%
src/pudl/analysis/fuel_by_plant.py                                37      0 100.0%
src/pudl/analysis/record_linkage/__init__.py                       1      0 100.0%
src/pudl/convert/__init__.py                                       1      0 100.0%
src/pudl/etl/__init__.py                                          22      0 100.0%
src/pudl/etl/eia_bulk_elec_assets.py                               9      0 100.0%
src/pudl/etl/epacems_assets.py                                    54      0 100.0%
src/pudl/etl/static_assets.py                                     24      0 100.0%
src/pudl/extract/__init__.py                                       1      0 100.0%
src/pudl/extract/csv.py                                           29      0 100.0%
src/pudl/extract/eia176.py                                         9      0 100.0%
src/pudl/extract/eia860.py                                        40      0 100.0%
src/pudl/extract/eia860m.py                                       30      0 100.0%
src/pudl/extract/eia923.py                                        51      0 100.0%
src/pudl/extract/eia_bulk_elec.py                                 46      0 100.0%
src/pudl/extract/epacems.py                                       47      0 100.0%
src/pudl/extract/ferc.py                                          14      0 100.0%
src/pudl/extract/ferc2.py                                         21      0 100.0%
src/pudl/extract/ferc6.py                                         15      0 100.0%
src/pudl/extract/ferc60.py                                        15      0 100.0%
src/pudl/extract/ferc714.py                                       21      0 100.0%
src/pudl/ferc_to_sqlite/__init__.py                               24      0 100.0%
src/pudl/glue/__init__.py                                          1      0 100.0%
src/pudl/metadata/__init__.py                                      1      0 100.0%
src/pudl/metadata/codes.py                                         5      0 100.0%
src/pudl/metadata/constants.py                                    21      0 100.0%
src/pudl/metadata/dfs.py                                          12      0 100.0%
src/pudl/metadata/enums.py                                        21      0 100.0%
src/pudl/metadata/fields.py                                       25      0 100.0%
src/pudl/metadata/labels.py                                       12      0 100.0%
src/pudl/metadata/resources/__init__.py                           12      0 100.0%
src/pudl/metadata/resources/allocate_gen_fuel.py                   4      0 100.0%
src/pudl/metadata/resources/eia.py                                 4      0 100.0%
src/pudl/metadata/resources/eia860.py                              3      0 100.0%
src/pudl/metadata/resources/eia861.py                              2      0 100.0%
src/pudl/metadata/resources/eia923.py                              4      0 100.0%
src/pudl/metadata/resources/eia_bulk_elec.py                       2      0 100.0%
src/pudl/metadata/resources/epacems.py                             3      0 100.0%
src/pudl/metadata/resources/ferc1.py                               4      0 100.0%
src/pudl/metadata/resources/ferc1_eia_record_linkage.py            3      0 100.0%
src/pudl/metadata/resources/ferc714.py                             3      0 100.0%
src/pudl/metadata/resources/glue.py                                3      0 100.0%
src/pudl/metadata/resources/mcoe.py                                4      0 100.0%
src/pudl/metadata/resources/pudl.py                                3      0 100.0%
src/pudl/metadata/sources.py                                       5      0 100.0%
src/pudl/output/__init__.py                                        1      0 100.0%
src/pudl/output/eia860.py                                         19      0 100.0%
src/pudl/output/eia_bulk_elec.py                                  14      0 100.0%
src/pudl/output/sql/helpers.py                                    12      0 100.0%
src/pudl/resources.py                                             17      0 100.0%
src/pudl/transform/__init__.py                                     1      0 100.0%
src/pudl/transform/eia_bulk_elec.py                               26      0 100.0%
src/pudl/transform/ferc714.py                                     76      0 100.0%
src/pudl/transform/params/__init__.py                              1      0 100.0%
src/pudl/transform/params/ferc1.py                                67      0 100.0%
src/pudl/workspace/__init__.py                                     1      0 100.0%
test/integration/console_scripts_test.py                          23      0 100.0%
test/integration/datasette_metadata_test.py                       22      0 100.0%
test/integration/etl_test.py                                      76      0 100.0%
test/integration/ferc1_eia_train_test.py                          30      0 100.0%
test/integration/ferc_dbf_extract_test.py                         38      0 100.0%
test/integration/zenodo_datapackage_test.py                       11      0 100.0%
test/unit/analysis/plant_parts_eia_test.py                        48      0 100.0%
test/unit/analysis/spatial_test.py                                67      0 100.0%
test/unit/analysis/state_demand_test.py                            8      0 100.0%
test/unit/analysis/timeseries_cleaning_test.py                    32      0 100.0%
test/unit/console_scripts_test.py                                 10      0 100.0%
test/unit/extract/csv_test.py                                     37      0 100.0%
test/unit/extract/eia_bulk_elec_test.py                           42      0 100.0%
test/unit/extract/xbrl_test.py                                    74      0 100.0%
test/unit/helpers_test.py                                        135      0 100.0%
test/unit/output/epacems_test.py                                   7      0 100.0%
test/unit/output/ferc1_test.py                                     9      0 100.0%
test/unit/transform/classes_test.py                              218      0 100.0%
test/unit/transform/eia923_test.py                                12      0 100.0%
test/unit/transform/eia_bulk_elec_test.py                         18      0 100.0%
test/unit/transform/epacems_test.py                                8      0 100.0%
test/unit/transform/ferc1_test.py                                133      0 100.0%
test/unit/transform/glue_test.py                                  11      0 100.0%
test/unit/workspace/datastore_test.py                             82      0 100.0%
test/unit/workspace/resource_cache_test.py                       140      0 100.0%
src/pudl/extract/eia861.py                                        36      1  97.2%
src/pudl/extract/ferc1.py                                        113      1  99.1%
test/integration/epacems_test.py                                  39      1  97.4%
test/integration/output_test.py                                   69      1  98.6%
test/unit/settings_test.py                                       145      1  99.3%
src/pudl/__main__.py                                               2      2   0.0%
src/pudl/analysis/mcoe.py                                         78      2  97.4%
src/pudl/analysis/record_linkage/classify_plants_ferc1.py         31      2  93.5%
src/pudl/etl/glue_assets.py                                      112      2  98.2%
src/pudl/glue/ferc1_eia.py                                       135      2  98.5%
src/pudl/output/censusdp1tract.py                                 28      2  92.9%
src/pudl/output/eia923.py                                         93      2  97.8%
src/pudl/workspace/setup.py                                       43      2  95.3%
test/unit/analysis/allocate_gen_fuel_test.py                      96      2  97.9%
test/unit/extract/excel_test.py                                   49      2  95.9%
test/unit/harvest_test.py                                         59      2  96.6%
test/unit/metadata_test.py                                        56      2  96.4%
src/pudl/extract/xbrl.py                                          52      3  94.2%
src/pudl/logging_helpers.py                                       15      3  80.0%
src/pudl/output/epacems.py                                        29      3  89.7%
src/pudl/settings.py                                             253      3  98.8%
src/pudl/transform/epacems.py                                     34      3  91.2%
test/integration/record_linkage_test.py                           72      3  95.8%
src/pudl/convert/censusdp1tract_to_sqlite.py                      35      4  88.6%
src/pudl/metadata/helpers.py                                     187      4  97.9%
test/integration/glue_test.py                                     42      4  90.5%
test/unit/glue.py                                                  7      4  42.9%
src/pudl/output/ferc714.py                                       150      5  96.7%
test/integration/jupyter_notebooks_test.py                        13      5  61.5%
src/pudl/analysis/record_linkage/embed_dataframe.py              124      6  95.2%
src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py     218      7  96.8%
src/pudl/convert/metadata_to_rst.py                               21      7  66.7%
src/pudl/analysis/spatial.py                                     112      8  92.9%
src/pudl/etl/check_foreign_keys.py                                20      8  60.0%
src/pudl/extract/excel.py                                        167      8  95.2%
src/pudl/transform/eia.py                                        300      8  97.3%
src/pudl/analysis/plant_parts_eia.py                             260      9  96.5%
src/pudl/workspace/resource_cache.py                             124      9  92.7%
src/pudl/transform/eia861.py                                     398     10  97.5%
test/unit/io_managers_test.py                                    206     10  95.1%
src/pudl/output/pudltabl.py                                       98     11  88.8%
src/pudl/analysis/epacamd_eia.py                                  18     12  33.3%
src/pudl/analysis/record_linkage/link_cross_year.py              106     12  88.7%
src/pudl/analysis/record_linkage/name_cleaner.py                  90     12  86.7%
src/pudl/analysis/allocate_gen_fuel.py                           323     13  96.0%
src/pudl/analysis/state_demand.py                                146     15  89.7%
src/pudl/ferc_to_sqlite/cli.py                                    44     15  65.9%
src/pudl/transform/classes.py                                    394     21  94.7%
src/pudl/etl/cli.py                                               48     22  54.2%
src/pudl/transform/eia860.py                                     205     23  88.8%
src/pudl/extract/dbf.py                                          258     25  90.3%
src/pudl/workspace/datastore.py                                  298     25  91.6%
src/pudl/io_managers.py                                          268     26  90.3%
src/pudl/analysis/service_territory.py                           123     31  74.8%
src/pudl/transform/eia923.py                                     231     34  85.3%
src/pudl/analysis/timeseries_cleaning.py                         460     44  90.4%
src/pudl/helpers.py                                              407     53  87.0%
src/pudl/output/ferc1.py                                         632     66  89.6%
src/pudl/transform/ferc1.py                                     1545     68  95.6%
src/pudl/output/eia.py                                           176     72  59.1%
src/pudl/analysis/record_linkage/eia_ferc1_train.py              172     79  54.1%
src/pudl/metadata/classes.py                                     798    101  87.3%
----------------------------------------------------------------------------------
TOTAL                                                          13072    943  92.8%
```

* `test/unit/glue.py` needs to be renamed so pytest collects tests from it, but that's not a new thing.
* `test/integration/console_scripts_test.py` doesn't get hit all in the CodeCov report but has 100% coverage if I run locally.
* `src/pudl/analysis/record_linkage/eia_ferc1_train.py` coverage is dramatically lower in CodeCov than locally.
* `src/pudl/analysis/service_territory.py` is also significantly lower. My guess is this is related to the scripts not getting tested?
* `src/pudl/output/pudltabl.py` is basically not getting hit at all in CodeCov which makes no sense to me, since the PudlTabl objects are used all over the place in the tests and have 88.8% coverage when run locally.

zaneselvans (commented):
Comparing locally generated test coverage and the CodeCov output from CI.

pyproject.toml
rousik: Sorry for delays, life has been busy. I will try to follow up on these in the coming week and try to figure out what is going on here. The fact that coverage is so brittle and it is so hard to figure out why it fluctuates so much is very discouraging; it shouldn't be that hard :-/

PR #2832, opened 2023-09-04
Convert the FERC exploded forest to a table for readability

cmgosnell:
# PR Overview
This PR adds a few methods to the forest class that converts a forest's relationships into a big ol table. Starting from the seed node, I grab that nodes' direct children and merge those children with their direct parents. I used `_level{n}` to indicate which step in the tree the nodes refer to.

I decided to use the `calculation_components_ferc1` as the core building blocks of the table instead of just using a decomposed forest. I think using only a forest could have done the job just fine, but I thought it would a) be a nice check on the forest and b) be slightly more straightforward bc the calc components table is already a table. Using the calc components table, each time you want to get a level of parent/child connections from the calc comps table I first select for the parent nodes and then select the child nodes from that subset of the table. This way you never get parents records that have children that got pruned from the forest and you never have children records that refer to a different parent.

Once the subset of the calc components table is selected each round, a new set of `_level{n+1}` nodes gets merged onto their `_level{n}` parents.

Other notes:
- I added in the `weight` of the children for every level.

# PR Checklist


zaneselvans:
I cancelled the CI since it was already failing. And the failure is on GitHub's side. Some issue with actions right now sitewide.

https://www.githubstatus.com/history

Issue #2834, opened 2023-09-05
Missing retirement dates from 860m data


arengel:
### Describe the bug
The `planned_generator_retirement_date` and `generator_retirement_date` columns are empty for monthly_update generator data.

### Bug Severity
How badly is this bug affecting you?
 - **Medium:** This bug is preventing me from adopting PUDL for 860m.

### To Reproduce
Read the data from the database downloaded Sept 5, 2023 at noon.

```python
import pandas as pd
import sqlalchemy as sa

gens = pd.read_sql_table(
    table_name="generators_eia860",
    con=sa.create_engine("sqlite:////.../pudl.sqlite").connect(),
).query("data_maturity == 'monthly_update'")
gens_denorm = pd.read_sql_table(
    table_name="denorm_generators_eia",
    con=sa.create_engine("sqlite:////.../pudl.sqlite").connect(),
).query("data_maturity == 'monthly_update'")
```

Then looking at the specific columns, all the values are `pd.NaT`.

```python
gens.planned_generator_retirement_date.unique()
Out[1]: array(['NaT'], dtype='datetime64[ns]')

gens_denorm.planned_generator_retirement_date.unique()
Out[2]: array(['NaT'], dtype='datetime64[ns]')

gens.generator_retirement_date.unique()
Out[3]: array(['NaT'], dtype='datetime64[ns]')

gens_denorm.generator_retirement_date.unique()
Out[4]: array(['NaT'], dtype='datetime64[ns]')
```

### Expected behavior
These columns would be populated with retirement and planned retirement data from EIA 860 monthly.

### Software Environment?
 - Operating System: MacOS 13.5.1 on ARM
 - Python version: 3.11.5 | packaged by conda-forge | (main, Aug 27 2023, 03:33:12) [Clang 15.0.7 ]
 - How did you install PUDL?
    - I didn't install PUDL just read the data in from the sqlite DB.

zaneselvans:
Whatever the issue ends up being here, we should make sure we add a new data validation test that checks for entirely NULL values in these columns when selecting just monthly update records.

aesharpe:
The monthly EIA data that gets subsumed into tables like the `generators_eia860` is supposed to have the months delineated in the report_date column, correct? Like `2023-`06-01`? Unless the `data_maturity` is not `monthly_updates` in which case it would all say `2023-01-01`. I assumed this was already happening, but it looks like the raw EIA860m data extraction does not preserve the month, just the year.
[...]

PR #2835, opened 2023-09-06
Add generator to the retirement_year / _month columns for the monthly…

aesharpe:
From issue #2834 

The monthly data retirement dates were getting mapped to `retirement_year` and `retirement_month` instead of `generator_retirement_year` and `generator_retirement_month` in the mapping spreadsheets. Only the `generator_retirement_` column is included in the metadata for the cleaned version of the `generators_eia860` so all the values for retirement date from the monthly data were getting dropped.

I updated the spreadsheet to map to `generator_retirement_year` and `generator_retirement_month` which should fix the problem.


zaneselvans (approved):
Aye, glad it was a simple fix.

I feel like this should have raised a warning or an error somewhere in the extract step due to mismatched column names.

zaneselvans:
I'm going to go ahead and merge this so it can be part of the builds tonight.

Issue #2836, opened 2023-09-07
Nightly Docker container build fails due to missing GLIBCXX library


zaneselvans:
For the last week or so the Docker container builds associated with our nightly deployments [have been failing](https://github.com/catalyst-cooperative/pudl/actions/runs/6106021922/job/16570419193).

Ultimately the error comes from attempting to run `pudl_setup` -- so the environment has been solved and the container is essentially built, but we get a runtime error, apparently having to do with pandas.  It may be that this error is related to the upgrade to pandas 2.0?

```
[stage-0 11/11] RUN --mount=type=bind,source=.git,target=/home/catalyst/pudl/.git     conda run --no-capture-output --prefix /home/catalyst/env pip install --no-cache-dir -e './[dev,doc,test,datasette]' &&     conda run --no-capture-output --prefix /home/catalyst/env pudl_setup:
122.2   File "/home/catalyst/env/lib/python3.11/site-packages/pandas/core/frame.py", line 182, in <module>
122.2     from pandas.core.generic import NDFrame
122.2   File "/home/catalyst/env/lib/python3.11/site-packages/pandas/core/generic.py", line 179, in <module>
122.2     from pandas.core.window import (
122.2   File "/home/catalyst/env/lib/python3.11/site-packages/pandas/core/window/__init__.py", line 1, in <module>
122.2     from pandas.core.window.ewm import (
122.2   File "/home/catalyst/env/lib/python3.11/site-packages/pandas/core/window/ewm.py", line 11, in <module>
122.2     import pandas._libs.window.aggregations as window_aggregations
122.2 ImportError: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /home/catalyst/env/lib/python3.11/site-packages/pandas/_libs/window/aggregations.cpython-311-x86_64-linux-gnu.so)
122.3 ERROR conda.cli.main_run:execute(41): `conda run pudl_setup` failed. (See above for error)
```

## Attempting to reproduce the problem locally

Building the container myself I noticed some things:

* The `mambaforge` image that we have been using is more than a year out of date.
* We're using Python 3.10 initially in creating the python environment.
* We've got the `google-cloud-sdk` package pinned to a very old version.
* My container build fails when attempting to create the `pudl-test` environment because it claims there's no `google-cloud-sdk` package available, but that package definitely appears to exist and installs just fine locally.  I have no trouble creating the `pudl-test` environment.
[...]

PR #2837, opened 2023-09-07
Fix docker builds

zaneselvans:
# PR Overview

* Debug docker build failures that started August 29th see #2836 
* Update `mambaforge` Docker image to a recent version -- we were a year out of date.
* Use Python 3.11 in the base environment in the Docker container.
* Pin `pyarrow<13` for the time being as that introduced the `GLIBCXX` issue.
* Allow a much more recent version of the `google-cloud-sdk` in our test environment.
* Create a `docker-build-test` workflow that runs on `pull_request` to avoid this kind of issue in the future.

# PR Checklist

### DIFF
--- a/.github/workflows/docker-build-test.yml
+++ b/.github/workflows/docker-build-test.yml
@@ -1,0 +1,36 @@
+---
+name: docker-build-test
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  pudl_docker_build:
+    name: Test building the PUDL ETL Docker image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Docker Metadata
+        id: docker_metadata
+        uses: docker/metadata-action@v4.4.0
+        with:
+          images: catalystcoop/pudl-etl
+          flavor: |
+            latest=auto
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2.5.0
+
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-build-test.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: docker/setup-buildx-action@v2.5.0
+
+      - name: Build image but do not push to Docker Hub
+        uses: docker/build-push-action@v4.0.0
+        with:
+          context: .
+          file: docker/Dockerfile
+          push: false
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
jdangerx, line 36: We'll need to be careful about hitting our cache capacity eventually, but this seems fine for now!

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "networkx>=2.2,<3.2",
     "numpy>=1.18.5,!=1.23.0,<1.26",
     "pandas[parquet,excel,fss,gcp,compression]>=2.0,<2.1",
-    "pyarrow>=5,<13.1",
+    "pyarrow>=7,<13",
```
jdangerx, line 32: Glad the fix was to bump the lower end instead of the bring the higher end down :)

jdangerx (approved):

pyproject.toml
zaneselvans: No no, sadly it was bringing the top down. We'll have to figure out wtf is going on here, or maybe we'll get lucky and it'll be fixed when we update to Pandas 2.1.  I raised the minimum version since I know that's going to be the minimum allowed version in Pandas when PyArrow becomes a mandatory dependency.

PR #2839, opened 2023-09-08
Skip integraton tests on draft PRs

jdangerx:
To save time/energy/$, we can only run the long CI jobs on PRs that are nominally ready for review.

Testing plan:
* [x] make this a draft PR, and see if ci-integration runs
* [x] turn this into a non-draft PR, and see if ci-integration runs

### DIFF
--- a/.github/workflows/tox-pytest.yml
+++ b/.github/workflows/tox-pytest.yml
@@ -1,7 +1,13 @@
 ---
 name: tox-pytest
 
-on: [pull_request]
+on:
+  pull_request:
+    types:
+      - created
+      - opened
+      - synchronize
+      - ready_for_review
 
 env:
   PUDL_OUTPUT: /home/runner/pudl-work/output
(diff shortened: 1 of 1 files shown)


rousik (approved):

.github/workflows/tox-pytest.yml
```diff
@@ -1,7 +1,13 @@
 ---
 name: tox-pytest
 
-on: [pull_request]
+on:
+  pull_request:
```
zaneselvans, line 5: Might we want to add a `workflow_dispatch` here so someone can manually kick off the integration tests once without needing to change the status of the PR?

jdangerx: Maybe. I'm not sure how the workflow_dispatch event will interact with the `if:` expression, since then presumably `pull_request.event` won't be defined.

I'm inclined to keep this change as-is and not deal with that, since it's pretty easy to change the status of the PR to get integration tests running if one wants them.

bendnorman:
This is great thanks @jdangerx !

Issue #2850, opened 2023-09-11
Add keywords to Zenodo archive to fix dataset validation issue


e-belfer:
Right now datasets can be archived to Zenodo without keywords. In the `datapackage.json` file for each dataset, this leaves an empty `keywords: []` field. When we run `pudl_datastore` on these archives, we are left with a validation error that prevents data extraction. To fix this issue, do the following:

- [x] add keywords to all datasets missing them in `pudl.metadata.constants`
- [x] remake environment in pudl-archiver
- [x] re-run archives
- [x] https://github.com/catalyst-cooperative/pudl-archiver/pull/167

In the long term, this should be more simply addressed through #1684. Partial duplicate of issue [#80](https://github.com/catalyst-cooperative/pudl-archiver/issues/80) in pudl-archiver repo.

Issue #80, opened 2017-04-01
Functionalize table CSV dump & load


zaneselvans:
In all the EIA923 ingest functions there's repeated code right now for dumping out the dataframe to CSV and using postgres_copy to load it into the DB. This should be replaced with a re-usable function that does it cleanly, and also allows you to specify what directory the CSV files get put into so they don't end up littering the repository the way they do now.

Also need to create a function to convert float fields in the dataframe that are actually integer fields... that have NaN values in them, to strings in the CSV file that is output (with empty strings as the NaN value) for loading with postgres.

PR #2851, opened 2023-09-11
Fix missing keywords in archived datasources

e-belfer:
Addresses #2850 and [issue #80 in the pudl-archiver repo](https://github.com/catalyst-cooperative/pudl-archiver/issues/80). The lack of keywords is a blocking issue for PHMSA data extraction using the pudl Datastore, so this iceboxed issue is now a priority again. While I'm at it, I may as well fix all the other datasets that are missing keywords / have inadequate ones.

# PR Checklist

### DIFF
--- a/src/pudl/metadata/constants.py
+++ b/src/pudl/metadata/constants.py
@@ -259,6 +259,34 @@ KEYWORDS: dict[str, list[str]] = {
         "government",
         "federal",
     ],
+    "msha": [
+        "msha",
+        "mshamines",
+        "mine safety and health administration",
+        "mines",
+        "mining",
+        "coal",
+        "metal",
+    ],
+    "phmsa": [
+        "phmsa",
+        "phmsagas",
+        "pipelines and hazardous materials safety administration",
+        "pipelines",
+        "natural gas",
+        "transmission",
+        "distribution",
+        "gathering",
+        "liquified natural gas",
+        "lng",
+        "underground natural gas storage",
+    ],
+    "eia_water": [
+        "eia thermoelectric cooling water",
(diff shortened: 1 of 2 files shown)


src/pudl/metadata/constants.py
```diff
@@ -259,6 +259,34 @@
         "government",
         "federal",
     ],
+    "msha": [
```
zaneselvans, line 262: Might add "department of labor" in here since MSHA is under them.

src/pudl/metadata/constants.py
```diff
@@ -259,6 +259,34 @@
...
+        "msha",
+        "mshamines",
+        "mine safety and health administration",
+        "mines",
+        "mining",
+        "coal",
+        "metal",
+    ],
+    "phmsa": [
```
zaneselvans, line 271: Might add "department of transportation" and "US DOT" since PHMSA is under them.

src/pudl/metadata/constants.py
```diff
@@ -259,6 +259,34 @@
...
+        "pipelines",
+        "natural gas",
+        "transmission",
+        "distribution",
+        "gathering",
+        "liquified natural gas",
+        "lng",
+        "underground natural gas storage",
+    ],
+    "eia_water": [
```
zaneselvans, line 284: "water energy nexus" and "energy water nexus" might also be common search terms for this kind of information.

zaneselvans (approved):
Thanks for picking this up and filling in the gaps!

It looked like you did another PR to assert that these fields are filled in when we create archives too which is great.

I suggested a few additional keywords if you feel like adding them.

PR #2855, opened 2023-09-12
Expand multi-dimensional totals correctly

jdangerx:
This fixes a couple issues:

1. We were getting duplicate rows for multi-dimensional totals.
2. We were also getting a number of components where we allocated too many components to a total value. In the example below you can see that only `plant_status_parent` is totalled, but rows 1325 - 1329 have `plant_function` that is different from `plant_function_parent`.

```python
>>> dimensions = ["utility_type", "plant_function", "plant_status"]
....
....wackos = []
....for dim in dimensions:
....    dim = "plant_function"
....    wackos.append(df[
....        (df[f"{dim}_parent"] != df[dim]) & (df[f"{dim}_parent"] != "total")
....        ][[f"{dim}_parent" for dim in dimensions] + dimensions])
....
....pd.concat(wackos)


     utility_type_parent                       plant_function_parent plant_status_parent utility_type      plant_function plant_status
1325            electric                            steam_production               total     electric             general       future
1326            electric                            steam_production               total     electric             general   in_service
1327            electric                            steam_production               total     electric             general       leased
1328            electric                            steam_production               total     electric  nuclear_production       future
1329            electric                            steam_production               total     electric  nuclear_production   in_service
...                  ...                                         ...                 ...          ...                 ...          ...
1584            electric  regional_transmission_and_market_operation               total     electric        transmission   in_service
1585            electric  regional_transmission_and_market_operation               total     electric        transmission       leased
1586            electric  regional_transmission_and_market_operation               total     electric        distribution       future
1587            electric  regional_transmission_and_market_operation               total     electric        distribution   in_service
1588            electric  regional_transmission_and_market_operation               total     electric        distribution       leased
```

This is the sort of thing where I think I would like to use `hypothesis` to generate some test data instead of hard-coding some. If I can get my head around what assertions to make about the input calc_comps that would be a bonus, but I think even just passing in an empty calc_comps list is useful.


test/unit/transform/ferc1_test.py
```diff
@@ -642,79 +638,113 @@ def test_adding_parent_dimensions():
...
-    )[list(expected_total_to_subdim)]
 
     pd.testing.assert_frame_equal(
-        out_total_to_subdim.convert_dtypes(), expected_total_to_subdim.convert_dtypes()
+        out_total_to_subdim,
+        expected_total_to_subdim,
     )
 
 
 def test_multi_dims_totals():
```
jdangerx, line 704: It would be sweet to generate these (relatively simple) data inputs with `hypothesis` but that seemed out of scope of this PR.

src/pudl/transform/ferc1.py
```diff
@@ -5877,133 +5921,57 @@ def add_dimension_total_calculations(
...
     )
-    # we assigned True for the specific flag boolean columns for each dim. now fill
-    # everything else in as false.
-    flag_cols = calc_components.filter(like="is_within_dimension_").columns
-    calc_components.loc[:, flag_cols] = calc_components.loc[:, flag_cols].fillna(False)
     assert calc_components[calc_components.duplicated()].empty
     return calc_components
-
-
-def remove_non_totals_from_impiled_child_calculation_components(
```
jdangerx, line 5959: This should be taken care of already via the new merge methodology.

src/pudl/transform/ferc1.py
```diff
@@ -5877,133 +5921,57 @@ def add_dimension_total_calculations(
...
-            (calc_components[f"{dim}_parent"].isin([pd.NA, "total"]))
-            & (calc_components[f"{dim}_has_totals"])
-            & (calc_components[dim] != "total")
-        )
-        logger.info(f"{dim}: Removing {len(calc_components[remove_mask])}")
-        calc_components = calc_components[~remove_mask]
-    return calc_components
-
-
-def add_has_totals_column_for_dim(
```
jdangerx, line 5994: These cols were only used in the above method, afaict, so I removed them entirely.

src/pudl/transform/ferc1.py
```diff
@@ -5849,12 +5877,28 @@ def add_dimension_total_calculations(
...
-    In addition to calculations defining how values reported in one set of facts can
-    be aggregated resulting in a value in another fact, there are implied calculation
-    relationships between values reported within a dimension. In particular, when a
-    dimension reports a ``total`` value, we assume it is the sum of all the non-total
-    values for that dimension. This function makes these within-dimension calculation
-    relationships explicit by defining new calculations for each dimension's ``total``.
+    If a factoid is marked as a total along some dimension, then that's
+    implicitly a calculation. Make these explicit.
+
+    The multi-dimension totals logic is a little hard to follow, so a concrete
```
jdangerx, line 5883: I'm happy to pair to walk through this logic - it makes more sense to *me* than the old version but that is no guarantee that it makes sense to anyone else...

test/unit/transform/ferc1_test.py
zaneselvans: I always find it difficult to understand what the structure of the test inputs are just by reading them. Does Hypothesis have strategies for generating tables with specific relational DB style constraints between the columns? e.g. many-to-one or one-to-one or ensuring some property holds true within groupbys? Or do you have some idea of how it would be easy to specify those kinds of relationships generically?

zaneselvans (approved):
This looks much simpler than the previous process, which is great.  I think I would benefit from walking through the new logic together if you have a few minutes. But I also didn't understand how it worked before so... if it's more correct now and I still don't understand it that's still an improvement. Also made a few tiny readability suggestions.

zaneselvans:
Any idea what's up with the integration test failure? So far as I know it passes on my `explode_tree_fixes` branch.

src/pudl/transform/ferc1.py
```diff
@@ -5849,12 +5877,31 @@ def add_dimension_total_calculations(
...
+    * possible plant status: "total", "in_service", "future"
+
+    In calc_components, we'll see these children for the "total", "total" parent.
+
+    * child:"factoid", "electric", "in_service"; parent: "factoid", "total", "total"
+    * child:"factoid", "electric", "future"; parent: "factoid", "total", "total"
+    * child:"factoid", "gas", "in_service"; parent: "factoid", "total", "total"
+    * child:"factoid", "gas", "future"; parent: "factoid", "total", "total"
+
+    We will also see these which are redundant with the above, and should be filtered out:
```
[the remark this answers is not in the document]
jdangerx: Yes, this is all confusing. I think factoids that have `total` along any dimension should not be included in components - instead, we should be unrolling *those* into their own component parts. That actually seems like a pretty simple check to make, also - let me write that up.

Here are the situations where the parent and child both have one total dimension, and what I expect:

* different `total` dimensions:
  ```
  child: "factoid", "gas", "total"; parent: "factoid", "total", "leased"
  ```
  Should have no component relationship, because the `total` child status is bigger than the `leased` parent status. This holds whenever there's a `total` child dimension that is not also a `total` parent dimension.
* same `total` dimension:
  ```
  child: "factoid", "gas", "total"; parent: "factoid", "electric", "total"
  child: "factoid", "gas", "total"; parent: "factoid", "gas", "total"
  ```
  * The first one is not a match because the child utility type `gas` is not a conceptual subset of the parent utility type `electric`.
  * The second one is not a match because they're the same dimension

src/pudl/transform/ferc1.py
```diff
@@ -5681,10 +5680,39 @@ def calculation_components_xbrl_ferc1(**kwargs) -> pd.DataFrame:
...
+
+    For example, if utility_type_parent is not "total", then utility_type must
+    be the same as utility_type_parent.
+
+    Args:
+        calc_comps: calculation component join table
+        dimensions: list of dimensions we resolved "total" values for
+    """
+    parent_dimensions = [f"{dim}_parent" for dim in dimensions]
+    totals_mask = (calc_comps[parent_dimensions] == "total").any(axis=1)
```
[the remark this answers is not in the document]
jdangerx: Ah, yes! I'll roll these into the updates I'm about to make :) I always have to look up what 0 and 1 mean in any given axis situation so I'm glad that words exist.

src/pudl/transform/ferc1.py
```diff
@@ -5681,10 +5680,39 @@ def calculation_components_xbrl_ferc1(**kwargs) -> pd.DataFrame:
             f"Found {len(parent_child_dupes)} calcuations where the parent and child "
             f"columns are identical and expected 0.\n{parent_child_dupes=}"
         )
+
+    assert unexpected_total_components(calc_components, dimensions).empty
     # Remove convert_dtypes() once we're writing to the DB using enforce_schema()
     return calc_components.convert_dtypes()
 
 
+def unexpected_total_components(
```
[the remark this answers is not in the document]
jdangerx: Second best time to document your expectations about the data is when you're reverse-engineering what those expectations are!

test/unit/transform/ferc1_test.py
jdangerx: What you would have to do is something like:

1. make a couple dataframes using the basic strategies for value generation (random choice from a predetermined list, random value within a range, etc.)
2. add a foreign key column to one dataframe which pulls values from the primary key column in the referent dataframe

We could maybe make some wrappers around that process to make it easier to use, or if we had some sort of SQLAlchemy <--> Pydantic <--> Pandas inter-op working, maybe we could use a plugin like [hypothesis-sqlalchemy](https://pypi.org/project/hypothesis-sqlalchemy/)

jdangerx: In this scenario, I think we just need to draw stuff from the same possible value sets, e.g. `table_name` is always from "table_1", "table_2", etc. We could restrict table_dimensions to always be a subset of metadata, but :shrug: maybe it's good to know how our code would behave if something shows up in table_dimensions but not metadata.

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
jdangerx: That would be an error, also - I updated the docs / code to hopefully be more clear that we should *never* infer a calculation-component relationship where the child is a total in this function. 

We might still run into those in calculations that cross factoid boundaries, but that's not what we're dealing with here.

src/pudl/transform/ferc1.py
```diff
@@ -5681,10 +5680,57 @@ def calculation_components_xbrl_ferc1(**kwargs) -> pd.DataFrame:
...
+    calc_comps: pd.DataFrame, dimensions: list[str]
+) -> pd.DataFrame:
+    """Find unexpected components in the "total" calculations.
```
jdangerx, line 5692: TODO: make clear that these are about the *inferred* total calculations

src/pudl/transform/ferc1.py
```diff
@@ -5877,133 +5945,61 @@ def add_dimension_total_calculations(
...
-    for dim in dimensions:
-        calc_components = add_has_totals_column_for_dim(
-            calc_components,
-            idx=["table_name_parent", "xbrl_factoid_parent"],
-            dim=f"{dim}_parent",
-        )
-    calc_components = calc_components.assign(
-        total_count_parent=lambda x: x.filter(like="_has_total").sum(axis=1)
+
+    calc_components = pd.concat([calc_components] + total_comps).reset_index(drop=True)
```
jdangerx, line 5981: TODO: restructure code to make clear that we don't actually *use* `calc_components`, we are just smacking a bunch of additional stuff on at the end

zaneselvans:
Okay my best guess as to what's going wrong with the integration test here is that conceptually need to `df.["tags"].fillna({})`.  On the `explode_tree_fixes` branch this happens in `XbrlCalculationForestFerc1.annotated_forest` in this line:

```py
        node_attrs["tags"] = node_attrs["tags"].apply(lambda x: {} if x != x else x)
```

and it works because of the fact that NA, uniquely, does not respect the reflexive property: NA != NA

jdangerx:
Ok, so if I use the new `explode_tree_fixes` branch, and then I also add `weight=1` and `is_within_table_calc=True` to all the inferred relationships, everything works! Albeit I have to tweak some missing data thresholds.

If I just add the weight and within_table_calc, I can uncomment the `set_forest_attributes()` call, and get most of the way - I just run into issues with a few values having multiple conflicting weights of -1 and 1.

I'm tempted to pull in just the "handle multiple component weights" logic from `explode_tree_fixes` so that we can merge this branch into `explode_ferc1` without having to wait for `explode_tree_fixes` to get merged in.

zaneselvans:
If I understand correctly, all of these inferred within-fact totals will **always** have a weight of 1.0, and will **always** be within table calculations, so those values should be set at some point for sure.

There are still some conflicting +1 and -1 weights on the `explode_tree_fixes` branch which are not longer assertion failures, just logged events, so if you want to set the weights and within-table-calc values (which should probably be done programmatically if and only if the parent and child tables are the same) and log rather than assert on the conflicting weights that's probably fine?

What are the missing data thresholds you're talking about? Is there a good reason why they need to be tweaked or is this indicative of a problem?

jdangerx:
Sorry, by missing data threshold I meant `calculation_tolerance` (it went from 0.27 to 0.28). I think it's because we have slightly fewer calculations in the denominator, is all.

PR #2885, opened 2023-09-22
Update expected row counts in 2 altered FERC 1 tables.

zaneselvans:
# PR Overview

There were 2 FERC 1 tables that have changed row counts as a result of the `explode_ferc1` merge.

## `plant_in_service_ferc1`

The difference between the last successful `plant_in_service_ferc1` and the new post-explosion version of the table looks like:
* The new version of the table added a `utility_type` column, which is entirely set to electric (seems correct & expected)
* There are 3220 additional records in the new version of the table.
* None of these records has a record_id
* All of these records are from the DBF era, and the overwhelming majority of them are from 1999-2003. Only 2 of them are from after 2003.
* Only 32 of the new records contain any non-null data, all of them are from 1994-1998
* All of the non-null data are corrections with an `xbrl_factoid` of `electric_plant_in_service_and_completed_construction_not_classified_electric_correction` (aka the combined FERC accounts 101 and 106 fact).
* Of the facts appearing in the new records, the one above with non-null data is the only one that is a correction.
* Seems like these are real corrections and we want to keep them.

## `purchased_power_ferc1`

* The new version of the table is missing 29 records that were in the previous version of the table.
* All of these records are from 2021 (the only year of XBRL data)
* It looks like the issue is we split `purchased_mwh` into `purchased_for_storage_mwh` and `purchased_other_than_storage_mwh` but did not update the parameters in `drop_invalid_rows` to reflect the new column names, so now there are some "invalid" rows getting through.
* We should check that the required valid columns actually exist in the dataframe and throw an error if not to catch this kind of discreapancy.
* I think we may not actually want to drop these records as "invalid".  They look like transactions that were not consummated -- they have valid quantities of MWh purchased, but NA or 0 for the MWh delivered/received, and NA or 0 for `total_settlement` (dollars).
* It looks like in the column renaming we may have accidentally eliminated the `purchased_for_storage_mwh` and `purchased_other_than_storage_mwh` columns from the XBRL data, since those columns don't appear to be getting transformed / summed into the older generic `purchased_mwh` column, and also were not added to the schema for the table, so they just get dropped at the end of the transform.

# PR Checklist


e-belfer:
@zaneselvans I'm surprised the other correction records from other tables didn't break the validations. What are the new null factoids from old DBF data that you're seeing?

zaneselvans:
## `plant_in_service_ferc1`

@e-belfer I could be wrong but I think we were already applying corrections on `dev` using the dictionary rather than tabular version of the calculation fixes, so all of the intra-table corrections should already have been present.

It looks like the "new" null factoids all refer to subtotals or totals and I would guess that they're null just because some utilities don't have anything to report in that whole category?

```
ferc_account_label
nuclear_production_plant                                                                        799
hydraulic_production_plant                                                                      716
other_production_plant                                                                          520
steam_production_plant                                                                          440
distribution_plant                                                                              250
intangible_plant                                                                                189
transmission_plant                                                                              127
general_plant_excluding_other_tangible_property_and_asset_retirement_costs_for_general_plant     71
general_plant                                                                                    68
electric_plant_in_service_and_completed_construction_not_classified_electric_correction          32
electric_plant_in_service                                                                         6
transmission_and_market_operation_plant_regional_transmission_and_market_operation_plant          1
production_plant                                                                                  1
Name: count, dtype: int64
```

## `purchased_power_ferc1`

It seems like what happened in the `purchased_power_ferc1` table is the OLD version we were renaming `purchased_other_than_storage_mwh` to be just `purchased_mwh` matching the old DBF column name, and ultimately discarding the `purchased_for_storage_mwh` column.  Then we split out the storage/not-storage into the two distinct columns that are used in the XBRL data since really it's the sum of them that should be used to create something equivalent to the old `purchased_mwh` column, but we didn't update the table schema to include those new columns, and we didn't update the `drop_invalid_rows` criteria to refer to the new column names, which is why we now see a difference in which record are present.

I still think we may want to be less stringent here about what columns we require to have valid values in them -- there are clearly some unusual transactions where a "purchase" happens, but then doesn't go through -- no power is sent or delivered and no money changes hand.  Which is weird, but seems real.

cmgosnell:
i'm confused why `purchased_power_ferc1` was changed at all bc it is not an explosion table... but this sounds like the columns in the drop invalid rows need to be updated/that check sounds good!

on the `plant_in_service_ferc1`... i'm wondering if these new records are being introduced in `calculate_values_from_components` when there are calculation components but not a parent record.

zaneselvans:
@cmgosnell I think we just happened to notice this error in the `purchased_power_ferc1` and "fixed" it in a half-assed way.

There *is* already a check that all valid columns are present, but we didn't trip it because we weren't attempting to look at the new columns (which are not present)

e-belfer:
Yeah this storage factoid fix originates from us handling the DBF-only factoids I believe: https://github.com/catalyst-cooperative/pudl/pull/2633/commits/9daa77ac88ebdc63dcf0193b4d75a506172b1a35

zaneselvans:
Ahhh yep.  Well, I propose:

* Leaving the `plant_in_service_ferc1` table as it is. (unless one of you think the null records will cause issues?)
* Adding the `purchased_for_storage_mwh` and `purchased_other_than_storage_mwh` columns to the `purchased_power_ferc1` table schema and adding them to the list of `required_valid_columns` (@cmgosnell do you remember how `drop_invalid_rows` works? It'll keep any row that has a valid value in AT LEAST one of these columns, right?)

e-belfer:
Looks like `drop_invalid_rows` drops rows where invalid values are in all specified columns. This plan of action looks good to me with respect to the fuel table, I'm wondering whether we want to just drop the created null subcomponents but maybe this will cause problems with our calculations and it doesn't seem invalid to keep them in the table.

zaneselvans:
Adding in the storage / not-storage columns, we get exactly the same set of records in the pre- and post-explosion outputs (albeit with 2 additional columns now, which contain a bunch of information).

In case anybody is curious this is all I'm doing to check the difference between the old and new data:

```py
table_name = "purchased_power_ferc1"
#table_name = "plant_in_service_ferc1"

pk_cols = {
    "purchased_power_ferc1": ["record_id"],
    "plant_in_service_ferc1":  Resource.from_id(table_name).schema.primary_key,
}

# Use the real paths to your 2 DBs obvs:
left_db_url = "sqlite:///" + "/Users/zane/code/catalyst/pudl-work/output/pudl.sqlite"
left_engine = sa.create_engine(left_db_url)
left_df = pd.read_sql(f"denorm_{table_name}", left_engine)
left_df = left_df.set_index(pk_cols[table_name])

# Use the real paths to your 2 DBs obvs:
right_db_url = "sqlite:///" + "/Users/zane/code/catalyst/pudl-work/output/pudl-dev-2023-09-21.sqlite"
right_engine = sa.create_engine(right_db_url)
right_df = pd.read_sql(f"denorm_{table_name}", right_engine)
right_df = right_df.set_index(pk_cols[table_name])

left_only_index = left_df.index.difference(right_df.index)
right_only_index = right_df.index.difference(left_df.index)

left_only_df = left_df.loc[left_only_index]
right_only_df = right_df.loc[right_only_index]
```

src/pudl/metadata/fields.py
```diff
@@ -2059,7 +2059,17 @@
     },
     "purchased_mwh": {
         "type": "number",
-        "description": "Megawatt-hours shown on bills rendered to the respondent.",
+        "description": "Megawatt-hours shown on bills rendered to the respondent. Includes both electricity purchased for storage and non-storage purposes, which were lumped together prior to 2021.",
+        "unit": "MWh",
+    },
+    "purchased_other_than_storage_mwh": {
+        "type": "number",
+        "description": "Number of megawatt hours purchased for other than energy storage during the period.",
```
e-belfer, line 2067: "Number of megawatt hours purchased for non-storage purposes during the period" would be clearer here.

e-belfer (approved):
One minor suggestion on the field description, otherwise looks good to go.

src/pudl/metadata/fields.py
zaneselvans: I cut-and-pasted these from the FERC Taxonomy but I agree they read weird.

PR #2886, opened 2023-09-23
standardize the calc checks for the total to subtotal calcs

cmgosnell:
# PR Overview
I don't love this overall... but it is a first attempt to standardize the way we are checking the total to subtotal calcs. I:
- added a `is_total_to_subdimensions_calc` bool column back into the calc components table
- in both the transform step (in `reconcile_table_calculations`) and in the explode step (in `reconcile_intertable_calculations`) we now use `calculate_values_from_components` and `check_calculation_metrics` twice: once for the "normal" calcs and once for the total-to-subdimension calcs.

## Notes:
- this did not fix the high error rate on the balance sheet asset explosion.
- I still think it feels nicer overall bc is removes almost duplicate code for the old version of this check.
- I think it could be a cleaner if we thought through a way to just have these two calc-ing/metrics-ing functions deal with the logic of what is a total-to-sub-dim calc instead of calling them twice. I tried adding the `is_total_to_subdimensions_calc` into the pks of the groupby and ran into a slight snag.. i still think its possible but i wanted to get a proof out before leaving.
- Because this didn't solve the asset explosions' problem... my suspicion is this is because this `electric_plant_depreciation_change_ferc1` has... ALLL OF THE DIMENSIONS. and perhaps we are introducing some double counting into the calc components because of it? not sure at all just a hunch.

# PR Checklist


src/pudl/output/ferc1.py
```diff
@@ -1477,6 +1485,22 @@ def generate_intertable_calculations(
...
+            calc_idx=calc_idx,
+            value_col=self.value_col,
+        )
+        subtotal_calcs = pudl.transform.ferc1.check_calculation_metrics(
+            calculated_df=subtotal_calcs,
+            value_col=self.value_col,
+            calculation_tolerance=self.calculation_tolerance.intertable_calculation_errors,
+            table_name=self.root_table,
+            add_corrections=True,
+        )
```
cmgosnell, line 1503: there are no total to sub-dimension inter-table calcs rn but this felt like a good thing to run nonetheless

zaneselvans (approved):
Definitely an improvement!

PR #2900, opened 2023-09-27
Use sphinx-autoapi 3.0 to work with astroid 3.0

zaneselvans:
# PR Overview

Bump the version of `sphinx-autoapi` and remove astroid pin that is hopefully no longer available.

Also test our environment solving in CI... to see if the problems I'm having locally are platform dependent.

# PR Checklist

- [ ] Merge the most recent version of the branch you are merging into (probably `dev`).
- [ ] All CI checks are passing. [Run tests locally to debug failures](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#running-tests-with-tox)
- [ ] Make sure you've included good docstrings.
- [ ] For major data coverage & analysis changes, [run data validation tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#data-validation)
- [ ] Include unit tests for new functions and classes.
- [ ] Defensive data quality/sanity checks in analyses & data processing functions.
- [ ] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/latest/release_notes.html) and reference reference the PR and related issues.
- [ ] Do your own explanatory review of the PR to help the reviewer understand what's going on and identify issues preemptively.

Issue #2896, opened 2023-09-27
Use lockfile to specify a reproducible python environment


zaneselvans:
We've had a spate of random breakage due to downstream dependencies so maybe it is finally time to figure out lockfiles. See #2140 for some prior discussion.

Since we are treating PUDL like an application and not a library, we just need **one** set of dependencies & their versions which is guaranteed to work, not an expansive range of every dependency. See #1669 for some rationale and discussion.

There are several possibilities with various pros and cons. Earlier this year @lwasser wrote [this post](https://www.pyopensci.org/blog/demystifying-python-packaging.html) about putting together the @pyOpenSci  [Python Package Guide](https://www.pyopensci.org/python-package-guide) which might be helpful.

## Considerations:
* Do we want `catalystcoop.pudl` to continue being distributed as a package, or will it **only** be installable from the git repository?
* Does it matter how the downloaded package are cached, e.g. for building Docker containers or on GitHub runners.
* How easily can we automatically update our dependency versions, regenerate the lockfile, and see if the tests still pass? Will it be compatible with the dependabot?
* How will we manage non-Python package dependencies like `pandoc`, `nodejs`, `sqlite`, `libsnappy`?  Conda and Docker seem to be the most common solutions here.
* How will we manage the version & installation of Python itself?

### Pros
* Conda can install and manage versions of non-python dependencies.
* Conda can manage and install various versions of Python itself.
* Allows locking of platform-specific binary distributions including hashes.
* Conda ensures shared compiled libraries have compatible versions too, which is nice for the geospatial stack.
* The `mamba` solver is extremely fast and reliable.
* We're already using `conda`/`mamba` to manage our environments.
* Dependabot is happy to update project dependencies & optional dependencies in `pyproject.toml`
* `conda-lock` and dependabot both work with Poetry, configured in `pyproject.toml`
* You can specify that particular packages that aren't available on `conda-forge` be installed from PyPI.

### Cons
* Dependabot doesn't cover `environment.yml` files, so we'd like all to read all dependencies in `pyproject.toml`.
* Doesn't appear to pick up on `package[extras]` specified in `pyproject.toml`
[...]

PR #2901, opened 2023-09-28
Set up reproducible Python environments with conda-lock

zaneselvans:
# PR Overview

This is the `conda-lock` specific stuff from #2900. See #2896 for some more general discussion of using lockfiles to make our environments reproducible.

## Draft `conda-lock` setup for conversation...

* Added configuration of `conda-lock` in `pyproject.toml` -- it still gets the newest versions of `recordlinkage` and `sphinx-autoapi` from PyPI.
* See `conda-lock.yml` as an example of the lockfiles it generates.

To create a new lockfile:
```bash
conda-lock --mamba -f pyproject.toml
```

To install all the packages specified in that lockfile in an environment named `pudl-dev`:
```bash
conda-lock install --mamba -n pudl-dev --dev -E doc -E datasette
```
* The `--dev` installs the somewhat standardized `test` and `dev` package extras.
* Note that this requires having `conda-lock` installed in your base conda environment.

To get into the environment and work with PUDL:
```bash
mamba activate pudl-dev
pip install -e ./
```

You should see that absolutely every dependency has already been satisfied, and the only new package that gets installed is `catalystcoop.pudl` itself.

# PR Checklist


pyproject.toml
```diff
@@ -276,3 +277,23 @@ doctest_optionflags = [
...
+channel-priority = "strict"
+name = "pudl-dev"
+
+[tool.conda-lock.dependencies]
+recordlinkage = {source = "pypi"}
+sphinx-autoapi = {source = "pypi"}
+pip = ">=23,<24"
+python = ">=3.11,<3.12"
+python-snappy = ">=0.6,<1"
+sqlite = ">=3.36,<4"
```
zaneselvans, line 299: This is all the configuration for `conda-lock`

With this `pyproject.toml` all of our dependencies except for `recordlinkage` and `sphinx-autoapi` are installed from `conda-forge` (and those two packages alone are singled out for installation from PyPI).

Our programmatic non-python dependencies (snappy, sqlite..) also get installed from `conda-forge`.

We could add `jupyterlab` and the other interactive dependencies from `devtools/environment.yml` into the `dev` optional dependencies above, and `pandoc` / `nodejs` here if we wanted, and eliminate `devtools/environment.yml` and probably our other `*-environment.yml` files.

Would need to add some infrastructure for updating the lockfile on a regular basis and get used to using it to install our environment, but it does seem to be working.

pyproject.toml
```diff
@@ -276,3 +277,23 @@ doctest_optionflags = [
...
+]
+channel-priority = "strict"
+name = "pudl-dev"
+
+[tool.conda-lock.dependencies]
+recordlinkage = {source = "pypi"}
+sphinx-autoapi = {source = "pypi"}
+pip = ">=23,<24"
+python = ">=3.11,<3.12"
+python-snappy = ">=0.6,<1"
```
jdangerx, line 298: Why do we need to explicitly include `python-snappy` here instead of in our normal dependencies?

pyproject.toml
zaneselvans: Yes, that's right -- we need a version of the package that can only be sourced from PyPII.  I've requested to be made a maintainer on one, and someone from `conda-forge` has commented on the potential abandonment of `sphinx-autoapi` in the most recent bot PR, so hopefully they'll both be available from conda.  But at the same time it's good to know that in the event there's some package we can really only get from PyPI, there's a way to do it without breakage.

zaneselvans: And yes it would be really great to have `pyproject.toml` be the ONLY source file, with dependencies broken out into groups / extras that can be selectively installed from the lockfile.  I didn't figure it out but I suspect there's a way to categorize the conda-only dependencies so that they can also be associated with a particular grouping, so we could have `jupyterlab`, `pandoc, `nodejs` etc. be part of the `dev` or a `jupyter` group -- they'd be invisible to `pip` (which couldn't satisfy many of them) but available in the lockfile for installation with `mamba`

pyproject.toml
zaneselvans: Hmm. Here I just pulled over the packages that we needed to have in `environment.yml` because they had non-python package components that we needed to satisfy.  But I think `python-snappy` could go into the main dependencies here. I think it would mean that `libsnappydev` would need to be installed on the system if the package were getting installed with `pip` rather than `conda` via the lockfile, but that's already true.

zaneselvans:
Oh I hadn't seen `conda-lock render` that's great to know!

For Tox we could use `tox-conda` and let it use the environment generated by the lockfile / rendered environment files... or I think we could just let it keep using `pip` to install the package based on `pyproject.toml` couldn't we?  If it's running inside a `conda` environment that's satisfying the binary dependencies like `libsnappydev` then in theory the `pip` installed virtual environment should work fine, and would test whether it can actually be solved / installed using `pip`.  But it does seem like it would be less duplicative / more reproducible to use the locked conda versions for tox.

zaneselvans:
I do think we should think a bit about what kind of installation scenarios and environmental isolation we care about.

If we're treating PUDL as an application, and only expect it to be reliably installable in a conda environment corresponding to the lockfile, then what do we think about publishing tagged releases to `PyPI` and `conda-forge`?  And what do we think about using Tox to create an isolated environment into which a built package is installed, separate from the code that's stored in the repository?

One could imagine a world in which we assume PUDL will only ever be run from a checked-out repository, and only ever run in a conda environment built from a lockfile, which would imply a different set of things to test.

PR #2903, opened 2023-09-29
Fix for harvesting owner utilities

katie-lamb:
# PR Overview

After I added harvesting of owner utilities a bug arose from harvesting owner utilities as "operator utilities" for the plant and generator entities and annual tables. 

Previously `_core_eia860__ownership` had `utility_id_eia` referring to operator utility. This utility ID would get harvested for generators and plants.

Now `_core_eia860__ownership` has `utility_id_eia` referring to owner utility and `operator_utility_id_eia` referring to the operator utility. The issue was that the owner utilities shouldn't be added to the generator and plant tables, where `utility_id_eia` refers to the operator.

The least quickest, least janky fix I could come up with for this is to skip the harvesting of the ownership table altogether. Ideally, we'd use `map_cols_dict` (I added this in the PPE PR) to grab operator utilities from the ownership table to add to the generator and plant tables, but this required dropping the owner utility columns from the ownership table. Ultimately, this seemed less foolproof and more janky in the short term than just skipping the ownership table.

Ultimately, I think we might want to rethink this harvesting process a little. It's already sort of weird that we have to harvest plants before utilities, so that we can subsequently drop some columns from the plants tables and harvest the utility columns that share names with the dropped plants columns. I'll highlight where this happens in the harvesting process. I'd actually like to do this for the current situation, but in reverse, where we harvest utilities first, drop owner columns, then harvest plants. But both can't happen in the current setup.

I still need change the validation row counts back.
 

# PR Checklist

### DIFF
--- a/src/pudl/transform/eia.py
+++ b/src/pudl/transform/eia.py
@@ -318,6 +318,12 @@ def _compile_all_entity_records(
             logger.debug(f"        {table_name}...")
             # create a copy of the df to muck with
             df = transformed_df.copy()
+            # TODO: take care of this elewhere, make less janky
+            # don't harvest owner utilities for plant and generator annual operator utility
+            if ((entity == EiaEntity.PLANTS) or (entity == EiaEntity.GENERATORS)) and (
+                table_name == "_core_eia860__ownership"
+            ):
+                continue
             # we know these columns must be in the dfs
             cols = []
             # check whether the columns are in the specific table


cmgosnell:
this is probably oos for this pr but now that every single table gets run through `enforce_schema`, it might be worthwhile to think about not dropping the columns in the harvesting process at all (the dropping was done originally in the way way pre-historic past long before we (which is to say i) knew what we were doing).

I thiiiink if we added some suffixes when columns with similar data reference different entities (like `zip_code_plant` and `zip_code_utility`) plus didn't drop any columns, the ordering of the harvesting process wouldn't matter.

zaneselvans:
> the owner utilities shouldn't be harvested for the generator and plant tables, where utility_id_eia refers to the operator.

What does it mean for the owner utilities to be harvested **for** the generator and plant tables?

Why is there a need to differentiate between the owner & operator utility IDs in any context outside of the ownership table (where there are two utility ID columns, referring to utilities that play different roles with respect to the generator in that record). Shouldn't all the utility IDs end up in the utility table, regardless of whether they were observed as owners or operators?

I thought we had already stopped dropping columns as they were harvested -- we call `enforce_schema()` at the end of `harvest_entity_tables()` and each of the harvesting assets gets its own copy of the inputs, so we don't have to worry about whether they are altered in the process -- there's no impact on any of the other harvesting processes. Are columns being dropped in some other part of the harvesting process?

> grab operator utilities from the ownership table to add to the generator and plant tables

In what sense are these utilities being added to the plant & generator tables?  Shouldn't they only appear in those tables if they're actually listed as the operator of the plant/generator?

If we skip the ownership table entirely, we'll miss collecting any owner-only utility IDs, as well as any operator utility IDs that only show up in this table (though hopefully there are zero of these because they've already been found in the plant or generator tables).

Would it be straightforward to just drop the `operator_utility_id_eia` at the beginning and make sure we harvest the owner utility ID, since that's the information we expect to be novel in this table?

Also, without capturing the owner utility IDs, we're going to go back to having FK constraint violations, and need to disable a bunch of FK relationships again.

zaneselvans:
* In `denorm_by_plant()` (and gen, and boiler, etc) we're currently **silently** dropping data with dropna() if there's an NA in one of the ID columns -- this is how we ended up losing data all over the place e.g. in FRC table.  Need a more robust check.
* `map_cols_dict` is currently creating two columns of the same name -- both `utility_id_eia` -- which is probably fucking things up.  We could either swap them in `map_cols_dict` **or** just rename `owner_utility_id_eia` to `utility_id_eia` in the original column maps for the spreadsheets, and make `operator_utility_id_eia` that has a funny name and doesn't get harvested or show up anywhere.
* Alternatively, the most correct solution -- that would harvest both sets of utility IDs and use the existing infrastructure -- would be to run two copies of the ownership table through the harvesting process as a special case -- it could be added to the special case we already have for `UTILITIES`.  Where we harvest the reported attributes in association with the ownership ID, and then remove all attributes and only include the ID columns when harvesting an additional copy based on the operator IDs. In this case it would still make sense to rename the original columns in the mapping spreadsheets.

katie-lamb:
> Alternatively, the most correct solution -- that would harvest both sets of utility IDs and use the existing infrastructure -- would be to run two copies of the ownership table through the harvesting process as a special case -- it could be added to the special case we already have for UTILITIES. Where we harvest the reported attributes in association with the ownership ID, and then remove all attributes and only include the ID columns when harvesting an additional copy based on the operator IDs. In this case it would still make sense to rename the original columns in the mapping spreadsheets.

Running two copies of the ownership table through the harvesting process is what `map_cols_dict` is supposed to do. It only renames `utility_id_eia` etc. on a copy of the dataframe, so it doesn't fuck things up.

The issue comes when you're harvesting the plant (or generator) entity and use the `plant_id_eia`, `utility_id_eia` (owner utility) relationship from the ownership table, which adds a bunch of inconsistency and subsequently leads to null values that are dropped in that `dropna` in `denorm_plants_eia`.

katie-lamb:
Another idea: leave the `_core_eia860__ownership` table with columns names as is in the raw data (`utility_id_eia`) is the operator utility, `owner_utility_id_eia` is the owner utility. Then harvest as usual, using `map_cols_dict` to rename the owner utility for the utility entity (generator and plant will be harvested as they previously were). In the normalized ownership and denormalized ownership table, rename the operator and owner columns.

zaneselvans:
Ahhhh, okay okay, that makes sense where the NA values were coming from then.

src/pudl/metadata/resources/__init__.py
```diff
@@ -289,8 +295,15 @@
...
-            "operator_utility_id_eia": "utility_id_eia",
-            "operator_name": "utility_name_eia",
+            "owner_utility_id_eia": "utility_id_eia",
+            "owner_name": "utility_name_eia",
+            "owner_state": "state",
+            "owner_country": "country",
+            "owner_street_address": "street_address",
+            "owner_zip_code": "zip_code",
+            "owner_city": "city"
+            # currently can't harvest operator utilities in the same dict
```
katie-lamb, line 305: Ideally you'd be able to harvest the operator utilities from the ownership table as well, but as this currently functions, `map_cols_dict` would rename both the `operator_utility_id_eia` and `owner_utility_id_eia` columns to `utility_id_eia` and have colliding column names. Maybe it's fine to just not harvest the operator utilities from the ownership table.

src/pudl/metadata/resources/eia860.py
```diff
@@ -290,13 +290,13 @@
                 "plant_id_eia",
                 "plant_id_pudl",
                 "plant_name_eia",
-                "utility_id_eia",
-                "utility_id_pudl",
-                "utility_name_eia",
+                "owner_utility_id_eia",
+                # "utility_id_pudl",
```
katie-lamb, line 294: See comment in the transform step for this table about utility_id_pudl

src/pudl/output/eia860.py
```diff
@@ -29,21 +29,23 @@ def denorm_ownership_eia860(
             "plant_id_eia",
             "plant_id_pudl",
             "plant_name_eia",
-            "utility_id_pudl",
+            # "utility_id_pudl",
```
katie-lamb, line 32: I think that bringing in `utilty_id_pudl` from the `plants_utililities_eia` table will bring in the operator `utility_id_pudl` for that plant, not the owner utility PUDL ID. Maybe grab the `utilities_eia` table as well to bring in the correct `utility_id_pudl`?

test/validate/eia_test.py
```diff
@@ -45,16 +45,16 @@ def test_no_null_cols_eia(pudl_out_eia, live_dbs, cols, df_name):
 @pytest.mark.parametrize(
     "df_name,raw_rows,monthly_rows,annual_rows",
     [
-        ("bf_eia923", 1_473_364, 1_473_364, 123_369),
+        ("bf_eia923", 1_521_340, 1_521_340, 127_415),
         ("bga_eia860", 141_638, 141_638, 141_638),
         ("boil_eia860", 80_059, 80_059, 80_059),
-        ("frc_eia923", 584_724, 241_751, 23_737),
-        ("gen_eia923", None, 4_765_282, 398_747),
+        ("frc_eia923", 632_035, 258_151, 25_373),
```
katie-lamb, line 51: These row counts should likely be a little higher than they were previous to merging the PPE because we're also harvesting owner utilities now and there are fewer null values in `utility_id_eia`.

katie-lamb:
@zaneselvans I think this is close to working if the foreign key rules and alembic stuff can be resolved. I tried to leave some comments about what I was thinking and any potential messiness.

katie-lamb:
Still need to debug a key error and update row counts.

katie-lamb:
Working and ready for review!

test/validate/eia_test.py
```diff
@@ -45,16 +45,16 @@ def test_no_null_cols_eia(pudl_out_eia, live_dbs, cols, df_name):
...
-        ("gen_eia923", None, 4_765_282, 398_747),
+        ("frc_eia923", 631_975, 258_115, 25_370),
+        ("gen_eia923", None, 5_178_892, 433_286),
         ("gens_eia860", 556_578, 556_578, 556_578),
-        ("gf_eia923", 2_598_272, 2_598_272, 222_356),
+        ("gf_eia923", 2_859_269, 2_859_269, 244_796),
         ("own_eia860", 89_687, 89_687, 89_687),
         ("plants_eia860", 200_177, 200_177, 200_177),
-        ("pu_eia860", 180_342, 180_342, 180_342),
+        ("pu_eia860", 199_332, 199_332, 199_332),
```
zaneselvans, line 57: Can you look up the expected number of rows that were being used on `dev` prior to the previous PR getting merged, so we can compare directly between the new expectations, and the expectations we had prior to the changes that resulted in the loss of all these rows? That will let us evaluate the net impacts of the ownership harvesting changes and see if there's anything suspicious that should still be investigated / compared with the last good `dev` output.  Here and in the `mcoe_test` outputs below.

src/pudl/transform/eia860.py
```diff
@@ -169,10 +169,7 @@ def _core_eia860__ownership(raw_eia860__ownership: pd.DataFrame) -> pd.DataFrame
             "utility_id_eia": "operator_utility_id_eia",
             "utility_name_eia": "operator_name",
             "state": "operator_state",
-            "owner_name": "utility_name_eia",
-            "owner_utility_id_eia": "utility_id_eia",
-            "owner_state": "state",
-            "owner_country": "country",
+            "owner_name": "owner_utility_name_eia",
```
zaneselvans, line 172: Is there a reason why we need or want for these renames to happen in the transform, rather than ensuring that the columns have the desired name in the initial spreadsheet mapping CSVs?

src/pudl/metadata/resources/__init__.py
zaneselvans: @cmgosnell did we still want to try adding a stripped down 2nd version of the ownership table with just the operator utility IDs into the harvesting for utilities?

zaneselvans (changes requested):
Mostly nits, but two things I think we should definitely do:
* If we're **not** going to harvest the operator IDs from the ownership table, make sure we're run the foreign key checks on a full DB locally before merging into `dev`
* Do a side-by-side comparison of the new expected row counts, and the expected row-counts from before the merge into `dev` that caused the records to get lost.
* We need to remove all of those `dropna()` calls in the `denorm_` assets and replace them with a more subtle and vocal check that complains if there are unacceptable NA values in the required columns.

test/validate/eia_test.py
katie-lamb: Yep, should have said this earlier but these row counts as well as MCOE are the same as in `dev` except for the `utils_eia860` table, which has all the extra owner utility rows.

src/pudl/transform/eia.py
```diff
@@ -312,45 +312,36 @@ def _compile_all_entity_records(
...
-                if column in df.columns:
-                    cols.append(column)
-            df = df[(base_cols + cols)]
-            df = df.dropna(subset=id_cols)
-            # add a column with the table name so we know its origin
-            df["table"] = table_name
-            dfs.append(df)
-            # check if there are any columns that should be renamed and harvested
+        if "annual" not in table_name:
+            if set(base_cols).issubset(transformed_df.columns):
```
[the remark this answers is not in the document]
katie-lamb: Previously, the `if map_cols_dict and ...` logic was nested inside this `if "annual" not in table_name and set(base_cols).issubset...` statement. However, in the case of the utility entity, the base columns are `report_date` and `utility_id_eia`, which aren't in the ownership table because the columns have been renamed to `operator_utility_id_eia` and `owner_utility_id_eia`. This meant that the ownership table wasn't making it past the first if statement (because the base cols weren't present) and no records were being harvested for the utility entity. So instead, I had to make it so we first check if the normal base columns are present, and then also check if the map_cols_dict maps to base column names in a separate statement.

I could make this `if "annual" not in table_name and base_cols.issubset ...` and then `if "annual" not in table_name and map_cols_dict...`. Not sure what is more conventional.

src/pudl/transform/eia860.py
katie-lamb: Not sure where the initial spreadsheet mapping happens, but you mean so that `_raw_eia860__ownership` also has these renamed columns?

katie-lamb: No preference from my end.

src/pudl/metadata/resources/__init__.py
katie-lamb: How do I run just the foreign key checks on the full DB? I thought they were part of the CI because I had one fail for `plant_parts_eia` in GitHub.

katie-lamb: Nevermind figured it out

katie-lamb: I haven't generated the FERC side but the `glue` tables and EIA tables passed the foreign key checks.

src/pudl/transform/eia.py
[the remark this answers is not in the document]
cmgosnell: is the `if "annual"` even required anymore? i think it was when we were passing around and editing the same  dict of dfs but if i understand correctly what is currently happening, "annual" should never be in the table name bc that is only in the table names of the harvested annual entity tables which should no long ever show up in `clean_dfs`

src/pudl/transform/eia.py
```diff
@@ -312,45 +312,36 @@ def _compile_all_entity_records(
...
+                if base_cols_to_add.union(set(map_cols_dict.keys())).issubset(
+                    transformed_df.columns
+                ):
+                    mapped_df = transformed_df[
+                        list(base_cols_to_add) + list(map_cols_dict.keys())
+                    ]
+                    mapped_df = mapped_df.rename(columns=map_cols_dict)
+                    mapped_df = mapped_df.dropna(subset=id_cols)
+                    mapped_df["table"] = table_name + "_mapped"
+                    dfs.append(mapped_df)
```
cmgosnell, line 344: i think i still see this as adding two sets of columns when there are `owner` and `operator` columns in one table. or at least it does not seem like it is going to be able to pick up both id's from one table.

hm okay nvm so the ownership table now has `owner` and `operator` prefixes and the `owner` prefixed columns are the ones getting renamed and added. the ownership table skips over the entire `if set(base_cols).issubset(transformed_df.columns):` section bc `base_cols` includes `utility_id_eia` and that is actually not yet in the table.

cmgosnell: i think it would still generally be preferable if we could grab both sets of ids from this ownership table for the harvesting process. although if there are truly no new operator ids' from this table i guess it is fine to basically ignore the operator column from the ownership table.

src/pudl/metadata/resources/__init__.py
cmgosnell: i think if you wanted to harvest both `map_cols_dict` would need to be a list of map dicts that you'd add both renamed versions into `dfs` -> `compiled_df`

src/pudl/transform/eia860.py
cmgosnell: i think it would be at least marginally nicer to rename these in the mapping CSVs. it is in `package_data/eia860/column_maps/ownership.csv`

src/pudl/metadata/resources/eia860.py
```diff
@@ -240,6 +240,8 @@
                     "heat_rate_by_generator_monthly",
                     "mcoe_monthly",
                     "mcoe_generators_monthly",
+                    "ownership_eia860",
+                    "denorm_ownership_eia860",
```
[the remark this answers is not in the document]
cmgosnell: ha i just asked the same thing for the plant fk relationship. it doesn't seem like it should be necessary anymore.

src/pudl/transform/eia.py
katie-lamb: > the ownership table skips over the entire `if set(base_cols).issubset(transformed_df.columns):` section bc `base_cols` includes `utility_id_eia` and that is actually not yet in the table.

Yep exactly.



> i think it would still generally be preferable if we could grab both sets of ids from this ownership table for the harvesting process. although if there are truly no new operator ids' from this table i guess it is fine to basically ignore the operator column from the ownership table.

About to push a change that should do this.

src/pudl/metadata/resources/eia860.py
```diff
@@ -391,6 +393,8 @@
                     "capacity_factor_by_generator_monthly",
                     "mcoe_monthly",
                     "mcoe_generators_monthly",
+                    "ownership_eia860",
+                    "denorm_ownership_eia860",
```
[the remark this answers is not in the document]
katie-lamb: Yep, it's only annual. I thought this was what we were talking about after our latest call, but maybe I misinterpreted. I can take this out.

test/validate/eia_test.py
zaneselvans: Oh great, good to know!

src/pudl/metadata/resources/eia860.py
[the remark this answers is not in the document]
zaneselvans: No, it shouldn't need to be excluded any more, but I think we do need to **add** `owner_utility_id_eia` as one of the columns which is used to construct the automatic FK relationships. This would be done in the FK rules section of the utility entity and annual utility tables.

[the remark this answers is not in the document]
zaneselvans: See this section in the `utilities_eia860` resource definition below:
```py
            "foreign_key_rules": {
                "fields": [
                    ["utility_id_eia", "report_date"],
                    # Failing because this column is not harvested in the old
                    # system. TODO: re-enable when we switch to new system.
                    # https://github.com/catalyst-cooperative/pudl/issues/1196
                    # ["owner_utility_id_eia", "report_date"],
```

src/pudl/transform/eia.py
```diff
@@ -312,45 +312,36 @@ def _compile_all_entity_records(
...
-                base_cols_to_add = set(base_cols) - set(map_cols_present_dict.values())
-                mapped_df = mapped_df[
-                    list(base_cols_to_add) + list(map_cols_present_dict.keys())
-                ]
-                logger.debug(f"mapped_df columns: {mapped_df.columns}")
-                mapped_df = mapped_df.rename(columns=map_cols_present_dict)
-                mapped_df = mapped_df.dropna(subset=id_cols)
-                mapped_df["table"] = table_name + "_mapped"
-                dfs.append(mapped_df)
+            if map_cols_dict:
```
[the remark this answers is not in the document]
katie-lamb: Updated the docstring in the ENTITIES constant.

cmgosnell (approved):
this looks great!! i am slightly tempted to suggest that we figure out a way to run the renamed df through as much of the same part of the ifs in there bc there is some duplication in there but it is minimal and I'm excited to get this in!

also, you didn't change the row counts again, which makes me assume the harvesting of both of these id's didn't result in any new util ids (which makes sense to me tbh). but just checking that you tested the row counts!

re: @zaneselvans 's
> We need to remove all of those dropna() calls in the denorm_ assets and replace them with a more subtle and vocal check that complains if there are unacceptable NA values in the required columns.

can we spin this off into a new smaller pr? mostly so we can get the builds to pass again bc this is holding up other merges.

zaneselvans:
@cmgosnell I agree we should spin the `dropna()` off but we should make create an issue and PR for that right now so we don't forget about it and run into this again...

zaneselvans (approved):
Not sure why the Read The Docs build failed. I am re-running it now.

katie-lamb:
I created an issue about the dropna calls (#2919 ) and am going to merge this!

PR #2915, opened 2023-10-03
Improve calculation error checking

zaneselvans:
# PR Overview

* Split the calculation checks into 3 steps: applying the calculations, checking the calculations, and adding correction records to the data.
* Switch to using a `CalculationTolerance` object to pass around a standardized set of expected error levels.
* Add a collection of error checking functions that can run on whole dataframes or via `GroupBy.apply()` and some helper functions that use them to calculate a matrix of different error metrics across different groupings, to be run in `check_calculation_metrics()`

# Review Questions
- We're using `np.isclose()` to determine whether reported & calculated values match. Depending on the scale of the values and the values of `rtol` and `atol` this means we may have some values that aren't exactly the same, but that still count as "matching" Do we want to correct all values even if `isclose()` says they're the same?  Right now we're using the default `atol=1e-5` which I think will only ever catch floating point math differences, rather than values that are off by say $1.00, or $0.001. Is that the intention?
- Should the `CalculatonTolerance` and `ReconcileTableCalculations` classes be consolidated into a single parameter? Seems like the `ReconcileTableCalculations` class contains parameters that only really apply to the intra-table calculation case.
- Note that the across-dimension calculation checking should be refactored to use these parameters too. See #2688 and #2886  
- `reconcile_table_calculations()` has a bunch of prep work happening before it gets to the part where it does the calculations. It might be better if this was made into its own function that can be run in a modular way. Similarly the part after it checks & corrects the intra-table calculations, dealing with the dimension-to-total calculations, but I think @cmgosnell is already working on that.

# Error Exploration
- There many cases where the reported value we're looking at is non-null, but the calculated value is NA. This seems like the more common arrangement than having a value for both! In `balance_sheet_assets` about 85% of records in 1994-2004 and 2021 have a non-null reported value but a null calculated value.
- Even weirder, in `balance_sheet_assets` for 2005-2020, 100% of the reported versions of the calculated values are showing up as NA in the calculation checks, but there are still a bunch of non-null corresponding calculated values.
- There are also 5 years where the number of null calculated values jumps up dramatically from 5-6k to 12-13k.
- There are instances where a single calculation has an error that is thousands of times larger than the reported value, which can have a significant impact on overall aggregations, even to the point of being several percent of the value reported by all utilities in a year.
  - `balance_sheet_liabilities` where (`utility_id_ferc1=165`, `report_year=1995`): error is $10B, which is 2.5% of all value reported by all utilities in that year.
  - `balance_sheet_assets` where (`utility_id_ferc1=172`, `report_year=1996`): error is 83% of reported value.
  - `balance_sheet_assets` where (`utility_id_ferc1=292`, `report_year=2004`):  error is 32% of reported value.

# PR Checklist


src/pudl/transform/ferc1.py
```diff
@@ -726,6 +727,116 @@ def combine_axis_columns_xbrl(
...
+class CalculationMetricTolerance(TransformParams):
+    """Tolerances for all data checks to be preformed within a grouped df."""
+
+    error_frequency: confloat(ge=0.0, le=1.0) = 0.01
+    relative_error_magnitude: confloat(ge=0.0) = 0.04
+    absolute_error_magnitude: confloat(ge=0.0) = np.inf
+    null_calculation_frequency: confloat(ge=0.0, le=1.0) = 0.7
+    """Fraction of records with non-null reported values and null calculated values."""
+    null_reported_value_frequency: confloat(ge=0.0, le=1.0) = 1.0
+    # ooof this one is just bad
```
cmgosnell, line 759: suggestion is to remove these default bc storing the defaults here and in `CalculationGroupChecks` and in the table params is too many place

src/pudl/transform/ferc1.py
```diff
@@ -960,75 +1124,332 @@ def calculate_values_from_components(
...
+class AbsoluteErrorMagnitude(GroupMetricCheck):
+    """Check relative magnitude of errors in XBRL calculations.
+
+    These numbers may vary wildly from table to table so no default values for the
+    expected errors are provided here...
+    """
+
+    def metric(self: Self, gb: DataFrameGroupBy) -> pd.Series:
+        """Calculate the absolute mangnitude of XBRL calculation errors."""
+        return gb.abs_diff.sum()
```
zaneselvans, line 1366: As above, I think there's a similar sign issue in here in that any positive and negative values within `abs_diff` would act to cancel each other out, so I suspect this should be

```py
return gb.abs_diff.abs().sum()
```

src/pudl/transform/ferc1.py
```diff
@@ -960,75 +1124,332 @@ def calculate_values_from_components(
...
+        return df.set_index(["group", "table_name", "group_value"])
+
+
+class ErrorFrequency(GroupMetricCheck):
+    """Check error frequency in XBRL calculations."""
+
+    def metric(self: Self, gb: DataFrameGroupBy) -> pd.Series:
+        """Calculate the frequency with which records are tagged as errors."""
+        try:
+            out = gb[gb.is_not_close].shape[0] / gb.shape[0]
```
zaneselvans, line 1334: I used the `shape` attributes here rather than `len()` because it doesn't require any computation to look up the shape, and we're doing this potentially a bazillion times, inside every group.

src/pudl/transform/ferc1.py
```diff
@@ -960,75 +1124,332 @@ def calculate_values_from_components(
...
+        """Calculate the frequency with which records are tagged as errors."""
+        df["is_not_close"] = self.is_not_close(df)
+        return df.groupby(by=self.groupby_by()).apply(self.metric)
+
+    def snake_case_metric_name(self: Self) -> str:
+        """Convert the TitleCase class name to a snake_case string."""
+        class_name = self.__class__.__name__
+        return re.sub("(?!^)([A-Z]+)", r"_\1", class_name).lower()
+
+    def check(self: Self, calculated_df) -> pd.DataFrame:
```
zaneselvans, line 1291: I think `check()` is probably a good name for this method, if the classes are named after the error metric then we'll have things like:

```py
# the generic case
error_metric.check(calculated_df)
error_frequency.check(calculated_df)
relative_error_magnitude.check(calculated_df)
```

which seems legible.

src/pudl/transform/ferc1.py
```diff
@@ -1670,12 +2091,12 @@ def add_calculation_corrections(
...
-        correction_parents = correction_components.assign(
-            table_name_parent=lambda t: t.table_name,
-            xbrl_factoid_parent=lambda x: x.xbrl_factoid,
-        ).drop(columns=["table_name", "xbrl_factoid"])
-        return pd.concat([calc_components, correction_components, correction_parents])
+        # # for every calc component, also make the parent-only version
+        # correction_parents = correction_components.assign(
+        #     table_name_parent=lambda t: t.table_name,
+        #     xbrl_factoid_parent=lambda x: x.xbrl_factoid,
+        # ).drop(columns=["table_name", "xbrl_factoid"])
```
zaneselvans, line 2098: Should this still be commented out?

src/pudl/transform/ferc1.py
```diff
@@ -5565,6 +5985,15 @@ def table_to_xbrl_factoid_name() -> dict[str, str]:
     }
 
 
+def table_to_column_to_check() -> dict[str, str]:
```
zaneselvans, line 5988: The docstring in this helper function is identical to the above helper function.

We already have some helper methods that are defined as properties built into `Ferc1TableTransformParams` that expose values from the bowels of the nested data structure for external use. Would it make sense to add this column in there? Maybe as `params.calculated_column`?

src/pudl/transform/ferc1.py
```diff
@@ -960,75 +1124,332 @@ def calculate_values_from_components(
...
+    These numbers may vary wildly from table to table so no default values for the
+    expected errors are provided here...
+    """
+
+    def metric(self: Self, gb: DataFrameGroupBy) -> pd.Series:
+        """Calculate the absolute mangnitude of XBRL calculation errors."""
+        return gb.abs_diff.sum()
+
+
+class NullCalculationFrequency(GroupMetricCheck):
```
zaneselvans, line 1369: To match the name naming of the next class below, this should probably be called something like `NullCalculatedValueFrequency`

Clarify in the docstring that we're calculating the frequency relative to the number of non-null reported values.

"""Fraction of non-null reported values that have null corresponding calculated values"""

src/pudl/transform/ferc1.py
```diff
@@ -960,75 +1124,332 @@ def calculate_values_from_components(
...
+    def calculate_metric(self: Self, df: pd.DataFrame) -> pd.Series:
+        """Calculate the frequency with which records are tagged as errors."""
+        return (
+            df[df.row_type_xbrl == "calculated_value"]
+            .groupby(self.groupby_by())
+            .apply(self.metric)
+        )
+
+    def metric(self: Self, gb: DataFrameGroupBy) -> pd.Series:
+        """Frequency of null calculated values when reported values are not."""
```
zaneselvans, line 1381: """Fraction of non-null reported values that have null corresponding calculated values"""

src/pudl/transform/ferc1.py
```diff
@@ -960,75 +1124,332 @@ def calculate_values_from_components(
...
+    def metric(self: Self, gb: DataFrameGroupBy) -> pd.Series:
+        """Calculate the absolute mangnitude of XBRL calculation errors."""
+        return gb.abs_diff.sum()
+
+
+class NullCalculationFrequency(GroupMetricCheck):
+    """Check the frequency of null calculated values."""
+
+    def calculate_metric(self: Self, df: pd.DataFrame) -> pd.Series:
+        """Calculate the frequency with which records are tagged as errors."""
```
zaneselvans, line 1373: """Only apply metric to rows that contain calculated values."""

zaneselvans (commented):
Mostly naming and documentation suggestions, but also I think I may have made a mistake in a couple of the error metrics initially, and we need to use some absolute values.

zaneselvans:
Playing with the results in a notebook using those snippets you sent, I'm wanting to drill down and identify which combinations of groupby columns identify the most egregious errors, but I don't think that's possible with just the summary output.

For example, the relative error magnitude has a huge spike in 2006, and I'd like to know what combinations of table, fact, and utility IDs are responsible for that.  Is it just a single entry that's off by a huge amount? Or is it a handful of utility filings that are super wrong in a single table? (maybe a table that changed its line number meanings in that 2006?)

Can we imagine an all-tables concatenated output that allows this kind of dynamic slicing and dicing of the data for diagnostic purposes?  Would it just be all of the tables with the standard names that get passed into the error checking infrastructure (with `reported_value`, `calculated_value`, `abs_diff`, `rel_diff` etc.) including all of their rows and all of the intact groupby columns (`report_year`, `xbrl_factoid`, `table_name`, `utility_id_ferc`)?

With such a table, is there a straightforward way to manually apply the different error metrics with multiple groupby columns and selections, so we can answer questions like "Looking just at 2006, what values of `utility_id_ferc1`, `xbrl_factoid`, and `table_name` are responsible for the biggest errors?" or "Given that `utility_id_ferc1==152` has a big relative error magnitude overall, is that error coming from a single year? A single `xbrl_factoid`? Or a range of years? Or a whole table of facts?

src/pudl/transform/ferc1.py
```diff
@@ -726,6 +727,116 @@ def combine_axis_columns_xbrl(
     return df
 
 
+class MetricInputs(TransformParams):
```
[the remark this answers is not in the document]
cmgosnell: I think we should go with `IsCloseTolerance` until it becomes clear we need more params!

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
cmgosnell: I tried this but ran into a snag that I think makes this whole migration pretty annoying.

Because these param classes are nested within the coordinating `ReconcileTableCalculations` param, when you need to override a metric's tolerance within a group, you currently get the default tolerances for all of the other metrics. which is to say you don't get any of the default tolerances for your group, assigned in `CalculationGroupChecks` which is a little annoying.

If you remove the defaults altogether and you want to edit any of the defaults for a table, you need to define all of the tolerances for all of the metrics for that table. Which is a little more than a little annoying.

Any idea about how to navigate this?

src/pudl/transform/ferc1.py
```diff
@@ -960,75 +1124,332 @@ def calculate_values_from_components(
...
+        "ungrouped", "table_name", "xbrl_factoid", "utility_id_ferc1", "report_year"
+    ]  # right now this only works with one column! bc of index rename in check()
+    """Name of group to check the metric based on.
+
+    Most groups are column names to use in groupby along with the ``table_name``. Both
+    ``ungrouped`` and ``table_name`` have some special exceptions in order to output the
+    metric results with the same index of ``table_name`` and ``group_value``. For the
+    ``ungrouped`` group, the values in ``table_name`` and ``group_value`` will be
+    ``ungrouped``. For the ``table_name`` group, the values in ``table_name`` and
+    ``group_value`` will be the table name values.
```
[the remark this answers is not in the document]
cmgosnell: i updated it. does it make sense to you now? and if not i'd like to :phone: about it so i can understand what is confusing to you.

src/pudl/transform/ferc1.py
```diff
@@ -960,75 +1124,332 @@ def calculate_values_from_components(
...
+        reasonable to only look at records where ``abs_diff`` is not Null?
+        """
+        return pd.Series(
+            ~np.isclose(
+                df["calculated_value"],
+                df["reported_value"],
+                rtol=self.metric_inputs.isclose_rtol,
+                atol=self.metric_inputs.isclose_atol,
+            )
+            & df["abs_diff"].notnull()
```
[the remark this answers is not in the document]
cmgosnell: I think the real intention here is to say: only check this if this is is actually a calculated value and this `df["abs_diff"].notnull()` is a proxy for that. I think it is possible to use `df.row_type_xbrl == "calculated_value"` instead as a more direct way of doing this.

I will test this.

src/pudl/transform/ferc1.py
cmgosnell: is this just an fyi? or should i add this into the docs/comments?

src/pudl/transform/ferc1.py
```diff
@@ -960,75 +1124,332 @@ def calculate_values_from_components(
...
+
+
+class RelativeErrorMagnitude(GroupMetricCheck):
+    """Check relative magnitude of errors in XBRL calculations."""
+
+    def metric(self: Self, gb: DataFrameGroupBy) -> pd.Series:
+        """Calculate the mangnitude of the errors relative to total reported value."""
+        try:
+            # Should we be taking the absolute value of the reported column?
+            return gb[gb.is_not_close].abs_diff.sum() / gb["reported_value"].abs().sum()
```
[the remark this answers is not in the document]
cmgosnell: yea these are good points thanks for noticing this! I've updated. It didn't change the # of errors for this metric but i'm sure it had some small influence

src/pudl/transform/ferc1.py
cmgosnell: i truly don't remember why i commented these out.... but i put em back in there

src/pudl/transform/ferc1.py
```diff
@@ -726,6 +727,116 @@ def combine_axis_columns_xbrl(
...
+        null_calculation_frequency=0.50,
+        null_reported_value_frequency=0.68,
+    )
+
+    # @validator("ungrouped", "utility_id_ferc1", "report_year", "table_name", "xbrl_factoid")
+    # def grouped_tol_ge_ungrouped_tol(v, values):
+    #    """Grouped tolerance should always be greater than or equal to ungrouped."""
+    #    if v is not None:
+    #        assert v >= values["ungrouped"]
+    #    return v
```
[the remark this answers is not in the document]
cmgosnell: i added the validator into `GroupMetricChecks`. but its very much a pain to have on because of the aforementioned overriding of the defaults.

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
cmgosnell: I added an explanation in the docs

src/pudl/transform/ferc1.py
cmgosnell: i've updated the doc string. I'm going to hold off on this other change just because there is so so much going on in this pr already.

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
cmgosnell: we decided to keep this as is even though its a little confusing to need to tweak the defaults in 3 places. it seems more annoying to need to completely specify these tolerances any time you want to tweak one.

src/pudl/transform/ferc1.py
```diff
@@ -770,129 +880,183 @@ def reconcile_table_calculations(
...
+def _check_subtotal_calculations(
+    df: pd.DataFrame,
+    intra_table_calcs: pd.DataFrame,
+    xbrl_metadata: pd.DataFrame,
+    params: "Ferc1TableTransformParams",
+    dim_cols: list[str],
+    table_name: str,
+    table_dims: pd.DataFrame,
+    calc_idx: list[str],
+) -> None:
```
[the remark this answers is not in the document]
cmgosnell: i broke it out into two steps. there are still a total of 8 params but they are more segmented which is nicer imo.

src/pudl/output/ferc1.py
```diff
@@ -1402,7 +1432,7 @@ def boom(self: Self, tables_to_explode: dict[str, pd.DataFrame]) -> pd.DataFrame
 
         Args:
             tables_to_explode: dictionary of table name (key) to transfomed table (value).
-            calculation_tolerance: What proportion (0-1) of calculated values are
+            group_metric_checks: What proportion (0-1) of calculated values are
             allowed to be incorrect without raising an AssertionError.
```
zaneselvans, line 1436: Update docstring.

src/pudl/output/ferc1.py
```diff
@@ -16,74 +16,104 @@
...
 
 logger = pudl.logging_helpers.get_logger(__name__)
 
 
-EXPLOSION_CALCULATION_TOLERANCES: dict[str, CalculationChecks] = {
-    "income_statement_ferc1": CalculationChecks(
-        group_checks=CalculationGroupChecks(
-            ungrouped=CalculationMetricTolerance(
-                error_frequency=0.20, null_calculation_frequency=1.0
+EXPLOSION_CALCULATION_TOLERANCES: dict[str, GroupMetricChecks] = {
```
zaneselvans, line 27: Curious if you have any initial thoughts about the errors observed in the exploded tables?

src/pudl/transform/ferc1.py
```diff
@@ -824,11 +824,28 @@ class CalculationChecks(TransformParams):
...
+    #         metric_tolerances = group_metric_tolerances.dict().get(group)
+    #         for metric_name, tolerance in metric_tolerances.items():
+    #             ungrouped_tolerance = group_metric_tolerances.dict()["ungrouped"].get(
+    #                 metric_name
+    #             )
+    #             if tolerance < ungrouped_tolerance:
+    #                 raise AssertionError(
+    #                     f"In {group=}, {tolerance=} for {metric_name} should be greater than {ungrouped_tolerance=}."
+    #                 )
+    #     return values
```
zaneselvans, line 848: Did this end up having other problems that weren't simple?

zaneselvans (commented):
I'm still fuzzy on that one docstring, but I left suggested language that reflects my understanding of what it's supposed to be saying.

I left another larger question in the comments on the PR, about how we can make it easy to interactively explore errors in more than one dimension to narrow down the exact source of the problems, which I think may require another concatenated asset, but that can be done in a separate PR.

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
cmgosnell: thread continued below w/ updated docs.

src/pudl/transform/ferc1.py
cmgosnell: The mechanics of the check are okay imo, but the substance of the check itself is a pain because of the various ways to set these tolerances. I think it would have been simpler if we could have removed one layer of defaults but as we discussed that was less simple.

src/pudl/output/ferc1.py
cmgosnell: no i didn't investigate deeply. mostly was just working on the conversion itself, not the errors

zaneselvans:
Looks like the `ungrouped` `error_frequency` tolerance for the `balance_sheet_assets_ferc1` table was a bit too low (at least for the fast ETL) so I bumped it from 0.00013 to 0.0002.

PR #2917, opened 2023-10-04
Rename `heat_rate_mmbtu_mwh -> heat_rate_mmbtu_mwh_by_unit`

bendnorman:
# PR Overview

This PR rename the `heat_rate_mmbtu_mwh -> heat_rate_mmbtu_mwh_by_unit` so it's clear heat rate is a unit level attribute. We decided to make this change because assets downstream of `_out_eia__{freq}_heat_rate_by_unit` broadcast `heat_rate_mmbtu_mwh` across generators.

@katie-lamb I replaced all instances of `heat_rate_mmbtu_mwh` with `heat_rate_mmbtu_mwh_by_unit`. Are there parts of the code where it doesn't make sense to use `heat_rate_mmbtu_mwh_by_unit`? I think it might be a little duplicative to have a `heat_rate_mmbtu_mwh_by_unit` column in the `heat_rate_mmbtu_mwh_by_unit` asset but I thought it would be nice for the column name to be consistent / traceable throughout all of the assets.

# PR Checklist


zaneselvans:
Maybe it's too late to catch this, but [our convention](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/naming_conventions.html) has been to have the physical units be the last component of column names.

Having `_by_unit` at the end also kind of clashes with the occasional `_per_unit` that we've used  when a column has mixed units (e.g. fuel costs which can be per ton, per mcf, per gallon, per barrel) with `_unit` standing in for the ambiguous physical unit, whereas here the `unit` is a generation unit.

The physical units here are really `[MMBtu / MWh]` and in [almost all composite unit column names](https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/pudl_db.html) we've made the `per` explicit as in `avg_customers_per_month`, `min_fuel_mmbtu_per_unit`, `fuel_cost_per_mmbtu`, `fuel_cost_per_mwh` etc. which I think is clearer. In the context of all those other names containing `_per_` it would be easy for someone to interpret `_mmbtu_mwh` as units of `[MMBtu * MWh]`

So it seems like this column should end with `_mmbtu_per_mwh`, maybe like `unit_heat_rate_mmbtu_per_mwh`

bendnorman:
Ah great catch thank you @zaneselvans! I'm a total naming convention hypocrite! `unit_heat_rate_mmbtu_per_mwh` makes more sense, thanks for explaining. I'll update the name.

zaneselvans (approved):

katie-lamb (approved):
Looks good!

PR #2931, opened 2023-10-10
Fix too-many-paths error in nightly build

jdangerx:
The nightly build was breaking for the `exploded_balance_sheet_assets_ferc1` asset, due to having too many paths between a particular root node & leaf node. Now the asset builds, at least on *my* machine. I don't believe there are validation tests affected by this change.

After talking with @zaneselvans we decided that we could handle multiple paths, and only raise an error if the paths conflict.

Not in scope:
* renaming the `forest` stuff since we don't have a forest anymore - maybe call it `digraph` or `dag`?

Maybe in scope / maybe a good follow-on PR:
* writing some unit tests - this calc graph stuff seems like a pretty good candidate for property based testing.

### DIFF
--- a/src/pudl/output/ferc1.py
+++ b/src/pudl/output/ferc1.py
@@ -2023,6 +2023,13 @@ class XbrlCalculationForestFerc1(BaseModel):
             stepparents = stepparents.union(graph.predecessors(stepchild))
         return list(stepparents)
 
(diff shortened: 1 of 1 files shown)


zaneselvans (approved):
This looks good to me if you ran it locally on the Full ETL and it passed.

All this Explosion stuff is heinously under-tested. 😬

PR #2934, opened 2023-10-11
2899 merge keys

### LINKED ISSUE #2899: Non-unique merge keys in FERC1 assets with 2021-2022 data
```
pandas.errors.MergeError: Merge keys are not unique in left dataset; not a one-to-one merge

  File "/Users/dazhong-catalyst/work/pudl/src/pudl/transform/ferc1.py", line 948, in calculate_values_from_components
    calculated_df = pd.merge(
```

* [ ] electric_operating_expenses_ferc1
* [ ] electric_operating_revenues_ferc1
* [ ] income_statement_ferc1
* [ ] utility_plant_summary_ferc1

jdangerx:
Closes #2899 .

# Overview
The 2021 XBRL filings for a utility included 2020 and 2021 data, and the 2022 filings included 2021 and 2022 data. This caused a ton of duplicate data for 2021. Additionally, many utilities filed multiple reports for each year, causing further duplication. We handled this duplication in `ferc-xbrl-extractor<1.2` but it makes more sense to have the deduplication centralized in PUDL.

After implementing this, most core FERC1 assets build with 2022 data, except for `retained_earnings_ferc1` and all the downstream assets in the output layer.

# !!!
We need https://github.com/catalyst-cooperative/ferc-xbrl-extractor/pull/151 to get merged, and then to push an actual `v1.2.0` tag, before we can actually run this code - I've been testing by running `pip install -e .` in the `ferc-xbrl-extractor` repository.

# Testing strategy

In addition to a traditional "manual example test," I added some property-based testing for the deduplication logic, to better communicate the actual expectations we have of the data.

There's an open issue in `pandera`, though, about the hypothesis strategies being very slow: https://github.com/unionai-oss/pandera/issues/1116 . It doesn't seem like it will quite get solved quickly. And there's also a corresponding issue in `hypothesis`: https://github.com/HypothesisWorks/hypothesis/issues/3701

This manifests as the single test taking several seconds to run, which is... pretty dang slow for a unit test. If that feels too slow to other folks I can lower the example count, or even drop the whole test. Though, you'll note that the property-based test is half the size and encapsulates a lot more information than the manual test....

# For reviewers

@zaneselvans this is the 'per-filing' approach, not the 'per-fact' approach - that seemed simpler to implement. let me know if you feel this is adequate, or if we need to dig in more to figure out which approach to use.

@zschira this is mostly an FYI about changes to how we are using the output of extractor / why that other PR matters.

@e-belfer this is mostly an FYI, since you've been thinking a bit about Panderas in the PHMSA context.


zaneselvans:
@jdangerx by "per filing" you mean taking the most recent filing for each (utility, year) and assuming that it will contain all pertinent facts? Rather than taking the last reported value for each fact in each (utility, year) even if they come from different filings?

src/pudl/io_managers.py
```diff
@@ -726,6 +770,13 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
                 },
             ).assign(sched_table_name=sched_table_name)
 
+        primary_key = self._get_primary_key(table_name)
+        context_cols = [
+            col for col in primary_key if col not in {"filing_name", "publication_time"}
+        ]
+        deduped = self.use_latest_filing_for_context(df, context_cols)
```
zaneselvans, line 777: Just for readability / ease of mental modeling I find named keyword arguments easier, or using the same variable names where it's arbitrary -- e.g. maybe rename `context_cols` to `unique_cols` here, or:

```py
        deduped = self.use_latest_filing_for_context(df, unique_cols=context_cols)
```

src/pudl/io_managers.py
```diff
@@ -683,6 +685,47 @@ class FercXBRLSQLiteIOManager(FercSQLiteIOManager):
     metadata.
     """
 
+    @staticmethod
+    def use_latest_filing_for_context(
+        table: pd.DataFrame, unique_cols: list[str]
```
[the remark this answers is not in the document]
zaneselvans: Since we have a `context` in PUDL that refers to the Dagster execution context I think it would be helpful to add a note in the docstring differentiating this "context".  Like we need a little context context. This method name makes me think that it's going to like, get one record that can be used as a template or as a representative sample of all the records.

src/pudl/transform/ferc1.py
```diff
@@ -4762,7 +4762,8 @@ def add_previous_year_factoid(self, df: pd.DataFrame) -> pd.DataFrame:
         ]
         # If previous_year type factoids aren't in all report_years, make factoids
         # for these years. Raise exception if more than one year.
```
zaneselvans, line 4764: Is this comment still relevant? I"m not seeing the assertion it refers to anywhere.

test/unit/io_managers_test.py
```diff
@@ -249,3 +255,232 @@ def test_error_when_reading_view_without_metadata(pudl_sqlite_io_manager_fixture
...
+def test_ferc_xbrl_sqlite_io_manager_dedupes(mocker, tmp_path):
+    db_path = tmp_path / "test_db.sqlite"
+    datapackage = json.dumps(
+        {
+            "profile": "tabular-data-package",
+            "name": "test_db",
+            "title": "Ferc1 data extracted from XBRL filings",
+            "resources": [
+                {
+                    "path": f"sqlite:///{db_path}",
```
zaneselvans, line 269: * I don't think a `sqlite:///` path is valid in a datapackage, at least not until using Frictionless v5.
* According to the spec I don't think these can have `tabular-data-package` or `tabular-data-resource` profiles because it's not referring to either CSV or inline JSON data.

test/unit/io_managers_test.py
```diff
@@ -249,3 +255,232 @@ def test_error_when_reading_view_without_metadata(pudl_sqlite_io_manager_fixture
...
+        },
+    )
+    io_manager = FercXBRLSQLiteIOManager(base_dir=tmp_path, db_name="test_db")
+    observed_table = io_manager.load_input(input_context)
+
+    assert len(observed_table) == 1
+    assert observed_table.str_factoid.to_numpy().item() == "updated 2021 EOY value"
+
+
+example_schema = pa.DataFrameSchema(
```
zaneselvans, line 365: We should be careful to as we roll out Pandera -- we also currently do:

```py
import pyarrow as pa
```

which could get confusing.

test/unit/io_managers_test.py
```diff
@@ -249,3 +255,232 @@ def test_error_when_reading_view_without_metadata(pudl_sqlite_io_manager_fixture
...
+        "publication_time": pa.Column("datetime64[ns]", nullable=False),
+        "int_factoid": pa.Column(int),
+        "float_factoid": pa.Column(float),
+        "str_factoid": pa.Column("str"),
+    }
+)
+
+
+@hypothesis.given(example_schema.strategy(size=3))
+def test_get_unique_row_per_context(df):
```
zaneselvans, line 381: More context context... I'm not clear on what that means in XBRL-land.

And maybe add a note indicating what other tests in here this duplicates so we don't get confused later about why things show up twice and wonder if they're trying to test different things.  Or are you thinking of removing one or the other of them?

test/unit/io_managers_test.py
```diff
@@ -249,3 +255,232 @@ def test_error_when_reading_view_without_metadata(pudl_sqlite_io_manager_fixture
...
+    assert (deduped.merge(df, how="left", indicator=True)._merge != "left_only").all()
+    # for every [entity_id, utility_type, date] - th"true"e is only one row
+    assert (~deduped.duplicated(subset=context_cols)).all()
+    # for every *context* in the input there is a corresponding row in the output
+    original_contexts = df.groupby(context_cols, as_index=False).last()
+    paired_by_context = original_contexts.merge(
+        deduped, on=context_cols, how="outer", suffixes=["_in", "_out"], indicator=True
+    ).set_index(context_cols)
+    hypothesis.note(original_contexts)
+    hypothesis.note(deduped)
```
zaneselvans, line 396: What does `hypothesis.note()` do? I read the docs but I'm still confused.

> The note is printed for the minimal failing example of the test in order to include any additional information you might need in your test.

zaneselvans (changes requested):
Just a few readability / docs things and questions about how things work...

src/pudl/io_managers.py
```diff
@@ -643,6 +644,7 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
             context: dagster keyword that provides access output information like asset
                 name.
         """
+        # TODO (daz): this is hard-coded to FERC1, though this is nominally for all FERC datasets.
```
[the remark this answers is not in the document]
jdangerx: Yeah, it's currently definitely just a FERC1 extractor. I think we can safely wait until we start extracting from more than one FERC form to revisit this design. One thing we'll end up dealing with eventually is that FERC 714 has a different identification table than the others, which means the SQL will have to be different.

I have a hunch that once we start actually making multiple different FERC XBRL extractors, we'll want to collapse the type hierarchy a bit - to figure out what a `FercXBRLSQLiteIOManager` does you also need to see what `FercSQLiteIOManager` and `SQLiteIOManager` do, and figure out the method-override relationships between them. But again, I think we can wait until we actually need to do this refactor to do it.

src/pudl/io_managers.py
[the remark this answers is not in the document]
jdangerx: Yeah, this XBRL context thing is confusing as hell, good to call it out so that we're not completely 🤯 in a month!

src/pudl/io_managers.py
```diff
@@ -683,6 +685,47 @@ class FercXBRLSQLiteIOManager(FercSQLiteIOManager):
...
+    ) -> pd.DataFrame:
+        """Get facts from the latest filing that reported for each context.
+
+        We treat two XBRL contexts that are the same except for their IDs as the same context.
+        """
+        strict_deduped = table.drop_duplicates()
+        logger.debug(
+            f"Dropped {len(table) - len(strict_deduped)} completely duplicated rows."
+        )
+        chrono_order = table.sort_values("publication_time")
```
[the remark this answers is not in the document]
jdangerx: > I'm not sure it makes a functional difference since the next `drop_duplicates()` is more liberal, but Is this supposed to be `strict_deduped.sort_values()` rather than `table.sort_values()`? It doesn't look like `strict_deduped` is ever referred to again after it is set above, other than to log the number of dropped records.

Yep, thanks for catching!

> It seems like it would be possible to condense all of these iterative `drop_duplicates()` into a single sort-and-drop_dupes. We're just doing in stages so we can log and have a sense of how many duplicates we're getting at each stage?

Yeah, I was doing that to log and understand the duplicate counts. We had wanted this at some point. I turned this into a loop to be a little clearer about what's going on, but I also think it's reasonable to just run it all in one stage and log its drop rate instead of one per step. LMK if you have a preference either way.

src/pudl/transform/ferc1.py
jdangerx: nope!

test/unit/io_managers_test.py
jdangerx: Oooh, yeah. I can keep them all as `pandera` for now.

test/unit/io_managers_test.py
jdangerx: I was thinking of removing the manual one.

test/unit/io_managers_test.py
jdangerx: Hypothesis runs a lot of iterations of each test, right - and once there's a failure, they try to find the simplest failing input.

So if you were to log things out, you'd have a lot of output to sift through. And if you were to stick a debugger in the test, you'd end up hitting the breakpoint hundreds of times.

`note()` is like `print()` but instead of printing *every* time, it prints output *only* for the minimal failing example, so you can understand what's going on when it fails.

test/unit/io_managers_test.py
jdangerx: This is a fake datapackage that only exists to mimic the structure of a datapackage descriptor and hold the primary keys, so I'll just make this a "csv" and add a comment to that effect here.

Separately, was looking at the spec & I guess all datapackages [MUST](https://specs.frictionlessdata.io/data-package/#descriptor) be named `datapackage.json`? I guess we're out of spec there, too.

jdangerx:
If CI + code review passes, we should merge this into `2811` so we have a fully building/test-passing version there. Then we can make any validation test-related fixes on the main branch.

jdangerx:
> @jdangerx by "per filing" you mean taking the most recent filing for each (utility, year) and assuming that it will contain all pertinent facts? Rather than taking the last reported value for each fact in each (utility, year) even if they come from different filings?

Yes, exactly!

test/unit/io_managers_test.py
zaneselvans: Yeah, there's that too...

We should eventually be able to annotate all of the outputs -- multiple SQLite DBs and parquet files -- with a single datapackage.json that refers to the individual tables in all of the different files as resources.

But good to switch it to CSVs for now in this testing setup.

test/unit/io_managers_test.py
zaneselvans: Ah okay that's really cool!

zaneselvans: So if `hypothesis.note()` results in output, that means there's a failure, so if all is well we shouldn't be seeing any output.

test/unit/io_managers_test.py
zaneselvans: I think that's a fine idea.

src/pudl/io_managers.py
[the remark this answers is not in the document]
zaneselvans: I don't have strong feelings either way. Just wanted to make sure I understood that this was an iterative narrowing, and that there wasn't something tricky going on that meant it had to be done in several stages.

zaneselvans (approved):

zaneselvans:
Unfortunately CI is currently doomed to fail by Zenodo, but I think we can probably merge into the feature branch.

Issue #2930, opened 2023-10-10
EIA923M: Add data maturity flag and 2023 quarterly data


aesharpe:
Our current 923 archiver contains annual and YTD data. We need to come up with a way to distinguish the data maturity of each of the partitions. The way we currently determine data_maturity is via the `add_data_maturity()` function in the `extract/excel.py` module. In the future it might make sense to add a `data_maturity` field to the archive metadata, but **for now, we'll update the `add_data_maturity()` to account for the EIA923 data.** There's a category for the `data_maturity` column called `incremental_ytd` that was created for the purpose of this type of data.

The raw files are formatted as such (examples): 

`EIA923_Schedules_2_3_4_5_M_07_2023_20SEP2023.xlsx` 
`EIA923_Schedules_2_3_4_5_M_12_2022_Early_Release.xlsx`

Where the value after `M` is the the last month the have data. Data with `12` is annual, data with values less than `12` are YTD. We can use regex to extract the month from here and add the `incremental_ytd` flag if it's not 12.

This issue will also inherently add the 2023 quarterly data for 923.

aesharpe:
Close with #2936

e-belfer:
Marked done but not closed, closing

PR #2936, opened 2023-10-12
Add data maturity for 923m

aesharpe:
# Explanation:

From the issue #2930: 

> Our current 923 archiver contains annual and YTD data. We need to come up with a way to distinguish the data maturity of each of the partitions. The way we currently determine data_maturity is via the add_data_maturity() function in the extract/excel.py module. In the future it might make sense to add a data_maturity field to the archive metadata, but for now, we'll update the add_data_maturity() to account for the EIA923 data. There's a category for the data_maturity column called incremental_ytd that was created for the purpose of this type of data.

> The raw files are formatted as such (examples):

> EIA923_Schedules_2_3_4_5_M_07_2023_20SEP2023.xlsx
EIA923_Schedules_2_3_4_5_M_12_2022_Early_Release.xlsx

> Where the value after M is the the last month the have data. Data with 12 is annual, data with values less than 12 are YTD. We can use regex to extract the month from here and add the incremental_ytd flag if it's not 12.

> This issue will also inherently add the 2023 quarterly data for 923.

# Errors to Fix:

- Unit tests are failing because there are more years of EIA923 data, now that we have the 2023 monthly data in there, than EIA860. How do we want to handle this miss-match? `test/unit/settings_test.py::test_eia923_dependency`.

```
=================================== FAILURES ===================================
____________________ TestEiaSettings.test_eia923_dependency ____________________

self = <test.unit.settings_test.TestEiaSettings object at 0x281055ad0>

    def test_eia923_dependency(self):
        """Test 860 is added if 923 is specified and 860 is not."""
        eia923_settings = Eia923Settings()
>       settings = EiaSettings(eia923=eia923_settings)

test/unit/settings_test.py:139: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pydantic.error_wrappers.ValidationError: 1 validation error for EiaSettings
E   __root__ -> __root__
E     '[2023]' years are not available. (type=value_error)

pydantic/main.py:341: ValidationError
```

- Validation `test_minmax_rows` tests for `gen_eia923` monthly and annual are negative which seems wrong. When I looked at the table equivalent in the database, the row numbers were completely different, so I need to investigate that too. Would like some help understanding what I should be looking at because this test still seems tailored to the old `pudl_out` system with monthly and annual resolution, and it's a bit confusing.

```
eia_test.py::test_minmax_rows[eia_annual-gen_eia923-None-5178892-433286] - ValueError: gen_eia923: found 433277 rows, expected 433286. Off by -0.002%, allowed margin of 0.000%
FAILED test/validate/

eia_test.py::test_minmax_rows[eia_monthly-gen_eia923-None-5178892-433286] - ValueError: gen_eia923: found 5178828 rows, expected 5178892. Off by -0.001%, allowed margin of 0.000%
FAILED test/validate/
```
It appears that 64 ["report_date", "plant_id_eia", "utility_id_eia", "generator_id"] groupings are dropped from this gen table at some point in the process.


src/pudl/output/eia923.py
```diff
@@ -126,6 +126,30 @@ def _fill_fuel_costs_by_state(
...
+            data_maturity = incremental_ytd.
+    """
+    if freq == "AS":
+        logger.info(
+            "Removing rows where data_maturity is incremental_ytd to avoid "
+            "aggregation errors."
+        )
+        return df.loc[df["data_maturity"] != "incremental_ytd"]
+    else:
+        return df
```
cmgosnell, line 150: this is a total nit but:
```suggestion
        df = df.loc[df["data_maturity"] != "incremental_ytd"]
    return df
```

irrc its best practice to attempt to keep the return only in one place.

and maybe it needs a  `.copy()`?

src/pudl/extract/excel.py
```diff
@@ -195,6 +196,17 @@ def add_data_maturity(self, df: pd.DataFrame, page, **partition) -> pd.DataFrame
...
             maturity = "monthly_update"
+        elif self._dataset_name == "eia923":
+            # this conditional is a hacky way to get around the fact that the formatting
+            # for the file names didn't always look like "EIA923_Schedules_2_3_4_5_M_"
+            # But, because we only want YTD data, it's fine to set a year floor.
+            if partition["year"] > 2022:
+                release_month = re.search(
+                    r"EIA923_Schedules_2_3_4_5_M_(\d{2})",
+                    self.excel_filename(page, **partition),
+                ).group(1)
```
cmgosnell, line 207: if the filename for the YTD file always contains `EIA923_Schedules_2_3_4_5_M_` you could do something like:
```suggestion
            file_name = self.excel_filename(page, **partition)
            ytd_file_name_start = "EIA923_Schedules_2_3_4_5_M_"
            if ytd_file_name_start in file_name:
                release_month = re.search(
                    r"ytd_file_name_start(\d{2})", file_name,
                ).group(1)
```

i think you could probably even use the `if ytd_file_name_start in file_name:` as the `elif` instead of checking for 923 first.

src/pudl/package_data/eia923/column_maps/emissions_control.csv
cmgosnell, line 1: I thiiink these extra lines at the end of the csvs are why the linters are failing.

src/pudl/package_data/eia923/file_map.csv
cmgosnell, line 1: Are we still on early release for 923 for 2022?? ah okay the most recent archive has june 2023 data and ER 2022. while the most fresh data from eia looks like its final 2022 and july 2023. i'd suggest getting this june 2023 stuff through the finish line bc it seems close and then maybe we can jump back in and make a fresh archive.

test/validate/eia_test.py
```diff
@@ -45,15 +45,15 @@ def test_no_null_cols_eia(pudl_out_eia, live_dbs, cols, df_name):
...
         ("bga_eia860", 141_638, 141_638, 141_638),
-        ("boil_eia860", 80_059, 80_059, 80_059),
-        ("frc_eia923", 631_975, 258_115, 25_370),
+        ("boil_eia860", 83_354, 83_354, 83_354),
+        ("frc_eia923", 639_647, 261_583, 25_370),
         ("gen_eia923", None, 5_178_892, 433_286),
-        ("gens_eia860", 556_578, 556_578, 556_578),
-        ("gf_eia923", 2_859_269, 2_859_269, 244_796),
+        ("gens_eia860", 556_586, 556_586, 556_586),
+        ("gf_eia923", 2_879_896, 2_879_896, 244_796),
```
cmgosnell, line 54: do you have any sense of why the annual rows didn't increase for some of these?

cmgosnell (changes requested):
this all looks pleasantly straightforward! I sprinkled a few minor suggestions.

also i didn't do a super tight look at the CSVs but it looks like they were basically all copy/paste?!? cool if so! i wonder how straightforward adding the non-ytd release is gonna be in this regard.

src/pudl/extract/excel.py
aesharpe: This makes sense, but also it won't fail if there is no capture group found. I feel like it would be good if this failed if, for example, the file naming changed and we needed to come up with a new way to grab it.

src/pudl/package_data/eia923/column_maps/emissions_control.csv
aesharpe: Oh yea for sure - I keep committing with a `--no-verify` flag because the other tests are failing, but that means that the linters don't automatically fix stuff.

aesharpe: Oh yea for sure - I keep committing with a `--no-verify` flag because the other tests are failing, but that means that the linters don't automatically fix stuff.

src/pudl/package_data/eia923/file_map.csv
aesharpe: I think that's a good idea!!

test/validate/eia_test.py
aesharpe: I don't. But I was also thinking about this. When I dig into the issues with the `gen_eia923` table I'll look at this too

aesharpe:
> * Validation `test_minmax_rows` tests for `gen_eia923` monthly and annual are negative which seems wrong. When I looked at the table equivalent in the database, the row numbers were completely different, so I need to investigate that too. Would like some help understanding what I should be looking at because this test still seems tailored to the old `pudl_out` system with monthly and annual resolution, and it's a bit confusing.

This is an ongoing and unknown issue documented here

test/validate/eia_test.py
aesharpe: This is now a non-issue because we remove the ytd records from the annual data anyways.

src/pudl/transform/eia.py
```diff
@@ -799,7 +799,6 @@ def boiler_generator_assn_eia860(context, **clean_dfs) -> pd.DataFrame:  # noqa:
             "steam_plant_type_code",
             "net_generation_mwh",
             "missing_from_923",
-            "data_maturity",
```
cmgosnell, line 802: this is probably related to why this guy is removed from the schema i assume.

test/unit/settings_test.py
```diff
@@ -133,21 +132,21 @@ def test_860m(self):
...
-    def test_eia860_dependency(self):
-        """Test 923 tables are added to eia860 if 923 is not specified."""
-        eia860_settings = Eia860Settings()
-        settings = EiaSettings(eia860=eia860_settings)
-        assert settings.eia923.years == eia860_settings.years
+    # def test_eia860_dependency(self):
+    #     """Test 923 tables are added to eia860 if 923 is not specified."""
+    #     eia860_settings = Eia860Settings()
+    #     settings = EiaSettings(eia860=eia860_settings)
+    #     assert settings.eia923.years == eia860_settings.years
```
cmgosnell, line 149: we chatted about this on the ☎️  and came to the understanding that despite this restriction not being technically necessary across the board, it'd be good to require an overlap between 923 and 860 years.

src/pudl/output/eia923.py
```diff
@@ -19,6 +18,7 @@
     "utility_id_eia",
     "utility_id_pudl",
     "utility_name_eia",
+    "data_maturity",
```
cmgosnell, line 21: this is a total nit but i don't think this lil guy should be in the first cols.

src/pudl/output/eia923.py
```diff
@@ -230,11 +255,12 @@ def denorm_boiler_fuel_eia923(
         boiler_fuel_eia923["fuel_consumed_units"]
         * boiler_fuel_eia923["fuel_mmbtu_per_unit"]
     )
-    return denorm_by_boil(
+    dd = denorm_by_boil(
         boiler_fuel_eia923,
-        pu=denorm_plants_utilities_eia,
+        pu=denorm_plants_utilities_eia.drop(columns=["data_maturity"]),
         bga=boiler_generator_assn_eia860,
     )
```
cmgosnell, line 262: this could be a bad suggestion bc i don't know all the places `denorm_by_X` are used but is it possible to do these drop `data_maturity`'s in there instead of having them sprinkled throughout?

cmgosnell (commented):
There are several nits + the ask to update the settings test + an ask to keep the `data_maturity` in the boiler generators table + two question below:

* Do you need to update the `package_data/settings` files? (i thought i saw that updated the first time around)
* Do you need to push a schema update as well??

src/pudl/metadata/resources/eia860.py
```diff
@@ -134,7 +134,6 @@
                 "boiler_generator_assn_type_code",
                 "steam_plant_type_code",
                 "bga_source",
-                "data_maturity",
```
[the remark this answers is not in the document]
aesharpe: Added it back! It was related to this table merging with other tables that already had `data_maturity` fields and them canceling each other out. I found a way around this.

src/pudl/output/eia.py
```diff
@@ -137,7 +137,7 @@ def denorm_generators_eia(
 
     # Bring in some generic plant & utility information:
     pu_eia = denorm_plants_utilities_eia.drop(
-        ["plant_name_eia", "utility_id_eia"], axis="columns"
+        ["plant_name_eia", "utility_id_eia", "data_maturity"], axis="columns"
```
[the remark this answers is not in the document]
aesharpe: yes, but I changed where this column get's dropped!

src/pudl/output/eia923.py
aesharpe: I think this is a good idea. I restructured the code to reflect this.

test/unit/settings_test.py
cmgosnell: I think both of these tests could be converted to use `isdisjoint`. We want to make sure they are not disjoint, so i think we cold do something like this for both tests:

```
# assign both EIA form years
eia860_years = settings.eia860.years
eia923_years = data_source.working_partitions["years"]
# assert that there is some overlap between EIA years
assert not set(eia860_years).isdisjoint(eia923_years)
```

cmgosnell (commented):
I think this all looks great! It looks like the settings tests weren't converted but I made some suggestions over there. Once the tests all pass let's merge this bb

test/unit/settings_test.py
aesharpe: I'm not sure what the difference is between these two tests? Did you mean to reverse the 860 and the 923 labels? Also, what is the difference between `eia923_years = data_source.working_partitions["years"]` and `eia923_years = settings.eia923.years`?

cmgosnell: they just some from different places: the working partitions comes from the stored working partitions while the `settings.eia923.years` come from the settings file that the test is checking.

aesharpe: I guess I don't understand what you mean by 1st and 2nd test then. They seem to be doing the same thing?

cmgosnell: i believe these are both checked because of the way the settings classes themselves are constructed. in the `pudl.settings.EiaSettings.check_eia_dependencies` we are setting default values for 860 and 923 years which can either come from the settings file or is by default grabbed from the working partitions,

aesharpe: Should we also test this in the reverse as in: 

```
set(eia923_years).isdisjoint(eia860_years)
```

test/unit/settings_test.py
```diff
@@ -132,21 +132,18 @@ def test_860m(self):
...
+        settings = EiaSettings()
+        data_source = DataSource.from_id("eia860")
+        assert settings.eia860
+        # assign both EIA form years
+        eia860_years = settings.eia860.years
+        eia923_years_partition = data_source.working_partitions["years"]
+        eia923_years_settings = settings.eia923.years
+        # assert that there is some overlap between EIA years
+        assert not set(eia860_years).isdisjoint(eia923_years_partition)
+        assert not set(eia860_years).isdisjoint(eia923_years_settings)
```
cmgosnell, line 146: The previous version of these two test were checking if the **default** years of 923 and 860 were getting added in when there was only 860 or 923 data in the settings respectively.

The first `test_eia923_dependency` test was checking this scenario:
* **Previously**: If you fed PUDL settings EIA 923 only, our `EiaSettings` knows that 860 also needs to exist and gives it some default years. Are the 860 years the same as the 923 years?
* **What we could update it to now**: If you fed PUDL settings EIA 923 only, our `EiaSettings` knows that 860 also needs to exist and gives it some default 860 years. Are the 860 years appropriate i.e.: do they have some overlap w/ 923? OR are the assigned default years the years in the partitions?

(honestly i'm not sure now why the previous setup for `test_eia923_dependency` didn't work bc you didn't touch the working partitions for 860 in this PR)

for `test_eia860_dependency`:
* **Previously**: If you fed PUDL settings w/ EIA 860 only, our `EiaSettings` knows that 923 also needs to exist and gives it some default years. Are the 923 years the same as the 860 years?
* **Now**: If you fed PUDL settings w/ EIA 860 only, our `EiaSettings` knows that 923 also needs to exist and gives it some default years. Do the 923 default years at least overlap with the 860 years?


tl;dr version is i think the combining of these two tests removes some important information. Your test builds a default `EiaSettings` object without feeding it anything specific with ends ups triggering `EiaSettings.default_load_all`. The intention previously as I understand it was to test the `EiaSettings.check_eia_dependencies`.

cmgosnell (changes requested):
see my comment re the settings tests.

also i still don't see any updates to the `package_data/settings/etl_fast` or `package_data/settings/etl_full` so i don't think the CI will have actually tested the ytd info yet.

test/unit/settings_test.py
aesharpe: Hm ok, my version doesn't combine the two tests (`test_eia860_dependency` and `test_eia923_dependency`) it just combines the two tests that you suggested above? I fixed the weird empty EiaSettings thing which was a mistake on my end. I also re-added the `test_eia860_dependency` test with the same but opposite code from the `test_eia923_dependency`. 

I also had to update the `check_eia_dependencies` in `settings.py` to create default values for eia860 that were based off of whatever was in EIA923 but also what is actually available in 860. It was failing because it was trying to create a default suite of 860 data based on 923 data that included 2023 (which 860 doesn't have). Lmk if this feels like an ok fix.

cmgosnell (approved):
great updates to the settings!! wahoo 🤞🏻  that the ci does the things

PR #2938, opened 2023-10-12
dagsterification of ferc1_eia

cmgosnell:
# PR Overview
- [x] convert the table to a non-db asset
- [x] convert the inputs to the assets
- [x] convert the pudl_out method into a default magic in the rename (idk how this works i just replicated it 🤷🏻)
- [x] check for PKs
   - [x] there were some dupes bc of the null override file: dedupe the override file.
   - [x] there were more dupes bc there were records in both the null override file & the training data. they all seemed to be labeled with 1:m in the null sheet so I removed them from the training sheet
   - [x] there were even more dupes bc of a weird merge choice that i cleared up & made a comment about
- [x] add the table metadata in resources (added draft ###ed out)
- [x] add any new columns: `record_id_ferc1`
- [x] convert the non-db asset into an db asset
- [x] fixed two deprecation warnings bc they were not cute in my logs

# Questions
- do we want alllll of the columns from both `plant_parts_eia` and `plants_all_ferc1`? rn i just added em all... i think that's data marty but idk it is a lot. ANSWER: [all of em](https://github.com/catalyst-cooperative/pudl/pull/2938#issuecomment-1761818266)
- asset_group: it feels weird to make a whole new asset group for this one guy... any input on that? CURRENT ANSWER: a new group named 🥁 `ferc1_eia`. a group of 1... this seems like our current pattern. i don't love that tbh but i'm going to move forward w/ that and if we come up with a collective good idea around this we should standardize.
- NAMES: should i try to go with the _new names_ for this table??

# PR Checklist


src/pudl/analysis/classify_plants_ferc1.py
```diff
@@ -681,7 +681,7 @@ def fuel_by_plant_ferc1(
                 "fuel_type_code_pudl",
             ]
         )
-        .agg(sum)
+        .sum()
```
cmgosnell, line 684: deprecation warning # 1

src/pudl/metadata/classes.py
```diff
@@ -1466,7 +1466,7 @@ def format_df(self, df: pd.DataFrame | None = None, **kwargs: Any) -> pd.DataFra
                 and pd.api.types.is_integer_dtype(df[field.name])
             ):
                 df[field.name] = pd.to_datetime(df[field.name], format="%Y")
-            if pd.api.types.is_categorical_dtype(dtypes[field.name]):
+            if isinstance(dtypes[field.name], pd.CategoricalDtype):
```
cmgosnell, line 1469: deprecation warning # 2 (the loud one omigosh so many warnings shhhhh)

zaneselvans:
Hmm, I had thought that this was just going to be glue / crosswalk table that provided the necessary FERC & EIA columns for merging together data that had both FERC & EIA planty records.

src/pudl/package_data/glue/ferc1_eia_null.csv
cmgosnell, line 1: I was getting dupes in the pks of this table with only overridden records, which were coming from this file so i:

```
file = importlib.resources.files("pudl.package_data.glue") / "ferc1_eia_null.csv"
null_override_raw = pd.read_csv(file)
null_override_dedupe = null_override_raw.drop_duplicates(keep="first")
null_override_dedupe.to_csv(file, index=False)
```

cmgosnell:
> Hmm, I had thought that this was just going to be glue / crosswalk table that provided the necessary FERC & EIA columns for merging together data that had both FERC & EIA planty records.

this is my one question about the pr. i haven't been in y'all's convos about the intention for this table so idk what the intention was here.

I believe could pretty easily make a shinny glue table and a wide marty table.

src/pudl/metadata/classes.py
zaneselvans: Oh I fixed this one somewhere too. Yeah it was loud.

zaneselvans:
If the big wide table is what we've been generating, then I assume that's what RMI is depending on downstream, in which case we need to continue providing it in some form. How does this data currently get used?  It must have a bajillion columns.

Regardless of that, it also seems like the glue table would be a generally useful thing. You could use it to stick together FERC + EIA data in a variety of contexts without needing to read this monster table.  What columns would be part of the glue?

cmgosnell:
I added all of the extra columns a while back while working in the rmi-ferc1-eia repo because I kept going back and merging in more columns from the EIA or FERC data. If the whole point of this merge is to connect the financial data in FERC and the operational data in EIA... it seems like we should probably publish this big table. Especially because the EIA merge is a merge from the `plant_parts_eia`, which is not super hard but requires some understanding of the ppe.

I think the skinny table could just have: `["record_id_eia", "record_id_ferc1", "match_type"]` (match_type is the only truly new column from the matching process)

src/pudl/metadata/resources/glue.py
```diff
@@ -87,6 +87,148 @@
...
+    # via record linkage. It is reported
+    # FERC plant records are reported... kind of messily. In the same table there are records
+    # that are reported as whole plants, generators, collections of prime movers. So we have
+    # this heterogeneously reported collection of parts of plants in FERC1.
+    # EIA on the other hand is reported in a much cleaner way. The are generators with ids and
+    # plants with ids reported in *seperate* tables. What a joy. In plant_parts_eia, we've
+    # generated the EIA plant-parts. The EIA plant-parts generated records for various levels
+    # or granularities of plant parts.
+    # For each of the FERC1 plant records we want to figure out which EIA plant-parts record
+    # is the corresponding record.
```
zaneselvans, line 101: I think we should work on this description. It'll probably be long but this is a valuable and also complex table.

zaneselvans (approved):
I think we need to talk more about the structure of the glue.

src/pudl/analysis/ferc1_eia.py
```diff
@@ -881,7 +873,11 @@ def prettyify_best_matches(
             train_df,
             match_set=match_set,
         )
-
+    connects_ferc1_eia = (
+        Package.from_resource_ids()
+        .get_resource("ferc1_eia")
+        .enforce_schema(connects_ferc1_eia)
+    )
```
zaneselvans, line 880: I think this can be done with just the Resource rather than the Package:

```py
Resource.from_id("ferc1_eia").enforce_schema(connects_ferc1_eia)
```

src/pudl/package_data/glue/ferc1_eia_train.csv
zaneselvans, line 1: What's going on with all the changes to the training data here?

cmgosnell:
okay @zaneselvans  and i chatted about table design and landed on keeping the wide table because the `record_id_eia` is only linkable to the `plant_parts_eia` and making users interact with that table to get the EIA data is... probably not a good idea. This table is at the "level" of the ferc records which necessarily means it includes a variety of EIA ppe records (i.e. aggregations of EIA generators).

in an ideal world we would follow up with this table with an addition table that is at the level of EIA generators (or EIA generator-owners). it would be a 1:m association table with one FERC1 record being linked to many EIA generator records. and the FERC1 data could be proportionally allocated across those generators. we did a lot of that work over in [this repo](github.com/catalyst-cooperative/rmi-ferc1-eia/).

src/pudl/metadata/resources/glue.py
```diff
@@ -87,6 +87,134 @@
...
+                "opex_production_before_pumping",
+                "opex_pumped_storage",
+                "opex_pumping",
+                "total_fuel_cost_ferc1",
+                "total_mmbtu_ferc1",
+                "fuel_type_code_pudl_ferc1",
+                "plant_id_report_year",
+                "heat_rate_mmbtu_mwh_ferc1",
+            ],
+            "primary_key": ["record_id_eia", "record_id_ferc1"],
```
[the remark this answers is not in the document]
cmgosnell: i think that's the right call.

src/pudl/package_data/glue/ferc1_eia_null.csv
cmgosnell: i think it was probably a semi-sloppy conversion thing. its vv possible this came from the groups of temp excel files were we had labeled them in a one:many way (which ended up being used to build a solution around the one:many problem). and we just weren't checking.

src/pudl/package_data/glue/ferc1_eia_train.csv
cmgosnell: we've had a few solutions to one ferc record being associated with many ppe records over the time we've been developing the ppe/ferc_eia connection. These were stragglers where we mapped these one:many FERC records in the training data to the whole plant.

we've since come up with a nice solution for the one:many's but never removed them from the training data so we were getting duplicates.

src/pudl/metadata/resources/glue.py
cmgosnell: halp

zaneselvans:
On the Naming of Things, I do think that getting a more descriptive table name in here would be great.  Under the new naming convention what would it be?  What "data source" are we using for the composite analytical tables? `pudl`? Or nothing at all?  Would `out__yearly_plants_all_ferc1_plant_parts_eia_assn` be appropriate? Is there a better way to indicate what the two sides of the table are?  I guess it's not strictly an association table.

On the asset groups, to me it feels like `plant_parts_eia`, `mega_gens` and `ferc1_eia` are all part of the same constellation. Maybe just add this module to the list (currently of length 1) of modules that's used to build the plant parts asset group?

Should we create an issue to get the already-existing FERC1 data allocated to EIA generator-ownership slices table migrated over from the RMI repo and/or the coarse-grained primary-fuel based linkage for the remaining 20% of FERC1 records that aren't getting associated in the current process.

Do you want me to take a stab at a table description, and then you can tell me how it's wrong because I don't totally understand this table?

src/pudl/etl/__init__.py
```diff
@@ -73,6 +73,7 @@
     *load_assets_from_modules(
         [pudl.analysis.plant_parts_eia], group_name="plant_parts_eia"
     ),
+    *load_assets_from_modules([pudl.analysis.ferc1_eia], group_name="ferc1_eia_stuff"),
```
zaneselvans, line 76: What if we combined all the assets that are currently part of the `plant_parts_eia` group, and the new one from this group into a single `ferc1_eia_record_linkage` or `ferc1_eia_entity_matching` group, since that's what they're all aimed at conceptually, and that's more intelligible to an outsider than `plant_parts_list` and we anticipate integrating additional tables within that theme into the database from the work that's already been done in the RMI repo.

There's also clear dependencies between these big tables, and in development we're probably going to need to regenerate them together, or one at a time as we futz with things, and being able to display them all and work with them in a single asset group will be nice.

```py
*load_assets_from_modules(
    [
        pudl.analysis.plant_parts_eia,
        pudl.analysis.ferc1_eia,
    ],
    group_name="ferc1_eia_record_linkage",
)
```

src/pudl/metadata/classes.py
```diff
@@ -204,7 +204,7 @@ def __repr_args__(self) -> list[tuple[str, Any]]:
 """Non-empty :class:`str` with no trailing or leading whitespace."""
 
 SnakeCase = pydantic.constr(
-    min_length=1, strict=True, regex=r"^[a-z][a-z0-9]*(_[a-z0-9]+)*$"
+    min_length=1, strict=True, regex=r"^[a-z_][a-z0-9_]*(_[a-z0-9]+)*$"
```
cmgosnell, line 207: in order to use the new double underscore name I stole this line from [this pr](https://github.com/catalyst-cooperative/pudl/pull/2818/files#)

zaneselvans (approved):
Dazhong says the Zenodo fix CI should be done in about 15 minutes!

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -2,8 +2,117 @@
...
+                "generator_operating_year",
+                "operational_status_pudl",
+                "capacity_eoy_mw",
+                "fraction_owned",
+                "ownership_record_type",
+            ],
+        },
+        "sources": ["eia860", "eia923"],
+        "etl_group": "outputs",
+        "field_namespace": "eia",
```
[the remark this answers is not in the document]
cmgosnell: i'm sure it could have some gross composite key but i just copied this over when migrating it.

PR #2942, opened 2023-10-16
New zenodo api bandaid

### LINKED ISSUE #2939: Update datastore to read from Zenodo's new InvenioRDM API
On October 13th [Zenodo switched over to using InvenioRDM as their backend](https://help.zenodo.org/docs/about/whats-new/). This switch seems to have broken all of the URLs in all of the `datapackage.json` files in every archive we've ever made, rendering them no longer usable by our current software.

* [InvenioRDM API Documentation](https://inveniordm.docs.cern.ch/reference/rest_api_index/)
* [Barebones API example repository](https://github.com/inveniosoftware/docs-invenio-rdm-restapi-example)

In the immediate term, it would be great if we can just change the software to work with both the old and new link formats, so we don't *have* to update all of our archives right now. But then we should update the archivers to start creating archives with the new link format, and ideally also to use the new API for the work they do, but that may not be absolutely necessary, if the old API has been preserved -- which they say it should have been but... we'll see.

```[tasklist]
- [x] Update the datastore to construct and use a valid InvenioRDM file URLs when it encounters an old archive with an obsolete URL.
- [ ] https://github.com/catalyst-cooperative/pudl-archiver/issues/183
```

Valid file download links take the form:

```
https://zenodo.org/api/records/{record_id}/files/{filename}/content
```

It appears that an API key is no longer required to download data which is great!  You can do:

```bash
curl https://zenodo.org/api/records/8346646/files/datapackage.json/content
```

Downloads from these URLs appears to be relatively performant. Working from a GCP VM I got about 10 Mbit/s download speeds, which I think was similar to what we had with the direct storage links before. IIRC, we originally decided to use the direct links to the storage buckets because downloads were much faster than with the API constructed links.

jdangerx:
Closes #2939 .

* Use new record files API endpoint to get datapackage.json
* Add shim to re-construct legacy URLs that are in `datapackage.json` / pretend that we were using the "best, most stable URLs" the whole time
* remove access keys, yay

### DIFF
--- a/src/pudl/workspace/datastore.py
+++ b/src/pudl/workspace/datastore.py
@@ -10,6 +10,7 @@ from collections import defaultdict
 from collections.abc import Iterator
 from pathlib import Path
 from typing import Any, Self
+from urllib.parse import ParseResult, urlparse
 
 import datapackage
 import requests
@@ -38,7 +39,7 @@ class ChecksumMismatchError(ValueError):
 class DatapackageDescriptor:
     """A simple wrapper providing access to datapackage.json contents."""
 
-    def __init__(self, datapackage_json: dict, dataset: str, doi: str):
+    def __init__(self, datapackage_json: dict, dataset: str, doi: ZenodoDoi):
         """Constructs DatapackageDescriptor.
 
         Args:
@@ -56,7 +57,15 @@ class DatapackageDescriptor:
         res = self._get_resource_metadata(name)
         # remote_url is sometimes set on the local cached version of datapackage.json
         # so we should be using that if it exists.
-        return res.get("remote_url") or res.get("path")
+        resource_path = res.get("remote_url") or res.get("path")
+        parsed_path = urlparse(resource_path)
+        if parsed_path.path.startswith("/api/files"):
+            record_number = self.doi.lower().rsplit("zenodo.", 1)[-1]
+            new_path = f"/api/records/{record_number}/files/{name}/content"
+            new_url = ParseResult(**(parsed_path._asdict() | {"path": new_path}))
+
+            return new_url.geturl()
+        return resource_path
 
     def _get_resource_metadata(self, name: str) -> dict:
         for res in self.datapackage_json["resources"]:
@@ -257,7 +266,7 @@ class ZenodoFetcher:
             api_root = "https://zenodo.org/api"
         else:
             raise ValueError(f"Invalid Zenodo DOI: {doi}")
-        return f"{api_root}/deposit/depositions/{zenodo_id}"
+        return f"{api_root}/records/{zenodo_id}/files"
 
     def _fetch_from_url(self: Self, url: HttpUrl) -> requests.Response:
         logger.info(f"Retrieving {url} from zenodo")
@@ -274,9 +283,9 @@ class ZenodoFetcher:
         doi = self.get_doi(dataset)
         if doi not in self._descriptor_cache:
             dpkg = self._fetch_from_url(self._get_url(doi))
-            for f in dpkg.json()["files"]:
-                if f["filename"] == "datapackage.json":
-                    resp = self._fetch_from_url(f["links"]["download"])
+            for f in dpkg.json()["entries"]:
+                if f["key"] == "datapackage.json":
+                    resp = self._fetch_from_url(f["links"]["content"])
                     self._descriptor_cache[doi] = DatapackageDescriptor(
                         resp.json(), dataset=dataset, doi=doi
                     )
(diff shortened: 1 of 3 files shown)


test/unit/settings_test.py
```diff
@@ -263,7 +262,7 @@ def test_partitions_with_json_normalize(pudl_etl_settings):
 
 def test_partitions_for_datasource_table(pudl_etl_settings):
     """Test whether or not we can make the datasource table."""
-    ds = Datastore(local_cache_path=PudlPaths().data_dir)
+    ds = Datastore(local_cache_path=None)
```
zaneselvans, line 265: We don't really want to leave this change in place right?

src/pudl/workspace/datastore.py
```diff
@@ -56,7 +57,15 @@ def get_resource_path(self, name: str) -> str:
         res = self._get_resource_metadata(name)
         # remote_url is sometimes set on the local cached version of datapackage.json
         # so we should be using that if it exists.
-        return res.get("remote_url") or res.get("path")
+        resource_path = res.get("remote_url") or res.get("path")
+        parsed_path = urlparse(resource_path)
+        if parsed_path.path.startswith("/api/files"):
+            record_number = self.doi.lower().rsplit("zenodo.", 1)[-1]
+            new_path = f"/api/records/{record_number}/files/{name}/content"
+            new_url = ParseResult(**(parsed_path._asdict() | {"path": new_path}))
```
zaneselvans, line 65: See the response from Zenodo support about what the best pseudo-persistent URL to use is.  His suggestion does seem like the simplest possible URL, without attributes that seem dependent on any API specifics...  here it would be:

```py
new_path = f"records/{record_number}/files/{name}"
```

Given that we're not **really** using the API here and are instead constructing the URL we would expect to receive from the API, does it make sense to go with the non-API URL support suggested? Or is there a good reason to go with the expected API URL?

src/pudl/workspace/datastore.py
```diff
@@ -38,7 +39,7 @@ class ChecksumMismatchError(ValueError):
 class DatapackageDescriptor:
     """A simple wrapper providing access to datapackage.json contents."""
 
-    def __init__(self, datapackage_json: dict, dataset: str, doi: str):
+    def __init__(self, datapackage_json: dict, dataset: str, doi: ZenodoDoi):
```
zaneselvans, line 42: Ah good catch!

src/pudl/workspace/datastore.py
```diff
@@ -274,9 +283,11 @@ def get_descriptor(self: Self, dataset: str) -> DatapackageDescriptor:
         doi = self.get_doi(dataset)
         if doi not in self._descriptor_cache:
             dpkg = self._fetch_from_url(self._get_url(doi))
-            for f in dpkg.json()["files"]:
-                if f["filename"] == "datapackage.json":
-                    resp = self._fetch_from_url(f["links"]["download"])
+            dpkg.json()["entries"]
```
zaneselvans, line 286: This line seems like it might be a copy-paste oops that does nothing.

src/pudl/workspace/datastore.py
```diff
@@ -274,9 +283,11 @@ def get_descriptor(self: Self, dataset: str) -> DatapackageDescriptor:
...
+            self._descriptor_cache[doi] = self._descriptor_from
+            for f in dpkg.json()["entries"]:
+                if f["key"] == "datapackage.json":
+                    resp = self._fetch_from_url(f["links"]["content"])
```
zaneselvans, line 290: If our whole setup is contingent on being able to construct a path to a file given the filename and record ID, is there a reason to use the API to obtain the `datapackage.json`?

I guess if we want to verify the checksum on `datapackage.json` we need to get the checksum from Zenodo, since the file can't contain its own checksum.

zaneselvans (commented):
Is this PR really supposed to be going into `main` rather than `dev`?

Another simplifying change I think we can make with the new API is getting rid of the API keys / tokens entirely, since it seems like read-only operations are fine without them (at least, `curl` works without them which I don't think was the case before!)

src/pudl/workspace/datastore.py
jdangerx: My reason for doing this is that "keeping the datapackage.json around lets me not think too hard about how to construct a `DatapackageDescriptor`."

src/pudl/workspace/datastore.py
jdangerx: Pretending that we used this most-stable URL all along seems like a fine idea!

zaneselvans (approved):

PR #2949, opened 2023-10-18
Start EIA-176 pipelines: company data

davidmudrauskas:
# PR Overview

* This introduces the basic extraction components for EIA-176, starting with reading `company` data. We can extend this pattern to process all EIA-176 data, but that is not included in this PR.
* This also includes integration with Dagster, where the assets materialized successfully.

# PR Checklist


src/pudl/extract/csv.py
```diff
@@ -0,0 +1,153 @@
...
+import pandas as pd
+import sqlalchemy as sa
+
+from pudl import logging_helpers
+from pudl.workspace.datastore import Datastore
+
+logger = logging_helpers.get_logger(__name__)
+
+
+class CsvReader:
```
davidmudrauskas, line 17: This is a simple first iteration of `FercDbfReader`, but for CSVs, attempting to reuse and establish consistent patterns. Some of the methods are copy-pasted in this draft and should be extracted to common functionality or otherwise reconciled to ward against complexity and drift.

I also could have used the objects in the `excel` module as the basis, which have their own patterns. But they appear to handle more complexity than I'm seeing at this point for CSVs. The overlap between the functionality needed for CSVs and DBFs appears stronger.

src/pudl/extract/csv.py
```diff
@@ -0,0 +1,153 @@
...
+        """Read the data from the CSV source and return as dataframes."""
+        logger.info(f"Extracting {filename} from CSV into pandas DataFrame.")
+        zipfile = self._cache_zipfile()
+        with zipfile.open(filename) as f:
+            # TODO: Define encoding
+            df = pd.read_csv(f)
+        return df
+
+
+class CsvExtractor:
```
davidmudrauskas, line 60: This implementation was based on `FercDbfExtractor`. So far I've been able to slim it down some compare to that class.

src/pudl/extract/eia176.py
```diff
@@ -0,0 +1,47 @@
+"""Extract EIA Form 176 data from CSVs.
+
+The EIA Form 176 archive also contains CSVs for EIA Form 191 and EIA Form 757.
```
davidmudrauskas, line 3: There's one source for the three forms, so I've coupled processing here.

davidmudrauskas:
I have an extractor I can run locally that populates the company table associated with EIA-176 (screenshot). Next I'll make sure there's sufficient test coverage and graduate this out of a draft so it doesn't become too big. I'll aim to do that over the next week. I think the next iterations would be making sure to integrate with Dagster properly, populating the tables for other data, and any refactoring around the repeated patterns between the CSV-related objects and DBF-related objects.

Screenshot 2023-10-29 at 8 39 56 PM

test/unit/extract/csv.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def get_csv_table_schema():
+    table_schema = CsvTableSchema(TABLE_NAME)
+    table_schema.add_column(COL1_NAME, COL1_TYPE)
+    table_schema.add_column(COL2_NAME, COL2_TYPE, COL2_SHORT_NAME)
+    return table_schema
+
+
+def test_csv_table_schema_get_columns():
```
davidmudrauskas, line 25: @jdangerx FYI I didn't see exactly the pattern you all were using for unit test coverage of methods so I've started tackling it like this. It's relatively easy for me to knock out tests for each new method (using mocks as needed), so I'll proceed that way unless told otherwise.

src/pudl/extract/csv.py
jdangerx: So the `CsvReader` provides:

* a list of tables that we know how to extract from an archive
* a `CsvArchive` object that we can get actual table data from

Is that right?

If so, what do you think about merging the `Archive` and the `Reader` into one class that handles "get a table from an archive that contains CSV files?"

src/pudl/extract/eia176.py
jdangerx: As in, the same ZIP file contains all three forms, but 191/757 aren't being extracted here, yet? That seems fine to me - do you think the empty 191/757 extractor files are useful from a documentation standpoint?

src/pudl/extract/csv.py
jdangerx: Ah, here's something confusing about our system:

* FERC DBF data gets converted to tables `ferc{form_number}.sqlite` outside of the normal ETL process - I don't know why, actually. FERC XBRL gets converted to a separate table also, that's because we don't know what the schema is going to be until we've read in the dang XML file.
* EIA data gets extracted to `raw_eia176__....` pickled dataframes in the Dagster asset storage directory - see `src/pudl/extract/eia860.py` for how that gets executed. They then feed into `_core_eia176__...` assets which also are pickled, which *finally* feed into `output_eia176__...` assets which get written to `pudl.sqlite` ( see `src/pudl/transform/eia860.py` and `src/pudl/output/eia860.py` for details).

Long story short, we don't need to worry about where the data is getting persisted on disk at this juncture, we just need to pull out a bunch of dataframes. I think if you want to test the `CsvExtractor` through testing the `Eia176CsvExtractor`, you'd want to make a dagster asset in ``extract/eia176.py` that looks something like....


```python
@asset
def raw_eia176__company(context):
    extractor = Eia176CsvExtractor(context.resource.datastore)
    # ... get the company table as dataframe
    # return that dataframe
```

And then you should be able to just test `raw_eia176_company` like a normal function, though you would need to mock out the `context.resource.datastore`.

jdangerx (changes requested):
Thanks for the easy-to-read code + accompanying tests! I think we led you a bit astray with the FERC DBF extractors - hopefully my comment about using the EIA 860 patterns makes sense. If not, I'm happy to get on a call with you and hash stuff out!

src/pudl/extract/csv.py
davidmudrauskas: Yeah I was thinking similar. Will do!

src/pudl/extract/eia176.py
davidmudrauskas: I was planning to call the 191 and 757 extractors from within the 176 extractor module, to mirror and communicate the coupling in the data source (unless we decide to start grouping and naming pipelines in a way that's abstracted from specific EIA forms). I'll introduce unimplemented extractors in my next iteration at the very least, and the calls to them, so the structure and intention is better communicated without just adding noise with empty files.

src/pudl/extract/csv.py
davidmudrauskas: That's helpful, thanks. I'll work on an iteration along those lines and get something surfaced by Monday.

jdangerx: Sweet! Just an FYI - a bunch of us are gonna be at a conference Monday and Tuesday. But ping on Slack if you're ready and I'll pull some time to take a look!

src/pudl/extract/csv.py
```diff
@@ -0,0 +1,106 @@
...
+        ds = context.resources.datastore
+        return extractor_cls(ds).extract()
+
+    return op(
+        required_resource_keys={"datastore", "dataset_settings"},
+        name=f"extract_single_{name}_year",
+    )(extract)
+
+
+def raw_df_factory(extractor_cls: type[CsvExtractor], name: str) -> AssetsDefinition:
```
davidmudrauskas, line 91: This follows the pattern established in [excel.raw_df_factory](https://github.com/catalyst-cooperative/pudl/blob/b50ea1d8b8fd5b15e888b8fdb0fea7047cb9e7fa/src/pudl/extract/excel.py#L471), i.e., what we do for extracting EIA-860 in Dagster. Looks like the preexisting logic is covered by the Dagster nightly tests and I'm inclined to rely on the same here.

src/pudl/extract/eia176.py
davidmudrauskas: I ended up remove these stub files for EIA forms 191 and 757. They won't need to be coupled in the way I thought and they're only adding noise at this point.

src/pudl/package_data/eia176/table_file_map.csv
```diff
@@ -0,0 +1,2 @@
+table,filename
+company,all_company_176.csv
```
davidmudrauskas, line 2: There will be more tables to add here once we get the simple piece for `company` going.

test/integration/etl_test.py
```diff
@@ -68,6 +68,17 @@ def test_ferc1_xbrl2sqlite(ferc1_engine_xbrl, ferc1_xbrl_taxonomy_metadata):
     )
 
 
+class TestCsvExtractor:
+    """Verify that we can lead CSV files as provided via the datastore."""
+
+    def test_extract_eia176(self, pudl_datastore_fixture):
```
davidmudrauskas, line 74: Followed the preexisting pattern for this (e.g., per `test_extract_eia860`). I was able to quickly and successfully run this spot check in isolation with
```
 pytest test/integration/etl_test.py -k test_extract_eia176
```

davidmudrauskas:
I was able to slim down the CSV extraction classes and tests a bunch. I also materialized the assets successfully in Dagster (screenshot).

Screenshot 2023-11-16 at 3 59 49 PM

src/pudl/extract/csv.py
```diff
@@ -0,0 +1,106 @@
...
+from dagster import AssetsDefinition, OpDefinition, graph_asset, op
+
+import pudl.logging_helpers
+from pudl.workspace.datastore import Datastore
+
+logger = pudl.logging_helpers.get_logger(__name__)
+
+
+class CsvExtractor:
+    """Generalized class for extracting dataframes from CSV files.
```
jdangerx, line 15: So this CsvExtractor class requires subclassing to set the `DATASET` - that seems a little clunky. If we can pass the dataset name to `__init__`, we'll have a `CsvExtractor` class which exposes:

```python
extractor = CsvExtractor(datastore, dataset_name)
extractor.read_source(filename) # get one file
extractor.extract() # get all files
```

Which seems like a nice generic CSV extraction interface that doesn't require subclassing to read a variety of different collections of CSV files.

I think we might want to tweak that API a bit to expose table-level operations:

```python
extractor.extract_one(table_name) # pd.DataFrame
extractor.extract_all() # dict[str, pd.DataFrame]
```

Because then we can read multiple tables' files in, in parallel - each table could have its own asset like

```python
@asset 
def raw_eia176__table_name(context):
    ...
    extractor.extract_one(table_name)
```

Which is pretty straightforward to factory-ize if you want to make a bunch of these assets.

In the event we need to read & combine multiple files for a single table (like we see in EIA 860), we can turn that simple asset above into a graph-backed asset. But for EIA176 that seems like overkill.

src/pudl/extract/csv.py
jdangerx: EIA 860 is a little more complicated than EIA 176, since each table corresponds to a number of files spread across a number of different zip files (see `src/pudl/package_data/eia860/file_map.csv` for a look into the mess...)

The upshot is, I think this `raw_df_factory` and `extractor_factory` can be combined into one layer of abstraction. See above for more details.

src/pudl/extract/dbf.py
```diff
@@ -159,7 +159,7 @@ def get_table_schema(self, table_name: str) -> DbfTableSchema:
         table_columns = self.get_db_schema()[table_name]
         dbf = self.get_table_dbf(table_name)
         dbf_fields = [field for field in dbf.fields if field.name != "_NullFlags"]
-        if len(table_columns) != len(table_columns):
+        if len(dbf_fields) != len(table_columns):
```
jdangerx, line 162: Ha! Good catch 🕵️

src/pudl/extract/csv.py
jdangerx: I think there's also an argument for passing in the table -> file(s) map and the path to file as init params, and then using the datastore at the call site - this lets people use the extractor to explore data sets that we haven't actually integrated into our upstream work yet, and lets you pass stuff in for testing without extensive patching. Could be something like

```
class CsvExtractor:
    def __init__(self, path: pathlib.Path | zipfile.Path, table_files_map: dict[str, list[str]]):
        ...
    @classmethod
    def from_resource(cls, datastore: Datastore, resource_id: str):
        # existing logic to get zipfile + table/file map
        return cls(...)
```

What do you think?

jdangerx (changes requested):
Sweet! I ran this in Dagster and got a very reasonable looking dataframe out when I then ran `defs.load_asset_value`. Our first gas dataset!

I think we could save you from some annoying subclassing in future table extractions + provide a slightly nicer generic CSV extraction API, but I could definitely be missing something here so let me know what you think!

I could be convinced to merge this with no changes - my core worry is that the `CsvExtractor` interface doesn't cleanly reflect its purpose (turn a zipfile + table -> file map into a table name -> dataframe map), so I'm happy to merge once that worry is assuaged.

src/pudl/extract/csv.py
davidmudrauskas: > So this CsvExtractor class requires subclassing to set the DATASET - that seems a little clunky. If we can pass the dataset name to __init__

Meant to comment on this. I actually started using dataset as a construction parameter but stopped because I was introducing a competing pattern, e.g., vs `FercDbfExtractor`'s inheritance tree. I'll proceed with parameterizing over inheritance now that someone else is also inclined.

> I think there's also an argument for passing in the table -> file(s) map and the path to file as init params, and then using the datastore at the call site

I don't know what the datastore would be used for in that case, since right now it's only to get the zipfile path. But generally you're talking about providing a class that lets a user/client point to any zip file and get dataframes out of it based on a `table_files_map`? Right now the zipfile path and the table-file(s) map are coupled on the dataset name, e.g., `eia176`. Decoupling opens a wider window for invalid combos, i.e., table filenames that do not exist in the zip archive, but yeah, it would be nice to be able to develop against data without needing the source published on Zenodo. I'll take a pass in that direction.

src/pudl/extract/csv.py
```diff
@@ -0,0 +1,72 @@
...
+from zipfile import ZipFile
+
+import pandas as pd
+
+import pudl.logging_helpers
+
+logger = pudl.logging_helpers.get_logger(__name__)
+
+
+def open_csv_resource(dataset: str, base_filename: str) -> DictReader:
```
davidmudrauskas, line 13: This can be moved to an even more general space at some point but I didn't want to introduce another moving part in this PR.

src/pudl/extract/eia176.py
```diff
@@ -0,0 +1,26 @@
...
+
+from dagster import asset
+
+from pudl.extract.csv import CsvExtractor, get_table_file_map
+
+DATASET = "eia176"
+
+
+@asset(required_resource_keys={"datastore"})
+def raw_eia176__company(context):
```
davidmudrauskas, line 14: I wrote a concrete asset like this for now, after trying to get a factory pattern down. Happy to learn more about the Dagster components and write a factory at some point.

davidmudrauskas:
I was able to materialize in Dagster and my updated integration test passed for this latest iteration.

src/pudl/extract/csv.py
jdangerx: Yeah, I could see this moving to `pudl.helpers` or something but this is a fine place for it.

src/pudl/extract/csv.py
```diff
@@ -0,0 +1,72 @@
...
+        """Create a new instance of CsvExtractor.
+
+        This can be used for retrieving data from CSV files.
+
+        Args:
+            zipfile: zipfile object containing source files
+            table_file_map: map of table name to source file in zipfile archive
+        """
+        self._zipfile = zipfile
+        self._table_file_map = table_file_map
```
jdangerx, line 52: If you're worried about the table file map and the zipfile not matching up, you could validate that the files in `table_file_map` do show up in `zipfile.namelist()`. Doesn't have to happen in this pass, but should probably happen at *some* point.

jdangerx (approved):
🎉 looks great! Definitely a good point re: trying to make sure we can't represent illegal states.

jdangerx:
The integration tests failures appear to be due to Zenodo flakiness + outdatedly including `/draft` in the download URLs they return. Should get handled if we merge https://github.com/catalyst-cooperative/pudl-archiver/pull/192 and recreate some data packages on Zenodo with better URLs.

The zenodo-cache-sync failed on git pull? Which seems funky. And the notification failed because we're missing some Slack webhook secret, which I can set up.

jdangerx:
@davidmudrauskas  I think if you merge `dev` one more time, we should be able to try re-running the integration tests. Sorry that took so long 😬 

We also took a look at the zenodo-cache-sync errors from a different PR - the root cause is some credential stuff that isn't particularly important to fix now, so we can just ignore those checks.

davidmudrauskas:
FYI `test_filter_for_freshest_data` failed in my last build's unit tests:

```
Unreliable test timings! On an initial run, this test took 558.62ms, which exceeded the deadline of 400.00ms, but on a subsequent run it took 47.77 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.

You can reproduce this example by temporarily adding @reproduce_failure('6.91.0', b'AXicY2BgYGBhYFjAwMDIzMDAycjAzMB4moGRkQEOgEwglx1IMTGC1DCygIRsgTKrQICR4R0DszgjEwODD7MUyKgjLiD1UQxIAABbSwgI') as a decorator on your test case
== 1 failed, 284 passed, 1 skipped, 9 xfailed, 8 warnings in 71.09s (0:01:11) ==
```

It did fine locally, so hopefully it's more consistent and the above was an anomaly. Running a new build with latest `dev`.

jdangerx:
I think you're good to merge! Might need to pull our latest `dev` into your branch. I see a button for me to do it but not sure if that will mess with your flow: 

image

src/pudl/etl/__init__.py
```diff
@@ -34,6 +34,7 @@
 default_assets = (
     *load_assets_from_modules([eia_bulk_elec_assets], group_name="core_eia_bulk_elec"),
     *load_assets_from_modules([epacems_assets], group_name="core_epacems"),
```
davidmudrauskas, line 36: Captured the recent addition of the `core_` prefix here in the resolution of the last merge conflict.

zaneselvans:
Hey @davidmudrauskas I'm sorry I didn't realize that when I got rid of the `dev` branch it would close all the PRs from forks of the repo. Can you re-open this against the `main` branch?

zaneselvans:
Hey @davidmudrauskas I'm sorry I didn't realize that when I got rid of the `dev` branch it would close all the PRs from forks of the repo. Can you re-open this against the `main` branch?

PR #2953, opened 2023-10-19
Have-nightly-builds-output-cache-keys

ggurjar333:
# PR Overview

# PR Checklist


.github/workflows/tox-pytest.yml
```diff
@@ -99,6 +99,21 @@ jobs:
...
+        run: |
+          if [ -f commit_hash.txt ] && [ -f sha256_hash.txt ]; then
+            CACHED_COMMIT_HASH=$(cat commit_hash.txt)
+            CACHED_SHA256_HASH=$(cat sha256_hash.txt)
+            if [ "$CACHED_COMMIT_HASH" = "$GITHUB_SHA" ] && [ "$CACHED_SHA256_HASH" = "$SHA256_HASH_OF_DB" ]; then
+              echo "Cache is up to date. Skipping download."
+            else
+              echo "Cache is outdated. Downloading a new DB."
+              # Download the PUDL DB here
+            fi
```
e-belfer, line 112: Probably the source of your syntax error @ggurjar333

.github/workflows/tox-pytest.yml
```diff
@@ -99,6 +99,21 @@ jobs:
           which sqlite3
           sqlite3 --version
 
+      - name: Check Cache and Download PUDL DB
```
zaneselvans, line 102: This isn't one of the repositories where we need to download and cache a full copy of the PUDL DB to use for other analysis.  This is the repository that produces the full PUDL DB itself.  So this section should be removed.

Something like this would be needed in any other repository that wants to have a fresh cached version of the PUDL DB available for use in CI, like the `catalyst-cooperative/ccai-entity-matching` repository.

The cache check would need to be integrated with GitHub's caching system, and check the hash of the currently cached version of the database(s) against the hash that's been published in the nightly build outputs.

docker/gcp_pudl_etl.sh
```diff
@@ -19,6 +26,18 @@ function authenticate_gcp() {
 }
 
 function run_pudl_etl() {
+    if [ -f commit_hash.txt ] && [ -f sha256_hash.txt ]; then
```
zaneselvans, line 29: This script runs inside the Docker container that produces the PUDL DB, so it doesn't make sense for it to be checking whether there's a cached copy to download.

docker/gcp_pudl_etl.sh
```diff
@@ -89,6 +108,8 @@ run_pudl_etl 2>&1 | tee $LOGFILE
 # Notify slack if the etl succeeded.
 if [[ ${PIPESTATUS[0]} == 0 ]]; then
     notify_slack "success"
+    echo $COMMIT_HASH > commit_hash.txt
+    echo $SHA256_HASH > sha256_hash.txt
```
zaneselvans, line 112: This is the only thing that really needs to happen in the script -- once the build is complete, record both of these hashes and make sure they get distributed with the other build outputs.

Because there are several databases being produced, the name of the file recording the hash of the database should indicate which database it applies to, and while we're doing this we might as well produce hashes for all of the SQLite DBs and the big single-file EPA CEMS Parquet output.

zaneselvans (changes requested):

.github/workflows/tox-pytest.yml
ggurjar333: Ok

docker/gcp_pudl_etl.sh
ggurjar333: the commit hash of the current Git branch to a text file named commit_hash.txt and the SHA256 hash of the PUDL SQLite output to a text file named sha256_hash.txt.

ggurjar333:
Please review @zaneselvans @e-belfer

ggurjar333:
@bendnorman any suggestions what I should do on this?

PR #2964, opened 2023-10-20
Improve flexibility for publishing options

rousik:
This change slightly improves upon the flexibility of publishing outputs to remote destinations.
1. `PUDL_GCS_OUTPUT` env variable is introduced to `gcp_pudl_etl.sh` runner script to possibly override the destination used to publish stuff.
2. EtlSettings can load publish_destinations from `PUDL_PUBLISH_DESTINATIONS` env variable to immediately ship outputs after ETL finishes.

Option (2) should eventually replace the logic contained within (1), once we integrate data validation into the ETL or decouple it and do it later on.


.github/workflows/build-deploy-pudl.yml
```diff
@@ -110,6 +115,7 @@ jobs:
             --container-env AWS_ACCESS_KEY_ID=${{ secrets.AWS_ACCESS_KEY_ID }} \
             --container-env AWS_SECRET_ACCESS_KEY=${{ secrets.AWS_SECRET_ACCESS_KEY }} \
             --container-env AWS_DEFAULT_REGION=${{ secrets.AWS_DEFAULT_REGION }} \
+            --container-env PUDL_GCS_OUTPUT=gs://nightly-build-outputs.catalyst.coop/${{ env.COMMIT_TIME }}-${{ env.ACTION_SHA}-${{ env.COMMIT_BRANCH }}
```
jdangerx, line 118: Nice to have the time, too, a bit more useful information than just the SHA!

docker/gcp_pudl_etl.sh
```diff
@@ -89,8 +95,12 @@ run_pudl_etl 2>&1 | tee $LOGFILE
 # Notify slack if the etl succeeded.
 if [[ ${PIPESTATUS[0]} == 0 ]]; then
     notify_slack "success"
+    copy_outputs_to_gcs
```
jdangerx, line 98: In the old version we copy to GCS regardless of success, so we can see the failed outputs etc. - do you think that behavior still makes sense?

jdangerx (changes requested):
Nice and clean :) I just wonder if we want to keep the upload in the general cleanup function vs. trigger it on build success.

docker/gcp_pudl_etl.sh
rousik: Yes, indeed it does as was demonstrated in the recent rounds of failures.

rousik:
Please take another look, I have retained the original behavior of copying to GCS after the ETL finishes, regardless of the outcome.

rousik:
I have merged `dev` branch and also added empty `success` file to output directory when ETL finishes successfully. That way we should be able to tell which builds are successful by looking for that file.

docker/gcp_pudl_etl.sh
```diff
@@ -124,4 +134,5 @@ else
     notify_slack "failure"
 fi
 
+copy_outputs_to_gcs
```
zaneselvans, line 137: Do we want this to be the very last thing that happens? Or do we want to copy the outputs to GCS  before we cleanup the output directory for distribution? (removing Datasette's `metadata.yml`, the partitioned CEMS parquet files, etc.) so that we still have that information if we need it for debugging?  Seems like we probably want to do the gzipping before shoving it into the nightly build bucket.

docker/gcp_pudl_etl.sh
```diff
@@ -109,6 +116,9 @@ if [[ $ETL_SUCCESS == 0 ]]; then
     ETL_SUCCESS=${PIPESTATUS[0]}
```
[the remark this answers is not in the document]
rousik: If we replace `success` file with, say, json file that has some runtime statistics from how the data was generated, would we want to include it in distributed files, or also not?

docker/gcp_pudl_etl.sh
rousik: This retains current behavior in that we used to copy things to nightly-buckets as part of `vm_shutdown`. Happy to move this to earlier stage if we think it's a good idea.

docker/gcp_pudl_etl.sh
```diff
@@ -2,6 +2,8 @@
 # This script runs the entire ETL and validation tests in a docker container on a Google Compute Engine instance.
 # This script won't work locally because it needs adequate GCP permissions.
 
+: "${PUDL_GCS_OUTPUT:=gs://nightly-build-outputs.catalyst.coop/$ACTION_SHA-$GITHUB_REF}"
```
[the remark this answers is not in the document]
rousik: `:=` sets env variable if it's unset, i.e. provides default value (see https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)

So, with this change, the runner will set env variable and direct outputs to new paths and this line ensures that when it's run without that env variable set, current behavior remains unchanged.

docker/gcp_pudl_etl.sh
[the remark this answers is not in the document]
zaneselvans: I think a little build metadata file would be worth including. We're distributing the `pudl-etl.log` file. It would kind of be a structured accessory to that output.

docker/gcp_pudl_etl.sh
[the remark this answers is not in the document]
zaneselvans: Ah okay. That makes sense. I tried searching for this information but small strings of puncutation are bad search terms.

Is the leading colon `:` at the beginning of the line doing something?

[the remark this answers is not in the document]
rousik: Ah,  yes. I had to figure that out myself, apparently it cleans up the outputs: https://aplawrence.com/Basics/leading-colon.html

docker/gcp_pudl_etl.sh
```diff
@@ -27,26 +29,26 @@ function run_pudl_etl() {
...
         --etl-settings $PUDL_SETTINGS_YML \
-        --live-dbs test/integration test/unit && \
-    pytest \
+        --live-dbs test/integration test/unit \
+    && pytest \
         -n auto \
         --gcs-cache-path gs://internal-zenodo-cache.catalyst.coop \
         --etl-settings $PUDL_SETTINGS_YML \
         --live-dbs test/validate
+    && touch ${PUDL_OUTPUT}/success
```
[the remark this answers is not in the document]
rousik: Yes

zaneselvans (approved):
I merged `dev` in and added a comment to explain the bash conditional environment variable assignment magic.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -83,6 +85,11 @@ jobs:
       - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@v1
 
+      - name: Determine commit information
+        run: |-
```
jdangerx, line 89: TIL about the [block chomping operator](https://yaml.org/spec/1.2.2/#8112-block-chomping-indicator).  Does GHA complain when there's an extra newline at the end here?

docker/gcp_pudl_etl.sh
```diff
@@ -109,6 +120,9 @@ if [[ $ETL_SUCCESS == 0 ]]; then
     ETL_SUCCESS=${PIPESTATUS[0]}
 
     # Dump outputs to s3 bucket if branch is dev or build was triggered by a tag
+    # TODO: this behavior should be controlled by on/off switch here and this logic
```
jdangerx, line 123: Agree - having the action just pass in a switch would be nice. I think at some point we should replace this whole nightly build harness with a Python script that's more robust and that would be a nice time to fix this too.

jdangerx (approved):

PR #2996, opened 2023-10-31
Set up Cloud SQL Postgres database for dagster storage

bendnorman:
# PR Overview

This PR sets up a Cloud SQL Postgres database for dagster storage in our nightly builds. This will hopefully prevent the dagster backend storage database from locking up and killing our builds.

Changes:
- Creates a Cloud SQL Postgres console with 2 CPUs and 8 GB of RAM. @jdangerx and I decided handling the instance lifecycle from within the container might cause some jankyness so we will keep the instance running 24/7 which will cost [about $101 a month](https://cloud.google.com/products/calculator/#id=62b1a523-8911-4b66-ac56-a5e7a4b5cbc3). We will explore running postgres within our VM container to save cost and the overhead of managing an external db #3003.
- Adds logic in `docker/gcp_pudl_etl.sh` to start and stop the `dagster-storage` Cloud SQL instance and create and delete a `dagster-storage-$ACTION_SHA-$GITHUB_REF` database. This way, we have a fresh database for each nightly build run and won't incur Cloud SQL instance charges when we aren't running a nightly build.
- Add static external IP addresses to the `pudl-deployment-tag` and `pudl-deployment-dev` VMs so they can be whitelisted by the Cloud SQL instance.
- Points the dagster deployment to the Postgres database
- Passes the `PUDL_SETTINGS_YAML` to the container. I think this var was set manually in the GCP Console.

# PR Checklist


.github/workflows/build-deploy-pudl.yml
```diff
@@ -85,6 +85,8 @@ jobs:
 
       # Deploy PUDL image to GCE
       - name: Deploy
+        env:
+          DAGSTER_PG_PASSWORD: ${{ secrets.DAGSTER_PG_PASSWORD }}
```
jdangerx, line 89: In the future we should probably store this secret in Google Secret Manager instead - but this is fine for now. We need to get our nightly builds working consistently again, so there's urgency. And, I think we're building up momentum to revamp the infra completely - what's one more piece of jankiness at this juncture?

.github/workflows/build-deploy-pudl.yml
```diff
@@ -110,6 +112,11 @@ jobs:
             --container-env AWS_ACCESS_KEY_ID=${{ secrets.AWS_ACCESS_KEY_ID }} \
             --container-env AWS_SECRET_ACCESS_KEY=${{ secrets.AWS_SECRET_ACCESS_KEY }} \
             --container-env AWS_DEFAULT_REGION=${{ secrets.AWS_DEFAULT_REGION }} \
+            --container-env DAGSTER_PG_USERNAME="postgres" \
+            --container-env DAGSTER_PG_PASSWORD="$DAGSTER_PG_PASSWORD" \
+            --container-env DAGSTER_PG_HOST="104.154.182.24" \
```
jdangerx, line 117: It might be nice to run Postgres in the container itself, so we don't have to do all the weird Cloud SQL lifecycle management stuff.

On the other hand, it is weird to run your application and database in the same container. But gcloud compute only supports one container per VM (unless you want to run docker compose manually on the VM). So I'm OK with the Cloud SQL stuff for now as well.

docker/gcp_pudl_etl.sh
```diff
@@ -4,6 +4,16 @@
...
 set -x
 
+function setup_dagster_storage() {
+    SHORT_GITHUB_REF=$(git rev-parse --short $GITHUB_REF)
+    export DAGSTER_PG_DB="dagster-storage-$SHORT_GITHUB_REF"
+
+    # Start dagster-storage Cloud SQL instance
+    gcloud sql instances patch dagster-storage --activation-policy=ALWAYS
+    # Create database
+    gcloud sql databases create $DAGSTER_PG_DB --instance=dagster-storage
```
jdangerx, line 14: 1. Does this wait for the database to spin up before returning?
2. Does create throw an error if the DB already exists? Do we expect the DB to persist data over runs, or do we want to drop everything between runs? I see we delete the database below, but what happens if we error out and get interrupted before we delete the database?

jdangerx (changes requested):
This is fine, though lots of aspects of the postgres setup aren't actually in this PR.

I have some blocking questions about the lifecycle management stuff.

Separately, if we're starting to use Cloud SQL to store our Dagster state, it feels like we should talk about whether we want to keep this batched execution model or if we want to run Dagster as a persistent service that we trigger job runs on. I'll make a Discussion about it.

.github/workflows/build-deploy-pudl.yml
bendnorman: You're saying pass secrets from Google Secret Manager directly to the VM instead of storing GCP secrets in Github Actions and passing them to the VM?

.github/workflows/build-deploy-pudl.yml
bendnorman: Based on my experience, it seems like it's best to keep the app and database containers separate. I went with Cloud SQL because it's what [dagster recommends](https://docs.dagster.io/deployment/guides/gcp#using-cloud-sql-for-run-and-event-log-storage) and it seemed simpler than figuring out how to launch multiple containers.

I guess it's a little strange to spin the database up and down but I don't want to pay for the time it goes unused.

docker/gcp_pudl_etl.sh
bendnorman: 1. I'll double check the database is created before returning.
2. I figured it was fine to delete the database even if the build fails because the dagster event logs aren't actually that helpful when debugging build failures.

bendnorman:
I removed the Cloud SQL lifecycle logic from the `gcp_pudl_etl.sh` script.

jdangerx (approved):

PR #2999, opened 2023-11-01
Fix calculating the report_date in demand_hourly_pa_ferc714

rousik:
For some reason `report_date` could be missing for some number of records (0.7%). This addresses that problem by changing the way we align the reports to year boundaries.

### DIFF
--- a/src/pudl/metadata/fields.py
+++ b/src/pudl/metadata/fields.py
@@ -3005,7 +3005,12 @@ FIELD_METADATA_BY_RESOURCE: dict[str, dict[str, Any]] = {
                     "Pacific/Honolulu",
                 ]
             }
-        }
+        },
+        "report_date": {
+            "constraints": {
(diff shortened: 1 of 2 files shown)


zaneselvans (approved):
Did you happen to run the full FERC-714 ETL locally for all years with these changes?

It looks like there were `NaT` values included in the 2020 data which got processed by the integration tests in CI, and it passed, so it seems like it will probably work in general.  🤞🏼

rousik:
> Did you happen to run the full FERC-714 ETL locally for all years with these changes?
> 
> It looks like there were `NaT` values included in the 2020 data which got processed by the integration tests in CI, and it passed, so it seems like it will probably work in general. 🤞🏼

I have ran the affected table locally with dagster with `etl_fast.yml`. I'm expecting this to be low risk WRT possibility of this breaking on additional data.

Issue #3001, opened 2023-11-01
Run archiver for EIA 923 2022 final release data


e-belfer:
Run archive to get final release EIA 923 data.

zaneselvans:
# Notes on review of draft archive at https://zenodo.org/uploads/10064652

Overall this looks very good. Just a few small ergonomics issues.

I ran `frictionless validate datapackage.json` and the only errors I got were because the URLs don't (yet) point to valid publicly accessible files.

## `datapackage.json`
- The `datapackage.json` file is all one line, not pretty printed, making it extremely hard for a human to parse. Previously we were outputting them with an indent of 4 spaces to make them both human and machine readable.
- The Catalyst Cooperative collaborator has a `null` ORCID which has caused problems in the past. I think we should be not including this key if the value is NA.
- The resources don't seem to have any kind of rational ordering in the `datapackage.json`. Not technically a bug, but also not super ergonomic if someone is trying to look at the metadata. Probalby sorting by `name` or `title` would work across the board on all archives.
- It looks like the US Government has changed its open data license ID / link.  The one we've been using (https://www.usa.gov/government-works) now goes to a scary page about US Govt. Copyrights. Based on [this page](https://resources.data.gov/open-licenses/) it looks like the correct license URL is now http://www.usa.gov/publicdomain/label/1.0/ which... also unfortunately resolves to that scary page about government copyright. 🙄 

## Uploaded Files
- File sizes look reasonable.
- The files are ordered in the record using the same random / arbitrary ordering that shows up in the list of resources in `datapackage.json` so hopefully sorting them earlier in the process will also ensure that they're listed in a human readable order here.

## Zenodo metadata
- List of creators / contributors etc. in the Zenodo metadata is different from in the `datapackage.json` and probably inaccurate, since they include Steve. Might want to synchronize those and just have it refer to Catalyst like the `datapackage.json` does? Or include all current members.

e-belfer:
I'm opening a bug report for the ORCID, since this issue predates the migration to Zenodo and I think it's out of scope for the moment: https://github.com/catalyst-cooperative/pudl/issues/3005

The reordering of files is tracked by this issue: https://github.com/catalyst-cooperative/pudl/issues/3006

e-belfer:
Published here: https://zenodo.org/records/10067550

Issue #3000, opened 2023-11-01
Run archiver for EIA 860 2022 final release data


e-belfer:

zaneselvans:
# Notes on reviewing draft archive
Archive at https://zenodo.org/uploads/10064757

Ran `frictionless validate datapackage.json` and got the same (non-) issues as in the EIA-923.

## Uploaded files
- Same random ordering issue as with 923.
- File sizes look reasonable.

## Zenodo Metadata
- Same issue of out-of-date creators on the archive. Also makes it hard for a human to see if all files are present.
- Otherwise same issues as mentioned in the new 923 archive #3001 

## `datapackage.json`
- same ordering issue as EIA-923
- same null ORCID as EIA-923

zaneselvans:
# Notes on the EIA-860M archive...
Found at: https://zenodo.org/uploads/10064760

## Uploaded Files
- **We have hit the 100 file limit** and there are probably some missing files. What a pain. I guess we need to zip by year?
- Same random ordering problem as the previous archives.

## Zenodo Metadata
- Looks good.

## `datapackage.json`
- Same ordering + pretty-printing issues as EIA-923/EIA-860

e-belfer:
860 published here: https://zenodo.org/records/10067566
Closing this issue as 860M archiver updates require archiver reconfiguration and will be tracked here: https://github.com/catalyst-cooperative/pudl-archiver/issues/190

Issue #184, opened 2018-06-21
Add a view for the CEMS data


karldw:
I can add this.

Design decisions:
- What variables should be in the view?
  - Date
  - EIA info?
  - UTC time, when that's ready
- Should the view be materialized?
- It seems like SQLAlchemy doesn't natively know about views, but there are packages on pypi that add support. Do you want to add another dependency?  

See: https://github.com/catalyst-cooperative/pudl/issues/171#issuecomment-398653563

karldw:
While we're at it, should we drop all of the *_rate_lbs_mmbtu variables from the table and put them in a view?

zaneselvans:
I think this should be rolled in with #171 and the normalization & naming stuff that @cmgosnell is moving into.

PR #3004, opened 2023-11-02
Update sources, DOI and copyright link in PUDL

e-belfer:
In response to comments on #3000 and #3001 and the Zenodo migration in [pudl-archiver PR #184](https://github.com/catalyst-cooperative/pudl-archiver/pull/184) this PR makes some minor modifications to the way we write and read datapackages and source data metadata on the PUDL side. Specifically:

- Removes individual contributors from PUDL data sources, including only Catalyst Cooperative as a data source for our main raw archived datasets
- Adds a `DOI` field with pydantic validation to the datapackage class, using regex from the `pudl-archiver`
- Adds a new `zenodo-role` field to the contributors which defaults to `project member`, and encodes the list of possible roles in the new Zenodo API (a newly required field).
- Updates the US copyright link

# PR Checklist


zaneselvans (changes requested):
Is there a good reason not to re-use our existing `ZenodoDoi` pydantic type here?

src/pudl/metadata/classes.py
```diff
@@ -228,6 +228,7 @@ def __repr_args__(self) -> list[tuple[str, Any]]:
 
 HttpUrl = pydantic.AnyHttpUrl
 """Http(s) URL."""
+"""DOIs from Zenodo archives."""
```
zaneselvans, line 231: I think this comment is a leftover?

zaneselvans (approved):
There's a cut-and-paste leftover comment string but other than that go for it.

PR #3007, opened 2023-11-02
Update ferc-ferc plant matching with ccai implementation.

zschira:
Closes catalyst-cooperative/ccai-entity-matching#109 .

This PR pulls @katie-lamb's CCAI implementation of the FERC-FERC inter-year plant matching process. This new implementation works well, and seems to be running much faster than the old implementation (~2 seconds vs ~8 seconds on the `etl_fast` dataset).

### Remaining work:
- [x] Verify new model runs in full etl
- [ ] Clean up code after dagster integration
- [ ] Get review of current implementation after model optimizations and dagster integration
- [ ] Run validation tests locally


katie-lamb:
This looks good so far! Is it the PCA or the clustering that blows up memory?

katie-lamb:
@zschira I made some somewhat confusing but hopefully helpful plots to better understand the distance between clusters. With the fitted`sklearn Agglomerative Clustering` model one can make a `scipy` dendrogram chart (see [this example](https://scikit-learn.org/stable/auto_examples/cluster/plot_agglomerative_dendrogram.html)). To understand what a dendrogram is, I thought [this example](https://www.displayr.com/what-is-dendrogram/) was helpful. That being said, I'm still not entirely sure how to interpret these plots, and would love some other ideas if I've missed the mark here.

These plots are run on the small subset of data I've been using (2000 records), and shows the clusters that are p=40 merges from the final merge. The y-axis shows the distance at which two nodes are merged into one, indicated by a bracket connecting them. I recommend ignoring the labels on the x-axis, but it basically represents the size of each node. The red horizontal line indicates the threshold I've been using for two clusters to be merged (currently .5). The merges made above this line did it happen and aren't represented in the labels. The merges below this bar are represented by the labels and were clusters that were merged to form a new cluster. I progressively zoomed in on the y-axis so we can see what's going on.

There's a big jump from threshold of 10 to threshold of 2. I think this means that 10 is way too large of a cluster distance threshold.    
Screen Shot 2023-11-09 at 12 47 51 AM

Zooming in on a y axis of 0 to 1, it still seems like a lot of merging happens at a much smaller distance. Maybe it's an indication that the threshold should actually be lower? Maybe it doesn't matter too much if the threshold is .5 or .1, still thinking about that and not entirely sure what to make of it.  
Screen Shot 2023-11-09 at 12 47 58 AM

More merging happening in the <.05 range. But that's also to be expected. There's maybe more clustering of larger nodes (bigger on the x-axis), which is good.  
Screen Shot 2023-11-09 at 12 48 05 AM

It's a little harder to visualize results with the model run on the full dataset, but for the most part it seems like results align with the smaller sample dataset. I ran this with p=20, so 20 merges from the final merge, because it was impossible to tell what was going on with p=40.

Screen Shot 2023-11-09 at 12 59 34 AM
Screen Shot 2023-11-09 at 12 59 40 AM
Screen Shot 2023-11-09 at 12 59 47 AM

zschira:
@zaneselvans I've spot checked a number of plant id's and so far they've all looked like the same plant to me. 

I think doing some focused spot checking using the dendrogram as a guide would be interesting. For example, look at clusters that merge just above our threshold at the 0.5-0.6 range and see if they look like the same plant or not, or do the same just below the threshold, and then maybe zoom way in to clusters with very small distances, and see if we can find any that don't seem to be general matches. I guess generally, it would be interesting to find some cases where we're clearly failing (matching plants that shouldn't be matched or not matching plants that should), and see where/why those might have gone wrong.

katie-lamb:
@zaneselvans 

> I'm definitely confused here. With several thousand expected clusters (one for each FERC plant) it seems hard to visualize all of them at once.
> 
You can think of the dendrograms as a sample of the several thousand clusters that are created by the model. In our case, the p=40 parameter is not that meaningful, since I'm instead using a distance threshold to decide when merging should stop.  I think the dendrogram is mostly helpful to understand if the distance threshold for merging is appropriate. As @zschira pointed out, the second step for validating is probably to spot check nodes that are merged right below or right above that .5 mark. 

This also verifies why the average distance between records in a cluster is always very small (.05 or less) even when I experiment with a distance threshold in the .2-1 range. The vast majority of merges happen between nodes with a distance <.05, so looking at the average distance between records in a cluster isn't a very helpful metric for verifying whether the threshold is good in the .2-1 range.

> Have you made histograms of the cluster sizes in the old vs. new systems?
> 
Yes, currently in the notebook in the CCAI repo. Here's a comparison from the smaller sample dataset (2000 records, I didn't have a screenshot of the full model histogram), where the average new cluster size is ~5.8 and the average old `plant_id_ferc1` size is ~4.3. This disparity is more pronounced on the whole dataset, where the new average cluster size is ~7.3.

Screen Shot 2023-11-09 at 10 15 54 AM


Screen Shot 2023-11-09 at 10 15 24 AM

I'll do some spot checking around the current distance threshold as Zach suggests.

zaneselvans:
I think the y-axis label might belong on the x-axis?

It's interesting that there's a bump at the very high end of the length spectrum (like, one record for every possible year) in the old version, but not in the new version.  I wonder why that would have happened, and whether we've actually lost some good long time series, or if they were bad for some reason and the new algorithm does a better job of distinguishing them?

katie-lamb:
> I think the y-axis label might belong on the x-axis?
> 
Oh yep, I did that too fast, but switch the label from y-axis to x-axis. 

> It's interesting that there's a bump at the very high end of the length spectrum (like, one record for every possible year) in the old version, but not in the new version. I wonder why that would have happened, and whether we've actually lost some good long time series, or if they were bad for some reason and the new algorithm does a better job of distinguishing them?

That's a good point, and that disparity at the high end of the length spectrum happens in the full dataset as well. I'll spot check some of those.

src/pudl/analysis/record_linkage/cross_year.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def transform(self, X, y=None, **fit_params):  # noqa: N803
+        """Compute distance between records then add penalty to records from same year."""
+        return pairwise_distances(X, metric=self.metric)
+
+
+class DistancePenalizeSameYear(PrecomputeDistance):
+    """Custom estimator to compute distances used to identify clusters of plants."""
+
+    def __init__(self, report_years: np.array, metric="euclidean", penalty=100):
```
katie-lamb, line 199: I actually updated the penalty to 1000 to eliminate long runs with duplicate report years.

src/pudl/analysis/record_linkage/cross_year.py
```diff
@@ -0,0 +1,241 @@
...
+        """Initialize estimator with configurable parameters.
+
+        Args:
+            report_years: reporty_year column used to penalize distance for records
```
katie-lamb, line 203: I feel confused about this `report_years` parameter, it's an array of all the records that share the same report year? I think some clarification in this docstring could be helpful

src/pudl/analysis/record_linkage/cross_year.py
```diff
@@ -0,0 +1,241 @@
...
+class DenseTransformer(BaseEstimator, TransformerMixin):
+    """Convert sparse numpy matrix to dense numpy array."""
+
+    def fit(self, X, y=None, **fit_params):  # noqa: N803
+        """No modifications made during fitting."""
+        return self
+
+    def transform(self, X, y=None, **fit_params):  # noqa: N803
+        """No modifications made during fitting."""
+        return np.asarray(np.float32(X.todense()))
```
katie-lamb, line 241: Now I'm not remembering if converting to `np.float32` actually matters. I think that was only required if the dense matrix was going to be an input to an `faiss` index. I guess it's fine to keep as float32?

katie-lamb (approved):
I think this looks good! I left a few small comment in the cross_year module. It might be worth taking a look at the validations that are currently in there and maybe adding a couple more metrics from the FERC to FERC notebook. I'll leave a separate comment with what might specifically be helpful to use as a validation metric.

src/pudl/analysis/record_linkage/cross_year.py
```diff
@@ -0,0 +1,241 @@
...
+    # This can be handled more elegantly in Pydantic 2.0.
+    class Config:
+        """To allow a BaseEstimator for 'transformer', arbitrary types must be allowed."""
+
+        arbitrary_types_allowed = True
+
+    def clean_columns(self, df):
+        """Perform configurable set of cleaning operations on inputs before pipeline."""
+        for cleaning_op in self.cleaning_ops:
+            df[self.columns] = _CLEANING_FUNCTIONS[cleaning_op](df, self.columns)
```
katie-lamb, line 73: I wonder if the string cleaning stuff that I do with the CompanyNameCleaner for both the EIA to FERC and FERC to FERC match should happen as part of these cleaning operations. It's more of a pre-processing step but it's done for the sake of better record linkage. If you want to punt that to a later FERC to EIA PR that's also okay.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,309 @@
...
+                        ),
+                        "cleaning_ops": ["null_to_zero"],
+                    }
+                ),
+            ],
+        }
+    )
+    ferc1_steam_df = clf.fit_predict(ferc1_steam_df)
+
+    # Set the construction year back to numeric because it is.
```
katie-lamb, line 265: Should the NA values that were filled with blank strings or 0's be reverted or is that no longer necessary? In the old module, that happens in `revert_filled_in_string_nulls` and `revert_filled_in_float_nulls` of the old `classify_plants_ferc1` module

katie-lamb (changes requested):
Sorry, didn't finish making all comments at once. Here's a few more things.

src/pudl/analysis/record_linkage/cross_year.py
zschira: I changed the comment a bit, let me know if it's any clearer. I made this just take the report year column instead of the whole dataframe, because I wasn't sure if the dataframe was going to be modified during the pipeline. We could try to change it back if you think that'd be clearer though

src/pudl/analysis/record_linkage/cross_year.py
zschira: Oh yeah I'd been meaning to pull the company name cleaning in, so I just did that. I think I might refactor how some of the cleaning steps work, so I didn't think too hard about how to hook it in here

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
zschira: Good call

zaneselvans:
I think you might need to add an `__init__.py` to the new `record_linkage` subpackge to make it importable, and bring in all of the new modules, and add an `import record_linkage` into `src/pudl/analysis/__init__.py` to reflect the new subpackage.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,359 @@
...
+    )
+    # We don't actually want to save the fuel fractions in this table... they
+    # were only here to help us match up the plants.
+    ferc1_steam_df = ferc1_steam_df.drop(ffc, axis=1)
+    ferc1_steam_df = revert_filled_in_string_nulls(ferc1_steam_df)
+
+    return ferc1_steam_df
+
+
+def revert_filled_in_string_nulls(df: pd.DataFrame) -> pd.DataFrame:
```
katie-lamb, line 287: BTW I think I'm going to use these two "revert" functions for the FERC to EIA match, so it might be worth pulling them out into a more general module (`cleaning_steps`? not really a cleaning step but an uncleaning step)

zschira: Good to know. I'm working on some refactoring/reorganization in the ferc-eia integration, so I'll probably leave that for the next PR

src/pudl/analysis/record_linkage/cross_year.py
```diff
@@ -0,0 +1,246 @@
...
+class DistancePenalizeSameYear(PrecomputeDistance):
+    """Custom estimator to compute distances used to identify clusters of plants."""
+
+    def __init__(self, report_years: np.array, metric="euclidean", penalty=1000):
+        """Initialize estimator with configurable parameters.
+
+        Args:
+            report_years: Used to find records with same report year and add significant
+                          distance penalty to these records to avoid matching records.
+                          from the same year.
```
zaneselvans, line 209: Indentation should be 4 spaces in from the name of the argument.

src/pudl/analysis/record_linkage/cross_year.py
```diff
@@ -0,0 +1,246 @@
...
+    Defines a scikit-learn transformer to be used by a ColumnTransformer. See
+    https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html#sklearn.compose.ColumnTransformer
+    for more information.  The full set of column transforms will transform the
+    input DataFrame into a feature matrix that will be to cluster records.
+
+    The 'transformer' should be a string or an initialized scikit-learn BaseEstimator.
+    If it is a string, it should be one of the following:
+        'string' - Applies a TfidfVectorizer to the column.
+        'category' - Applies a OneHotEncoder to the column.
+        'number' - Applies a MinMaxScaler to the column.
```
zaneselvans, line 55: I suspect this indentation is what's causing the docs build to fail.

src/pudl/analysis/record_linkage/__init__.py
```diff
@@ -0,0 +1 @@
+"""This module impolements models for various forms of record linkage."""
```
zaneselvans, line 1: The pattern we've adopted in the other subpackages is to have each `__init__.py` import all of the modules in that subpackage.  Should we do that here? Or should we be doing something different everywhere?

test/integration/record_linkage.py
```diff
@@ -0,0 +1,211 @@
...
+        }
+    )
+
+    # Add random edits to plant name
+    generated_df["plant_name_ferc1"] = generated_df["base_plant_name"].apply(
+        _randomly_modify_string
+    )
+
+    # Modify capacity rows based on probability and sigma
+    generated_df["capacity_mw"] = _noisify(
```
zschira, line 128: Probably not the best way to model capacity. Adding a bit of random noise to ~1% of rows doesn't account for expected capacity changes if generators were added or removed. It also doesn't really reflect cases where the value is just totally bogus

test/integration/record_linkage.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+    generated_df = _modify_categorical(
+        generated_df,
+        "plant_type",
+        list(PLANT_TYPE_CATEGORIES["categories"]),
+        plant_type_error_prob,
+    )
+
+    # Generate l1 unit vector to represent fuel fractions (all add up to 1)
+    fuel_frac_vec = abs(_RANDOM_GENERATOR.normal(size=len(_FUEL_COLS)))
```
zschira, line 155: Also probably not a great way to model the fuel fractions where plants tend to be heavily weighted towards one fuel, not distributed radomly across all types

test/integration/record_linkage.py
```diff
@@ -0,0 +1,211 @@
...
+    return generated_df
+
+
+@pytest.fixture
+def mock_ferc1_plants_df():
+    """Returns a test DataFrame for use in generic record linkage testing."""
+    # Creates test dataframes for a bunch of plants and concatenates them
+    return pd.concat(
+        [
+            _generate_random_test_df("fox lake, mn"),
```
zschira, line 172: I picked some random names that seemed somewhat representative of the types of plant names we see, and random edits will be applied to certain rows. I think this is a decent reflection of some of the types of errors we see in names, but there's probably other types that this misses

zschira (commented):
Left some of my thoughts on the weaknesses in the random df generation. I'm sure there's plenty more, though.

src/pudl/analysis/record_linkage/__init__.py
zschira: I don't think that the imports with `__init__.py`, are strictly necessary, and in my experience it's pretty common practice to have empty `__init__.py` files in sub-packages. I always thought it was more if you want to make code from a module available in the package level namespace you could import it in `__init__.py`, but python doesn't have a problem finding modules within a package. All that said, I don't really feel like I have a firm grasp on import best practices, and I'm totally fine with sticking with the pattern used elsewhere for consistency.

src/pudl/analysis/record_linkage/cross_year.py
zschira: Fixed

src/pudl/analysis/record_linkage/cross_year.py
zschira: Yeah I think you're right, should be fixed now.

zaneselvans:
I tried to run the full ETL and ended up with a failure in the `plants_steam_ferc1` asset, with the process crashing, so maybe it's running out of memory? I've got 32GB. @zschira how much memory do you have locally? Have you had any recent issues running the full ETL?

```
Multiprocess executor: child process for step plants_steam_ferc1 was terminated by signal 11 (SIGSEGV).
dagster._core.executor.child_process_executor.ChildProcessCrashException

Stack Trace:
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/executor/multiprocess.py", line 249, in execute
    event_or_none = next(step_iter)
                    ^^^^^^^^^^^^^^^
,  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/executor/multiprocess.py", line 360, in execute_step_out_of_process
    for ret in execute_child_process_command(multiproc_ctx, command):
,  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/executor/child_process_executor.py", line 174, in execute_child_process_command
    raise ChildProcessCrashException(exit_code=process.exitcode)
```

zaneselvans:
Giving this a try locally in the context of a full ETL run and:
* It's been running for more than 2.5 hours.  The last 40 minutes or so has been only the two assets that are parallelized internally `plants_steam_ferc1` and `imputed_hourly_demand_ferc714`.  Given that each of them runs in 10-20 minutes in isolation, I wonder if they aren't fighting over resources and getting bogged down.
* It would be interesting to see what the run times look like if each of these assets is only allowed to use at most 50% of the available CPUs.
* For the `imputed_hourly_demand_ferc714` asset, there's some internal logging that indicates it's making progress (whenever it starts imputing a new year, it prints out the year).
* It would be nice if there were some logging output from the `plants_steam_ferc1` asset as well, so we can see what kind of progress it's making, if it's going to take a very long time. Are there any internal steps it can report out about? As it is, it hasn't produced any output for 2.5 hours, and the last thing it said was:
```2023-12-01 11:45:34 [    INFO] catalystcoop.pudl.analysis.record_linkage.classify_plants_ferc1:47 Identifying distinct large FERC plants for ID assignment.```

**Comment by zaneselvans** [comment]
Re-running just the `plants_steam_ferc1` asset in isolation, I get to the same crash in about 12 minutes.

**Comment by zaneselvans** [comment]
```py
bad_plants = {
    # APS Yucca 3 misses the plant in 2021-2022 solely due to name change and a blip in fuel fraction splits. Everything else is consistent. Seems too sensitive.
    "yucca3": lambda steam: (steam.plant_id_pudl == 1013),
    # The columbia power station in Wisconsin seems genuinely complicated. It's a mishmash of different units with different owners.
    # Sometimes discontinuous years. Getting multiple records from the same year assigned the same plant_id_ferc1.
    # Sometimes big changes in the capacity within the same plant_id_ferc1.
    "columbia": lambda steam: (steam.plant_id_pudl == 124),
    # APS West Phoenix 5, data for 2021-2022 are not associated with other records due to different name.
    "west_phoenix_5": lambda steam: (steam.plant_name_ferc1.str.contains(r"west phoenix.*5", regex=True, case=False)),
    # sterling avenue plant is very consistent except for construction_type, which is usually null but has values in a couple of years,
    # so in those years it gets split off from all the other years, e.g. in 2003
    "sterling": lambda steam: (steam.plant_name_ferc1.str.contains(r"sterling", regex=True, case=False)),
    # Jeffrey Energy Center is getting split up badly due to minor changes in the name
    "jeffrey": lambda steam: (steam.plant_name_ferc1.str.contains(r"jeffrey", regex=True, case=False)) & (steam.utility_id_ferc1==255),
    "jeffrey_8pct": lambda steam: (steam.plant_name_ferc1.str.contains(r"jeffrey.*8\%", regex=True, case=False)) & (steam.utility_id_ferc1==255),
    # Jeffrey Energy Center NOT getting split up when it should due to large variations in capacity
    "jeffrey_capacity": lambda steam: (steam.plant_name_ferc1.str.contains(r"jeffrey energy cntr", regex=True, case=False)) & (steam.utility_id_ferc1==255),
    # Belews Creek record for 2022 gets lost. The only change is gas/coal fuel ratio
    "belews": lambda steam: (steam.plant_name_ferc1.str.contains(r"belews", regex=True, case=False)),
    # Minor name changes split valmy 1&2 records.
    # NA fuel fractions in 2013 split valmy 1&2 records.
    "valmy12": lambda steam: (steam.plant_name_ferc1.str.contains(r"valmy.*1.*2", regex=True, case=False)),
    # Many cases of more than one record from the same year getting assigned the same plant_id_ferc1
    "valmy_duplicate_years": lambda steam: (steam.plant_name_ferc1.eq(r"valmy")),
    # Minor name change + null fuel fractions split this plant:
    "niles": lambda steam: (steam.plant_name_ferc1.str.contains(r"niles")) & (steam.plant_type.eq("combustion_turbine")),
    # Minor name change and flaky fuel categorization splits the plant
    "manatee": lambda steam: (steam.plant_name_ferc1.str.contains(r"manatee")) & (steam.plant_type.eq("steam")),
    # HB Robinson steam plant records get split due to whitepsace change in name, flaky capacity & fuel fraction reporting.
    "hb_robinson_steam": lambda steam: (steam.plant_name_ferc1.str.contains(r"h.*b.*robinson", case=False, regex=True)) & (steam.plant_type.eq("steam")),
}
```

src/pudl/analysis/__init__.py
```diff
@@ -11,6 +11,7 @@
     ferc1_eia_record_linkage,
     mcoe,
     plant_parts_eia,
+    record_linkage,
```
katie-lamb, line 14: Just making a note that this module name is pretty close to `recordlinkage` which is a dependency in the FERC to EIA match, but I'm planning on removing the `recordlinkage` dependency there, so it's fine.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,361 @@
...
+_FUEL_COLS = [
+    "coal_fraction_mmbtu",
+    "gas_fraction_mmbtu",
+    "nuclear_fraction_mmbtu",
+    "oil_fraction_mmbtu",
+    "waste_fraction_mmbtu",
+]
+
+
+ferc_datframe_embedder = embed_dataframe.dataframe_embedder_factory(
```
katie-lamb, line 32: Typo: `ferc_datframe_embedder` -> `ferc_dataframe_embedder`

src/pudl/output/ferc1.py
```diff
@@ -891,12 +892,16 @@ def drop_other_fuel_types(df):
...
         )
-        .pipe(pudl.analysis.classify_plants_ferc1.revert_filled_in_float_nulls)
-        .pipe(pudl.analysis.classify_plants_ferc1.revert_filled_in_string_nulls)
+        .pipe(
+            pudl.analysis.record_linkage.classify_plants_ferc1.revert_filled_in_float_nulls
+        )
+        .pipe(
+            pudl.analysis.record_linkage.classify_plants_ferc1.revert_filled_in_string_nulls
+        )
         .merge(
```
katie-lamb, line 905: Maybe I'm confused but are these revert function calls still necessary now that the actual ID classifier is run later? All that happens is the fuel columns are merged on so I don't think any nullifying is done yet?

src/pudl/analysis/record_linkage/link_cross_year.py
```diff
@@ -0,0 +1,273 @@
...
+    """Compute a distance matrix and penalize records from the same year."""
+    return DistanceMatrix(feature_matrix.matrix, original_df, config)
+
+
+class DBSCANConfig(Config):
+    """Configuration for DBSCAN step."""
+
+    #: See :class:`sklearn.cluster.DBSCAN` for details.
+    eps: float = 0.1
+    min_samples: int = 2
```
katie-lamb, line 126: Probably out of scope for this PR but it would be nice to potentially set up some experiment tracking infrastructure so we can do some cross validation on model parameters. Harder to do in this context since we have no training data, but in the FERC to EIA process we're running a full grid search that probably shouldn't happen every time, but should instead be documented somewhere else and revisited when data changes?

katie-lamb (changes requested):
I think this looks good and I'm excited about the infrastructure that you set up through this PR! Left a bunch of little comments, some of them are more just thoughts versus requests for changes.

katie-lamb:
@zschira Looking at the "bad plants" that @zaneselvans found, it seems like the lowest hanging fruit are those which have name changes. I wonder if it makes sense to play with the TfidfVectorizer parameters. The old model used character n grams from 2 to 10, but in my experience with FERC to EIA I've had more success with word level tokenization and shorter n grams. Otherwise, some stricter or specific rules in the Company Name Cleaner to address the cases Zane found seems like it would help. It might be worth trying to strip digits from plant names?

jdangerx:
Current state - we're tuning parameters to address some of these specific cases, while trying to avoid over-fitting to the specific test cases.

We want to have some sort of metric for when this is "good enough," so we can avoid tuning forever.


Options:

- time-box remaining development
- flesh out the generative testing that @zschira has already set up - and then use match % as a comparison metric with existing FERC-FERC
- potentially use a [metamorphic test](https://www.hillelwayne.com/post/metamorphic-testing/) and *its* match % to compare with existing FERC-FERC as well - definitely timebox this if we decide to do it at all

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,364 @@
...
+        fuel_fractions[["utility_id_ferc1", "plant_name_ferc1", "report_year"] + ffc],
+        on=["utility_id_ferc1", "plant_name_ferc1", "report_year"],
+        how="left",
+    ).astype({"plant_type": str, "construction_type": str})
+
+
+@graph_asset
+def _out_ferc1__yearly_steam_plants_sched402_with_plant_ids(
+    core_ferc1__yearly_steam_plants_sched402: pd.DataFrame,
+    out_ferc1__yearly_steam_plants_fuel_by_plant_sched402: pd.DataFrame,
```
katie-lamb, line 148: I updated the name of this asset and the input assets after merging `dev` with the big PUDL rename

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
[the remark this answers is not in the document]
zschira: Fixed!

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,361 @@
...
+    function replaces empty strings with null values for specific columns that are known
+    to contain empty strings introduced for the classifier.
+    """
+    for col in [
+        "plant_type",
+        "construction_type",
+        "fuel_type_code_pudl",
+        "primary_fuel_by_cost",
+        "primary_fuel_by_mmbtu",
+    ]:
```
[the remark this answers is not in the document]
zschira: The revert functions are a bit weird right now. With the new architecture, they never actually need to be used during the matching process, because it doesn't modify the original dataframe. They are still being used in the fuel by plant asset though, so I've kept them for legacy reasons.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,361 @@
...
+
+def revert_filled_in_float_nulls(df: pd.DataFrame) -> pd.DataFrame:
+    """Revert the filled nulls from float columns.
+
+    Many columns that are used for the classification in
+    :func:`plants_steam_assign_plant_ids` have many nulls. The classifier can't handle
+    nulls well, so we filled in nulls with zeros for float columns. This function
+    replaces zeros with nulls for all float columns.
+    """
+    float_cols = list(df.select_dtypes(include=[float]))
```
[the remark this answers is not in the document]
zschira: I broadly agree, but I'm not sure if this change should be in scope of this PR, given my above comment about the current usage of the revert functions.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,361 @@
...
+    label_df = link_ids_cross_year(input_df, feature_matrix)
+
+    return plants_steam_validate_ids(plants_steam_ferc1, label_df)
+
+
+def revert_filled_in_string_nulls(df: pd.DataFrame) -> pd.DataFrame:
+    """Revert the filled nulls from string columns.
+
+    Many columns that are used for the classification in
+    :func:`plants_steam_assign_plant_ids` have many nulls. The classifier can't handle
```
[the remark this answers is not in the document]
zschira: I've removed the docstring because for the same reason about the revert functions.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+@op
+def plants_steam_validate_ids(
+    ferc1_steam_df: pd.DataFrame, label_df: pd.DataFrame
+) -> pd.DataFrame:
+    """Tests that plant_id_ferc1 times series includes one record per year.
+
+    Args:
+        ferc1_steam_df: A DataFrame of the data from the FERC 1 Steam table.
```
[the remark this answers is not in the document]
zschira: Added

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,361 @@
...
+    nulls well, so we filled in nulls with zeros for float columns. This function
+    replaces zeros with nulls for all float columns.
+    """
+    float_cols = list(df.select_dtypes(include=[float]))
+    if float_cols:
+        df.loc[:, float_cols] = df.loc[:, float_cols].replace(0, np.nan)
+    return df
+
+
+def fuel_by_plant_ferc1(
```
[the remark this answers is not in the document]
zschira: I've moved this plus the revert functions to their own analysis module to stop from cluttering `pudl.output.ferc1.py`, let me know if this seems reasonable.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,361 @@
...
+        )
+        .sum()
+        .reset_index()
+        # Set the index to the keys, and pivot to get per-fuel columns:
+        .set_index(["utility_id_ferc1", "plant_name_ferc1", "report_year"])
+        .pivot(columns="fuel_type_code_pudl")
+        .fillna(0.0)
+    )
+
+    # Undo pivot. Could refactor this old function
```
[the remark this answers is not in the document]
zschira: Yeah I've been trying to avoid touching this for now, as this is already a large enough PR.

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -0,0 +1,181 @@
...
+    return embed_dataframe
+
+
+class TextVectorizer(TransformStep):
+    """Implement TransformStep for :class:`sklearn.feature_extraction.text.TfidfVectorizer`."""
+
+    name: str = "tfidf_vectorizer"
+
+    #: See sklearn documentation for all options
+    options: dict = {"analyzer": "char", "ngram_range": (2, 10)}
```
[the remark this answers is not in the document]
zschira: I was kind of viewing these as our defaults, and intentionally not mirroring `sklearn`. Particularly if these values work for the ferc-eia match, I think it would be good to use our own set of defaults.

src/pudl/analysis/record_linkage/link_cross_year.py
```diff
@@ -0,0 +1,273 @@
...
+        """Get new unique cluster id."""
+        while True:
+            max_cluster_id += 1
+            yield max_cluster_id
+
+    cluster_id_generator = _generate_cluster_ids(id_year_df.record_label.max())
+    classifier = AgglomerativeClustering(
+        metric="precomputed",
+        linkage="average",
+        distance_threshold=config.distance_threshold,
```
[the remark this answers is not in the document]
zschira: Yeah `eps` and `distance_threshold` are definitely not identical, however they both do apply an upper bound on the distances within a cluster, so I agree it would make sense to make them the same by default.

src/pudl/analysis/record_linkage/link_cross_year.py
```diff
@@ -0,0 +1,273 @@
...
+            continue
+
+        cluster_inds = id_year_df[
+            id_year_df.record_label == duplicated_id
+        ].index.to_numpy()
+        cluster_distances = get_cluster_distance_matrix(
+            distance_matrix.distance_matrix, cluster_inds
+        )
+
+        new_labels = classifier.fit_predict(cluster_distances)
```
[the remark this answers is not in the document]
zschira: Yeah this will split clusters with a duplicate year record into two or more new clusters. The benefit of doing it like this is that after the `dbscan` step, there are many clusters that seem to contain two distinct groupings within a single cluster, and this efficiently separates them.

src/pudl/analysis/record_linkage/link_cross_year.py
```diff
@@ -0,0 +1,273 @@
...
+import pudl
+from pudl.analysis.record_linkage.embed_dataframe import FeatureMatrix
+
+logger = pudl.logging_helpers.get_logger(__name__)
+
+
+class PenalizeReportYearDistanceConfig(Config):
+    """Compute distance between records and add penalty to records from same year."""
+
+    distance_penalty: float = 10000.0
```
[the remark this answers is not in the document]
zschira: Yeah I messed with this a bit, and it doesn't really have any impact once it's sufficiently large. DBSCAN is just not really sensitive to it after a certain point, which is why the other two steps are necessary

src/pudl/output/ferc1.py
zschira: In the fuel by plant fuction, there's some `fillna`'s being done that I think do make this necessary. It seems like there's definitely room to refactor the fuel by plant stuff, but I've been hesitant to change anything too much given the large size of this PR already.

src/pudl/analysis/record_linkage/link_cross_year.py
zschira: Yeah I think some experiment tracking would be really nice, but I'm not sure when the best time to do that would be. Maybe it's something we could spend some remaining hours in the new year on if time permits, or during the mozilla grant?

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -0,0 +1,181 @@
...
+    name: str = "one_hot_encoder_vectorizer"
+
+    options: dict = {"categories": "auto"}
+
+    def as_transformer(self):
+        """Return configured OneHotEncoder."""
+        return OneHotEncoder(**self.options)
+
+
+class NumericalVectorizer(TransformStep):
```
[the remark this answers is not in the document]
zschira: Fixed

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -0,0 +1,361 @@
...
+    # Test to make sure that we don't have any plant_id_ferc1 time series
+    # which include more than one record from a given year. Warn the user
+    # if we find such cases (which... we do, as of writing)
+    year_dupes = (
+        ferc1_steam_df.groupby(["plant_id_ferc1", "report_year"])["utility_id_ferc1"]
+        .count()
+        .reset_index()
+        .rename(columns={"utility_id_ferc1": "year_dupes"})
+        .query("year_dupes>1")
+    )
```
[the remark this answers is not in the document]
zschira: That seems reasonable to me, changed.

test/integration/record_linkage.py
```diff
@@ -0,0 +1,240 @@
...
+            _generate_random_test_df("keystone *", capacity_mean=1872.0, size=3),
+            _generate_random_test_df("keystone", capacity_mean=50.0, size=21),
+            _generate_random_test_df(
+                "keystone 1&2 (3.70%)", capacity_mean=69.3, size=5
+            ),
+        ]
+    ).reset_index()
+
+
+def test_classify_plants_ferc1(mock_ferc1_plants_df):
```
katie-lamb, line 217: I think this testing seems good enough! One last idea I had from that article on Metamorphic Testing: what if you run the clustering once on the test dataframe, then shuffle the input test dataframe, rerun the clustering, and check that assigned IDs are the same.

src/pudl/output/ferc1.py
katie-lamb: That makes sense to me about not doing that refactor here! Just wanted to be careful about making "0's" in float columns null when the true value is actually 0.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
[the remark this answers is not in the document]
katie-lamb: Got it. That sounds good then.

src/pudl/output/ferc1.py
katie-lamb: Saw your comment above about using these reverts for legacy reasons.

test/integration/record_linkage.py
katie-lamb: Never mind, I forgot you're starting with DBSCAN and that seems agnostic to the order of the input dataframe.

zaneselvans:
It looks like when the modules under `pudl.analysis` were rearranged, the new import paths were not added to `pudl.analysis.__init__.py` so they were not importable by other modules, which was causing the use of `pudl.analysis.fuel_by_plant` functions to fail in the steam table processing.  the `record_linkage` module was also removed from this subpackage (and turned into its own subpackage) but it wasn't 

I'm running the full ETL locally using the `etl_full` job and also seeing that EPA CEMS is using all 10 of my CPUs rather than limiting itself to a concurrency of 2.  Did something get changed that makes that safe? Is the EPA CEMS concurrency setting not working in the Dagster UI?

zaneselvans:
After including the new import paths, I get a failure on the plant parts EIA:

```
AssertionError: Consistency of all matches across years dipped below 75.0% to 53.5%
  File "/Users/zane/code/catalyst/pudl/src/pudl/analysis/eia_ferc1_record_linkage.py", line 94, in out_pudl__yearly_assn_eia_ferc1_plant_parts
    connects_ferc1_eia = prettyify_best_matches(
                         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/analysis/eia_ferc1_record_linkage.py", line 888, in prettyify_best_matches
    check_match_consistency(
  File "/Users/zane/code/catalyst/pudl/src/pudl/analysis/eia_ferc1_record_linkage.py", line 998, in check_match_consistency
    raise AssertionError(
```

zaneselvans:
Hmm, attempting to re-run just the EPA CEMS it seems to do them two-by-two. So maybe it was just that the whole EPA CEMS graph job had failed and left the unstarted ghost assets in the UI.

zaneselvans:
@zschira The builds passed last night so this could go into `dev` cleanly now.

Do we know why the test coverage drops by half a percent between this PR and `dev`? That seems like kind of a lot.

zaneselvans (approved):

PR #3025, opened 2023-11-07
Fix validation `test_fbp_ferc1_mismatched_fuels` error

cmgosnell:
# PR Overview
We were getting the following error:

```python
FAILED test/validate/fbp_ferc1_test.py::test_fbp_ferc1_mismatched_fuels[ferc1_annual] - AssertionError: Too many records (9.69%) have mismatched primary fuel types.
```

I determined that this was happening bc the nulls in the fuel table were somehow getting converted to `"nan"`:

image

### first solution:
A part of the `pudl.analysis.classify_plants_ferc1.fuel_by_plant_ferc1` step does convert nulls to `""` and there is a post-processing step that we were already applying (`pudl.analysis.classify_plants_ferc1.revert_filled_in_string_nulls`). I simply added `"nan"` in here.

Still tbh do no know why these `"nan"`s were all of a sudden appearing but hey.. they are gone now 👀 

### current solution:
it looooks as if the "nan" is added in right here...
```python
for fuel_str in fuel_categories:
    try:
        mmbtu_mask = df[f"{fuel_str}_fraction_mmbtu"] > thresh
        df.loc[mmbtu_mask, "primary_fuel_by_mmbtu"] = fuel_str
    except KeyError:
        pass

    try:
        cost_mask = df[f"{fuel_str}_fraction_cost"] > thresh
        df.loc[cost_mask, "primary_fuel_by_cost"] = fuel_str
    except KeyError:
        pass
```
So I stopped em in their tracks 🚥 . these columns don't exist before this loop and not all of the records get labeled with a `fuel_str` bc not all of the records meet the `thresh` so I added the whole column before the loop w/ values of `pd.NA` 

# PR Checklist

### DIFF
--- a/src/pudl/analysis/classify_plants_ferc1.py
+++ b/src/pudl/analysis/classify_plants_ferc1.py
@@ -546,7 +546,7 @@ def revert_filled_in_string_nulls(df: pd.DataFrame) -> pd.DataFrame:
         if col in df.columns:
             # the replace to_replace={column_name: {"", pd.NA}} mysteriously doesn't work.
             df[col] = df[col].replace(
-                to_replace=[""],
+                to_replace=["", "nan"],
                 value=pd.NA,
             )
     return df
--- a/src/pudl/analysis/classify_plants_ferc1.py
+++ b/src/pudl/analysis/classify_plants_ferc1.py
@@ -736,6 +736,12 @@ def fuel_by_plant_ferc1(
 
     # Label each plant-year record by primary fuel:
     df.loc[:, ["primary_fuel_by_cost", "primary_fuel_by_mmbtu"]] = pd.NA
+    df = df.astype(
+        {
+            "primary_fuel_by_cost": pd.StringDtype(),
+            "primary_fuel_by_mmbtu": pd.StringDtype(),
+        }
+    )
     for fuel_str in fuel_categories:
         try:
             mmbtu_mask = df[f"{fuel_str}_fraction_mmbtu"] > thresh
--- a/src/pudl/analysis/classify_plants_ferc1.py
+++ b/src/pudl/analysis/classify_plants_ferc1.py
@@ -546,7 +546,7 @@ def revert_filled_in_string_nulls(df: pd.DataFrame) -> pd.DataFrame:
         if col in df.columns:
             # the replace to_replace={column_name: {"", pd.NA}} mysteriously doesn't work.
             df[col] = df[col].replace(
-                to_replace=["", "nan"],
+                to_replace=[""],
                 value=pd.NA,
             )
     return df
@@ -654,9 +654,11 @@ def fuel_by_plant_ferc1(
     ]
 
     # Ensure that the dataframe we've gotten has all the information we need:
-    for col in keep_cols:
-        if col not in fuel_df.columns:
-            raise AssertionError(f"Required column {col} not found in input fuel_df.")
+    missing_cols = [col for col in keep_cols if col not in fuel_df.columns]
+    if missing_cols:
+        raise AssertionError(
+            f"Required columns not found in input fuel_df: {missing_cols}"
+        )
 
     # Calculate per-fuel derived values and add them to the DataFrame
     df = (
@@ -679,7 +681,8 @@ def fuel_by_plant_ferc1(
                 "plant_name_ferc1",
                 "report_year",
                 "fuel_type_code_pudl",
-            ]
+            ],
+            observed=True,
         )
         .sum()
         .reset_index()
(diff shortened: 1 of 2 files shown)


zaneselvans:
This is disconcerting. Previously the ID assignment process was expecting empty strings in there. Rather than turning the mysterious "nan” strings into `np.nan` at the end did you look at the place where the `np.nan` values are turned into empty strings on the way in?

src/pudl/analysis/classify_plants_ferc1.py
```diff
@@ -679,7 +681,8 @@ def fuel_by_plant_ferc1(
                 "plant_name_ferc1",
                 "report_year",
                 "fuel_type_code_pudl",
-            ]
+            ],
+            observed=True,
```
zaneselvans, line 685: Are you sure we want `observed=True` here?  This column is pivoted such that there's a column per fuel type later on, and if there are missing values they'll become NA. Ah. Okay but it looks like we fill NA with 0.0 so I guess it's fine. As long as every fuel type is observed at least once.

src/pudl/analysis/classify_plants_ferc1.py
```diff
@@ -732,6 +735,7 @@ def fuel_by_plant_ferc1(
     ).reset_index()
 
     # Label each plant-year record by primary fuel:
+    df.loc[:, ["primary_fuel_by_cost", "primary_fuel_by_mmbtu"]] = pd.NA
```
zaneselvans, line 738: So here you are overwriting the string "nan" values with real NA values? But we still don't know what changed to introduce them here?  Oh wait didn't @jdangerx do something with null or empty strings / NA values in the XBRL extractor recently? But in your screenshot there are "nan" values in both 1994 and 2022 data.

zaneselvans (commented):
In the code snipped you cite as the source of the "nan" values, how are they getting there?

```py
for fuel_str in fuel_categories:
    try:
        mmbtu_mask = df2[f"{fuel_str}_fraction_mmbtu"] > thresh
        df2.loc[mmbtu_mask, "primary_fuel_by_mmbtu"] = fuel_str
    except KeyError:
        pass

    try:
        cost_mask = df2[f"{fuel_str}_fraction_cost"] > thresh
        df2.loc[cost_mask, "primary_fuel_by_cost"] = fuel_str
    except KeyError:
        pass
```

Is it that "nan" is somehow showing up the `fuel_categories`? That doesn't seem right, since it's just the keys of the `pudl.transform.params.ferc1.FUEL_CATEGORIES["categories"]` dictionary.

src/pudl/analysis/classify_plants_ferc1.py
cmgosnell: these two columns don't exist before the for/try/except loop. they are being introduced in this loopy masky setup.

WHICH SEEMS NORMAL/FINE. but for some reason the records which are not getting labeled w/ a fuel (i.e. the  highest cost/mmbtu fuel does not meet the 50% threshold) at the end of this loopy loopy had `"nan"` primary fuels.

src/pudl/analysis/classify_plants_ferc1.py
cmgosnell: We were previously converting `fuel_type_code_pudl` into a `str` column before this because this column in our metadata is (correctly) a categorical. I suspected early on that the `str` type conversion was a part of the problem (it wasn't). But the only reason we needed that conversion was because of the semi-new `groupby` behavior regarding categoricals.

cmgosnell: the outcome of the new version of the table passes the `pd.testing.assert_frame_equal` so i feel pretty confident that the new behavior is the same.

cmgosnell:
> Is it that "nan" is somehow showing up the fuel_categories? That doesn't seem right, since it's just the keys of the pudl.transform.params.ferc1.FUEL_CATEGORIES["categories"] dictionary.

no, the fuel cats have a specific `na_category` but that definitely isn't the way the "nan"s are coming from. see my comment above

src/pudl/analysis/classify_plants_ferc1.py
zaneselvans: Okay thanks for showing me your debugging. I'm convinced this is a new pandas behavior somehow and now an upstream data issue. Seems very strange though.

src/pudl/analysis/classify_plants_ferc1.py
zaneselvans: Okay thanks for showing me your debugging. I'm convinced this is a new pandas behavior somehow and now an upstream data issue. Seems very strange though.

zaneselvans (approved):

PR #3026, opened 2023-11-07
Fix XBRL extraction clobber

jdangerx:
On dev, if you set `clobber: True` in the `ferc_to_sqlite` job config, it will clobber the existing database when extracting 2021 data, and then clobber it again when extracting 2022 data.

* clobbering happens on PUDL side now, instead of on ferc-xbrl-extractor side - so it gets called once per op instead of once per form
* you can still clobber when running ferc-xbrl-extractor on its own

This was *probably* the cause of the missing 2021 XBRL data that @cmgosnell ran into yesterday.

### DIFF
--- a/src/pudl/extract/xbrl.py
+++ b/src/pudl/extract/xbrl.py
@@ -88,8 +88,13 @@ def xbrl2sqlite(context) -> None:
 
         sql_path = Path(urlparse(PudlPaths().sqlite_db(f"ferc{form.value}_xbrl")).path)
 
-        if clobber:
-            sql_path.unlink(missing_ok=True)
+        if sql_path.exists():
+            if clobber:
+                sql_path.unlink()
+            else:
+                raise RuntimeError(
+                    f"Found existing DB at {sql_path} and clobber was set to False. Aborting."
+                )
 
         convert_form(
             settings,
@@ -117,8 +122,8 @@ def convert_form(
         form_settings: Validated settings for converting the desired XBRL form to SQLite.
         form: FERC form number.
         datastore: Instance of a FERC XBRL datastore for retrieving data.
-        pudl_settings: Dictionary containing paths and database URLs
-            used by PUDL.
+        output_path: PUDL output directory
+        sql_path: path to the SQLite DB we'd like to write to.
         batch_size: Number of XBRL filings to process in a single CPU process.
         workers: Number of CPU processes to create for processing XBRL filings.
 
@@ -131,11 +136,12 @@ def convert_form(
     for year in form_settings.years:
         taxonomy_archive, taxonomy_entry_point = datastore.get_taxonomy(year, form)
         filings_archive = datastore.get_filings(year, form)
-
+        # if we set clobber=True, clobbers on *every* call to run_main;
+        # we already delete the existing base on `clobber=True` in `xbrl2sqlite`
         run_main(
             instance_path=filings_archive,
             sql_path=sql_path,
-            clobber=False,  # if we set clobber=True, clobbers on *every* call to run_main
+            clobber=False,
             taxonomy=taxonomy_archive,
             entry_point=taxonomy_entry_point,
             form_number=form.value,
--- a/src/pudl/extract/xbrl.py
+++ b/src/pudl/extract/xbrl.py
@@ -2,7 +2,6 @@
 import io
 from datetime import date
 from pathlib import Path
-from urllib.parse import urlparse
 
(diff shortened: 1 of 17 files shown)


src/pudl/io_managers.py
```diff
@@ -426,8 +426,9 @@ def __init__(
 
         super().__init__(base_dir, db_name, md, timeout)
 
-        existing_schema_context = MigrationContext.configure(self.engine.connect())
-        metadata_diff = compare_metadata(existing_schema_context, self.md)
+        with self.engine.connect() as conn:
```
jdangerx, line 429: Found an un-closed connection when testing - figured I'd clean that up while I'm here.

src/pudl/extract/xbrl.py
```diff
@@ -128,10 +134,8 @@ def convert_form(
 
         run_main(
             instance_path=filings_archive,
-            sql_path=PudlPaths()
-            .sqlite_db(f"ferc{form.value}_xbrl")
-            .removeprefix("sqlite:///"),  # Temp hacky solution
-            clobber=clobber,
+            sql_path=sql_path,
+            clobber=False,  # if we set clobber=True, clobbers on *every* call to run_main
```
zaneselvans, line 138: It seems like this makes the `clobber` option non-functional, at least in this location. The idea with `clobber` is to functionally `rm -f the_database.sqlite` before starting a new ETL run. Maybe it should only exist at the highest level in the `ferc_to_sqlite` script and actually unlink the paths if set -- and then everything in here can just be like "You gave me a DB URL and I'm going to write to it. Not my problem if there was already something incompatible there."

jdangerx: I think the `unlink()` on line 92 does the `rm -f` you're looking for?

jdangerx: Previously when I failed to clobber, the XBRL extractor would write duplicate data -  when I ran `xbrl_to_sqlite` multiple times I still only had one copy of the data afterwards. I guess it could have not written any data at all, the later times. I can run again and see if the database is gone between the `unlink()` and the database write.

jdangerx: Yeah, the database gets deleted at the `unlink()` step if `clobber` is set to `True`.

zaneselvans: DELETED. Strongbad would be proud. I guess the other half of the logic we've previously associated with `--clobber` is that if the DB already exists and you attempt to run the ETL, it bails out early, to avoid unintentionally modifying an existing DB. It's not trying to be anything fancy -- just a binary switch "Do you want me to make something new and blow everything that exists away if it's there? Or do you expect that there's a clean slate, and there should be no possible conflicts.

src/pudl/extract/xbrl.py
```diff
@@ -85,12 +86,17 @@ def xbrl2sqlite(context) -> None:
             logger.info(f"Dataset ferc{form}_xbrl is disabled, skipping")
             continue
 
+        sql_path = Path(urlparse(PudlPaths().sqlite_db(f"ferc{form.value}_xbrl")).path)
+
+        if clobber:
+            sql_path.unlink(missing_ok=True)
+
```
cmgosnell, line 93: okay so this is now the guy that removes the db and we always run `clobber=True` in pudl when we call the extractor's `run_main` bc each run is per year. cool cool!

src/pudl/extract/xbrl.py
```diff
@@ -101,7 +107,7 @@ def convert_form(
     form: XbrlFormNumber,
     datastore: FercXbrlDatastore,
     output_path: Path,
-    clobber: bool,
+    sql_path: Path,
```
cmgosnell, line 110: i know the arg docs are currently a lil out of date but it'd be good to update em with this change.

cmgosnell (approved):
tbh i don't understand all of the whys for the changes in here but the broad banner of make pudl do a big clobber, always `clobber=False` our run of the year-based extractor makes sense to me and looks good!

src/pudl/extract/xbrl.py
zaneselvans: I don't think this is the current behavior though. If you don't say clobber and there is an existing DB what happens?

jdangerx: In the past if you don't say clobber and there's an existing DB, we actually just append to the existing DB. Which is pretty bad. I'll add a quick check to just bail out if there's an existing DB instead.

PR #3027, opened 2023-11-08
Test that DB schema matches the Alembic migrations

jdangerx:
If the DB schema as defined in `metadata.classes.Package.from_resource_ids()` doesn't match the schema you get from running `alembic upgrade head`, **AND** `PudlPaths().pudl_db` is a pre-existing file, the `pudl_sqlite_io_manager` throws an error at startup.

HOWEVER!
1. in integration tests, when `pudl_sqlite_io_manager` is initialized, `PudlPaths().pudl_db` does not actually point to an existing file, so we use `metadata.create_all()` - which masks any differences between the migrations and the Package
2. in nightly builds, the `pudl_sqlite_io_manager` isn't initialized until we actually want to write to the DB, which is pretty late in the DAG execution. This is a super avoidable nightly build failure.

So, to catch any divergence between the alembic migrations and the package metadata, I added a unit test that runs the migrations on a test db and tries to initialize the io manager.

### DIFF
--- a/test/unit/io_managers_test.py
+++ b/test/unit/io_managers_test.py
@@ -218,12 +218,24 @@ def fake_pudl_sqlite_io_manager_fixture(tmp_path, test_pkg, monkeypatch):
 
 
 def test_migrations_match_metadata(tmp_path, monkeypatch):
-    pkg = Package.from_resource_ids()
+    """If you create a `PudlSQLiteIOManager` that points at a non-existing
+    `pudl.sqlite` - it will initialize the DB based on the `package`.
+
+    If you create a `PudlSQLiteIOManager` that points at an existing
+    `pudl.sqlite`, like one initialized via `alembic upgrade head`, it
+    will compare the existing db schema with the db schema in `package`.
+
+    We want to make sure that the schema defined in `package` is the same as
+    the one we arrive at by applying all the migrations.
+    """
+    # alembic wants current directory to be the one with `alembic.ini` in it
     monkeypatch.chdir(Path(__file__).parent.parent.parent)
     # alembic knows to use PudlPaths().pudl_db - so we need to set PUDL_OUTPUT env var
     monkeypatch.setenv("PUDL_OUTPUT", tmp_path)
+    # run all the migrations on a fresh DB at tmp_path/pudl.sqlite
     alembic.config.main(["upgrade", "head"])
 
+    pkg = Package.from_resource_ids()
     PudlSQLiteIOManager(base_dir=tmp_path, db_name="pudl", package=pkg)
 
     # all we care about is that it didn't raise an error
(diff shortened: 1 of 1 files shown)


test/unit/io_managers_test.py
```diff
@@ -204,40 +206,53 @@ def test_missing_schema_error(sqlite_io_manager_fixture):
 
 
 @pytest.fixture
-def pudl_sqlite_io_manager_fixture(tmp_path, test_pkg):
-    """Create a SQLiteIOManager fixture with a PUDL database schema."""
-    db_path = tmp_path / "pudl.sqlite"
+def fake_pudl_sqlite_io_manager_fixture(tmp_path, test_pkg, monkeypatch):
```
cmgosnell, line 209: does this guy actually need `monkeypatch`?

test/unit/io_managers_test.py
```diff
@@ -204,40 +206,53 @@ def test_missing_schema_error(sqlite_io_manager_fixture):
...
     md = test_pkg.to_sql()
     md.create_all(engine)
-    return PudlSQLiteIOManager(base_dir=tmp_path, db_name="pudl", package=test_pkg)
+    return PudlSQLiteIOManager(base_dir=tmp_path, db_name="fake", package=test_pkg)
 
 
-def test_error_when_handling_view_without_metadata(pudl_sqlite_io_manager_fixture):
+def test_migrations_match_metadata(tmp_path, monkeypatch):
+    pkg = Package.from_resource_ids()
+    monkeypatch.chdir(Path(__file__).parent.parent.parent)
```
cmgosnell, line 222: why do you need to do this `chdir`?

test/unit/io_managers_test.py
```diff
@@ -204,40 +206,53 @@ def test_missing_schema_error(sqlite_io_manager_fixture):
...
+    # alembic knows to use PudlPaths().pudl_db - so we need to set PUDL_OUTPUT env var
+    monkeypatch.setenv("PUDL_OUTPUT", tmp_path)
+    alembic.config.main(["upgrade", "head"])
+
+    PudlSQLiteIOManager(base_dir=tmp_path, db_name="pudl", package=pkg)
```
cmgosnell, line 227: so this guy is running `compare_metadata` which is what is actually doing the checking?? why wasn't this doing it before in the `pudl_sqlite_io_manager_fixture`??

cmgosnell: or is it `alembic.config.main(["upgrade", "head"])` that is doing the checking? or the combo or the upgrade/hear w/ generating the io manager?

cmgosnell (commented):
I left some clarifying questions. (lol maybe i am not the right person to review this! but i learn things so maybe it good)

jdangerx:
I added comments etc. to hopefully address your questions in a more durable way - lmk what you think!

cmgosnell (approved):
okay okay so we are doing the second if option in the `test_migrations_match_metadata`. initialize the db w/ alembic then creat the IO manager. okay that makes sense to me

PR #3029, opened 2023-11-08
Rename `heat_rate_mmbtu_mwh -> unit_heat_rate_mmbtu_per_mwh`... again

bendnorman:
# PR Overview
We renamed the `heat_rate_mmbtu_mwh` column to `unit_heat_rate_mmbtu_per_mwh ` but merged the changes into a branch that didn't get merged in the main feature branch `rename-core-assets`. This PR rebases the changes from https://github.com/catalyst-cooperative/pudl/pull/2917 onto `rename-core-assets`.

# PR Checklist


migrations/versions/1ceb9897fd34_add_data_maturity_to_eia923m_tables.py
zaneselvans, line 1: What's up with there being several migration versions in here changing? Including older ones? Is this search-and-replace gone wild? It seems like the older migrations might actually need to refer to the older column name.

Alternatively, should we wipe all the migrations periodically to keep too many layers from building up? Since I don't think we actually intend for it to be possible to migrate between all database versions.

zaneselvans (changes requested):
The only question I have is about the alembic migrations that have been changed. Should they really all be getting changed piecemeal like this? Or should the old ones keep referring to the old column name since... old DBs will have the old column name.

Or should we just wipe the migrations out and start anew?

bendnorman:
I've occasionally been clobbering and recreating the migrations when branches off of `rename-core-assets` start to diverge. I figured this was ok because we'll probably clobber and recreate a revision once we merge `rename-core-assets` into `dev`.

I don't think there is a ton of value keeping all of these small revision changes. Because we're not maintaining the schema of a production database, migrations are mostly helpful on single branches when you adjusting schemas and don't want to rerun the entire ETL to test your changes.

zaneselvans:
My concern is that it seems like there's a pre-existing auto-generated migration that's being edited here to use the new column name, which seems a bit odd -- shouldn't the old migration use the old column name?

bendnorman:
Good catch! I must have been a little too trigger-happy and accidentally renamed the columns in old revision files.

I deleted `1ceb` and `3313` because they are from `dev` and couldn't be easily merged with the "base" revision on the `rename-core-assets` branch. The changes in those two revisions are captured in `9ccf`.

zaneselvans (approved):

PR #3045, opened 2023-11-14
Fix the override paths when running in github actions

rousik:
This is an independent change that is part of https://github.com/catalyst-cooperative/pudl/pull/2825. Trying to separate it to ensure that this doesn't affect codecov results.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -323,8 +323,8 @@ def configure_paths_for_tests(tmp_path_factory, request):
     gha_override_input = False
     gha_override_output = False
     if os.environ.get("GITHUB_ACTIONS", False):
-        gha_override_input = "PUDL_INPUTS" not in os.environ
-        gha_override_output = "PUDL_OUTPUTS" not in os.environ
+        gha_override_input = "PUDL_INPUT" not in os.environ
+        gha_override_output = "PUDL_OUTPUT" not in os.environ
         logger.info(
             "Running in GitHub Actions environment, using"
             f" temporary input dir: {gha_override_input}, and"


zaneselvans (approved):

test/conftest.py
```diff
@@ -323,8 +323,8 @@ def configure_paths_for_tests(tmp_path_factory, request):
     gha_override_input = False
     gha_override_output = False
     if os.environ.get("GITHUB_ACTIONS", False):
-        gha_override_input = "PUDL_INPUTS" not in os.environ
-        gha_override_output = "PUDL_OUTPUTS" not in os.environ
+        gha_override_input = "PUDL_INPUT" not in os.environ
+        gha_override_output = "PUDL_OUTPUT" not in os.environ
```
zaneselvans, line 327: Sigh. Sorry @rousik I shouldn't have merged this. I thought it was just fixing a typo but actually the subsequent logic seems broken. The docstring doesn't explain what the desired behavior is, so I don't know what it's supposed to be doing.

* The logging below will always print out `False` for both the input and output directories, since `PUDL_INPUT` and `PUDL_OUTPUT` are set in the GitHub actions workflow file, which doesn't seem like the desired behavior.
* I think that overriding the value of `PUDL_INPUT` that's set in the workflow will mean that the action has to repopulate the datastore (downloading multiple GB of data) every time the action runs, since the data that's restored from the GitHub Actions cache is put in a known location (the `PUDL_INPUT` defined in the workflow) not the randomly generated temporary directory.
* The first `if gha_override_output` statement only configures the **input** directory, which seems like it might be a cut-and-paste error. Also I think we'll never enter the body of this if statement since at this point `gha_override_output` can only ever be `False`, and we don't run the CI with `--tmp-data`.
* The final logging message says that the unit tests are starting, but the fixture is session scoped. Why would this only be confined to the unit tests?

rousik: Yeah, this is actually fixing a typo. In case env variables are set (which I think is pretty much every time), overrides should not be used. So I think we may have been using incorrect logic in ghas until now.

zaneselvans: I think I've fixed it on the `pydantic-v2` branch where it was breaking things.

PR #3051, opened 2023-11-15
Refactor PUDL to use Pydantic v2

zaneselvans:
# PR Overview

## Refactor PUDL to use Pydantic v2
* Fix some unit test failures that were incidental -- due to differences in the way Pydantic v2 converts types when multiple basic types are acceptable (v2 will generally preserve the types, if they match any of the allowable types)
* Create the specified `PUDL_INPUT` and `PUDL_OUTPUT` directories in our docs build environments. The new `PotentialDirectoryPath` requires that the path either be a directory, or be a path where you could immediately create a new directory (without needing to create parent directories).
* Replace `model.dict()` with `model.model.dump()` across the board.
* Suppress warnings coming from Pydantic v1 syntax in other packages.
* Switch to explicitly using `@classmethod` decorators for pydantic validators where appropriate.
* Swapped a lot of `root_validator` for `model_validator`
* Swapped a lot of `validator` for `field_validator` and in some cases turned them into `model_validator(mode="after")` if they were accessing many attributes and not mutating the class.
* Tried to use more immediately legible names for the parameters being validated in many of the validators (rather than `v` or `value` or `values` -- since most of our validators only check a single attribute).
* Switched a bunch of constrained type factories like `constr` or `conint` for `Annotated[]` types.

## Other deprecations and warnings:
* Replaced deprecated `df.applymap()` with `df.map()`
* Remove the zombie Zenodo sandbox DOIs which got wiped.
* Suppress the 50,000 warnings we're getting about inconsistent date formats in the EIA Bulk Electricity data.
* Don't use deprecated `pd.unique()` on types it will stop working with soon.
* Wrap inputs to `pd.ExcelFile()` with `BytesIO` as indicated by deprecation warnings.

## Outstanding Issues/Questions:
* Not sure if the new `PudlPaths` / `PotentialDirectoryPath` behavior is actually ideal. But it works and is simple to define.
* In retrospect, not sure if the changes that got made in the settings classes were necessary, given that we didn't end up trying to turn them into Dagster's `ConfigurableResurce` classes... but it's working as it is.

# PR Checklist


zaneselvans:
Woo! I got the unit and integration tests to pass locally.

src/pudl/settings.py
```diff
@@ -49,9 +54,65 @@ class GenericDatasetSettings(BaseModel):
...
+
+            # If partition is None, default to working_partitions
+            if getattr(self, name) is None:
+                setattr(self, name, working_partitions)
+
+            nonworking_partitions = list(set(partition) - set(working_partitions))
+            if nonworking_partitions:
+                raise ValueError(f"'{nonworking_partitions}' {name} are not available.")
+            setattr(self, name, sorted(set(partition)))
+        return self
```
bendnorman, line 109: Do we actually need to be mutating the object in this validator? We call `setattr` for two reasons:
1. If `partition` is `None. This will never be true because the default partition values will be used right?
2. Sorting the partitions. I don't think we have code that expects the partitions to be sorted.

Just my first thoughts, still need to play around with the code.

zaneselvans: I guess you're right. If every child class sets the default partition values based on the working partitions, then this is superfluous. Let me try making this change...

zaneselvans: It turns out this is currently how we're deduplicating input partitions.

Also, it woule be nice if we didn't have to reference `data_source.working_partitions` both when setting the data source and the defaults for each partition, since just having the data source should be enough to figure out what the default values are.

If we did it with `mode="before"` then when we get `None` for the partitions, we should be able to just fill it in with the working partitions from the class attribute.

src/pudl/settings.py
```diff
@@ -689,26 +743,20 @@ def _convert_settings_to_dagster_config(d: dict) -> None:
...
             _convert_settings_to_dagster_config(v)
         else:
-            try:
-                d[k] = Field(type(v), default_value=v)
-            except DagsterInvalidDefinitionError:
-                # Dagster config accepts a valid dagster types.
-                # Most of our settings object properties are valid types
-                # except for fields like taxonomy which are the AnyHttpUrl type.
-                d[k] = Field(Any, default_value=v)
+            d[k] = DagsterField(type(v), default_value=v)
```
bendnorman, line 746: I don't think my doc string for `_convert_settings_to_dagster_config()` is accurate. This function does not create Noneable Dagster fields and instead sets the default value of the partition. Therefore I don't think we need to be setting the partition values to the default if partition is `None` in GenericDatasetSettings.

src/pudl/settings.py
bendnorman: If we want to keep the values sorted we could programmatically create getters for each partition attribute that return sorted values?

src/pudl/settings.py
zaneselvans: Meaning we should just do:
```py
d[k] = DagsterField(type(v))
```

bendnorman: I think we can leave it as is we just need to update the doc string.

src/pudl/settings.py
bendnorman: I agree I'd be nice if we didn't have to reference `data_source.working_partitions` both when setting the data source and the defaults for each partition. We could probably create the partition attributes from the data source information too though might not be worth the complexity. Something to think about if we want to do broader rewrite of our settings validation.

Instead of allowing duplicate partitions we could throw a validation error instead?

zaneselvans: Okay, now this is just raising errors if anything is wrong. Hopefully we don't have any waiting for us out there in the real world!

src/pudl/settings.py
zaneselvans: Okay, I tried to update the docstring to reflect my understanding of what's going on here, let me know if I am wrongo.

src/pudl/metadata/classes.py
```diff
@@ -1778,15 +1717,23 @@ def _check_foreign_keys(cls, value):  # noqa: N805
...
         for key in ("keywords", "contributors", "sources", "licenses"):
-            values[key] = _unique(
-                values[key], *[getattr(r, key) for r in values["resources"]]
-            )
-        return values
+            package_value = getattr(self, key)
+            resource_values = [getattr(resource, key) for resource in self.resources]
+            deduped_values = _unique(package_value, *resource_values)
+            setattr(self, key, deduped_values)
+        return self
```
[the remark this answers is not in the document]
zaneselvans: However, this creates an infinite recursion when `validate_assignment` is enabled for the model because of [changes in how model validation works](https://docs.pydantic.dev/latest/migration/#changes-to-validators).

For the moment I've disabled `validate_assignment` but that doesn't seem ideal.

src/pudl/metadata/classes.py
```diff
@@ -303,53 +212,71 @@ def _validator(*names, fn: Callable) -> Callable:
...
+# PUDL Metadata Classes
+########################################################################################
+class PudlMeta(BaseModel):
+    """A base model that configures some options for PUDL metadata classes."""
+
+    model_config = ConfigDict(
+        extra="forbid",
+        validate_all=True,
+        validate_assignment=True,
+    )
```
zaneselvans, line 238: This class is a simplified generic version of the old `Base` class, which sets some model configurations shared across all of our metadata classes below.

bendnorman (approved):
LGTM! Lots of good upgrading and clean up. Thanks @zaneselvans!

pyproject.toml
```diff
@@ -258,6 +252,15 @@ filterwarnings = [
...
     "ignore:The Shapely GEOS version:UserWarning:geopandas[.*]",
     "ignore:Unknown extension:UserWarning:openpyxl.worksheet[.*]",
+    "ignore:The `__fields__` attribute is deprecated:pydantic.PydanticDeprecatedSince20:unittest.mock",
+    "ignore:The `__fields_set__` attribute is deprecated:pydantic.PydanticDeprecatedSince20:unittest.mock",
+    "ignore:The `__fields__` attribute is deprecated:pydantic.PydanticDeprecatedSince20:pydantic.main",
+    "ignore:The `__fields_set__` attribute is deprecated:pydantic.PydanticDeprecatedSince20:pydantic.main",
+    "ignore:The `update_forward_refs` method is deprecated:pydantic.PydanticDeprecatedSince20:pydantic.main",
+    "ignore:Support for class-based `config` is deprecated:pydantic.PydanticDeprecatedSince20:pydantic._internal._config",
+    "ignore:Pydantic V1 style `@validator` validators are deprecated:pydantic.PydanticDeprecatedSince20:ferc_xbrl_extractor.instance",
+    "ignore:The `update_forward_refs` method is deprecated:pydantic.PydanticDeprecatedSince20:ferc_xbrl_extractor.taxonomy",
```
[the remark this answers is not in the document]
zaneselvans: Okay I lied that point seemed like Saturday apparently.

src/pudl/metadata/classes.py
```diff
@@ -139,151 +163,36 @@ def _get_jinja_environment(template_dir: DirectoryPath = None):
...
+PositiveFloat = Annotated[float, pydantic.Field(ge=0, strict=True)]
 """Positive :class:`float`."""
 
-Email = pydantic.EmailStr
-"""String representing an email."""
-
-HttpUrl = pydantic.AnyHttpUrl
-"""Http(s) URL."""
-
-
```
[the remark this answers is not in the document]
zaneselvans: It seemed more straightforward to me to just use the pydantic types directly in the classes where these types are used, rather than renaming them and creating a layer of indirection.

src/pudl/metadata/classes.py
```diff
@@ -749,32 +688,33 @@ def encode(self, col: pd.Series, dtype: type | None = None) -> pd.Series:
...
-class ForeignKeyReference(Base):
+class ForeignKeyReference(PudlMeta):
     """Foreign key reference (`resource.schema.foreign_keys[...].reference`).
 
     See https://specs.frictionlessdata.io/table-schema/#foreign-keys.
     """
 
     resource: SnakeCase
-    fields_: StrictList(SnakeCase) = pydantic.Field(alias="fields")
+    fields: StrictList(SnakeCase)
```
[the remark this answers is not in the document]
zaneselvans: Yes.

src/pudl/settings.py
```diff
@@ -454,7 +464,7 @@ class FercGenericXbrlToSqliteSettings(BaseSettings):
         disabled: if True, skip processing this dataset.
     """
 
-    taxonomy: AnyHttpUrl
+    taxonomy: str
```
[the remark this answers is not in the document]
zaneselvans: Maybe we'll just see how their Pythonic / Pydantic parametrization evolves.

src/pudl/settings.py
```diff
@@ -671,44 +675,42 @@ def from_yaml(cls, path: str) -> "EtlSettings":
...
+    """Recursively convert a dictionary of dataset settings to dagster config in place.
 
-    For each partition parameter in a GenericDatasetSettings subclass, create a Noneable
-    Dagster field with a default value of None. The GenericDatasetSettings
-    subclasses will default to include all working paritions if the partition value
-    is None. Get the value type so dagster can do some basic type checking in the UI.
+    For each partition parameter in a :class:`GenericDatasetSettings` subclass, create a
+    corresponding :class:`DagsterField`. By default the :class:`GenericDatasetSettings`
+    subclasses will default to include all working paritions if the partition value is
+    None. Get the value type so dagster can do some basic type checking in the UI.
```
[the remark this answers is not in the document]
zaneselvans: Famous last words!

PR #3058, opened 2023-11-18
Fix full build notification logic

bendnorman:
# PR Overview

I noticed my full build `271ffc37d3e7a7253beb9bb74ec2ea6585b92e0a-rename-core-assets` passed all of the validation tests but the `pudl-deployments` channel was received a failure notification. I think this happened because #3018 changed the script to catch any errors that happened during the data distribution and datasette publishing:

https://github.com/catalyst-cooperative/pudl/blob/2493454122018f396b89c9566398ea4db5c7b7c9/docker/gcp_pudl_etl.sh#L92-L116

This logic does not work if we have a successful build that isn't kicked off from the `dev` branch. The first if statement will evaluate to true because the build succeeded, the inner two if statements will evaluate to false because the branch is not dev. The final if statement to determine if slack should be notified will also be false because the previous two if statements exit codes are non-zero. 

To resolve this issue, I created a `ETL_SUCCESS` variable to track failures during the distribution and publishing steps.

# PR Checklist


jdangerx (approved):
Storing that value as a variable makes way more sense than the weird house of cards I was building, thanks!!

Issue #3064, opened 2023-11-20
add a special calculation correction when a sizable minority of utilities report in a non-standard way


cmgosnell:
```[tasklist]
### Tasks
- [ ] identify list of `xbrl_factoid`'s that have a sizable minority of utils reporting non-standardly... see notes in #2703 for some specific examples
- [x] add a step in `reconcile_intertable_calculations` (or before) that id's the non-standard utils
- [x] add special-correction `xbrl_factoid` values in the concatenated data tables for the non-standard utils
- [x] add a special-correction for all of these id-ed `xbrl_factoid`'s that have non-standard reporter into the calc components table
```

## questions
* I was previously imagining manually compiling the list of factoids that need this special correction... but all of the code required to find these special factoids to fix are required for the fix itself. 
   * it seems i do definitely need to enumerate the list of facts to fix here. with both the parent and missing child's dimensions bc nothing in these calcs are ever straightforward it seems.
* how to deal with updating `exploded_calcs`? it is currently a `cached_property` of `Exploder` and fully independent of the data tables (the concatenated data itself). this update will most likely couple the data and `exploded_calcs`. Unless I hard-code the facts to fix.... then i think I could make these updates independently.
   *  bc the facts to fix should almost certainly be enumerated, we don't need to worry about this coupling bc we can use the mannually compiled facts to fix to generate new correction records in the `exploded_calcs` table.

PR #3078, opened 2023-11-22
wip: add data corrections for sizable minority utilities

cmgosnell:
# PR Overview
Working on  #3064

# PR Checklist

### DIFF
--- a/src/pudl/output/ferc1.py
+++ b/src/pudl/output/ferc1.py
@@ -1552,8 +1552,10 @@ class Exploder:
         """
         exploded = (
             self.initial_explosion_concatenation(tables_to_explode)
-            .pipe(self.reconcile_intertable_calculations)
-            .pipe(self.calculation_forest.leafy_data, value_col=self.value_col)
+            # .pipe(self.indentify_and_correct_sizable_minority_utility_reporting)
+            .pipe(self.reconcile_intertable_calculations).pipe(
+                self.calculation_forest.leafy_data, value_col=self.value_col
+            )
         )
         # Identify which columns should be kept in the output...
         # TODO: Define schema for the tables explicitly.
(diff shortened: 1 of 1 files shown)


src/pudl/output/ferc1.py
```diff
@@ -1159,9 +1174,10 @@ def _aggregatable_dimension_tags(
 
 def exploded_table_asset_factory(
     root_table: str,
-    table_names_to_explode: list[str],
+    table_names: list[str],
```
cmgosnell, line 1177: i changed this so it's easier to copy/paste `explosion_args` in `create_exploded_table_assets` into a notebook to do:
```
args_explode = explosion_args[1]
args_generic = {
    "metadata_xbrl_ferc1":meta_w_dims,
    "calculation_components_xbrl_ferc1":calc_components,
    "tags":tags_df
}

exploder = Exploder(
    **args_explode|args_generic
)
```

src/pudl/output/ferc1.py
```diff
@@ -1495,7 +1583,7 @@ def calculation_forest(self: Self) -> "XbrlCalculationForestFerc1":
         )
 
     @cached_property
-    def other_dimensions(self: Self) -> list[str]:
+    def dimensions(self: Self) -> list[str]:
```
cmgosnell, line 1586: i changed this for brevity + because other didn't really add any information.

src/pudl/output/ferc1.py
```diff
@@ -1632,6 +1719,104 @@ def initial_explosion_concatenation(
...
+        The data corrections are identified by calculating the absolute difference
+        between the reported value and calculable value from the standard set of
+        subcomponents (via :func:`pudl.transform.ferc1.calculate_values_from_components`)
+        and finding the child factiods that have the same value as the absolute difference.
+        This indicates that the calculable parent factiod is off by that cooresponding
+        child fact.
+
+        Relatedly, :meth:`add_sizable_minority_corrections_to_calcs` adds these
+        :attr:`self.off_by_facts` to :attr:`self.exploded_calcs`.
+        """
```
cmgosnell, line 1743: this is my (current) main attempt at an explanation here.

src/pudl/output/ferc1.py
```diff
@@ -1632,6 +1719,104 @@ def initial_explosion_concatenation(
...
+            value_col=self.value_col,
+        )
+
+        off_by = pd.DataFrame(self.off_by_facts)
+        # we calculate the non-abs diff here because many times the
+        # bad utility reporters include components that are not in the
+        # stock calculation so we need to remove it. the value of the
+        # correction record needs to be the raw diff.
+        not_close = (
+            calculated_df[calculated_df.abs_diff != 0]
```
e-belfer, line 1763: We use `np.isclose` elsewhere, why not here?

cmgosnell:
> Blocking: Confused about the presence of environment changes in this PR.
I was also confused about this! I asked @zaneselvans  and it seems like small changes to the lock files with minor version updates should be expected in general. I believe this is because we have so many dependencies that it is likely that from day to day we'll see small version updates.

> Non-blocking: We calculate, add correction factoids, then calculate everything again. I am not sure whether it'd just be simpler to calculate once, add the correction factoids and change the corrected abs_diffs to 0 before proceeding to reconciling calcs?

I was wondering about this. I think I could break up `reconcile_intertable_calculations` into two steps so `boom` could include something like:
```
    exploded = (
            self.initial_explosion_concatenation(tables_to_explode)
            .pipe(self.calculate_values_from_components)
            .pipe(self.add_sizable_minority_corrections) # which would have to correct the abs_diffs...
            .pipe(self.reconcile_intertable_calculations)
            .pipe(self.calculation_forest.leafy_data, value_col=self.value_col)
        )
```

I will try this. I'm not sure this will be simpler even though it would mean we would run `pudl.transform.ferc1.calculate_values_from_components` only once.

src/pudl/output/ferc1.py
cmgosnell: ah good point.

in a previous commit i used the `is_not_close` from the ungrouped error frequency metric... but that felt like a little much setup wise so i reverted to a simpler selection here. but i'll conver to `~np.isclose` bc that is more in line with `is_not_close`/what we are doing in other places.

removed in [this commit](https://github.com/catalyst-cooperative/pudl/pull/3078/commits/b7328a05f9d414b22148e1de965713170e0413e2#diff-0c9cd108cbf26a3701799dcb4888a7d09d7c657eb9de859f4383b631e4042f7aL1742-L1747)

```python
       # add the ungrouped is_not_close column
        calculated_df["is_not_close"] = pudl.transform.ferc1.ErrorFrequency(
            by="ungrouped",
            is_close_tolerance=self.group_metric_checks.is_close_tolerance.error_frequency,
            metric_tolerance=self.group_metric_checks.group_metric_tolerances.ungrouped.error_frequency,
        ).is_not_close(calculated_df)
```

src/pudl/output/ferc1.py
```diff
@@ -1632,6 +1719,104 @@ def initial_explosion_concatenation(
...
+                ~self.exploded_calcs.is_within_table_calc
+                & ~self.exploded_calcs.is_total_to_subdimensions_calc
+            ],
+            data=exploded,
+            calc_idx=list(NodeId._fields),
+            value_col=self.value_col,
+        )
+
+        off_by = pd.DataFrame(self.off_by_facts)
+        # we calculate the non-abs diff here because many times the
```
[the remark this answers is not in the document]
cmgosnell: i'm actually going to remove this comment altogether bc i'm moving the `diff` assignment into calculate_values_from_components and this comment is below:

```python
            # the correction value is the diff - not the abs_diff from not_close or value_col
            # from off_by_fact bc often the utility included a fact so this diff is negative
```

cmgosnell:
okay @e-belfer I was able to address the non-blocking concern so I tink its ready for a re-review.

src/pudl/output/ferc1.py
```diff
@@ -1075,7 +1075,13 @@ class NodeId(NamedTuple):
 
 
 class OffByFactoid(NamedTuple):
-    """A factoid which is reported incorrectly as off-by another factoid."""
+    """A calculated factoid which is off by one other factoid.
+
+    A factoid where a sizeable majority of utilities using a non-standard and
```
e-belfer, line 1080: "are using a non-standard"?

src/pudl/output/ferc1.py
```diff
@@ -1815,7 +1833,26 @@ def add_sizable_minority_corrections(
...
             f"Adding {len(data_corrections)} from {len(bad_utils)} utilities that report differently."
         )
-        return pd.concat([exploded, data_corrections]).reset_index(drop=True)
+        dtype_calced = {
+            col: dtype
+            for (col, dtype) in calculated_df.dtypes.items()
+            if col in data_corrections.columns
+        }
+        corrected = (
+            pd.concat([calculated_df, data_corrections.astype(dtype_calced)])
```
e-belfer, line 1842: Is this to get rid of NA columns for the concatenations?

e-belfer (approved):
One tiny documentation correction but otherwise this looks much cleaner to me.

src/pudl/output/ferc1.py
cmgosnell: yes! #3125

PR #3086, opened 2023-11-25
Clickety click

zaneselvans:
# PR Overview

I've been messing around with the [Click CLI framework](https://click.palletsprojects.com/en/8.1.x/) and it's pretty great. Much more ergonomic and compact, and we depend on it indirectly already, so I though I would (for fun) migrate our `argparse` based CLI modules and do a little related cleanup because it feels good.

# PR Checklist


src/pudl/etl/cli.py
zaneselvans, line 1: I renamed this module so that we were using the same naming pattern here and in the `ferc_to_sqlite` CLI. And also because the CLI subpackage was very lonely.

Once all of our existing CLIs are using Click (or have been retired, where appropriate) I think I may create a single top-level `cli.py` that unifies them all into a `pudl` command with several subcommands.

src/pudl/analysis/service_territory.py
```diff
@@ -536,77 +551,145 @@ def plot_all_territories(
...
-    pudl_engine = sa.create_engine(PudlPaths().pudl_db)
-    # Load the US Census DP1 county data:
-    county_gdf = pudl.etl.defs.load_asset_value(AssetKey("county_censusdp1"))
-
-    kwargs_dicts = [
-        {"entity_type": "util", "limit_by_state": False},
-        {"entity_type": "util", "limit_by_state": True},
-        {"entity_type": "ba", "limit_by_state": True},
-        {"entity_type": "ba", "limit_by_state": False},
-    ]
```
zaneselvans, line 591: Rather than hard-coding that the script runs through all combinations of params, let the user select on the command line.

src/pudl/analysis/service_territory.py
```diff
@@ -424,25 +440,25 @@ def dagster_compile_geoms(
 ################################################################################
 # Functions for visualizing the service territory geometries
 ################################################################################
-def plot_historical_territory(gdf, id_col, id_val):
+def plot_historical_territory(
+    gdf: gpd.GeoDataFrame,
+    id_col: str,
+    id_val: str | int,
+) -> None:
```
zaneselvans, line 447: I have not yet tested any of the old plotting infrastructure.

src/pudl/analysis/service_territory.py
```diff
@@ -255,22 +256,41 @@ def get_territory_geometries(
...
-        # States & counties only remain at this point if we didn't dissolve
-        for col in ("county_id_fips", "state_id_fips"):
-            # pandas.NA values are not compatible with Parquet Strings yet.
-            gdf[col] = gdf[col].fillna("")
-    limited = ""
-    if limit_by_state:
-        limited = "_limited"
-    # Save the geometries to a GeoParquet file
-    fn = f"{entity_type}_geom{limited+dissolved}.pq"
-    gdf.to_parquet(fn, index=False)
```
zaneselvans, line 273: This got much simpler since we last touched it, and now it kind of "just works" with a small number of records in each row group.  Not sure what the memory blowup was, but it's gone now.

zaneselvans (commented):
This PR has a lot of line changes, but most of it is either deletions or simple refactors from `argparse` to `click`.  Test coverage is a bit better and lots of little crufty bits have been removed.

jdangerx (commented):
I've reviewed everything except for the service territory changes so far. Nothing blocking here!

I will need a bit more time to investigate the service territory changes / think about why the test directory seems to be messed up for you (I'm running integration tests locally now to see if I can repro, but obviously that's a pretty long iteration cycle). We could split those changes into a separate PR and merge everything else in now if you want!

pyproject.toml
```diff
@@ -300,3 +291,34 @@ prettier = ">=3.0"
...
+source = [
+    "src/pudl/",
+    "test/integration/",
+    "test/unit/",
+]
+omit = [
+    # Never hit by integration tests:
+    "src/pudl/validate.py",
+]
+
```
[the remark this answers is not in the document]
zaneselvans: Yeaaaah, this was definitely annoying and also seems like note great design. Agree that comments would be helpful. Will add.

src/pudl/workspace/setup.py
```diff
@@ -26,6 +24,12 @@ class PudlPaths(BaseSettings):
     pudl_output: PotentialDirectoryPath
     model_config = SettingsConfigDict(env_file=".env", extra="ignore")
 
+    @model_validator(mode="after")
+    def create_directories(self: Self):
+        """Create PUDL input and output directories if they don't already exist."""
+        self.input_dir.mkdir(parents=True, exist_ok=True)
+        self.output_dir.mkdir(parents=True, exist_ok=True)
+
```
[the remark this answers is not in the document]
zaneselvans: The directory creation has to be managed somewhere, and right now it's all over the place:
- In `docs/conf.py` I create directories if `$READTHEDOCS`.
- In our github workflows I have to create the directories 3 times: for `ci-docs`, `ci-unit` and `ci-integration`
- We have to create the directories inside the temporary directory when we run pytest locally.
- Users would need to create their input/output directories manually wherever they're supposed to be.

Avoiding all that flaky cruft and assuming that these paths have been set to something reasonable and creating them if they don't already exist sounds easier to keep track of -- and if there's a file there already, or the path isn't writable, then it'll fail.  I would love to get rid of all those freaking `mkdir -p` commands.

src/pudl/workspace/datastore.py
```diff
@@ -411,114 +412,26 @@ def get_zipfile_file_names(self, zip_file: zipfile.ZipFile):
...
-        parts = dstore.get_datapackage_descriptor(single_ds).get_partitions()
+        partitions = dstore.get_datapackage_descriptor(single_ds).get_partitions()
 
-        print(f"\nPartitions for {single_ds} ({dstore.get_doi(single_ds)}):")
-        for pkey in sorted(parts):
-            print(f'  {pkey}: {", ".join(str(x) for x in sorted(parts[pkey]))}')
-        if not parts:
+        print(f"\nPartitions for {single_ds} ({ZenodoFetcher().get_doi(single_ds)}):")
+        for partition_key in sorted(partitions):
+            # try-except required because ferc2 has parts with heterogenous types that
```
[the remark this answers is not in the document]
zaneselvans: Frictionless is promiscuous -- you can add extra stuff wherever you want, and we added `parts` so it's up to us to define it. Recently @e-belfer has also been messing around with the types used in partitions over in `pudl-archiver` land. We should probably talk about it from both ends of the process together.

pyproject.toml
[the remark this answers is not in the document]
zaneselvans: Explanation added.

.codecov.yml
```diff
@@ -1,9 +1,12 @@
 ---
 coverage:
   range: 70..100
-  round: down
+  round: nearest
   precision: 1
 
+ignore:
+  - "src/pudl/validate.py"
+
```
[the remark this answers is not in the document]
zaneselvans: I've tried many times to integrate collecting coverage into `make nuke` (which uses Dagster to run the ETL) and it never seems to come out right. Probably for reasons related to why Jan has had such a hard time collecting coverage in his CI revamp. So I've kind of given up on it.

I think the general plan is to move the data validation into the ETL process itself, rather than having it be a post-processing / testing step, which would bring this (or whatever it evolves into) into the fold of the main coverage collection system. Also this file has the *most* uncovered lines of any file and I hate seeing it, knowing I can never fix it.  I realize this is not particularly rational.  I just want us to be an A- student, not a B+ student okay man?

test/integration/console_scripts_test.py
```diff
@@ -1,20 +1,47 @@
...
 
-@pytest.mark.parametrize("script_name", PUDL_SCRIPTS)
+
+# @pytest.mark.skip("Test exceeds memory or disk limits of GitHub runner.")
 @pytest.mark.script_launch_mode("inprocess")
-def test_pudl_scripts(script_runner, script_name):
-    """Run each console script in --help mode for testing."""
-    ret = script_runner.run([script_name, "--help"], print_result=False)
+def test_pudl_service_territories(
+    script_runner, tmp_path: Path, pudl_engine: sa.Engine
```
zaneselvans: Duh, I can just use a temporary directory here!

test/integration/console_scripts_test.py
```diff
@@ -1,20 +1,47 @@
...
+):
+    """CLI tests specific to the pudl_service_territories script.
+
+    Depends on the ``pudl_engine`` fixture to ensure that the censusdp1tract.sqlite
+    database has been generated, since that data is required for the script to run.
+    """
+    out_path = tmp_path / "balancing_authority_geometry.parquet"
+    assert not out_path.exists()
+    command = (
+        f"pudl_service_territories --entity-type ba -y 2022 --no-dissolve -o {tmp_path}"
```
zaneselvans: Hopefully doing this on a single year will make it easy enough for the runner.

src/pudl/analysis/service_territory.py
```diff
@@ -536,77 +564,161 @@ def plot_all_territories(
...
+    type=click.IntRange(min=2001),
+    default=[],
+    multiple=True,
+    help=(
+        "Limit service territories generated to those from the given year. This can "
+        "dramatically reduce the memory and CPU intensity of the geospatial "
+        "operations. Especially useful for testing. Option can be used multiple times "
+        "toselect multiple years."
+    ),
+)
```
zaneselvans, line 607: Added the option to restrict the geometry compilation to a subset of the available years to keep resource usage down when necessary.

test/integration/console_scripts_test.py
```diff
@@ -1,20 +1,57 @@
...
+@pytest.mark.parametrize(
+    "command,filename",
+    [
+        (
+            "pudl_service_territories --entity-type ba -y 2022 --no-dissolve -o ",
+            "balancing_authority_geometry.parquet",
+        ),
+        (
+            "pudl_service_territories --entity-type util -y 2022 --dissolve -o ",
+            "utility_geometry_dissolved.parquet",
```
zaneselvans: With just a single year of data, we can actually test both ba/util and dissolved/undissolved geometries in a few seconds.

zaneselvans:
@jdangerx I've resolved the comments where it seemed like everything was okay, and responded to a few that weren't.

src/pudl/analysis/service_territory.py
zaneselvans: Created issue #3117 for this work.

docker/gcp_pudl_etl.sh
```diff
@@ -44,7 +43,7 @@ function run_pudl_etl() {
         -n auto \
         --gcs-cache-path gs://internal-zenodo-cache.catalyst.coop \
         --etl-settings $PUDL_SETTINGS_YML \
-        --live-dbs test/validate
+        --live-dbs test/validate \
```
jdangerx, line 46: non-blocking: Was this missing `\` breaking things?

jdangerx (approved):
OK - given that the service territory stuff is still in an interim state of "running but not verified," and the previous version was in the "completely broken" state, I have satisfied myself with reading what the code does and then mostly not worrying too much about it.

docker/gcp_pudl_etl.sh
zaneselvans: Yes. This is the fix that @bendnorman added on his tiny core-dump debugging PR that unfortunately didn't fix the core-dump, but did mean the script could actually run.

PR #3103, opened 2023-12-01
Hide diffs in lock files by default.

jdangerx:
Right now, we often get some lockfile updates in PRs - these are automatically generated based on the pyproject.toml and in the vast majority of cases we don't have to look at them, so we should hide them by default. We'll still be able to expand them if necessary.

I tested this by adding and removing `mypy` from the dev dependencies in `pyproject.toml`. This triggered some re-locking and WOW look the diffs don't show up unless you manually expand them!

image


zaneselvans (approved):
Looks good to me!

PR #3123, opened 2023-12-05
Replace ferc714 @multi_assert with asset factory 

rousik:
`@multi_asset` is a bit unwieldy and can prevent selective materialization of downstream assets. While this can be addressed for `@multi_asset` instances, I have opted to replace this with asset factory pattern which is also used elsewhere, because the raw ferc714 tables are truly independent and there's no need to bundle them together.

I suspect we should review other cases where we use `@multi_assets` and see what design pattern makes sense. In general, it would be valuable to ensure that `@multi_assets` we keep can be subsetted which should enable us to selectively materialize arbitrary assets from dagster-webserver UI.


src/pudl/extract/ferc714.py
```diff
@@ -2,7 +2,8 @@
 from collections import OrderedDict
 
 import pandas as pd
-from dagster import AssetOut, Output, multi_asset
+from dagster import AssetsDefinition, asset
+from tomlkit import table
```
jdangerx, line 6: What do you need this for?

jdangerx (changes requested):
Overall this looks good, barring the (possibly erroneous?) import. 

I personally think the asset factory is a nicer pattern than multi-assets, but I'll defer to @bendnorman's opinion on multi-asset vs. single-asset.

bendnorman:
I like this! I agree single assets are generally better than multi assets. They make it easier to subset and parallelize things.

src/pudl/extract/ferc714.py
rousik: I don't need this, just came across the hurdle of trying to run some downstream asset and failing that because dagster would refuse to materialize anything that had multi-asset upstream of it unless you manually select all of the constituents of multi-asset.

zaneselvans (approved):
Agree that we should use asset factories when they're simple like this, and this matches the patterns we've got elsewhere.

Off topic but... if you're excited about asset factories in general, there are 20+ individual assets currently defined at the top of `pudl.extract.ferc1` which could be consolidated into one or two asset factories.

src/pudl/extract/ferc714.py
rousik: Ah, I haven't realized that the comment specifically refers to this import. Huh, no idea how this came to be.

rousik:
> Agree that we should use asset factories when they're simple like this, and this matches the patterns we've got elsewhere.
> 
> Off topic but... if you're excited about asset factories in general, there are 20+ individual assets currently defined at the top of `pudl.extract.ferc1` which could be consolidated into one or two asset factories.

@zaneselvans do you mean the `SourceAssets` defined there that are constructed by iterating over the `TABLE_NAME_MAP_FERC1`? Those pretty much look like factory pattern already, because `SourceAssets` have pretty much no substance to them, other than specifying the right io manager to pull the data from the right sqlite file. I suppose this may eventually change once/if we integrate `ferc_to_sqlite` and `pudl_etl` into a single graph.

rousik:
> > Agree that we should use asset factories when they're simple like this, and this matches the patterns we've got elsewhere.
> > Off topic but... if you're excited about asset factories in general, there are 20+ individual assets currently defined at the top of `pudl.extract.ferc1` which could be consolidated into one or two asset factories.
> 
> @zaneselvans do you mean the `SourceAssets` defined there that are constructed by iterating over the `TABLE_NAME_MAP_FERC1`? Those pretty much look like factory pattern already, because `SourceAssets` have pretty much no substance to them, other than specifying the right io manager to pull the data from the right sqlite file. I suppose this may eventually change once/if we integrate `ferc_to_sqlite` and `pudl_etl` into a single graph.

After further reading, I suspect that you were referring to the assets that are in the `output/ferc1.py` which define the actual FERC1 assets with slight variations to how they're constructed from the incoming dfs?

jdangerx (approved):

jdangerx:
@zaneselvans maybe it makes sense to turn "factory-ize ferc1.py" into an issue? Seems like a good-first-issue to me.

zaneselvans:
@jdangerx I created #3147 and added it to the Community board.

Issue #2756, opened 2023-07-26
Automate the data release process


bendnorman:
Once #1973 and the implementation of #2517 are complete, we can move to data-only releases. There is some automation we'd like to create to make this process as smooth as possible.

The plan right now is to do a semi-manual data release based on the [`v2023.12.01` tag](https://github.com/catalyst-cooperative/pudl/releases/tag/v2023.12.01) and take notes on the process, so we can do an automatic release within the next 2 weeks containing all the post-rename tables.  It's just a draft at the moment, but the v2023.12.01 release will be available at:
* On Zenodo: https://zenodo.org/records/10275052
* With DOI: `10.5281/zenodo.10275052`

## Resolved Questions
* The auto-generated release notes on the GitHub release got truncated. Regenerating them manually did not truncate. I have modified our release configuration to exclude all of the bot PRs and hopefully we will never wait a year between releases, so this shouldn't be a problem anymore.

## Open Questions
* Do we want an "oops" wait time or required reviews on the release? I've already restricted it to only being possible to deploy from `main` and only if the tag starts with `v20*`.
* Do we or don't we want to push releases to PyPI? It's an obvious place to look for things, but we're publishing an application not a library. If we do publish release software to PyPI should we also continue updating on `conda-forge`? I'm inclined to find a low overhead but not guaranteed to work / be reproducible way to do this just so the package can ge installed without git reference gymnastics.
* If we continue doing software release, note that PyPI removes leading zeroes in the calendar version, turning `v2023.12.01` into `v2023.12.1` However, GitHub and the outputs on GCS & S3 don't do this, and the tag we applied on GitHub is `v2023.12.01` so this could result in some confusion / annoyance. **Do we want to use the no-leading-zeroes version of CalVer?**

## For Future Consideration
* Ideally we would probably have some more cautious flow-control in the release process. Like as it is once you push the tag, all of these things kick off, but there should really be dependencies. Right now if the build fails for some reason, then the data won't be released, but the software would still be pushed out.
[...]

PR #3124, opened 2023-12-05
Add GHA workflow for release-on-tag

zaneselvans:
# PR Overview

- Sets up a workflow to do GitHub (and possibly PyPI) releases automatically any time a tag is pushed starting with `v20` (matching calendar versions like `v2023.12.05` for the next 77 years).
- This includes a new job that attempts to build valid source & binary (wheel) distribution of the repo on every push, creating the files which would then be distributed if the right tag is pushed to `main`
- The publication of a GitHub release triggers the archiving of the PUD **repository** on Zenodo (through the GitHub-Zenodo integration app we have set up already), which is one of the goals of the automated releases. See #2756 
- We already have a build-on-tag action set up. The data release work (creating a new **data** archive on Zenodo, pushing to persistent locations on GCP/S3) will be done in the nightly build system.

# Questions / Issues
- In an ideal world it seems like none of the distribution would happen until after the build initiated by the tag had completed successfully. What kinds of issues/changes can creep into the system between getting a good nightly build, and tagging the associated commit for release?

# PR Checklist

### DIFF
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
   push:
     tags:
-      - "v**"
+      - "v20*"
   schedule:
     - cron: "0 6 * * 1-5" # Weekdays at midnight on MST
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,121 @@
+---
+name: release
+
+on: push
+
+jobs:
+  build-distribution:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - name: Build source and wheel distributions
+        run: |
+          python -m pip install --upgrade build twine
+          python -m build
+          twine check --strict dist/*
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
+
+  publish-test-pypi:
+    name: Upload release to PyPI Test Server
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v20')
+    needs:
+      - build-distribution
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://test.pypi.org/p/catalystcoop.pudl
+    permissions:
+      id-token: write
+    steps:
+      - name: Download all the dists
+        uses: actions/download-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
+      - name: Publish distribution to PyPI Test Server
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: https://test.pypi.org/legacy/
+
+  publish-pypi:
+    name: Upload release to PyPI
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v20')
+    needs:
+      - build-distribution
+    runs-on: ubuntu-latest
+    environment:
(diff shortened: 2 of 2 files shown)


.github/workflows/build-deploy-pudl.yml
```diff
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
   push:
     tags:
-      - "v**"
+      - "v20*"
```
jdangerx, line 6: Sadly it seems like we can't do a regex like `v\d{4}.\d{2}.\d{2`, just glob matching.

.github/workflows/release.yml
```diff
@@ -0,0 +1,121 @@
...
+          --repo '${{ github.repository }}'
+
+  notify-slack:
+    runs-on: ubuntu-latest
+    needs:
+      - publish-github
+    if: ${{ always() }}
+    steps:
+      - name: Inform the Codemonkeys
+        uses: 8398a7/action-slack@v3
```
jdangerx, line 104: non-blocking: I think we can use the official slack-github-action here, so long as we set `SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK` in the env as well.

.github/workflows/release.yml
```diff
@@ -0,0 +1,121 @@
...
+        uses: actions/download-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
+      - name: Publish distribution to PyPI Test Server
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: https://test.pypi.org/legacy/
+
+  #publish-pypi:
```
jdangerx, line 51: This is commented out since you're just testing that the pypi upload works right now, right?

.github/workflows/release.yml
```diff
@@ -0,0 +1,121 @@
+---
+name: release
+
+on: push
```
jdangerx, line 4: This works for testing, since you can just push to a branch, but...
- Do you want to run this on all pushes or just tags that match the pattern? Or pushes to branches that match a pattern?
- Do you want to add a workflow_dispatch trigger? Might be nice for debugging.

jdangerx (approved):
Yeah, let's get this into `main` so you can start testing. A few comments that don't block but might make your testing easier... 🤷

.github/workflows/release.yml
zaneselvans: Yes, I just want to make sure it all works first.

.github/workflows/release.yml
zaneselvans: I want to attempt to build the installable package (the first job below) on push so that if it breaks we know immediately and can fix it while we still have context.

The filter which prevents any of the actual publication from happening is in the first of the subsequent publishing jobs.  If we want to do filtering in the `on:` step then we need to create a separate `build-package.yml` workflow that just attempts to create the package as part of the CI, and duplicate that logic in here, which seemed less clear & concise.

.github/workflows/release.yml
zaneselvans: I have no idea how any of the slack notifications work. I'm just cutting and pasting from other jobs...

.github/workflows/release.yml
zaneselvans: As far as the `workflow_dispatch` I don't think it will help with the testing, since so much of what happens downatream is tied to having the version-compatible github ref to work with.  We use `setuptools_scm` to automatically generate a software version from the tags, and if we don't do that we've got a garbled mess of git stuff as the "version" which PyPI rejects (not sure what GitHub thinks about it).

Mostly what I need to test here is that these actions are hooked up to the tagging and we have all the authentication setup. I've used these actions in `ferc-xbrl-extractor` before so hopefully they'll work here as well..

.github/workflows/build-deploy-pudl.yml
zaneselvans: Yeah, I looked around a bit too.

PR #3127, opened 2023-12-06
Script to sync a local directory up to a Zenodo DOI

zaneselvans:
# PR Overview

Just struggling with the Zenodo API for the moment. Can easily do a manual upload of the files as well.

# PR Checklist

### DIFF
--- a/devtools/zenodo/zenodo_data_release.py
+++ b/devtools/zenodo/zenodo_data_release.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+import os
+from pathlib import Path
+
+import click
+import requests
+from pydantic import AnyHttpUrl
+
+from pudl.workspace.datastore import ZenodoDoi
+
+# All of these pertaining
+ZENODO_TOKEN: str = os.environ["ZENODO_SANDBOX_TOKEN_PUBLISH"]
+ZENODO_API_BASE: AnyHttpUrl = AnyHttpUrl("https://sandbox.zenodo.org/api")
+CONCEPT_DOI: ZenodoDoi = ZenodoDoi("10.5072/zenodo.5563")
+CONCEPT_REC_ID: str = CONCEPT_DOI.split(".")[-1]
+HEADERS = {"Content-Type": "application/json"}
+AUTH_PARAM = {"access_token": ZENODO_TOKEN}
+
+
+def get_current_record():
+    response = requests.get(
+        f"{ZENODO_API_BASE}/records{CONCEPT_REC_ID}",
+        params=AUTH_PARAM,
+        headers=HEADERS,
+    )
+    return response
+
+
+def discard_draft():
+    rec_id = get_current_record.json()["id"]
+    response = requests.get(
+        f"{ZENODO_API_BASE}/deposit/depositions/{rec_id}/actions/discard",
+        params=AUTH_PARAM,
+        headers=HEADERS,
+    )
+    return response
+
+
+def create_draft():
+    """Create a new draft deposit based on the current version."""
+    response = requests.get(
+        str(ZENODO_API_BASE),
+    )
+    return response
+
+
+def wipe_draft():
+    """Remove all files from the current draft."""
+
+
+@click.command()
+@click.option(
+    "--sandbox/--production",
+    "sandbox",
+    type=bool,
+    default=True,  # Change this to False once it's really working...
+    help="Use the Zenodo sandbox rather than the production server.",
+)
+def pudl_zenodo_data_release(sandbox: bool):
+    """Publish a new PUDL data release to Zenodo."""
+    pudl_out = Path(os.environ["PUDL_OUTPUT"])
+
+    pass
--- a/devtools/zenodo/zenodo_data_release.py
+++ b/devtools/zenodo/zenodo_data_release.py
@@ -194,7 +194,7 @@ class LocalSource(SourceData):
 
     def iter_files(self) -> Iterable[tuple[str, BinaryIO]]:
         """Loop through all files in the directory."""
-        for f in self.path.iterdir():
+        for f in sorted(self.path.iterdir()):
             if f.is_file():
                 yield f.name, f.open("rb")
             else:
@@ -323,13 +323,14 @@ class CompleteDraft(State):
     "--env",
     type=click.Choice([SANDBOX, PRODUCTION], case_sensitive=False),
     default=SANDBOX,
-    help="Use the Zenodo sandbox rather than the production server.",
+    help="Whether to use the Zenodo sandbox server (for testing) or the production server.",
+    show_default=True,
 )
 @click.option(
     "--rec-id",
     type=int,
     required=True,
-    help="The record ID we'd like to sync to.",
+    help="Record ID of the Zenodo archive to create a new version of.",
 )
 @click.option(
     "--source-dir",
@@ -337,14 +338,15 @@ class CompleteDraft(State):
         exists=True, file_okay=False, dir_okay=True, readable=True, path_type=Path
     ),
     required=True,
-    help="What directory to sync up to this concept DOI.",
+    help="Path to a directory whose contents will be uploaded to Zenodo. Subdirectories are ignored.",
 )
 @click.option(
     "--publish/--no-publish",
     default=False,
-    help="Whether to publish automatically after syncing, or to give you a chance to review.",
+    help="Whether to publish the new record without confirmation, or leave it as a draft to be reviewed.",
+    show_default=True,
 )
-def pudl_zenodo_data_release(env: bool, rec_id, source_dir, publish):
+def pudl_zenodo_data_release(env: str, rec_id: int, source_dir: Path, publish: bool):
     """Publish a new PUDL data release to Zenodo."""
     # TODO (daz): if the source-dir is actually an S3 or GCS uri, then, well. Do that instead.
     source_data = LocalSource(path=source_dir)
--- a/devtools/zenodo/zenodo_data_release.py
+++ b/devtools/zenodo/zenodo_data_release.py
@@ -92,7 +92,7 @@ class ZenodoClient:
         response = requests.get(
             f"{self.base_url}/deposit/depositions/{deposition_id}",
             headers=self.auth_headers,
-            timeout=10,
+            timeout=5,
         )
         logger.debug(
             f"License from JSON for {deposition_id} is "
@@ -108,7 +108,7 @@ class ZenodoClient:
         response = requests.get(
             f"{self.base_url}/records/{record_id}",
             headers=self.auth_headers,
-            timeout=10,
+            timeout=5,
         )
         return _NewRecord(**response.json())
 
@@ -121,7 +121,7 @@ class ZenodoClient:
         response = requests.post(
             f"{self.base_url}/records/{record_id}/versions",
             headers=self.auth_headers,
-            timeout=10,
+            timeout=5,
         )
         return _NewRecord(**response.json())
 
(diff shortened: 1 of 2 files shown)


jdangerx:
So you should be able to use this script to sync a local directory up to a Zenodo DOI - and it will leave a draft for you to manually approve, or just YOLO it and publish (with the `--publish` flag).

I think this functionality is useful enough to merge into `dev`, and the next PR will have to handle:

* [ ] getting Zenodo creds into nightly build runner - either through GH secrets or Google Secret Manager
* [ ] only publishing when `v20xx` tag is pushed

I wonder if that is enough funkiness to warrant moving `gcp_pudl_etl.sh` into a Python script...
I also wonder how much effort it would be to use Google Batch for this instead of a VM, since we already have this all Dockerized.

devtools/zenodo/zenodo_data_release.py
```diff
@@ -0,0 +1,295 @@
...
+)
+@click.option(
+    "--concept-rec-id",
+    type=int,
+    required=True,
+    help="The concept record ID we'd like to sync to.",
+)
+@click.option(
+    "--source-dir",
+    type=Path,
```
zaneselvans, line 267: Could be worth using the `click.Path` type here, which makes it easy to verify that this path exists, is a directory, and is readable.

zaneselvans: Ah, I guess if we want this to work with `gs://` or `s3://` paths then we'd need to use an `fsspec` path here. It looks like they're done some work to try and provide a `pathlib.Path` like interface for remote filesystems as well. But it's not super mature. https://github.com/fsspec/universal_pathlib

devtools/zenodo/zenodo_data_release.py
```diff
@@ -0,0 +1,295 @@
...
+        makes a new version unless one exists already.
+        """
+        return requests.post(
+            f"{self.base_url}/records/{record_id}/versions",
+            headers=self.auth_headers,
+            timeout=10,
+        )
+
+    def update_deposition_metadata(
+        self, deposition_id: int, metadata: dict
```
zaneselvans, line 87: It would be nice to have a Pydantic model or dataclass that defines what fields we want to require be populated in the Zenodo metadata. It would help make sure we're filling this out "completely" for our purposes.

The only element that I think we may need to change substantively between releases or generate somehow is the description, which I think is HTML, and would be nice to have a Jinja template for that we can fill in based on metadata like the start/end dates of the datasets that are included in the release.

devtools/zenodo/zenodo_data_release.py
```diff
@@ -0,0 +1,295 @@
...
+        self, deposition_id: int, file_name: str, file_content: BinaryIO
+    ) -> requests.Response:
+        """LEGACY API: Create file in deposition."""
+        url = f"{self.base_url}/deposit/depositions/{deposition_id}/files"
+        files = {"file": file_content, "name": file_name}
+        response = requests.post(
+            url,
+            headers=self.auth_headers,
+            files=files,
+            timeout=1_000,
```
zaneselvans, line 120: If it takes longer than 1000 seconds to upload one of the large files what happens?

If any of the files fail to upload on the first try, does the creation of the new archive necessarily fail? Or can individual files be retried if they don't work the first time?

zaneselvans (commented):
Mostly questions.

devtools/zenodo/zenodo_data_release.py
jdangerx: We can worry about `fsspec` when we get there... for now `click.Path` seems totally fine.

devtools/zenodo/zenodo_data_release.py
jdangerx: Yeah, I sort of wanted to avoid this because I didn't want to make any more promises about the Zenodo API & attendant data shapes than I had to. But I think the promises are being made regardless, and I can codify them with Pydantic models.

devtools/zenodo/zenodo_data_release.py
jdangerx: If it takes longer than 1,000 seconds to upload one of the large files, it will time out! We can set this to some ridiculously high number.

Currently there aren't retries.

We could add retry handling and also pre-publishing hash checks for sure, do you think that's blocking this merge or more of a follow-up sort of situation?

zaneselvans: I'm trying to think of the things that I would check as a human who had just run the script in reviewing the draft.

My expectation is that most of the time we will get some failures when uploading a new archive, usually on the larger files, and will have to retry.

We can hold off the retry logic and try running the script to see what happens, but I think we should clearly define what we want a failure to look like.

- Should it leave the files that did upload successfully in a draft? Or should it wipe everything and set it back to a no-draft state?
- Should it keep going and attempt to upload additional files after one of them has failed? Or should it bail as soon as there's a failure?
- If a failure is detected, it should never publish the archive, even if `--publish` has been set.

For now while we're testing and developing it, if one of the files fails to upload I kinda thing we still want it to iterate through the other files and try at least once to get them uploaded so we can see what happens with all of them.  And then leave the draft in whatever state it ended up in so we can inspect it.

zaneselvans:
## Notes on attempted use:

- We should display the default values for all the command line args.
- It's a little unusual to have required arguments also require a `--flag` to identify them.  The `--source-dir` feels like an argument not an option to me. Not sure about the `--record-id`. It kinda feels like the source directory is the "from" and the record ID is the "to", but "from" a directory "to" an integer is kinda weird.
- We should log each filename as it's being uploaded.
- I tried making an update to [this sandbox archive](https://sandbox.zenodo.org/records/5564) using the nightly build outputs from last night with the following command: 
```
./zenodo_data_release.py --env sandbox --concept-rec-id 5563 --no-publish --source-dir ../../../no-backups/nightly
```
and got the following error:

```
INFO:root:Getting new version for 5563
INFO:root:Draft 6640 has 22 existing files, deleting...
INFO:root:Syncing files from LocalSource(path=PosixPath('../../../no-backups/nightly')) to draft 6640...
  File "/Users/zane/code/catalyst/pudl/devtools/zenodo/./zenodo_data_release.py", line 343, in <module>
    pudl_zenodo_data_release()
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/click/core.py", line 783, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/devtools/zenodo/./zenodo_data_release.py", line 331, in pudl_zenodo_data_release
    .sync_directory(source_data)
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/devtools/zenodo/./zenodo_data_release.py", line 236, in sync_directory
    self.zenodo_client.create_deposition_file(self.record_id, name, blob)
  File "/Users/zane/code/catalyst/pudl/devtools/zenodo/./zenodo_data_release.py", line 150, in create_deposition_file
    response = requests.post(
               ^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/requests/api.py", line 115, in post
    return request("post", url, data=data, json=json, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/requests/sessions.py", line 575, in request
    prep = self.prepare_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/requests/sessions.py", line 486, in prepare_request
    p.prepare(
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/requests/models.py", line 371, in prepare
    self.prepare_body(data, files, json)
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/requests/models.py", line 556, in prepare_body
    (body, content_type) = self._encode_files(files, data)
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/requests/models.py", line 191, in _encode_files
    fdata = fp.read()
            ^^^^^^^^^
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa2 in position 82: invalid start byte
```
- Logging in, I looked at the drafts and found 6640 still there, unpublished. Two small JSON files had successfully uploaded. No metadata was filled in. So maybe Requests doesn't like getting binary data here?
- Attempting to run the update again without altering the draft at all, it successfully finds the draft with 2 files and removes them.
- Logging the filenames, I can see that the next file it's attempting to upload is the big EPA CEMS Parquet file, and it fails with the same error as above again.
- I tried a 3rd time, but now using the actual record ID, rather than the ID associated with the concept DOI, and it all worked the same, so it seems like we don't need to restrict the ID being used to be the concept record ID and could use `--record-id` instead, which feels more legible to me, and it's easy to pull the ID out of the URL by hand, whereas finding the concept record ID in the sidebar of the page can be a little obscure. For the automated uploads we'll want to actually use the concept record ID so we never have to update it.

jdangerx:
I do generally prefer "keyword" vs. "positional" arguments because they are much harder to misuse - so keeping that in for now.

The Whole Dang Nightly Builds Output also made it to [Sandbox](https://sandbox.zenodo.org/uploads/6640)! With the following command:

```
./devtools/zenodo/zenodo_data_release.py --rec-id 5563 --source-dir ~/scratch/pudl_2023.12.01/
```

And took about 2 hours, 15 minutes.

devtools/zenodo/zenodo_data_release.py
```diff
@@ -0,0 +1,367 @@
...
+
+@dataclass
+class LocalSource(SourceData):
+    """Get files from local directory."""
+
+    path: Path
+
+    def iter_files(self) -> Iterable[tuple[str, BinaryIO]]:
+        """Loop through all files in the directory."""
+        for f in self.path.iterdir():
```
zaneselvans, line 197: It's much easier to visually inspect a record if the files are sorted, so related FERC files are next to each other.

```suggestion
        for f in sorted(self.path.iterdir()):
```

zaneselvans (commented):
I suggested some clarifications on the help messages, and would like to know what you think about embedding the target record IDs in the script, at least as defaults.

devtools/zenodo/zenodo_data_release.py
```diff
@@ -0,0 +1,367 @@
...
+    type=click.Choice([SANDBOX, PRODUCTION], case_sensitive=False),
+    default=SANDBOX,
+    help="Use the Zenodo sandbox rather than the production server.",
+)
+@click.option(
+    "--rec-id",
+    type=int,
+    required=True,
+    help="The record ID we'd like to sync to.",
+)
```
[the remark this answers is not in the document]
jdangerx: I found it useful in testing to create new records in sandbox and then use this tool to sync files up to arbitrary records. Now that that phase is over, we could hard-code them based on environments.

zaneselvans:
✅

jdangerx (approved):
Weirdly, this became a PR where I wrote the code and Zane was reviewing. But he made a ✅ so I'm making the official GH PR-approval.

zaneselvans:
@jdangerx Will the auto-merge go through if the branch is out of date with the base branch? I feel like several PRs of mine have gotten stuck lately because of this.

jdangerx:
I think it will work **iff** the PR is still "auto-mergeable" / there are no conflicts. We'll see!

jdangerx:
Looks like the answer, from another PR, is "no." So I'm gonna click that 'update branch' button.

PR #3128, opened 2023-12-06
Table diff tools

jdangerx:
This adds a little helper function that makes a diff of two tables that have the same schema - plus some extra wrapper code in a notebook that handles ignoring some columns, caring extra about some others, etc.

```[tasklist]
## TODO
- [ ] Add more to the example to show a bit more of what you can do / update the helper functions to do what we need to do
- [x] Add test for empty case
- [ ] future? add support for dagster's `load_asset_value` - combined with Code Locations we might be able to compare across branches pretty easily...
```


jdangerx:
I do also want to make it so that this can diff different versions of *any* dagster asset, because then we can follow causality backwards. But I think that has to come after we get good at managing multiple DAGSTER_HOMEs.

PR #3129, opened 2023-12-07
Include sub-annual updates in annual_updates docs

### LINKED ISSUE #3130: Update documentation about how to integrate new data updates

aesharpe:
The current documentation page for adding new data is called "Annual Updates". This is misleading because the page caters to all sorts of annual and sub-annual updates to existing data. 

I changed the name to "Existing Data Updates" and updated some of the language in the documents to be more inclusive of all types of data updates.

Closes #3130

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
   # Formatters: hooks that re-write Python & documentation files
   ####################################################################################
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.6
+    rev: v0.1.7
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
--- a/devtools/sqlite-table-diff.ipynb
+++ b/devtools/sqlite-table-diff.ipynb
@@ -1,0 +1,102 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "\"\"\"Example of diffing tables across multiple different SQLite DBs.\n",
+    "\n",
+    "The tables must have the same name/schema. This is intended for use in\n",
+    "investigating validation test errors.\n",
+    "\"\"\"\n",
+    "import sqlite3\n",
+    "from pathlib import Path\n",
+    "from typing import Iterable\n",
+    "\n",
+    "import pandas as pd\n",
+    "\n",
+    "from pudl.helpers import diff_wide_tables, TableDiff\n",
+    "from pudl.metadata.classes import Resource\n",
+    "from pudl.metadata.fields import apply_pudl_dtypes\n",
+    "\n",
+    "\n",
+    "def table_diff(\n",
+    "        table_name: str,\n",
+    "        old_conn,\n",
+    "        new_conn,\n",
+    "        ignore_cols: Iterable[str] = (\"plant_id_ferc1\",),\n",
+    "        addl_key_cols: Iterable[str] = (),\n",
+    "    ) -> TableDiff:\n",
+    "    \"\"\"Diff two versions of the same table that live in SQL databases.\n",
+    "\n",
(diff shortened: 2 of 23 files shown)


aesharpe:
Currently `dev` is failing the ci so I pushed all these commits with a `no-verify` flag. Once `dev` is back up and running we'll see if the tests pass here too.

.github/ISSUE_TEMPLATE/existing_data_updates.md
```diff
@@ -0,0 +1,24 @@
...
+about: Check-list for integrating a new year of data
+title: ''
+labels: new-data
+assignees: ''
+
+---
+
+### New year of data integration check-list:
+
+Based on the [Existing Data Updates Docs](https://catalystcoop-pudl.readthedocs.io/en/dev/dev/existing_data_updates.html)
```
aesharpe, line 12: This whole template is showing up as deleted and then re-added I think because the percentage content change is large enough that git doesn't recognize it as the same file? I tried to prevent this by running seperate commits - one for the file changes and one for the file rename - but that didn't seem to help.

https://stackoverflow.com/questions/7938582/how-does-git-detect-similar-files-for-its-rename-detection

cmgosnell: that's annoying but of well :-/

docs/dev/existing_data_updates.rst
```diff
@@ -1,14 +1,25 @@
...
-As of spring 2023 the annual updates include:
+Many of the raw data inputs for PUDL are published on a annual or monthly basis. These
+instructions explain the process for integrating new versions of existing data into
+PUDL.
+
+We update EIA monthly data and EPA CEMS hourly data on a quarterly basis.
+
+EIA typically publishes an "early release" version of their annual data in the summer
+followed by a final release in the fall. Our ``data_maturity`` column indicates
+which version has been integrated into PUDL.
```
cmgosnell, line 13: the `data_maturity` column also indicates whether the data is coming from the monthly or otherwise YTD data. it might be good to add that here in well bc it sounds like this is just an indicator or ER or final

docs/dev/existing_data_updates.rst
```diff
@@ -31,19 +39,20 @@ archivers themselves.
 **1.2)** Update the dictionary of production DOIs in :mod:`pudl.workspace.datastore` to
 refer to the new raw input archives.
 
-**1.3)** In :py:const:`pudl.metadata.sources.SOURCES`, update the ``working_partitions``
-to reflect the years of data that are available within each dataset and the
-``records_liberated`` to show how many records are available. Check to make sure other
-fields such as ``source_format`` or ``path`` are still accurate.
+**1.3)** If you're adding a new year, in :py:const:`pudl.metadata.sources.SOURCES`,
```
cmgosnell, line 42: "adding a new year" could be "increasing coverage" or something that is more inclusive of non-year partitions

docs/dev/existing_data_updates.rst
```diff
@@ -31,19 +39,20 @@ archivers themselves.
...
 
   If you're updating EIA861, you can skip the rest of the steps in this section and
   all steps after step two because 861 is not yet included in the ETL.
 
-**1.4)** Update the years of data to be processed in the ``etl_full.yml`` and
-``etl_fast.yml`` settings files stored under ``src/pudl/package_data/settings`` in the
-PUDL repo.
+**1.4)** If you're adding a new year, update the years of data to be processed in
+the ``etl_full.yml`` and ``etl_fast.yml`` settings files stored under
+``src/pudl/package_data/settings`` in the PUDL repo.
```
cmgosnell, line 55: same comment from above about increasing coverage

docs/dev/existing_data_updates.rst
aesharpe: That's true...the reason I used "year" here specifically is because sometimes (as in the case for 923 quarterly updates) when we add the monthly data we don't update the year partitions because the new data is within the same year. So this step is unnecessary.

PR #3132, opened 2023-12-07
Skip slow tests in pre-commit hook

jdangerx:
The property-based test that makes sure our XBRL deduplication logic is working is [quite slow](https://github.com/unionai-oss/pandera/issues/1116). This test alone adds ~15-20s onto our full unit test suite - which is quite a lot! Additionally, it sometimes errors because test case generation time is erratic.

So:

* I increased the deadline on the test case generation
* I tagged all 3 tests that took >1s to execute with the `pytest.mark.slow` mark. This allows us to skip them in the `pre-commit` hook, but we will still run them in CI as well as when you just run `pytest test/unit` or `pytest test/unit/the-file.py`. This cuts the unit test timing on my machine from ~60s to ~10s.

I think it's reasonable to say that we should remove the property-based test and revisit when the performance isn't such an issue, or write a data generation strategy that makes a bunch of records and turns them into DataFrames.


zaneselvans:
Side-note, based on #3133 investigations, it looks like the PostgreSQL weirdness is somehow related to OpenSSL 3.2.0 weirdness, which is causing trouble for other projects as well, and that's why we've got all the downgrades here.

bendnorman (approved):
This is great, thank you!

Side note: I kicked off a full build from this branch to make sure the version downgrades solve #3133 which they do! So, hopefully, dev passes tonight 🤞

PR #3154, opened 2023-12-13
TESTONLY - sharded ferc_to_sqlite in ci-integration

rousik:
Splits off `ferc_to_sqlite` step into sharded step that generates each dataset independently on a smaller runner. This way we should be able to speed up the process significantly.


rousik:
This is combination of pre-existing PR that reworks pytest and refactoring of `ferc_to_sqlite` into separate matrix-based job. I'm aiming to turn this into separate PR when `parallel-pytest` is merged, but this gives a preview into this proof of concept. I think that it would make sense to move the setup steps (micromamba + other steps) into reusable workflow that can be used here repeatedly now that we have more than just one instance that needs this.

PR #3155, opened 2023-12-13
Update lockfiles on pre-commit

### LINKED ISSUE #3133: Double Free error
We're experiencing a frustrating `double free or corruption (out)` error that initially did not provide any helpful stack traces. I enabled the python [`faulthandler`](https://docs.python.org/3/library/faulthandler.html) and got a more useful stacktrace: 
[Full double free stack trace.log](https://github.com/catalyst-cooperative/pudl/files/13603321/Full.double.free.stack.trace.log)

The last successful build was from commit `7cc80dab59437b0dda3dbac3bed28be5f4e33858` and this error popped up from commit `58bb0880b1ef6de8d3e1341b8c3f481e5a44d9f9`.

Here is what I know so far:
- The stack trace from `faulthandler` suggests the the double free is occurring while trying to connect to the postgres database.
- The ETL does not fail in CI or locally. CI and local deployment do not run in a docker container and use sqlite for dagster storage.
- The ETL does not fail when I run the ETL locally while using the Cloud SQL postgres as the dagster backend.
- This error does not occur when using sqlite as the dagster storage backend in the docker container and VM.
- This error does not occur while running the code from `58bb088` with the conda lock files from `7cc80da`. Here is the diff between the two [diff_env.txt](https://github.com/catalyst-cooperative/pudl/files/13603376/diff_env.txt). A few packages `psycopg2` depends on changed between the two commits (I think the versions from this command are out of date because I've been mucking with the version numbers):
```
✦ ❯ mamba repoquery depends psycopg2
Using local repodata...

Loaded current active prefix: "/Users/bendnorman/mambaforge/envs/pudl-dev"

 Name       Version Build                Channel     Subdir
────────────────────────────────────────────────────────────
 libpq      16.1    h6dd4ff7_0           conda-forge osx-64
 openssl    3.1.4   hd75f5a5_0           conda-forge osx-64
 psycopg2   2.9.7   py311h187f0af_1      conda-forge osx-64
 python     3.11.6  h30d4d87_0_cpython   conda-forge osx-64
 python_abi 3.11    4_cp311              conda-forge osx-64
```

Here are the version changes for the dependencies that changed:
```
-  - libpq=16.1=hfc447b1_0
+  - libpq=16.1=hfc447b1_2
-  - openssl=3.1.4=hd590300_0
+  - openssl=3.2.0=hd590300_1
```

Seems one of these indirect `psycopg2` dependencies is the culprit though I'm not sure how to proceed:
- Revert to `openssl=3.1.4`
- I'm not sure how to revert libpq
- Update `psycopg2=v2.9.9`
- We got a double free error a couple of weeks ago because of some conflict in `curl` installations. We resolved this by installing curl in the mamba env instead of using `apt-get` #3090. I could try to move our other packages installed using `apt-get` into the mamba env. I don't think every `apt-get` package is available via mamba though.

jdangerx:
# Overview

Closes #3133.

We introduced `conda-lock` and lockfiles in an attempt to avoid instability caused by frequent dependency version changes. 

However, we have been re-locking very eagerly, which means that we are actually still very susceptible to dependency version changes that break our ETL. It also chews through our CI budget.

At least it's very easy to see what changed in the environment between runs!

So we decided to only re-lock:
* before committing a change to `pyproject.toml`
* once a week

This adds a simple hook to `.pre-commit-config.yaml` which runs our re-locking `make` command when `pyproject.toml` changes.

If `pyproject.toml` doesn't change, the hook is skipped; if `pyproject.toml` changes, it runs `make conda-clean conda-lock.yml`.

If `make` doesn't actually change any of the `environment/*` files, pre-commit will pass; otherwise the check will fail and you'll have to re-add & re-commit, much like any other pre-commit check we run.

Sometimes the lock will have updated between the first time you ran the hook and the second time, because some package update happened - in that case you'll have to re-add & re-commit a third time.

# Testing

To make sure this worked, I added some comments to `pyproject.toml` and attempted to commit - that triggered a re-lock. Then I added the lockfile changes and re-committed - this time it still re-locked (because `pyproject.toml` diffs were still in the commit) but because the lockfiles didn't change, the pre-commit check passed. 

```[tasklist]
# To-do list
```

### DIFF
--- a/.github/workflows/update-conda-lockfile.yml
+++ b/.github/workflows/update-conda-lockfile.yml
@@ -5,12 +5,6 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "0 9 * * 1" # Mondays at 9AM UTC
-  push:
-    paths:
-      - "Makefile"
-      - "pyproject.toml"
-      - "environments/*"
-      - ".github/workflows/update-conda-lockfile.yml"
 
 # What branch does this action run on?
 # - workflow_dispatch: Whatever branch it was run against.
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -101,5 +101,5 @@ ci:
   autoupdate_branch: dev
   autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
   autoupdate_schedule: weekly
-  skip: [unit-tests, nb-output-clear]
+  skip: [unit-tests, nb-output-clear, conda-lock]
   submodules: false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,6 +82,15 @@ repos:
         always_run: true
         entry: pytest --doctest-modules src/pudl test/unit -m "not slow"
 
+      - id: conda-lock
+        name: conda-lock
+        stages: [commit]
+        language: system
+        verbose: false
+        files: "pyproject.toml"
+        always_run: false
+        entry: make conda-clean conda-lock.yml
+
 # Configuration for pre-commit.ci
 ci:
   autofix_commit_msg: |


.github/workflows/update-conda-lockfile.yml
```diff
@@ -5,12 +5,6 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "0 9 * * 1" # Mondays at 9AM UTC
```
jdangerx, line 7: Should we make this run less frequently than weekly? I guess we should see how this cadence goes. We could do [1st/15th of each month](https://crontab.guru/#0_9_1,15_*_*) if weekly feels too frequent still.

.pre-commit-config.yaml
```diff
@@ -82,6 +82,15 @@ repos:
...
         entry: pytest --doctest-modules src/pudl test/unit -m "not slow"
 
+      - id: conda-lock
+        name: conda-lock
+        stages: [commit]
+        language: system
+        verbose: false
+        files: "pyproject.toml"
+        always_run: false
+        entry: make conda-clean conda-lock.yml
```
jdangerx, line 92: Surprisingly straightforward to get the behavior I wanted... a little suspicious but I'll take it.

jdangerx:
Aiyah, the pre-commit.ci checker doesn't have mamba.

.pre-commit-config.yaml
zaneselvans: I made a trivial change to my local `pyproject.toml` and tried to commit and it regenerated the lockfiles, refused to commit, and then I added them to the staging area, and tried to commit again, and it regenerated them again with no changes and the commit went through so, it seems like it works.

I guess if someone makes a change to `pyproject.toml` that is NOT related to dependencies, after the first run through the pre-commit hooks if everything else passes they can do a `--no-verify` to avoid unnecessarily relocking all the dependencies.

zaneselvans (approved):

zaneselvans:
@jdangerx Looking at RTD it seems to think that the docs build passed. Not sure why it's showing up here as having failed.

zaneselvans:
Re-ran RTD and it passed, but auto-merge didn't run with it. 🤷🏼

PR #3157, opened 2023-12-14
Alter output path for the nightly builds

rousik:
Try to fix `COMMIT_BRANCH` calculation (doesn't work) and use runtime timestamp instead of commit timestamp when constructing the output name.


zaneselvans:
Blurgh it looks like maybe now we're running out of disk space when downloading the newly chonky CEMS data? But the big runners have more disk too.

zaneselvans (approved):

PR #3158, opened 2023-12-14
Attempt to create sandbox data release in nightly builds.

zaneselvans:
# Overview

Attempt to create a sandbox data release after the nightly builds succeed.  This is just meant to test the data release publication process until we switch over to creating real releases from tagged builds.

I think the workflow needs to be added to `main` before it'll work in the scheduled builds though.

I added the `ZENODO_SANDBOX_TOKEN_PUBLISH` to the organization secrets.

### DIFF
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -78,6 +78,11 @@ function copy_outputs_to_distribution_bucket() {
     aws s3 cp "$PUDL_OUTPUT/" "s3://intake.catalyst.coop/$GITHUB_REF" --recursive
 }
 
+function zenodo_data_release() {
+    echo "Creating a new PUDL data release on Zenodo."
+    ~/devtools/zenodo/zenodo_data_release.py --publish --env sandbox --source-dir $PUDL_OUTPUT
+}
+
 
 function notify_slack() {
     # Notify pudl-builds slack channel of deployment status
@@ -125,8 +130,13 @@ if [[ $ETL_SUCCESS == 0 ]]; then
     if [ $GITHUB_ACTION_TRIGGER = "push" ] || [ $GITHUB_REF = "dev" ]; then
         copy_outputs_to_distribution_bucket
         ETL_SUCCESS=${PIPESTATUS[0]}
+        zenodo_data_release 2>&1 | tee -a $LOGFILE
+        ETL_SUCCESS=${PIPESTATUS[0]}
     fi
 fi
+
+# This way we also save the logs from latter steps in the script
+gsutil cp $LOGFILE ${PUDL_GCS_OUTPUT}
 
 # Notify slack about entire pipeline's success or failure;
 # PIPESTATUS[0] either refers to the failed ETL run or the last distribution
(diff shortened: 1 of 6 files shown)


docker/gcp_pudl_etl.sh
```diff
@@ -122,6 +127,8 @@ if [[ $ETL_SUCCESS == 0 ]]; then
     if [ $GITHUB_ACTION_TRIGGER = "push" ] || [ $GITHUB_REF = "dev" ]; then
         copy_outputs_to_distribution_bucket
         ETL_SUCCESS=${PIPESTATUS[0]}
+        zenodo_data_release 2>&1 | tee -a $LOGFILE
```
bendnorman, line 130: Given our current logic, the logs from the release script will show up in log file that is sent to slack but they won't show up in the log file copied to the GCS bucket. I think `copy_outputs_to_gcs` should be called towards the end of the script.

bendnorman (changes requested):
Mostly looks good to me. Left a comment about log handling.

docker/gcp_pudl_etl.sh
zaneselvans: There's a problem here now that we're doing some post-processing for distribution -- removing files we don't want to distribute, gzipping the SQLite DBs.  What we copy to GCS is for forensic purposes, which is different from what we're currently shipping to AWS, Kaggle, Zenodo, etc.

bendnorman: Ah that's right. I guess it's not a huge deal if we don't include the zenodo release logs in the GCS bucket. We could also change the script to dump the outputs to GCS after the ETL runs then dump the logs after most of the post-processing/distribution logic has happened.

p.s. would love to rework this system so all of the logs are collected for us.

zaneselvans: Do you think it'd be better to capture the additional steps in the logs, or only save the outputs we distribute to the nightly build buckets?

docker/gcp_pudl_etl.sh
```diff
@@ -122,9 +127,14 @@ if [[ $ETL_SUCCESS == 0 ]]; then
...
+        ETL_SUCCESS=${PIPESTATUS[0]}
     fi
 fi
 
+# This way we also save the logs from latter steps in the script
+gsutil cp $LOGFILE ${PUDL_GCS_OUTPUT}
```
zaneselvans, line 136: Okay I added a re-copy of the logfile after everything.

docker/gcp_pudl_etl.sh
zaneselvans: I think an overhaul of the nightly build system is definitely on the docket for next year.

bendnorman: I don't think I understand your question:
> Do you think it'd be better to capture the additional steps in the logs, or only save the outputs we distribute to the nightly build buckets?

Are you asking which logs should be written to the log file?

bendnorman (approved):
Looks good! This should be merged into `main` because we are updating a workflow file.

docker/gcp_pudl_etl.sh
zaneselvans: Nevermind I was seeing it as either or based on the ordering, but re-saving the logfile fixes everything.

PR #3160, opened 2023-12-14
Add EPA CEMS concurrency limit to `pudl_etl`

bendnorman:
# Overview

#3096 limited concurrency for the `hourly_emissions_epacems` asset for the jobs being loaded into the `dagster-webserver`. This PR applies the limit to the `pudl_etl` command. 

I ran into an OOM issue when running a full build on the `rename-core-asset` branch. We haven't run into this issue yet on a nightly `dev` build which is a little suspicious. Maybe we've just been getting lucky on `pudl-deployment-dev`? I diffed the tag and dev machines specs and they are working with the same amount of CPUs and memory. Either way, I think we should be able to limit CEMS concurrency for the `pudl_etl` command.
# Testing

I ran `pudl_etl` locally and made sure only two partitions were executed at once.

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/etl/cli.py
+++ b/src/pudl/etl/cli.py
@@ -145,7 +145,13 @@ def pudl_etl(
             },
         },
     }
-    run_config.update(get_dagster_execution_config(dagster_workers))
+
+    run_config.update(
+        get_dagster_execution_config(
+            num_workers=dagster_workers,
+            tag_concurrency_limits=pudl.etl.default_tag_concurrency_limits,
+        )
+    )
 
     result = execute_job(
         pudl_etl_reconstructable_job,
--- a/src/pudl/etl/cli.py
+++ b/src/pudl/etl/cli.py
@@ -77,6 +77,15 @@ def pudl_etl_job_factory(
     help="Max number of processes Dagster can launch. Defaults to the number of CPUs.",
 )
 @click.option(
+    "--epacems-workers",
+    default=2,
+    type=int,
+    help=(
+        "Max number of processes Dagster can launch for EPA CEMS assets. Defaults "
+        "to max number of processes our typical local machines can handle."
+    ),
+)
+@click.option(
     "--gcs-cache-path",
     type=str,
     help=(
@@ -107,6 +116,7 @@ def pudl_etl_job_factory(
 def pudl_etl(
     etl_settings_yml: pathlib.Path,
     dagster_workers: int,
+    epacems_workers: int,
     gcs_cache_path: str,
     logfile: pathlib.Path,
     loglevel: str,
@@ -146,10 +156,18 @@ def pudl_etl(
         },
     }
 
+    tag_concurrency_limits = [
+        {
+            "key": "datasource",
+            "value": "epacems",
+            "limit": epacems_workers,
+        }
+    ]
+
     run_config.update(
         get_dagster_execution_config(
             num_workers=dagster_workers,
-            tag_concurrency_limits=pudl.etl.default_tag_concurrency_limits,
+            tag_concurrency_limits=tag_concurrency_limits,
         )
     )
 
--- a/src/pudl/helpers.py
+++ b/src/pudl/helpers.py
@@ -1796,7 +1796,9 @@ def scale_by_ownership(
     return gens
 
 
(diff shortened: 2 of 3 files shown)


src/pudl/etl/cli.py
```diff
@@ -145,7 +145,13 @@ def pudl_etl(
             },
         },
     }
-    run_config.update(get_dagster_execution_config(dagster_workers))
+
+    run_config.update(
+        get_dagster_execution_config(
+            num_workers=dagster_workers,
+            tag_concurrency_limits=pudl.etl.default_tag_concurrency_limits,
```
bendnorman, line 152: Should I add EPA CEMS concurrency value as a command line arg with a default value of 2 for `pudl_etl`?

src/pudl/helpers.py
```diff
@@ -1808,12 +1810,19 @@ def get_dagster_execution_config(num_workers: int = 0):
...
+            that key-value pair. If no value is set, the limit is applied across
+            all values of that key. If the value is set to a dict with
+            `applyLimitPerUniqueValue: true`, the limit will apply to the number
+            of unique values for that key. Note that these limits are per run, not global.
 
     Returns:
         A dagster execution config.
     """
-    if not num_workers:
-        return {}
```
bendnorman, line 1816: Just using the default values of the function is simpler now that we have multiple execution config arguments.

src/pudl/etl/cli.py
zaneselvans: Would we want to run it with higher concurrency in the nightly builds because the machine has so much memory? I have no trouble on my laptop with 32GB and 2 concurrent processes.

Although, having several ways to configure the same parameters can get confusing and make it harder to debug when something goes wrong.  As it is, it's hard-coded to 2, unless someone edits it in the UI right? How does Dagster typically expect people to configure this kind of thing?

src/pudl/helpers.py
zaneselvans: I think maybe I would benefit from a walkthrough on the config system. I don't really understand how these settings are flowing into the run, when they get used vs. not.

zaneselvans: Does this change mean that the asset groups would have a concurrency of 2 for cems now too?

src/pudl/etl/cli.py
bendnorman: I agree tt would be nice to be able to change concurrency for CEMS from the command line, given the GCP machines have a bunch of memory. I'll add it as an arg.

src/pudl/helpers.py
bendnorman: No :/ I'm not sure of a way to configure execution config for asset groups. Basically, you need to construct a job from a collection of assets then specify the execution configuration for that job.

src/pudl/etl/cli.py
bendnorman: We could just move towards using the dagster CLI commands to execute jobs specified in `pudl.etl` instead of using `pudl_etl` and `ferc_to_sqlite` and reconstructing the run configs.

```
dagster job execute -m pudl.etl -j [etl_fast|etl_fast_no_cems|etl_full|etl_full_no_cems]

dagster job execute -m pudl.etl -j [ferc_to_sqlite_fast|ferc_to_sqlite_full]
```

The PUDL commands currently use `dagster.build_reconstructable_job()` which is required to execute multiproccess jobs from python. The function is still experimental and not super straightforward. We kept `pudl_etl` and `ferc_to_sqlite` around as people migrated to executing from the UI. Seems like we're at the point where we are mostly executing jobs from the UI and just using the CLI commands for CI.

bendnorman: Also, if we use the dagster CLI we won't have to translate click arguments into dagster config.

zaneselvans: If there's a clean and well supported way to run the things that we run using dagster directly it seems like that's probably the direction to move in.

bendnorman: Sounds good 👍 I'll create an issue to replace the commands

zaneselvans (approved):

Issue #3140, opened 2023-12-08
Move to trunk-based development


jdangerx:
Earlier context in [this discussion](https://github.com/orgs/catalyst-cooperative/discussions/2506)

We spend a bunch of time juggling the `dev` and `main` branches which complicates some of our automation and development flows. It also means that `main` and `dev` are divergent. Which one is the authoritative history of the project? Who really knows?

The desired end state is to get rid of `dev` and branch everything off of `main`.

We want to keep building nightly builds, and have tagged release versions.

The RTD updates appear to be mostly about configuring the readthedocs project.

```[tasklist]
### Success criteria
- [x] only archive vXXX tags on Zenodo
- [x] `nightly-build-XXXX-XX-XX` tags created every night - see #3164
- [x] `nightly` branch always points at latest `nightly-build-XXXX-XX` tag - see #3164
- [x] `stable` branch always points at latest `v20XX.XX.XX` tag - see #3164
- [x] Datasette gets redeployed on nightly build success - see #3164
- [ ] #3193
- [ ] https://github.com/catalyst-cooperative/pudl/pull/3195
- [x] RTD versions: `nightly` -> `nightly`, `stable` -> `stable`, `latest` -> `main`
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3179
```

zaneselvans:
On the archive to Zenodo on tag, I think you're talking about the data releases, but there will be 2 kinds of archives that get sent to Zenodo: software and data.  The software archive-on-tag is working now, and is only triggered on `v20*` tags (because GitHub releases are only triggered on `v20*` tags, and that's what the GitHub to Zenodo integration watches for).

By unifying the scheduled and tagged builds, do you mean that we only ever build on a schedule, and every tagged release would just be adding an additional tag to the successful nightly build that we've decided to use as a new versioned release?  So we could just package up that commit and its build outputs without needing to re-run anything?  That sounds nice and clean to me!  One issue that might come up there is that typically there's a post-build edit to the release notes that happens to close off the old section (which is changes that go into the release) and open up a new section.  But I imagine lots of other software also has this problem so there must be some way around it.

jdangerx:
@zaneselvans with RTD, we're currently set up to include these versions on our website:

<img width="337" alt="image" src="https://github.com/catalyst-cooperative/pudl/assets/2495794/2f33f091-bc5d-4830-beda-c46ab2f3baf4">

- `latest` corresponds to `main`;
[...]

PR #3164, opened 2023-12-15
Improve nightly build and deployment logic

zaneselvans:
# Overview

- Remove minor versions from the GHA we depend on to minimize dependabot churn.
- Remove the logic that was updating conda lockfiles on push.
- Purge old files from distribution paths before uploading new outputs. This prevents stale files from hanging around after we change the name of a file.
- Tag commits that are used for nightly builds.
- Update `stable` branch to point at newly tagged release
- Update `nightly` branch to point at newly successful nightly build tag.
- Simplify Dockerfile / make it more internally consistent

## Issues / Questions:
- The nightly build jank continues. We need to dedicate some resources to consolidating it!

Part of #3140

### DIFF
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -39,7 +39,7 @@ jobs:
         run: |
           git config user.email "pudl@catalyst.coop"
           git config user.name "PudlBot"
-          git tag -a -m "$NIGHTLY_TAG" $NIGHTLY_TAG
+          git tag -a -m "$NIGHTLY_TAG" $NIGHTLY_TAG $GITHUB_REF
           git push origin $NIGHTLY_TAG
 
       - name: Get HEAD of the branch (main or dev)
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -137,6 +137,7 @@ jobs:
             --container-env DAGSTER_PG_HOST="104.154.182.24" \
             --container-env DAGSTER_PG_DB="dagster-storage" \
             --container-env FLY_ACCESS_TOKEN=${{ secrets.FLY_ACCESS_TOKEN }} \
+            --container-env PUDL_BOT_PAT=${{ secrets.PUDL_BOT_PAT }} \
             --container-env ZENODO_SANDBOX_TOKEN_PUBLISH=${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }} \
             --container-env PUDL_SETTINGS_YML="/home/mambauser/src/pudl/package_data/settings/etl_full.yml" \
             --container-env PUDL_GCS_OUTPUT=${{ env.PUDL_OUTPUT_PATH }}
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -140,7 +140,7 @@ jobs:
             --container-env FLY_ACCESS_TOKEN=${{ secrets.FLY_ACCESS_TOKEN }} \
             --container-env PUDL_BOT_PAT=${{ secrets.PUDL_BOT_PAT }} \
             --container-env ZENODO_SANDBOX_TOKEN_PUBLISH=${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }} \
-            --container-env PUDL_SETTINGS_YML="/home/mambauser/src/pudl/package_data/settings/etl_full.yml" \
+            --container-env PUDL_SETTINGS_YML="/home/mambauser/pudl/src/pudl/package_data/settings/etl_full.yml" \
             --container-env PUDL_GCS_OUTPUT=${{ env.PUDL_OUTPUT_PATH }}
 
       # Start the VM
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -26,6 +26,14 @@ jobs:
         if: ${{ (github.event_name == 'schedule') }}
         run: |
           echo "This action was triggered by a schedule." && echo "GCE_INSTANCE=pudl-deployment-dev" >> $GITHUB_ENV && echo "GITHUB_REF=dev" >> $GITHUB_ENV
+          echo "RUN_DATE=$(date +%Y-%m-%d)" >> $GITHUB_ENV
+
+      - name: Tag nightly build
+        if: ${{ (github.event_name == 'schedule') }}
+        uses: EndBug/latest-tag@1
+        with:
+          ref: nightly-${{ env.RUN_DATE }}
+          description: "Nightly build for ${{ env.RUN_DATE }}"
 
       - name: Checkout Repository
         uses: actions/checkout@v4
@@ -45,7 +53,7 @@ jobs:
 
       - name: Docker Metadata
         id: docker_metadata
-        uses: docker/metadata-action@v5.3.0
+        uses: docker/metadata-action@v5
         with:
           images: catalystcoop/pudl-etl
           flavor: |
@@ -55,17 +63,17 @@ jobs:
             type=ref,event=tag
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3.0.0
+        uses: docker/setup-buildx-action@v3
 
       - name: Login to DockerHub
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@v3.0.0
+        uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       - name: Build image and push to Docker Hub
-        uses: docker/build-push-action@v5.1.0
+        uses: docker/build-push-action@v5
         with:
           context: .
           file: docker/Dockerfile
@@ -137,7 +145,7 @@ jobs:
 
       - name: Post to a pudl-deployments channel
         id: slack
-        uses: slackapi/slack-github-action@v1.24.0
+        uses: slackapi/slack-github-action@v1
         with:
           channel-id: "C03FHB9N0PQ"
           slack-message: "build-deploy-pudl status: ${{ job.status }}\n${{ env.GCS_OUTPUT_BUCKET }}/${{ env.RUN_TIMESTAMP}}-${{ env.SHORT_SHA }}-${{ env.COMMIT_BRANCH }}"
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -124,6 +124,7 @@ jobs:
             --container-env-file="./docker/.env" \
             --container-env ACTION_SHA=$ACTION_SHA \
             --container-env GITHUB_REF=${{ env.GITHUB_REF }} \
+            --container-env NIGHTLY_TAG=${{ env.NIGHTLY_TAG }} \
             --container-env GITHUB_ACTION_TRIGGER=${{ github.event_name }} \
             --container-env SLACK_TOKEN=${{ secrets.PUDL_DEPLOY_SLACK_TOKEN }} \
             --container-env GCE_INSTANCE=${{ env.GCE_INSTANCE }} \
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -26,19 +26,21 @@ jobs:
         if: ${{ (github.event_name == 'schedule') }}
         run: |
           echo "This action was triggered by a schedule." && echo "GCE_INSTANCE=pudl-deployment-dev" >> $GITHUB_ENV && echo "GITHUB_REF=dev" >> $GITHUB_ENV
-          echo "RUN_DATE=$(date +%Y-%m-%d)" >> $GITHUB_ENV
-
-      - name: Tag nightly build
-        if: ${{ (github.event_name == 'schedule') }}
-        uses: EndBug/latest-tag@1
-        with:
-          ref: nightly-${{ env.RUN_DATE }}
-          description: "Nightly build for ${{ env.RUN_DATE }}"
+          echo "NIGHTLY_TAG=nightly-$(date +%Y-%m-%d)" >> $GITHUB_ENV
+          echo "NIGHTLY_TAG: $NIGHTLY_TAG"
 
       - name: Checkout Repository
         uses: actions/checkout@v4
         with:
           ref: ${{ env.GITHUB_REF }}
+
+      - name: Tag nightly build
+        if: ${{ (github.event_name == 'schedule') }}
+        run: |
+          git config user.email "pudl@catalyst.coop"
+          git config user.name "PudlBot"
+          git tag -a -m "$NIGHTLY_TAG" $NIGHTLY_TAG
+          git push origin $NIGHTLY_TAG
 
       - name: Get HEAD of the branch (main or dev)
         run: |
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -25,33 +25,35 @@ jobs:
       - name: Use pudl-deployment-dev vm and dev branch if running on a schedule
         if: ${{ (github.event_name == 'schedule') }}
         run: |
-          echo "This action was triggered by a schedule." && echo "GCE_INSTANCE=pudl-deployment-dev" >> $GITHUB_ENV && echo "GITHUB_REF=dev" >> $GITHUB_ENV
-          echo "NIGHTLY_TAG=nightly-$(date +%Y-%m-%d)" >> $GITHUB_ENV
-          echo "NIGHTLY_TAG: $NIGHTLY_TAG"
+          echo "This action was triggered by a schedule."
+          echo "GCE_INSTANCE=pudl-deployment-dev" >> $GITHUB_ENV
+          echo "GCE_INSTANCE: $GCE_INSTANCE"
+          echo "GITHUB_REF=dev" >> $GITHUB_ENV
+          echo "GITHUB_REF: $GITHUB_REF"
 
       - name: Checkout Repository
         uses: actions/checkout@v4
         with:
           ref: ${{ env.GITHUB_REF }}
 
+      - name: Set action environment variables
+        run: |
+          echo "NIGHTLY_TAG=nightly-$(date +%Y-%m-%d)" >> $GITHUB_ENV
+          echo "NIGHTLY_TAG: $NIGHTLY_TAG"
+          echo "SHORT_SHA=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+          echo "SHORT_SHA: $SHORT_SHA"
+          echo "BUILD_TIMESTAMP=$(date +%Y-%m-%d-%H%M)" >> $GITHUB_ENV
+          echo "BUILD_TIMESTAMP: $BUILD_TIMESTAMP"
+          echo "BUILD_ID=${BUILD_TIMESTAMP}-${SHORT_SHA}-${GITHUB_REF}
+          echo "BUILD_ID: $BUILD_ID"
+
       - name: Tag nightly build
         if: ${{ (github.event_name == 'schedule') }}
         run: |
           git config user.email "pudl@catalyst.coop"
-          git config user.name "PudlBot"
+          git config user.name "pudlbot"
           git tag -a -m "$NIGHTLY_TAG" $NIGHTLY_TAG $GITHUB_REF
           git push origin $NIGHTLY_TAG
-
-      - name: Get HEAD of the branch (main or dev)
-        run: |
-          echo "ACTION_SHA=$(git rev-parse HEAD)" >> $GITHUB_ENV
-          echo "SHORT_SHA=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
-
-      - name: Print action vars
-        run: |
-          echo "ACTION_SHA: $ACTION_SHA" && \
-          echo "GITHUB_REF: $GITHUB_REF" && \
-          echo "GCE_INSTANCE: $GCE_INSTANCE"
 
       - name: Docker Metadata
         id: docker_metadata
@@ -95,17 +97,11 @@ jobs:
       - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@v2
 
-      - name: Determine commit information
-        run: |-
-          echo "COMMIT_BRANCH=$(echo ${GITHUB_REF#refs/heads/} | tr / -)" >> $GITHUB_ENV
-          echo "COMMIT_TIME=$(git log -1 --format=%cd --date=format:%Y-%m-%d-%H%M)" >> $GITHUB_ENV
-          echo "RUN_TIMESTAMP=$(date +%Y-%m-%d-%H%M)" >> $GITHUB_ENV
-
       # Deploy PUDL image to GCE
       - name: Deploy
         env:
           DAGSTER_PG_PASSWORD: ${{ secrets.DAGSTER_PG_PASSWORD }}
-          PUDL_OUTPUT_PATH: ${{ env.GCS_OUTPUT_BUCKET }}/${{ env.RUN_TIMESTAMP }}-${{ env.SHORT_SHA }}-${{ env.COMMIT_BRANCH }}
+          PUDL_OUTPUT_PATH: ${{ env.GCS_OUTPUT_BUCKET }}/${{ env.BUILD_ID }}
         run: |-
           gcloud compute instances add-metadata "$GCE_INSTANCE" \
             --zone "$GCE_INSTANCE_ZONE" \
@@ -152,7 +148,7 @@ jobs:
         uses: slackapi/slack-github-action@v1
         with:
           channel-id: "C03FHB9N0PQ"
-          slack-message: "build-deploy-pudl status: ${{ job.status }}\n${{ env.GCS_OUTPUT_BUCKET }}/${{ env.RUN_TIMESTAMP}}-${{ env.SHORT_SHA }}-${{ env.COMMIT_BRANCH }}"
+          slack-message: "build-deploy-pudl status: ${{ job.status }}\n${{ env.GCS_OUTPUT_BUCKET }}/${{ env.BUILD_ID }}"
         env:
           channel-id: "C03FHB9N0PQ"
           SLACK_BOT_TOKEN: ${{ secrets.PUDL_DEPLOY_SLACK_TOKEN }}
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -118,8 +118,8 @@ jobs:
             --container-arg="bash" \
             --container-arg="./docker/gcp_pudl_etl.sh" \
             --container-env-file="./docker/.env" \
-            --container-env ACTION_SHA=$ACTION_SHA \
             --container-env GITHUB_REF=${{ env.GITHUB_REF }} \
+            --container-env BUILD_ID=${{ env.BUILD_ID }} \
             --container-env NIGHTLY_TAG=${{ env.NIGHTLY_TAG }} \
             --container-env GITHUB_ACTION_TRIGGER=${{ github.event_name }} \
             --container-env SLACK_TOKEN=${{ secrets.PUDL_DEPLOY_SLACK_TOKEN }} \
--- a/.github/workflows/docker-build-test.yml
+++ b/.github/workflows/docker-build-test.yml
@@ -17,20 +17,22 @@ jobs:
 
       - name: Docker Metadata
         id: docker_metadata
-        uses: docker/metadata-action@v5.3.0
+        uses: docker/metadata-action@v5
         with:
           images: catalystcoop/pudl-etl
           flavor: |
             latest=auto
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3.0.0
+        uses: docker/setup-buildx-action@v3
 
       - name: Build image but do not push to Docker Hub
-        uses: docker/build-push-action@v5.1.0
+        uses: docker/build-push-action@v5
         with:
           context: .
           file: docker/Dockerfile
           push: false
           cache-from: type=gha
           cache-to: type=gha,mode=max
+          secrets: |
+            "GITHUB_REF=${{ github.ref_name }}"
--- a/.github/workflows/docker-build-test.yml
+++ b/.github/workflows/docker-build-test.yml
@@ -34,5 +34,3 @@ jobs:
           push: false
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          secrets: |
-            "GITHUB_REF=${{ github.ref_name }}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,7 +83,7 @@ jobs:
           name: python-package-distributions
           path: dist/
       - name: Sign dist with Sigstore
-        uses: sigstore/gh-action-sigstore-python@v2.1.0
+        uses: sigstore/gh-action-sigstore-python@v2
         with:
           inputs: ./dist/*.tar.gz ./dist/*.whl
       - name: Upload artifact signatures to GitHub Release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,9 +98,10 @@ jobs:
       - name: Tag the current stable commit
         run: |
           git config user.email "pudl@catalyst.coop"
-          git config user.name "PudlBot"
-          git tag -a -m "stable" stable
-          git push origin stable
+          git config user.name "pudlbot"
+          git checkout stable
+          git merge --ff-only ${{ github.ref_name }}
+          git push
 
   notify-slack:
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -95,11 +95,12 @@ jobs:
           --discussion-category Announcements
           --generate-notes
           --repo '${{ github.repository }}'
-      - name: Tag stable release
-        uses: EndBug/latest-tag@1
-        with:
-          ref: stable
-          description: "The current stable release of catalystcoop.pudl"
+      - name: Tag the current stable commit
+        run: |
+          git config user.email "pudl@catalyst.coop"
+          git config user.name "PudlBot"
+          git tag -a -m "stable" stable
+          git push origin stable
 
   notify-slack:
     runs-on: ubuntu-latest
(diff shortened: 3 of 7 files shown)


.github/workflows/build-deploy-pudl.yml
```diff
@@ -26,6 +26,14 @@ jobs:
         if: ${{ (github.event_name == 'schedule') }}
         run: |
           echo "This action was triggered by a schedule." && echo "GCE_INSTANCE=pudl-deployment-dev" >> $GITHUB_ENV && echo "GITHUB_REF=dev" >> $GITHUB_ENV
+          echo "RUN_DATE=$(date +%Y-%m-%d)" >> $GITHUB_ENV
+
+      - name: Tag nightly build
+        if: ${{ (github.event_name == 'schedule') }}
+        uses: EndBug/latest-tag@1
```
jdangerx, line 33: Do you think we would be able to use `run: git tag | git push <tag>` here instead of a third-party action? Or are there weird issues with that?

jdangerx (commented):
Stoked for this to get in, I think probably making a limited bot token makes the most sense for auth!

.github/workflows/build-deploy-pudl.yml
zaneselvans: What does piping the output of `git tag` to `git push` do?

I imagine there's some way we can run `git tag nightly-2023-12-20` and `git push nightly-2023-12-20` but I guess I don't know whether there's other authentication or git configuration magic that has to be done to make that work. Who is the user that's running these commands and what permissions do they have on the repository and how do you communicate that authentication information to the shell that's being used to run the commands? I guess I can try with a test tag and then delete it.

zaneselvans: It does like that: 
```
Committer identity unknown
*** Please tell me who you are.
Run
  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"
to set your account's default identity.
```

.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
   # Formatters: hooks that re-write Python & documentation files
   ####################################################################################
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.7
+    rev: v0.1.8
```
zaneselvans, line 32: I just happened to `pre-commit autoupdate` nothing to see here.

.github/workflows/update-conda-lockfile.yml
```diff
@@ -4,17 +4,16 @@ name: update-conda-lockfile
...
 
 # What branch does this action run on?
 # - workflow_dispatch: Whatever branch it was run against.
 # - schedule: Always the same branch (will be dev or main)
-# - push: Base branch of the PR.
 
 jobs:
   update-conda-lockfile:
     runs-on: ubuntu-latest
-    if: ${{ (github.event_name == 'push' && github.actor != 'pudlbot') || (github.event_name == 'schedule' && github.repository == 'catalyst-cooperative/pudl') || (github.event_name == 'workflow_dispatch') }}
```
zaneselvans, line 17: I removed all of the on `push` logic, since the `pre-commit` hooks seem to be working well, so this workflow will just be used to ensure our dependencies get updated once a week on a schedule.

docker/gcp_pudl_etl.sh
```diff
@@ -68,9 +68,13 @@ function copy_outputs_to_gcs() {
 }
 
 function copy_outputs_to_distribution_bucket() {
+    echo "Removing old outputs from GCP distributon bucket."
+    gsutil -m -u $GCP_BILLING_PROJECT rm -r "gs://pudl.catalyst.coop/$GITHUB_REF"
     echo "Copying outputs to GCP distribution bucket"
     gsutil -m -u $GCP_BILLING_PROJECT cp -r "$PUDL_OUTPUT/*" "gs://pudl.catalyst.coop/$GITHUB_REF"
 
+    echo "Removing old outputs from AWS distributon bucket."
+    aws s3 rm "s3://pudl.catalyst.coop/$GITHUB_REF" --recursive
```
zaneselvans, line 77: We've had a couple of instances where we've removed a file, or changed its name, and the old version hangs around because we're relying on new files overwriting old ones to get rid of the old versions.  These explicit purges of all previously distributed files from the `dev` or soon `nightly` paths should avoid that problem.

docker/gcp_pudl_etl.sh
```diff
@@ -107,6 +111,12 @@ copy_outputs_to_gcs
 
 # if pipeline is successful, distribute + publish datasette
 if [[ $ETL_SUCCESS == 0 ]]; then
+    if [ $GITHUB_ACTION_TRIGGER = "schedule" ]; then
+        # Tag the nightly build
+        git config user.email "pudl@catalyst.coop"
+        git config user.name "PudlBot"
+        git tag -a -m "The most recent successful nightly build." nightly $GITHUB_REF
+        git push origin nightly
```
zaneselvans, line 119: - I don't think this will work, because I don't think what we have inside the Docker container is a git repo.
- We also need to somehow make use of the `PUDL_BOT_PAT` to authenticate here.
- Also, we want `nightly` to be a branch, not a tag, so the action we take here needs to be a merge or a rebase or something rather than just a tag.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -127,6 +137,7 @@ jobs:
             --container-env DAGSTER_PG_HOST="104.154.182.24" \
             --container-env DAGSTER_PG_DB="dagster-storage" \
             --container-env FLY_ACCESS_TOKEN=${{ secrets.FLY_ACCESS_TOKEN }} \
+            --container-env PUDL_BOT_PAT=${{ secrets.PUDL_BOT_PAT }} \
```
zaneselvans, line 140: Pass in a token so we can do `git` stuff within the container.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -45,7 +55,7 @@ jobs:
 
       - name: Docker Metadata
         id: docker_metadata
-        uses: docker/metadata-action@v5.3.0
+        uses: docker/metadata-action@v5
```
zaneselvans, line 58: I removed the minor versions to reduce the frequency with which dependabot makes PRs. We haven't (yet) had issues with github actions breaking during minor version updates.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -26,12 +26,22 @@ jobs:
...
           ref: ${{ env.GITHUB_REF }}
 
+      - name: Tag nightly build
+        if: ${{ (github.event_name == 'schedule') }}
+        run: |
+          git config user.email "pudl@catalyst.coop"
+          git config user.name "PudlBot"
+          git tag -a -m "$NIGHTLY_TAG" $NIGHTLY_TAG $GITHUB_REF
+          git push origin $NIGHTLY_TAG
+
```
zaneselvans, line 44: I *think* this is doing what we want it to now.  It's a tag, it's applied to the branch that we're going to build (not `main` which is checked out here).

zaneselvans (commented):
I've got *something* here in all the places I think we'd need to update to get the build and release tagging working:
- `vYYYY.MM.DD` tags are created manually.
- `nightly-YYYY-MM-DD` tags are created by the workflow, before the build is kicked off.
- The `nightly` branch (which is currently a tag) has to be updated from within the Docker container, but we don't currently have a git repo inside the Docker container.
- The `stable` branch (also currently a tag) will be updated by the `release` workflow.

Questions in review comments.

docker/gcp_pudl_etl.sh
bendnorman: The dockerfile does copy over `.git/` so it might work?

.github/workflows/release.yml
```diff
@@ -95,6 +95,12 @@ jobs:
           --discussion-category Announcements
           --generate-notes
           --repo '${{ github.repository }}'
+      - name: Tag the current stable commit
+        run: |
+          git config user.email "pudl@catalyst.coop"
+          git config user.name "PudlBot"
+          git tag -a -m "stable" stable
+          git push origin stable
```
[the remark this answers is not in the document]
jdangerx: I think what we want is: `stable` as a branch that always points at the latest `v202X.XX.XX` tag.

As to their relationship to `main`, I think we need to zoom out and think about branches and tags as **pointers to commits**. So if the branch `stable` points at the tag `v2023.12.01`, which points at the commit `f96336`, then `stable` will have the same history as `f96336`. So, to answer your questions:

> What do we do to update them?

`git checkout stable && git merge --ff-only <TAG> && git push` will update `stable` to point at the same commit as `<TAG>`, **IF** `stable` is in the history of `<TAG>`. This should work for our use case almost all of the time.

When we destroy `dev` and only use `main` we might have to do a little bit of manual work with `git checkout stable && git reset <TAG> && git push -f`.

> Do we merge `main` into them? How does that work if we're only occasionally bringing over work from main?

Not explicitly. When we merge `main` into `dev`, and then tag a release from `dev`, `stable` will eventually point at that tag, which will have the merge from `main` in its history.

> Do we want all of the intermediate commits?

Yeah, those will be along for the ride too. Again, `stable` is just a pointer to an existing commit. We're not in the business of making some sort of parallel branch that has the same code but different commits from `dev`.

> Or only snapshots of the state of main when a nightly build was run, or a release was cut?

No, I think this is served by the tags which are supposed to stay stable.

> And are those branches essentially read-only, except for the automated processes that update them?

We would want to make the `nightly` and `stable` branches only pushable by repo admins + the bot. I think we can achieve this with GitHub branch protections.

.github/workflows/update-conda-lockfile.yml
jdangerx: Sweet. Thanks, I forgot to clean this up when I removed the `push` trigger.

docker/gcp_pudl_etl.sh
jdangerx: The dockerfile *mounts* `.git` explicitly to avoid copying the repo history etc. over - so I think you're right about the not-a-repo thing.

I think you're right about using the PAT in place of a password: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#using-a-personal-access-token-on-the-command-line

We can use the PAT to do a fresh `git clone` and do all our fancy git stuff using that copy of the repo, I think.

docker/gcp_pudl_etl.sh
zaneselvans: On second thought, this is actually a dangerous operation. In the event that `$GITHUB_REF` is unset, it will wipe out the entire bucket, so I am inclined to hard-code `dev` here and update it to `nightly`/`stable` when we switch over to using those.

docker/gcp_pudl_etl.sh
zaneselvans: @jdangerx do you mean that **instead** of copying over the file hierarchy when building the container, we should do like `git clone --depth 2` inside the Docker container when it's running, since we're going to need to have git credentials inside it anyway?  Or that we should do the `git clone` in the Dockerfile so it happens when we're building the image? Or that we should both copy the repository files in, and then also `git clone` on top of them?

Looking at the `Dockerfile`:

```Dockerfile
# Copy dagster configuration file
COPY docker/dagster.yaml ${DAGSTER_HOME}/dagster.yaml

# Create a conda environment based on the specification in the repo
COPY environments/conda-lock.yml environments/conda-lock.yml
RUN micromamba create --prefix ${CONDA_PREFIX} --yes --file environments/conda-lock.yml && \
    micromamba clean -afy
# Copy the cloned pudl repository into the user's home directory
COPY --chown=${MAMBA_USER}:${MAMBA_USER} . ${CONTAINER_HOME}
```

- It seems like we need to copy the `dagster.yaml` into the appropriate place, since it's not inside the repo. But it could be copied to the correct location from within the container rather than from a cloned repo outside the container. Is there still a need to check out the repo in the github workflow? Or could it only happen inside the Docker container?
- If we're copying the entire repository into the image, why is `conda-lock.yml` called out specifically?
- Does the final `COPY` of `.` **not** get the `.git` directory? If it **does** get the `.git` directory, then why are we subsequently mounting it before installing PUDL:

```Dockerfile
# We need information from .git to get version with setuptools_scm so we mount that
# directory without copying it into the image.
RUN --mount=type=bind,source=.git,target=${PUDL_REPO}/.git \
    ${CONDA_RUN} pip install --no-cache-dir --no-deps --editable .
```

If we need the whole repository inside the Docker container, could we do something like this in the Dockerfile?

```Dockerfile
RUN git clone --depth 2 --single-branch --branch $GITHUB_REF https://github.com/catalyst-cooperative/pudl.git
RUN git remote set-url origin https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git
RUN git config user.email "pudl@catalyst.coop"
RUN git config user.name "pudlbot"
```

.github/workflows/build-deploy-pudl.yml
zaneselvans: I think that's close, except that we **only** distribute the most recent nightly build and tagged releases to AWS. I imagine we would also have a copy of the most recent release in a `stable/` directory.  So if you want to refer to `nightly-202x-xx-xx` you would need to look at the `gs://builds.catalyst.coop/nightly-202x-xx-xx` path where the last 30 days of nightly build outputs are stashed, regardless of whether they were successful or not. So if you want to look at nightly build outputs + the code that generated them, you can look at the GCS output, and the tagged commit that matches its name.

.github/workflows/release.yml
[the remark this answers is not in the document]
zaneselvans: I think I'm confused about what it means to merge a tag into a branch?

```
git checkout stable && git merge --ff-only <TAG>
```

If I understand correctly:

- We're going to start running nightly builds from the HEAD of `main`.
- Before we kick off the build, we'll tag the commit to be built from with `nightly-YYYY-MM-DD`
- If the build succeeds, we'll update the `nightly` branch to point at the `nightly-YYYY-MM-DD` tag.
- Occasionally, we'll manually push a `vYYYY.MM.DD` tag to a recent commit that was already tagged `nightly-YYYY-MM-DD` and which was successful (so once upon a time it was the HEAD of `nightly`) and also update `stable` to point at that tag.

So if we update the `nightly` or `stable` branches by merging a newly tagged `nightly-YYYY-MM-DD` or `vYYYY.MM.DD` commit into them, and all of these tagged commit were at one point the HEAD of `main`, then is the idea that the `nightly` and `stable` branches are selected old versions of `main`?

It seems like we won't know which commits are destined to become releases until they are no longer at the HEAD of `main`, and so any final cleanup (like wrapping up the release notes) would mean merging new changes into an old commit. Or is there a neater way to deal with that?

docker/gcp_pudl_etl.sh
zaneselvans: Okay, now I'm checking that `$GITHUB_REF` is a nonzero length string at least.

.github/workflows/docker-build-test.yml
```diff
@@ -17,20 +17,22 @@ jobs:
...
-        uses: docker/build-push-action@v5.1.0
+        uses: docker/build-push-action@v5
         with:
           context: .
           file: docker/Dockerfile
           push: false
           cache-from: type=gha
           cache-to: type=gha,mode=max
+          secrets: |
+            "GITHUB_REF=${{ github.ref_name }}"
```
zaneselvans, line 38: Is there no way to just pass an environment variable in?

https://docs.docker.com/build/ci/github-actions/secrets/

.github/workflows/build-deploy-pudl.yml
```diff
@@ -127,8 +138,9 @@ jobs:
             --container-env DAGSTER_PG_HOST="104.154.182.24" \
             --container-env DAGSTER_PG_DB="dagster-storage" \
             --container-env FLY_ACCESS_TOKEN=${{ secrets.FLY_ACCESS_TOKEN }} \
+            --container-env PUDL_BOT_PAT=${{ secrets.PUDL_BOT_PAT }} \
             --container-env ZENODO_SANDBOX_TOKEN_PUBLISH=${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }} \
-            --container-env PUDL_SETTINGS_YML="/home/mambauser/src/pudl/package_data/settings/etl_full.yml" \
+            --container-env PUDL_SETTINGS_YML="/home/mambauser/pudl/src/pudl/package_data/settings/etl_full.yml" \
```
zaneselvans, line 143: The cloned repo is copied into its own directory now, rather than having the input/output/dagster_home directories munged with whatever is in the repo.

.github/workflows/release.yml
```diff
@@ -95,6 +95,13 @@ jobs:
           --discussion-category Announcements
           --generate-notes
           --repo '${{ github.repository }}'
+      - name: Tag the current stable commit
+        run: |
+          git config user.email "pudl@catalyst.coop"
+          git config user.name "pudlbot"
+          git checkout stable
+          git merge --ff-only ${{ github.ref_name }}
+          git push
```
zaneselvans, line 104: Update the `stable` branch as part of publishing the release to github.

docker/Dockerfile
```diff
@@ -30,28 +30,27 @@ ENV PUDL_INPUT=${CONTAINER_PUDL_WORKSPACE}/input
 ENV PUDL_OUTPUT=${CONTAINER_PUDL_WORKSPACE}/output
 ENV DAGSTER_HOME=${CONTAINER_PUDL_WORKSPACE}/dagster_home
 
-RUN mkdir -p ${PUDL_INPUT} ${PUDL_OUTPUT} ${DAGSTER_HOME}
+RUN mkdir -p ${PUDL_INPUT} ${PUDL_OUTPUT} ${DAGSTER_HOME} ${PUDL_REPO}
```
zaneselvans, line 33: Explicitly create a directory for the PUDL git repo.

docker/Dockerfile
```diff
@@ -30,28 +30,27 @@ ENV PUDL_INPUT=${CONTAINER_PUDL_WORKSPACE}/input
...
 
 # Copy dagster configuration file
 COPY docker/dagster.yaml ${DAGSTER_HOME}/dagster.yaml
 
+# Copy the cloned pudl repository into the container
+# This includes the .git directory, so it is a whole repo
+COPY --chown=${MAMBA_USER}:${MAMBA_USER} . ${PUDL_REPO}
+
```
zaneselvans, line 41: Copy the whole cloned repo into its own directory within the container.

docker/Dockerfile
```diff
@@ -30,28 +30,27 @@ ENV PUDL_INPUT=${CONTAINER_PUDL_WORKSPACE}/input
...
 # Create a conda environment based on the specification in the repo
-COPY environments/conda-lock.yml environments/conda-lock.yml
-RUN micromamba create --prefix ${CONDA_PREFIX} --yes --file environments/conda-lock.yml && \
+RUN micromamba create --prefix ${CONDA_PREFIX} --yes --file ${PUDL_REPO}/environments/conda-lock.yml && \
```
zaneselvans, line 43: Rather than copying just the `conda-lock.yml` file in by itself, use the version which has been copied in along with the whole repository.

docker/Dockerfile
```diff
@@ -30,28 +30,27 @@ ENV PUDL_INPUT=${CONTAINER_PUDL_WORKSPACE}/input
...
-# Copy the cloned pudl repository into the user's home directory
-COPY --chown=${MAMBA_USER}:${MAMBA_USER} . ${CONTAINER_HOME}
 
 # TODO(rousik): The following is a workaround for sudden breakage where conda
 # can't find libraries contained within the environment. It's unclear why!
 ENV LD_LIBRARY_PATH=${CONDA_PREFIX}/lib
-# We need information from .git to get version with setuptools_scm so we mount that
-# directory without copying it into the image.
-RUN --mount=type=bind,source=.git,target=${PUDL_REPO}/.git \
-    ${CONDA_RUN} pip install --no-cache-dir --no-deps --editable .
```
zaneselvans, line 51: This mount was superfluous, since we were in fact copying the whole repo, including the `.git` directory into the image.

docker/Dockerfile
```diff
@@ -30,28 +30,27 @@ ENV PUDL_INPUT=${CONTAINER_PUDL_WORKSPACE}/input
...
+RUN ${CONDA_RUN} pip install --no-cache-dir --no-deps --editable ${PUDL_REPO}
```
zaneselvans, line 49: Install PUDL from the pudl repo directory.

docker/Dockerfile
```diff
@@ -61,6 +60,6 @@ USER $MAMBA_USER
 RUN ${CONDA_RUN} bash -c 'curl -L https://fly.io/install.sh | sh'
 ENV PATH="${CONTAINER_HOME}/.fly/bin:$PATH"
 
-
+WORKDIR ${PUDL_REPO}
```
zaneselvans, line 63: The GCP ETL script expects to be run from within the repository, so make sure that's where we end up at the end of the build.

docker/gcp_pudl_etl.sh
```diff
@@ -48,16 +48,12 @@ function run_pudl_etl() {
...
 function shutdown_vm() {
-    # Copy the outputs to the GCS bucket
-
     upload_file_to_slack $LOGFILE "pudl_etl logs for $ACTION_SHA-$GITHUB_REF:"
-
+    # Shut down the vm instance when the etl is done.
     echo "Shutting down VM."
-    # # Shut down the vm instance when the etl is done.
-    ACCESS_TOKEN=`curl \
+    ACCESS_TOKEN=$(curl \
```
zaneselvans, line 54: I started using `ShellCheck` a bash linter, and apparently backticks are deprecated.

docker/gcp_pudl_etl.sh
```diff
@@ -68,19 +64,25 @@ function copy_outputs_to_gcs() {
 }
 
 function copy_outputs_to_distribution_bucket() {
-    echo "Copying outputs to GCP distribution bucket"
-    gsutil -m -u $GCP_BILLING_PROJECT cp -r "$PUDL_OUTPUT/*" "gs://pudl.catalyst.coop/$GITHUB_REF"
-
-    echo "Copying outputs to AWS distribution bucket"
-    aws s3 cp "$PUDL_OUTPUT/" "s3://pudl.catalyst.coop/$GITHUB_REF" --recursive
+    # Only attempt to update outputs if we have a real value of GITHUB_REF
+    if [ -n "$GITHUB_REF" ]; then
```
zaneselvans, line 68: Only deploy or delete old files if we have a real git ref -- avoids accidentally blowing away the whole bucket if this variable ends up unset somehow.

docker/gcp_pudl_etl.sh
```diff
@@ -97,6 +99,17 @@ function notify_slack() {
...
+    echo "Deployed via workflow_dispatch, testing git authentication!"
+    git config user.email "pudl@catalyst.coop"
+    git config user.name "pudlbot"
+    git remote set-url origin "https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git"
+    git config -l
+    git tag -a -m "Nightly test tag" nightly-tag-test
+    git push origin nightly-tag-test
+    shutdown_vm
+fi
+
```
zaneselvans, line 112: This is temporary -- the idea was to test whether this authentication and tag push worked without running a bunch of unnecessary stuff but... it never seems to have run and I didn't get any error output in the container logs so I don't know what is going on here.

docker/gcp_pudl_etl.sh
```diff
@@ -107,6 +120,15 @@ copy_outputs_to_gcs
...
 if [[ $ETL_SUCCESS == 0 ]]; then
+    if [ $GITHUB_ACTION_TRIGGER = "schedule" ]; then
+        # Update the nightly branch to point at newly successful nightly build tag
+        git config user.email "pudl@catalyst.coop"
+        git config user.name "pudlbot"
+        git remote set-url origin https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git
+        git checkout nightly
+        git merge --ff-only $NIGHTLY_TAG
+        git push
+    fi
```
zaneselvans, line 131: Update the `nightly` branch if/when the nightly builds succeed.

zaneselvans (commented):
@jdangerx Tried to implement the merge-tag-into-branch thing here and cleaned up the Dockerfile a bit. My attempt to test the git authentication from within the container failed though.

zaneselvans:
I went ahead and created `nightly` and `stable` branches and a recently `nightly-2023-12-19` tag where they would currently be, so that if this stuff runs it finds them.

.github/workflows/build-deploy-pudl.yml
jdangerx: Ah, right - that makes sense! Are there any barriers to distributing a few recent nightly builds on AWS as well as GCS?

docker/gcp_pudl_etl.sh
jdangerx: I do like the idea of having this sort of short circuit test thing in the script for now, I think that will make testing this way less painful.

I see some logs corresponding to this in the [Log Explorer...](https://console.cloud.google.com/logs/query;query=resource.type%3D%22gce_instance%22%20resource.labels.zone%3D%22us-west1-a%22%20%2528resource.labels.instance_id%3D%222074811782323017117%22%20OR%20protoPayload.resourceName%3D%22projects%2Fcatalyst-cooperative-pudl%2Fzones%2Fus-west1-a%2Finstances%2Fpudl-deployment-tag%22%2529%0Aseverity%3E%3DDEFAULT%0AjsonPayload.%22cos.googleapis.com%2Fcontainer_name%22%3D%22klt-pudl-deployment-tag-luui%22;cursorTimestamp=2023-12-20T05:33:35.290711021Z;startTime=2023-12-20T05:11:00.000Z;endTime=2023-12-20T06:45:00.000Z?referrer=search&project=catalyst-cooperative-pudl)

That shows a couple things:

1. Hah, it logs out some sensitive information. Oops.
2. It does appear to run, just failing authentication. Are you able to use that `pudlbot:<TOKEN>` method to authenticate locally? I can give it a shot now.

jdangerx: I was able to `git remote add origin-pudlbot https://pudlbot:<TOKEN>@github.com/...` and then push a new branch with `git push -u origin-pudlbot <branch-name>`.

jdangerx: The `git config -l` output in the logs does show an extra config value that I don't have locally: `http.https://github.com/.extraheader=AUTHORIZATION: basic <token>` - that seems suspicious.

jdangerx: I think that basic auth token probably gets copied in as part of the docker build process, since we're copying in the whole `.git` directory which was checked out by the `checkout` GitHub action.

I think we should avoid baking in any sensitive tokens to the Docker image 😉 but we could potentially try removing that `http.https://github.com....` config value and see if it chooses the right auth.

docker/gcp_pudl_etl.sh
```diff
@@ -107,6 +122,15 @@ copy_outputs_to_gcs
...
     # Deploy the updated data to datasette
     if [ $GITHUB_REF = "dev" ]; then
         python ~/devtools/datasette/publish.py 2>&1 | tee -a $LOGFILE
```
jdangerx, line 136: This datasette publishing should also happen whenever we finish a nightly build - not just when `$GITHUB_REF = "dev"`.

zaneselvans: There's a lot of fuzziness around these conditional statements, about when which actions need to happen.  I think we'll need to go over it all in more detail and make the reasoning explicit.

I think the idea here was that we wanted the nightly builds to update datasette, and they always by definition run on `dev`.  But we also wanted to be able to re-deploy Datasette with a `workflow_dispatch` if we run in on `dev` manually.

zaneselvans:
@jdangerx Cleaned up the build script to remove ad-hoc tests. Integrated some of the bugfixes from #3171 here and extended them into the build script -- we construct a single `$BUILD_ID` in the workflow now and pass it in to the container, rather than reconstructing the ID in different ways in different places leading to broken URLs etc...

.github/workflows/build-deploy-pudl.yml
```diff
@@ -25,27 +25,39 @@ jobs:
...
         run: |
-          echo "ACTION_SHA=$(git rev-parse HEAD)" >> $GITHUB_ENV
+          echo "NIGHTLY_TAG=nightly-$(date +%Y-%m-%d)" >> $GITHUB_ENV
+          echo "NIGHTLY_TAG: $NIGHTLY_TAG"
           echo "SHORT_SHA=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+          echo "SHORT_SHA: $SHORT_SHA"
+          echo "BUILD_TIMESTAMP=$(date +%Y-%m-%d-%H%M)" >> $GITHUB_ENV
+          echo "BUILD_TIMESTAMP: $BUILD_TIMESTAMP"
+          echo "BUILD_ID=${BUILD_TIMESTAMP}-${SHORT_SHA}-${GITHUB_REF}
+          echo "BUILD_ID: $BUILD_ID"
```
zaneselvans, line 48: The `BUILD_ID` was being constructed in a couple of different ways in different places which was breaking things. See #3171.  Now it's constructed here and passed around as a unified entity.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -85,17 +97,11 @@ jobs:
       - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@v2
 
-      - name: Determine commit information
-        run: |-
-          echo "COMMIT_BRANCH=$(echo ${GITHUB_REF#refs/heads/} | tr / -)" >> $GITHUB_ENV
-          echo "COMMIT_TIME=$(git log -1 --format=%cd --date=format:%Y-%m-%d-%H%M)" >> $GITHUB_ENV
-          echo "RUN_TIMESTAMP=$(date +%Y-%m-%d-%H%M)" >> $GITHUB_ENV
-
```
zaneselvans, line 93: Some of this isn't necessary, and some of it was consolidated above.

docker/gcp_pudl_etl.sh
```diff
@@ -12,128 +12,141 @@ function send_slack_msg() {
...
         ETL_SUCCESS=${PIPESTATUS[0]}
     fi
 
     # Compress the SQLite DBs for easier distribution
     # Remove redundant multi-file EPA CEMS outputs prior to distribution
-    gzip --verbose $PUDL_OUTPUT/*.sqlite && \
-    rm -rf $PUDL_OUTPUT/hourly_emissions_epacems/ && \
-    rm -f $PUDL_OUTPUT/metadata.yml
+    gzip --verbose "$PUDL_OUTPUT"/*.sqlite && \
+    rm -rf "$PUDL_OUTPUT/core_epacems__hourly_emissions/" && \
```
zaneselvans, line 132: Whoops, manual renaming of things succccks.  Gotta go delete these directories in the deployed outputs.

docker/Dockerfile
jdangerx: ah! how confusing.

docker/gcp_pudl_etl.sh
```diff
@@ -12,128 +12,141 @@ function send_slack_msg() {
...
+        git remote set-url origin "https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git"
+        # Update the nightly branch to point at newly successful nightly build tag
+        git checkout nightly
+        git merge --ff-only "$NIGHTLY_TAG"
+        git push
+    fi
     # Deploy the updated data to datasette
-    if [ $GITHUB_REF = "dev" ]; then
-        python ~/devtools/datasette/publish.py 2>&1 | tee -a $LOGFILE
+    if [ "$GITHUB_REF" = "dev" ]; then
```
jdangerx, line 124: Should this be rolled in to the above `if` block? We're using `"$GITHUB_REF" = "dev"` to mean "is scheduled nightly build", right?

jdangerx (changes requested):
Just one blocking question about the datasette build trigger. I think we can knock out like half of #3140 in one PR though once this gets merged!

.github/workflows/build-deploy-pudl.yml
zaneselvans: We have a limited storage quota on AWS, and I don't think we actually want the public looking at our nightly build outputs, which may or or may not correspond to successful runs.

docker/Dockerfile
zaneselvans: I'm not sure I understand.

You mean that by copying in the single file, and then building the container, we get a cached layer in the docker image that corresponds to the python environment having been installed (so long as the `conda-lock.yml` didn't change since last time we did it) but that precedes the copying of the rest of the  files in the repository over (which will change any time anything changes in the repo)?

docker/gcp_pudl_etl.sh
zaneselvans: We could do that, but that would mean that the *only* way to redeploy Datasette is for the scheduled build to run. Whereas with this condition, if you deploy `dev` with `workflow_dispatch` it will redeploy Datasette upon success (and it will also be redeployed if the nightly builds succeed, since they only run on `dev`)

docker/Dockerfile
zaneselvans: Okay, I've reinstated this ordering that allows the conda environment to be cached, and added a comment explaining why it's separate.

docker/Dockerfile
zaneselvans: Switched this back to allow caching of conda env.

docker/gcp_pudl_etl.sh
zaneselvans: But in general there's a lot of fuzziness with many of these conditionals, and it's not always clear what the intended behavior is, with some of them looking at `GITHUB_REF` and others looking at the action trigger.

jdangerx: OK. We can leave this as is for now, but as we are consolidating onto `main` we'll have to change it.

I liked @rousik 's suggestion (somewhere...) of having a script with a lot of actual logical arguments, something sort of like
```
./run_the_frickin_nightly_build.py --etl-fast --deploy-to-datasette --publish-to-gcp --publish-to-aws --update-branch nightly
```

And then building that command line in the GHA workflow. So then at least it's clear what logical thing we mean by `"$GITHUB_REF" = "dev"`

jdangerx (approved):

docker/gcp_pudl_etl.sh
zaneselvans: There are a bunch of hard-coded references to `dev` that we'll need to update.  I agree that setting all of the options in one place would be nice. And getting all of this into Python would be nice (I think we could ditch installing the AWS CLI and Google Cloud SDK with a simple deploy-to-buckets script). Definitely having 3-4 places with logic (workflow, Dockerfile, shell script, pudl_etl python script) isn't the best!

PR #3171, opened 2023-12-18
Restructure build-deploy action

rousik:
Move computation of various env variables into single step, print them out and use slightly different logic for calculating branch name.


zaneselvans (commented):
There are some additional references to these environment variables / paths that happen inside `gcp_pudl_etl.sh` which also need to be updated.

zaneselvans:
I adapted these fixes and updates to the build script into #3164 which just got merged into `main` along with some other changes to get the `nightly` and `stable` branches automatically updating w/ builds and tagged releases, so I'm going to close this.

PR #3178, opened 2023-12-19
Update EPA CEMS docs for quarterly data, new table name.

zaneselvans:
# Overview

- Remove references to defunct PUDL Catalog
- Remove link to defunct EPA FTP site.
- Update link to EPA CEMS nightly build outputs
- Remove duplicative and flaky source vs. working partitions in data source template. Just show temporal coverage.
- Update EPA CEMS row count

# Testing

Ran the docs build locally and inspected the output.

```[tasklist]
# To-do list
```

### DIFF
--- a/docs/templates/epacems_child.rst.jinja
+++ b/docs/templates/epacems_child.rst.jinja
@@ -60,8 +60,8 @@ The files are available in a prepackaged format, accessible via a `user interfac
 
 {% block notable_irregularities %}
 
-CEMS is enourmous
------------------
+CEMS is enormous
+----------------
 CEMS is by far the largest dataset in PUDL what with hourly records for thousands of
 plants spanning decades. For this reason, we distribute CEMS data using `Apache Parquet
 <https://parquet.apache.org/>`__ files rather than in the main PUDL database.
--- a/docs/templates/epacems_child.rst.jinja
+++ b/docs/templates/epacems_child.rst.jinja
@@ -1,11 +1,10 @@
 {% extends "data_source_parent.rst.jinja" %}
 {% block database_text %}
 Clicking on the links will show you a description of the table as well as
-the names and descriptions of each of its fields. You can access the data via the
-EPACEMS Intake catalog.
+the names and descriptions of each of its fields.
 {% endblock %}
 
-{% block browse_online %}- Available via `PUDL Data Catalog <https://github.com/catalyst-cooperative/pudl-catalog>`__{% endblock %}
+{% block browse_online %}- Not browsable due to file size.{% endblock %}
 
 {% block background %}
 `Continuous Emissions Monitoring Systems
@@ -25,10 +24,9 @@ Irregularities"; it gets complicated.
 {% endblock %}
 
 {% block accessible %}
-All of it! Data is currently available via the `PUDL Data Catalog <https://github.com/catalyst-cooperative/pudl-catalog>`__.
+All of it! We typically update the EPA CEMS dataset quarterly, as new data releases are
+made.
 
-Thanks to `Karl Dunkle Werner <https://github.com/karldw>`__ for contributing
-much of the EPA CEMS Hourly ETL code!
 {% endblock %}
 
 {% block fill_out_form_title %}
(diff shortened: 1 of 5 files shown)


docs/templates/epacems_child.rst.jinja
```diff
@@ -56,18 +54,17 @@ A plain English explanation of the requirements of Part 75 is available in secti
 
 {% block original_data %}
 EPA CAMD publishes the CEMS data in an online `data portal <https://campd.epa.gov/>`__.
-The files are available in a prepackaged format, accessible via a `user interface <https://campd.epa.gov/data/custom-data-download>`__
-or `FTP site <https://gaftp.epa.gov/DMDnLoad/>`__ with each downloadable zip file
-encompassing a year of data.
+The files are available in a prepackaged format, accessible via a `user interface
+<https://campd.epa.gov/data/custom-data-download>`__.
```
e-belfer, line 58: Worth mentioning the API here, which is how we actually access the data?
https://www.epa.gov/power-sector/cam-api-portal

docs/templates/epacems_child.rst.jinja
```diff
@@ -56,18 +54,17 @@ A plain English explanation of the requirements of Part 75 is available in secti
...
 {% block notable_irregularities %}
 
-CEMS is enourmous
------------------
-CEMS is by far the largest dataset in PUDL what with hourly records for
-thousands of plants spanning decades. For this reason, we distribute CEMS data using `Apache
-Parquet <https://parquet.apache.org/>`__ files rather than in the main PUDL database.
+CEMS is enormous
+----------------
+CEMS is by far the largest dataset in PUDL what with hourly records for thousands of
```
e-belfer, line 65: Can we also delete the "what" in "what with" here? It's confusing and extraneous

docs/templates/epacems_child.rst.jinja
```diff
@@ -56,18 +54,17 @@ A plain English explanation of the requirements of Part 75 is available in secti
...
+plants spanning decades. For this reason, we distribute CEMS data using `Apache Parquet
+<https://parquet.apache.org/>`__ files rather than in the main PUDL database.
```
e-belfer, line 67: I find this terminology confusing - it is part of our 'main database', just not part of the main PUDL SQLite file. Can we just say, "rather than using SQLite" here?

src/pudl/metadata/classes.py
```diff
@@ -927,7 +931,6 @@ def add_datastore_metadata(self) -> None:
             partitions["years"] = partitions["year"]
         elif "year_month" in partitions:
             partitions["year_month"] = max(partitions["year_month"])
-        self.source_file_dict["source_years"] = self.get_temporal_coverage(partitions)
```
e-belfer, line 930: Is this function getting called elsewhere? If not, why get rid of it here? Is this the change referenced in this line of your PR, "Remove duplicative and flaky source vs. working partitions in data source template. Just show temporal coverage"?

src/pudl/metadata/sources.py
```diff
@@ -268,7 +268,7 @@
         ),
         "source_file_dict": {
             "respondents": "Coal and high-sulfur fueled plants over 25MW",
-            "records_liberated": "~800 million",
+            "records_liberated": "924 million",
```
e-belfer, line 271: I suspect we're going to be inconsistent about updating this each quarter, unless we add it to our "add new data" issue template task list - suggest either rounding to ~900 million or ~1 billion here or adding it to the template to update with regularity.

e-belfer (changes requested):
A few tiny changes to make and one clarifying question. Otherwise all looks good.

docs/templates/epacems_child.rst.jinja
zaneselvans: Updated language.

docs/templates/epacems_child.rst.jinja
zaneselvans: Added a link to the CAM API.

src/pudl/metadata/classes.py
zaneselvans: It's called in several places, including the Jinja templates for the Datasette Metadata and the individual data source documentation pages.

I removed it here because it was in effect being called twice -- once to generate these source years, and again to generate the "working years" which used to be referenced separately (since there are some years for which data is available, where we haven't extracted that data) but when we went to using the templates, both of them were set to reference the working years. So the same range showed up more than once on the data pages, and also were always called "years" even though we now have quarters, months, and years.

src/pudl/metadata/sources.py
zaneselvans: Rounded to order of magnitude. Would be nice to generate these things automatically.

docs/templates/epacems_child.rst.jinja
zaneselvans: Yep.

e-belfer (approved):
Looks great, thanks for catching this and cleaning it up!

src/pudl/metadata/sources.py
e-belfer: Agreed, definitely out of scope but would be nice.

Issue #3179, opened 2023-12-20
Deprecate and remove `dev` branch


jdangerx:
```[tasklist]
### Tasks
- [x] scour documentation for references to `dev` branch
- [x] remove `dev` from RTD
- [x] Update nightly build infrastructure to run from `main` instead of `dev`
- [x] Set `update-conda-lockfile` to run against `main` rather than `dev`
- [x] Set `pre-commit-ci` to run against `main` rather than `dev`
- [x] Set dependabot to run against `main` rather than `dev`
- [x] Update nightly build script to output to `/nightly` rather than `/dev` when running on a schedule
- [x] Explain the new git workflow to internal folks
- [x] Reset `main` to `dev` so that they share history, protect the `dev` branch so nobody can push to it
- [x] Set up appropriate branch protections for `main`, `nightly`, and `stable`
- [x] Delete the old `gs://nightly-build-outputs.catalyst.coop` bucket
- [x] Update header information in Datasette to point at `nightly` not `dev`
- [x] Update PUDL dataset on Kaggle to point at `nightly` outputs
- [x] Tell external people to pin to `main`, `nightly`, or `stable`, instead of `dev` branch
- [x] Get #2949 reopened against `main`
- [x] Remove `/dev` outputs from GCS and S3 in favor of `/stable` and `/nightly` outputs
```

zaneselvans:
What do we think the [appropriate branch protections](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#about-branch-protection-rules) are for `main`, `nightly`, and `stable`?

It seems like `nightly` and `stable` should only every be modified by the automated processes (unless something goes awry).

The protections we have on `dev` right now seem like they would probably get moved to `main`?  Namely that it has to be changed by a PR with at least one reviewer, and all the "required" CI checks need to pass?

zaneselvans:
Verifying that our first successful main/nightly/stable build behaved as expected:
- [ ] nightly-YYYY-MM-DD tag was created
- [ ] nightly branch was updated to point at main
- [ ] build outputs were written to gs://builds.catalyst.coop/BUILD_ID
- [ ] appropriate build outputs were copied to gs://pudl.catalyst.coop/nightly
- [ ] appropriate build outputs were copied to s3://pudl.catalyst.coop/nightly
- [ ] Datasette was redeployed with build outputs

zaneselvans:
## Issues:

[...]

PR #3181, opened 2023-12-20
Remove references to `dev` branch in preparation for deprecation

jdangerx:
# Overview

Part of #3179.

Remove references to `dev` branch, since we'll be using `main`/`stable`/`nightly` instead.

* in git workflow docs (replaced with main)
* in readthedocs URLs (replaced with /stable/)
* in GHA workflows (replaced with main)
* in gcp_pudl_etl.sh (replaced with main)
* in datasette template (replaced with main, but maybe should point to latest or nightly?)

If the builds from #3164 pass, and the `nightly` branch gets updated, we will merge this as the final PR into `dev` and lock the `dev` branch. Then we'll force `main` to point at the tip of `dev` and all future development will work from there.

# Testing

We need to see if the nightly builds do successfully build on `main` instead of `dev` after we do all the finagling above. We can't use a tag build to test that before we merge this PR, because a lot of the stuff we're testing is specific to `main`, but we can kick off a tag build immediately after we move `main`.

```[tasklist]
# To-do list
```


zaneselvans:
We'll also need to change the `dependabot` and `pre-commit-ci` target branch from `dev` to `main`

PR #3183, opened 2023-12-21
Fix errors in new build-deploy-pudl workflow

zaneselvans:
# Overview
Note, many changes originally here were merged in #3188 

- Update the `nightly` branch if the scheduled nightly build is successful.
- Use `[[ modern bash conditionals ]]` throughout the script.
- Track success of various different steps in separate environment variables.
- Use the `&&` to combine exit statuses of the commands inside functions so we can track the success/failure of the entire chain of commands.
- Send output from all of the functions to the logfile in the same way.

```[tasklist]
# To-do list
```


.github/workflows/build-deploy-pudl.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 env:
   GCP_BILLING_PROJECT: ${{ secrets.GCP_BILLING_PROJECT }}
-  GITHUB_REF: ${{ github.ref_name }} # This is changed to dev if running on a schedule
+  BUILD_REF: ${{ github.ref_name }} # This is changed to dev if running on a schedule
```
zaneselvans, line 12: It seemed like our use of `GITHUB_REF` as a variable name was shadowing an existing environment variable. When we re-set `GITHUB_REF` to equal `dev` below, it would be `dev`, but when it wasn't overridden, `GITHUB_REF` would contain the full ref e.g. `/refs/heads/main` rather than the shortened version (`main`) that's supposed to be found in `github.ref_name`.  In any case, re-using the name and making it refer to something with different properties was confusing, so I renamed it to `BUILD_REF`

.github/workflows/build-deploy-pudl.yml
```diff
@@ -19,40 +19,41 @@ jobs:
     name: Build Docker image, push to Docker Hub and deploy to a GCE VM
     runs-on: ubuntu-latest
     permissions:
-      contents: read
+      contents: write
```
zaneselvans, line 22: Write permissions required to push the `NIGHTLY_TAG`

.github/workflows/build-deploy-pudl.yml
```diff
@@ -19,40 +19,41 @@ jobs:
...
         uses: actions/checkout@v4
         with:
-          ref: ${{ env.GITHUB_REF }}
+          ref: ${{ env.BUILD_REF }}
+          fetch-depth: 0
 
       - name: Set action environment variables
         run: |
           echo "NIGHTLY_TAG=nightly-$(date +%Y-%m-%d)" >> $GITHUB_ENV
+          echo "BUILD_ID=$(date +%Y-%m-%d-%H%M)-$(git rev-parse --short HEAD)-${BUILD_REF}" >> $GITHUB_ENV
```
zaneselvans, line 41: We never re-use these subcomponents, and having them lying around is cluttered & tempts someone to try and reconstruct this ID later, which was a source of bugs / internal inconsistencies, so I just set the `BUILD_ID` once here and it's used throughout.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -19,40 +19,41 @@ jobs:
...
-          echo "SHORT_SHA: $SHORT_SHA"
-          echo "BUILD_TIMESTAMP=$(date +%Y-%m-%d-%H%M)" >> $GITHUB_ENV
-          echo "BUILD_TIMESTAMP: $BUILD_TIMESTAMP"
-          echo "BUILD_ID=${BUILD_TIMESTAMP}-${SHORT_SHA}-${GITHUB_REF}
           echo "BUILD_ID: $BUILD_ID"
 
       - name: Tag nightly build
-        if: ${{ (github.event_name == 'schedule') }}
+        # TESTING -- SHOULD ONLY RUN ON SCHEDULE
+        if: ${{ (github.event_name == 'schedule') || (github.event_name == 'workflow_dispatch') }}
```
zaneselvans, line 52: Necessary for testing interactively.

docker/gcp_pudl_etl.sh
```diff
@@ -48,7 +48,7 @@ function run_pudl_etl() {
 }
 
 function shutdown_vm() {
-    upload_file_to_slack "$LOGFILE" "pudl_etl logs for $BUILD_ID:"
+    # upload_file_to_slack "$LOGFILE" "pudl_etl logs for $BUILD_ID:"
```
zaneselvans, line 51: Temporarily commented out, because we're not actually running the build below and thus there's no logfile.

docker/gcp_pudl_etl.sh
```diff
@@ -99,29 +99,43 @@ function notify_slack() {
...
+}
+
+# Short circut the script to debug the nightly branch update
+update_nightly_branch
+if [[ $ETL_SUCCESS == 0 ]]; then
+    notify_slack "success"
+else
+    notify_slack "failure"
+fi
+shutdown_vm
```
zaneselvans, line 123: Only attempting to test the git branch mechanics right now.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -19,40 +19,41 @@ jobs:
...
-          echo "GITHUB_REF=dev" >> $GITHUB_ENV
-          echo "GITHUB_REF: $GITHUB_REF"
+          echo "BUILD_REF=dev" >> $GITHUB_ENV
 
       - name: Checkout Repository
         uses: actions/checkout@v4
         with:
-          ref: ${{ env.GITHUB_REF }}
+          ref: ${{ env.BUILD_REF }}
+          fetch-depth: 0
```
zaneselvans, line 36: Setting the `fetch-depth` to 0 tells the action to check out the whole repo, rather than a shallow copy of just the `BUILD_REF`.  This is necessary in order to have access to the `nightly` branch for merging inside the running container.  I didn't see a way to only check out the two refs we need, and this works fine and doesn't add much time.

docker/gcp_pudl_etl.sh
```diff
@@ -99,29 +99,43 @@ function notify_slack() {
...
+    git config user.name "pudlbot"
+    git remote set-url origin "https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git"
+    echo "WTAF: Updating nightly branch to point at $NIGHTLY_TAG."
+    git checkout nightly
+    git branch --set-upstream-to=origin/nightly
+    git pull
+    git merge --ff-only "$NIGHTLY_TAG"
+    ETL_SUCCESS=${PIPESTATUS[0]}
+    git push
+}
```
zaneselvans, line 114: This feels more involved than it needs to be. Is there a more concise way to do the same thing?  However, when it runs it seems to have the desired effect.

zaneselvans (commented):
@jdangerx @bendnorman The git branch update is nominally working, but something weird is going on with the version of the container.

Something weird is going on that affects what version of the container ends up being deployed to GCE.  To reproduce the behavior:

- Commit a trivial change to the `echo` statement in `update_nightly_branch()` so you can tell whether your new version of the code/container is being used (e.g. `s/WTAF/DUDE/`).
- Re-set the `nightly` branch to point at `main` locally and on GitHub
- Delete the `NIGHTLY_TAG` locally and on GitHub
- Kick off the `build-deploy-pudl` action with a `workflow_dispatch`

The `NIGHTLY_TAG` gets updated reliably, so the action is definitely running. However:

- Sometimes no logs will be appear in the GCP console, and the `nightly` branch will not be updated.
- After a run in which no logs appear, if you delete the `NIGHTLY_TAG` locally and on GitHub, and re-run the action without making any changes to the code or the status of the `nightly` branch, the build script will appear to run producing logs in the GCP console, and updating the `nightly` branch, but the code won't reflect the trivial change you made in the most recent commit.
- It's like `HEAD~1` of `BUILD_REF` is being run, rather than `HEAD`. Like the container that's being run is lagging one version behind after a hiccup that results in nothing running.

docker/gcp_pudl_etl.sh
zaneselvans: I struggled a bit when trying to ensure that the `nightly` branch was able to be fast-forwarded to the new `$NIGHTLY_TAG`. For testing, I pointed `nightly` at `main` and since this branch is coming off of `main` it works. But I don't feel like I really understand how this will work in production.

jdangerx: If:
- `nightly-XXXX-XX` is always an ancestor of `nightly-XXXX-XX+1`
- `nightly` always points at the current `nightly-XXXX` or a (chronologically) earlier one

Then, `nightly` will always be able to be fast-forwarded to the most current `nightly-XXXX-XX`. This is pretty easy to guarantee if `nightly-XXXX-XX` is always a commit on `main` and we don't rewrite history on `main`.

But, right now, since there's the whole `main`/`dev` split, it's a little harder to reason about. Thinking out loud:

- `nightly-XXXX-XX` is always at the tip of `dev` when this action runs.
- we don't move the history of `dev` around
- `nightly` always points at current or earlier `nightly-XXXX`

So I think the `--ff-only` should work.

docker/gcp_pudl_etl.sh
```diff
@@ -99,29 +99,43 @@ function notify_slack() {
...
+function update_nightly_branch() {
+    git config --unset http.https://github.com/.extraheader
+    git config user.email "pudl@catalyst.coop"
+    git config user.name "pudlbot"
+    git remote set-url origin "https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git"
+    echo "WTAF: Updating nightly branch to point at $NIGHTLY_TAG."
+    git checkout nightly
+    git branch --set-upstream-to=origin/nightly
+    git pull
+    git merge --ff-only "$NIGHTLY_TAG"
```
jdangerx, line 111:
```suggestion
    git fetch origin nightly:nightly
    git checkout nightly
    git merge --ff-only "$NIGHTLY_TAG"
```

.github/workflows/build-deploy-pudl.yml
jdangerx: This seems fine, though also the suggestion I added below works with a shallow clone on my machine:

```
> git clone https://github.com/catalyst-cooperative/pudl.git --depth 1 
> cd pudl
> git fetch origin nightly:nightly
> git checkout nightly
```

jdangerx (commented):
Things look fine, I added a tiny git suggestion to use slightly more normal git commands to do what we need to do. I'll take a look at the strange container behavior.

docker/gcp_pudl_etl.sh
zaneselvans: The first time I tried running it with `nightly` pointing at a prior nightly build and it couldn't fast-forward but I don't understand why.  Then I moved `nightly` to `main` for testing.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -141,7 +142,9 @@ jobs:
 
       # Start the VM
       - name: Start the deploy-pudl-vm
-        run: gcloud compute instances start "$GCE_INSTANCE" --zone="$GCE_INSTANCE_ZONE"
+        run: |
+          gcloud compute instances start "$GCE_INSTANCE" --zone="$GCE_INSTANCE_ZONE"
+          gcloud compute ssh "$GCE_INSTANCE" --zone="$GCE_INSTANCE_ZONE" -- "docker image ls catalyst-coop/pudl-etl --digests"
```
jdangerx, line 147: Added this in an attempt to get the actual hash that's running on the VM.

.github/workflows/build-deploy-pudl.yml
zaneselvans: Ah okay great. I'll take the suggestion and switch this back to doing a shallow clone.

docker/gcp_pudl_etl.sh
zaneselvans: Accepted.

docker/gcp_pudl_etl.sh
jdangerx: Now that I've played with this more, I think it's because we're testing on this `debug-nightly-deploy` branch, and if `nightly` is pointing at a commit that's not in the `debug-nightly-deploy` history, the `--ff-only` will fail.

.github/workflows/build-deploy-pudl.yml
jdangerx: Didn't work due to gcloud authorization issues that I didn't want to work through. Reverted.

jdangerx:
Some notes on what I've found so far after re-running the GHA a bunch of times:

1. The VM fails boot integrity checks, but continues booting anyways - so that's a red herring.
2. We seem to receive a "VM stop" request almost immediately after booting - but then we ignore it until a second stop request comes from the `shutdown_vm` function itself
3. Sometimes logs show up a long time after they are initially written out - I was thinking about why a bunch of logs showed up in the 17:57 UTC run but not in the 17:44 UTC run, but then after refreshing one more time those logs showed up in the earlier run too.
4. The last time we actually logged out "Launching user container" and had all the attendant logs about attempting to download a container appears to have been at 17:33 UTC - the later runs afterwards, even ones that appeared to have executed the script, did not actually appear to try to pull a container image. ???
5. the images seem to make it to dockerhub with no problem.

Useful snippets to search for in the logs:
"image.revision" - gets the git hash that we associated with the docker image.
"Launching user container" - when google appears to log, well, doing that

jdangerx:
More things I've learned:

- it can take a long time for logs to surface in Logs Explorer, >30m
- the log timestamps can be jumbled up in a confusing way, making the logstream appear to start in the middle of some operation / combine the logstreams of multiple runs
- we *are* actually pulling the container every time...
- we are *not* successfully pulling the newest version of the container, as shown by the `image.revision` tag

I'm exploring what it would look like to kick off these jobs in Batch instead, since those logs seem to be more timely and also easier to separate from each other.

zaneselvans:
Wow, this seems kind of messy. I really thought I was crazy last night.

If this mess is the impetus we need to switch over to Batch and a more cohesive build system overall that doesn't seem bad. Probably better to pour learning and debugging time into that than continuing to hack on our homebrew solution.

jdangerx:
> Wow, this seems kind of messy. I really thought I was crazy last night.

Man, I thought I was crazy this afternoon 😅 

> If this mess is the impetus we need to switch over to Batch and a more cohesive build system overall that doesn't seem bad. Probably better to pour learning and debugging time into that than continuing to hack on our homebrew solution.

I'll write up a little plan presently and shop it around, see if that seems outré to anyone.

zaneselvans:
@jdangerx It seems kind of unlikely that the VM flakiness is related to the attempt to update the `nightly` branch -- it was just the process of debugging the new code that the flakiness was identified. So maybe it makes sense to remove the testing-only parts of this PR and merge it in, to see if we can get the `nightly` branch updating automatically.

zaneselvans:
Moved the remaining work from this PR into #3195

PR #3184, opened 2023-12-21
FERC to EIA Entity Matching Refactor

katie-lamb:
# Overview

I'm not sure how to pin the issue since it's in a different repo, but it closes https://github.com/catalyst-cooperative/ccai-entity-matching/issues/108 in the CCAI repo.

This PR incorporates the new vectorization infrastructure for entity matching problems into the FERC to EIA matching model.

**Major Changes**
- Removed the `recordlinkage` dependency that was previously being used to vectorize the model input dataframes and replaced that vectorization process with the infrastructure that newly went in with the FERC to FERC model update.
- Added cleaning steps to the columns to improve results.
- Had to rework the dataframe embedding graph and ops because I couldn't get it to cooperate with a factory function (see below inline comment)

**Still need to do:**
- Update the doc string in the new vectorization functions that I added to the `embed_dataframe` module to be more descriptive.

**Out of scope**
- Experiment tracking
- Integration of blocking with `faiss` clustering
- Splink model
- More advanced imputations of nulls in columns.

# Testing

Accuracy is now at .74 on the test set (previously .73). Consistency of matches is still .74. I'm going to compare the output matches from the model in `dev` and this model and see if there are any changes in what got matched.

```[tasklist]
# To-do list
```


src/pudl/analysis/plant_parts_eia.py
```diff
@@ -790,7 +790,9 @@ def add_additional_cols(self, plant_parts_eia, plants_eia860, utils_eia860):
                 ],
             )
             .merge(
-                utils_eia860[["utility_id_eia", "utility_id_pudl"]].drop_duplicates(),
+                utils_eia860[
+                    ["utility_id_eia", "utility_id_pudl", "utility_name_eia"]
```
katie-lamb, line 794: I added `utility_name_eia` as a column in the plant parts list so that it can be used for matching.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -146,6 +139,12 @@ def merge_steam_fuel_dfs(
     ).astype({"plant_type": str, "construction_type": str})
 
 
+@op
+def get_vectorizers():
+    """Get the dictionary of vectorizer transforms for each column."""
+    return dataframe_vectorizers
```
katie-lamb, line 145: Is there a better way to get the dictionary of vectorizers as an input to an `op` within the below `graph_asset` besides returning it from this type of shell op? Maybe use `context`?

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -32,6 +35,7 @@ class FeatureMatrix:
     """
 
     matrix: np.ndarray | scipy.sparse.csr_matrix
+    index: pd.Index
```
katie-lamb, line 38: I added this variable to keep track of the index of the matrix, because in the FERC to EIA context, it's useful to know what `record_id_eia` x `record_id_ferc1` pair each row in the matrix corresponds to.

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -69,37 +73,36 @@ def as_pipeline(self):
         )
 
 
-def dataframe_embedder_factory(vectorizers: dict[str, ColumnVectorizer]):
```
katie-lamb, line 72: I'm pretty confused why I was unable to use this `dataframe_embedder_factory`. Basically, when I created another instance of the factory in the FERC to EIA model, it created duplicate nodes with the same name as the ops in the FERC to FERC model. When I moved the two ops and graph outside of this factory function then I was able to call the graph twice, once in each of the two different models. Maybe it should somehow be made into a class? It would be nice to bundle the two ops and graph into one container.

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -69,37 +73,36 @@ def as_pipeline(self):
...
-        return apply_dataframe_embedder(df, transformer)
+@op
+def apply_dataframe_embedder(df: pd.DataFrame, transformer: ColumnTransformer):
+    """Use :class:`sklearn.compose.ColumnTransformer` to transform input."""
+    return FeatureMatrix(matrix=transformer.transform(df), index=df.index)
 
-    return embed_dataframe
+
+@graph
+def embed_dataframe_graph(df: pd.DataFrame, vectorizers) -> FeatureMatrix:
```
katie-lamb, line 102: Maybe `embed_dataframe_graph` isn't the best name for this graph. It was previously just `embed_dataframe` but now that you actually call this graph instead of just the `dataframe_embedder_factory`, I wanted to make sure that this graph did't ahve the same name as the module (`embed_dataframe`).

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -150,12 +153,13 @@ def as_transformer(self):
 
 
 def _apply_cleaning_func(df, function_key: str = None):
+    category_cols = df.select_dtypes(include="category").columns
+    df[category_cols] = df[category_cols].astype("str")
```
katie-lamb, line 157: I'm not sure why, but I was getting errors when cleaning the `category` type columns because converting nulls to the empty string adds another category value that's not set in `categories`. Maybe I should reset the categories themselves instead of converting all category columns to strings?

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -181,3 +185,111 @@ class NameCleaner(TransformStep):
...
+    function_key: str,
+    col1: str,
+    col2: str,
+    scale: float,
+    offset: float,
+    origin: float,
+    missing_value: float,
+    label: str,
+):
+    def _exp_sim(df, col1, col2, scale, offset, origin, missing_value, label):
```
katie-lamb, line 227: These functions were taken from the python `recordlinkage` package.

src/pudl/analysis/record_linkage/name_cleaner.py
```diff
@@ -266,4 +266,11 @@ def apply_name_cleaning(
         Returns:
             df (dataframe): the clean version of the input dataframe
         """
+        if isinstance(df, pd.DataFrame) and len(df.columns) > 1:
+            clean_df = pd.DataFrame()
+            for col in df.columns:
+                clean_df = pd.concat(
+                    [clean_df, df[col].apply(self.get_clean_data)], axis=1
+                )
+            return clean_df
```
katie-lamb, line 275: This allows the name cleaner to work on a dataframe with multiple columns, so if you pass in a dataframe with `plant_name_ferc1` and `plant_name_eia` then it will clean both.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
```diff
@@ -35,25 +35,277 @@
...
+
+
+@graph_multi_asset(
+    outs={
+        "out_pudl__yearly_assn_eia_ferc1_plant_parts": AssetOut(
+            io_manager_key="pudl_sqlite_io_manager"
+        )
+    }
 )
 def out_pudl__yearly_assn_eia_ferc1_plant_parts(
```
katie-lamb, line 310: Maybe some of the ops that are called in this graph asset should be combined so there are fewer tiny ops. I wasn't really sure what the best practice was here, so I made a bunch of little ops.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
```diff
@@ -499,25 +590,16 @@ def run_model(
...
-            left_index=True,
-            right_index=True,
-            indicator=True,
-        )["_merge"]
-        == "both",
-        1,
-        0,
+    X = features_train.matrix  # noqa: N806
+    X_train, X_test, y_train, y_test = train_test_split(  # noqa: N806
+        X, y_df, test_size=0.25, random_state=16
```
katie-lamb, line 595: I added a train and test split, since previously accuracy was just being measured on the train set. Accuracy on the test set is .74 and the results of grid search cross validation on the train set are also outputted (.94)

zaneselvans:
Hey @katie-lamb if we know that the integration tests can't pass, I'm going to switch this back to a draft so that every new push doesn't cost us a few dollars in GitHub runner time.

katie-lamb:
@zschira The CI is passing locally for me, so I'm not sure why it's not passing here. Investigating!

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
zschira: There's some overhead to each `op`, so it can improve performance to combine them, but it's probably not too big of a deal as long as it's not excessive.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
```diff
@@ -35,25 +35,277 @@
...
+def get_compiled_input_manager(plants_all_ferc1, fbp_ferc1, plant_parts_eia):
+    """Get :class:`InputManager` object with compiled inputs for model."""
+    inputs = InputManager(plants_all_ferc1, fbp_ferc1, plant_parts_eia)
+    # compile/cache inputs upfront. Hopefully we can catch any errors in inputs early.
+    inputs.execute()
+    return inputs
+
+
+@op
+def get_all_pairs_df(inputs):
```
zschira, line 208: It would be nice to be able to use the same op for `get_all_pairs_df` and `get_train_pairs_df` since they're doing the same thing. Maybe we should pass the inputs to an op that just does the `merge`?

zschira (approved):
Glad it seemed pretty straightforward to adapt to the new infrastructure! I don't have anything blocking, just some minor comments/suggestions.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
katie-lamb: I changed this so that `inputs` is passed into one op that returns both `all_pairs_df` and `train_pairs_df`.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
```diff
@@ -68,39 +345,44 @@ def out_pudl__yearly_assn_eia_ferc1_plant_parts(
...
+)
+def out_pudl__yearly_assn_eia_ferc1_plant_parts(
+    _out_pudl__yearly_assn_eia_ferc1_plant_parts: pd.DataFrame,
+) -> pd.DataFrame:
+    """Linkage between FERC1 plants and EIA plant parts for persistent storage.
+
+    Args:
+        out_eia__yearly_plant_parts: The EIA plant parts list.
+    """
+    return _out_pudl__yearly_assn_eia_ferc1_plant_parts
```
katie-lamb, line 385: @zschira I found out that you can't write an output to persistent storage from a `graph_asset` so I made this separate `asset` which takes the intermediary output from the `graph_asset` `_out_pudl__yearly_assn_eia_ferc1_plant_parts` and basically just outputs as an asset to the DB. Maybe there's a better way to do this besides having this wrapper `asset` which is almost identically named? It's nice to keep the graph asset though so you can run individual ops, and not have to run the whole linking process at once.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
katie-lamb: Changed so that the dictionary is contained within the op.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
zschira: I think the problem is that we're not setting the io manager in the graph asset. I remember messing with this awhile back for something, and I think I could only get it to work by using a `graph_multi_asset` because the API is a little weird. I can see if I can get it to work that way in the morning, and if not it's probably ok as is

katie-lamb: Hmm I think I tried setting it in the `graph_multi_asset` and didn't have success. I think `graph_multi_asset` doesn't take `io_manager_key` as a parameter, and so I tried to set it within the `outs` parameter:

`outs={"out_pudl__yearly_assn_eia_ferc1_plant_parts" : AssetOut(io_manager_key="pudl_sqlite_io_manager")` 

But maybe that's not the right way to go about it.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
```diff
@@ -35,27 +35,264 @@
...
+def run_matching_model(features_train, features_all, y_df):
+    """Run model to match EIA to FERC records."""
+    return run_model(
+        features_train=features_train,
+        features_all=features_all,
+        y_df=y_df,
+    )
+
+
+@op
```
zschira, line 280: @katie-lamb I see what you mean, I tried using the `graph_multi_asset` and specifying the io-manager through the `outs`, but it's still not actually writing to the DB for some reason. I did you find you can specify the io-manager for the final `op` in the graph, and that works. IDK if this is actually better than just using the second asset, but it's an option.

```suggestion
@op(out={"out_pudl__yearly_assn_eia_ferc1_plant_parts": Out(io_manager_key="pudl_sqlite_manager_key"})
```

I tested this locally and I was able to remove the wrapper asset and get this to write to the DB as expected. Either way is a bit janky, it would be nice if dagster just let us specify the asset manager directly in a `graph_asset`.

src/pudl/analysis/record_linkage/embed_dataframe.py
zschira: The plant and construction types are both being converted to strings in the ferc-ferc model before vectorizing, so I think this is pretty much equivalent to how that works

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
katie-lamb: Ya I think I like this better, just because then there's no confusion with what out_pudl__yearly... vs _out_pudl__yearly... is. Will make this change.

katie-lamb:
After comparing results of the current vs incoming matching model, there are 519 out of the 24430 matches (2% of matches) that change from the old to the new model. After spot checking these changed matches and looking at the average change in each of the "matching columns" between the new and old results, it doesn't seem like the 519 new matches are decidedly better or worse. The change likely comes from the name cleaning and the null filling that happens in this PR.

zaneselvans:
Merging this manually as an admin since it's passed CI 3 times and keeps getting scooped by other PRs.

PR #3188, opened 2023-12-22
Clean up nightly build/deploy w/o nightly branch update.

zaneselvans:
Takes the working elements of #3183 and gets them merged into `main` so the builds are working again.

- Fix syntax error in build-deploy-pudl workflow.
- Move envvar printing into separate step
- Rename GITHUB_REF to BUILD_REF
- set the nightly tag, so we have something to merge
- Allow write permissions for git tagging in nightly builds workflow
- Functionalize udpate_nightly_branch()
- Set origin for nightly when pushing.
- Temporarily disable nightly branch updates to fix nightly builds.

### DIFF
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -47,6 +47,7 @@ jobs:
           echo "BUILD_ID: $BUILD_ID"
 
       - name: Tag nightly build
+        # TESTING -- SHOULD ONLY RUN ON SCHEDULE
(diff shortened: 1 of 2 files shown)


jdangerx (approved):

zaneselvans:
The `ci-integration` runner got killed when attempting to cache an `epacems` archive. I've seen this happen occasionally elsewhere too, but it doesn't seem to be consistent/repeatable.  I went ahead and overrode it to merge, but I'm a little worried that with the new annual `epacems` archives we may see this problem more frequently. @e-belfer

PR #3189, opened 2023-12-22
update the 860m doi

### LINKED ISSUE #3186: EIA 860M: Retool extraction to handle listed partitions

cmgosnell:
# Overview

Closes #3186.

We had to make a new zenodo archive with eia860m a years worth of monthly files zipped together into one resource because *file upload limits*. So this pr 

What did you change?
basically nothing except the DOI. bc @e-belfer added a little logic into the datastore to work with partitions that are lists of partitions. And the excel extractor/the datastore combined  already knows how to grab a file out of a zipped file bc of course it does bc so many of our one partition resources have many files. The main place where this is happening is `load_excel_file`.

We could remove the first `try` in `load_excel_file` because the old eia860m archive being individual files was actually the edge case.

# Testing

I ran the fast etl locally. But first I thought I was going to have to muck with the excel extractor so I setup a little notebook testing situation and the simplest setup gave me the eia860m outputs:

```python
from pudl.extract.eia860m import Extractor
from pudl.workspace.datastore import Datastore

ds = Datastore(local_cache_path=pudl.workspace.setup.PudlPaths().pudl_input)
self = Extractor(ds=ds)
raw_eia860m_dfs = self.extract(year_month=pudl.metadata.sources.SOURCES["eia860m"]["working_partitions"]["year_month"])
```

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/workspace/datastore.py
+++ b/src/pudl/workspace/datastore.py
@@ -141,7 +141,10 @@ class DatapackageDescriptor:
             if name and res["name"] != name:
                 continue
             for k, v in res.get("parts", {}).items():
-                partitions[k].add(v)
+                if isinstance(v, list):
+                    partitions[k] |= set(v)  # Add all items from list
+                else:
+                    partitions[k].add(v)
         return partitions
 
     def get_partition_filters(self, **filters: Any) -> Iterator[dict[str, str]]:
--- a/src/pudl/workspace/datastore.py
+++ b/src/pudl/workspace/datastore.py
@@ -104,7 +104,6 @@ class DatapackageDescriptor:
                     f"Resource filter values should be all lowercase: {k}={v}"
                 )
         parts = res.get("parts", {})
-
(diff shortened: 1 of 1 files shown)


e-belfer (approved):
Looks good! I'll do the final validation in the CEMS branch.

zaneselvans:
The change to allow lists in the partitions has broken the docs build script, which reads those partitions to generate the dataset docs and.

e-belfer:
> The change to allow lists in the partitions has broken the docs build script, which reads those partitions to generate the dataset docs and.

Already on it!

zaneselvans (changes requested):
The documentation build script isn't expecting to find lists in the partitions, and so is failing when it attempts to build the data source specific docs pages using the Jinja templates.  It needs to be updated to accommodate the new metadata structure associated with the newly bundled raw archives.

PR #3193, opened 2023-12-27
Tell setuptools_scm to ignore non-version tags

zaneselvans:
# Overview

Update configuration for `setuptools_scm` such that it only considers tags matching `v20*` when trying to determine the current version.

Related to #3140

# Testing

After altering `pyproject.toml` I attempted to install the PUDL package using `pip`

```[tasklist]
# To-do list
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,13 +152,13 @@ where = ["src"]
 [tool.setuptools_scm]
 tag_regex = "^v(?P<version>20\\d{2}\\.\\d{1,2}\\.\\d{1,2})$"
 git_describe_command = [
-    'git',
-    'describe',
-    '--dirty',
-    '--tags',
-    '--long',
-    '--match',
-    'v20*',
+    "git",
+    "describe",
+    "--dirty",
+    "--tags",
+    "--long",
+    "--match",
+    "v20*",
 ]
 
 [tool.ruff]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,6 +150,16 @@ include-package-data = true
 where = ["src"]
 
 [tool.setuptools_scm]
+tag_regex = "^v(?P<version>20\\d{2}\\.\\d{1,2}\\.\\d{1,2})$"
+git_describe_command = [
+    'git',
+    'describe',
+    '--dirty',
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -150,6 +150,16 @@ include-package-data = true
 where = ["src"]
 
 [tool.setuptools_scm]
+tag_regex = "^v(?P<version>20\\d{2}\\.\\d{1,2}\\.\\d{1,2})$"
```
zaneselvans, line 153: This tells `setuptools_scm` how to extract the "version" string.
- Require 4 digit year
- Allows 1 or 2 digit month and day strings (since PyPI and some other tools seem to default to 1 digit rather than allowing a leading zero)

pyproject.toml
```diff
@@ -150,6 +150,16 @@ include-package-data = true
...
+git_describe_command = [
+    "git",
+    "describe",
+    "--dirty",
+    "--tags",
+    "--long",
+    "--match",
+    "v20*",
+]
```
zaneselvans, line 162: This changes the command which is used by `setuptools_scm` to gather information from `git` about the tags and current state of the repository. By using the `v20*` glob, it will only search for tags which match our current versioning scheme.

pyproject.toml
```diff
@@ -263,9 +273,7 @@ filterwarnings = [
     "once:The behavior of DataFrame concatenation with empty or all-NA entries is deprecated.:FutureWarning",
 ]
 
-markers = [
-    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
-]
+markers = ["slow: marks tests as slow (deselect with '-m \"not slow\"')"]
```
zaneselvans, line 276: Not sure why this is all in red. Just a change in line-wrapping from my autoformatter.

pyproject.toml
jdangerx: We can deal with the eventual failure in 77 years :)

pyproject.toml
jdangerx: Not sure either but it's very scary.

jdangerx (approved):
Aha! Nice work.

PR #3194, opened 2023-12-27
Clarify consistency check and lower threshold from 75% to 74%

zaneselvans:
# Overview

Addresses `nightly-2023-12-27` build failure by:
- lowering expected consistency between FERC-FERC and FERC-EIA entity matching from 75% to 74%.
- documents what's being checked in the assertion that was failing.
- also makes implicit weights of 1.0 explicit in the FERC 1 plant classification pipeline.

# Testing

Ran the full ETL locally.

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/analysis/eia_ferc1_record_linkage.py
+++ b/src/pudl/analysis/eia_ferc1_record_linkage.py
@@ -939,24 +939,31 @@ def check_match_consistency(
     train_df: pd.DataFrame,
     match_set: Literal["all", "overrides"] = "all",
 ) -> pd.DataFrame:
-    """Check how consistent matches are across time.
+    """Check how consistent FERC-EIA matches are with FERC-FERC matches.
+
+    We have two record linkage processes: one that links FERC plant records across time,
+    and another that links FERC plant records to EIA plant-parts. This function checks
+    that the two processes are as consistent with each other as we expect.  Here
+    "consistent" means that each FERC plant ID is associated with a single EIA plant
+    parts ID across time. The reverse is not necessarily required -- a single EIA plant
+    part ID may be associated with various FERC plant IDs across time.
 
     Args:
         connects_ferc1_eia: Matches of FERC1 to EIA.
         train_df: training data.
         match_set: either ``all`` - to check all of the matches - or ``overrides`` - to
-            check just the overrides. Default is'``all``. The overrides are less
+            check just the overrides. Default is ``all``. The overrides are less
             consistent than all of the data, so this argument changes the consistency
             threshold for this check.
     """
     # these are the default
-    consistency = 0.75
-    consistency_one_cap_ferc = 0.85
+    expected_consistency = 0.74
+    expected_uniform_capacity_consistency = 0.85
     mask = connects_ferc1_eia.record_id_eia.notnull()
 
     if match_set == "overrides":
-        consistency = 0.39
-        consistency_one_cap_ferc = 0.75
+        expected_consistency = 0.39
+        expected_uniform_capacity_consistency = 0.75
         train_ferc1 = train_df.reset_index()
         # these bbs were missing from connects_ferc1_eia. not totally sure why
         missing = [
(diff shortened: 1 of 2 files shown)


src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -50,24 +50,28 @@
                 embed_dataframe.ColumnCleaner(cleaning_function="null_to_empty_str"),
                 embed_dataframe.CategoricalVectorizer(),
             ],
+            weight=1.0,
```
zaneselvans, line 53: In debugging this issue I played around with these weights, and found having them explicit more legible.

I was surprised that even substantial changes didn't seem to affect the consistency check much at all.  Doubling the weight of `capacity_mw` and `utility_id_ferc1` only increased it from 74.8% to 74.9%.

src/pudl/analysis/eia_ferc1_record_linkage.py
```diff
@@ -939,24 +939,31 @@ def check_match_consistency(
...
 ) -> pd.DataFrame:
-    """Check how consistent matches are across time.
+    """Check how consistent FERC-EIA matches are with FERC-FERC matches.
+
+    We have two record linkage processes: one that links FERC plant records across time,
+    and another that links FERC plant records to EIA plant-parts. This function checks
+    that the two processes are as consistent with each other as we expect.  Here
+    "consistent" means that each FERC plant ID is associated with a single EIA plant
+    parts ID across time. The reverse is not necessarily required -- a single EIA plant
+    part ID may be associated with various FERC plant IDs across time.
```
zaneselvans, line 949: Looking at the code below I *think* this is right but a second set of 👀 would be helpful.

```py
    count = (
        connects_ferc1_eia[mask]
        .groupby(["plant_id_ferc1"])[["plant_part_id_eia", "capacity_mw_ferc1"]]
        .nunique()
    )
    actual_consistency = len(count[count.plant_part_id_eia == 1]) / len(count)
```

The only material change in here is that the required consistency is 74% rather than 75%.  Everything else is just docs and more explanatory variable names.

src/pudl/analysis/eia_ferc1_record_linkage.py
```diff
@@ -939,24 +939,31 @@ def check_match_consistency(
...
-            check just the overrides. Default is'``all``. The overrides are less
+            check just the overrides. Default is ``all``. The overrides are less
             consistent than all of the data, so this argument changes the consistency
             threshold for this check.
     """
     # these are the default
-    consistency = 0.75
-    consistency_one_cap_ferc = 0.85
+    expected_consistency = 0.74
+    expected_uniform_capacity_consistency = 0.85
```
zaneselvans, line 961: It is surprising to me given how many parts of the overall FERC-FERC matching process changed, that we would end up within a fraction of a percent of the old threshold.

I suspect that `capacity_mw` should be weighted more heavily in the FERC-FERC matching, since a big change (like, 10% or more) in the capacity *probably* means that which chunk of plant is being reported has changed.

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
katie-lamb: Ya that does seem weird. Maybe that's an indication that inconsistent matches are more attributable to mistakes in the FERC to EIA matching process?

katie-lamb (approved):
The updated docstring makes sense to me!

katie-lamb:
@zaneselvans I found results from a FERC to EIA run that I did on Friday 12/21, before the FERC to FERC got merged in:

```
2023-12-21 22:19:59 [    INFO] catalystcoop.pudl.analysis.eia_ferc1_record_linkage:998 Matches with consistency across years of all matches is 88.9%
2023-12-21 22:19:59 [    INFO] catalystcoop.pudl.analysis.eia_ferc1_record_linkage:1011 Matches with completely consistent FERC capacity have a consistency of 97.6%
2023-12-21 22:20:00 [    INFO] catalystcoop.pudl.analysis.eia_ferc1_record_linkage:998 Matches with consistency across years of overrides matches is 83.7%
2023-12-21 22:20:00 [    INFO] catalystcoop.pudl.analysis.eia_ferc1_record_linkage:1011 Matches with completely consistent FERC capacity have a consistency of 97.6%
```

So the match consistency definitely went down with the new FERC to FERC.  Also, did we check what the new consistency is on the "override" matches as well as "all" matches? I wasn't sure if that failure in the builds was for the "all" or "override" check. 

I still think we should merge this change to get the nightly builds working again.

zaneselvans:
Do you if your run with the higher consistency was for all years?

katie-lamb:
Yep, it was all years (2001-2023).

PR #3195, opened 2023-12-28
Update nightly branch after successful build

zaneselvans:
# Overview

- Update the `nightly` branch if the scheduled nightly build is successful.
- Use `[[ modern bash conditionals ]]` throughout the script.
- Track success of various different steps in separate environment variables.
- Use the `&&` to combine exit statuses of the commands inside functions so we can track the success/failure of the entire chain of commands.
- Send output from all of the functions to the logfile in the same way.

Part of #3140 

# Testing

Kinda need to run a real scheduled deployment... 

```[tasklist]
# To-do list
```


zaneselvans:
This *seems* to be working, but it's unable to actually do the FF merge to update the `nightly` tag for reasons that I don't understand.

I'm currently attempting to run the `build-deploy-pudl` action after:
- setting `nightly` to point at `dev`
- rebasing the `update-nightly-branch` branch off of `dev`

So `nightly`/`dev` is definitely in the history of this branch, and it should be possible to `merge --ff-only`.  HOWEVER, for some reason the commit that's associated with the `testing-2023-12-29` tag inside the container when the action runs is DIFFERENT from the commit that's associated with that tag on GitHub and in my local repo after the action runs:

In my local repo / on GitHub:
```
commit 898837b683f4fb312848683bd7e903c4608cedad (HEAD -> update-nightly-branch, tag: testing-2023-12-29, origin/update-nightly-branch)
Author: Zane Selvans <zane.selvans@catalyst.coop>
Date:   Thu Dec 28 09:17:02 2023 -0600

    Don't fetch NIGHTLY_TAG
```

But if I check out the hash that's associated with the `testing-2023-12-29` tag inside the container:
```
commit d47cacb9a6d75a544c4d737a0078e2bd6ff4ced7 (HEAD)
Author: Zane Selvans <zane.selvans@catalyst.coop>
Date:   Thu Dec 28 09:17:02 2023 -0600

    Don't fetch NIGHTLY_TAG
```

It seems like it is caching a version of the repo from before I did the rebase.

Could this be the stale container problem? Where it's not using the most recent version of the container to run the build?  But shouldn't it be updating the local repository inside the container when it runs?

```bash
function update_nightly_branch() {
    git config --unset http.https://github.com/.extraheader && \
    git config user.email "pudl@catalyst.coop" && \
    git config user.name "pudlbot" && \
    git remote set-url origin "https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git" && \
    git checkout "$NIGHTLY_TAG" && \
    git tag && \
    echo "Updating nightly branch to point at $NIGHTLY_TAG." && \
    git fetch origin nightly:nightly && \
    git checkout nightly && \
    git merge --ff-only "$NIGHTLY_TAG" && \
    git push -u origin
}
```

I guess maybe it's not. It's just checking out whatever `$NIGHTLY_TAG` is present in the repo in the container already and if the container is stale, then the commit associated with the tag will be stale.

zaneselvans:
I updated the nightly build script to delete the local (possibly stale) `$NIGHTLY_TAG` and re-fetch it from `origin` and now the hash for the container, local, and GitHub versions of the tag are all identical, and it is still unable to `merge --ff-only` so I don't understand what the problem is.

jdangerx:
Woohoo! There’s probably some rebase in there on your feature branch that
makes it so that the nighty branch isn’t actually pointing at an ancestor
of the nightly tag. I can take a look when I’m at a computer. Have the
flaky execution problems resolved themselves?

Zane Selvans ***@***.***>于2023年12月29日 周五上午11:03写道：

jdangerx:
Hmm but after reading your earlier comments…  this is still weird and I
will poke around when I get home.

Dazhong Xia ***@***.***>于2023年12月29日 周五下午1:42写道：

jdangerx:
@zaneselvans what's the logging output you see when it's trying to merge `--ff-only` and failing?

zaneselvans:
@jdangerx  Maybe this doesn’t make sense but I wonder if there’s any way that the test-docker-build which builds the image and doesn’t push it to DockerHub and does cache the image locally on GitHub is interfering with the GCE deployment somehow. Like if we kick off build-deploy-pudl before test-docker-build has finished then there’s a conflict?

```
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
```

.github/workflows/build-deploy-pudl.yml
```diff
@@ -33,29 +33,46 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ env.BUILD_REF }}
+          fetch-depth: 0
```
zaneselvans, line 36: I set this to 0 to ensure access to all the tags & branches so we can check whether there are any changes necessitating a build.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -33,29 +33,46 @@ jobs:
...
+
+      - name: Skip the build if no changes since the last successful nightly build.
+        run: |
+          CURRENT_COMMIT=$(git rev-parse HEAD)
+          NIGHTLY_COMMIT=$(git rev-parse origin/nightly)
+          if [[ "$CURRENT_COMMIT" == "$NIGHTLY_COMMIT" ]]; then
+            echo "::notice::No changes since last successful nightly build. Skipping."
+            echo "SKIP_BUILD=true" >> $GITHUB_ENV
+            exit 0
+          fi
```
zaneselvans, line 46: See if anything has actually changed since the last nightly build, and if not, skip doing another one.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -33,29 +33,46 @@ jobs:
...
 
       - name: Set action environment variables
+        if: ${{ env.SKIP_BUILD != 'true' }}
```
zaneselvans, line 49: Seems like there should be a more elegant way to skip everything once we've decided to bail but what is it?  Maybe split into 2 jobs, rather than using steps?

.github/workflows/build-deploy-pudl.yml
```diff
@@ -33,29 +33,46 @@ jobs:
...
-        if: ${{ (github.event_name == 'schedule') }}
+        if: ${{ (github.event_name == 'schedule') && (env.SKIP_BUILD != 'true') }}
         run: |
           git config user.email "pudl@catalyst.coop"
           git config user.name "pudlbot"
           git tag -a -m "$NIGHTLY_TAG" $NIGHTLY_TAG $BUILD_REF
           git push origin $NIGHTLY_TAG
+          git tag
+          git fetch --tags origin
+          git tag
```
zaneselvans, line 71: Maybe this isn't necessary now that I'm doing `fetch-depth: 0`?

docker/gcp_pudl_etl.sh
```diff
@@ -2,11 +2,6 @@
 # This script runs the entire ETL and validation tests in a docker container on a Google Compute Engine instance.
 # This script won't work locally because it needs adequate GCP permissions.
 
-# Set PUDL_GCS_OUTPUT *only* if it is currently unset
-: "${PUDL_GCS_OUTPUT:=gs://nightly-build-outputs.catalyst.coop/$BUILD_ID}"
```
zaneselvans, line 6: Moved this down to where other variables are initialized.

docker/gcp_pudl_etl.sh
```diff
@@ -2,11 +2,6 @@
...
-
-set -x
```
zaneselvans, line 8: Removed to avoid printing out the PAT.

docker/gcp_pudl_etl.sh
```diff
@@ -57,38 +52,38 @@ function shutdown_vm() {
...
+        echo "Copying outputs to AWS distribution bucket" && \
         aws s3 cp "$PUDL_OUTPUT/" "s3://pudl.catalyst.coop/$BUILD_REF" --recursive
     fi
 }
 
 function zenodo_data_release() {
-    echo "Creating a new PUDL data release on Zenodo."
-    ~/pudl/devtools/zenodo/zenodo_data_release.py --publish --env sandbox --source-dir "$PUDL_OUTPUT"
+    echo "Creating a new PUDL data release on Zenodo." && \
+    ~/pudl/devtools/zenodo/zenodo_data_release.py --publish --env "$1" --source-dir "$PUDL_OUTPUT"
```
zaneselvans, line 78: Added an argument here so we can specify which Zenodo server to release to.

docker/gcp_pudl_etl.sh
```diff
@@ -100,60 +95,95 @@ function notify_slack() {
...
+    git remote set-url origin "https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git" && \
+    echo "Updating nightly branch to point at $NIGHTLY_TAG." && \
+    git tag -d "$NIGHTLY_TAG" && \
+    git fetch --tags origin "$NIGHTLY_TAG" && \
+    git checkout "$NIGHTLY_TAG" && \
+    git tag && \
+    git fetch origin nightly:nightly && \
+    git checkout nightly && \
+    git merge --ff-only "$NIGHTLY_TAG^0" && \
+    git push -u origin nightly
```
zaneselvans, line 110: Is there anything else in here that we don't need now that it's working?  What does the `^0` do in the emrge?

docker/gcp_pudl_etl.sh
```diff
@@ -100,60 +95,95 @@ function notify_slack() {
...
+# MAIN SCRIPT
+########################################################################################
+# Initialize our success variables so they all definitely have a value to check
+ETL_SUCCESS=0
+SAVE_OUTPUTS_SUCCESS=0
+UPDATE_NIGHTLY_SUCCESS=0
+DATASETTE_SUCCESS=0
+CLEAN_UP_OUTPUTS_SUCCESS=0
+DISTRIBUTION_BUCKET_SUCCESS=0
+ZENODO_SUCCESS=0
```
zaneselvans, line 131: Using separate variables and checking that they are all 0 before declaring success at the end. Could probably do some better communication of why things failed if they fail.

docker/gcp_pudl_etl.sh
```diff
@@ -100,60 +95,95 @@ function notify_slack() {
...
+        update_nightly_branch 2>&1 | tee -a "$LOGFILE"
+        UPDATE_NIGHTLY_SUCCESS=${PIPESTATUS[0]}
     fi
 
-    # Compress the SQLite DBs for easier distribution
-    # Remove redundant multi-file EPA CEMS outputs prior to distribution
-    gzip --verbose "$PUDL_OUTPUT"/*.sqlite && \
-    rm -rf "$PUDL_OUTPUT/core_epacems__hourly_emissions/" && \
-    rm -f "$PUDL_OUTPUT/metadata.yml"
-    ETL_SUCCESS=${PIPESTATUS[0]}
```
zaneselvans, line 135: Moved this into its own function, and put it inside the conditional below which only happens if we're doing distribution.

zaneselvans (commented):
@jdangerx I think this is ready to merge? But a line-by-line look would be helpful to make sure all of the testing hacks have been removed.

.github/workflows/build-deploy-pudl.yml
jdangerx: Perfect! And it probably means we don't have to do as many gymnastics to get the git stuff working, either!

.github/workflows/build-deploy-pudl.yml
jdangerx: Sure seems like it, but I guess what you've done is the best solution for now: https://github.com/orgs/community/discussions/82744

.github/workflows/build-deploy-pudl.yml
jdangerx: I think the most useful debug output would be something like

```
...
git push origin $NIGHTLY_TAG
git fetch --tags origin $NIGHTLY_TAG
git show-ref -d $NIGHTLY_TAG # did the tag make it to origin? 
```

But I sort of feel like all this debug output unnecessary. I expect git push to work, you know? Only when we are questioning the very nature of reality would we need debug output here, and then we can just add it back in.

docker/gcp_pudl_etl.sh
jdangerx: Sweet, that's useful! Do we need to pass this in as a `--container-arg` in `build-deploy-pudl.yml`?

docker/gcp_pudl_etl.sh
jdangerx: Apparently, [if you merge a tag that *could* be fast-forwarded, `git` will still create a merge commit by default.](https://git-scm.com/docs/git-merge#_merging_tag)

The `^0` "unwraps" the tag, so that git doesn't create a merge commit. But, `--ff-only` should have the same effect, in that it forces a fast-forward merge. I don't think we need to double-dip here, and `--ff-only` is much more legible, so I think we can get rid of `^0`.

docker/gcp_pudl_etl.sh
```diff
@@ -100,60 +95,95 @@ function notify_slack() {
...
-    git config user.email "pudl@catalyst.coop"
-    git config user.name "pudlbot"
-    git remote set-url origin "https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git"
-    echo "BOGUS: Updating nightly branch to point at $NIGHTLY_TAG."
-    git fetch origin nightly:nightly
-    git checkout nightly
-    git merge --ff-only "$NIGHTLY_TAG"
-    ETL_SUCCESS=${PIPESTATUS[0]}
-    git push -u origin
+    git config --unset http.https://github.com/.extraheader && \
```
jdangerx, line 98:
```suggestion
    # When building the image, GHA adds an HTTP basic auth header in git
    # config, which overrides the auth we set below. So we unset it.
    git config --unset http.https://github.com/.extraheader && \
```

docker/gcp_pudl_etl.sh
jdangerx, line 110:
```suggestion
    git fetch --force --tags origin "$NIGHTLY_TAG" && \
    git fetch origin nightly:nightly && \
    git checkout nightly && \
    git show-ref -d nightly "$NIGHTLY_TAG" && \
    git merge --ff-only "$NIGHTLY_TAG" && \
    git push -u origin nightly
```
* use `--force` flag on git fetch to overwrite any local version of the $NIGHTLY_TAG, instead of manually deleting
* I think you don't have to check out the tag to be able to merge it later
* `show-ref -d` shows us "which commit hashes are we pointing at", which is both more useful & less verbose than `git tag`
* not sure why the diff isn't rendering nicely, but the merge/push and the nightly branch fetch/checkout all seem fine.

docker/gcp_pudl_etl.sh
jdangerx: WOW this is a great idea and I can't believe I hadn't thought of this. Will definitely remember that there's a whole slew of steps & separate states when we port to Python!

jdangerx: I suppose we could have different numbers for "not started, attempted, succeeded, failed". But we can do without that for now.

jdangerx (changes requested):
This mostly looks great! Added a couple suggestions around the `git` stuff & a question about the sandbox/prod argument, and then I think we're good to go!

.github/workflows/build-deploy-pudl.yml
zaneselvans: Ah okay, if that's as good as it gets, it does work!

docker/gcp_pudl_etl.sh
zaneselvans: It's just hard-coded in the main body of the script below right now since we don't yet want to try doing a production release, but I guess when we go to doing production releases, maybe we would like to try and do sandbox releases any time that we're not on a version tag? Just to keep exercising the code?  In which case I guess we could set the value to `sandbox` or `production` depending on which kind of build we're doing in the workflow.  Is that what you were thinking?

.github/workflows/build-deploy-pudl.yml
zaneselvans: Okay I'll remove it for now and cross my fingers...

docker/gcp_pudl_etl.sh
jdangerx: Yeah, I think eventually setting sandbox/production on every build would be smart. But right now just always pushing to sandbox on dev seems fine. Another improvement for python-script land, I think.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -36,6 +36,7 @@ jobs:
           fetch-depth: 0
 
       - name: Skip the build if no changes since the last successful nightly build.
+        if: ${{ (github.event_name == 'schedule') }}
```
jdangerx, line 39: Ha. yes, good catch.

jdangerx (approved):
Sweet, I hope it all still works 😬

.github/workflows/build-deploy-pudl.yml
zaneselvans: Yeah, whoops!

zaneselvans:
I will be amazed if it all still works on the first try. 😆 

I went ahead and merged it all the way into `main` so that the nightly build will try and use the new workflow, and so (hopefully) the `--ff-only` merge will work.

PR #3200, opened 2023-12-29
cleanup the excel extractor to no longer look for a single file

cmgosnell:
# Overview
In #3189 we started to use a eia860m archive with zipped files instead of stand xlsx files. which means that we can remove some cruft that handled this in the excel extractor.

What did you change?
I mostly just removed the option for an archive to go grab an individual file. now it always looks for a zipped resource

# Testing

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/workspace/datastore.py
+++ b/src/pudl/workspace/datastore.py
@@ -407,6 +407,7 @@ class Datastore:
 
     def get_zipfile_resource(self, dataset: str, **filters: Any) -> zipfile.ZipFile:
         """Retrieves unique resource and opens it as a ZipFile."""
+        # TODO: add some check if the result of get_unique_resource isn't a zip file.
         return zipfile.ZipFile(io.BytesIO(self.get_unique_resource(dataset, **filters)))
 
     def get_zipfile_resources(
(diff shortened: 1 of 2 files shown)


src/pudl/workspace/datastore.py
```diff
@@ -407,6 +407,7 @@ def get_unique_resource(self, dataset: str, **filters: Any) -> bytes:
 
     def get_zipfile_resource(self, dataset: str, **filters: Any) -> zipfile.ZipFile:
         """Retrieves unique resource and opens it as a ZipFile."""
+        # TODO: add some check if the result of get_unique_resource isn't a zip file.
         return zipfile.ZipFile(io.BytesIO(self.get_unique_resource(dataset, **filters)))
```
cmgosnell, line 411: hey @rousik ! since the excel extractor no longer needs to deal with a single file in the case of eia860m and can always expect a zipped resource, I stripped out that possibility from the excel extractor. But i thought it might be good to have a very clear message to let you know if you somehow are pointing to a non-zipped resource. Do you know of an easy way to check for that? also, this may be a moot question bc maybe the error message that would result from this would be clear enough.

cmgosnell:
closing bc maybe we want this functionality in the future.

PR #3205, opened 2024-01-03
Only run release workflow on v20xx tags.

jdangerx:
While playing with actions, I noticed that we're running the release workflow on *every* push which seems excessive... any reason to not filter at the trigger level?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,10 @@
 ---
 name: release
 
-on: push
+on:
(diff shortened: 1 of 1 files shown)


zaneselvans (commented):
I was doing this to test that the python distribution can still be built. All the pypi and GitHub package release steps should be short circuited unless we're on a version tag.

jdangerx:
Coolio. I'll close this then.

PR #3206, opened 2024-01-03
Use digest instead of tag so VM uses right image.

jdangerx:
# Overview

Follow-up to #3195 ; Related to #3140 .

We were getting inconsistent behavior pulling the latest image for a certain tag off of Dockerhub - often we would get an out-of-date image running on the VM.

While investigating Google Batch, I found that this was due to Dockerhub not updating its tag references quickly enough. However, we can just refer to the actual image digest we just built, which is unambiguously the correct image.

Given how small this change is, I decided to pull this logic over to the non-Batch world as well, so we don't have to wait for me to get all the Batch stuff working.

# Testing
I short-circuited `gcp_pudl_etl.sh` and had it echo a short message to the logs.

Then:
1. change the line
2. push
3. run `build-deploy-pudl.yml`
4. observe that the output includes the changed line!

Then, I went back to how it works on `main`/`dev` now - using the branch name as the image tag, instead of the image digest. Repeating the flow above updated the changed line on the first iteration, but subsequent changes were not picked up.

So I'm convinced that this should alleviate our "VM doesn't pick up the latest image version" woes.

```[tasklist]
# To-do list
- [x] Manual testing detailed above
```

### DIFF
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -92,6 +92,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       - name: Build image and push to Docker Hub
+        id: docker-build
         if: ${{ env.SKIP_BUILD != 'true' }}
         uses: docker/build-push-action@v5
         with:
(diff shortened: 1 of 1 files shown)


jdangerx (commented):
When this gets merged we should probably merge it all the way down to `main`.

bendnorman (approved):
Great! Using the digest seems like a better way to ensure we're pulling the correct image. 

Seems weird that there is a delay between the image getting uploaded and the digest updating. I'm assuming there are plenty of systems that pull images as soon as they are uploaded.

zaneselvans (approved):

zaneselvans:
I think @jdangerx has gone AFK for the night and it would be great to get this integrated with the other nightly build stuff that was merged today so I'm going to push the button.

PR #3211, opened 2024-01-03
Use Google Batch for full ETL runs

### LINKED ISSUE #3208: Port nightly build process to Google Batch
This lets us run multiple tag builds at once, and provides an environment where the dang logs show up on time.

Let's try to leave the rest of the system as untouched as possible.

```[tasklist]
## Success criteria
- [ ] ETL runs on Google Batch
- [ ] Slack notifications on success/failure include links to Batch job
```

jdangerx:
# Overview

Closes #3208 .

This lets us kick off nightly builds via Google Batch.

Opens up:
- prompter/searchable/filterable logging
- scaling up/down the instances easily
- kicking off multiple builds at once without stepping on each other's toes
- trying to use spot instances - we'd probably want to set this job up in a more "restartable" way, so that we don't end up having to re-run the whole dang thing when our VM gets interrupted

What did you change?
- add script to create batch configuration, based on the same CLI args we were using before for VM configuration
- run Postgres within the container, so we can stop running the Cloud SQL instance
- slightly nicer Slack message outputs
- exit 1 on failure so that the batch job dashboard correctly identifies a failure

# Testing

Ran lots of nightly builds using Batch :) You can kick off your own via `workflow_dispatch`.

```[tasklist]
# To-do list
- [ ] remove the Cloud SQL instances + Compute Engine instances once this gets merged in
```


devtools/generate_batch_config.py
```diff
@@ -51,14 +53,11 @@ def to_config(
...
-                        {
-                            "deviceName": "local-ssd",
-                            "mountPath": "/home/mambauser/pudl_work/output",
-                            "mountOptions": "rw,async",
-                        }
-                    ],
+                    "computeResource": {
+                        "cpuMilli": 8000,
+                        "memoryMib": 32768,
+                        "bootDiskMib": 204800,
```
zaneselvans, line 59: holycrap how do we end up needing 200GB?

jdangerx: This was just a guess since we ran out of a small boot disk of 10GB. I'll shrink it down to 80GB to match the existing VM.

zaneselvans: I was just surprised. Disk is cheap! We do have a lot of data.

docker/gcp_pudl_etl.sh
```diff
@@ -133,9 +134,11 @@ function notify_slack() {
     message+="DISTRIBUTION_BUCKET_SUCCESS: $DISTRIBUTION_BUCKET_SUCCESS\n"
     message+="ZENODO_SUCCESS: $ZENODO_SUCCESS\n\n"
 
-    message+="See https://console.cloud.google.com/storage/browser/builds.catalyst.coop/$BUILD_ID for logs and outputs."
+    message+="Logs on <https://console.cloud.google.com/batch/jobsDetail/regions/us-west1/jobs/run-etl-$BUILD_ID/logs?project=catalyst-cooperative-pudl|Google Batch Console>.\n\n"
+    message+="See <https://console.cloud.google.com/storage/browser/builds.catalyst.coop/$BUILD_ID|GCS> for outputs."
 
     send_slack_msg "$message"
+    upload_file_to_slack "$LOGFILE" "pudl_etl logs for $BUILD_ID:"
```
jdangerx, line 141: Do we want to keep this here? It's kind of nice to be able to click and download logs right from the failure notification. But we can already just click that Google Batch Console link...

devtools/generate_batch_config.py
```diff
@@ -0,0 +1,102 @@
...
+    """Munge arguments into a configuration dictionary."""
+    complete_env = sorted(_flat(container_env))
+    env_dict = OrderedDict(
+        (name, value.strip('"'))
+        for name, value in (pair.split("=", maxsplit=1) for pair in complete_env)
+    )
+
+    # NOTE (daz): the best documentation of the actual data structure I've found is at
+    # https://cloud.google.com/python/docs/reference/batch/latest/google.cloud.batch_v1.types.Job
+    config = {
```
jdangerx, line 43: I briefly thought about doing some dataclass thing here instead of a dictionary, but that seemed *super* verbose so I abandoned that idea.

zaneselvans (changes requested):
Looks great to the extent that I understand what it's doing :)

I have some desires around what shows up in the Slack notifications so we can easily tell which build a message is coming from and directly access the outputs.

docker/gcp_pudl_etl.sh
zaneselvans: I guess what I'm saying with the links to logs & outputs is different people will prefer different access methods so maybe we should just provide them all.

devtools/generate_batch_config.py
```diff
@@ -0,0 +1,102 @@
...
+                            "container": {
+                                "imageUri": container_image,
+                                "commands": [container_command] + _flat(container_arg),
+                            },
+                            "environment": {"variables": env_dict},
+                        }
+                    ],
+                    "computeResource": {
+                        "cpuMilli": 8000,
+                        "memoryMib": 32 * 1024,
```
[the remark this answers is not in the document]
jdangerx: Seems like it, the 32GB guy has been running fine.

jdangerx:
I think I know how to get the Slack notifications working even better but I figured that was out of scope for this PR. Just decided to update the wording & make sure that you can (a) click the links and go straight to the correct console page (b) copy the links directly into `gsutil` or `gcloud storage` commands.

zaneselvans (approved):

zaneselvans:
I note that since we we distribute the log file associated with the nightly builds, it's also possible to provide a clickable HTTP download link, though that would only work if the build has gotten as far as distributing the outputs.  In which case I guess we'll usually not need to look at the logs.

https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/2024-01-27-0602-52a2741ff-main-pudl-etl.log

Another nice bonus of giving the logfile a readable and unique name is we can now at a glance see what build the nightlies came from.

devtools/generate_batch_config.py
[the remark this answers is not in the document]
zaneselvans: Well that's wonderful.

PR #3212, opened 2024-01-04
It's german. It means "the dev, the"

zaneselvans:
# Overview

https://www.youtube.com/watch?v=gaXigSu72A4

Covers most of the tasks in #3179 which is the last bit of #3140

- Update docs to refer to main, nightly, and stable branches rather than dev in explaining our git workflow and nightly build / data distribution setup.
- Change all direct links to Read The Docs to use the newly default `nightly` docs rather than `dev`.
- Update nightly build scripts and other github actions workflows to run off of main rather than dev.
- Update dependabot and pre-commit-ci to run against main instead of dev.
- Switch to using `gs://builds.catalyst.coop` for our build outputs.
- Remove obsolete `docker/.env` file and define those envvars either in the container / workflow / script.

# Testing

- Manually ran a build off of this branch.

```[tasklist]
# To-do list
```


.github/workflows/build-deploy-pudl.yml
```diff
@@ -9,7 +9,6 @@ on:
 
 env:
   GCP_BILLING_PROJECT: ${{ secrets.GCP_BILLING_PROJECT }}
-  BUILD_REF: ${{ github.ref_name }} # This is changed to dev if running on a schedule
```
zaneselvans, line 12: I think we can pretty much get rid of the `BUILD_REF` variable now, since we don't have to do the "check out one branch when the workflow is running on another" thing.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -22,17 +21,15 @@ jobs:
       contents: write
       id-token: write
     steps:
-      - name: Use pudl-deployment-dev vm and dev branch if running on a schedule
+      - name: Use pudl-deployment-dev vm if running on a schedule
         if: ${{ (github.event_name == 'schedule') }}
         run: |
           echo "This action was triggered by a schedule."
           echo "GCE_INSTANCE=pudl-deployment-dev" >> $GITHUB_ENV
```
zaneselvans, line 28: Hopefully we'll be on Batch shortly and the archaic naming of this VM won't matter.

README.rst
```diff
@@ -68,7 +68,7 @@ PUDL is comprised of three core components:
     to the original inputs. Each of the data inputs may have several different versions
     archived, and all are assigned a unique DOI and made available through the REST API.
     You can read more about the Raw Data Archives in the
-    `docs <https://catalystcoop-pudl.readthedocs.io/en/dev/intro.html#raw-data-archives>`__.
+    `docs <https://catalystcoop-pudl.readthedocs.io/en/nightly/#raw-data-archivess>`__.
```
zaneselvans, line 71: These links to `intro.hml` were broken.  And also I updated all the direct docs links to use the newly default `nightly` docs.

docker/.env
```diff
@@ -1,11 +1,2 @@
...
-CONTAINER_HOME=/home/mambauser
-PUDL_INPUT=/home/mambauser/pudl_work/input
-PUDL_OUTPUT=/home/mambauser/pudl_work/output
-DAGSTER_HOME=/home/mambauser/pudl_work/dagster_home
-CONDA_PREFIX=/home/mambauser/env
-PUDL_SETTINGS_YML=/home/mambauser/src/pudl/package_data/settings/etl_full.yml
+PUDL_SETTINGS_YML=/home/mambauser/pudl/src/pudl/package_data/settings/etl_full.yml
 LOGFILE=/home/mambauser/pudl_work/output/pudl-etl.log
-CONDA_RUN="micromamba run --prefix /home/mambauser/env --attach ''"
-GCS_CACHE=gs://zenodo-cache.catalyst.coop
```
zaneselvans, line 11: I removed all the variables which were also being defined inside the Dockerfile to avoid ambiguity. This only left 2 lines. Do we think this file should still exist?  When would we expect the Dockerfile to be falling back on these values? Why shouldn't we be defining them either inside the container, or in the GitHub Actions workflow file?

docs/index.rst
```diff
@@ -34,15 +34,15 @@ We also publish SQLite databases containing relatively pristine versions of our
...
-* `FERC Form 714 (XBRL) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/dev/ferc714_xbrl.sqlite.gz>`__
+* `FERC Form 1 (DBF) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc1_dbf.sqlite.gz>`__
+* `FERC Form 1 (XBRL) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc1_xbrl.sqlite.gz>`__
+* `FERC Form 2 (DBF) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc2_dbf.sqlite.gz>`__
+* `FERC Form 2 (XBRL) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc2_xbrl.sqlite.gz>`__
+* `FERC Form 6 (DBF) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc6_dbf.sqlite.gz>`__
+* `FERC Form 6 (XBRL) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc6_xbrl.sqlite.gz>`__
+* `FERC Form 60 (DBF) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc60_dbf.sqlite.gz>`__
+* `FERC Form 60 (XBRL) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc60_xbrl.sqlite.gz>`__
+* `FERC Form 714 (XBRL) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc714_xbrl.sqlite.gz>`__
```
zaneselvans, line 45: I copied the current `dev` into `nightly` and `v2023.12.01` into `stable` so these URLs at least work. After the PR gets merged the nightly outputs should start getting updated nightly.

docker/gcp_pudl_etl.sh
```diff
@@ -62,14 +62,22 @@ function copy_outputs_to_distribution_bucket() {
...
-        echo "Removing old $BUILD_REF outputs from GCP distributon bucket." && \
-        gsutil -m -u "$GCP_BILLING_PROJECT" rm -r "gs://pudl.catalyst.coop/$BUILD_REF" && \
+        if [[ "$GITHUB_ACTION_TRIGGER" == "schedule" ]]; then
+            # If running nightly builds, copy outputs to the "nightly" directory
+            # TODO: if doing a tagged release, copy outputs to "stable" as well.
+            DIST_PATH="nightly"
+        else
+            # Otherwise we want to copy them to a directory named after the tag/ref
+            DIST_PATH="$BUILD_REF"
+        fi
```
zaneselvans, line 72: This feels janky. And where/when should we be doing the update to the `stable` outputs?

docs/data_access.rst
```diff
@@ -97,63 +97,66 @@ Nightly Builds
...
+  * `FERC-714 Datapackage (JSON) describing SQLite derived from XBRL <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc714_xbrl_datapackage.json>`__
+  * `FERC-714 XBRL Taxonomy Metadata as JSON (2021-2022) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc714_xbrl_taxonomy_metadata.json>`__
 
 .. note::
 
    To reduce network transfer times, we ``gzip`` the SQLite database files, which can
-   be quite large when uncompressed. To decompress them locally, you can use the
-   ``gunzip`` command.
+   be quite large when uncompressed. To decompress them locally, at the command line
+   on Linux, MacOS, or Windows you can use the ``gunzip`` command.
```
zaneselvans, line 151: It turns out Git for Windows installs `gzip`!  But not helpful if you aren't using the command line.

docs/dev/nightly_data_builds.rst
```diff
@@ -184,13 +180,7 @@ are configured to run the ``docker/gcp_pudl_etl.sh`` script. This script:
 5. Notifies the ``pudl-deployments`` Slack channel with the final build status.
 
 The ``gcp_pudl_etl.sh script`` is only intended to run on a GCE VM with adequate
-permissions. The full ETL and tests can be run locally by running these commands
-from the ``pudl`` directory:
-
-.. code-block::
-
-    docker compose -f docker/docker-compose.yml build
-    docker compose -f docker/docker-compose.yml up
```
zaneselvans, line 193: We removed the local docker compose setup a while ago but forgot to cut this.

docs/dev/nightly_data_builds.rst
```diff
@@ -292,37 +284,58 @@ that are available:
 
 .. code::
 
-   gsutil ls gs://nightly-build-outputs.catalyst.coop
+   gsutil ls -lh gs://nightly-build-outputs.catalyst.coop
 
-You should see a list of directories with the naming convention
-``<git commit SHA>-<git branch>``.
+You should see a list of directories with build IDs that have a naming convention:
+``<YYYY-MM-DD-HHMM>-<short git commit SHA>-<git branch>``.
```
zaneselvans, line 290: Updated the Build ID format

docs/dev/nightly_data_builds.rst
```diff
@@ -292,37 +284,58 @@ that are available:
...
 .. code::
 
-   gsutil ls -l gs://nightly-build-outputs.catalyst.coop/<build name of interest>
-
-      1152800  2022-11-22T12:51:02Z  gs://nightly-build-outputs.catalyst.coop/<build name of interest>/pudl-etl.log
-                                 gs://nightly-build-outputs.catalyst.coop/<build name of interest>/parquet/
-                                 gs://nightly-build-outputs.catalyst.coop/<build name of interest>/pudl_out/
-                                 gs://nightly-build-outputs.catalyst.coop/<build name of interest>/sqlite/
-   TOTAL: 1 objects, 1152800 bytes (1.1 MiB)
+    gsutil ls -lh gs://nightly-build-outputs.catalyst.coop/2024-01-03-0605-e9a91be-dev/
```
zaneselvans, line 296: I'd love if we could shorten this URL. I can never remember what it's supposed to be exactly.  Maybe just `gs://builds.catalyst.coop`?

docs/dev/project_management.rst
zaneselvans, line 1: Is there additional explanation needed in here with respect to the git workflow we'll be adopting?

Other than the fact that we branch off of and merge into `main` what do we need to be aware of? Do we need to do anything special to ensure that `nightly` can always be fast-forwarded to `main` in the nightly builds?

.github/workflows/build-deploy-pudl.yml
jdangerx: Yes! I think so too.

.github/workflows/build-deploy-pudl.yml
jdangerx: I also think, at some point, we just trigger the whole `build-deploy` only on tag push and workflow dispatch, and just make the nightly tag on a schedule. Which will mean we don't have to do all these conditionals on the event name.

.github/workflows/update-conda-lockfile.yml
```diff
@@ -21,21 +21,9 @@ jobs:
       - name: Get today's date
         run: |
           echo "TODAY=$(date +%Y-%m-%d)" >> $GITHUB_ENV
-      - name: Set GITHUB_REF for use with workflow_dispatch
```
jdangerx, line 24: Getting rid of state we have to track, you love to see it.

README.rst
jdangerx: No, looks like a broken link. Well - it just sends you to the main page without popping you to the anchor.

jdangerx, line 71:
```suggestion
    `docs <https://catalystcoop-pudl.readthedocs.io/en/nightly/#raw-data-archives>`__.
```

zaneselvans: Yesss, my preciousssss.

docker/.env
jdangerx: I see no reason for this file to exist anymore, I think it was probably being used in some docker compose situation but we certainly don't seem to use that. We should define the log file in the Dockerfile and the settings location in GHA since it's much more likely we want to change the settings file than the logfile location.

docker/gcp_pudl_etl.sh
jdangerx: Yeah, it does seem janky, since we're trying to decide where to send the data products based on "is this on a schedule, or not?"

Long-term, we should either pass in "this is nightly/stable/ad-hoc," or "update *this* location in GCS, *this* location in AWS, ..." - flags that actually correspond to our concepts. I'm sort of partial to "nightly/stable/ad-hoc" because that lets us hide the implementations of those within the build script.

But, short-term, this works - though I think we need to add another case to update `stable` outputs when the build ref is `v20xx...`

.github/workflows/update-conda-lockfile.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 # What branch does this action run on?
 # - workflow_dispatch: Whatever branch it was run against.
-# - schedule: Always the same branch (will be dev or main)
+# - schedule: Always runs on main
```
bendnorman, line 11: Oh yay I'm glad we won't have push workflow files into `dev` and again into `main` to get the changes.

docker/vm_startup_script.sh
```diff
@@ -1,3 +0,0 @@
-#! /bin/bash
-# Delete old catalystcoop/pudl-etl images
-docker rmi -f $(docker images -q catalystcoop/pudl-etl)
```
jdangerx, line 3: Is this no longer necessary? Or do we still need it while we're not on Batch?

README.rst
```diff
@@ -77,13 +77,13 @@ PUDL is comprised of three core components:
     Python package is embedded with a set of of DOIs to indicate which version of the
     raw inputs it is meant to process. This process helps ensure that the ETL and it's
     outputs are replicable. You can read more about the ETL in the
-    `docs <https://catalystcoop-pudl.readthedocs.io/en/dev/intro.html#the-etl-process>`__.
+    `docs <https://catalystcoop-pudl.readthedocs.io/en/nightly/#the-etl-process>`__.
```
bendnorman, line 80: Do we eventually want our docs, and datasette to by pointed at `stable`? I feel like most projects default to the latest stable version with the option to check out bleeding edge docs.

docs/data_access.rst
jdangerx: 😅 😅 😅

docs/dev/nightly_data_builds.rst
jdangerx: I'm down, can we just... do it?

docker/.env
bendnorman: I think the `.env` file was only useful when we were debugging this image locally using `docker-compose`. We should probably add the last two variables to the dockerfile.

docker/vm_startup_script.sh
bendnorman, line 3: Why did you remove this? I added this so that different image tags don't build up on the VM.

docs/dev/project_management.rst
jdangerx: Since `nightly` and `stable` both point at an ancestor of `main` right now, we should be in good shape so long as:

- *people* don't push to `nightly`/`stable` - fine for our nightly build bot to do it bc it only fast-forwards. And I guess in emergency situations we should allow ourselves to `git reset nightly <the-right-hash>`. But in normal operations this should be disallowed.
- nobody rewrites history on `main` - i.e., all changes to main happen through a PR (you can still squash/rebase the PR, and probably should, but once the commits hit `main` they are there for good)

docs/dev/nightly_data_builds.rst
bendnorman: I'm down to shorten it. Maybe `gss://pudl-builds.catalyst.coop`? Adding `pudl` might be redundant because everything in the `catalyst-cooperative-pudl` project should be related to `pudl`

jdangerx (changes requested):
This looks great!

A few loose ends here and there, which I guess are technically not blocking the deprecation of `dev`. But, since we still also need to communicate to people about the new process, I think we're not in *such* a hurry. So maybe it's worth just fixing them.

In terms of communication, I think what we need to communicate is:

"New process, same as the old process, except now you branch off of and merge into `main`. Which means you don't have to manually change the base branch each time you make a PR, yay! Also don't mess with `nightly`/`stable` and still don't push things willy-nilly to main."

If that sounds good to you, feel free to copy-paste that right into Slack, hah.

docs/dev/project_management.rst
```diff
@@ -19,23 +19,29 @@ them easily.
...
-* At the end of each sprint, or any time a particularly significant feature has
-  been merged in and the nightly builds are passing, **dev** is merged into **main**.
+* We have 3 persistent branches: ``main`` (the default branch), ``nightly``, and
+  ``stable``.
+* We create temporary feature branches off of ``main`` and make pull requests into
+  ``main`` throughout our 2 week long sprints. All code that's merged into ``main``
+  should have passed our CI tests and been reviewed by at least one other person.
+* Every night the ``main`` branch is used to run the :ref:`nightly-data-builds`. If the
+  builds are successful, then the ``nightly`` branch is automatically updated to point
+  to the latest commit on ``main``. If the builds fail, then the ``nightly`` branch is
```
bendnorman, line 29: Will `nightly` have branch protections so only a VM can update it?

jdangerx: Just added these for `nightly` and `stable`: 
image

README.rst
jdangerx: Yeah, `stable` seems less surprising, but also `nightly` seems fine to me. We could maybe move it all to `latest` and then configure `latest` to point at `stable` via RTD - then if we want to switch things around we can.

docs/dev/nightly_data_builds.rst
e-belfer: can I advocate for `gss://nightly.catalyst.coop`? I think that makes it clearer that it's the nightly updated data, personally.

.github/workflows/build-deploy-pudl.yml
zaneselvans: I've been thinking it might be nice to completely separate the build and deploy processes.

Like the build runs the ETL and tests / validations and fills up a directory in `gs://builds.catalyst.coop` as it does now, and then a totally separate script could be used to update the `nightly` branch, and deploy the outputs from a selected `BUILD_ID` to any combination of:

* Datasette on Fly.io
* AWS Open Data Registry S3 buckets
* Our GCS buckets that mirror AWS
* ...

README.rst
zaneselvans: Fixed it locally will push in a sec.

README.rst
zaneselvans: I think once we have successfully moved to an almost automated monthly release schedule, switching to stable seems fine.  But we have intended to do something like that many times, and I don't think having our default documentation be months out of date with what many people are using is a great idea.

docker/.env
zaneselvans: Okay, I'll add them as "set defaults but only if they aren't already set by the script that called the container" variables inside the container.

docs/dev/nightly_data_builds.rst
zaneselvans: But it's not just the nightly builds -- it's also the stable builds and the ad-hoc builds. Any data build will deposit its outputs in here.

docker/gcp_pudl_etl.sh
zaneselvans: Okay, I think I've got it set up to udpate the `stable` branch on a versioned tag push.

docker/vm_startup_script.sh
zaneselvans: Oh, I removed it because I didn't see it getting run anywhere and figured it was a stale leftover from when we had a local setup.  Does this script have a special magical meaning somewhere else?  If so I can add it back with a comment in the file that explains what it does.

zaneselvans: I just didn't see the script being called anywhere and saw no documentation as to its being needed so I figured it was another stale leftover script from our local docker-compose setup.

bendnorman: Yes! It is magical. It gets run when the VM starts. It feels a little wacky but when I was setting this up originally the disk was filling up with stale images.

@jdangerx any chance this could have played a part in the image not getting updated properly?

README.rst
bendnorman: I agree. I think most people will gravitate towards `nightly` because it seems like the prefer freshness over stability.

I'm curious when people will point to `stable`. I feel like folks will either want the bleeding edge or a pinned version.

docker/vm_startup_script.sh
jdangerx: I don't think so, since this wasn't removed in the branches we were working on. I think it's just Dockerhub <> GCP slowness.

jdangerx (approved):

docs/dev/nightly_data_builds.rst
zaneselvans: Okay, I've updated this all to point at `gs://builds.catalyst.coop` and have copied over all of our existing build outputs to a newly created bucket, so hopefully it works.

docker/vm_startup_script.sh
zaneselvans: Okay! I'll restore it and add a note in the comments inside the file so it's clear it's necessary even though it's not called anywhere in our codebase.

zaneselvans: Added this script back.

bendnorman (approved):

zaneselvans:
Hmm @bendnorman @jdangerx it looks like Google was really hoping for that `.env` file. Or is this a setting that's associated with the VM somewhere?

```
Updated [https://www.googleapis.com/compute/v1/projects/***/zones/***/instances/pudl-deployment-tag].
ERROR: (gcloud.compute.instances.update-container) Unable to read file [./docker/.env]: [Errno 2] No such file or directory: './docker/.env'
Error: Process completed with exit code 1.
```

Issue #3203, opened 2024-01-02
Make a utility rate base table using exploded assets


cmgosnell:
```[tasklist]
### Tasks
- [x] pull in both balance sheet explosions & filter by in_rate_base
- [x] ensure all corrections have tags
- [x] add the one additional element from a non-exploded table: working capital xbrl_factoid * (1/8?)
```

```[tasklist]
### Tasks
- [x] make unit tests build a forest
- [x] write unit tests for tag propagating (3 passes - leaf-ward, root-ward, correction-ward) in annotated_forest
- [x] add 3 ways of propagation - leaf-ward, root-ward, correction-ward in annotated_forest
- [x] determine whether we want to expect to have all nodes tagged even if they have no tags (i.e. do they show up in `annotated_tags = nx.get_node_attributes(annotated_forest, "tags")`) -> No. It doesn't seem like we need this at all and it required an additional input into the forest builder so I removed it.
- [x] Remove the metadata df as an input into the forest builder. it wasn't strictly necessary.
- [x] are we enforcing tree structure now? (are we removing stepparents/stepchildren?) YES
- [x] bunch of nothing in assets and liabilities from 2005-2020 #3300
- [x] pull in the three propagations into individual _functions/_methods
- [x] add validation test: does rate base table have more than one _correction factoid?
- [x] add validation test: do exploded tables have more tags?
```

PR #3214, opened 2024-01-04
ensure all the corrections get tags and add the begining of a rate base asset

cmgosnell:
# Overview

Closes [#3203](https://github.com/catalyst-cooperative/pudl/issues/3203).

..
What did you change?
..
# Testing

```[tasklist]
# To-do list
```


src/pudl/output/ferc1.py
```diff
@@ -2774,3 +2792,70 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+            dollar_value=lambda x: x.dollar_value / 8,
+            xbrl_factoid="cash_on_hand",  # newly definied (do we need to add it anywhere?)
+            tags_rate_base_category="net_working_capital",
+            tags_aggregatable_utility_type="electric",
+            table_name="core_ferc1__yearly_operating_expenses_sched320",
+        )
+        .drop(columns=[xbrl_factoid_name])
+        # the assets/liabilites both use ending_balance for its main $$ column
+        .rename(columns={"dollar_value": "ending_balance"})
+    )
```
cmgosnell, line 2842: @jrea-rmi  this is the little process to grab the `operations_and_maintenance_expenses_electric` line from the operating expense table and convert it into working capital/cash on hand. Before adding it to the `in_rate_base` exploded data. Does this look right to you generally? Do the tags look okay & sufficient? I figured this should be `in_service` as plant status or have a plant function. but lmk it's obviously easy to add in there as an additional column.

jrea-rmi: Yes this looks right to me generally!

I'd go with xbrl_factoid as "cash_working_capital" rather than "cash_on_hand".

I don't know if it needs plant status or plant function; my initial thought is those don't need to be added. I thought in_service etc. only applied to utility plant assets, and plant function only applied to plants in service. But I could see an expanded definition of those if you think there should be.

cmgosnell: okay great! I'll change it to "cash_working_capital". And I think your right on the status and function - or at least that sounds plausible to me. I just wanted to make sure!

src/pudl/output/ferc1.py
```diff
@@ -1191,7 +1191,22 @@ def _out_ferc1__explosion_tags(table_dimensions_ferc1) -> pd.DataFrame:
         .reset_index()
         .drop(columns=["notes"])
     )
-    return tags_all
+    # Add the correction records to the tags with the same tags as the parent
+    idx = list(NodeId._fields)
```
jdangerx, line 1195: I think if we're re-using this `list(NodeId._fields)` in a bunch of places, including line 1187, we might as well give it a meaningful name (`set_index(idx)` doesn't really tell you more than `set_index(adfadfadfa)`).

This could be `node_id_fields = list(NodeId._fields)` maybe?

src/pudl/output/ferc1.py
```diff
@@ -1191,7 +1191,22 @@ def _out_ferc1__explosion_tags(table_dimensions_ferc1) -> pd.DataFrame:
...
+    correction_index = (
```
jdangerx, line 1196: This is "all un-corrected factoids, indexed by the NodeId fields," right? And we want to generate a bunch of tags for the corrected factoids by:

1. look at all the tags we have
2. inner merge "tags we have" with "all factoids we know about", using the NodeId fields as a join key + dropping all the random `table_dimensions` columns
3. add "_correction" to each factoid name & concat it onto all of our tags - tada!

Some questions:
1. Do we expect there to be tags for factoids that aren't in the table_dimensions? If not, can we get away with taking every tag + making a `_correction` version of it?
2. Are there factoids we expect to not have a `_correction` partner?
3. If we still need to filter out the tags we have that don't correspond to something in `table_dimensions`, should we use something more like `table_dimensions[node_id_fields].join(tags_all)` to make the intent of the code clearer?

src/pudl/output/ferc1.py
```diff
@@ -1236,7 +1251,10 @@ def _aggregatable_dimension_tags(
         )
         .set_index(idx)
     )
-    table_dimensions_ferc1 = table_dimensions_ferc1.set_index(idx)
+    # don't include the corrections
```
jdangerx, line 1254: We're filtering these out since we always want the correction to have the same tags as the parent, and will add those tags in later, right?

src/pudl/output/ferc1.py
```diff
@@ -2774,3 +2792,70 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+    concatenates it with the rest of the assets and liabilities from the granular
+    exploded data.
+
+    """
+    # get the factoid name to grab the right part of the table
+    xbrl_factoid_name = pudl.transform.ferc1.FERC1_TFR_CLASSES[
+        "core_ferc1__yearly_operating_expenses_sched320"
+    ]().params.xbrl_factoid_name
+    # First grab the cash on hand out of the operating expense table.
+    # then prep it for concating. Calculate cash on hand & add tags
```
jdangerx, line 2826: Seems like we use `cash_on_hand` in the XBRL factoid name but `cash_working_capital` in the code - should we standardize?

src/pudl/output/ferc1.py
```diff
@@ -2774,3 +2792,70 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+    cash_working_capital = (
+        core_ferc1__yearly_operating_expenses_sched320[
+            core_ferc1__yearly_operating_expenses_sched320[xbrl_factoid_name]
+            == "operations_and_maintenance_expenses_electric"
+        ]
+        .assign(
+            dollar_value=lambda x: x.dollar_value / 8,
```
jdangerx, line 2833: This seems like it could use the vectorized division operator, should we?

src/pudl/package_data/ferc1/xbrl_factoid_plant_status_tags.csv
```diff
@@ -26,14 +26,4 @@ core_ferc1__yearly_utility_plant_summary_sched200,depreciation_and_amortization_
 core_ferc1__yearly_utility_plant_summary_sched200,abandonment_of_leases,total
 core_ferc1__yearly_utility_plant_summary_sched200,amortization_of_plant_acquisition_adjustment,total
 core_ferc1__yearly_utility_plant_summary_sched200,utility_plant_in_service_classified_and_property_under_capital_leases,in_service
-core_ferc1__yearly_utility_plant_summary_sched200,utility_plant_in_service_plant_purchased_or_sold_correction,in_service
```
jdangerx, line 29: 🔪

jdangerx (changes requested):
I know this is in draft form still so you obviously still plan on changing it. Overall this makes sense to me, I think. I might be totally off base w my understanding, so please correct me!

Also! We should write tests for this behavior 🙂

src/pudl/output/ferc1.py
cmgosnell: do you mean just use `x.dollar_value.divide(8)`? or do you mean taking this out of the assign altogether and doing `cash_working_capital.loc[:,"dollar_value"] = cash_working_capital.dollar_value.divide(8)`?

src/pudl/output/ferc1.py
cmgosnell: yes! i'm going to convert them all to `cash_working_captial`. i was waiting to hear back from jon about desired name for the factoid.

src/pudl/output/ferc1.py
cmgosnell: ah yes you are so right `idx` is a completely meaningless name. this `list(NodeId._fields)` is used 13 (!) times throughout this module. In this context it's really just in index columns or primary key columns of the tags. so maybe `tag_idx`? in other contexts its used as the pk/idx of a calculation component or to identify a node in the calculation forest. but in the tag space we aren't really working with networks and so caling them node_id's feels a little wrong.

src/pudl/output/ferc1.py
cmgosnell: right! I'll add a more descriptive comment to confirm that.

src/pudl/output/ferc1.py
cmgosnell: hmm these are all excellent questions and i think your suggestion in your first question is correct. IF what i was trying to do here was actually the right thing to do be doing. I think i could replace all of this stuff in here with `corrections = tags_all.assign(xbrl_factoid=lambda x: x.xbrl_factoid + "_correction")`

BUT I think what we really need to be doing is saying for a given calculated record, do all of the child component records contain those same tags? if so give the correction record for the parent fact those tags.

I think i was starting to attempt that using the table dims but i don't think that's the right method. hm... i could probably do it with the calculation component table but it might be simpler actually using the tree/network methods.

src/pudl/output/ferc1.py
```diff
@@ -1249,6 +1259,64 @@ def _aggregatable_dimension_tags(
...
+) -> pd.DataFrame:
+    """Make tags for correction records.
+
+    We need to check to see if any of the tags in each of the calculated
+    parent factoids are the same for all of their child components. So in this
+    function, we're going to merge on the tags to the children then groupby the
+    parents. For each tag, see if the childrens'tags contains only one unique value.
+    If so grab the tag to associate with the correction record of the parent. If not,
+    no tag will be associated with the record.
+    """
```
cmgosnell, line 1273: this is my new attempt at adding tags to parent facts when all of their children share the same tag. BUT i think this could probably be done more switfly and more completely using the networkx graphs. if all ancestors of a node share tag X, give all ancestor nodes tag X.

cmgosnell:
hey @jrea-rmi ! do you know if we need to apply any sign changes to any of these input tables before squishing them together in this rate base table? I am mostly asking because the annual sum of both the assets and liabilities are positive. (The gaps in here are being fixed via #3300 because 🙃 )

image
image

src/pudl/output/ferc1.py
```diff
@@ -2774,3 +2811,122 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+
+    def _get_tag(annotated_forest, node, tag_name):
+        return annotated_forest.nodes.get(node, {}).get("tags", {}).get(tag_name, pd.NA)
+
+    logger.info(f"propagaging tags leafward from {node}")
+    tag = pd.NA
+    # i'm a leaf so i stop looking
+    if not list(annotated_forest.successors(node)):
+        tag = _get_tag(annotated_forest, node, tag_name)
+        logger.info(f"    We found a leaf people. w/ {tag=}")
```
cmgosnell, line 2832: ow i need to remove these lil logs 👀

src/pudl/output/ferc1.py
```diff
@@ -2252,6 +2235,55 @@ def annotated_forest(self: Self) -> nx.DiGraph:
...
+                    }
+                    | existing_tags.get(desc, {})
+                }
+                for desc in descendants
+            }
+            nx.set_node_attributes(annotated_forest, descendant_tags)
+
+        # Rootward propagation
+        root_node = self.roots(annotated_forest)[0]
+        _ = recursively_propagate_tags_leafward(
```
cmgosnell, line 2263: lol i named this wrong!! its rootward

test/unit/output/ferc1_test.py
```diff
@@ -18,59 +18,262 @@
...
+            (self.child1, self.grand_child11),
+            (self.child1, self.child1_correction),
+        ]
+        tags = pd.DataFrame([self.child1]).assign(in_rate_base=["yes"])
+        annotated_tags = self.build_forest_and_annotated_tags(
+            edges, tags, seeds=[self.child1]
+        )
+        assert annotated_tags[self.child1]["in_rate_base"] == "yes"
+        assert annotated_tags[self.grand_child11]["in_rate_base"] == "yes"
+        assert annotated_tags[self.child1_correction]["in_rate_base"] == "yes"
```
cmgosnell, line 238: these should be same as child1 tag

jrea-rmi:
> hey @jrea-rmi ! do you know if we need to apply any sign changes to any of these input tables before squishing them together in this rate base table? I am mostly asking because the annual sum of both the assets and liabilities are positive.
> 
> ## total rate base
> (i would add a legend but there are almost 200 utilities and so the legend makes the graphs un-readable) image
> 
> ## Components of Rate base by source table
> image image image image image image
> 
> ## code to make the above plots
> ```python
> import matplotlib.colors as mcolors
> import matplotlib.pyplot as plt
> from dagster import AssetKey
> from pudl.etl import defs
> ```
> 
> if you have a rate base table materialized:
> 
> ```python
> out_ferc1__yearly_rate_base = defs.load_asset_value(AssetKey("out_ferc1__yearly_rate_base"))
> ```
> 
> otherwise you can grab this pickled table: [out_ferc1__yearly_rate_base.pkl.zip](https://github.com/catalyst-cooperative/pudl/files/14068981/out_ferc1__yearly_rate_base.pkl.zip)
> 
> ```python
> annual_rb_by_util = out_ferc1__yearly_rate_base.groupby(["report_year", "utility_id_ferc1"],as_index=False)[["ending_balance"]].sum()
> 
> non_white_colors = [color for color in mcolors.CSS4_COLORS.keys() if "white" not in color]
> new_colors = non_white_colors * 4
> new_colors = new_colors[0:len(annual_rb_by_util_wide.columns)]
> 
> annual_rb_by_util_wide = annual_rb_by_util.pivot(index='report_year', columns='utility_id_ferc1', values='ending_balance')
> annual_rb_by_util_wide.plot(kind='bar', stacked=True, color=new_colors)
> plt.legend([])
> plt.title("Annual Sum of Rate Base by Utility")
> plt.show()
> ```
> 
> ```python
> annual_rb_by_util = out_ferc1__yearly_rate_base.groupby(["report_year", "utility_id_ferc1","table_name"],as_index=False)[["ending_balance"]].sum()
> 
> annual_rb_by_util = out_ferc1__yearly_rate_base.groupby(["report_year", "utility_id_ferc1","table_name"],as_index=False)[["ending_balance"]].sum()
> annual_rb_by_util_wide = annual_rb_by_util.pivot(index=["table_name",'report_year'], columns='utility_id_ferc1', values='ending_balance')
> for table in annual_rb_by_util.table_name.unique():
>     annual_rb_by_util_wide.loc[table].plot(kind='bar', stacked=True, color=new_colors)
>     plt.legend([])
>     plt.title(f"Annual Sum of Rate Base by Utility from {table}")
>     plt.show()
> ```
> 
> ## notes to investigate
> ```
> - [x] bunch of nothing in assets and liabilities from 2005-2020 #3300
> - [ ] why basically nothing in `core_ferc1__yearly_depreciation_by_function_sched219`in 2021??
> - [ ] why is 2004 so weird looking?
> ```

yes, this needs sign changes applied to the liabilities side of the balance sheet, everything from that table that's labeled as in rate base is an offset to rate base rather than a positive contribution to it.

I am confused by the magnitude of what's in rate base from the liabilities side of the balance sheet - I expect it to be much lower than the assets side.

I am also confused by the plant in service table values being so small - I expect those to be larger than accumulated depreciation and the largest component of rate base.

And finally the operating expenses table shouldn't have anything in rate base - these expenses would show up in revenue requirement but not in capital cost breakdown.

src/pudl/output/ferc1.py
```diff
@@ -2496,11 +2496,16 @@ def forest_leaves(self: Self) -> list[NodeId]:
...
+        orphans = []
+        for idx_cols in [self.calc_cols, self.parent_cols]:
+            orphans.extend(
+                [
+                    NodeId(*n)
+                    for n in self.exploded_calcs.set_index(idx_cols).index
+                    if n not in nodes
+                ]
+            )
+        return list(set(orphans))
```
cmgosnell, line 2508: this is the only other place that the `explode_meta` was being used. I think it actually makes more sense to use the `exploded_calcs` here to determine the og set of nodes here because that's what the forest was actually build on and the metadata could be quite different if we weren't being careful in the prep

src/pudl/output/ferc1.py
```diff
@@ -2774,3 +2779,239 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+                }
+                | existing_tags.get(desc, {})
+            }
+            for desc in descendants
+        }
+        nx.set_node_attributes(annotated_forest, descendant_tags)
+    return annotated_forest
+
+
+def _propagate_tag_rootward(
```
cmgosnell, line 2805: this one is named with `_tag_` instead of `_tags_` rn because it only takes one. if we wanted to keep it consistent i would call this one inside of a `_tags_` function that loops over the tags. but rn this is all only really for the `in_rate_base` tag so that felt premature/necessary.

src/pudl/output/ferc1.py
```diff
@@ -2774,3 +2779,239 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+    annotated_forest: nx.DiGraph, node, tag_name: Literal["in_rate_base"]
+) -> str:
+    """Set the tag for nodes when all of its children have same tag.
+
+    This function returns the value of a tag, but also sets node attributes
+    down the tree when all children of a node share the same tag.
+    """
+
+    def _get_tag(annotated_forest, node, tag_name):
+        return annotated_forest.nodes.get(node, {}).get("tags", {}).get(tag_name, pd.NA)
```
cmgosnell, line 2815: this was just a lil helper function to get the tag or a null because omigosh as you can see it is a lil complicated because of the layered-ness and option for the node to not exist or the tag to not exist etc. I suppose it could also be:

```python
annotated_forest.nodes.get(node, {"tags", {tag_name: pd.NA}})["tags"][tag_name]
```

src/pudl/output/ferc1.py
```diff
@@ -2774,3 +2779,239 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+        parent = parents[0]
+        correction_tags[correction_node] = {
+            "tags": existing_tags.get(parent, {})
+            | existing_tags.get(correction_node, {})
+        }
+    nx.set_node_attributes(annotated_forest, correction_tags)
+    return annotated_forest
+
+
+def check_tag_propagation_compared_to_compiled_tags(
```
cmgosnell, line 2868: i added two `check_*` functions here. they are both being called rn in `out_ferc1__yearly_rate_base`. really these should be validation tests i believe for both the rate base table and the two exploded tables detailed in the args. i could put it into the `exploded_table_asset_factory` and just skip it for the income table. thoughts?

cmgosnell:
@jrea-rmi  thanks for these insights!! this is great info.

> yes, this needs sign changes applied to the liabilities side of the balance sheet, everything from that table that's labeled as in rate base is an offset to rate base rather than a positive contribution to it.

Okay, grand i'll change that sign. Is that the only input that needs a sign change?
 
> I am confused by the magnitude of what's in rate base from the liabilities side of the balance sheet - I expect it to be much lower than the assets side.

yea i was also a little surprised by that. I will try to investigate to see if there anything weird there and will report back.

> I am also confused by the plant in service table values being so small - I expect those to be larger than accumulated depreciation and the largest component of rate base.

ditto on the will investigate and report back!

> And finally the operating expenses table shouldn't have anything in rate base - these expenses would show up in revenue requirement but not in capital cost breakdown.

the operating expense table values are just the `operations_and_maintenance_expenses_electric` / 8 values that got added after the fact for `cash_working_capital`.

src/pudl/output/ferc1.py
```diff
@@ -2774,3 +2775,239 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+    """Set the tag for nodes when all of its children have same tag.
+
+    This function returns the value of a tag, but also sets node attributes
+    down the tree when all children of a node share the same tag.
+    """
+
+    def _get_tag(annotated_forest, node, tag_name):
+        return annotated_forest.nodes.get(node, {}).get("tags", {}).get(tag_name, pd.NA)
+
+    logger.info(f"propagaging tags leafward from {node}")
```
jdangerx, line 2813: Here's a non-recursive solution that uses `nx.topological_generations()`. I sort of think it's a little easier to read, since the flow is:

```
for each generation (starting from last):
    for each node in generation:
        set node tags, if children agree
```

We could skip the *very* last generation here since we know they have no children.

```diff
diff --git a/src/pudl/output/ferc1.py b/src/pudl/output/ferc1.py
index ecd6bfae3..4c9773bc8 100644
--- a/src/pudl/output/ferc1.py
+++ b/src/pudl/output/ferc1.py
@@ -1,5 +1,6 @@
 """A collection of denormalized FERC assets and helper functions."""
 import importlib
 import re
 from copy import deepcopy
 from functools import cached_property
@@ -2243,7 +2244,7 @@ class XbrlCalculationForestFerc1(BaseModel):
         annotated_forest = _propagate_tags_leafward(annotated_forest, ["in_rate_base"])
         # Rootward propagation
         root_node = self.roots(annotated_forest)[0]
-        _ = _propagate_tag_rootward(annotated_forest, root_node, "in_rate_base")
+        _propagate_tag_rootward(annotated_forest, "in_rate_base")
         # Correction Records
         annotated_forest = _propagate_tags_to_corrections(annotated_forest)
         return annotated_forest
@@ -2799,8 +2800,8 @@ def _propagate_tags_leafward(
 
 
 def _propagate_tag_rootward(
-    annotated_forest: nx.DiGraph, node, tag_name: Literal["in_rate_base"]
-) -> str:
+    annotated_forest: nx.DiGraph, tag_name: Literal["in_rate_base"]
+) -> None:
     """Set the tag for nodes when all of its children have same tag.
 
     This function returns the value of a tag, but also sets node attributes
@@ -2808,37 +2809,34 @@ def _propagate_tag_rootward(
     """
 
     def _get_tag(annotated_forest, node, tag_name):
-        return annotated_forest.nodes.get(node, {}).get("tags", {}).get(tag_name, pd.NA)
-
-    logger.info(f"propagaging tags leafward from {node}")
-    tag = pd.NA
-    # i'm a leaf so i stop looking or
-    # if i have a value you don't need to keep looking at this node's childern
-    if not list(annotated_forest.successors(node)) or not pd.isna(
-        _get_tag(annotated_forest, node, tag_name)
-    ):
-        tag = _get_tag(annotated_forest, node, tag_name)
-
-    else:
-        child_tags = set()
-        for child_node in annotated_forest.successors(node):
-            if not child_node.xbrl_factoid.endswith("_correction"):
-                child_tags.add(
-                    _propagate_tag_rootward(annotated_forest, child_node, tag_name)
-                )
-        # if all the children tags are the same and non-null
-        if (len(child_tags) == 1) and {t for t in child_tags if not pd.isna(t)}:
-            new_node_tag = child_tags.pop()
-            # actually assign the tag here but don't wipe out any other tags
-            existing_tags = nx.get_node_attributes(annotated_forest, "tags")
-            node_tags = {
-                node: {"tags": {tag_name: new_node_tag} | existing_tags.get(node, {})}
+        return annotated_forest.nodes.get(node, {}).get("tags", {}).get(tag_name)
+
+    generations = list(nx.topological_generations(annotated_forest))
+    for gen in reversed(generations):
+        untagged_nodes = {
+            node_id
+            for node_id in gen
+            if _get_tag(annotated_forest, node_id, tag_name) is None
+        }
+        for parent_node in untagged_nodes:
+            child_tags = {
+                _get_tag(annotated_forest, c, tag_name)
+                for c in annotated_forest.successors(parent_node)
+                if not c.xbrl_factoid.endswith("_correction")
             }
-            nx.set_node_attributes(annotated_forest, node_tags)
-            tag = new_node_tag
-        # elif the children disagree then the node's tag shouldn't be set and
-        # the og null tag should be returned
-    return tag
+            non_null_tags = child_tags - {None}
+            # sometimes, all children can share same tag but it's null.
+            if len(child_tags) == 1 and non_null_tags:
+                # actually assign the tag here but don't wipe out any other tags
+                new_node_tag = non_null_tags.pop()
+                existing_tags = nx.get_node_attributes(annotated_forest, "tags")
+                node_tags = {
+                    parent_node: {
+                        "tags": {tag_name: new_node_tag}
+                        | existing_tags.get(parent_node, {})
+                    }
+                }
+                nx.set_node_attributes(annotated_forest, node_tags)
 
 
 def _propagate_tags_to_corrections(annotated_forest: nx.DiGraph) -> nx.DiGraph:
```


It would probably be good, also, to break some of this logic out into its own little named functions - e.g. "do the children agree, on a non-null value?" is a surprisingly gnarly thing to figure out, and it seems very likely that we'd change the implementation in the future. By pulling it out we'd save a little time puzzling over "what code decides if the children agree?" in the future.

src/pudl/output/ferc1.py
jdangerx: Calling them in out_ferc1__yearly_rate_base seems fine, though you could try stuffing them in an [asset check](https://docs.dagster.io/concepts/assets/asset-checks) in the factory. I think either is fine.

src/pudl/output/ferc1.py
```diff
@@ -2774,3 +2775,239 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+            ``xbrl_factoids`` in ``df`` than found in ``_out_ferc1__explosion_tags``.
+    """
+    # the tag df has all tags - not just those in a specific explosion
+    # so we need to drop
+    node_idx = list(NodeId._fields)
+    df_filtered = df.filter(node_idx).drop_duplicates()
+    df_tags = _out_ferc1__explosion_tags.merge(
+        df_filtered, on=list(df_filtered.columns), how="right"
+    )
+    mannually_tagged = df_tags[df_tags[propogated_tag].notnull()].xbrl_factoid.unique()
```
jdangerx, line 2895: Couple typos here - `manually_tagged` and `propagated_tag`

jdangerx (changes requested):
Mostly looks good! Spent a bit of time messing with a non-recursive solution to rootward propagation, which I *think* is easier to reason about / maintain going into the future. Let me know what you think!

Also, I have one major question just about what our expected behavior is, plus some typo cleanup.

test/unit/output/ferc1_test.py
```diff
@@ -18,59 +18,287 @@
...
-# CALC_TREE_1 = XbrlCalculationForestFerc1(
-#    exploded_meta=TEST_EXPLODED_META,
-#    seeds=[NodeId("table_1", "calc_1")],
-# )
-# CALC_TREE_2 = XbrlCalculationForestFerc1(
-#    exploded_meta=TEST_EXPLODED_META,
-#    seeds=[NodeId("table_2", "calc_2")],
-# )
+
+# TODO: combine these into a class because we have a lot of similar method names
```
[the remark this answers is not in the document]
cmgosnell: TODONE!

test/unit/output/ferc1_test.py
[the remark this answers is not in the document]
cmgosnell: ow yea i hadn't touched the module docs but they are oooold 👀

test/unit/output/ferc1_test.py
```diff
@@ -18,59 +18,287 @@
...
+            annotated_tags[self.child2]["in_rate_base"]
+            == annotated_tags[self.parent]["in_rate_base"]
+        )
+
+    def test_leafward_prop_preserve_non_propagating_tags(self):
+        """Don't force the diagreeing child to follow the parent."""
+        edges = [(self.parent, self.child1), (self.parent, self.child2)]
+        tags = pd.DataFrame([self.parent, self.child1, self.child2]).assign(
+            in_rate_base=["yes", "no", pd.NA],
+            in_root_boose=["yus", "nu", "purtiul"],
```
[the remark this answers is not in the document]
cmgosnell: oh this is a good point

test/unit/output/ferc1_test.py
```diff
@@ -18,59 +18,287 @@
...
+        for node in null_tag_edges:
+            assert not annotated_tags.get(node)
+
+        tags = pd.DataFrame([self.parent, self.child1, self.child2]).assign(
+            in_rate_base=[pd.NA, pd.NA, pd.NA],
+            a_non_propped_tag=["hi", "hello", "what_am_i_doing_here_even"],
+        )
+        annotated_tags = self.build_forest_and_annotated_tags(edges, tags)
+        for node in null_tag_edges:
+            assert not annotated_tags[node].get("in_rate_base")
```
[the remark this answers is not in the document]
cmgosnell: this is a very good point! i will change here and elsewhere

test/unit/output/ferc1_test.py
```diff
@@ -18,59 +18,287 @@
...
+            (self.parent, self.child1),
+            (self.parent, self.child2),
+            (self.child1, self.grand_child11),
+            (self.child1, self.grand_child12),
+        ]
+        tags = pd.DataFrame([self.grand_child11, self.grand_child12]).assign(
+            in_rate_base=["yes", "yes"]
+        )
+        annotated_tags = self.build_forest_and_annotated_tags(edges, tags)
+        assert not annotated_tags.get(self.parent)
```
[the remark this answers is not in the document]
cmgosnell: we expect parent to only have "yes" if *all* its children have the same tag

src/pudl/output/ferc1.py
cmgosnell: lol ty

src/pudl/output/ferc1.py
```diff
@@ -2173,32 +2177,9 @@ def node_attrs(self: Self) -> dict[NodeId, dict[str, dict[str, str]]]:
             .reset_index()
             # Type conversion is necessary to get pd.NA in the index:
             .astype({col: pd.StringDtype() for col in self.calc_cols})
-            # We need a dictionary for *all* nodes, not just those with tags.
```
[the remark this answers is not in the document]
cmgosnell: yes and! this was enforcing that we had all the nodes in the graph having null tags (the `{"tags": {}}`) even when those nodes weren't included in the tags we passed in.

but the addition of the `.dropna(how="all")` above is how we are now not added null tags when those nodes are in the tags we pass in but those tags are all null.

src/pudl/output/ferc1.py
```diff
@@ -2496,11 +2492,16 @@ def forest_leaves(self: Self) -> list[NodeId]:
...
-            for n in self.exploded_meta.set_index(self.calc_cols).index
-            if n not in nodes
-        ]
+        orphans = []
+        for idx_cols in [self.calc_cols, self.parent_cols]:
+            orphans.extend(
+                [
+                    NodeId(*n)
+                    for n in self.exploded_calcs.set_index(idx_cols).index
+                    if n not in nodes
```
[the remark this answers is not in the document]
cmgosnell: oh this is a good question... This has definitely picked up orphans without the `NodeId(*n) not in nodes` so i assume it works as is but i don't think it would hurt to add this guy (on a broader note i am already thinking about ways to adopt the unit test setup we added here to test other things o/s of tag propagation like this or #3309

jrea-rmi:
> @jrea-rmi thanks for these insights!! this is great info.
> 
> > yes, this needs sign changes applied to the liabilities side of the balance sheet, everything from that table that's labeled as in rate base is an offset to rate base rather than a positive contribution to it.
> 
> Okay, grand i'll change that sign. Is that the only input that needs a sign change?
> 
> > I am confused by the magnitude of what's in rate base from the liabilities side of the balance sheet - I expect it to be much lower than the assets side.
> 
> yea i was also a little surprised by that. I will try to investigate to see if there anything weird there and will report back.
> 
> > I am also confused by the plant in service table values being so small - I expect those to be larger than accumulated depreciation and the largest component of rate base.
> 
> ditto on the will investigate and report back!
> 
> > And finally the operating expenses table shouldn't have anything in rate base - these expenses would show up in revenue requirement but not in capital cost breakdown.
> 
> the operating expense table values are just the `operations_and_maintenance_expenses_electric` / 8 values that got added after the fact for `cash_working_capital`.

I believe so yes, the only sign change when combining tables into rate base is to flip sign of liabilities side of balance sheet.

I'll stay tuned for investigation of the liabilities and plant in service values.

And okay, that makes sense on inclusion of operating expenses table for cash working capital component of rate base!

src/pudl/output/ferc1.py
```diff
@@ -2809,3 +2809,241 @@ def nodes_to_df(calc_forest: nx.DiGraph, nodes: list[NodeId]) -> pd.DataFrame:
...
+            }
+            for desc in descendants
+        }
+        nx.set_node_attributes(annotated_forest, descendant_tags)
+    return annotated_forest
+
+
+def _propagate_tag_rootward(
+    annotated_forest: nx.DiGraph, tag_name: Literal["in_rate_base"]
+) -> str:
```
jdangerx, line 2841:
```suggestion
) -> nx.DiGraph:
```

src/pudl/output/ferc1.py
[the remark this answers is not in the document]
jdangerx: nit: did you still want to add this? either way is fine w/ me

test/unit/output/ferc1_test.py
```diff
@@ -12,54 +10,305 @@
 - pruning of passthrough nodes & associated corrections
 - propagation of weights
 - conflicting weights
-- propagation of tags
 - conflicting tags
 - validation of calculations using only leaf-nodes to reproduce root node values
 
+Stuff we are testing:
+- propagation of tags
```
jdangerx, line 17: 🎉

jdangerx (approved):
Looks good to me! You had mentioned wanting to coerce a tuple to a `NodeId`, though the code works now without coercion - I think it's probably worth changing, but either way works for me.

src/pudl/output/ferc1.py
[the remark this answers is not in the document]
cmgosnell: i didn't transition it bc it seemed hard to test initially. but i rejiggered the tests a bit to make sure i could triple check that this was going to work okay. turns out the change in using the `exploded_calcs` here instead of the old metadata table makes this whole `orphans` thing I believe redundant and unnecessary. but i'd like to double check that with zane before ripping it out.

zaneselvans:
@cmgosnell @jdangerx If this table isn't ready to be published in the DB, maybe we should give it a `_out` prefix, so that people aren't confused when they see it in the data dictionary and can't find it in the database?

Issue #3102, opened 2023-12-01
Output PUDL as Parquet as well as SQLite


zaneselvans:
We've [heard feedback from users](https://github.com/orgs/catalyst-cooperative/discussions/3071) that the monolithic 12GB PUDL SQLite DB is a bit chonky for download an integration into other workflows when they only need a couple of tables, and that some kind of remote, cloud-optimized outputs would be very handy.  One way to address these needs is to publish Parquet files representing the tables in our DB.

## Pros of Parquet:
- can be read remotely very efficiently due to their chunked row-groups and columnar format.
- compressed internally so take up very little space (around 3 GB for everything in the uncompressed 12 GB SQLite DB)
- provide rich PyArrow data types that can be read into Pandas, Dask, and other dataframe libraries without requiring any additional schema information.
- are platform independent and can be used by many tools
- optimized for fast read/write and analytical workflows
- can be queried like database tables using DuckDB

## Cons of Parquet:
- no foreign key relationships and much less in the way of constraints that ensure everything is internally consistent
- will necessitate a large directory of ~200 files, which users could easily end up mixing and matching from different versions, resulting in internal incompatibilities.
- We don't currently know how or if we can collect enough information about what data is being used, by who, and for what purpose through our completely open, file-based data distribution system. AWS S3 offers a bewildering array of logging tools, which we haven't really explored. See: [logging with S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/logging-with-S3.html), [CloudTrail Logging](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cloudtrail-logging.html), [CloudWatch Monitoring](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cloudwatch-monitoring.html), [ServerLogs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerLogs.html)

On balance, we've decided that this would be a valuable addition to the data we publish, and should be relatively straightforward given the infrastructure we already have in place.

## Current Situation:

After merging #3222:

- We can create a `pyarrow` schema for and output any PUDL `Resource` as a Parquet file.
- The `pyarrow` schema uses this mapping of simplified data types:
```py
FIELD_DTYPES_PYARROW: dict[str, pa.lib.DataType] = {
    "boolean": pa.bool_(),
    "date": pa.date32(),  # date with a resolution of days.
    "datetime": pa.timestamp("ms", tz="UTC"),
    "integer": pa.int32(),
[...]

PR #3222, opened 2024-01-08
Tweak metadata and pyarrow schema methods to work for all tables

zaneselvans:
# Overview

## What problem/need does this address?

Previously, we were unable to use the `Resource.to_pyarrow()` method to generate PyArrow schemas for all of our tables, and had historically only used it for the billion-row EPA CEMS table. It turns out that the issues preventing the schemas for being valid were minor and easy to fix:

- Some tables and fields were missing a `description` field, which we were assuming would be present and trying to write into the Parquet metadata.
- Some tables and fields don't have a `primary_key` but we were assuming one always existed, and trying to write that into the Parquet metadata as well.

## What did you change?

- I added a few missing titles / descriptions and also added a conditional that uses an empty string if the description is `None`.
- I added a conditional that only writes the `primary_key` to the Parquet metadata if it is not `None`.
- Added a unit test ensuring that all Resources can generate their PyArrow schemas.
- Added a unit test ensuring that all defined fields can actually be instantiated (we had some bad ones with typos that were never used)
- Removed the unfinished and problematic `entity_types_eia` table (and renamed its vestigial form `core_eia__codes_entity_types`)
- Added some currently `xfail` tests for: unused fields, fields without descriptions, and resources without descriptions.

Part of #3102
Followup in #3224 

# Testing

I iterated through every `Resource` we generate, and attempted to create a PyArrow schema from it, and then attempted to use that schema to write the table out to a Parquet file successfully.

```[tasklist]
# To-do list
- [x] Disable the confusing `entity_types_eia` table which does not actually exist.
- [x] Add an integration test that attempts to create PyArrow schemas for all resources.
- [x] Make sure full ETL runs locally (especially EPA CEMS)
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -137,7 +137,7 @@ def data_dictionary_metadata_to_rst(app):
     """Export data dictionary metadata to RST for inclusion in the documentation."""
     # Create an RST Data Dictionary for the PUDL DB:
     print("Exporting PUDL DB data dictionary metadata to RST.")
-    skip_names = ["datasets", "accumulated_depreciation_ferc1", "entity_types_eia"]
+    skip_names = ["datasets", "accumulated_depreciation_ferc1"]
     names = [name for name in RESOURCE_METADATA if name not in skip_names]
     package = Package.from_resource_ids(resource_ids=tuple(sorted(names)))
     # Sort fields within each resource by name:
--- a/src/pudl/metadata/classes.py
+++ b/src/pudl/metadata/classes.py
@@ -624,7 +624,9 @@ class Field(PudlMeta):
             name=self.name,
             type=self.to_pyarrow_dtype(),
             nullable=(not self.constraints.required),
-            metadata={"description": self.description},
+            metadata={
+                "description": self.description if self.description is not None else ""
+            },
         )
 
     def to_sql(  # noqa: C901
@@ -1318,9 +1320,10 @@ class Resource(PudlMeta):
         """Construct a PyArrow schema for the resource."""
         fields = [field.to_pyarrow() for field in self.schema.fields]
         metadata = {
-            "description": self.description,
-            "primary_key": ",".join(self.schema.primary_key),
+            "description": self.description if self.description is not None else ""
         }
+        if self.schema.primary_key is not None:
+            metadata |= {"primary_key": ",".join(self.schema.primary_key)}
         return pa.schema(fields=fields, metadata=metadata)
 
     def to_pandas_dtypes(self, **kwargs: Any) -> dict[str, str | pd.CategoricalDtype]:
--- a/src/pudl/metadata/codes.py
+++ b/src/pudl/metadata/codes.py
@@ -795,101 +795,6 @@ CODE_METADATA: dict[str, dict[str, Any]] = {
         "code_fixes": {"5": "F"},
         "ignored_codes": [],
     },
-    "entity_types_eia": {
-        "df": pd.DataFrame(
-            columns=[
-                "code",
-                "label",
-                "description",
-            ],
-            data=[
-                (
-                    "A",
-                    "municipal_marketing_authority",
-                    "Municipal Marketing Authority. Voted into existence by the residents of a municipality and given authority for creation by the state government. They are nonprofit organizations",
-                ),
-                (
-                    "B",
-                    "behind_the_meter",
-                    "Behind the Meter. Entities that install, own, and/or operate a system (usually photovoltaic), and sell, under a long term power purchase agreement (PPA) or lease, all the production from the system to the homeowner or business with which there is a net metering agreement. Third Party Owners (TPOs) of PV solar installations use this ownership code.",
-                ),
-                ("C", "cooperative", "Cooperative. Member-owned organizations."),
-                ("COM", "commercial", "Commercial facility."),
-                (
-                    "D",
-                    "nonutility_dsm_administrator",
-                    "Non-utility DSM Administrator. Only involved with Demand-Side Management activities.",
-                ),
-                (
-                    "F",
-                    "federal",
-                    "Federal. Government agencies with the authority to deliver energy to end-use customers.",
-                ),
-                ("G", "community_choice_aggregator", "Community Choice Aggregator."),
-                (
-                    "I",
-                    "investor_owned",
-                    "Investor-owned Utilities. Entities that are privately owned and provide a public service.",
-                ),
-                ("IND", "industrial", "Industrial facility."),
-                (
-                    "M",
-                    "municipal",
-                    "Municipal: Entities that are organized under authority of state statute to provide a public service to residents of that area.",
-                ),
-                ("O", "other", "Other entity type."),
-                (
-                    "P",
-                    "political_subdivision",
-                    'Political Subdivision. (also called "public utility district"): Independent of city or county government and voted into existence by a majority of the residents of any given area for the specific purpose of providing utility service to the voters. State laws provide for the formation of such districts.',
-                ),
-                ("PO", "power_marketer", "Power marketer."),
-                ("PR", "private", "Private entity."),
-                (
-                    "Q",
-                    "independent_power_producer",
-                    "Independent Power Producer or Qualifying Facility. Entities that own power plants and sell their power into the wholesale market.",
-                ),
-                (
-                    "R",
-                    "retail_power_marketer",
-                    "Retail Power Marketer or Energy Service Provider: Entities that market power to customers in restructured markets.",
-                ),
-                (
-                    "S",
-                    "state",
-                    "State entities that own or operate facilities or provide a public service.",
-                ),
-                (
-                    "T",
-                    "transmission",
-                    "Transmission: Entities that operate or own high voltage transmission wires that provide bulk power services.",
-                ),
-                ("U", "unknown", "Unknown entity type."),
-                (
-                    "W",
-                    "wholesale_power_marketer",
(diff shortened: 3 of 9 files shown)


src/pudl/metadata/classes.py
```diff
@@ -624,7 +624,9 @@ def to_pyarrow(self) -> pa.Field:
             name=self.name,
             type=self.to_pyarrow_dtype(),
             nullable=(not self.constraints.required),
-            metadata={"description": self.description},
+            metadata={
+                "description": self.description if self.description is not None else ""
```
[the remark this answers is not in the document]
zaneselvans: Ideally, no. We want them all to have descriptions.

Practically, this is a bit of work. In the case of the columns, many of the ones that didn't have descriptions were extremely generic, and had notes about the need to disambiguate them between the many different tables that they show up in, since they don't actually always mean the same thing (so they at least need specific descriptions, and maybe they need to be renamed so that there's closer to a 1:1 mapping between column name and description).

And then there's the EIA-861, which has dozens of tables and hundreds of columns with no descriptions.

We can certainly use Pydantic to force them all to have descriptions (and we probably should) but there's probably a couple of days worth of work to get everything filled in with clear and accurate metadata.  Since we don't actually know what all of these columns are.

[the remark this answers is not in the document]
e-belfer: I'm ok to split this off into its own issue if we're able to catalog all the columns that don't currently have descriptions.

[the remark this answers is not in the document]
zaneselvans: I've created #3224 enumerating:
- resources without descriptions
- unused fields
- fields without descriptions

docs/conf.py
```diff
@@ -137,7 +137,7 @@ def data_dictionary_metadata_to_rst(app):
     """Export data dictionary metadata to RST for inclusion in the documentation."""
     # Create an RST Data Dictionary for the PUDL DB:
     print("Exporting PUDL DB data dictionary metadata to RST.")
-    skip_names = ["datasets", "accumulated_depreciation_ferc1", "entity_types_eia"]
```
zaneselvans, line 140: The never-quite-finished `entity_types_eia` table (now `core_eia__codes_entity_types`) was malingering and creating weird special cases so I excised it.

src/pudl/metadata/codes.py
```diff
@@ -2391,3 +2296,103 @@
         "ignored_codes": [],
     },
 }
+
+# The entity type codes were never fully reconciled. Preserving this work for reference.
+# See https://github.com/catalyst-cooperative/pudl/issues/1392
+DISABLED_CODE_METADATA = {
```
zaneselvans, line 2302: I didn't want to just delete this because it has useful information, but it's not finished.

src/pudl/metadata/fields.py
```diff
@@ -1728,15 +1729,15 @@
...
+        "description": "The name of the EIA operator utility.",
     },
     "operator_state": {
         "type": "string",
         "description": "The state where the operator utility is located.",
     },
     "operator_utility_id_eia": {
         "type": "integer",
-        "descrption": "The EIA utility Identification number for the operator utility.",
+        "description": "The EIA utility Identification number for the operator utility.",
```
zaneselvans, line 1740: Typos which were not causing errors because these fields aren't actually used anywhere, and so are never instantiated.

src/pudl/metadata/resources/pudl.py
```diff
@@ -56,6 +56,8 @@
         "sources": ["pudl"],
     },
     "core_pudl__assn_utilities_plants": {
+        "title": "PUDL Utility-Plant Associations",
+        "description": "Associations between PUDL utility IDs and PUDL plant IDs. This table is read in from a spreadsheet stored in the PUDL repository: src/pudl/package_data/glue/pudl_id_mapping.xlsx",
```
zaneselvans, line 60: A few tables that I added titles and descriptions to before realizing there were dozens, an implemented the hacky workaround.

test/unit/metadata_test.py
```diff
@@ -39,3 +42,60 @@ def test_get_etl_group_tables() -> None:
...
+@pytest.mark.parametrize("resource_name", sorted(PUDL_RESOURCES.keys()))
+def test_pyarrow_schemas(resource_name: str):
+    """Verify that we can produce pyarrow schemas for all defined Resources."""
+    _ = PUDL_RESOURCES[resource_name].to_pyarrow()
+
+
+@pytest.mark.parametrize("field_name", sorted(FIELD_METADATA.keys()))
+def test_field_definitions(field_name: str):
+    """Check that all defined fields are valid."""
+    _ = Field(name=field_name, **FIELD_METADATA[field_name])
```
zaneselvans, line 56: Would have caught those typos I mentioned above.

We can remove this after purging the unused fields.

test/unit/metadata_test.py
```diff
@@ -39,3 +42,60 @@ def test_get_etl_group_tables() -> None:
     """Test that a Value error is raised for non existent etl group."""
     with pytest.raises(ValueError):
         Package.get_etl_group_tables("not_an_etl_group")
+
+
+@pytest.mark.parametrize("resource_name", sorted(PUDL_RESOURCES.keys()))
+def test_pyarrow_schemas(resource_name: str):
+    """Verify that we can produce pyarrow schemas for all defined Resources."""
+    _ = PUDL_RESOURCES[resource_name].to_pyarrow()
```
zaneselvans, line 50: This was the main new test I wanted to add -- just so we catch any changes which happen to invalidate our PyArrow schemas (especially since we aren't actually using them yet, so it wouldn't break the builds)

test/unit/metadata_test.py
```diff
@@ -39,3 +42,60 @@ def test_get_etl_group_tables() -> None:
...
+    resources_without_description = []
+    for resource_name, resource in PUDL_RESOURCES.items():
+        if resource.description is None:
+            resources_without_description.append(resource_name)
+    resources_without_description = sorted(resources_without_description)
+    if len(resources_without_description) > 0:
+        raise AssertionError(
+            f"Found {len(resources_without_description)} resources without descriptions: "
+            f"{resources_without_description}"
+        )
```
zaneselvans, line 101: Once #3224 has been finished and we are requiring all resources and fields to have descriptions we can remove these, but they were useful in compiling the lists of things that need to be fixed.

test/unit/metadata_test.py
```diff
@@ -39,3 +42,60 @@ def test_get_etl_group_tables() -> None:
...
+    """Check that all fields which are defined are actually used."""
+    used_fields = set()
+    for resource in PUDL_RESOURCES.values():
+        used_fields |= {f.name for f in resource.schema.fields}
+    defined_fields = set(FIELD_METADATA.keys())
+    unused_fields = sorted(defined_fields - used_fields)
+    if len(unused_fields) > 0:
+        raise AssertionError(
+            f"Found {len(unused_fields)} unused fields: {unused_fields}"
+        )
```
zaneselvans, line 70: I don't *think* there are good reasons to have unused fields laying around.  I suspect that the ones we've got are just leftovers where we renamed a column and never purged the old definition.

src/pudl/metadata/resources/eia.py
```diff
@@ -444,15 +444,17 @@
...
+    # See https://github.com/catalyst-cooperative/pudl/issues/1392
+    # "core_eia__codes_entity_types": {
+    #    "description": "Descriptive labels for EIA entity type and ownership codes, taken from the EIA-861 form instructions, valid through 2023-05-31.",
+    #    "schema": {"fields": ["code", "label", "description"], "primary_key": ["code"]},
+    #    "encoder": CODE_METADATA["core_eia__codes_entity_types"],
+    #    "sources": ["eia861"],
+    #    "etl_group": "static_eia_disabled",  # currently not being loaded into the db
+    #    "field_namespace": "eia",
+    #    "create_database_schema": False,
+    # },
```
zaneselvans, line 457: This doesn't feel great, but the table is 99% functional and we really should get it over the line, so I didn't want to delete this definition which indicates what it should look like.

test/unit/metadata_test.py
zaneselvans, line 1: I realize the tests I've added here aren't truly "unit" tests, but I think it would be better to catch these kinds of errors earlier rather than later, and they only add 2 seconds to the unit tests.

src/pudl/metadata/codes.py
bendnorman: Oh yay thank you for dealing with this

test/unit/metadata_test.py
bendnorman: I think these tests are worth keeping around after we complete #3224. We might add typos and extra fields into the system again.

bendnorman (approved):
Yay! I'm glad this was an easy fix.

src/pudl/metadata/resources/eia.py
bendnorman: Yeah this seems fine for now.

test/unit/metadata_test.py
zaneselvans: Those errors would be caught by Pydantic when it attempts to instantiate all of the Resources in the Package, and all of the Fields in the Resources, which it currently does (for better or worse) on import, if we make `description` a required field, and extra fields are already disallowed.

zaneselvans: We just can't enforce that now since we have 15 resources with no description and 150 fields with no description, and 42 fields that we aren't using so they never get validated by Pydantic.

bendnorman: Oh I see. That makes sense 👍

PR #2995, opened 2023-10-31
Rename FERC Form 1 core and output assets

cmgosnell:
# PR Overview
See tasklist and decision on names over in: #2992

# PR Checklist

- [x] Merge the most recent version of the branch you are merging into (probably `dev`).
- [x] All CI checks are passing. [Run tests locally to debug failures](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#running-tests-with-tox)
- [x] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/latest/release_notes.html) and reference reference the PR and related issues.
- [ ] Make sure you've included good docstrings.
- [ ] For major data coverage & analysis changes, [run data validation tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#data-validation)
- [ ] Include unit tests for new functions and classes.
- [ ] Defensive data quality/sanity checks in analyses & data processing functions.
- [ ] Do your own explanatory review of the PR to help the reviewer understand what's going on and identify issues preemptively.

PR #3226, opened 2024-01-09
Hide `_out` tables and restore metadata in Datasette

### LINKED ISSUE #3223: Remove _out tables from datasette and sort tables
Right now tables are ordered alphabetically on data.catalyst.coop (`_out`, `core`, `out`). This isn't ideal because it promotes the table in the reverse of the ideal order. We should reorder the tables so `out` is at the top of the page, then `core`, then `_out`.

We could also just remove `_out` tables but then there will be a discrepancy between the data on data.catalyst.coop and the data in the sqlite databases. The `_out` tables will be removed from the database once the #3215 is completed.

```[tasklist]
### Tasks
- [ ] Hide `_out` tables on datasette
- [ ] Resole bug that resulted in most output tables not having any metadata on datasette
- [ ] Sort tables so `out` tables are first
```

bendnorman:
# Overview

This PR:
- Hides `_out` tables from datasette
- Removes some resources sorting logic that didn't work
- Resolve a bug that resulted in most of our output tables not having any metadata on datasette
- Adds to test to make sure all tables in the databases have metadata

This PR will not deal with sorting the tables on datasette.

# Testing

```[tasklist]
# To-do list
- [x] Write a test that ensures `metadata.yaml` has metadata for all tables
```


src/pudl/metadata/sources.py
```diff
@@ -299,7 +299,7 @@
         "license_raw": LICENSES["us-govt"],
         "license_pudl": LICENSES["cc-by-4.0"],
     },
-    "core_epa__assn_eia_epacamd": {
```
bendnorman, line 302: This was an oopsie from #2995

src/pudl/metadata/classes.py
```diff
@@ -1971,19 +1977,18 @@ def from_data_source_ids(
...
         data_sources = [DataSource.from_id(ds_id) for ds_id in data_source_ids]
 
         # Instantiate all possible resources in a Package:
-        pkg = Package.from_resource_ids()
-        # Grab a list of just the resources we want to output:
-        resources = [
-            res
-            for res in pkg.resources
-            if res.etl_group in data_source_ids + extra_etl_groups
-        ]
```
bendnorman, line 1986: I removed this filtering by source id and etl_group because all of our output table metadata was being excluded. Also, I think we want all resources to be loaded into datasette. 

[`out_eia860__yearly_emissions_control_equipment`](https://data.catalyst.coop/pudl/out_eia860__yearly_emissions_control_equipment) is the only table with metadata right now on datasette.

src/pudl/metadata/classes.py
```diff
@@ -1971,19 +1977,18 @@ def from_data_source_ids(
...
+        resources = Package.from_resource_ids().get_sorted_resources()
```
bendnorman, line 1986: The resources are correctly sorted in the metadata.yml file but it doesn't have an effect on the final ordering in the datasette UI. Any idea how to do this @zaneselvans?

bendnorman: Looks like I might need to adjust the [HTML templates datasette uses.](https://docs.datasette.io/en/stable/custom_templates.html#custom-templates) I think it's currently just pulling the table names from the database directly: https://github.com/simonw/datasette/blob/2ff4d4a60a348c143f79d63c48c329ffd0c1f02f/datasette/templates/database.html#L50-L82

zaneselvans: The layout of Datasette pages (and the set of pages which exist) can be customized extensively with custom JavaScript, CSS, and [Jinja templates](https://docs.datasette.io/en/stable/custom_templates.html#custom-templates)

It seems like maybe we want a child-template that extends the database page that lists all of the tables, and sorts / filters them.

zaneselvans: And if we want to hide all of the `_out` tables, then it seems like we can add them to [the list of `hidden_tables`](https://docs.datasette.io/en/stable/metadata.html#hiding-tables) in the metadata that we generate.

zaneselvans: Seems like we need to create a directory of our own `templates` and create a `database-pudl.html` Jinja template that overrides the default `database.html` template, at least for the content block?

src/pudl/metadata/classes.py
```diff
@@ -2004,15 +1975,10 @@ def from_data_source_ids(
...
-    def to_yaml(self, exclude_intermediate_resources: bool = False) -> None:
+    def to_yaml(self) -> None:
         """Output database, table, and column metadata to YAML file."""
         template = _get_jinja_environment().get_template("datasette-metadata.yml.jinja")
-        if exclude_intermediate_resources:
-            [
-                resource
-                for resource in self.resources
-                if not resource.name.startswith("_")
-            ]
```
bendnorman, line 2015: Removing `_out` tables from the metadata.yaml file is not the way to go! Using the [`hidden` argument in the `metadata.yaml` file](https://docs.datasette.io/en/stable/metadata.html#hiding-tables) is the way.

src/pudl/metadata/classes.py
```diff
@@ -1863,22 +1863,6 @@ def to_sql(
...
-        docs: output, core, intermediate.
-
-        In the future we might want to have more fine grain control over how
-        Resources are sorted.
-
-        Returns:
-            A sorted list of resources.
-        """
-        return sorted(self.resources, key=lambda r: r.name, reverse=True)
-
```
bendnorman, line 1881: I added this a while back hoping it would resolve the datasette sorting issue. It does not and I'm going to tackle it in a separate PR.

devtools/datasette/publish.py
```diff
@@ -82,9 +83,78 @@ def inspect_data(datasets: list[str], pudl_out: Path) -> str:
...
+
+def metadata(
+    pudl_out: Path,
+    datasets: list[str],
+) -> str:
     """Return human-readable metadata for Datasette."""
-    return DatasetteMetadata.from_data_source_ids(pudl_out).to_yaml()
+    metadata = DatasetteMetadata.from_data_source_ids(pudl_out)
+
+    # DBF databases do not have any metadata
```
bendnorman, line 150: Just want to double check the raw DBF databases do not have metadata.

devtools/datasette/publish.py
```diff
@@ -82,9 +83,78 @@ def inspect_data(datasets: list[str], pudl_out: Path) -> str:
     return inspect_output
 
 
-def metadata(pudl_out: Path) -> str:
+def check_tables_have_metadata(
```
bendnorman, line 86: This checks all tables in the databases have metadata. This feels like a check that should be available in datasette. I left a [question on the datasette github](https://github.com/simonw/datasette/discussions/2233).

src/pudl/metadata/classes.py
bendnorman: I added a test in `devtools/datasette/publish.py`.

devtools/datasette/publish.py
zaneselvans: They super duper do not have metadata. 🗑️

devtools/datasette/publish.py
```diff
@@ -82,9 +83,78 @@ def inspect_data(datasets: list[str], pudl_out: Path) -> str:
...
+) -> None:
+    """Check to make sure all tables in the databases have metadata.
+
+    This function fails if there are tables present in a database
+    that do not have structure metadata.
+
+    Args:
+        metadata: The structure metadata for the datasette deployment
+        pudl_out: The directory that contains the pudl outputs
+        datasets: The list of datasets to test.
```
zaneselvans, line 99: * What is a "dataset" here? What set of strings are you referring to? Is this like `eia923` and `epacems` or something else?
* Can we not use `pudl_out` for anything else for a while, until the memory of it has faded?

devtools/datasette/publish.py
zaneselvans: I think we should add this test to `test/integration/datasette_metadata_test.py` so that we catch new errors in CI rather than only in the nightly builds. Is there some reason that won't work? Maybe create a helper function that can be used both in the tests and here.

devtools/datasette/publish.py
```diff
@@ -82,9 +83,78 @@ def inspect_data(datasets: list[str], pudl_out: Path) -> str:
...
+
+        resources_in_database = {table_name[0] for table_name in cursor.fetchall()}
+        if dataset_exceptions:
+            resources_in_database = resources_in_database - dataset_exceptions
+        resources_with_metadata = {
+            resource.name for resource in resources[dataset_name]
+        }
+
+        cursor.close()
+        connection.close()
```
zaneselvans, line 128: Can we use a context manager instead of needing to close these connections?

devtools/datasette/publish.py
```diff
@@ -82,9 +83,78 @@ def inspect_data(datasets: list[str], pudl_out: Path) -> str:
...
+    """
+    resources = {}
+    resources |= metadata.xbrl_resources
+    resources["pudl"] = metadata.resources
+
+    resource_exceptions = {"pudl": {"alembic_version"}}
+
+    tables_missing_metadata_results = {}
+
+    for dataset in datasets:
```
zaneselvans, line 109: I'm having a hard time following here because I don't know what "dataset" refers to.  It seems like maybe it's actually a database, not a dataset?

src/pudl/metadata/classes.py
zaneselvans: I think this test would be more useful in the integration tests if it's easy to put it there instead. Otherwise we'll only discover that we've broken something in the nightly builds rather than in CI.

zaneselvans (changes requested):

devtools/datasette/publish.py
bendnorman: - Yes! "dataset" is confusing. Will update to databases.
- What do you mean by `pudl_out`? Are we trying to remove it?

devtools/datasette/publish.py
bendnorman: Yes there is a clean way to grab all tables using sqlalchemy.

devtools/datasette/publish.py
bendnorman: I'll update the names to make it clearer.

devtools/datasette/publish.py
bendnorman: That makes sense. I'll make the change.

devtools/datasette/publish.py
zaneselvans: Ah sorry about the confusion.

For years, we have used `pudl_out` as the name for a `PudlTabl` output object that's passed around everywhere, so re-using that name while `PudlTabl` is still floating around feels confusing/distracting to me. So I would recommend calling it `pudl_output` instead, which is the same as the `$PUDL_OUTPUT` environment variable which typically defines it now.

bendnorman: Oh I see. Thanks!

devtools/datasette/publish.py
```diff
@@ -82,9 +83,75 @@ def inspect_data(datasets: list[str], pudl_out: Path) -> str:
...
+    tables_missing_metadata_results = {}
+
+    for database in databases:
+        database_path = pudl_out / database
+        database_name = database_path.stem
+
+        # Grab all tables in the database
+        engine = sa.create_engine(f"sqlite:///{str(database_path)}")
+        inspector = sa.inspect(engine)
+        tables_in_database = set(inspector.get_table_names())
```
zaneselvans, line 116: This is a fairly large new entanglement -- previously we did not need to have the actual databases to generate the `metadata.yml`. Now it can **only** be produced if you've already run the ETL and have the DBs laying around. Of course that's fine for the actual deployment, but is it really necessary? It will make testing or generating the metadata in any other non-deployment environment annoying or impossible.

bendnorman: That makes sense. I can move check_tables_have_metadata()` so that `publish.py` can create metadata without actually needing complete databases.

devtools/datasette/publish.py
```diff
@@ -123,24 +118,27 @@ def deploy_datasette(deploy: str, fly_args: tuple[str]) -> int:
...
         ["pudl.sqlite"]
-        + sorted(str(p.name) for p in pudl_out.glob("ferc*.sqlite"))
+        + sorted(str(p.name) for p in pudl_output.glob("ferc*.sqlite"))
         + ["censusdp1tract.sqlite"]
     )
-
+    # Make sure we have the expected metadata for databases
+    # headed to deployment.
+    if deploy != "metadata":
+        check_tables_have_metadata(metadata_yml, databases)
```
bendnorman, line 132: I think this solves the entanglement issue you mentioned @zaneselvans . We only check if all of the databases tables have datasette metadata if we are actually deploy datatsette.

src/pudl/helpers.py
```diff
@@ -1918,3 +1923,84 @@ def diff_wide_tables(
...
+        metadata_yml: datasette metadata as yml.
+
+    Returns:
+        Parsed datasette metadata as JSON.
+    """
+    metadata_json = json.dumps(yaml.safe_load(metadata_yml))
+    return datasette.utils.parse_metadata(metadata_json)
+
+
+def check_tables_have_metadata(
```
bendnorman, line 1952: This check gets run when we deploy to datasette using `devtools/datasette/publish.py` and after the ETL is run in the integration tests. This way we'll catch errors in the CI and during datasette deployments.

src/pudl/helpers.py
```diff
@@ -1918,3 +1923,84 @@ def diff_wide_tables(
     return TableDiff(
         deleted=deleted, added=added, changed=changed, old_df=old, new_df=new
     )
+
+
+def create_datasette_metadata_yaml() -> str:
```
bendnorman, line 1928: I made the datasette metadata generation a helper function so parts of the code base that generate datasette metadata are all working with the same logic.

devtools/datasette/publish.py
zaneselvans: The way the Datasette metadata is generated, and the checks rely on bespoke logic for constructing paths to the databases we expect to have metadata, and the dependence on the XBRL datapackage.json sometimes and the SQLite files other times and the PUDL Package other times all feels pretty fragile and janky, but that's beyond the scope of this PR.

src/pudl/helpers.py
zaneselvans: Great, I think bringing as much (maybe all...) of our devtools / scripty stuff under the main package hierarchy is probably a good idea. It's starting to grow over there.

src/pudl/helpers.py
```diff
@@ -1918,3 +1923,84 @@ def diff_wide_tables(
...
+    """Create datasette metadata yaml.
+
+    Returns:
+        Datasette metadata for all PUDL resources as YAML
```
zaneselvans, line 1932: Is it just metadata for the PUDL resources? Or is it for the PUDL Resources and also all of the XBRL databases, as derived from their datapackage.json files too?

src/pudl/metadata/classes.py
```diff
@@ -1872,14 +1872,24 @@ def get_sorted_resources(self) -> StrictList(Resource):
...
         Resources are sorted.
 
         Returns:
             A sorted list of resources.
         """
-        return sorted(self.resources, key=lambda r: r.name, reverse=True)
+        resources = self.resources
+
+        def sort_resource_names(resource: Resource):
+            pattern = re.compile(r"(_out_|out_|core_)")
```
zaneselvans, line 1884: I think this regex should specify that these are the beginnings of the table names, otherwise we might get some inadvertent matches with tables like `core_eia__flout_the_authority` or out_ferc1__hardcore_energy`
```suggestion
            pattern = re.compile(r"^(_out_|out_|core_)")
```

src/pudl/metadata/classes.py
zaneselvans: I think this test is now being run both when we deploy and in the integration tests.

test/integration/datasette_metadata_test.py
```diff
@@ -56,3 +57,17 @@ def test_datasette_metadata_to_yml(ferc1_engine_xbrl):
...
+def test_database_metadata(
+    pudl_engine: sa.Engine,
+):
+    """Test to make sure all of the tables in the databases have metadata."""
+    pudl_output = PudlPaths().pudl_output
+    metadata_yml = create_datasette_metadata_yaml()
+    databases = (
+        ["pudl.sqlite"]
+        + sorted(str(p.name) for p in pudl_output.glob("ferc*.sqlite"))
+        + ["censusdp1tract.sqlite"]
```
zaneselvans, line 71: And it's okay to check all of `*.sqlite` because actually internally `check_tables_have_metadata()` has the secret knowledge of which databases we expect to have metadata?

src/pudl/helpers.py
```diff
@@ -1918,3 +1923,84 @@ def diff_wide_tables(
...
+
+def check_tables_have_metadata(
+    metadata_yml: str,
+    databases: list[str],
+) -> None:
+    """Check to make sure all tables in the databases have metadata.
+
+    This function fails if there are tables present in a database
+    that do not have structure metadata.
+
```
zaneselvans, line 1960:
```suggestion
    This function fails if there are tables lacking Datasette metadata in one of the
    databases we expect to have that kind of metadata. Note that we currently do
    not have this kind of metadata for the FERC databases derived from DBF or the
    Census DP1.

```

zaneselvans (changes requested):
Added suggestions for a docstring and a more restrictive table name regex for sorting.

src/pudl/helpers.py
bendnorman: It's both! I'll update the doc string.

test/integration/datasette_metadata_test.py
bendnorman: Yup! Maybe not the best place to store this information but seems good enough for now.

zaneselvans (approved):

Issue #2842, opened 2023-09-10
Update PUDL to use Pydantic 2


zaneselvans:
We can't switch over to using Pydantic 2 until [Dagster supports it](https://github.com/dagster-io/dagster/issues/15162) but I wanted a place to collect tasks for once it's supported.

```[tasklist]
- [x] Run the [`bump-pydantic`](https://github.com/pydantic/bump-pydantic) tool.
- [x] Fix doctest failure in `pudl.metadata.classes._validator` or obsolete the need for this custom validator getter.
- [x] `pytest test/unit/transform/classes_test.py::test_multicol_transform_factory` gets dtype mismatches
- [x] Fix or work around failing `PudlPaths()` validations when input/output dirs are None
- [x] Replace all instances of `model.dict()` with `model.model_dump()`
- [x] Replace all instances of the within-model `Config` classes with the `config = ConfigDict` attribute.
- [x] Remove unnecessary `fields_` and `schema_` aliases b/c v2 uses `model_fields` and `model_schema`.
- [x] Use pydantic strict types directly instead of aliasing in `pudl.metadata.classes`
- [x] Replace all instances of `update_forward_refs` with `model_rebuild()`
- [x] Use `re.Pattern` and `Annotated` types for the simplest models in `pudl.metadata.classes`
- [x] replace use of `__fields_set__` with `model_fields_set`
- [x] replace use of `__fields__` with `model_fields`
- [x] Replace all instances of `@validator` with `@field_validator`
- [x] Suppress warning about shadowing deprecated `BaseModel.schema` attribute
- [x] Fix `calc_cols` + `parent_cols` validation errors in exploded tables
- [x] Replace all instances of `@root_validator` with `@model_validator`
- [x] Add `extra`, `validate_all`, and `validate_assignment` back to metadata class configs.
- [x] Update release notes.
```
```[tasklist]
### Issues to Create
- [ ] Use simple `Annotated` types to validate single-element FERC 1 `TransformParams`.
- [x] Update `ferc-xbrl-extractor` to use Pydantic v2
- [x] Update `pudl-archiver` to use Pydantic v2
```

zaneselvans:
I'm currently getting an error in the conversion of our settings classes into Dagster configs, I think because of the taxonomy URL being part of our settings.  But that element is getting a type of `Any` in the conversion right now, so not sure why Dagster is still finding it to be a URL:

```
TypeError: Object of type Url is not JSON serializable
```

[...]

PR #3232, opened 2024-01-10
Add basic parquet export capability to PudlSqliteIOManager

rousik:
This is an experimental support for emitting data in parquet format. It is currently controlled via env variables `PUDL_WRITE_TO_PARQUET` and `PUDL_READ_FROM_PARQUET` and both these modes are currently disabled.

### DIFF
--- a/src/pudl/io_managers.py
+++ b/src/pudl/io_managers.py
@@ -9,6 +9,7 @@ import dask.dataframe as dd
 import pandas as pd
 import pyarrow as pa
 import sqlalchemy as sa
+import pyarrow.parquet as pq
 from alembic.autogenerate.api import compare_metadata
 from alembic.migration import MigrationContext
 from dagster import (
@@ -23,6 +24,7 @@ from dagster import (
 from packaging import version
 from sqlalchemy.exc import SQLAlchemyError
 from upath import UPath
+from pydantic_settings import BaseSettings, SettingsConfigDict
 
 import pudl
 from pudl.metadata.classes import Package, Resource
@@ -83,6 +85,27 @@ class ForeignKeyErrors(SQLAlchemyError):  # noqa: N818
     def __getitem__(self, idx):
         """Index the fk errors."""
         return self.fk_errors[idx]
+
+
+class ParquetSettings(BaseSettings):
+    """Controls how parquet files are used by pudl infrastructure.
+
+    This settings structure is used to control how parquet file formats are integrated with the PUDL
+    ETL pipeline. In addition to the original sqlite format, we may emit data into parquet files, using
+    pyarrow with schemas. In addition, we may also choose, whether sqlite or parquet should be treated
+    as the source of truth for subsequent data reads. I.e. when subsequent steps need to read the data,
+    we can control whether this data is read from sqlite or parquet files.
+
+    Eventually, we may also envision this mechanism to be used to suppress the use of sqlite files
+    altogether, if needed.
+    """
+    model_config = SettingsConfigDict(env_prefix='pudl_')
+
+    write_to_parquet: bool = False
+    """If True, data will be also written to parquet files into PUDL_OUTPUT/parquet/{db_name}/{table_name}.pq"""
+
+    read_from_parquet: bool = False
+    """If True, data will be read from parquet files instead of being read from sqlite tables."""
 
 
 class SQLiteIOManager(IOManager):
@@ -278,7 +301,6 @@ class SQLiteIOManager(IOManager):
         """
         table_name = self._get_table_name(context)
         sa_table = self._get_sqlalchemy_table(table_name)
-
         column_difference = set(sa_table.columns.keys()) - set(df.columns)
         if column_difference:
             raise ValueError(
@@ -377,90 +399,15 @@ class SQLiteIOManager(IOManager):
             return df
 
 
-# TODO(rousik): we can add env variable that will redirect read/writes between sqlite and parquet
-# formats. Alternatively, we could also just do this choice in the resource instantiation and use
-# a different class altogether for this.
-
-class PudlParquetIOManager(IOManager):
-    base_dir: str
-
-    def __init__(self, base_dir: str):
-        """Initializes PUDL parquet IO manager.
-
-        Creates base_dir where parquet files will be written.
-        """
-        self.base_dir = base_dir
-        Path(self.base_dir).mkdir(parents=True, exist_ok=True)
-
-    def _get_table_name(self, context) -> str:
-        """Get asset name from dagster context object."""
-        if context.has_asset_key:
-            table_name = context.asset_key.to_python_identifier()
-        else:
-            table_name = context.get_identifier()
-        return table_name
-
-    def handle_output(self, context: OutputContext, obj: pd.DataFrame | str):
-        """Handle an op or asset output.
-
-        If the output is a dataframe, write it to the database. If it is a string
-        execute it as a SQL query.
-
-        Args:
-            context: dagster keyword that provides access output information like asset
-                name.
-            obj: a sql query or dataframe to add to the database.
-
-        Raises:
-            Exception: if an asset or op returns an unsupported datatype.
-        """
-        if isinstance(obj, str):
-            raise Exception(
-                f"PudlParquetIOManager only supports pandas DataFrames, got: {str(obj)}"
-            )
-        assert isinstance(obj, pd.DataFrame)
-        table_name = self._get_table_name(context)
-        parquet_path = Path(self.base_dir) / f"{table_name}.parquet"
-        row_count = len(obj)
-        context.log.info(f"Row count for table {table_name}: {row_count}")
-        obj.to_parquet(path=parquet_path.as_posix(), index=False)
-
-    def load_input(self, context: InputContext) -> pd.DataFrame:
-        """Load a dataframe from a sqlite database.
-
-        Args:
-            context: dagster keyword that provides access output information like asset
-                name.
-        """
-        table_name = self._get_table_name(context)
-
-        # Check if there is a Resource in self.package for table_name
-        try:
-            res = self.package.get_resource(table_name)
-        except ValueError:
-            raise ValueError(
-                f"{table_name} does not appear in pudl.metadata.resources. "
-                "Check for typos, or add the table to the metadata and recreate the "
-                f"PUDL SQlite database. It's also possible that {table_name} is one of "
-                "the tables that does not get loaded into the PUDL SQLite DB because "
-                "it's a work in progress or is distributed in Apache Parquet format."
-            )
-        df = res.enforce_schema(
-            pd.read_parquet(path=(self.base_dir / f"{table_name}.parquet").as_posix())
-        )
-        if df.empty:
-            raise AssertionError(
-                f"The {table_name} table is empty. Materialize the {table_name} "
-                "asset so it is available in the database."
-            )
-        return df
-
 class PudlSQLiteIOManager(SQLiteIOManager):
     """IO Manager that writes and retrieves dataframes from a SQLite database.
 
     This class extends the SQLiteIOManager class to manage database metadata and dtypes
     using the :class:`pudl.metadata.classes.Package` class.
     """
+    # TODO(rousik): now that this experimentally supports also writing to parquet
+    # as an alternative storage format, we should probably rename this class
+    # to be less sqlite-centric.
 
     def __init__(
         self,
@@ -468,6 +415,7 @@ class PudlSQLiteIOManager(SQLiteIOManager):
         db_name: str,
         package: Package | None = None,
         timeout: float = 1_000.0,
+        settings: ParquetSettings = ParquetSettings(),
     ):
         """Initialize PudlSQLiteIOManager.
 
@@ -494,12 +442,14 @@ class PudlSQLiteIOManager(SQLiteIOManager):
                 exception, if the database is locked by another connection.  If another
                 connection opens a transaction to modify the database, it will be locked
                 until that transaction is committed.
-            use_parquet: if True, instead of actually writing to sqlite databases, we will
-                use parquet format.
+            settings: controls how parquet files are used by the PUDL ETL, in particular,
+                whether they should be used for input/output of dataframes.
         """
         if package is None:
             package = Package.from_resource_ids()
         self.package = package
+        self.settings = settings
+        logger.info(f"Instantiated pudl io manager with settings: {settings}")
         md = self.package.to_sql()
         sqlite_path = Path(base_dir) / f"{db_name}.sqlite"
         if not sqlite_path.exists():
@@ -507,7 +457,7 @@ class PudlSQLiteIOManager(SQLiteIOManager):
                 f"{sqlite_path} not initialized! Run `alembic upgrade head`."
             )
 
-        super().__init__(base_dir, db_name, md, timeout, use_parquet=use_parquet)
+        super().__init__(base_dir, db_name, md, timeout)
 
         existing_schema_context = MigrationContext.configure(self.engine.connect())
         metadata_diff = compare_metadata(existing_schema_context, self.md)
@@ -559,6 +509,20 @@ class PudlSQLiteIOManager(SQLiteIOManager):
 
         df = res.enforce_schema(df)
 
+
+        if self.settings.write_to_parquet:
+            logger.info(f"Writing {self.db_name}/{table_name} to parquet.")
+            parquet_path = PudlPaths().parquet_path(self.db_name, table_name)
+            parquet_path.parent.mkdir(parents=True, exist_ok=True)
+            schema = res.to_pyarrow()
+            with pq.ParquetWriter(
+                where=parquet_path,
+                schema=schema,
+                compression="snappy",
+                version="2.6",
+            ) as writer:
+                writer.write_table(df)
+
         with self.engine.begin() as con:
             # Remove old table records before loading to db
             con.execute(sa_table.delete())
@@ -592,6 +556,13 @@ class PudlSQLiteIOManager(SQLiteIOManager):
                 "the tables that does not get loaded into the PUDL SQLite DB because "
                 "it's a work in progress or is distributed in Apache Parquet format."
             )
+        if self.settings.read_from_parquet:
+            logger.info(f"Reading table {self.db_name}/{table_name} from parquet.")
+            schema = res.to_pyarrow()
+            return pq.read_table(
+                source=PudlPaths().parquet_path(self.db_name, table_name),
+                schema=schema,
+            ).to_pandas()
 
         with self.engine.begin() as con:
             try:
@@ -622,11 +593,6 @@ def pudl_sqlite_io_manager(init_context) -> PudlSQLiteIOManager:
     """Create a SQLiteManager dagster resource for the pudl database."""
     # TODO(rousik): this is now controlled with env variable, but that should be turned into
     # dagster configuration value.
-    if os.environ.get("PUDL_USE_PARQUET", "").lower() == "true":
-        logger.info("PudlIOManager: Using parquet")
-        return PudlParquetIOManager(base_dir=(PudlPaths().output_dir / "pudl").as_posix())
-
-    logger.info("PudlIOManager: using sqlite")
     return PudlSQLiteIOManager(base_dir=PudlPaths().output_dir, db_name="pudl")
 
 
@@ -659,6 +625,10 @@ class FercSQLiteIOManager(SQLiteIOManager):
                 connection opens a transaction to modify the database, it will be locked
                 until that transaction is committed.
         """
+        # TODO(rousik): Note that this is a bit of a partially implemented IO manager that
+        # is not actually used for writing anything. Given that this is derived from base
+        # SqliteIOManager, we do not support handling of parquet formats. This is probably
+        # okay for now.
         super().__init__(base_dir, db_name, md, timeout)
 
     def _setup_database(self, timeout: float = 1_000.0) -> sa.Engine:
--- a/src/pudl/io_managers.py
+++ b/src/pudl/io_managers.py
@@ -1,5 +1,6 @@
 """Dagster IO Managers."""
 import json
+import os
 import re
 from pathlib import Path
 from sqlite3 import sqlite_version
@@ -17,6 +18,7 @@ from dagster import (
     OutputContext,
     UPathIOManager,
     io_manager,
+    ConfigurableIOManager,
 )
 from packaging import version
 from sqlalchemy.exc import SQLAlchemyError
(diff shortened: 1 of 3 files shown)


rousik:
@zaneselvans once we actually support dual output formats (sqlite+parquet) it might be good time to do a little cleanup in the io manager codebase as this has been clearly designed with sqlite only in mind, and the class structure seems little clunky here, esp. when it comes to overlap/differences between pudl/ferc io managers (the latter being only partial and fixed to sqlite formats) and pudl/epacems (that might now share more due to both writing to parquet). I will think about how we could improve the situation, but this could get us going in the meantime.

src/pudl/io_managers.py
```diff
@@ -477,6 +509,21 @@ def _handle_pandas_output(self, context: OutputContext, df: pd.DataFrame):
 
         df = res.enforce_schema(df)
 
+        if self.settings.write_to_parquet:
+            logger.info(f"Writing {self.db_name}/{table_name} to parquet.")
+            parquet_path = PudlPaths().parquet_path(self.db_name, table_name)
+            parquet_path.parent.mkdir(parents=True, exist_ok=True)
+            schema = res.to_pyarrow()
```
zaneselvans, line 516: Along with the schema, another resource-specific parameter we might want in here is information about how to organize the row groups for larger assets, which I guess could be another method we add to `Resource`

src/pudl/io_managers.py
```diff
@@ -510,6 +557,13 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
                 "the tables that does not get loaded into the PUDL SQLite DB because "
                 "it's a work in progress or is distributed in Apache Parquet format."
             )
+        if self.settings.read_from_parquet:
+            logger.info(f"Reading table {self.db_name}/{table_name} from parquet.")
+            schema = res.to_pyarrow()
+            return pq.read_table(
+                source=PudlPaths().parquet_path(self.db_name, table_name),
+                schema=schema,
+            ).to_pandas()
```
zaneselvans, line 566: I don't think this pandas -> parquet -> pandas round-trip will necessarily preserve dtypes, since we're not using pyarrow dtypes in our dataframes, so I suspect we'll still need to`res.enforce_schema()` here as we do when reading out of SQLite below to avoid unexpected dtype problems downstream.

zaneselvans (changes requested):
I'm fuzzy on where the configuration **should** go, but I'm pretty sure that Dagster already has a place where it expects to receive this kind of resource configuration, including via environment variables. But maybe @bendnorman knows off the top of his head.

It feels like it would be cleaner to keep the SQLite IO Manager dedicated to SQLite, and have another IO Manager that is dedicated to the Parquet logic, and then compose them together into a hybrid that can do both / either depending on how it's configured. Does that seem reasonable?

src/pudl/io_managers.py
```diff
@@ -382,12 +405,17 @@ class PudlSQLiteIOManager(SQLiteIOManager):
...
+    # as an alternative storage format, we should probably rename this class
+    # to be less sqlite-centric.
+
     def __init__(
         self,
         base_dir: str,
         db_name: str,
         package: Package | None = None,
         timeout: float = 1_000.0,
+        settings: ParquetSettings = ParquetSettings(),
```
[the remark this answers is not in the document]
rousik: I'm now attempting to switch to dagster configuration system, but it seems to me that `EnvVar` that dagster has is kind of crude compared to pydantic. They only support str/int types for env variables and AFAIK there's no easy way to say what the default should be when the env variable is not actually set. This seems really strange to me and I'm still investigating what the options are here.

src/pudl/io_managers.py
```diff
@@ -382,12 +405,17 @@ class PudlSQLiteIOManager(SQLiteIOManager):
     using the :class:`pudl.metadata.classes.Package` class.
     """
 
+    # TODO(rousik): now that this experimentally supports also writing to parquet
+    # as an alternative storage format, we should probably rename this class
+    # to be less sqlite-centric.
+
```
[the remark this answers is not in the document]
rousik: I dug deeper into how we deal with io_managers and I'm a bit perplexed by the apparent complexity here. There are several partial implementations, some of which support reads, some of which only writes, and there's more duplicate code than ideally should be, but I think this is a problem for a little later and shouldn't be necessarily solved as part of this.

After some more tinkering, I think that the idea of composite IO manager that simply hands-off reads/writes to either sqlite or parquet should work well and should be relatively clean too.

The only remaining snag is that dagster support for loading configuration from env variables seems kind of... incomplete... in that there doesn't seem to be support for booleans or for supplying default value for when the env variable is not set. 

I was expecting something along the lines of 1. use built-in default, 2. replace this with env-variable if set, and 3. replace env default with custom value in case it is supplied via dagster ui/job configuration, but either I'm misreading or this is not actually supported?!

rousik:
Moving the conversation ahead a notch. The modified code that addresses some of the issues outlined here is still on my local machine and need to be cleaned up before publishing - this is pending open issue with dagster `EnvVar`.

rousik:
I've refactored IO managers so that there's sqlite, parquet and a combined one. Reading from env variables is kind of crummy, because dagster has much worse support for this than pydantic `BaseSettings`, but I don't think this is a major problem. parquet writes/reads will be enabled whenever the associated env variables are non-empty (i.e. truth value is based on strings, not interpreted as boolean, which could lead to some confusion).

I feel like there's still some work to be done to clean up things, and, ideally we would also use the same parquet io manager for epacems, but the presence of sharding will require some custom logic.

pyarrow parquet writer does support sharding, but the naming schema is usually `base_dir/key=value/{uuid}.parquet` i.e. in our case it would be `epacems/state=CA/...` which may not be what we want.

src/pudl/io_managers.py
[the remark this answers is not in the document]
zaneselvans: Among us I think @bendnorman knows the most about how Dagster resource configuration is supposed to work. I only have a vague understanding, and am just hoping that we can avoid a proliferation of configuration systems, and ideally use whatever the canonical system is for the framework we've adopted.

src/pudl/io_managers.py
zaneselvans: This doesn't currently exist. The only big table we've written out to Parquet historically is the EPA CEMS hourly emissions, and that's been done in a bespoke way.

zaneselvans:
@rousik It looks like it's not actually able to write to the DB because it has no connection engine?

rousik:
Oh that would be tests reaching into the IO manager and assuming it's internal structure. I'll fix later today and when at the keyboard.

ETL itself should be working.

src/pudl/io_managers.py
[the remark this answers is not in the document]
bendnorman: Do we have to use an env var to configure the `PudlMixedFormatIOManager`? I think just using vanilla boolean attributes would be fine:

```python
write_to_parquet: bool = False
"""If true, data will be written to parquet files."""

read_from_parquet: bool = False
"""If true, data will be read from parquet files instead of sqlite."""
```

We could create a `etl_full_gcp_build` that configures the `PudlMixedFormatIOManager` to write and read from parquet.

You can also set a default value for a `EnvVar` using [`EnvVar.get_value()`](https://github.com/dagster-io/dagster/blob/9aa58e22e68e8455573cc04c7fb6a3448ce933c5/python_modules/dagster/dagster/_config/field_utils.py#L564C16-L568). We could do something like this:

```python
write_to_parquet: bool = bool(EnvVar("PUDL_WRITE_TO_PARQUET").get_value(default=True))
```

Not the most elegant thing ever but I tested it out and it allows you to have a default, read from an env var and overwrite the value in the UI. 

There is an [`IntEnvVar`](https://github.com/dagster-io/dagster/blob/9aa58e22e68e8455573cc04c7fb6a3448ce933c5/python_modules/dagster/dagster/_config/field_utils.py#L502) that casts env var values to ints. Maybe we could make one for booleans?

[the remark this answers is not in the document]
bendnorman: I think my preference is for `PudlMixedFormatIOManager` to be a subclass of `IOManager` and to configure it using the `@io_manager(config_schema=...)` decorator. [These are the legacy concepts](https://docs.dagster.io/concepts/io-management/io-managers-legacy) but I don't think it's great to have our IO managers use a mix of the legacy and current resource concepts.

Eventually we should move to using the new pydantic configurable resources. We explored migrating our setting configurations in https://github.com/catalyst-cooperative/pudl/issues/2842 but decided not to because it was a bigger change than we expected. We could create separate issues for migrating the resources in `pudl.resources` and `pudl.io_managers`.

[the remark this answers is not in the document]
zaneselvans: Ah okay, so there *is* a way to do all this configuration using the more ergonomic Pydantic settings system. But we would need to do some migration work to get the system there.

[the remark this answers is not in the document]
rousik: @bendnorman `ConfigurableIOManager` is a subclass of `IOManager` and works just fine with the old-style `@io_manager` annotated method. The old-style approach just hands out instance of this io manager to whoever needs it, so I don't think there's a practical problem with how it's done now. Or maybe I'm missing something.

Using `ConfigurableIOManager` will AFAIK make the migration easier once we're ready for it because this is basically already prepared for the new world.

src/pudl/io_managers.py
rousik: Yeah, I think we can encode this in the resource, or in the asset metadata once we're ready for it. The code should be flexible enough to accommodate that once we need it.

src/pudl/io_managers.py
```diff
@@ -536,9 +638,19 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
 
 
 @io_manager
-def pudl_sqlite_io_manager(init_context) -> PudlSQLiteIOManager:
+def pudl_sqlite_io_manager(init_context) -> IOManager:
```
rousik, line 641: @bendnorman and @zaneselvans - given that this now returns mixed format IO manager, could we rename this resource/io-manager from `pudl_sqlite_io_manager` to plain `pudl_io_manager`? I think that would make it more appropriate. There's a bit of a risk that very widely used word `pudl` refers both to *the primary output database* (`pudl.sqlite` up to this point) as well as the overall project/ETL/codebase. This ambiguity may be okay, because I'm not quite sure what would be the other word we could substitute here. Maybe `main_output` or something along those lines?

bendnorman: I think `pudl_io_manager` is appropriate for this IO Manager. We're currently ok with the naming ambiguity of the code and data outputs. We explain the different parts of pudl in the [readme](https://github.com/catalyst-cooperative/pudl?tab=readme-ov-file#what-is-pudl).

zaneselvans: I think `pudl_io_manager` makes sense.

This is probably beyond the scope of this PR and maybe more of a @bendnorman thing but I think we should also change the way the `default_resources` are defined in `pudl.etl.__init__.py`:

```py
default_resources = {
    "datastore": datastore,
    "pudl_sqlite_io_manager": pudl_sqlite_io_manager,
    "ferc1_dbf_sqlite_io_manager": ferc1_dbf_sqlite_io_manager,
    "ferc1_xbrl_sqlite_io_manager": ferc1_xbrl_sqlite_io_manager,
    "dataset_settings": dataset_settings,
    "ferc_to_sqlite_settings": ferc_to_sqlite_settings,
    "epacems_io_manager": epacems_io_manager,
}
```

It seems like the keys in this dictionary should not specify anything about the format / destination of the data, while the values (the actual IO managers) should have that specificity.  So `pudl_io_manager` could be any kind of IO manager... but regardless it's the one that's going to be used to output PUDL data, whether it's going to SQLite, BigQuery, both SQLite + Parquet, etc.  And the value associated with it determines the details.

I found this very confusing in thinking about how we were going to switch over to writing the data to not just SQLite, since we've hard-coded `pudl_sqlite_io_manager` all throughout the system.  But it's not really an SQLite IO Manager. It could be any kind of IO Manager.  But I didn't realize they were separable until reviewing this PR.

The FERC IO managers seem like they have a similar issue, and also, what IO Managers are used to output all of the FERC Forms that aren't Form 1?

src/pudl/io_managers.py
```diff
@@ -83,6 +87,72 @@ def __getitem__(self, idx):
...
+        )
+        self._parquet_io_manager = PudlParquetIOManager()
+
+        # TODO(rousik): Move the following diagnostics to debug level once
+        # we know what is happening.
+        write_fmt = "parquet" if self.write_to_parquet else "sqlite"
+        read_fmt = "parquet" if self.read_from_parquet else "sqlite"
+        logger.warning(
+            f"pudl_io_manager will be reading from {read_fmt}, writing to {write_fmt}."
+        )
```
zaneselvans, line 141: There's an asymmetry here in that we can only read from one source, and it can be either SQLite or Parquet, but we're always writing to SQLite, and optionally also writing to Parquet. But the way it's phrased it seems like there's one read format, and one write format, which is misleading.

src/pudl/io_managers.py
bendnorman: I agree Zane the keys shouldn't specify anything about format / destination. 

The FERC IO managers are just for reading data out of the dbf and xbrl sqlite databases. We don't process any of the other ferc forms in the main PUDL ETL so we don't have IO managers for them.

zaneselvans (changes requested):
Just one minor issue with what we're logging for the read/write formats.

zaneselvans:
Also, don't we need to update the resource configuration in `pudl/etl/__init__.py`? Otherwise it won't actually use the new IO manager will it?

src/pudl/io_managers.py
zaneselvans: Also don't we need to tell it to use the new IO Manager here? Otherwise it won't actually generate any Parquet files will it?

src/pudl/io_managers.py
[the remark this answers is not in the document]
rousik: This is fair point, the duplication of the config both in the `ConfigurableIOManager` and `config_schema` here would be tedious.

src/pudl/io_managers.py
rousik: We are instantiating `PudlMixedFormatIOManager` here which uses either of the formats depending on the configuration.

src/pudl/io_managers.py
rousik: Right, I have tweaked the log message. We probably want to drop the severity also. I have used `warning` for local testing to see that it correctly picks up env var overrides.

src/pudl/io_managers.py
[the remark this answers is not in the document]
rousik: @bendnorman, fair point re/ use of `config_schema` here. I will make the necessary changes.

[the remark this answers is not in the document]
rousik: Though it's tempting to keep using `ConfigurableIOManager` as that takes care of building constructor for me that would be identical to what I'm about to write :-)

src/pudl/io_managers.py
```diff
@@ -83,6 +87,72 @@ def __getitem__(self, idx):
...
+    # TODO(rousik): the folowing EnvVars should really be booleans but
+    # dagster EnvVar doesn't recognize anything but str/int types here
+    # and AFAIK there's no way to set default value if the env variable
+    # is not set.
+
+    write_to_parquet: bool = bool(EnvVar("PUDL_WRITE_TO_PARQUET").get_value(False))
+    """If true, data will be written to parquet files."""
+
+    read_from_parquet: bool = bool(EnvVar("PUDL_READ_FROM_PARQUET").get_value(False))
+    """If true, data will be read from parquet files instead of sqlite."""
```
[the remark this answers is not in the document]
rousik: In short, practical concerns.

I'm treating this as a "feature flag", where we start by having the new functionality (writing to parquet files) off by default, but toggle-able on demand. While toggling this in the dagster UI is useful for local development, in many cases where we might want to test this (e.g. in the CI, nightly builds and other dockerized/remote scenarios), it's very easy to set env variable and pass it to the ETL, and it's much more difficult to be tweaking dagster configuration from the outside IMO. 

This could be hooked into command-line flag, but that is also somewhat clunky and would require ad-hoc wiring to be piped through while env variables are easy and widely supported (both by github actions as well as when running docker image on vm/batch), so it should be fairly easy to toggle this new functionality just where we need it.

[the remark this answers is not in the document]
rousik: If all goes well and this feature will end up being enabled everywhere and always, we can drop this toggle and hardcode the new behavior (always on). I'm unsure what would be the timeline for this, given the early stage this is in.

rousik:
Okay. I have made some more changes, specifically renamed `pudl_sqlite_io_manager` to a generic `pudl_io_manager`. 

I have left `fake_pudl_sqlite_io_manager` in place, as that seems to be essentially unit tests for the `PudlSQLiteIOManager` itself. Personally I think that switching to `unittest.TestCase` here would make things a little cleaner instead of passing around these fixtures, e.g. (based on https://docs.python.org/3/library/unittest.html#basic-example)

```
class PudlSQLiteIOManagerTests(unittest.TestCase):

  def setUp(self):
    # spin up temp database (in TemporaryDirectory perhaps)
    self.io_manager = PudlSQLiteIOManager(...)

  def test_delete_statement(self):
    self.io_manager.handle_output(...)
    ...
```

We're already using `unittest` style tests for excel so that would not be first case :)

test/integration/etl_test.py
```diff
@@ -33,7 +33,7 @@ def test_pudl_engine(
 
     if check_foreign_keys:
         # Raises ForeignKeyErrors if there are any
-        pudl_sql_io_manager.check_foreign_keys()
+        pudl_sql_io_manager._sqlite_io_manager.check_foreign_keys()
```
rousik, line 36: This is dirty, but I'm not sure what would be the best way around it. Maybe implement `MixedFormatIOManager.get_sqlite_io_manager() -> PudlSqliteIOManager` method that we could call when we really need to access sqlite only features?

rousik:
Looks like failing validation when there are unknown columns is not the right thing to do as demonstrated by the failing `ci-integration`. Perhaps the assumption in the test `test_extra_column_error` is itself wrong? Halp!

zaneselvans:
@rousik whenever you introduce additional frameworks or systems that we're not using (like `poetry` or `unittest`) it creates a barrier to our understanding, maintaining, and extending the work, or it means we need to dedicate additional internal resources to educating everyone on how these other systems work, or to replacing what you've done with the systems that we've already agreed to use internally.

src/pudl/io_managers.py
```diff
@@ -570,6 +626,10 @@ def __init__(
                 connection opens a transaction to modify the database, it will be locked
                 until that transaction is committed.
         """
+        # TODO(rousik): Note that this is a bit of a partially implemented IO manager that
+        # is not actually used for writing anything. Given that this is derived from base
+        # SqliteIOManager, we do not support handling of parquet formats. This is probably
+        # okay for now.
```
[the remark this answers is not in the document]
bendnorman: Could you add these thoughts to the doc string? The doc string already states the class should be subclassed though it could be a little more clear.

test/integration/etl_test.py
bendnorman: I think implementing a getter that throws a helpful error if the `MixedFormatIOManager` doesn't have a `sqlite_io_manager`.

It might not be worth the complexity but we could make the `MixedFormatIOManager` more generalizable and accept an arbitrary number of IO managers to use. Then we could have some methods for viewing and grabbing specific io managers:

```python
>>> MixedFormatIOManager.list_io_manager_names()
["sqlite_io_manager", "parquet_io_manager", "bq_io_manager", ...]
>>> MixedFormatIOManager.get_io_manager("sqlite_io_manager")
```

I think for now a simple `MixedFormatIOManager.get_sqlite_io_manager() -> PudlSqliteIOManager` method seems fine.

test/unit/io_managers_test.py
```diff
@@ -268,46 +99,174 @@ def test_migrations_match_metadata(tmp_path, monkeypatch):
...
-
-def test_error_when_reading_view_without_metadata(fake_pudl_sqlite_io_manager_fixture):
-    """Make sure and error is thrown when a user loads a view without metadata."""
-    asset_key = "track_view"
-    input_context = build_input_context(asset_key=AssetKey(asset_key))
-    with pytest.raises(ValueError):
-        fake_pudl_sqlite_io_manager_fixture.load_input(input_context)
+# TODO(rousik): is there difference between fake_sqlite_io_manager
+# and real sqlite_io_manager in terms of functionality?!!
+class PudlSQLiteIOManagerTest(unittest.TestCase):
```
bendnorman, line 104: Why are you using unittest instead of pytest here?

bendnorman (changes requested):

rousik:
> @rousik whenever you introduce additional frameworks or systems that we're not using (like `poetry` or `unittest`) it creates a barrier to our understanding, maintaining, and extending the work, or it means we need to dedicate additional internal resources to educating everyone on how these other systems work, or to replacing what you've done with the systems that we've already agreed to use internally.

This is a fair point. I agree that adding new technologies to the mix have non-trivial barriers and we shouldn't do this haphazardly, which is, perhaps, how it looks from the outside. Let me try to provide a bit more detailed reasoning for `unittest`:

1. `unittest` style is already used in excel, so we're not strictly introducing it from scratch. It was me who added it back then, so it's still on me though :)
2. `unittest` plays well in `pytest` in that pytest can run unittest based tests just fine, so it boils down to different "style" of writing tests, but the overall usage remains unchanged and unaffected.
3. It feels to me that unittest offers a little more encapsulated and nicer ways to build tests, which is particularly nice for situations where we are testing specific classes and their expectations (i.e. rather than having flat collection of tests, we can group them into test-classes that are geared towards specific functionality) and provide very explicit setup/teardown logic rather than relying on the nebulous cloud of fixtures.

It is, in particular, the point (3) where I think that departing from `pytest` might result in cleaner and leaner looking test and it might very well be worth it to use that. I do think that we're also using fairly limited and intuitive features from `unittest` so it shouldn't be too hard to maintain those tests, but perhaps I'm underestimating the burden, and overestimating the lower maintenance of these "nicer" tests.

I can roll these changes back if this is undesirable.

I think we can have separate conversation re/ `poetry` - there, the use is definitely more experimental, but also non-binding in that it should be possible to revert back/switch to other dep management systems.

test/unit/io_managers_test.py
rousik: See other threads on the PR. I thought that `unitest` offers generally cleaner tests with less verbosity. However, if this breaks the convention (most of the code does use `pytest` even though there are some `unitest` based tests too), I can revert these changes.

rousik: I do like the ability to put all of the assumptions/expectations about single class-under-test into a parallel unittest test-class, as this provides nice encapsulation and visual separation compared to the flat undifferentiated pytest style approach, but if we think consistency is more valuable (pytest vs unittest) I can revert these last changes.

test/integration/etl_test.py
rousik: I think the generalization here is probably not warranted. I would only add that if/when we know we will need it but otherwise it may be just future proofing for something that will never happen and I agree that the complexity is not worth it.

What this does show, however, is that sqlite and pudl-sqlite io managers are special in that their APIs and functionality is much broader than that of the typical/format-agnostic io manager and this API is a bit vague.

It might be valuable to separate these concerns somewhat, perhaps by defining the interface that such sqlite thingy should provide and have a method on the mixed io manager that can return implementations of this clearly-defined and separate interface. It could return the io_manager that implements this interface, but in most cases I see, we don't actually need the io_manager but the specific sqlite validation/schema-management functionality.

It might even make sense to separate this sqlite/schema functionality to a standalone class that is embedded in the IO manager but with which IO manager communicated through this well defined API.

Note that the above may very well go past the scope of this PR and should be likely reserved to subsequent cleanup/refactoring. Providing `get_sqlite_io_manager()` function is simple and deals with this dirtiness reasonably well.

src/pudl/io_managers.py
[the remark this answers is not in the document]
zaneselvans: If reading from Parquet ends up significantly speeding up the ETL it's easy to imagine this getting turned on immediately.

Once it's there, I'd experiment with using it to speed up the integration tests and data validations (which would mean relying on SQLite primarily for bulk distribution, validation of referential integrity, and maybe other schema checks / constraints that are only implemented by the `to_sqlite()` schema). Since we want to remove `PudlTabl` (probably in the Spring) and that will mean reorganizing the tests, that would probably be a good time to play around with it. I don't know exactly what fraction of the time in the integration tests and data validations is spent reading data, but I suspect it's significant given the way we've wired up `PudlTabl` to pass the read requests through to the database and not cache anything in memory, and I think reading

I can't remember where we were discussing it but I think @bendnorman suggested that since the PUDL ETL script is now a relatively thin wrapper around some slightly nonstandard Dagster structures, we might want to switch to using the Dagster CLI directly, which I think would make these job configurations more straightforward with the Dagster configs.

src/pudl/etl/check_foreign_keys.py
```diff
@@ -51,7 +51,12 @@ def pudl_check_fks(logfile: pathlib.Path, loglevel: str):
     pudl.logging_helpers.configure_root_logger(logfile=logfile, loglevel=loglevel)
 
     context = build_init_resource_context()
-    io_manager = pudl_sqlite_io_manager(context)
+    io_manager: PudlSQLiteIOManager = pudl_io_manager(context)._sqlite_io_manager
+    # TODO(rousik): This reaches into the io_manager and assumes
+    # PudlSQLiteIOManager and the corresponding foreign key
+    # functionality. It's strange to be constructing dagster
+    # resources to achieve this, but alas, that's where the logic
+    # lies.
```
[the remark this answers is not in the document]
rousik: We have the foreign key check on the schemas/Packages (do the foreign keys make sense at all), and right now we have the validation on the actual data, which is very much format specific (relies on sqlite pragmas). Building something format agnostic may be well out of scope of this PR, but nevertheless, extracting the functionality out of io_manager does actually make sense and I will do so. That will simplify things.

zschira:
Hey @rousik, we decided we want to try to get the parquet support completed by the end of the week, so I'm going to branch off of here and see if I can get a parallel PR through over the finish line. Thanks for getting this most of the way there, though!

rousik:
> Hey @rousik, we decided we want to try to get the parquet support completed by the end of the week, so I'm going to branch off of here and see if I can get a parallel PR through over the finish line. Thanks for getting this most of the way there, though!

Thanks, that is awesome. I'm off the grid most of this week so it's a great decision.

zaneselvans:
Parquet outputs have been merrrrged in #3296

PR #3234, opened 2024-01-11
fill in the annual columns with quarterly balances

### LINKED ISSUE #3233: why is the `in_rate_base` records from the liabilities table all null from 2005-2020?
For some reason all of the records labeled `in_rate_base` from 2005 and 2020 are null:

```python
df=defs.load_asset_value(AssetKey("exploded_income_statement_ferc1"))
tbl = df[
    (df.ending_balance.notnull())
    & (df.tags_in_rate_base == "yes")
]
plt.scatter(
    tbl.report_year,
    tbl.ending_balance,
    alpha=.2
)
```

Image

removing the `tags_in_rate_base=="yes"` mask, there is a ton of non-null data in the table..

i will first check if this is a low label issue...

cmgosnell:
# Overview

Closes #3233.

What did you change?
* simple replace of contents of annual column with quarterly column's content

# Testing

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/output/ferc1.py
+++ b/src/pudl/output/ferc1.py
@@ -110,7 +110,7 @@ EXPLOSION_CALCULATION_TOLERANCES: dict[str, GroupMetricChecks] = {
                 null_calculated_value_frequency=1.0,
             ),
             xbrl_factoid=MetricTolerances(
-                error_frequency=0.028,
+                error_frequency=0.035,
                 relative_error_magnitude=0.019,
                 null_calculated_value_frequency=1.0,
             ),
--- a/src/pudl/transform/ferc1.py
+++ b/src/pudl/transform/ferc1.py
@@ -4720,7 +4720,7 @@ class BalanceSheetLiabilitiesTableTransformer(Ferc1AbstractTableTransformer):
             .assign(utility_type="total")
         )
 
-    def process_dbf(self: Self, df: pd.DataFrame):
+    def process_dbf_test(self: Self, df: pd.DataFrame):
         """Standard dbf process plus converting quarterly data to annual.
 
         For some reason in the dbf data for this table reported all of the
@@ -4739,8 +4739,10 @@ class BalanceSheetLiabilitiesTableTransformer(Ferc1AbstractTableTransformer):
                 "Found non-null records, so the annual columns may no longer need to "
                 "be filled in with quarterly data."
             )
-        for col in annual_cols:
-            df.loc[bad_years_mask, col] = df.loc[bad_years_mask, f"{col}_qtr"]
+
+        df.loc[bad_years_mask, annual_cols] = df.loc[
+            bad_years_mask, ["pri_yr_q4_bal", "end_qtr_bal"]
+        ].to_numpy()
         return df
 
     @cache_df(key="main")
--- a/src/pudl/transform/ferc1.py
+++ b/src/pudl/transform/ferc1.py
@@ -4733,7 +4733,7 @@ class BalanceSheetLiabilitiesTableTransformer(Ferc1AbstractTableTransformer):
         annual_cols = ["starting_balance", "ending_balance"]
         bad_years_mask = df.report_year.between(2005, 2020)
         # ensure this filling in treatment is necessary!
-        if not df.loc[bad_years_mask, annual_cols].isnull().all().all():
+        if not df.loc[bad_years_mask, annual_cols].isnull().all(axis=None):
             raise AssertionError(
                 "We expected that all balance data between 2005 and 2020 are all null. "
                 "Found non-null records, so the annual columns may no longer need to "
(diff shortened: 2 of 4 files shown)


src/pudl/output/ferc1.py
```diff
@@ -110,7 +110,7 @@
                 null_calculated_value_frequency=1.0,
             ),
             xbrl_factoid=MetricTolerances(
-                error_frequency=0.028,
+                error_frequency=0.035,
                 relative_error_magnitude=0.019,
                 null_calculated_value_frequency=1.0,
             ),
```
cmgosnell, line 116: adding ... data!... into the liabilities table for many years, we the calc checks got a little worse for `retained_earnings`. Still pretty good compared to the other inter-table calcs.

src/pudl/transform/params/ferc1.py
```diff
@@ -3447,8 +3447,8 @@
                     "begin_yr_balance": "starting_balance",
                     "end_yr_balance": "ending_balance",
                     "report_prd": "report_prd",
-                    "end_qtr_bal": "end_qtr_bal",
-                    "pri_yr_q4_bal": "pri_yr_q4_bal",
+                    "end_qtr_bal": "ending_balance_qtr",
+                    "pri_yr_q4_bal": "starting_balance_qtr",
```
zaneselvans, line 3451: The renamed `pri_yr_q4_bal` here is a bit misleading, since it does not pertain to the quarter that's specified in the record (Q4 of this year). Maybe we could just not rename the columns and not iterate over the column names, and use the original names in a multi-column assignment (can you do that?) in the loc `process_dbf()`?

src/pudl/transform/ferc1.py
```diff
@@ -4720,6 +4720,29 @@ def process_xbrl_metadata(
...
+        bad_years_mask = df.report_year.between(2005, 2020)
+        # ensure this filling in treatment is necessary!
+        if not df.loc[bad_years_mask, annual_cols].isnull().all().all():
+            raise AssertionError(
+                "We expected that all balance data between 2005 and 2020 are all null. "
+                "Found non-null records, so the annual columns may no longer need to "
+                "be filled in with quarterly data."
+            )
+        for col in annual_cols:
+            df.loc[bad_years_mask, col] = df.loc[bad_years_mask, f"{col}_qtr"]
```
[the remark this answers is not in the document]
zaneselvans: To avoid iterating and renaming to somewhat misleading column names could we do:

```py
df.loc[bad_years_mask, annual_cols] = df.loc[bad_years_mask, ["end_qtr_bal", "pri_yr_q4_bal"]]
```

src/pudl/transform/ferc1.py
```diff
@@ -4720,6 +4720,29 @@ def process_xbrl_metadata(
...
+        balance data as quarterly data between 2005 and 2020. We already choose
+        the end of the year in :meth:`select_annual_rows_dbf`.
+
+        https://github.com/catalyst-cooperative/pudl/issues/3233
+        """
+        df = super().process_dbf(df)
+        annual_cols = ["starting_balance", "ending_balance"]
+        bad_years_mask = df.report_year.between(2005, 2020)
+        # ensure this filling in treatment is necessary!
+        if not df.loc[bad_years_mask, annual_cols].isnull().all().all():
```
zaneselvans, line 4736: Rather than the double-all to reduce all axes you can use:

```py
if not df.loc[bad_years_mask, annual_cols].isnull().all(axis=None):
```

zaneselvans (changes requested):

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
cmgosnell: on your first comment about prd 12 i noted this in the docstring for the method. we always apply `select_annual_rows_dbf` so that's always going to be the case.

and on your second point, I tried this. and apparently you can't replace two columns at a time like this. I'm really not sure why when it works just fine with one column at a time.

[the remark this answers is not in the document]
cmgosnell: oh lol you just have to add `.values` at the end

```python
df.loc[bad_years_mask, annual_cols] = df.loc[bad_years_mask, ["end_qtr_bal", "pri_yr_q4_bal"]].values
```

src/pudl/transform/params/ferc1.py
cmgosnell: yeah this felt bad.  `ending_balance_qtr` feels better tbh but these are both temporary so i will revert.

src/pudl/transform/ferc1.py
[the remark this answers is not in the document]
zaneselvans: Huh, interesting. Well glad something worked!

zaneselvans (approved):

src/pudl/transform/ferc1.py
```diff
@@ -4720,6 +4720,31 @@ def process_xbrl_metadata(
             .assign(utility_type="total")
         )
 
+    def process_dbf_test(self: Self, df: pd.DataFrame):
```
cmgosnell, line 4723: oooohmigosh i forgot to rename this! this hasn't been being run this whole time

PR #3238, opened 2024-01-12
Propose bug triage process + nightly build mgmt process

jdangerx:
# Overview

Sometimes stuff goes wrong! We need to make decisions about how urgently they need to be fixed! It might be worth writing down some guidelines - but maybe we can just use "discuss as a team how urgent this is" as a process for a little longer. In any case, I tried to sketch out what some written guidelines could look like.

Separately, we have a bit more of an actual process around the nightly build failures now, so I wrote that down.

# Testing

Well, other people need to read it and give feedback ;)

```[tasklist]
# To-do list
```

### DIFF
--- a/docs/dev/nightly_data_builds.rst
+++ b/docs/dev/nightly_data_builds.rst
@@ -14,12 +14,29 @@ pushes a Docker image with PUDL installed to `Docker Hub <https://hub.docker.com
 and deploys the image as a container to a GCE instance. The container runs the ETL and
 tests, then copies the outputs to a public AWS S3 bucket for distribution.
 
+
 Breaking the Builds
 -------------------
 The nightly data builds based on the ``main`` branch are our comprehensive integration
-tests. When they pass, we consider the results fit for public consumption.  The builds
-are expected to pass. If they don't then someone needs to take responsibility for
-getting them working again with some urgency.
+tests. When they pass, we consider the results fit for public consumption.
+
+When they don't pass, we need to fix them.
+
+Every morning, someone from inframundo will check the #pudl-deployments slack
+channel to see if the build has succeeded.
+
+If it has succeeded, they will track it in `this spreadsheet <https://docs.google.com/spreadsheets/d/11YdknGi4br51kxz03nNmxD-lOvilWFgm3jkUpiEdAzU/edit#gid=1678819446>`__.
+
+If it has failed, they will:
+
+1. Create a GitHub issue for the nightly build failure to hold the investigation & discussion.
+2. Track the GitHub issue and the build status in the above spreadsheet.
+3. Look in the logs and determine whether it was an "infrastructure failure," i.e. something went wrong with the code that *runs*
+the nightly build, or a "PUDL failure," i.e. something that went wrong with the PUDL ETL itself.
+4. Investigate the source of the issue & explore ways to fix it. Get help from the folks whose PRs broke the build.
+
+Avoiding breaking the builds
+----------------------------
 
 Because of how long the full build & tests take, we don't typically run them
 individually before merging every PR into ``main``. However, running ``make nuke``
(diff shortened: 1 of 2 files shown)


docs/dev/nightly_data_builds.rst
```diff
@@ -14,12 +14,29 @@ pushes a Docker image with PUDL installed to `Docker Hub <https://hub.docker.com
...
 -------------------
 The nightly data builds based on the ``main`` branch are our comprehensive integration
-tests. When they pass, we consider the results fit for public consumption.  The builds
-are expected to pass. If they don't then someone needs to take responsibility for
-getting them working again with some urgency.
+tests. When they pass, we consider the results fit for public consumption.
+
+When they don't pass, we need to fix them.
+
+Every morning, someone from inframundo will check the #pudl-deployments slack
```
jdangerx, line 25: @catalyst-cooperative/inframundo is it ok to sign us up for this? I think it's mostly been @zaneselvans and @bendnorman checking in the past. I've been doing it in the new year. We could set up a rotation if we want.

docs/dev/nightly_data_builds.rst
```diff
@@ -14,12 +14,29 @@ pushes a Docker image with PUDL installed to `Docker Hub <https://hub.docker.com
...
+
+If it has succeeded, they will track it in `this spreadsheet <https://docs.google.com/spreadsheets/d/11YdknGi4br51kxz03nNmxD-lOvilWFgm3jkUpiEdAzU/edit#gid=1678819446>`__.
+
+If it has failed, they will:
+
+1. Create a GitHub issue for the nightly build failure to hold the investigation & discussion.
+2. Track the GitHub issue and the build status in the above spreadsheet.
+3. Look in the logs and determine whether it was an "infrastructure failure," i.e. something went wrong with the code that *runs*
+the nightly build, or a "PUDL failure," i.e. something that went wrong with the PUDL ETL itself.
+4. Investigate the source of the issue & explore ways to fix it. Get help from the folks whose PRs broke the build.
```
jdangerx, line 36: Do you think we need more guidance here? TBH I sort of ran out of doc-writing steam for this afternoon. But I also think this is probably good enough?

docs/dev/project_management.rst
```diff
@@ -15,6 +15,54 @@ track bugs, enhancements, support requests, and just about any other work that g
...
+
+Our core offering is complete, connected, and granular data. Issues that
+interrupt the availability of that data are of highest importance.
+
+However, not all datasets in PUDL are the same; some are mature and have many
+downstream users; others are more experimental. We've split them into "tier 1"
+and "tier 2" groups below.
+
+**Tier 1 datasets**
+* FERC 1 schedules XYZ
```
jdangerx, line 34: This is just some placeholder stuff. I think the main thrust of this proposal is:

1. We should decide which datasets are "important" and thus warrant firefighting action, vs. "unimportant" and thus get slotted in with the rest.
2. We should also decide what it means for something to be "broken" - i.e. X% data missing/incorrect, new data unincorporated after X time, etc. - **absolutely** I need help actually defining these
3. If we have 1. and 2. then it will be much easier for us to make prioritization decisions about random things that blow up!

docs/dev/nightly_data_builds.rst
e-belfer: I think this is a realistic description of this team's responsibilities, even if we wind up delegating the fix to someone outside of inframundo.

docs/dev/project_management.rst
```diff
@@ -15,6 +15,54 @@ track bugs, enhancements, support requests, and just about any other work that g
...
+* EPA CEMS
+
+**Tier 2 datasets**
+Everything else
+
+This then informs some reliability goals:
+
+For *tier 1* tables:
+
+- latest source data is incorporated into PUDL within 1 month of publication
```
e-belfer, line 46: This is faster than our actual funding levels allow for at present, we're running on a quarterly integration calendar for our sub-annual datasets for now.

docs/dev/project_management.rst
e-belfer: The issue here is that the most integrated data has the most testing infrastructure written into it and is the most likely to fail. I think more useful here would be that the first step of triage is actually to scope out the issue and write some proposed solutions. Then we make the step of deciding whether we want to fix it in the most minimal way (relax the restriction, xfail the test), actually fix the core issue, or implement a more extensive design change. I think the pause between "here's the problem and what needs to be done" and "which version of this fix should we implement now" is probably the thing we most often fail to do and would help us prioritize.

docs/dev/project_management.rst
```diff
@@ -15,6 +15,54 @@ track bugs, enhancements, support requests, and just about any other work that g
...
+Which then, in turn, informs our bug triage guidelines:
+
+**Urgent (find some way to address ASAP)**
+- nightly build failures
+- datasette not available
+- incorrect data in distribution buckets
+
+**High (prioritize in the upcoming sprint planning)**
+- missing/incorrect data in Tier 1 tables
+- new Tier 1 source data available
```
e-belfer, line 61: New data availability won't show up in our nightly builds, so I'm not 100% following the connection here?

e-belfer (commented):
Left some questions and suggestions.

zaneselvans:
@jdangerx is this still relevant with the new bug triage process that has been getting used lately? Does it need to be updated and merged? Or did the documentation end up somewhere already?

PR #3247, opened 2024-01-17
Fix dataframe embedder op factory and tune ferc-ferc model

zschira:
This PR gets the dataframe embedder op-factory working, and adds some tuning to the ferc-ferc model. It also adds some minor tooling to configure models using yaml files.


katie-lamb (approved):
This looks good! Glad you got the dataframe embedder factory working. No requested changes on my end besides updating the xfail in `test/validate/plants_steam_ferc1_test.py::test_dupe_years_in_plant_id_ferc1[ferc1_annual]` to be a passing test.

PR #3252, opened 2024-01-18
Update gas price upper bound; Enable null-row check for MCOE.

### LINKED ISSUE #2033: add newly allocated `fuel_consumed_mmbtu` into `pudl.analysis.mcoe` module
With @grgmiller 's new addition to the allocation module via #1608  (being merged into dev with some additional validation test checks via #1096 ), we will now have both net generation and fuel allocated at the generator level. The generation is a little more naturally integrated into the `mcoe` outputs because we can easily replace the original 923 table accessed by `pudl_out.gen_eia923()`. There is not 923 table that is fuel at the generator level, so we don't have the same straightforward way to integrate the fuel at the generator-level. But we should find a way to integrate it!

Note: without doing this, the `test_no_null_rows_mcoe` mcoe validation test failed at its previous threshold of 80%. I've set this test to xfail for now. So a part of this issue will need to make this validation test pass.

### LINKED ISSUE #3250: Failed nightly build 2024-01-18

zaneselvans:
# Overview

- Bumped the upper bound on natural gas prices per MMBTU to accommodate the integration of the new EIA Bulk Electricity data, which includes more high prices from 2022.
- Found an XPASSing test related to using the allocated fuel in MCOE which... should actualy be passing, so removed the XFAIL mark.

Closes #2033 
Closes #3250 

# Testing

Re-ran the failing tests with `--live-dbs` locally on my full ETL outputs locally and they passed.

```[tasklist]
# To-do list
- [x] Make sure full ETL runs
```

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -31,8 +31,13 @@ v2024.01.XX
 Data Coverage
 ^^^^^^^^^^^^^
 * Updated :doc:`data_sources/epacems` to switch to pulling the quarterly updates of
-  CEMS instead of the annual files. Integrates CEMS through 2023q3. See issue
+  CEMS instead of the annual files. Integrates CEMS through 2023Q3. See issue
   :issue:`2973` & PR :pr:`3096`.
+* Updated the EIA Bulk Electricity data archive so that the available data now to runs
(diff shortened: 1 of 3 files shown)


jdangerx (approved):
Thanks for the quick fix!

Issue #3256, opened 2024-01-18
Refactor `test_validate_override_fixes` to be a unit test


katie-lamb:
Currently `test_validate_override_fixes` is an integration test for the FERC to EIA entity matching module that tests the functionality of `validate_override_fixes`. The test has parameters with record IDs that need to be updated annually to ensure that those records are from report years that are processed in the fast ETL. The test uses the FERC to EIA training data within `validate_override_fixes`. 

Instead, this test should be refactored to be a unit test that takes synthetic training data that is passed into `validate_override_fixes`. Additionally, instead of XFAIL-ing, the tests should pass if an expected exception is raised or failure condition is observed. Finally, `validate_override_fixes` shouldn't allow duplicate FERC IDs to be matched to different EIA records after the overrides are complete. There are test parameters set up to test this, but the `validate_override_fixes` function won't actually catch that condition in it's current state.

- [ ] Refactor `test_validate_override_fixes` to be a unit test that uses synthetic FERC to EIA training data
- [ ] Change XFAIL-ing tests to tests that pass if an expected exception is raised
- [ ] Add a check in `pudl.analysis.record_linkage.eia_ferc1_record_linkage.validate_override_fixes` to make sure that there aren't duplicate FERC1 IDs matched to EIA records.

Issue #3251, opened 2024-01-18
Review XPASSing tests


zaneselvans:
We have a number of tests or data validations which are reporting XPASS in the nightly builds, which we might want to review, in case they should no longer be marked XFAIL:

- [x] `test/validate/mcoe_test.py::test_no_null_rows_mcoe[eia_annual-mcoe-0.8]` (I think this can be removed, see #2033)
- [x] `test/validate/mcoe_test.py::test_no_null_rows_mcoe[eia_monthly-mcoe-0.8]` (I think this can be removed, see #2033)
- [x] `test/integration/zenodo_datapackage_test.py::TestZenodoDatapackages::test_prod_datapackages` (This is here because Zenodo can be flaky w/ network hiccups)
- [x] `test/validate/plants_steam_ferc1_test.py::test_dupe_years_in_plant_id_ferc1[ferc1_annual]` (This should be passing now, @zschira will fix in #3247) 
- [x] `test/integration/ferc1_eia_train_test.py::test_validate_override_fixes[verified4-report_year4-record_id_eia_override_14-record_id_ferc14-utility_id_pudl_ferc14]` See #3257
- [x] `test/integration/ferc1_eia_train_test.py::test_validate_override_fixes[verified5-report_year5-record_id_eia_override_15-record_id_ferc15-utility_id_pudl_ferc15]` See #3257
- [x] `test/integration/ferc1_eia_train_test.py::test_validate_override_fixes[verified6-report_year6-record_id_eia_override_16-record_id_ferc16-utility_id_pudl_ferc16]` See #3257
- [ ] `test/integration/ferc1_eia_train_test.py::test_validate_override_fixes[verified3-report_year3-record_id_eia_override_13-record_id_ferc13-utility_id_pudl_ferc13]`

@katie-lamb are the XFAIL labels for these FERC-EIA test/train tests still required / appropriate?

@zschira or @katie-lamb what do you think about the duplicate FERC years in a plant ID test? Should that be passing now?

katie-lamb:
@zaneselvans A couple things re `test/integration/ferc1_eia_train_test.py`:

- The first of the XPASSing `test_validate_override_fixes` tests ("Duplicate FERC1 ids in overrides") doesn't seem to actually be a problem in the `validate_override_fixes` function that this test is testing. I think this is a legitimate issue with the `validate_override_fixes` function that should be fixed, but the bug seems to be have been there since it's inception, not from the recently merged refactor. 
[...]

PR #3257, opened 2024-01-18
Tiny fix to make FERC to EIA tests xfail

katie-lamb:
# Overview

Addresses 3 of the 4 `test/integration/ferc1_eia_train_test.py::test_validate_override_fixes` tests that should XFAIL but don't as mentioned in #3251 . If we don't care about this patch to make those 3/4 tests XFAIL and just want to address the larger problem in a larger PR, then no need to merge this. 

The following tests should XFAIL but don't because of a small issue with how the parameters are set in the `test_validate_override_fixes` test.
```
test/integration/ferc1_eia_train_test.py::test_validate_override_fixes[verified4-report_year4-record_id_eia_override_14-record_id_ferc14-utility_id_pudl_ferc14]
test/integration/ferc1_eia_train_test.py::test_validate_override_fixes[verified5-report_year5-record_id_eia_override_15-record_id_ferc15-utility_id_pudl_ferc15]
test/integration/ferc1_eia_train_test.py::test_validate_override_fixes[verified6-report_year6-record_id_eia_override_16-record_id_ferc16-utility_id_pudl_ferc16]
```

Based on a scan of the history with this test, I don't think these parameters have ever been correctly set, and these tests have never XFAILed, but I've never touched this test before so I'm not entirely sure.

What did you change?

Since it was a really simple change, I just updated those parameters. But that's really just a patch for the larger issue in this test, which I've moved into #3256 .

# Testing

```[tasklist]
# To-do list
```


zaneselvans:
It seems like it's more correct for the tests marked xfail to fail, and if it's just that these parameters were set incorrectly, that seems like a good change. We can refactor the tests in an actual sprint plan.

zaneselvans (approved):

PR #3258, opened 2024-01-19
PHMSA gas extract step for transmission part k

### LINKED ISSUE #3248: Map columns for transmission K
The goal of this PR is to map columns, file names, skipped rows and footers, and any other relevant partitions for the raw table that will eventually become `core_phmsa__yearly_miles_of_transmission_pipe_by_specified_minimum_yield_strength`. This includes Section K of the PHMSA transmission data from 2010-present.

cmgosnell:
# Overview

Closes #3248.

we didn't have the mapping files for Part K!

What did you change?
```[tasklist]
 - [x] add all the mapping files for part k 2010-present.
- [x] i also removed the `data_label` bc that was not being used anywhere and was probably resulting in a million warnings about a missing column 😬 
- [x] add old times - 2009 (no old times part k that was easy)
- [x] add in a notebook extractor into the devtools notebook
- [x] make the extractor warnings quiet/appropriate for this old year split up into new format thing.
```
 

# Testing

```[tasklist]
# To-do list
```


src/pudl/extract/phmsagas.py
```diff
@@ -35,18 +36,21 @@ def process_final_page(self, df, page):
...
-            if c not in self._metadata.get_all_columns(page)
-            and c not in self.cols_added
-        ]
-        if to_drop:
-            logger.warning(
-                f"Dropping columns {to_drop} that are not mapped to this asset."
-            )
-            df = df.drop(columns=to_drop, errors="ignore")
-        return df
+        if int(partition["year"]) < 2010:
```
e-belfer, line 39: This is also only true for transmission and not distribution data, so we want to filter by `form` as well. We should update the docstring to clarify. Otherwise this is super helpful and catches some columns I've missed in earlier mappings.

e-belfer (changes requested):
Part K: materialized the table successfully and it has all the rows I'd expect.

One blocking issue re what tables we're dropping columns for - should be transmission only.

One additional non-blocking suggestion: you've helpfully set up the column names to get parsed and split into categoricals, and maybe you want to use `nonsteel` instead of `non_steel` to make this even simpler?

src/pudl/extract/phmsagas.py
cmgosnell: this is a great suggestion! will do for sure

cmgosnell: @e-belfer is it only `page != "yearly_distribution"` that we should avoid for this?

i was going to add `and "_transmission_" in page` here but `yearly_inspections_and_assessments` is a part of the transmission zip but it just doesn't look like it is included in old years.

cmgosnell: could also look it up in the page->form map

e-belfer: I think that table should probably get renamed to include transmission in it, honestly.

e-belfer: But the way you've done it is more robust

e-belfer (approved):
Looks great, thanks for cleaning things up!

e-belfer:
Also completely forgot when I first reviewed this, but I actually think `data_label` is what we use to store "final" vs. "early-release" status and I'm not convinced we should be dropping it.

Issue #2603, opened 2023-05-25
Integrate EIA-176 into PUDL


zaneselvans:
# Description
Integrate the EIA-176 natural gas sources & dispositions data into PUDL.

What we've been calling the EIA-176 is actually 3 different related forms that are bundled together for bulk distribution:
- **[EIA-176: Annual Report of Natural and Supplemental Gas Supply & Disposition](https://www.eia.gov/survey/#eia-176)**. Data collected include the origin of natural gas supplies and the disposition of natural gas on a State basis. Respondents include natural and synthetic gas producers, processors, distributors, storage.
- **[EIA-191: Monthly Underground Natural Gas Storage Report](https://www.eia.gov/survey/#eia-191)**. Working and base gas in reservoirs, injections, withdrawals, and location of reservoirs are reported by operators of all underground natural gas storage fields on a monthly basis.
- **[EIA-757: Natural Gas Processing Plant Survey](https://www.eia.gov/survey/#eia-757)**. Detailed plant-level information is collected on the capacity, status, and operations of natural gas processing plants to monitor supply constraints during periods of disruption due to a natural disaster, such as a hurricane.

Each of these forms uses the same IDs to refer to the reporting companies, and that shared company ID information is provided as a separate standalone table with the company ID, name, and activity status. Note that there is no date information associated with the company information, so the activity status probably just pertains to when the reporting was done, and there are no historical archives, so this field is pretty useless. Presumably we'll be able to guess which companies are active based on whether they're reporting data in the other tables?

There's also a lot more [natural gas data available from EIA](https://www.eia.gov/naturalgas/data.php) that we might be able to download in bulk from their API or other hidden endpoints.

## Motivation
This gas source, disposition, storage, and processing plant data should help us target existing natural gas utilities and the capital locked up in existing infrastructure for early retirement, and may help advocates prevent new investments in natural gas facilities that would need to be decommissioned well before end of life to maintain a stable climate.

## In Scope
We'll know we're done when:
- EIA-176, 191, and 757 data are loaded into the PUDL DB by the main DAG. It looks like this will probably result in 4 tables: companies, sources & dispositions, underground storage, and processing facilities
- Tables have well-defined schemas, including data types and descriptions for all columns.
- Tables are in a tidy format, with each column representing a single variable, and each row representing a single observation.
- Updates to data should be easy, and as automated as practicable, alongside our other existing data.

## Out of Scope
[...]

PR #3264, opened 2024-01-19
Add `data` and `other` to EIA 176 raw data extraction

e-belfer:
# Overview

This PR builds on #3227 and adds `data` and `other` 176 data to the extraction process designed by @davidmudrauskas, transforming the extractor function into a dagster asset factory to enable simple creation of a dagster asset for each table.

Extracts raw EIA 176 "data" and "other" tables.

What did you change?

Converted the extractor in `pudl.extract.eia176` to a factory, and added to the `table_file_map.csv`. Updated the unit test to handle multiple files, rather than just one.

# Testing

Generate the raw assets in dagster and open them to review.

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/extract/eia176.py
+++ b/src/pudl/extract/eia176.py
@@ -3,24 +3,40 @@
 The EIA Form 176 archive also contains CSVs for EIA Form 191 and EIA Form 757.
 """
 
-from dagster import asset
+import pandas as pd
+from dagster import AssetsDefinition, asset
 
 from pudl.extract.csv import CsvExtractor, get_table_file_map
 
 DATASET = "eia176"
 
 
-@asset(required_resource_keys={"datastore"})
-def raw_eia176__company(context):
-    """Extract raw EIA company data from CSV sheets into dataframes.
+def eia_176_asset_factory(
+    table_name: str,
+) -> AssetsDefinition:
+    """Build bulk electricity tables from raw data."""
 
-    Args:
-        context: dagster keyword that provides access to resources and config.
+    @asset(
+        name=f"raw_eia176__yearly_{table_name}", required_resource_keys={"datastore"}
+    )
+    def eia_176_table(context) -> pd.DataFrame:
+        """Extract raw EIA company data from CSV sheets into dataframes.
 
-    Returns:
-        An extracted EIA dataframe with company data.
-    """
-    zipfile = context.resources.datastore.get_zipfile_resource(DATASET)
-    table_file_map = get_table_file_map(DATASET)
-    extractor = CsvExtractor(zipfile, table_file_map)
-    extractor.extract_one("company")
+        Args:
+            context: dagster keyword that provides access to resources and config.
+
+        Returns:
+            An extracted EIA dataframe.
+        """
+        zipfile = context.resources.datastore.get_zipfile_resource(DATASET)
+        table_file_map = get_table_file_map(DATASET)
+        extractor = CsvExtractor(zipfile, table_file_map)
+        return extractor.extract_one(table_name)
+
+    return eia_176_table
+
+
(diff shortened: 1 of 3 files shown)


zaneselvans:
> EIA 176 zipfiles also bundle a few other forms - 191 and 757. Where do we want to extract and process these datasets? As separate modules, or as part of the EIA 176 extraction?

I wouldn't interpret the "EIA-176" label narrowly.  There's nothing particularly special about that form, and as noted in #2603 initially we thought that all of the data in this bulk download zipfile was EIA-176, but it turns out there were these other associated forms related to gas production, storage, etc. IIRC in my initial digging around, it seemed like some of it looked usable, and some of it was both very messy and only had small number of years of data included. Almost like they accidentally dumped the data in this CSV once and then never looked at it again (which means the data is somewhere else, but... where?).

Rather than focusing on the forms in particular, I'd try and identify the subset of data that we've got archived and ready to process which are actually worth cleaning up and turning into tables -- like there's a significant amount of data and it's a tractable problem without a huge investment of manual effort.  Or at least prioritize them in terms of "person-hours per unit of data integrated"

Another thing that we should look at is how the data available from this obscure zipfile compares / relates to the [bulk natural gas API data](https://www.eia.gov/opendata/bulk/NG.zip) available from EIA.  My guess is that like the EIA-860 and EIA-923 the data in this zipfile is part of what gets fed into the bulk data they publish, but that there may be other data coming from other forms too, and the bulk API data may not reflect the full detail of the data that's in the original submitted form responses. Like I think we're looking for the long historical record of inputs that go into making their [natural gas data products](https://www.eia.gov/naturalgas/data.php)

The [EIA Natural Gas Annual Report](https://www3.eia.gov/naturalgas/annual/) refers to forms EIA-176, EIA-895, and EIA-910 in Appendix A: "summary of data collection operations and report methodology" though weirdly EIA-895 doesn't show up on their [big list of forms](https://www.eia.gov/survey).

I assume that the EIA-191 (monthly underground natural gas storage report) and EIA-191L (monthly LNG storage report) will have some relationship to the gas storage facility data that comes out of PHMSA, so those would be good to look at in tandem.  While the EIA-757 (Natural Gas Processing Plant Survey) will probably relate to some of the pipeline infrastructure reported in PHMSA. Compressor stations, NGL precipitators, H2S removal facilities. I assume linking the facilities and their owners/operators between the EIA and PHMSA data will be another entity matching circus that we probably don't want to get into now.

src/pudl/extract/eia176.py
```diff
@@ -3,24 +3,40 @@
...
 
 from pudl.extract.csv import CsvExtractor, get_table_file_map
 
 DATASET = "eia176"
 
 
-@asset(required_resource_keys={"datastore"})
-def raw_eia176__company(context):
-    """Extract raw EIA company data from CSV sheets into dataframes.
+def eia_176_asset_factory(
```
zaneselvans, line 14: I think we should stick with the dataset ID `eia176` and never introduce an `_` in here. It's nice to be able to search for these IDs reliably and use the same idioms for all the datasets like `eia860`, `eia861`, `eia923` etc.

e-belfer: You mean in the function names? Simple enough fix, that makes sense.

zaneselvans: Function names, variable names, tables, classes, anywhere we're naming something to refer to the dataset.

zaneselvans: Not sure what we'll want to do about the mish-mash of different forms that were hiding inside this zipfile.

zaneselvans:
@e-belfer can we delete the `eia176_extraction` branch?

PR #3274, opened 2024-01-22
Update conda lockfile for week of 2024-01-22

pudlbot:


zaneselvans (approved):

zaneselvans:
Note that the huge number of changes in this PR is due to `conda-lock` v2.5.2 sorting dependencies in the environment files alphabetically (which is much more readable!). There are only a normal number of actually updated packages.

PR #3279, opened 2024-01-22
PHMSA transmission part P

### LINKED ISSUE #3278: Map columns for PHMSA transmission part P

jdangerx:
# Overview

Closes #3278.

This maps part P of the PHMSA transmission/gathering pipeline data.

It has state-by-state information about pipeline materials, from 2010 onwards.

I added a column map for `yearly_miles_of_pipe_in_state_by_material` and added that to the various "where to find this data" CSVs.

I also made the release notes have a nested list of the small PRs so that we don't have a merge conflict on that same line over and over.

# Testing

I ran the debugging notebook and also materialized the PHMSA assets.

```[tasklist]
# To-do list
```


src/pudl/package_data/phmsagas/column_maps/yearly_miles_of_pipe_in_state_by_material.csv
```diff
@@ -0,0 +1,102 @@
...
+total_total_other,,,,,,,,,,,,,,,,,,,,,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal
+total_total_total,,,,,,,,,,,,,,,,,,,,,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles
+other_materials,,,,,,,,,,,,,,,,,,,,,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials,partpothermaterials
+gathering_other_materials,,,,,,,,,,,,,,,,,,,,,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials,partpgothermaterials
```
jdangerx, line 102: So for most of the field names, we have `part<letter><t or g or nothing><name>` - so I inferred that these were the "total other materials" and "gathering other materials" columns... but I think it sort of makes sense for these to be the "transmission other materials" & "gathering other materials" columns - to show what those other materials might be. I guess `partpgothermaterials` is pretty unambiguously "gathering", so my question is - do we think we should call this `total_other_materials` or `transmission_other_materials`?

jdangerx (commented):
One question that I need a bit of help with.

src/pudl/package_data/phmsagas/column_maps/yearly_miles_of_pipe_in_state_by_material.csv
cmgosnell: Ugh that's tough. I think I would go with `total_other_materials` because it is not specified.... even though with more context we might learn that this is transmission not total.

It looks like these two are outside of the core table in the PDF version of the table. The description is "Composite pipe requires a PHMSA Special Permit or waiver from a State2 specify Other material(s):"
I would definitely add something about `special_permit` or something to the names of both of these.
So I

src/pudl/package_data/phmsagas/column_maps/yearly_miles_of_pipe_in_state_by_material.csv
```diff
@@ -0,0 +1,102 @@
...
+total_total_wrought_iron,,,,,,,,,,,,,,,,,,,,,partpwitotal,partpwitotal,partpwitotal,partpwitotal,partpwitotal,partpwitotal,partpwitotal,partpwitotal,partpwitotal,partpwitotal,partpwitotal,partpwitotal,partpwitotal
+total_total_plastic,,,,,,,,,,,,,,,,,,,,,partpptotal,partpptotal,partpptotal,partpptotal,partpptotal,partpptotal,partpptotal,partpptotal,partpptotal,partpptotal,partpptotal,partpptotal,partpptotal
+total_total_composite,,,,,,,,,,,,,,,,,,,,,partpctotal,partpctotal,partpctotal,partpctotal,partpctotal,partpctotal,partpctotal,partpctotal,partpctotal,partpctotal,partpctotal,partpctotal,partpctotal
+total_total_other,,,,,,,,,,,,,,,,,,,,,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal,partpototal
+total_total_total,,,,,,,,,,,,,,,,,,,,,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles,partptotalmiles
```
cmgosnell, line 100: i love this one

src/pudl/package_data/phmsagas/column_maps/yearly_miles_of_pipe_in_state_by_material.csv
cmgosnell, line 1: you did a great job on these names they look great overall! we have been add a suffix of `_miles` for all the fields that contain miles of pipe, which i think all of these are.

cmgosnell (changes requested):
one request to add `_miles` suffixes and obvis we should settle on the name of the weird not in the table two you brought up but otherwise this looks grand

oh plus i think you need to add the pr into the release notes

src/pudl/package_data/phmsagas/column_maps/yearly_miles_of_pipe_in_state_by_material.csv
jdangerx: Indeed, I'll add the miles to all but the weird "other materials" columns.

src/pudl/package_data/phmsagas/column_maps/yearly_miles_of_pipe_in_state_by_material.csv
jdangerx: I read the PDF as saying two things:

1. Composite pipe requires a special permit / waiver
2. please specify "other materials" - we'll store this as `partpothermaterials` and `partpgothermaterials`

In which case we shouldn't have any `composite`/`special_permit` in the name.

Part of why I read it that way is that the rest of the form distinguishes between `composite` & `other` materials, and another is that there's a little superscript (1) by the Composite column header which seems like it would correspond to the first point.

image

image

What do you think?

e-belfer: Just a heads up that when there's a "specify" field I've been calling it "x_y_description" as a column name. (E.g., "gathering_other_materials_description").

cmgosnell: oh yes you are totally right here. good eye. no `composite` or `special_permit`!

hm okay well that changes my answer to you were originally asking because now we can check if `partpototal` shows up when there is any other pipe or just transmission and it looks like there are only values in this `partpototal` when there is other transmission miles! so i would go with `transmission_other_materials`

jdangerx: Sweet, thanks for all the discussion! Going with `transmission/gathering_other_materials_description` :)

cmgosnell (approved):
🚀

PR #3281, opened 2024-01-24
Remove datapackage, use frictionless.

### LINKED ISSUE #1420: Update PUDL from `datapackage==1.15.2` to `frictionless>=4.40,<5`
The pudl-zenodo-storage currently uses datapackage 1.15.2. It seems like frictionless has wrapped their packages up into one main package called [Frictionless Framework](https://framework.frictionlessdata.io/docs/development/migration/#from-datapackage):

From [PyPi](https://pypi.org/project/datapackage/):
_[Important Notice] We have released Frictionless Framework. This framework provides improved datapackage functionality extended to be a complete data solution. The change in not breaking for the existing software so no actions are required. Please read the Migration Guide from datapackage to Frictionless Framework._

Done when datapackges has been updated.

jdangerx:
# Overview

Closes #1420 

We were using a super old package to read frictionless datapackages, now we are merely using a fairly old package that is the same version we use in archiver.

What did you change?
- update package dep
- change package validation call
- also include the 5.x upgrade path.

# Testing

I ran `pudl_datastore --verify` and saw that nothing blew up.

```[tasklist]
# To-do list
```


src/pudl/workspace/datastore.py
```diff
@@ -167,8 +167,11 @@ def _validate_datapackage(self, datapackage_json: dict):
 
         Throws ValueError if invalid.
         """
-        dp = datapackage.Package(datapackage_json)
-        if not dp.valid:
+        # TODO (daz): when we upgrade to frictionless>=5.0, this will be:
+        # dp = frictionless.Package.validate_descriptor(datapackage_json)
+        # if not dp.valid:
+        dp = frictionless.Package(datapackage_json)
+        if not dp.metadata_validate():
```
zaneselvans, line 174: I could be remembering wrong, but I'm suspicious that this validation may **only** checking the validity of the topmost metadata elements, and not descending into any of the referenced resources to check whether they are also valid. If that's true, it would be informative to try validating the resources and see whether they also pass.

I think a lot of the rest of this class is also doing work that frictionless should really be doing, like validating the file sizes, MD5 checksums and potentially paths.

jdangerx:
I was fiddling around with this mostly in Frictionless 5, but there *is* a
way (package.validate()) to check that the SQL tables do what they say.
But, we have to change the “dialect” key in the datapackage to point at
{“sql”: {“tag: …}} instead of just {“tag: …}.

Not sure what the deal is w frictionless 4, or whether we could just update
frictionless across the board.

Zane Selvans ***@***.***>于2024年1月23日 周二下午8:56写道：

zaneselvans:
IIRC when @zschira was refactoring the archiver and building the initial ferc-xbrl-extractor, Frictionless 5 had just come out and there were some non-trivial API changes between it and 4, and we've just never taken the time to make the transition. I think it would definitely be desirable, since all the non-CSV tabular annotation is only available in 5, and it would be nice to be on the current release so we don't get left behind.

jdangerx:
@zaneselvans - I think we should probably move ahead with "just get this thing up-to-par with the rest of our code," i.e. #1420, then we can think about #3293 separately.

zaneselvans:
@jdangerx Oh yeah definitely agree we should get all the repos working with Frictionless v4 first (which will also purge some crusty old dependencies) and then switch everything over to v5 in a separate step.

src/pudl/workspace/datastore.py
jdangerx: Yes, this is not properly checking the referenced resources. But, since the referenced resources aren't defined with the correct `{"dialect": {"sql": {"table": "blah"}}}` format, and frictionless 4 doesn't know what SQL files are anyways, I think this is the best we can do at this point...

zaneselvans: In the datastore use case, we're using generic [data resources](https://specs.frictionlessdata.io/data-resource/) which don't have any schema / dialect attributes or know anything about the internal structure of the data. They're just files of any kind, with a name, description, path, filetype, size, hash, and potentially other freeform metadata. This is in contrast to the [tabular data resources](https://specs.frictionlessdata.io/tabular-data-resource/) which are intended to represent tables and can describe much more of the expected internal structure of the resource.

Since we're just trying to bump to the new version I think addressing missing validation step this is well outside the scope of this PR, but I think we can validate these resources and should try to get that working in a future PR so we don't have so much of our own infrastructure paralleling stuff that Frictionless can/should do.

jdangerx: Ah, right, since we're not downloading any actual SQL resources into $PUDL_INPUT - that only affects our fancy `fercX_xbrl_datapackage.json`s.

That being said, I've some write-up of what it would take to get all this stuff into Frictionless in #3293 .

zaneselvans:
@jdangerx I just regenerated the lockfiles locally using a newer version of `conda-lock` to minimize the diffs and keep the ones on `main` sorted.

zaneselvans (approved):
See clarifying comment about (non-tabular) data resource validation in the `datastore` module.

Issue #1572, opened 2022-04-06
Final data validation using dagster-pandera


bendnorman:

zaneselvans:
We've long since switched to using pandera for our schema asset checks, e.g. see #5432 (the dagster-pandera integration turned out not to be exactly what we needed) and other data validations are now handled by dbt

PR #3282, opened 2024-01-24
Prototype dagster-pandera integration

jdangerx:
# Overview

Relevant to #1572, but doesn't close it. Opens the possibility of closing it.

We need to integrate:
* our existing resource schema definition system
* Pandera
* dagster

Fortunately, there's already integration between pandera and dagster. We just need a shim layer between our existing schema definition system and pandera - hence, this PR.

### Changes:

* shim layer between our schema definitions & pandera
* create dagster asset checks for all assets that correspond to a Resource
* port boiler fuel validation checks to resource schema

See details for the original rant, but I ended up introducing some classes to define our existing `RESOURCE_METADATA` data structure. That lays the groundwork for moving towards using the `frictionless` library in the future, as well as further refactoring to make our metadata handling interoperate with various different libraries like we discussed in https://github.com/orgs/catalyst-cooperative/discussions/2546. 

<details>

Our existing schema definition system is hard to change for many reasons; one of which is that the shape of the data in `RESOURCE_METADATA` is actually not documented anywhere! Instead we have all sorts of weird logic scattered throughout to turn that undocumented shape into something resembling Frictionless Packages/Resources/Fields (but that are distinct classes...)

One thing we could do to make this whole thing a bit more tractable is to actually define some sort of `ResourceMetadata` or `ResourceSpec` class (and, potentially, companion `SchemaSpec`, `FieldSpec` classes) that:
- describes what we expect from `RESOURCE_METADATA`
- knows how to translate from itself into various different formats: frictionless, pandera, pyarrow, pandas, etc.

Right now, our `Resource` (and companion) classes happens to be "a copy of the frictionless class, sort of, that knows how to translate from the undefined shape of what's in `RESOURCE_METADATA` into our ersatz frictionless class." Which causes much entanglement/confusion.

It doesn't seem like it would be *too* big of a refactor, to split up `Resource` into `ResourceSpec` and the official `frictionless.Resource`, so maybe we should do that sometime soon.
</details>

# Testing

I ran the whole ETL, and saw that all the asset checks passed. I also happened to pick an asset which has some validation checks we expect to fail - so I saw that failing checks did actually show up in Dagster, before then adding machinery to xfail certain checks.

```[tasklist]
# To-do list
- [x] use Pandera to express real data type and validation tests for an asset in Dagster
```

### DIFF
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -120,10 +120,18 @@ def asset_check_from_schema(
     def pandera_schema_check(asset_value) -> AssetCheckResult:
         try:
             pandera_schema.validate(asset_value, lazy=True)
-        except pr.errors.SchemaErrors as e:
+        except pr.errors.SchemaErrors as schema_errors:
             return AssetCheckResult(
                 passed=False,
-                metadata={"schema_errors": [str(err) for err in e.schema_errors]},
+                metadata={
+                    "errors": [
+                        {
+                            "failure_cases": str(err.failure_cases),
+                            "data": str(err.data),
+                        }
+                        for err in schema_errors.schema_errors
+                    ],
+                },
             )
         return AssetCheckResult(passed=True)
 
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -1,13 +1,19 @@
 """Dagster definitions for the PUDL ETL and Output tables."""
 import importlib.resources
+import itertools
+import warnings
 
+import pandera as pr
 from dagster import (
+    AssetCheckResult,
+    AssetChecksDefinition,
     AssetKey,
     AssetsDefinition,
     AssetSelection,
     Definitions,
+    ExperimentalWarning,
+    asset_check,
     define_asset_job,
-    load_asset_checks_from_modules,
     load_assets_from_modules,
 )
 
@@ -31,6 +37,7 @@ from . import (
 )
 
 logger = pudl.logging_helpers.get_logger(__name__)
+
 
 default_assets = (
     *load_assets_from_modules([eia_bulk_elec_assets], group_name="core_eia_bulk_elec"),
@@ -96,13 +103,54 @@ default_assets = (
     ),
 )
 
-default_asset_checks = (
-    *load_asset_checks_from_modules(
-        [
-            pudl.output.eia923,
-        ],
-    ),
+
+def asset_check_from_schema(
+    asset_key: AssetKey,
+    package: pudl.metadata.classes.Package,
+) -> AssetChecksDefinition | None:
+    """Create a dagster asset check based on the resource schema, if defined."""
+    resource_id = asset_key.to_user_string()
+    if resource_id != "out_eia923__boiler_fuel":
+        return None
+    try:
+        resource = package.get_resource(resource_id)
+    except ValueError:
+        return None
+    pandera_schema = resource.schema.to_pandera()
+
+    @asset_check(asset=asset_key)
+    def pandera_schema_check(asset_value) -> AssetCheckResult:
+        try:
+            pandera_schema.validate(asset_value, lazy=True)
+        except pr.errors.SchemaErrors as e:
+            return AssetCheckResult(
+                passed=False,
+                metadata={"schema_errors": [str(err) for err in e.schema_errors]},
+            )
+        return AssetCheckResult(passed=True)
+
+    return pandera_schema_check
+
+
+def _get_keys_from_assets(asset_def):
+    try:
+        return asset_def.keys
+    except AttributeError:
+        return [asset_def.key]
+
+
+warnings.filterwarnings("ignore", category=ExperimentalWarning)
+_package = pudl.metadata.classes.Package.from_resource_ids()
+_asset_keys = itertools.chain.from_iterable(
+    _get_keys_from_assets(asset_def) for asset_def in default_assets
 )
+default_asset_checks = [
+    check
+    for check in (
+        asset_check_from_schema(asset_key, _package) for asset_key in _asset_keys
+    )
+    if check is not None
+]
 
 default_resources = {
     "datastore": datastore,
@@ -203,4 +251,7 @@ defs: Definitions = Definitions(
         ),
     ],
 )
+
+warnings.resetwarnings()
+
 """A collection of dagster assets, resources, IO managers, and jobs for the PUDL ETL."""
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -110,8 +110,6 @@ def asset_check_from_schema(
 ) -> AssetChecksDefinition | None:
     """Create a dagster asset check based on the resource schema, if defined."""
     resource_id = asset_key.to_user_string()
-    if resource_id != "out_eia923__boiler_fuel":
-        return None
     try:
         resource = package.get_resource(resource_id)
     except ValueError:
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -12,10 +12,12 @@ from dagster import (
     AssetSelection,
     Definitions,
     ExperimentalWarning,
+    SourceAsset,
     asset_check,
     define_asset_job,
     load_assets_from_modules,
 )
+from dagster._core.definitions.cacheable_assets import CacheableAssetsDefinition
 
 import pudl
 from pudl.io_managers import (
@@ -138,13 +140,31 @@ def asset_check_from_schema(
     return pandera_schema_check
 
 
-def _get_keys_from_assets(asset_def):
-    try:
-        return asset_def.keys
-    except AttributeError:
+def _get_keys_from_assets(
+    asset_def: AssetsDefinition | SourceAsset | CacheableAssetsDefinition,
+) -> list[AssetKey]:
+    """Get a list of asset keys.
+
+    Most assets have one key, which can be retrieved as a list from
+    ``asset.keys``.
+
+    Multi-assets have multiple keys, which can also be retrieved as a list from
+    ``asset.keys``.
+
+    SourceAssets always only have one key, and don't have ``asset.keys``. So we
+    look for ``asset.key`` and wrap it in a list.
+
+    We don't handle CacheableAssetsDefinitions yet.
+    """
+    if isinstance(asset_def, AssetsDefinition):
+        return list(asset_def.keys)
+    if isinstance(asset_def, SourceAsset):
         return [asset_def.key]
+    return []
 
 
+# Asset Checks are still Experimental, silence the warning since we use them
+# everywhere.
 warnings.filterwarnings("ignore", category=ExperimentalWarning)
 _package = pudl.metadata.classes.Package.from_resource_ids()
 _asset_keys = itertools.chain.from_iterable(
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -7,6 +7,7 @@ from dagster import (
     AssetSelection,
     Definitions,
     define_asset_job,
+    load_asset_checks_from_modules,
     load_assets_from_modules,
 )
 
@@ -95,6 +96,14 @@ default_assets = (
     ),
 )
 
+default_asset_checks = (
+    *load_asset_checks_from_modules(
+        [
+            pudl.output.eia923,
+        ],
+    ),
+)
+
 default_resources = {
     "datastore": datastore,
     "pudl_io_manager": pudl_mixed_format_io_manager,
@@ -153,6 +162,7 @@ def load_dataset_settings_from_file(setting_filename: str) -> dict:
 
 defs: Definitions = Definitions(
     assets=default_assets,
+    asset_checks=default_asset_checks,
     resources=default_resources,
     jobs=[
         define_asset_job(
--- a/src/pudl/metadata/classes.py
+++ b/src/pudl/metadata/classes.py
@@ -737,9 +737,10 @@ class Schema(PudlMeta):
 
     fields: StrictList[Field]
     missing_values: list[StrictStr] = [""]
-    primary_key: StrictList[SnakeCase] | None = None
+    primary_key: list[SnakeCase] = []
     foreign_keys: list[ForeignKey] = []
-    checks: dict[SnakeCase, Callable] = {}
+    df_checks: list[Callable] = []
+    field_checks: dict[SnakeCase, Callable] = {}
 
     _check_unique = _validator(
         "missing_values", "primary_key", "foreign_keys", fn=_check_unique
@@ -784,15 +785,18 @@ class Schema(PudlMeta):
 
     def to_pandera(self: Self) -> pr.DataFrameSchema:
         """Turn PUDL Schema into Pandera schema, so dagster can understand it."""
+        # 2024-02-09: pr.Check doesn't have interop with Pydantic type system
+        # yet, so we encode as Callable, then cast.
         return pr.DataFrameSchema(
             {
                 field.name: pr.Column(
                     FIELD_DTYPES_PANDAS[field.type],
-                    checks=cast(pr.Check, self.checks.get(field.name, None)),
+                    checks=cast(pr.Check, self.field_checks.get(field.name, None)),
                 )
                 for field in self.fields
             },
             unique=self.primary_key,
+            checks=cast(list[pr.Check], self.df_checks),
         )
 
 
@@ -1046,7 +1050,8 @@ class PudlResourceDescriptor(PudlMeta):
 
         field_ids: list[str] = pydantic.Field(alias="fields", default=[])
         primary_key_ids: list[str] = pydantic.Field(alias="primary_key", default=[])
-        checks: list[Callable] = []
+        df_checks: list[Callable] = []
+        field_checks: dict[str, Callable] = {}
         foreign_key_rules: PudlForeignKeyRules = PudlForeignKeyRules()
 
     class PudlCodeMetadata(PudlMeta):
@@ -1055,8 +1060,8 @@ class PudlResourceDescriptor(PudlMeta):
         class CodeDataframe(pr.DataFrameModel):
             """The DF we use to represent code/label/description associations."""
 
-            # TODO (daz): each of these | Nones are one-offs. Fix the frickin
-            # data instead.
+            # TODO (daz) 2024-02-09: each of these | Nones are one-offs. Fix
+            # the frickin data instead.
             code: pr.typing.Series[Any]
             label: pr.typing.Series[str] | None
             description: pr.typing.Series[str]
@@ -1068,12 +1073,12 @@ class PudlResourceDescriptor(PudlMeta):
         code_fixes: dict = {}
         ignored_codes: list = []
 
-    # TODO (daz): with a name like "title" you might imagine all resources
-    # would have one...
+    # TODO (daz) 2024-02-09: with a name like "title" you might imagine all
+    # resources would have one...
     title: str | None = None
     description: str
     schema_: PudlSchemaDescriptor = pydantic.Field(alias="schema")
-    encoder: PudlCodeMetadata = PudlCodeMetadata()
+    encoder: PudlCodeMetadata | None = None
     source_ids: list[str] = pydantic.Field(alias="sources")
     etl_group_id: str = pydantic.Field(alias="etl_group")
     field_namespace_id: str = pydantic.Field(alias="field_namespace")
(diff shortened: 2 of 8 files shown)


jdangerx:
cc @zaneselvans, who I think has done a lot of the recent touching of this, and @cmgosnell + @e-belfer , who have recently expressed some desire for using our hard-won schema information to annotate our asset types.

bendnorman:
Something to note is dagster's addition of [asset checks](https://docs.dagster.io/concepts/assets/asset-checks) which allow you to run arbitrary tests after an asset executes. The feature is still experimental but seems pretty useful. I [asked how](https://github.com/dagster-io/dagster/discussions/15880#discussioncomment-6788995) asset checks and something like dagster-pandera compare. If we want to use pandera + dagster asset checks they recommend validating the asset using pandera inside of the asset check instead of using dagster-pandera.

jdangerx:
Ooh, thanks @bendnorman ! Using something that's more "core" dagster is better than using something that's less core :)

zaneselvans:
I'm not sure if it matters but it seems like there's a conflation of 2 related functions here: asset type checking, and more complex data validations that look at the data contents, beyond just the schema (though I guess you could consider constraints on the values and relationships between columns part of a more complicated "type" with the validations that Pydantic does).

But if we use an asset check and happen to use Pandera inside it, will the dataframe typing information still be available to Dagster? Or to us in the IDE?  It sounded like there was a vague plan to updated `dagster-pandera` to incorporate asset checks

> For now you should use pandera directly. We'll update dagster-pandera it to add checks support but don't have a date yet

src/pudl/metadata/classes.py
```diff
@@ -190,13 +191,15 @@ def _get_jinja_environment(template_dir: DirectoryPath = None):
 """Positive :class:`float`."""
 
 
-def StrictList(item_type: type = Any) -> type:  # noqa: N802
-    """Non-empty :class:`list`.
+T = TypeVar("T")
+StrictList = Annotated[list[T], pydantic.Field(min_length=1)]
```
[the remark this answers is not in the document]
zaneselvans: Ahhh, I've been wondering how to do this right for years.  I think there are some ergonomics improvements for this in Python 3.12 too, which we're about to be able to start using.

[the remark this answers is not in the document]
jdangerx: Yes indeed! The mypy error message actually suggested using the 3.12 sugar, which didn't work - I banged my head against that for a bit before realizing what was going on.

src/pudl/etl/__init__.py
```diff
@@ -93,6 +101,61 @@
...
+                        for err in schema_errors.schema_errors
+                    ],
+                },
+            )
+        return AssetCheckResult(passed=True)
+
+    return pandera_schema_check
+
+
+def _get_keys_from_assets(asset_def):
```
jdangerx, line 139: This is a weird thing to handle both multi assets and source assets.

src/pudl/metadata/classes.py
```diff
@@ -288,6 +291,24 @@ def _check_max(cls, value, info: ValidationInfo):
                 raise ValueError("must be greater or equal to minimum")
         return value
 
+    def to_pandera_checks(self) -> list[pr.Check]:
```
jdangerx, line 294: Eventually, it would be nice to have this be a proper `frictionless` field definition, and use the Pandera built-in conversion method.

src/pudl/metadata/classes.py
```diff
@@ -1001,6 +1058,69 @@ class ResourceHarvest(PudlMeta):
     """Fraction of invalid fields above which result is considerd invalid."""
 
 
+class PudlResourceDescriptor(PudlMeta):
```
jdangerx, line 1061: Added this class so that we have a proper description/enforcement of the shape of `RESOURCE_METADATA`s. This should make it easier for someone to understand how to add a new resource.

Ran into a couple things that are a little funky, documented as TODOs.

src/pudl/metadata/classes.py
```diff
@@ -1185,8 +1305,18 @@ def _check_harvest_primary_key(cls, value, info: ValidationInfo):
         return value
 
     @staticmethod
-    def dict_from_id(x: str) -> dict:  # noqa: C901
-        """Construct dictionary from PUDL identifier (`resource.name`).
+    def dict_from_id(resource_id: str) -> dict:
```
jdangerx, line 1308: It was surprisingly simple to run all of our `RESOURCE_METADATA` through the new type and validation machinery!

src/pudl/validate.py
```diff
@@ -1533,6 +1554,7 @@ def plot_vs_agg(orig_df, agg_df, validation_cases):
         "low_bound": 0.95,
         "data_col": "fuel_mmbtu_per_unit",
         "weight_col": "fuel_consumed_units",
+        "xfail": True,
```
jdangerx, line 1557: It might be worthwhile to define a `BoundsCheckParams` class that all these bounds checks conform to, instead of having them be dictionaries that can be anything. It's not *necessary* for this code to work, though. What do people think?

zaneselvans: This module is a horror show written long ago, before we knew about Pydantic or data classes or typed dicts. It's long overdue for an overhaul. I don't even know if it should continue existing after we rip out `PudlTabl` and change how the tests and validations access the database.

zaneselvans (changes requested):
Wow this is great! Huge progress toward a better validation setup, and yet not huge changes. I had some minor clarifying questions and requests for docstrings, but I think we should get this in as incremental progress and think about whether we want to move on to anything mentioned below:

Since we have so much new data coming into the DB right now, I think it would be helpful to create some more aspirational templates / examples of how we actually want the data validations to look going forward, so that as we add new cases, we can be doing it the Right Way.  The factory wrapper for converting the old validation feels little opaque for this purpose.

So it would be nice to have examples of both new `df_checks` and `field_checks` which are not tied to the old data validation implementation (though the underlying content could be drawn from those validations if we want)

Probably this would mean implementing at least one parametrized check class and migrating at least some of the existing data validations into that container completely, not just by reference to `pudl.validate`

Running the validations in the ETL will require more compute, but I think there's likely to be a big overall performance speedup for the validations, given that the way the validations are currently organized, the same dataframes are being read out of the DB and into memory repeatedly (since the tests are grouped by type of validation, not by which dataframe they apply to). This also results in memory usage blowing up if you try and run the data validations on a machine with lots of cores with pytest-xdist.

And then of course we'll also discover violations of our validation checks immediately in the ETL instead of in the nightly builds, which means we'll fix them sooner.

IIRC by default failing asset checks don't cause the ETL to fail, do they? How will we be notified when a check fails? Do we want to set them to ERROR out on check failure? I think I saw that was a new option as of the most recent dagster release.

Are there any checks that we can't encode this way? I guess any check (like referential integrity) that involves more than one table won't work.

src/pudl/metadata/classes.py
jdangerx: Yes! Basically:
* It's helpful to have something like `PudlResourceDescriptor` to explicitly encode our bespoke resource descriptor data structure!
* It's... much less helpful if we don't actually enforce that that's what we're working with in the rest of the code.
* This should also blow up on incorrectly defined resources at validation time, instead of when we're then trying to do all this transformation logic to them.

src/pudl/validate.py
jdangerx: I think what makes the most sense is:

1. Use the PUDL resource schemas to define the data type validations. If e.g. "everything in this column should be >10" that seems like it lives in the schema.
2. More complicated validations should be defined alongside the assets - you can just define an `@asset_check` right in the module, in which case we'll have to add some `load_asset_checks_from_module` calls to `etl/__init__.py`. 
3. We should have utility functions + types related to validations in the `pudl.validations` module.

I don't *think* we need to then reference the more complicated validations alongside the schema in RESOURCE_METADATA.

src/pudl/etl/__init__.py
```diff
@@ -95,6 +103,61 @@
...
+    _get_keys_from_assets(asset_def) for asset_def in default_assets
+)
+default_asset_checks = [
+    check
+    for check in (
+        asset_check_from_schema(asset_key, _package) for asset_key in _asset_keys
+    )
+    if check is not None
+]
+
```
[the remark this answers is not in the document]
jdangerx: They're associated by explicitly linking them in the `asset` param you pass to `@asset_check`.

src/pudl/etl/__init__.py
```diff
@@ -95,6 +103,61 @@
     ),
 )
 
+
+def asset_check_from_schema(
+    asset_key: AssetKey,
+    package: pudl.metadata.classes.Package,
+) -> AssetChecksDefinition | None:
+    """Create a dagster asset check based on the resource schema, if defined."""
```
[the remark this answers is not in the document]
jdangerx: No, I don't think so. But I do expect there to be assets that are built without corresponding `Resource`s - e.g. `raw_*`.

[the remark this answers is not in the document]
zaneselvans: Ah, okay. I think the docstring could be clearer then. "if defined" is not "if there's a schema defined for this resource" but "if there's a resource defined for this asset."

jdangerx:
> Wow this is great! Huge progress toward a better validation setup, and yet not huge changes. I had some minor clarifying questions and requests for docstrings, but I think we should get this in as incremental progress and think about whether we want to move on to anything mentioned below:

Sweet! Yeah, I was a bit bummed that I let the scope creep up (from Do One Asset to Do A Bunch of Assets + Make Some Ergonomic Improvements) , but I think the extra work was very high bang for buck.

> Since we have so much new data coming into the DB right now, I think it would be helpful to create some more aspirational templates / examples of how we actually want the data validations to look going forward, so that as we add new cases, we can be doing it the Right Way. The factory wrapper for converting the old validation feels little opaque for this purpose.
>
> So it would be nice to have examples of both new `df_checks` and `field_checks` which are not tied to the old data validation implementation (though the underlying content could be drawn from those validations if we want)

Agreed - I think this would be a great follow-up! We should define those directly alongside the assets, I think, as mentioned above. Since `@asset_check(asset=foo)` will associate the function with the asset, we can define asset checks wherever we want and they'll get picked up. We'll just have to remember to run `load_asset_checks_from_module()` in `etl/__init__.py`. Might make sense to have those module-crawlers crawl everything automatically...

> Probably this would mean implementing at least one parametrized check class and migrating at least some of the existing data validations into that container completely, not just by reference to `pudl.validate`

Yeah, sounds right.

> Running the validations in the ETL will require more compute, but I think there's likely to be a big overall performance speedup for the validations, given that the way the validations are currently organized, the same dataframes are being read out of the DB and into memory repeatedly (since the tests are grouped by type of validation, not by which dataframe they apply to). This also results in memory usage blowing up if you try and run the data validations on a machine with lots of cores with pytest-xdist.
> 
> And then of course we'll also discover violations of our validation checks immediately in the ETL instead of in the nightly builds, which means we'll fix them sooner.

Yes! Agreed on these benefits :)

> IIRC by default failing asset checks don't cause the ETL to fail, do they? How will we be notified when a check fails? Do we want to set them to ERROR out on check failure? I think I saw that was a new option as of the most recent dagster release.

I think they don't cause the ETL to *stop running* but they do cause the whole run to be marked as a "failure". I can make a bogus failing test and run `pudl_etl` to see if things failed.

> Are there any checks that we can't encode this way? I guess any check (like referential integrity) that involves more than one table won't work.

Asset checks have [additional_ins](https://docs.dagster.io/_apidocs/asset-checks) which should let us encode checks across multiple tables.


I think the upshot of all this is that we need the following before we merge:
* [ ] check that a failing asset check does cause `pudl_etl` to exit with a non-zero status
* [ ] fix docstrings

And we want to encode some validations in-line instead of in-schema as a follow-up PR. For that one, we could even rip out the "df_checks" and "field_checks" machinery in this PR, and say:
* complicated checks live in `@asset_check`s that are defined alongside each asset
* simple dtype/min-max checks are auto-generated based on the field descriptions.

If we agree on ^, should we then rip out the df_checks and field_checks machinery in this PR pre-emptively?

src/pudl/etl/__init__.py
```diff
@@ -95,6 +103,61 @@
...
+    resource_id = asset_key.to_user_string()
+    try:
+        resource = package.get_resource(resource_id)
+    except ValueError:
+        return None
+    pandera_schema = resource.schema.to_pandera()
+
+    @asset_check(asset=asset_key)
```
zaneselvans, line 119: So this is the line that says "Asset X gets asset check Y?" I definitely didn't catch that the first time through.

This is efficient / concise but feels a little opaque. I think anything we can do to make the definition of new asset checks / validations intuitive and low-overhead will help us add lots of them incrementally over time, which will be good for data quality.  What are the different categories / modes of definition that we're going to enable?

- Autogenerated checks based on the `TableSchema` / SQL checks.
- Re-usable, parameterized classes of data validations that are specified in the Resource definition, and so also end up getting injected here (like the ones you created from our old data validation for the boiler fuel table, but with standalone check classes/parameters)
- Additional more bespoke per-asset checks which are defined close to the asset definition and added with an explicit `@asset_check` decorator on the asset?

Are there any others?

test/unit/metadata_test.py
```diff
@@ -88,3 +96,67 @@ def test_get_sorted_resources() -> None:
...
     ), f"{last_resource_name} is the last resource. Expected a resource with the prefix '_out'"
+
+
+def test_resource_descriptors_valid():
+    # just make sure these validate properly
+    descriptors = {
+        name: PudlResourceDescriptor.model_validate(desc)
+        for name, desc in RESOURCE_METADATA.items()
+    }
+    assert len(descriptors) > 0
```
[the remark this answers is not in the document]
jdangerx: The main failure case of this would be "the test hits a validation error." But, in case we really goof something up and `RESOURCE_METADATA.items()` is empty, this test would succeed... so I added some assertion that we really expect to be true all the time.

src/pudl/workspace/resource_cache.py
```diff
@@ -94,8 +94,9 @@ def _resource_path(self, resource: PudlResourceKey) -> Path:
 
     def get(self, resource: PudlResourceKey) -> bytes:
         """Retrieves value associated with a given resource."""
-        logger.debug(f"Getting {resource} from local file cache.")
-        return self._resource_path(resource).open("rb").read()
+        with self._resource_path(resource).open("rb") as res:
+            logger.debug(f"Getting {resource} from local file cache.")
+            return res.read()
```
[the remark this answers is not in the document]
jdangerx: Thank you!

src/pudl/validate.py
```diff
@@ -363,6 +364,26 @@ def vs_bounds(
             )
 
 
+def vs_bounds_checks(groups: dict[str, list[dict]]) -> list[pr.Check]:
+    """Turn case specs into Pandera checks."""
```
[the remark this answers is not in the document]
jdangerx: I think this is an argument to make that `BoundsCase` class.

src/pudl/metadata/classes.py
```diff
@@ -1001,6 +1058,69 @@ class ResourceHarvest(PudlMeta):
...
+
+    This is all very useful in that we can describe the resources more concisely!
+
+    TODO: In the future, we could convert from a :class:`PudlResourceDescriptor` to
+    various standard formats, such as a Frictionless resource or a :mod:`pandera`
+    schema. This would require some of the logic currently in :class:`Resource` to move
+    into this class.
+    """
+
+    class PudlSchemaDescriptor(PudlMeta):
```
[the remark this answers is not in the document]
jdangerx: IMO they don't really need to be accessed outside of the higher level class.

src/pudl/etl/__init__.py
jdangerx: I think that's right! I'm inclined to go one of two ways:

## Super Simple
* **only** autogenerate type checks from schema, and undo the `df_checks` and `field_checks` machinery in this PR
* all other type checks go into colocated `@asset_check` definitions, and clean up the `pudl.validations` module to make those re-usable data validations simple to add

Benefits being: less weird asset check meta-machinery to maintain. Costs being that checks such as "the 5th percentile of this column, weighted by this other column, should be less than X value" might be nice to see *in the schema itself*.

## Slightly more complicated
* keep the `df/field_checks` machinery, clean up `pudl.validations`
* complicated checks still live in colocated `@asset_check`s.

Benefits: for the most common cases, all the checks will live in the `RESOURCE_METADATA`
Costs: we have a bit more machinery to maintain.

I guess in either case we need to clean up pudl.validations a bit more. I think I slightly lean towards "super simple" because that's less custom logic for us to maintain.

zaneselvans: I'm not sure which of these arrangements sounds better to me. It would be nice if there were *some* centralized place where you can easily see all of the asset checks that are being applied to an asset. But maybe that's already available within the Dagster machinery somewhere? If so, then using that built-in functionality and avoiding building and maintaining something in parallel seems like a good idea.

One reason we haven't developed more and better validations, and haven't applied the same kinds of validations we do have for our older tables to the (hundreds of!) new ones we've created is that the current validation specification system is hard to work with, and with the change in caching (all data read from the DB every time, rather than being cached in memory inside `PudlTabl`) the validations are now very slow to run.

In converting the commonly applied validations that we've currently specified, it seems like we should be setting up a pattern that's easy for us to incrementally expand as we apply these kinds of validations to new tables, and also come up with new things to validate.  If it's low computational overhead (because we're applying them when the data is already in memory) and low human overhead (because it's easy to specify a new validation using the existing classes as templates) then we'll hopefully see a proliferation of validations, and having a clean pattern to accommodate them will help keep it from turning into spaghetti.

One problem that's going to come up here is that we'll expect different outcomes depending on whether we're running the full or fast ETL. E.g. looking at gas prices or coal heat content just for 2022 will yield a different expected distribution than looking at the last 20 years of data, so we'll either need to change the expectations to be more lenient and line up with all years / any individual year of data, or we'll need to be able to specify different asset check expectations for the fast vs. full ETL.

Translating all of the existing validations directly is also probably not ideal. I didn't really design them very well.  E.g. the min/max rows checks were originally meant to check that the number of rows were within a particular range, but now we really just check for an exact expected row count. I think there's also some unnecessary redundancy in the weighted distribution checks and some of them can probably just be discarded (I'm thinking of the `vs_self()` checks here). I think it may also be that we can re-use the same cumulative sums for calculating all of the weighted quantiles we're interested in, rather than recalculating them for every check, which could cut the time for doing these checks down by ~75%.

So... which of these many possible threads do we want to pull on right now? And which do we want to defer until we hopefully get some additional data validation / CI robustness support via POSE?

To me it feels like establishing an easy to use, extensible pattern for defining new data validations that makes use of Pandera and the asset checks, and starting to use that pattern for the new data we're integrating is a high priority (stop digging the janky data validation hole any deeper), but converting all of the existing jank into that new form might be lower priority, except insofar as those validations inform the needs we have for the new system.

jdangerx: > It would be nice if there were _some_ centralized place where you can easily see all of the asset checks that are being applied to an asset. But maybe that's already available within the Dagster machinery somewhere?

Yep! You can look in the `checks` tab in the asset catalog.

> To me it feels like establishing an easy to use, extensible pattern for defining new data validations that makes use of Pandera and the asset checks, and starting to use that pattern for the new data we're integrating is a high priority (stop digging the janky data validation hole any deeper), but converting all of the existing jank into that new form might be lower priority, except insofar as those validations inform the needs we have for the new system.

Agreed. I think that looks like utility methods in `pudl.validate` that make it easy to create asset checks for:
* row count
* weighted distribution - tail size, but also "peaks at X, Y, Z"
* no missing years of data, based on the dataset settings


I think the todo list then is:

* [ ] rip out the `df_checks` and `field_checks` machinery (though, keeping the frictionless compatibility stuff with uniqueness/min-max)
* [ ] merge this PR
* [ ] add these utility methods as a separate PR & use them to define a few checks

zaneselvans: ## Decisions
* Agree let's keep all of the Frictionless / schema / datapackage compatible validations specified in the schema / Resource area, and move the non-schema checks to live near the assets they apply to.
* With that, the threshold for getting this PR merged is just that all of the schema validation checks are being run, and we make a new PR to set up the classes and example structures for the existing re-usable non-schema checks.
* An additional reusable non-schema check that we do on many tables and should probably do on many more is verify that there are no entirely NULL columns. There's a `pudl.validate.no_null_cols()` function that does this I think, which should be added to the list of utility methods to parameterize.

## Thoughts & Questions
* Checking for the exact expected row counts should be extremely fast if we just use values from `df.shape`
* Agree no missing years (or maybe more precisely, no missing partitions?) is important. I think we've had some issues with this recently in the `eia860m`, where the settings and archive were updated to include the new months of data, but the `working_partitions` associated with the data source were not updated, so none of the new data actually made it through the ETL.
* There are also some cases in which we set an allowable missignness threshold for rows, via `pudl.validate.no_null_rows()` but I think that's only applied in the case of the MCOE outputs, so maybe we don't need to parameterize it right now.
* Should/Can the schema-based asset checks ensure uniqueness across the set of PK columns when there is a PK defined? It would be great to identify that kind of problem as soon as it appears rather than only later in the DB, and getting those constraints into the asset checks would make them independent of the storage medium, so if we were to only output Parquet, they'd still be applied.
* The ability to define asset checks that rely on more than one asset as inputs makes me wonder about creating a generic, storage-independent referential integrity check -- if we can check that FK relationships are satisfied even when we're writing out to Parquet or BigQuery that would be great.

jdangerx: Seems like we can validate joint uniqueness by taking the `primary_key` and shoving it into the Pandera schema's `unique` constructor param:

https://pandera.readthedocs.io/en/stable/dataframe_schemas.html#validating-the-joint-uniqueness-of-columns

jdangerx: Oh, I actually already do that in this PR. No changes necessary!

jdangerx:
Made a "minimal checks" issue here: https://github.com/catalyst-cooperative/pudl/issues/3412 - @zaneselvans this should be ready for re-review!

zaneselvans (approved):

Issue #3224, opened 2024-01-09
Add resource and field descriptions; purge unused fields


zaneselvans:
While working on Parquet outputs in #3102 we discovered a number of unused field definitions, fields with no descriptions, and resources with no descriptions.

```[tasklist]
- [x] Ensure all resources have complete and accurate descriptions.
- [x] Purge all unused field definitions.
- [x] Enable the integration test that verifies all defined fields are being used.
- [x] Ensure all remaining fields have complete and accurate descriptions.
- [x] Update our Pydantic `Field` and `Resource` types to enforce that `description` is a non-empty string.
- [x] Update the unit tests to reflect the new `Field` and `Resource` requirements
```

```py
UNUSED_FIELDS = [
    "active",
    "country",
    "credits_or_adjustments",
    "delivery_customers",
    "depreciation_amortization_value",
    "electric_plant",
    "energy_source",
    "environmental_equipment_name",
    "expense",
    "fuel_transportation_mode",
    "future_plant",
    "income",
    "is_total",
    "leased_plant",
    "line_id",
    "month",
    "notes",
    "operator_name",
    "operator_state",
    "operator_utility_id_eia",
    "other",
    "other_total",
    "owner_name",
    "peak_demand_summer_mw",
    "peak_demand_winter_mw",
    "period_nox",
    "period_particulate",
    "period_so2",
    "prime_mover",
    "retail_sales",
    "sales_for_resale",
    "status",
    "storage_capacity_mw",
    "storage_customers",
    "total",
    "total_meters",
    "transmission",
    "unbundled_revenues",
    "utility_attn",
    "utility_pobox",
    "virtual_capacity_mw",
    "virtual_customers",
]

RESOURCES_NO_DESC = [
    "core_eia861__yearly_demand_side_management_misc",
    "core_eia861__yearly_demand_side_management_sales",
    "core_eia861__yearly_distributed_generation_fuel",
    "core_eia861__yearly_distributed_generation_misc",
    "core_eia861__yearly_distributed_generation_tech",
    "core_eia861__yearly_net_metering_customer_fuel_class",
    "core_eia861__yearly_net_metering_misc",
    "core_eia861__yearly_non_net_metering_customer_fuel_class",
    "core_eia861__yearly_non_net_metering_misc",
    "core_eia861__yearly_operational_data_misc",
    "core_eia861__yearly_operational_data_revenue",
    "core_eia861__yearly_reliability",
[...]

PR #3283, opened 2024-01-24
Resource field description cleanup

aesharpe:
# This PR addresses the fixes laid out in #3224

* Remove unused fields
* Add descriptions to all fields
* Make descriptions mandatory for `Field` and `Resource` instances
* Update unit tests to reflect mandatory descriptions
* Update the docstring examples to reflect mandatory descriptions
* Remove the x-fail from the `test_defined_fields_are_used ` test so we make sure all fields are being used
* Add a new alembic migration for altering the metadata

## Notes on unused fields that were removed:
(I went through and checked when each description-less field was last edited to see what might have happened)

* `active` no record of addition
* `country` changed to owner_country out_eia860__yearly_ownership table from the ownership_eia860 table
* `credits_or_adjustments` no record of addition
* `delivery_customers` no record of addition
* `depreciation_amortization_value` from f1_dacs_epda table - I think changed to "dollar_value" in the out_ferc1__yearly_depreciation_summary_sched336 table
* `electric_plant` no record of addition 
* `energy_source` no record of addition 
* `environmental_equipment_name` from the raw_eia923__emissions_control table which hasn't been transformed yet
* `expense` changed to dollar_value in the core_ferc1__yearly_operating_expenses_sched320 table
* `fuel_transportation_mode` no record of addition
* `future_plant` no record of addition
* `income` changed to dollar_value in the core_ferc1__yearly_income_statements_sched114 table
* `is_total` used to be used to identify total rows in ferc1 tables
* `leased_plant` no record of addition
* `line_id` no record of addition
* `month` no record of addition
* `notes` was used to show any notes extracted from ferc1 tables
* `operator_name` not exactly sure, maybe dropped from eia860 ownership out_ table? 
* `operator_state` not exactly sure, maybe dropped from eia860 ownership out_ table? 
* `operator_utility_id_eia` not exactly sure, maybe dropped from eia860 ownership out_ table? 
* `other` no record of addition
* `other_total` no record of addition
* `owner_name` owner_name
* `peak_demand_summer_mw` old ferc714 table description_pa_ferc714
* `peak_demand_winter_mw` old ferc714 table description_pa_ferc714
* `period_nox` from the raw_eia860__boiler_info table that gets pulled into the _core_eia860__boilers table but I guess the columns get dropped
* `period_particulate` from the raw_eia860__boiler_info table that gets pulled into the _core_eia860__boilers table but I guess the columns get dropped idk if it's intentional?
* `period_so2` from the raw_eia860__boiler_info table that gets pulled into the _core_eia860__boilers table but I guess the columns get dropped idk if it's intentional?
* `prime_mover` no record of addition
* `retail_sales` no record of addition`
* `sales_for_resale` no record of addition
* `status` no record of addition
* `storage_capacity_mw` no record of addition
* `storage_customers` no record of addition
* `total` no record of addition
* `total_meters` no record of addition
* `transmission` no record of addition
* `unbundled_revenues` no record of addition
* `utility_attn` no record of addition
* `utility_pobox` no record of addition
* `virtual_capacity_mw` no record of addition, probably from pre-normalized EIA861 data
* `virtual_customers` no record of addition, probably from pre-normalized EIA861 data

### High priority fixes:
- [ ] `business_model` this is a weird column in core_eia861__yearly_sales which contains a letter I'm not sure if we've coded correctly.
- [ ] `fuel_pct` kind of funky carve out.
- [x] All peak_reduction columns should be in MW not MWH! (this means fixing the actual column name which is a bit of a pain

### Low priority fixes (probably out of scope):
- [ ] `demand_mwh` Could be more specific: add "hourly_demand"?
- [ ] `incremental_energy_savings_mwh` Appears to align with old column "energy_efficiency_incremental_effects_mwh" - maybe we want them to have the same name? (from 2012 split from DSM table to EE/DR tables).
- [ ] `incremental_peak_reduction_mw` Appears to align with old column "energy_efficiency_incremental_actual_peak_reduction_mw" - maybe we want them to have the same name? (from 2012 split from DSM table to EE/DR tables).
- [ ] `respondent_type` Maybe this should be more specific like "respondent_type_ferc714"
- [ ] `utc_datetime` There's another field similar to this - investigate whether they are the same thing or not. On a similar note: it might be worth it to go through some of the fields and make sure there are no duplicates. I feel like there are a lot of categorical fields like "respondent type" etc. that are very similar across data sources. We should either make them more specific by adding data source suffixes or make them more generic and apply to all tables. There's definitely an opportunity to go through some of the fields and prune a bit, but it's not a small task.
- [ ] `utility_owned_capacity_mw` Could make this col name more specific to non-net metered capacity

### Low priority fixes (probably out of scope)
- [ ] Come up with standard format for resource descriptions (at least within a given source) so there is some standardization.
- [ ] It looks like we currently don't integrate the EIA861 `dispersed_generation` tables (part of the raw `distributed_generation` spreadsheets). Is that something we want to do? (distributed = grid connected, dispersed = not grid connected).


aesharpe:
One question I have is how or whether we will create descriptions for the output tables (all tables with the `out` prefix). Right now it seems like the only table metadata we have exists in for `core` normalized tables.

zaneselvans:
I think given that we intend the `out_` tables to the be the ones that are most commonly used / user friendly, we'll want to make sure they all have descriptions. Though these could be very similar or even identical to the descriptions of the corresponding `core_` table if they're conceptually the same.

src/pudl/metadata/resources/eia861.py
```diff
@@ -388,6 +483,8 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_net_metering_misc": {
+        # TODO: I feel skeptical that the pv_current_flow_type field shouldn't be linked to the other net metering table.
```
zaneselvans, line 486: Why did it get split out? Agree it feels a little funny to have a single data column over there all lonely.  If it were consolidated, would it be weird that the "non-net metering" data has a misc table, but the net metering doesn't?

src/pudl/metadata/resources/eia861.py
```diff
@@ -439,6 +555,14 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_operational_data_misc": {
+        # TODO: misc might be a missleading name
+        "description": (
+            """The annual megawatt hours (MWH) for all a utility's sources of
```
zaneselvans, line 560: We've got a couple of electricity "sources and dispositions" tables in the FERC Form 1. Would it be easy to name this table so that it's clearly analogous? Would be interesting to compare them:

- `core_ferc1__yearly_energy_sources_sched401`
- `core_ferc1__yearly_energy_dispositions_sched401`

Maybe:

- `core_eia861__yearly_energy_sources_dispositions`

It seems a little confusing to list the sources, but not the dispositions in the description.  I think it's fine to not enumerate them at all, since if a user knows it's a sources & dispositions table, the column names are readily interpretable, and there's kind of a lot of them.

src/pudl/metadata/resources/eia861.py
```diff
@@ -476,6 +600,13 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_operational_data_revenue": {
```
zaneselvans, line 602: This is another table I think we might rename along the lines of FERC 1 to make their similarity clear. Given that we have `out_ferc1__yearly_operating_revenues_sched300` maybe

- `out_eia861__yearly_operating_revenues`

src/pudl/metadata/resources/eia861.py
```diff
@@ -492,6 +623,12 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_reliability": {
+        "description": (
+            """The system average interruption duration index (SAIDI), system average
```
zaneselvans, line 627: A first sentence just describing this table conceptually would be helpful, after which we can list the specific metrics as you have here. Maybe:

```
"""Standardized metrics of electricity system reliability and outage impacts.

Includes the system average interruption duration index (SAIDI), system average
interruption frequency index (SAIFI), and customer average interruption duration index
(CAIDI) aka SAIDI/SAIFI with and without major event days and loss of service. Includes
the standard (IEEE/other) and other relevant information."""
```

src/pudl/metadata/resources/eia861.py
```diff
@@ -599,6 +736,12 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_utility_data_misc": {
```
zaneselvans, line 738: A table rename that suggests itself to me here is `core_eia861__yearly_utility_activities`.

Maybe:
```
"""A table of boolean values indicating what kind of business activities each utility engages in.

For example, do they provide transmission, distribution, or generation services? Wholesale or retail marketing? Do they have an alternative fuel vehicle program?
"""
```

(there are a couple of typos in the current description, "includeing" and "bundeled")

src/pudl/metadata/resources/eia861.py
```diff
@@ -627,6 +770,7 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_utility_data_nerc": {
```
zaneselvans, line 772: Feels like `core_eia861__yearly_utility_nerc_regions` would be more descriptive as a table name.

src/pudl/metadata/resources/eia861.py
```diff
@@ -642,6 +786,7 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_utility_data_rto": {
```
zaneselvans, line 788: `core_eia861__yearly_utility_rtos`? `core_eia861__yearly_utility_iso_rto`?

src/pudl/metadata/resources/eia861.py
```diff
@@ -642,6 +786,7 @@
...
+        "description": "The RTOs that a utility operates in.",
```
zaneselvans, line 789: Is it really just RTOs? Or does EIA-861 just default to using RTO and not mention ISO?  I guess I'm still not 100% sure what the difference is between an ISO and an RTO if there is one.

zaneselvans (changes requested):
Mostly the resource descriptions look good! I made a few suggestions.

We should remove the `xfail` from the two unit tests that check whether all fields and resources have descriptions (they're currently XPASS in the unit tests, so seems like they should pass) or potentially remove them entirely since we're now requiring descriptions.

It does seem like there are some table name clarifications / simplifications that we would be good to do here while we're still cleaning up old names. @bendnorman what do you think about getting this into the Great Rename?  The old EIA-861 table names were kind of just pulled from the EIA-861 forms / spreadsheets without too much intention since they were "interim" 😆.

@aesharpe would you be willing to put your detailed per-field "suggested fixes" in context in the `fields.py` file as a self-review?  I don't have enough context to understand what's going on with all those fields and would end up trying to transfer all the comments myself to see what's going on, but you've got more context on them.

I have not yet reviewed `fields.py`

zaneselvans:
(converting to draft so the `ci-integration` doesn't cost us money while there's some additional back-and-forth)

test/unit/metadata_test.py
```diff
@@ -56,7 +56,6 @@ def test_field_definitions(field_name: str):
     _ = Field(name=field_name, **FIELD_METADATA[field_name])
 
 
-@pytest.mark.xfail(reason="Need to purge unused fields. See issue #3224")
```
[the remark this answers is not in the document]
zaneselvans: I went ahead and removed these `xfail` marks below just to verify that they pass and they do.

aesharpe:
@zaneselvans renaming tables is a bit more disruptive than just updating descriptions etc. Just want to confirm that it's ok to go ahead and update some of these table names as per your/my suggestions.

aesharpe:
> We should remove the `xfail` from the two unit tests that check whether all fields and resources have descriptions (they're currently XPASS in the unit tests, so seems like they should pass) or potentially remove them entirely since we're now requiring descriptions.

I think it makes sense to remove them because of the required description field.

src/pudl/metadata/resources/eia861.py
```diff
@@ -74,7 +74,18 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_demand_response": {
```
bendnorman, line 76: @zaneselvans we should rename them if we think the names don't accurately describe the contents of the table. However, IDK if we have the hours for it this quarter. 

I don't think it'd be the worst thing to do a data release with the current names and then update them in the future. The EIA 861 tables require more modeling/cleaning work, correct? Maybe we can update the names then?

zaneselvans: Yes, they'll definitely require more work if we want them to be comparable in normalization and consistency to the other EIA data. I think there are a handful of obvious changes here that will probably not be disruptive and will make the table names meaningful.  Let's defer anything that's not obvious. I don't think we can really get away from the fact that tables and columns will change over time, but I don't know how to gracefully deal with that kind of schema drift either, beyond offering stable releases so folks can choose when to migrate.

bendnorman: Ok! After looking at the table names more closely, I agree there are some easy fixes we can tackle in this PR.

src/pudl/metadata/fields.py
```diff
@@ -1092,7 +1373,10 @@
         "description": "Heat content of the fuel in millions of Btus per physical unit.",
         "unit": "MMBtu_per_unit",
     },
-    "fuel_pct": {"type": "number"},
+    "fuel_pct": {
+        "type": "number",
+        "description": "Percent of fuel",
+    },
```
aesharpe, line 1379: This field comes from the `core_eia861__yearly_distributed_generation_fuel ` table. In 2010, the form switched from reporting percent of total capacity to mw of capacity for each individual technology type. In the transform module, we convert all the pre-2010 pct_capacity values to mw of capacity so we can have a uniform column of values throughout the years. 

The form stops reported fuel_pct values in 2007. This means that there was no incentive to convert this column to fuel_mw (because there was no post-2010 fuel_mw columns to combine it with). 

This leaves me with 2 questions: 
1) The fuel_pct columns represent the percent of total capacity by fuel. We split this table off from the `_tech` and `_misc` tables so that we could normalize the data (i.e.: make `fuel_class` and `fuel_pct` columns instead of having individual columns for each fuel type). However, the `capacity_mw` field, that all the percentages are based on, lives in the `core_eia861__yearly_distributed_generation_tech` table. How should we solve this issue? It feels like the capacity field should be in the same table, but the whole point of normalization is to de-duplicate data for storage. It also doesn't necessarily make sense to recombine the tables, because that will result in tons of columns because we will have to make it a wide table again. 

2) Should we bother converting this `fuel_pct` to `capacity_mw` to match the newer `tech_pct` data that has been converted?

Will update the description for this field once we have more clarity on what to do here.

It helps to look at the column maps for the `distributed_generation_eia861` table.

src/pudl/metadata/fields.py
```diff
@@ -1257,31 +1561,77 @@
...
+    "incremental_life_cycle_peak_reduction_mwh": {
+        "type": "number",
+        "description": (
+            "The estimated total changes in peak load for incremental programs and "
+            "participants over the life of the programs. DSM programs have a useful "
+            "life, and the net effects of these programs will diminish over time. "
+            "Considers the useful life of energy efficiency technology by accounting "
+            "for building demolition, equipment degradation, and program attrition."
+        ),
+        "unit": "MWh",
```
aesharpe, line 1624: This field comes from the `core_eia861__yearly_energy_efficiency ` table. Peak demand reduction should be reported in MW not MWh. I think this is just a column name error and we should change it to be `incremental_life_cycle_peak_reduction_mw`

src/pudl/metadata/fields.py
```diff
@@ -2309,19 +2796,44 @@
     },
     "respondent_type": {
         "type": "string",
+        "description": (
+            "Whether a respondent to the FERC form 714 is a utility or a balancing "
+            "authority."
+        ),
```
aesharpe, line 2802: This is pretty specific to 714, maybe we should add a `_714` suffix to the column?

src/pudl/metadata/fields.py
```diff
@@ -2805,8 +3474,10 @@
         "type": "boolean",
         "description": "Was an uprate or derate completed on this generator during the reporting year?",
     },
-    "utc_datetime": {"type": "datetime"},
-    "utility_attn": {"type": "string"},
+    "utc_datetime": {
+        "type": "datetime",
+        "description": ("Date and time converted to Coordinated Universal Time (UTC)."),
+    },
```
aesharpe, line 3480: I think there's another field similar to this called `operating_datetime_utc`, maybe we should investigate how similar they are. Also, I'm blanking but is there a reason we don't just convert all dates to utc? Or do we? 

This reminds me that there is definitely room to do a more thorough review of columns and potential column duplication, but it's out of scope right now.

src/pudl/metadata/fields.py
```diff
@@ -1257,31 +1561,77 @@
...
-    "incremental_life_cycle_peak_reduction_mwh": {"type": "number", "unit": "MWh"},
-    "incremental_peak_reduction_mw": {"type": "number", "unit": "MW"},
+    "incremental_energy_savings_mwh": {
+        "type": "number",
+        "description": (
+            "energy savings in the given report year resulting from new participants "
+            "in existing demand response programs and all participants in new demand "
+            "response programs."
+        ),
+        "unit": "MWh",
```
aesharpe, line 1602: This field is from the `core_eia861__yearly_energy_efficiency ` table, the offspring table of the pre-2012 `core_eia861__yearly_demand_side_management_ee_dr ` table. We talked to EIA about whether we could or should merge these tables, and they said no. However, this column is suspiciously similar to the `energy_efficiency_incremental_effects_mwh` from the newer table. Maybe we should investigate whether they should have the same name at least? 

This also goes for the `incremental_peak_reduction_mw` and the `energy_efficiency_incremental_actual_peak_reduction_mw` columns.

src/pudl/metadata/fields.py
```diff
@@ -591,7 +773,11 @@
         "description": "Demand charges (USD).",
         "unit": "USD",
     },
-    "demand_mwh": {"type": "number", "unit": "MWh"},
+    "demand_mwh": {
+        "type": "number",
+        "description": "Demand in a given time frame.",
+        "unit": "MWh",
+    },
```
aesharpe, line 780: This field comes from the `out_ferc714__hourly_predicted_state_demand `, maybe we should clarify in the column name that it's hourly demand? Or maybe the timestamps make it obvious.

src/pudl/metadata/fields.py
```diff
@@ -2841,12 +3512,16 @@
         "type": "string",
         "description": "Utility name, chosen arbitrarily from the several possible utility names available in the utility matching process. Included for human readability only.",
     },
-    "utility_owned_capacity_mw": {"type": "number", "unit": "MW"},
+    "utility_owned_capacity_mw": {
+        "type": "number",
+        "description": "Total non-net-metered capacity owned by the respondent.",
+        "unit": "MW",
+        # TODO: Make this column name more specific to non-net metered capacity.
+    },
```
aesharpe, line 3520: This column comes from the `core_eia861__yearly_non_net_metering_misc` table, maybe we should clarify in the column name that it's from non-net_metered capacity?

src/pudl/metadata/resources/eia861.py
```diff
@@ -184,6 +223,16 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_demand_side_management_sales": {
+        "description": (
+            """Electricity sales for resale and to ultimate customer.
+
```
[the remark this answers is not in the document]
aesharpe: This is interesting. The more I dig into this the more I'm confused why these columns are in the DSM table to begin with.

They're clearly in the raw data, but when you look at the actual 861 Form from 2009 (the last year these columns show up in this table) there is no mention of sales data in Schedule 6 (DSM) -- granted, the PDF gets cut off...

Screen Shot 2024-01-25 at 5 00 17 PM

Rather, it shows up in Schedule 2 and Schedule 3 which correspond to the Operational Data table. I wonder if they are the same or not.

src/pudl/metadata/resources/eia861.py
aesharpe: I think they are effectively the same thing, RTOs are just bigger.

src/pudl/metadata/resources/eia861.py
aesharpe: Ok! I'll tackle the suggested table name changes. I'm sure there is a lot of nit-picking we can do to these tables at a later time.

src/pudl/metadata/resources/eia861.py
```diff
@@ -216,6 +274,18 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_distributed_generation_misc": {
+        # TODO: might want to rename this table to be _capacity
```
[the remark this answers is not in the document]
aesharpe: See my comment below https://github.com/catalyst-cooperative/pudl/pull/3283#discussion_r1467001741 before I fix this.

src/pudl/metadata/resources/eia861.py
aesharpe: We could rename this table, but it would distance the table from the original EIA861 table name (`Operational Data`). Which is fine, but it make make it more confusing for folks that have previously worked with the raw data.

Are you referring to both of the `operational_data` tables here or just the `misc` one?

src/pudl/metadata/resources/eia861.py
aesharpe: See comment from above

src/pudl/metadata/resources/eia861.py
aesharpe: I like this description, but it's not about whether the utility has alternative fuel vehicle programs, the form actually says "Did Your Company Operate Alternative-Fueled Vehicles During the Year?"

src/pudl/metadata/resources/eia861.py
aesharpe: On second thought, I kind of agree with @bendnorman here. Renaming tables opens a can of worms and makes it really hard not to dig into the content of the tables more meaningfully. I think we are due for an 861 revamp, but it deserves a bit more time/effort, and I don't think it makes sense to dedicate any time to doing that until we can do it in full.

src/pudl/metadata/resources/eia861.py
```diff
@@ -127,7 +138,21 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_demand_side_management_ee_dr": {
```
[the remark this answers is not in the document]
aesharpe: energy efficiency and demand response

src/pudl/metadata/resources/eia861.py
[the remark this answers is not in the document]
aesharpe: To be addressed in #3301

src/pudl/metadata/resources/eia861.py
aesharpe: To be addressed in #3301

src/pudl/metadata/resources/eia861.py
aesharpe: To be addressed in #3301

src/pudl/metadata/resources/eia861.py
aesharpe: To be addressed in #3301

src/pudl/metadata/resources/eia861.py
aesharpe: To be addressed in #3301

src/pudl/metadata/resources/eia861.py
aesharpe: To be addressed in #3301

src/pudl/metadata/resources/eia861.py
aesharpe: Using #3301 to collect table name and structure change information.

src/pudl/metadata/resources/eia861.py
```diff
@@ -294,7 +373,18 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_energy_efficiency": {
```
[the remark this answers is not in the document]
aesharpe: To be addressed in #3301

src/pudl/metadata/resources/eia861.py
```diff
@@ -492,6 +623,12 @@
         "etl_group": "eia861",
     },
     "core_eia861__yearly_reliability": {
```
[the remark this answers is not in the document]
aesharpe: To be addressed in #3301

src/pudl/metadata/fields.py
zaneselvans: Added to #3301

zaneselvans (changes requested):

zaneselvans:
@aesharpe I think there's only a handful of things left that are unclear and that we want to try and address in this PR. I made some small wording suggestions / typo fixes, fixed the one truly wrong column name (MWh => MW) and gathered together the issues we're punting until later in #3322.

src/pudl/metadata/fields.py
```diff
@@ -217,23 +268,61 @@
...
     "bypass_heat_recovery": {
         "type": "boolean",
         "description": "Can this generator operate while bypassing the heat recovery steam generator?",
     },
-    "caidi_w_major_event_days_minutes": {"type": "number", "unit": "min"},
+    "caidi_w_major_event_days_minutes": {
+        "type": "number",
+        "description": (
+            "Average number of minutes per interuption (SAIDI/SAIFI) including major "
```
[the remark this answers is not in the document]
aesharpe: found a few of these haha

src/pudl/metadata/fields.py
```diff
@@ -717,33 +916,103 @@
         "description": "Energy charges (USD).",
         "unit": "USD",
     },
-    "energy_displaced_mwh": {"type": "number", "unit": "MWh"},
+    "energy_displaced_mwh": {
+        "type": "number",
+        "description": "Amount of energy displaced from net metering.",
```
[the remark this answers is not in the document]
aesharpe: This is such a weird blip of data that only shows up in 2007-2009. I think it refers to energy fed back into the grid (i.e., energy the utility does not need to produce)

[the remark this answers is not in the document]
aesharpe: Actually, I think it is the same as `sold_to_utility_mwh` from later years (which gets split out by technology type) - I might change this

src/pudl/metadata/resources/eia861.py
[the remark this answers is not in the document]
aesharpe: I think we can leave this for a larger discussion about EIA861 table normalization like you have laid out in #3324.

src/pudl/package_data/eia861/column_maps/net_metering_eia861.csv
```diff
@@ -4,11 +4,6 @@ utility_id_eia,1,1,1,1,1,1,1,1,1,1,1,1,1,1,2,2,2,2,2,2,2
 utility_name_eia,2,2,2,2,2,2,2,2,2,2,2,2,2,2,3,3,3,3,3,3,3
 short_form,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,4,-1,-1,-1
 state,3,3,3,3,3,3,3,3,3,3,3,3,3,3,1,1,1,1,1,1,1
-residential_total_energy_displaced_mwh,-1,-1,-1,-1,-1,9,9,9,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1
-commercial_total_energy_displaced_mwh,-1,-1,-1,-1,-1,10,10,10,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1
-industrial_total_energy_displaced_mwh,-1,-1,-1,-1,-1,11,11,11,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1
-transportation_total_energy_displaced_mwh,-1,-1,-1,-1,-1,12,12,12,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1
-total_total_energy_displaced_mwh,-1,-1,-1,-1,-1,13,13,13,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1
```
zaneselvans, line 11: You feel pretty confident about this reorganization based on the form documentation?

zaneselvans (approved):
If you feel pretty confident about the "total energy displaced" and "sold to utility" re-org then I think this is good to go assuming all the tests pass.

src/pudl/package_data/eia861/column_maps/net_metering_eia861.csv
aesharpe: I do - the documentation says: "Net Metering programs allow customers to sell excess power they generate back to the electrical grid to offset consumption" and then lists the "Net Metering Displaced Energy (MWh)" for data entry.

Screen Shot 2024-01-30 at 1 32 10 PM

aesharpe: And there is no overlap between this "energy_displaced" column and the "sold_to_utiliy" column

src/pudl/metadata/resources/eia861.py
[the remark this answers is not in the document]
zaneselvans: Okay I added a link to this thread to #3224 for context.

PR #3286, opened 2024-01-24
Add issue template for nightly build

### LINKED ISSUE #3284: Nightly build failure 2024-01-24
# Data validation failure
The utility and balancing authority service territories derived from EIA-861 have significantly more records than expected.

* All of the real code changes between `nightly-2024-01-21` and `nightly-2024-01-24` seem to be related to the PHMSA extraction.
* However, the `conda` environment was also re-locked on Monday. Maybe something there changed the behavior of the service territory assets?

```
test/validate/service_territory_test.py::test_minmax_rows[compiled_geometry_balancing_authority_eia861-108436] 
-------------------------------- live log call ---------------------------------
2023-08-15 14:53:49 [    INFO] catalystcoop.pudl.validate:194 compiled_geometry_balancing_authority_eia861: found 112507 rows, expected 108436. Off by 3.754%, allowed margin of 0.000%
FAILED                                                                   [ 99%]
test/validate/service_territory_test.py::test_minmax_rows[compiled_geometry_utility_eia861-237872] 
-------------------------------- live log call ---------------------------------
2023-08-15 14:53:52 [    INFO] catalystcoop.pudl.validate:194 compiled_geometry_utility_eia861: found 247705 rows, expected 237872. Off by 4.134%, allowed margin of 0.000%
```

# Build script success criteria:

* In addition to the above issue, despite the fact that the **build failed** the script seems to think it succeeded in some ways.
* The `nightly` branch was updated.
* The `success` file was created in the output bucket
* The `/nightly` outputs were updated with last night's outputs.
* It seems like we aren't correctly capturing the exit status of the chain of commands in the `run_pudl_etl()` funciton. Maybe we should break them up into individual commands each with their own success variable, and have the function return the AND of all of them?

# Datasette doesn't match?

Curious whether the datasette deployment had taken place given the above combination of failure and perceived success, I counted the rows in the two affected tables, and to my surprise, it matched **neither** the expected or observed row counts above:

```sql
SELECT COUNT(*) FROM out_eia861__compiled_geometry_utilities;
-- 248987
```

```sql
SELECT COUNT(*) FROM out_eia861__compiled_geometry_balancing_authorities;
-- 112853
```

# ETL Logs:
[pudl-etl.log](https://github.com/catalyst-cooperative/pudl/files/14039695/pudl-etl.log)

jdangerx:
# Overview

Closes #3284 .

There was no action needed to resolve #3284, except for verifying that these things did indeed happen. I'm adding an issue template so we can handle these more smoothly in the future!

# Testing

I'm not actually sure how to test if issue templates are properly formatted, but we can verify this once merged by trying to create an issue!

```[tasklist]
# To-do list
- [ ] verify post-merge
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/nightly_build_failure.md
+++ b/.github/ISSUE_TEMPLATE/nightly_build_failure.md
@@ -1,0 +1,31 @@
+---
+name: Nightly build failure report
+about: Lightweight incident report for nightly build failure
+title: Nightly Build Failure XXXX-XX-XX
+labels:
+  - bug
+  - nightly-builds
+assignees: ''
+---
+
+# Overview
+
+What do you think happened, if anything?
+
+What next steps do we need to do?
+
+```[tasklist]
+# Next steps
+- [ ] verify that [S3 distribution bucket](https://s3.console.aws.amazon.com/s3/buckets/pudl.catalyst.coop?region=us-west-2&bucketType=general&prefix=nightly/&showversions=false) was updated at the expected time
+- [ ] verify that [GCP distribution bucket](https://console.cloud.google.com/storage/browser/pudl.catalyst.coop/nightly;tab=objects?project=catalyst-cooperative-pudl) was updated at the expected time
+- [ ] verify that [GCP internal bucket](https://console.cloud.google.com/storage/browser/builds.catalyst.coop) was updated at the expected time
+- [ ] verify that [Datasette PUDL version](https://data.catalyst.coop/pudl/core_pudl__codes_datasources) points at the same hash as [nightly](https://github.com/catalyst-cooperative/pudl/tree/nightly)
+- [ ] verify that [Zenodo sandbox record](https://sandbox.zenodo.org/doi/10.5072/zenodo.5563) was updated to the record number in the logs (search for `zenodo_data_release.py` and `Using metadata from` in the logs, to see what the new record number should be!)
+```
+
+
+# Relevant logs
+[link to build logs from internal distribution bucket]()
+
+```
+```
--- a/.github/ISSUE_TEMPLATE/nightly_build_failure.md
+++ b/.github/ISSUE_TEMPLATE/nightly_build_failure.md
@@ -18,16 +18,18 @@ What next steps do we need to do to understand or remediate the issue?
 
 # Verify that everything is fixed!
 
+Once you've applied any necessary fixes, make sure that the nightly build outputs are all in their right places.
+
 ```[tasklist]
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/nightly_build_failure.md
```diff
@@ -0,0 +1,31 @@
+---
+name: Nightly build failure report
+about: Lightweight incident report for nightly build failure
+title: Nightly Build Failure XXXX-XX-XX
+labels:
+  - bug
+  - nightly-builds
+assignees: ''
```
jdangerx, line 8: maybe worth auto-assigning inframundo?

e-belfer: Yes 100% worth auto-assigning Inframundo. We can always reassign.

.github/ISSUE_TEMPLATE/nightly_build_failure.md
```diff
@@ -0,0 +1,31 @@
...
+
+# Overview
+
+What do you think happened, if anything?
+
+What next steps do we need to do?
+
+```[tasklist]
+# Next steps
+- [ ] verify that [S3 distribution bucket](https://s3.console.aws.amazon.com/s3/buckets/pudl.catalyst.coop?region=us-west-2&bucketType=general&prefix=nightly/&showversions=false) was updated at the expected time
```
e-belfer, line 19: These steps seem more oriented towards "make sure the updates failed as expected / clean up after failure" and not "identify the source of the failure and fix it", which is the intended goal of this issue template. I'm a bit confused about why we'd want to check them here, rather than make an issue to address any existing issues with the nightly build failures working as expected, if that makes sense?

e-belfer (commented):
Left a question.

.github/ISSUE_TEMPLATE/nightly_build_failure.md
jdangerx: My thought was basically that we should *at least* always check this, and then any actual investigative tasks should be added by the creator.

jdangerx:
Updated this to hopefully be a bit clearer on what to do.

.github/ISSUE_TEMPLATE/nightly_build_failure.md
```diff
@@ -0,0 +1,33 @@
...
+
+# Overview
+
+What do you think happened, if anything?
+
+# Next steps
+
+What next steps do we need to do to understand or remediate the issue?
+
+# Verify that everything is fixed!
```
e-belfer, line 19: This would be the night after, after pushing the fix? Or after kicking off another build?

e-belfer (approved):
One more question for you, but it's non-blocking.

PR #3289, opened 2024-01-24
Experiment tracking

zschira:
## Background
This PR adds basic experiment tracking functionality to PUDL's record linkage models. It uses [mlflow](https://mlflow.org/docs/latest/tracking.html) for tracking, and provides useful tools for creating new experiments and logging model configuration. As of now this does not add functionality to deploy a cloud DB to log `mlflow` runs remotely, so this will be most useful for individual development.

## Further work
This base set of functionality could be expanded upon to use `mlflow` to efficiently tune hyper-parameters of a model, but this is not contained in the initial PR.


src/pudl/analysis/record_linkage/model_helpers.py
```diff
@@ -1,18 +1,134 @@
...
 
-    if not (model_config := config.get(model_key)):
-        raise RuntimeError(f"No {model_key} entry in {config_file}")
+    if not (model_config := config.get(experiment_name)):
+        raise RuntimeError(f"No {experiment_name} entry in {config_file}")
+
+    return {experiment_name: model_config}
+
+
+def flatten_model_config(model_config: dict) -> dict:
```
zschira, line 30: This converts a nested dictionary of configuration and turns it into a flat dictionary which works better in the `mlflow` UI. For example:

```py
config = {
    "ferc_to_ferc": {
        "link_ids_cross_year": {
            "compute_distance_matrix": {"distance_threshold": .5, "metric": "euclidean"},
            "match_orphaned_records": {"distance_threshold": 0.5},
            ...
        {
    }
}
```

becomes:
```py
config = {
    "ferc_to_ferc.link_ids_cross_year.compute_distance_matrix.distance_threshold": .5,
    "ferc_to_ferc.link_ids_cross_year.compute_distance_matrix.metric": "euclidean",
    "ferc_to_ferc.match_orphaned_records.compute_distance_matrix.distance_threshold": .5,
    ...
}

File: src/pudl/analysis/record_linkage/model_helpers.py [hunk 2]
```diff
@@ -1,18 +1,134 @@
...
+                    config = flatten_model_config(
+                        get_model_config(experiment_config.experiment_name)
+                    )
+                    mlflow.log_params(config)
+
+                run_id = run.info.run_id
+
+        return cls(tracker_config=experiment_config, run_id=run_id)
+
+    def execute_logging(self, logging_func: Callable):
```
zschira, line 89: Passing in a `Callable` here means that we can only execute it and do any logging when `tracking_enabled=True`, and the caller doesn't need to know whether tracking is enabled or not. This also means that tracking can be enabled/disabled from the dagster ui

src/pudl/analysis/record_linkage/model_helpers.py
```diff
@@ -1,18 +1,134 @@
...
+            experiment_id = mlflow.create_experiment(
+                experiment_name, artifact_location=artifact_location
+            )
+
+        return experiment_id
+
 
-    return model_config
+@op
+def create_experiment_tracker(config: ExperimentTrackerConfig) -> ExperimentTracker:
```
zschira, line 132: This creates an experiment tracker which can be passed around to various `op`s within a `graph`. It will maintain state and restart an `mlflow` everytime it logs something

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
```diff
@@ -142,6 +150,15 @@ def merge_steam_fuel_dfs(
...
+@graph
+def ferc_to_ferc(
+    input_df: pd.DataFrame, experiment_tracker: model_helpers.ExperimentTracker
+) -> pd.DataFrame:
+    """Graph implements core portion of the ferc-ferc plant matching model."""
+    feature_matrix = ferc_dataframe_embedder(input_df, experiment_tracker)
+    return link_ids_cross_year(input_df, feature_matrix)
+
+
 @graph_asset(config=model_helpers.get_model_config("ferc_to_ferc"))
```
katie-lamb, line 162: Maybe this is unrelated to the experiment tracking, but what does this Dagster config do? It gets a dictionary with the config for the ferc to ferc experiment and that's used by mlflow?

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
```diff
@@ -587,6 +604,18 @@ def run_model(
         f"    Precision: {precision:.02}\n"
         f"    Recall:    {recall:.02}\n"
     )
+    # Log model metrics
+    experiment_tracker.execute_logging(
```
katie-lamb, line 608: I think ultimately these won't be the metrics that we track but this model will likely get replaced by the splink model so let's make those changes in the splink PR

src/pudl/analysis/record_linkage/model_helpers.py
```diff
@@ -1,18 +1,134 @@
...
+                flattened_dict[flattened_param[1:]] = val
+        return flattened_dict
+
+    return _flatten_level(model_config, "")
+
+
+class ExperimentTrackerConfig(Config):
+    """Dagster config to setup experiment tracking with mlflow."""
+
+    tracking_uri: str = f"sqlite:///{PudlPaths().output_dir}/experiments.sqlite"
```
katie-lamb, line 49: So ultimately these experiments get logged to a SQLite DB in the output directory and then can be loaded into the mlflow UI? I think it would be helpful to have more instruction or a tutorial on how to actually look at the tracked experiments

src/pudl/analysis/record_linkage/model_helpers.py
```diff
@@ -1,18 +1,134 @@
 """This module provides shared tooling that can be used by all record linkage models."""
```
katie-lamb, line 1: Two things: 

1. Does it make sense for the new experiment tracking modules to live in record linkage model helpers? Right now we only have record linkage models, but in the near future I could see incorporating experiment tracking into other kinds of modeling - like table extraction from PDFs. How well does it generalize currently and would it make any sense to just make it a separate module in `pudl.analysis`? 

2. I think we should include some more detailed instructions on how to use these experiment tracking classes - like what is the high level workflow when adding it into a new model. I think the same could be done with instructions on how to use the dataframe embedding/cleaning module. Where do you think a docstring like this would be best placed? Or is it best formatted as a tutorial notebook?

src/pudl/analysis/record_linkage/model_helpers.py
katie-lamb: Oh interesting, so I could make a change, enable experiment tracking in the dagster UI, run that experiment, then make another change and disable experiment tracking all within the dagster UI?

katie-lamb (changes requested):
Most of my comments are just about making it more clear how to actually use this and view experiments. I think a high level docstring that walks through which experiment tracking classes to instantiate for a new model, then how to use it within Dagster, and how to then take the experiments SQLite DB and open up the mlflow UI (or something else?) to compare experiments and tracked metrics. I'm not exactly sure where docstrings like this would live? Maybe the module level docstring in a separate experiment tracking module?

Also maybe we want to think about where we want experiment tracking to live so that it generalizes to more than just record linkage models.

src/pudl/analysis/record_linkage/model_helpers.py
zschira: Yeah! We might end up just having it on all the time, but seemed like a nice to have

src/pudl/analysis/record_linkage/classify_plants_ferc1.py
zschira: Yeah I have a yaml file in `package_data/` that has configuration for the `ferc_to_ferc` model. Passing it to the `graph_asset` decorator will override the default values defined in `classify_plants_ferc1.py`. If you configure an `ExperimentTracker` with `log_yaml=True`, it will call `get_model_config` with the experiment name to look for configuration and log those parameters with mlflow

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
zschira: It doesn't has to be in this PR, but it would be probably be good to update the metrics while this model is in use, so we can compare them to splink once it's in place.

src/pudl/analysis/record_linkage/model_helpers.py
zschira: Yeah I realized the docstrings were quite light across the board, so I did a little overhaul. I added some info on using the UI in the `ExperimentTracker` docstring, but I could see moving that. I moved everything to `pudl/analysis/ml_tools.py` to be more generic. I could see creating a specific ml focused package within `pudl/analysis/` with submodules, at which point we could move this into something like `pudl/analysis/ml_tools/experiment_tracking.py` and put it in the top level docstring. I could also do that now, but it seemed maybe excessive to create a new package when there's not too much code actually there right now.

src/pudl/analysis/record_linkage/model_helpers.py
zschira: See my comment above on organization stuff.

I've tried to put more details in the docstrings, but I do think a tutorial notebook would be good. I think it would be cool to create a couple tutorial notebooks that show how to set experiment tracking up in a notebook, then how to migrate a model into PUDL. We could use these as sort of scaffolding when we work on new models.

src/pudl/analysis/ml_tools.py
```diff
@@ -0,0 +1,181 @@
+"""This module provides shared tooling that can be used by all record linkage models."""
```
katie-lamb, line 1: Does this tooling still only generalize to record linkage models, or can it just read "all PUDL models" or something?

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
katie-lamb: Yep, I'll do this in the splink PR so we can more easily compare results.

src/pudl/analysis/record_linkage/model_helpers.py
katie-lamb: Ya I think the `pudl/analysis/ml_tools.py` structure is fine for now.

src/pudl/analysis/ml_tools.py
```diff
@@ -0,0 +1,181 @@
...
+    log_yaml: bool
+    run_context: str
+    #: Location to store artifacts. Artifact storage not currently used.
+    artifact_location: str = str(PudlPaths().output_dir)
+
+
+class ExperimentTracker(BaseModel):
+    """Class to manage tracking a machine learning model using MLflow.
+
+    The following command will launch the mlflow UI to view model results:
```
katie-lamb, line 85: Would it make sense to move most of this class docstring to a module level docstring? Then it gives kind of a general workflow for how to use experiment tracking?

src/pudl/analysis/ml_tools.py
```diff
@@ -0,0 +1,181 @@
...
+    `mlflow ui --backend-store-uri {tracking_uri}`. From here, you can compare metrics
```
katie-lamb, line 86: It's kind of hard to figure out what the tracking_uri is, and then you have to actually go to this docstring to get the command for the MLflow UI. Could you spit out a logging message with the tracking URI and the command to open the UI? Maybe in `create_experiment_tracker`?

src/pudl/analysis/ml_tools.py
```diff
@@ -0,0 +1,181 @@
...
+    from multiple runs, and track performance.
+
+    This class is designed to be created using the `op` :func:`create_experiment_tracker`.
+    This allows the `ExperimentTracker` to be passed around within a Dagster `graph`,
+    and be used for mlflow logging in any of the `op`s that make up the `graph`. This
+    is useful because Dagster executes `op`s in separate processes, while mlflow does
+    not maintain state between processes. This design also allows configuration of
+    the ExperimentTracker to be set from the Dagster UI.
+
```
katie-lamb, line 95: How can you change the configuration from the Dagster UI? Maybe I've just never figured out how to change any of the Dagster config stuff from the UI. I think eventually it would be nice if there was an Experiment Tracking section in the PUDL Development docs but maybe that's for a later PR.

katie-lamb:
I left a few clarifying questions about the docstring that explains how to use MLflow. Also, would it be helpful to track more than just the number of year duplicates for the FERC-FERC linkage? Like for the three ops in `link_cross_year` we could track the number of unique record IDs after each step of the clustering process, if that seems like a helpful metric.

katie-lamb:
Also I think it would be great to set up a tracking server as part of the SEC to EIA record linkage work.

katie-lamb:
I created two spinoff issues, #3406 and #3407

src/pudl/analysis/ml_tools.py
zschira: Yeah I'm still figuring out dagster config, initially you had to create the structure yourself and it was pretty confusing to actually get right, but I've made some changes, so now if you click `Open Launchpad` in the UI, the config will pop up for each model, and you can easily edit values right there.

src/pudl/analysis/ml_tools.py
zschira: Added a log

src/pudl/analysis/ml_tools.py
zschira: I reworked the structure a bit, so now most of this info is in a module level docstring

katie-lamb (approved):
Little change to see if the config shows up in the Dagster launchpad but approving since everything else looks good.

PR #3280, opened 2024-01-23
Map PHMSA Natural Gas Transmission Part Q columns

cmgosnell:
# Overview

Closes #3275.

What problem does this address?

What did you change?

# Testing

How did you make sure this worked? How can a reviewer verify this?

```[tasklist]
# To-do list
- [x] Make sure full ETL runs & `make pytest-integration-full` passes locally
- [x] Update the [release notes](../docs/release_notes.rst): reference the PR and related issues.
- [ ] For major data coverage & analysis changes, [run data validation tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#data-validation)
- [ ] If updating analyses or data processing functions: make sure to update or write data validation tests
- [ ] Review the PR yourself and call out any questions or issues you have
```

PR #3291, opened 2024-01-24
PHMSA: fix to Part Q - fix the column names for the other materials (NOT MILES!)

cmgosnell:
# Overview

This is an addendum to  #3280 which i learned i needed to do via @jdangerx  over in #3279

i previously labeled the fields where phmsa respondents are supposed to list other materials as `_miles` and generally didn't realize these were linked to the cram in the column labeled as "Other" w/ the lil footnote.

What did you change?
removed miles and standardized the column names so they are more mirrored w/ their brethren from the table above.

# Testing

```[tasklist]
# To-do list
```


jdangerx (changes requested):
Nice catch! We should probably follow @e-belfer 's lead and add a `_description` suffix to these since they're "specify" fields, though.

And these are 192.619 (vs. 192.624) because 192.619 includes "other" and 192.624 only includes itself?

cmgosnell:
he yes i just read ella's comment over there and was coming back to add `_description`. and yes these are all 192.619 because this table is broken out by the two regulation sections and only 192.619 has the "other" category.

jdangerx (approved):

PR #3292, opened 2024-01-24
Push to prod Zenodo sometimes, fix pypi release flow, update docs

### LINKED ISSUE #3175: Create production data release on push of `v20*` tags.

jdangerx:
# Overview

Closes #3175:

- Tries to upload draft to Zenodo prod
- Doesn't run PyPI release unless test PyPI release goes through
- added docs

Side-note:

It *really* doesn't seem like it would be that hard to grab the outputs from GCS & just upload to Zenodo that way:

- use `git tag --points-at {{ github.ref_name }}` to figure out which nightly build to access on `builds.catalyst.coop`
- auth into GCP - we could add another WIF identity, or just use the one we use in `build-deploy-pudl.yml`
- sync the directory down to some location on the runner
- check out repo + install deps
- run zenodo_data_release.py

Some clean-up we'd have to do:
- stop running build-deploy-pudl on tag
- run `release.yml` on pushes to `nightly` as well as `vYYYY.MM.DD`, with the appropriate conditionals (e.g., never running prod PyPI release for nightly, always pushing to sandbox for nightly)
- rip out the zenodo publication logic in gcp_pudl_etl.sh 🎉 

So maybe it's worth doing that here? Would make testing this go a bit more smoothly I think.

# Testing

Here's the testing plan:

1. push `v2024.01.25.a1` tag
2. `release.yml` goes:
   - [x] doesn't publish to PyPI production - that one short-circuits
   - [x] does publish to PyPI test
   - [x] We should be able to click through the workflow and see the deps work correctly - e.g., pypi depending on test-pypi.
3. `build-deploy-pudl.yml` gets kicked off:
   - [x] `run_pudl_etl` gets short-circuited
   - [x] we don't publish `stable` to distribution buckets
   - [x] we *do* publish `v20204.01.25.test` in private/public buckets
   - [x] there's a draft record in Zenodo with some bogus files - log, test file

After verifying the behavior of the Test PyPI release I short-circuited so that we wouldn't repeatedly attempt to release v2024.1.25.

```[tasklist]
# To-do list
- [x] update with prod Zenodo token - not a publish token! 😅
```

If we put the Zenodo publication logic in the release workflow, we can comment out the prod PyPI stuff and just push some `vYYYY.MM.DD_rcX` tags, avoiding the whole "short-circuit `gcp_pudl_etl.sh`" step - but, let's first try to get this working for now.

### DIFF
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -154,6 +154,7 @@ jobs:
             --container-env PUDL_BOT_PAT=${{ secrets.PUDL_BOT_PAT }} \
             --container-env ZENODO_SANDBOX_TOKEN_PUBLISH=${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }} \
             --container-env ZENODO_TARGET_ENV=${{ (startsWith(github.ref_name, 'v20') && 'production') || 'sandbox' }} \
+            --container-env ZENODO_TOKEN_UPLOAD=${{ secrets.ZENODO_TOKEN_UPLOAD }} \
             --container-env PUDL_SETTINGS_YML="/home/mambauser/pudl/src/pudl/package_data/settings/etl_full.yml" \
             --container-env PUDL_GCS_OUTPUT=${{ env.PUDL_GCS_OUTPUT }}
 
--- a/.github/workflows/build-deploy-pudl.yml
+++ b/.github/workflows/build-deploy-pudl.yml
@@ -153,6 +153,7 @@ jobs:
             --container-env FLY_ACCESS_TOKEN=${{ secrets.FLY_ACCESS_TOKEN }} \
             --container-env PUDL_BOT_PAT=${{ secrets.PUDL_BOT_PAT }} \
             --container-env ZENODO_SANDBOX_TOKEN_PUBLISH=${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }} \
+            --container-env ZENODO_TARGET_ENV=${{ (startsWith(github.ref_name, 'v20') && 'production') || 'sandbox' }} \
             --container-env PUDL_SETTINGS_YML="/home/mambauser/pudl/src/pudl/package_data/settings/etl_full.yml" \
             --container-env PUDL_GCS_OUTPUT=${{ env.PUDL_GCS_OUTPUT }}
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,6 @@ jobs:
     name: Upload release to PyPI - only if test release is successful!
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v20')
     needs:
-      - build-distribution
       - publish-test-pypi
     runs-on: ubuntu-latest
     environment:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,10 +49,11 @@ jobs:
           repository-url: https://test.pypi.org/legacy/
 
   publish-pypi:
-    name: Upload release to PyPI
+    name: Upload release to PyPI - only if test release is successful!
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v20')
     needs:
       - build-distribution
+      - publish-test-pypi
     runs-on: ubuntu-latest
     environment:
       name: pypi
--- a/devtools/zenodo/zenodo_data_release.py
+++ b/devtools/zenodo/zenodo_data_release.py
@@ -71,18 +71,18 @@ class ZenodoClient:
         """
         if env == SANDBOX:
             self.base_url = "https://sandbox.zenodo.org/api"
-            self.auth_headers = {
-                "Authorization": f"Bearer {os.environ['ZENODO_SANDBOX_TOKEN_PUBLISH']}"
-            }
+            token = os.environ["ZENODO_SANDBOX_TOKEN_PUBLISH"]
         elif env == PRODUCTION:
             self.base_url = "https://zenodo.org/api"
-            self.auth_headers = {
-                "Authorization": f"Bearer {os.environ['ZENODO_TOKEN_PUBLISH']}"
-            }
+            token = os.environ["ZENODO_TOKEN_UPLOAD"]
         else:
             raise ValueError(
                 f"Got unexpected {env=}, expected {SANDBOX} or {PRODUCTION}"
             )
+
+        self.auth_headers = {"Authorization": f"Bearer {token}"}
+
+        logger.info(f"Using Zenodo token: {token[:4]}...{token[-4:]}")
 
     def get_deposition(self, deposition_id: int) -> _LegacyDeposition:
         """LEGACY API: Get JSON describing a deposition.
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -101,8 +101,13 @@ function copy_outputs_to_distribution_bucket() {
 }
 
 function zenodo_data_release() {
-    echo "Creating a new PUDL data release on Zenodo." && \
-    ~/pudl/devtools/zenodo/zenodo_data_release.py --publish --env "$1" --source-dir "$PUDL_OUTPUT"
+    echo "Creating a new PUDL data release on Zenodo."
+
+    if [[ "$1" == "production" ]]; then
+        ~/pudl/devtools/zenodo/zenodo_data_release.py --no-publish --env "$1" --source-dir "$PUDL_OUTPUT"
+    else
+        ~/pudl/devtools/zenodo/zenodo_data_release.py --publish --env "$1" --source-dir "$PUDL_OUTPUT"
+    fi
 }
 
 function notify_slack() {
@@ -195,7 +200,7 @@ if [[ $ETL_SUCCESS == 0 ]]; then
         # TODO: this currently just makes a sandbox release, for testing. Should be
         # switched to production and only run on push of a version tag eventually.
         # Push a data release to Zenodo for long term accessiblity
-        zenodo_data_release sandbox 2>&1 | tee -a "$LOGFILE"
+        zenodo_data_release "$ZENODO_TARGET_ENV" 2>&1 | tee -a "$LOGFILE"
         ZENODO_SUCCESS=${PIPESTATUS[0]}
     fi
 fi
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -65,8 +65,8 @@ function save_outputs_to_gcs() {
 }
 
 function upload_to_dist_path() {
-    GCS_PATH="gs://pudl.catalyst.coop/$1"
-    AWS_PATH="s3://pudl.catalyst.coop/$1"
+    GCS_PATH="gs://pudl.catalyst.coop/$1/"
+    AWS_PATH="s3://pudl.catalyst.coop/$1/"
 
     # If the old outputs don't exist, these will exit with status 1, so we
     # don't && them with the rest of the commands.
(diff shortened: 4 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -49,10 +49,11 @@ jobs:
           repository-url: https://test.pypi.org/legacy/
 
   publish-pypi:
-    name: Upload release to PyPI
+    name: Upload release to PyPI - only if test release is successful!
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v20')
     needs:
       - build-distribution
```
zaneselvans, line 55: I guess we don't technically need both of these, since `test-pypi` already depends on `build-distribution`

docs/dev/run_a_release.rst
```diff
@@ -0,0 +1,48 @@
...
+
+5. If *that* build passes, verify the following:
+
+   * GCS/AWS distribution buckets have the appropriate data
+   * ``stable`` and ``vYYYY.MM.DD`` point at the same Git ref
+   * ReadTheDocs for ``stable`` and ``vYYYY.MM.DD`` versions have the latest
+     changes in the release notes
+   * PyPI has the most current version of our code
+
+6. Update the Kaggle dataset.
```
zaneselvans, line 40: I think we can and should omit Kaggle from all this, at least for the moment and probably for the foreseeable future, unless they get their API fleshed out.  Right now Kaggle pulls new outputs once a week from whatever URLs are associated with the dataset.  Right now that's `/nightly` but if we want it to be `/stable` we just need to update the URLs, and we can let Kaggle pull down the new version on the schedule we set (can be as frequently as daily if we want, but that seemed a bit excessive)

docs/dev/run_a_release.rst
```diff
@@ -0,0 +1,48 @@
...
+4. If the nightly build passes, push a git tag ``vYYYY.MM.DD`` that points at
+   the commit which just passed the nightly build. This will kick off another build.
+
+5. If *that* build passes, verify the following:
+
+   * GCS/AWS distribution buckets have the appropriate data
+   * ``stable`` and ``vYYYY.MM.DD`` point at the same Git ref
+   * ReadTheDocs for ``stable`` and ``vYYYY.MM.DD`` versions have the latest
+     changes in the release notes
+   * PyPI has the most current version of our code
```
zaneselvans, line 38: One additional step here is that within ~24hrs of a new version showing up on PyPI we'll get a PR from the `conda-forge` bot updating the `conda` package, that will need to be attended to. The main human intervention required there will be updating the list of direct dependencies in the conda packaging if they have changed.

docs/data_access.rst
```diff
@@ -157,18 +157,33 @@ can also be installed using the conda package manager).
 If you're not familiar with using Unix command line tools in Windows you can also use a
 3rd party tool like `7zip <https://www.7-zip.org/download.html>`__.
 
-.. _access-zenodo:
+.. _access-stable:
 
 ---------------------------------------------------------------------------------------
-Zenodo
+Stable Builds
```
zaneselvans, line 163: Previously this section was trying to do double duty documenting both that the raw inputs are being archived on Zenodo, as well as the stable data releases.  The raw inputs part feels a little buried now. Should it go somewhere else entirely, since this page is more focused on accessing outputs?

docs/dev/index.rst
```diff
@@ -14,6 +14,7 @@ Development
   datastore
   clone_ferc1
   existing_data_updates
+  run_a_release
```
zaneselvans, line 17: I added the release instructions to the toctree here, otherwise they don't show up anywhere and Sphinx complains.

zaneselvans (changes requested):

docs/dev/index.rst
jdangerx: thank you!

jdangerx:
Oops, `v2024.01.25.test` didn't conform to PEP 440 and thus setuptools_scm got angry. Going to:

1. un-short-circuit the Test PyPI deployment
2. try pushing `v2024.01.25.a1` tag instead.

jdangerx:
Found out that our version regex in `pyproject.toml` was rejecting the `a1` suffix - and then, if I modified the regex to include that suffix, found that `setuptools_scm` was doing something funky with the tag `v2024.01.25a1`:

```
> python -m setuptools_scm
2024.1.25a2.dev0+g287a85d6c.d20240125
```

So I just used `v2024.01.25` instead since we're likely not doing an actual release today. I'll clean up the distribution outputs.

jdangerx:
Ok, @zaneselvans - I went through all that short-circuiting above, found some bugs, and this now creates a handy dandy draft in production Zenodo for human approval! And also pushes stuff out to the distribution buckets. I already went and deleted all the artifacts from the tests, to avoid people getting confused about weird empty `v2024.01.25` data.

docker/gcp_pudl_etl.sh
```diff
@@ -76,46 +93,46 @@ function copy_outputs_to_distribution_bucket() {
...
-            echo "Removing old stable outputs from GCS distributon bucket." && \
-            gsutil -m -u "$GCP_BILLING_PROJECT" rm -r "gs://pudl.catalyst.coop/stable" && \
-            echo "Copying tagged version outputs to stable GCS distribution bucket" && \
-            gsutil -m -u "$GCP_BILLING_PROJECT" cp -r "$PUDL_OUTPUT/*" "gs://pudl.catalyst.coop/stable" && \
-            echo "Removing old stable outputs from AWS S3 distributon bucket." && \
-            aws s3 rm "s3://pudl.catalyst.coop/stable" --recursive && \
-            echo "Copying tagged version outputs to stable AWS S3 distribution bucket" && \
-            aws s3 cp "$PUDL_OUTPUT/" "s3://pudl.catalyst.coop/stable" --recursive
-
+            upload_to_dist_path "stable"
```
zaneselvans, line 100: Oh duh, this is much better consolidated!

docs/data_access.rst
```diff
@@ -157,18 +160,47 @@ can also be installed using the conda package manager).
...
+s3://pudl.catalyst.coop/stable/`` to see what's available.
+
+.. _access-raw:
+
+---------------------------------------------------------------------------------------
+Raw Data
+---------------------------------------------------------------------------------------
+
+Sometimes you want to see the raw data that is published by the government, but
+it's hard to find or difficult to download.
```
zaneselvans, line 194:
```suggestion
Sometimes you want to see the raw data that is published by the government, but
it's hard to find or difficult to download, or you want to see what an older version of
the published data looked like prior to being revised or deleted.
```

docs/dev/run_a_release.rst
```diff
@@ -0,0 +1,52 @@
...
+
+7. Update the Zenodo metadata to include the description from the release
+   notes, and update the various links.
+
+8. Publish the Zenodo deposition! Wahoo! You're now done!
+
+9. Except... 24 hours after the PyPI version is updated, we'll get a PR from
+   the ``conda-forge`` bot updating our ``conda`` package. If the direct
+   dependencies have changed in this release, we need to update them in that
+   PR.
```
zaneselvans, line 52:
```suggestion
9. Except... within 24 hours after the PyPI version is updated, we'll get [a PR in the PUDL
   conda-forge feedstock repo](https://github.com/conda-forge/catalystcoop.pudl-feedstock/pulls)
   from the :user:`regro-cf-autotick-bot` updating our ``conda`` packaging. If PUDL's direct
   dependencies have changed in this release, we need to update them in the conda packaging
   recipe in that PR.
```

docker/gcp_pudl_etl.sh
```diff
@@ -64,6 +64,23 @@ function save_outputs_to_gcs() {
     rm "$PUDL_OUTPUT/success"
 }
 
+function upload_to_dist_path() {
+    GCS_PATH="gs://pudl.catalyst.coop/$1"
+    AWS_PATH="s3://pudl.catalyst.coop/$1"
```
zaneselvans, line 69:
```suggestion
    GCS_PATH="gs://pudl.catalyst.coop/$1/"
    AWS_PATH="s3://pudl.catalyst.coop/$1/"
```

docker/gcp_pudl_etl.sh
```diff
@@ -64,6 +64,23 @@ function save_outputs_to_gcs() {
...
+    # If the old outputs don't exist, these will exit with status 1, so we
+    # don't && them with the rest of the commands.
+    echo "Removing old outputs from $GCS_PATH."
+    gsutil -m -u "$GCP_BILLING_PROJECT" rm -r "$GCS_PATH"
+    echo "Removing old outputs from $AWS_PATH."
+    aws s3 rm "$AWS_PATH" --recursive
+
+    echo "Copying outputs to $GCS_PATH:" && \
+    gsutil -m -u "$GCP_BILLING_PROJECT" cp -r "$PUDL_OUTPUT/*" "$GCS_PATH" && \
+    echo "Copying outputs to AWS distribution bucket" && \
```
zaneselvans, line 80:
```suggestion
    echo "Copying outputs to $AWS_PATH" && \
```

.github/workflows/build-deploy-pudl.yml
```diff
@@ -153,6 +153,8 @@ jobs:
             --container-env FLY_ACCESS_TOKEN=${{ secrets.FLY_ACCESS_TOKEN }} \
             --container-env PUDL_BOT_PAT=${{ secrets.PUDL_BOT_PAT }} \
             --container-env ZENODO_SANDBOX_TOKEN_PUBLISH=${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }} \
+            --container-env ZENODO_TARGET_ENV=${{ (startsWith(github.ref_name, 'v20') && 'production') || 'sandbox' }} \
+            --container-env ZENODO_TOKEN_UPLOAD=${{ secrets.ZENODO_TOKEN_UPLOAD }} \
```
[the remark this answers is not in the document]
zaneselvans: (looks like below we're differentiating between publish and upload tokens so ✅

zaneselvans (changes requested):
Left some minor suggested changes.

.github/workflows/build-deploy-pudl.yml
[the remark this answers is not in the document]
jdangerx: Yep! Wanted to have as many safeguards against oopsie publishing as possible :)

devtools/zenodo/zenodo_data_release.py
```diff
@@ -71,19 +71,19 @@ def __init__(self, env: str):
...
-            }
+            token = os.environ["ZENODO_TOKEN_UPLOAD"]
         else:
             raise ValueError(
                 f"Got unexpected {env=}, expected {SANDBOX} or {PRODUCTION}"
             )
 
+        self.auth_headers = {"Authorization": f"Bearer {token}"}
+
+        logger.info(f"Using Zenodo token: {token[:4]}...{token[-4:]}")
```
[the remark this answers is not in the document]
jdangerx: Yeah, I figured it is not really a security risk to log 8 characters of it and helps diagnose issues when e.g. Zenodo decides to expire all the tokens without telling anyone :)

.github/workflows/build-deploy-pudl.yml
```diff
@@ -153,6 +153,8 @@ jobs:
             --container-env FLY_ACCESS_TOKEN=${{ secrets.FLY_ACCESS_TOKEN }} \
             --container-env PUDL_BOT_PAT=${{ secrets.PUDL_BOT_PAT }} \
             --container-env ZENODO_SANDBOX_TOKEN_PUBLISH=${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }} \
+            --container-env ZENODO_TARGET_ENV=${{ (startsWith(github.ref_name, 'v20') && 'production') || 'sandbox' }} \
```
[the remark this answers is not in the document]
jdangerx: Yeah, this is the GHA version of the ternary: `startsWith(github.ref_name, 'v20') ? 'production' : 'sandbox'` for

zaneselvans (approved):

PR #3295, opened 2024-01-25
Clean up and standardize PHMSA raw assets

### LINKED ISSUE #3277: Clean up and standardize column names
We highly distributed the mapping of PHMSA raw data. This means we probably have some inconsistencies across our column naming conventions we should fix. These include:

- [x] Fix ordering of onshore/offshore and transmission/gathering
- [x] Check that categorizations are consistent across tables
- [x] Check missing and extra column errors for all tables to make sure we haven't mismapped anything
- [x] Standardize file names - `transmission_gathering`, `miles`
- [x] Standardize placement of `pipe`
- [x] standardize treatment of time between transmission and distribution data

e-belfer:
# Overview

Closes #3277. See issue for task list.

Non-standard column and file naming practices caused by parallel PHMSA mapping, plus some mis-mapped columns.

What did you change?
File and column names, and a few missing/wrong column mappings in the CSVs.

# Testing

Generate all raw PHMSA assets in dagster and check the logs to make sure only a handful of `_other_size_miles` custom fill-in fields are missing 2010-present (they show up in the form and not the data, but we should keep them mapped anyways just in case this changes).

```[tasklist]
# To-do list
```


src/pudl/extract/phmsagas.py
cmgosnell, line 1: oh i'm so glad you standardized where the miles are in here that was driving me bananas! I am tempted to suggest tepidly that we remove the `transmission_gathering` from these asset names but it is informative and despite the very longness of these names i think they are good.

src/pudl/package_data/phmsagas/column_maps/yearly_transmission_gathering_pipe_miles_by_class_location.csv
```diff
@@ -60,7 +60,3 @@ total_pipe_class_2_miles,,,,,,,,,,,,b4total_2,b4total_2,b4total_2,b4total_2,b4to
...
-total_pipe_192_710_miles,,,,,,,,,,,,,,,,,,,,,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal,partltr192milestotal
-total_pipe_class_3_or_4_only_miles,,,,,,,,,,,,,,,,,,,,,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total,partltrclass34total
-total_pipe_class_1_or_2_only_miles,,,,,,,,,,,,,,,,,,,,,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total,partltrclass12total
```
cmgosnell, line 66: why are these guys gone? did they get double mapped or something?

src/pudl/package_data/phmsagas/column_maps/yearly_transmission_gathering_summary_by_commodity.csv
```diff
@@ -24,36 +24,36 @@ system_state,stop,stop,stop,stop,stop,stop,stop,stop,stop,stop,stop,stop,stop,st
...
+hca_method_total_relative_risk_miles,,,,,,,,,,,,,,,,,,,,,,,partb1rrt,partb1rrt,partb1rrt,partb1rrt,partb1rrt,partb1rrt,partb1rrt,partb1rrt,partb1rrt,partb1rrt,partb1rrt
+hca_method_total_quantitative_miles,,,,,,,,,,,,,,,,,,,,,,,partb1quant,partb1quant,partb1quant,partb1quant,partb1quant,partb1quant,partb1quant,partb1quant,partb1quant,partb1quant,partb1quant
+hca_method_total_probabilistic_miles,,,,,,,,,,,,,,,,,,,,,,,partb1probt,partb1probt,partb1probt,partb1probt,partb1probt,partb1probt,partb1probt,partb1probt,partb1probt,partb1probt,partb1probt
+hca_method_total_scenario_based_miles,,,,,,,,,,,,,,,,,,,,,,,partb1sbt,partb1sbt,partb1sbt,partb1sbt,partb1sbt,partb1sbt,partb1sbt,partb1sbt,partb1sbt,partb1sbt,partb1sbt
+hca_method_total_other_miles,,,,,,,,,,,,,,,,,,,,,,,partb1otht,partb1otht,partb1otht,partb1otht,partb1otht,partb1otht,partb1otht,partb1otht,partb1otht,partb1otht,partb1otht
+hca_method_total_miles,,,,,,,,,,,,,,,,,,,,,,,partb1mt,partb1mt,partb1mt,partb1mt,partb1mt,partb1mt,partb1mt,partb1mt,partb1mt,partb1mt,partb1mt
```
cmgosnell, line 56: this is not blocking but i think it will be helpful in our future transformations of these tables to include totals even when it seems silly/redundant. which i think in this structure would be `hca_method_total_total_miles`. but these guys might be easier to see and change when we actually get to transformations so i don't feel strongly at all about doing this right now because it would be relatively challenging and maybe preemptive to do it before we really know how its going to be reshaped.

src/pudl/package_data/phmsagas/column_maps/yearly_transmission_gathering_pipe_miles_by_material.csv
cmgosnell, line 1: this is just a note: the old name version of this guy got deleted and the new name version got added separately in gits eyeballs so i can't tell what if anything you changed in here. but it looks sensible on first glance!

cmgosnell (changes requested):
my only real blocking concern here is the deletion of some columns!

src/pudl/package_data/phmsagas/column_maps/yearly_transmission_gathering_pipe_miles_by_class_location.csv
e-belfer: Yes, they got double mapped. See `transmission_pipe_hca_miles`, e.g. There aren't gathering miles in this section so it's only transmission, we decided for now to keep the totals closer to the source (and we could fill in the `total_` when we're cleaning, if we want).

src/pudl/package_data/phmsagas/column_maps/yearly_transmission_gathering_summary_by_commodity.csv
e-belfer: Agreed, but I think this will be easier to do during the transform step.

src/pudl/package_data/phmsagas/column_maps/yearly_transmission_gathering_pipe_miles_by_material.csv
e-belfer: Ah, should've split the edits and renames into two commits, sorry!

src/pudl/package_data/phmsagas/column_maps/yearly_transmission_gathering_pipe_miles_by_class_location.csv
cmgosnell: ah okay great!!

cmgosnell (approved):
🚀

PR #3296, opened 2024-01-25
Output Parquet files as well as SQLite in PUDL ETL

zschira:
## Background
This PR is a branch off of #3232 . Most of the changes come from that branch/pr with a few additions which attempt to answer unresolved questions from that ongoing PR. Specific changes here include: refactoring where the `check_foreign_keys` logic lives, reverting the use of `unittest` for consistency (I don't feel too strongly either way, but consistency is nice), and removing the check for extra columns from `enforce_schema` which was causing CI to fail.


src/pudl/etl/check_foreign_keys.py
```diff
@@ -50,15 +61,143 @@ def pudl_check_fks(logfile: pathlib.Path, loglevel: str):
...
+    )
+    table_fks = table_fks[["id", "table", "fk"]].drop_duplicates()
+
+    # Rename the fields so we can easily merge with the foreign key errors.
+    table_fks = table_fks.rename(columns={"id": "fkid", "table": "parent"})
+    table_fks["table"] = table
+    return table_fks
+
+
+def check_foreign_keys(engine: sa.Engine):
```
zschira, line 150: Pulled the foreign key check logic out of the io-manager and put it here directly, which seemed as good a place as anywhere to put it.

test/conftest.py
```diff
@@ -316,13 +315,13 @@ def pudl_sql_io_manager(
...
-    return pudl_sqlite_io_manager(context)
+    return pudl_io_manager(context)
 
 
 @pytest.fixture(scope="session")
-def pudl_engine(pudl_sql_io_manager: PudlSQLiteIOManager) -> sa.Engine:
+def pudl_engine(pudl_mixed_format_io_manager: PudlMixedFormatIOManager) -> sa.Engine:
     """Get PUDL SQL engine from io manager."""
-    return pudl_sql_io_manager.engine
+    return pudl_mixed_format_io_manager._sqlite_io_manager.engine
```
zschira, line 324: I didn't add a getter because this is the only place we directly access the `_sqlite_io_manager` like this, and this fixture doesn't make sense if the SQLite manager doesn't exist, so it felt fine in this testing context to make that assumption.

test/unit/io_managers_test.py
```diff
@@ -277,6 +280,33 @@ def test_error_when_handling_view_without_metadata(fake_pudl_sqlite_io_manager_f
...
+    """Tests that two runs of the same asset overwrite existing contents."""
+    artist_df = pd.DataFrame({"artistid": [1], "artistname": ["Co-op Mop"]})
+    output_context = build_output_context(asset_key=AssetKey("artist"))
+    input_context = build_input_context(asset_key=AssetKey("artist"))
+
+    # Write then read.
+    fake_pudl_sqlite_io_manager_fixture.handle_output(output_context, artist_df)
+    read_df = fake_pudl_sqlite_io_manager_fixture.load_input(input_context)
+    pd.testing.assert_frame_equal(artist_df, read_df, check_dtype=False)
+    # check_dtype=False, because int64 != Int64. /o\
```
zaneselvans, line 300: Should we be getting matching dtypes here, if we're using the `pudl_sqlite_io_manager` since it uses `enforce_schema(df)` before writing and after reading? Is that having no effect because there's not actually any `Resource` defined for the tables being written/read with specific data types? I don't really understand the how or why of mocking.

zaneselvans (changes requested):
Mostly questions and a bit of documentation clarification.

One naming thing -- about where we use the generic `pudl_io_manager` vs. `pudl_{specific_type}_io_manager`

src/pudl/io_managers.py
```diff
@@ -31,56 +34,75 @@
...
-                and (self.foreign_key == other.foreign_key)
-                and (self.rowids == other.rowids)
-            )
-        return False
+class PudlMixedFormatIOManager(IOManager):
+    """Format switching IOManager that supports sqlite and parquet.
+
+    We currently support both sqlite and parquet formats. This io manager
+    allows us to choose which formats should be used for reading and which
+    should be used for reading.
```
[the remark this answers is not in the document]
zschira: Reworded the docstring.

src/pudl/io_managers.py
```diff
@@ -31,56 +34,75 @@
...
+
+        Args:
+            write_to_parquet: if True, all data will be written to parquet
+                files in addition to sqlite.
+            read_from_parquet: if True, all data reads will be using
+                parquet files as source of truth. Otherwise, data will be
+                read from the sqlite database.
+        """
+        self.write_to_parquet = write_to_parquet
+        self.read_from_parquet = read_from_parquet
```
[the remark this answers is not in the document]
zschira: I added a check to raise an error in this case

src/pudl/io_managers.py
```diff
@@ -182,84 +196,6 @@ def _get_sqlalchemy_table(self, table_name: str) -> sa.Table:
             )
         return sa_table
 
-    def _get_fk_list(self, table: str) -> pd.DataFrame:
```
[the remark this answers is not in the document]
zschira: I'm not sure, but I guess it could've made more sense when we were only using the sqlite io-manager

src/pudl/io_managers.py
```diff
@@ -375,6 +310,38 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
+            where=parquet_path,
+            schema=schema,
+            compression="snappy",
+            version="2.6",
+        ) as writer:
+            writer.write_table(
+                pa.Table.from_pandas(df, schema=schema, preserve_index=False)
+            )
+
+    def load_input(self, context: InputContext) -> Any:
```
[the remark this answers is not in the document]
zschira: Yeah this should only return dataframes, updated.

src/pudl/workspace/setup.py
```diff
@@ -66,6 +66,10 @@ def sqlite_db_uri(self, name: str) -> str:
         # sqlite://{credentials}/{db_path}
         return f"sqlite:///{self.sqlite_db_path(name)}"
 
+    def parquet_path(self, db_name: str, table_name: str) -> Path:
```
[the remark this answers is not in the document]
zschira: Jan had a reason on his PR, but I think it's a bit confusing to have an unused param that might be used in the future, so I  removed it

src/pudl/etl/__init__.py
```diff
@@ -95,7 +95,7 @@
 
 default_resources = {
     "datastore": datastore,
-    "pudl_sqlite_io_manager": pudl_sqlite_io_manager,
+    "pudl_io_manager": pudl_io_manager,
```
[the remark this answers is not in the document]
zschira: Agreed, changed the name

src/pudl/etl/check_foreign_keys.py
```diff
@@ -50,15 +61,147 @@ def pudl_check_fks(logfile: pathlib.Path, loglevel: str):
...
+    """Raised when data in a database violates a foreign key constraint."""
+
+    def __init__(
+        self, child_table: str, parent_table: str, foreign_key: str, rowids: list[int]
+    ):
+        """Initialize a new ForeignKeyError object."""
+        self.child_table = child_table
+        self.parent_table = parent_table
+        self.foreign_key = foreign_key
+        self.rowids = rowids
```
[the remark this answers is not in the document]
zschira: Add docstring descriptions. @bendnorman can you check these and make sure all the descriptions are correct?

src/pudl/etl/check_foreign_keys.py
```diff
@@ -50,15 +61,147 @@ def pudl_check_fks(logfile: pathlib.Path, loglevel: str):
...
     pudl.logging_helpers.configure_root_logger(logfile=logfile, loglevel=loglevel)
 
-    context = build_init_resource_context()
-    io_manager = pudl_sqlite_io_manager(context)
+    # Using PudlPaths to get default value for CLI causes validation issues
+    if not db_path:
+        db_path = PudlPaths().output_dir / "pudl.sqlite"
 
-    database_path = io_manager.base_dir / f"{io_manager.db_name}.sqlite"
-    logger.info(f"Checking foreign key constraints in {database_path}")
```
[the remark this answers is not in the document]
zschira: This is logged directly in the `check_foreign_keys` function, so this was just removing a duplicate log

test/unit/io_managers_test.py
zschira: It is applying `enforce_schema` because earlier in this file the `Package` is created with the schema, so I think it's the original `DataFrame` where the types are off. We could set those explicitly so we can check dtypes, but if that would be better.

src/pudl/etl/check_foreign_keys.py
```diff
@@ -50,15 +61,154 @@ def pudl_check_fks(logfile: pathlib.Path, loglevel: str):
...
+    check for foreign key failures once all of the data has been loaded into
+    the database using the `foreign_key_check` and `foreign_key_list` PRAGMAs.
+
+    You can learn more about the PRAGMAs in the `SQLite docs
+    <https://www.sqlite.org/pragma.html#pragma_foreign_key_check>`__.
+
+    Raises:
+        ForeignKeyErrors: if data in the database violate foreign key constraints.
+    """
+    logger.info(f"Running foreign key check on {engine.name} database.")
```
zaneselvans, line 175: I think this is the logging message you're referring to above, but unlike the logging message which was removed, it doesn't provide the path to the database whose FK relationships it's checking, which could lead to confusion in testing or development if say, the `$PUDL_OUTPUT` envvar isn't set correctly, or isn't set to the value that the user thinks it's set to.

src/pudl/etl/check_foreign_keys.py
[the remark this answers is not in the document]
zaneselvans: Yes, but it only includes the `engine.name` and not the path, which is less informative. See note below.

src/pudl/etl/check_foreign_keys.py
```diff
@@ -50,15 +61,154 @@ def pudl_check_fks(logfile: pathlib.Path, loglevel: str):
...
+    def __init__(
+        self, child_table: str, parent_table: str, foreign_key: str, rowids: list[int]
+    ):
+        """Initialize a new ForeignKeyError object.
+
+        Args:
+            child_table: The table that a foreign key constraint is applied to.
+            parent_table: The table that a foreign key constraint refers to.
+            foreign_key: Comma seperated string of key(s) that make up foreign key.
+            rowids: Row(s) of child_table where constraint failed.
```
zaneselvans, line 86: Like the entire rows?  Or the dataframe index of the rows? Or the primary keys of the rows? Or the SQLite autoincrementing integer pseudokey associated with the rows?

src/pudl/etl/check_foreign_keys.py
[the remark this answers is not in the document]
zaneselvans: I'm still not sure what the `rowids` refer to here. I could imagine this wording meaning:

- All of the data in the bad rows.
- The dataframe indices of the bad rows.
- The primary keys of the bad rows.
- The SQLite database's internal integer id associated with the bad rows.

src/pudl/io_managers.py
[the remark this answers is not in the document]
zaneselvans: Is it updated tho?

src/pudl/io_managers.py
```diff
@@ -414,6 +388,8 @@ def __init__(
                 exception, if the database is locked by another connection.  If another
                 connection opens a transaction to modify the database, it will be locked
                 until that transaction is committed.
+            settings: controls how parquet files are used by the PUDL ETL, in particular,
+                whether they should be used for input/output of dataframes.
```
zaneselvans, line 392: This looks like it might be a leftover from the previous PR that should be removed?

src/pudl/io_managers.py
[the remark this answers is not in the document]
zaneselvans: Is it updated tho?

src/pudl/etl/check_foreign_keys.py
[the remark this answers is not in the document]
zaneselvans: Yes, but it only includes the `engine.name` and not the path, which is less informative. See note below.

src/pudl/etl/check_foreign_keys.py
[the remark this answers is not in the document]
zaneselvans: I'm still not sure what the `rowids` refer to here. I could imagine this wording meaning:

- All of the data in the bad rows.
- The dataframe indices of the bad rows.
- The primary keys of the bad rows.
- The SQLite database's internal integer id associated with the bad rows.

src/pudl/etl/check_foreign_keys.py
zaneselvans, line 175: I think this is the logging message you're referring to above, but unlike the logging message which was removed, it doesn't provide the path to the database whose FK relationships it's checking, which could lead to confusion in testing or development if say, the `$PUDL_OUTPUT` envvar isn't set correctly, or isn't set to the value that the user thinks it's set to.

In addition, by default this is checking the `pudl.sqlite` database, so the log message says:

"Running foreign key check on pudl database."

But we're now publishing 11 different SQLite databases (9 for FERC + Census DP1 + PUDL), which are all kinda part of "PUDL" so the message is ambiguous.

zaneselvans (changes requested):

zaneselvans:
Looks like there's an issue with accessing `_sqlite_io_manager`

```
AttributeError: 'IOManagerDefinition' object has no attribute '_sqlite_io_manager'
```

src/pudl/io_managers.py
```diff
@@ -535,10 +503,29 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
+        "read_from_parquet": Field(
+            bool,
+            description="""If True, the canonical source of data for reads
+                will be parquet files. Otherwise, data will be read from the
+                SQLite database.""",
+            default_value=bool(os.getenv("PUDL_READ_FROM_PARQUET")),
+        ),
+    }
+)
+def pudl_io_manager(init_context) -> IOManager:
```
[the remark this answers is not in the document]
bendnorman: Agreed! I think the resource key for assets should be `pudl_io_manager` but the IO managers should have more specific names like `pudl_mixed_format_io_manager` the type of IO manager the `pudl_io_manager` key refers to.

src/pudl/etl/check_foreign_keys.py
rousik: Yeah, that was my original idea as well. I do not like the name clash in that you end up with `pudl.etl.check_foreign_keys.check_foreign_keys()`.

In the end, I moved this to `helpers.py`. That said, there doesn't really seem to be a good place in the codebase for miscellaneous stuff and that might be useful, e.g. `pudl.misc` where you can have assortment of small modules that don't naturally fit elsewhere. Putting this into `check_foreign_keys.py` is a bit awkward IMO, because of the naming as well as the fact that this module is CLI and I think that mixing CLI/binary capabilities and library capabilities (something you'd be importing elsewhere) seems a little dirty.

This is just a thought and I would not block this PR for this, as this can be easily refactored in the future.

src/pudl/io_managers.py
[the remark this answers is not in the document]
rousik: Wouldn't this be achieved via migrating to the new dagster configuration/resources model where we just instantiate whatever iomanager we want under `pudl_io_manager` key?

That said, I do not expect that we will really want to swap out iomanagers here that often if at all. I anticipate that we want pretty uniform behavior and pretty fixed formats. Adding support for dual writes/reads is one case where this is getting violated, but I'm unsure if we really need to be planning on too much flexibility here just yet.

src/pudl/etl/check_foreign_keys.py
rousik: We can use `{engine.url}` here.

rousik (approved):
Looks great, thanks for pushing this past the finish line.

src/pudl/etl/check_foreign_keys.py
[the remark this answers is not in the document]
zschira: I added a link to SQLite's documentation on `rowid`'s, let me know if this feels sufficient. I tried to summarize their docs at first, but it seemed clearer to just add the link.

src/pudl/etl/check_foreign_keys.py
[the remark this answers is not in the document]
zschira: Ah yeah that makes sense, I've added the log back in.

src/pudl/etl/check_foreign_keys.py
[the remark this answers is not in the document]
bendnorman: Yes! These descriptions look good to me. Sorry, I didn't add the descriptions originally. it's not the most straightforward collection of data.

src/pudl/etl/check_foreign_keys.py
bendnorman: This is fine for now. I think I added `check_foreign_keys` to the `SQLiteIOManager` class because I imaged we might have different methods for checking referential integrity for each IO Manager. For example, if we decided to drop sqlite and rely on a combination of parquet files and duckdb we'd need to create a new method of enforcing referential integrity.

src/pudl/io_managers.py
```diff
@@ -31,56 +34,82 @@
...
+        if self.write_to_parquet:
+            self._parquet_io_manager.handle_output(context, obj)
 
-    def __getitem__(self, idx):
-        """Index the fk errors."""
-        return self.fk_errors[idx]
+    def load_input(self, context: InputContext) -> pd.DataFrame:
+        """Reads input from the appropriate IO manager instance."""
+        if self.read_from_parquet:
+            return self._parquet_io_manager.load_input(context)
```
bendnorman, line 111: It'd be helpful to add a comment or docstring that says we prefer to load inputs from parquet files because of speed and type preservation.

src/pudl/io_managers.py
```diff
@@ -535,10 +508,29 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
+        ),
+        "read_from_parquet": Field(
+            bool,
+            description="""If True, the canonical source of data for reads
+                will be parquet files. Otherwise, data will be read from the
+                SQLite database.""",
+            default_value=bool(os.getenv("PUDL_READ_FROM_PARQUET")),
+        ),
+    }
+)
```
bendnorman, line 527: Do we want to parameterize this using environment variables? I'm mildly against it because we don't use env vars to parameterize our other dagster resources.

bendnorman (changes requested):

src/pudl/io_managers.py
zschira: I think it makes sense to keep parquet support turned off by default for local development, so I generally agree with @rousik's reasoning that using environment variables will make it easy to turn on in a nightly build context. I do imagine this could be removed if/when parquet support gets past the experimental phase and we want to turn it on by default, but until then I personally think this would be useful.

zaneselvans: I think if it's turned off by default it'll see much less exercise and we may not catch issues as quickly. The entire PUDL DB is only ~1.3GB in Parquet. If it speeds up the ETL (or later, the full integration tests, or data validations) significantly, It'd be great to have them around.

src/pudl/io_managers.py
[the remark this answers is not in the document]
zaneselvans: I spent a long time confused about where the IO Manager was coming from because the same name was used in every context, so just for legibility I'd prefer that we keep the names distinct here. I agree we don't anticipate swapping in lots of different IO Managers in different test/prod environments, but given that the change is tiny and also helps with legibility I see no reason not to do it.

src/pudl/etl/check_foreign_keys.py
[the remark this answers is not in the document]
zaneselvans: Oh I think having a single logging message is fine, it should just include the path. @rousik notes below that we can use `engine.url` rather than `engine.name`

src/pudl/etl/check_foreign_keys.py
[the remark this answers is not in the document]
zaneselvans: Thanks, this is much clearer.

src/pudl/etl/check_foreign_keys.py
zaneselvans: There's also some actual data transformation creeping into the `etl` subpackage which I thought was just supposed to be laying out the Dagster assets / config stuff (see `glue_assets.py`)

src/pudl/etl/check_foreign_keys.py
```diff
@@ -50,15 +61,156 @@ def pudl_check_fks(logfile: pathlib.Path, loglevel: str):
...
+    check for foreign key failures once all of the data has been loaded into
+    the database using the `foreign_key_check` and `foreign_key_list` PRAGMAs.
+
+    You can learn more about the PRAGMAs in the `SQLite docs
+    <https://www.sqlite.org/pragma.html#pragma_foreign_key_check>`__.
+
+    Raises:
+        ForeignKeyErrors: if data in the database violate foreign key constraints.
+    """
+    logger.info(f"Running foreign key check on {engine.name} database.")
```
zaneselvans, line 177:
```suggestion
    logger.info(f"Running foreign key check on {engine.url} database.")
```

src/pudl/etl/check_foreign_keys.py
```diff
@@ -50,15 +61,156 @@ def pudl_check_fks(logfile: pathlib.Path, loglevel: str):
...
-    context = build_init_resource_context()
-    io_manager = pudl_sqlite_io_manager(context)
+    # Using PudlPaths to get default value for CLI causes validation issues
+    if not db_path:
+        db_path = PudlPaths().output_dir / "pudl.sqlite"
 
-    database_path = io_manager.base_dir / f"{io_manager.db_name}.sqlite"
-    logger.info(f"Checking foreign key constraints in {database_path}")
+    logger.info(f"Checking foreign key constraints in {db_path}")
+    engine = sa.create_engine(f"sqlite:///{db_path}")
```
zaneselvans, line 69:
```suggestion
```

zaneselvans (commented):
It seems like the primary remaining points of discussion include:
- Should we allow configuration of output formats using environment variables?
- Should we enable Parquet outputs by default?

Is there anything else?

src/pudl/io_managers.py
bendnorman: I'm in favor of turning on parquet outputs by default for the reasons Zane listed.

If we wanted to be consistent with how we are configuring the nightly build jobs right now we wouldn't use env vars and instead create a new argument for the `pudl_etl` cli called `--write-parquet-outputs` and pass the value to the `run_config` variable in `pudl.etl.cli`: https://github.com/catalyst-cooperative/pudl/blob/aaa46cbf98be6bf94303015eb358c5234837d3f1/src/pudl/etl/cli.py#L149-L158

I'm worried we'll end up with a confusing mix of methods for configuring runs: env vars, pudl_etl args and hard coded config in python code. I think eventually we should deprecate `pudl_etl` and create a preconfigued job in python for nightly build runs that can be executed using `dagster job execute nightly_build_job`. Something like this:

```python
defs: Definitions = Definitions(
    assets=default_assets,
    resources=default_resources,
    jobs=[
        define_asset_job(
            name="etl_full_nightly_build",
            config=default_config
            | {
                "resources": {
                  "datastore": {
                      "config": {
                          "gcs_cache_path": gcs_cache_path,
                      },
                  },
                 "pudl_io_manager": {
                     "config": {
                          "write_to_parquet": True,
                          "read_from_parquet": True
                     },
                  },
              },
            },
            description="This job the full ETL in a cloud VM.",
        ),
      # Other types of jobs: etl_fast, etl_fast_no_cems ...
    ],
)
```

zaneselvans:
Just switching this to draft for the moment so every small changes doesn't incur costs.

src/pudl/io_managers.py
zschira: Good call, I added to the argument description in the docstring

src/pudl/io_managers.py
zschira: I'm fully convinced into parquet on-by-default, and not using environment variables, so I've up gone that direction.

@bendnorman it would definitely be nice to have all of the configuration for nightly builds consolidated in one place.

zaneselvans: I think switching to using the dagster CLI and their job configuration system directly sounds like a great medium term simplification.

src/pudl/io_managers.py
zaneselvans: Hopefully this ends up being true. 🤞🏼

zaneselvans:
# I'm running the full ETL locally and getting some errors:

## Stale `parquet_path()` argument in `PudlParquetIOManager`
- Removed the `db_name="pudl"` arguments

## Stale `check_foreign_keys()` argument
- replaced `engine` with constructed engine using SQLite URI.

## Reading TZ aware datetime type

### Affected assets
```
Error occurred while loading input "core_ferc714__hourly_demand_pa" of step "load_hourly_demand_matrix_ferc714"
Error occurred while loading input "core_ferc714__hourly_demand_pa" of step "out_ferc714__summarized_demand"
```
### Error:
```
TypeError: Cannot use .astype to convert from timezone-aware dtype to timezone-naive dtype. Use obj.tz_localize(None) or obj.tz_convert('UTC').tz_localize(None) instead.
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/utils.py", line 54, in op_execution_error_boundary
    yield
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/dagster/_core/execution/plan/inputs.py", line 831, in _load_input_with_input_manager
    value = input_manager.load_input(context)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/io_managers.py", line 112, in load_input
    return self._parquet_io_manager.load_input(context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/io_managers.py", line 350, in load_input
    return res.enforce_schema(df)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/metadata/classes.py", line 1482, in enforce_schema
    df = self.format_df(df)
         ^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/metadata/classes.py", line 1463, in format_df
    .astype(dtypes, copy=False)
     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/generic.py", line 6513, in astype
    res_col = col.astype(dtype=cdt, copy=copy, errors=errors)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/generic.py", line 6534, in astype
    new_data = self._mgr.astype(dtype=dtype, copy=copy, errors=errors)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/internals/managers.py", line 414, in astype
    return self.apply(
           ^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/internals/managers.py", line 354, in apply
    applied = getattr(b, f)(**kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/internals/blocks.py", line 616, in astype
    new_values = astype_array_safe(values, dtype, copy=copy, errors=errors)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/dtypes/astype.py", line 238, in astype_array_safe
    new_values = astype_array(values, dtype, copy=copy)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/dtypes/astype.py", line 180, in astype_array
    values = values.astype(dtype, copy=copy)
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.11/site-packages/pandas/core/arrays/datetimes.py", line 708, in astype
    raise TypeError(
```

## EPA CEMS output location

- Previously we had an epacems directory at the top level of `$PUDL_OUTPUT` which contained the quarterly EPA CEMS parquet outputs prior to consolidation.
- Those outputs are consolidated into a single top-level epacems parquet file.
- For uniformity, it seems like moving all of that process into the `parquet` directory and then only plucking the CEMS output from the `parquet` directory for distribution in the nightly build outputs might be better?
- This would require a small tweak to the nightly build script. Would it affect anything else?
- We could continue to have the (5GB) epacems parquet output distributed both as a top level file alongside the SQLite DBs, and in the parquet-only output bucket.

zschira:
@zaneselvans I'll take a look at the errors you've outlined. 

> Defunct PandasParquetIOManager?

As for the `PandasParquetIOManager`, we are using it for reads, but to avoid running out of memory it returns Dask DataFrame, so we definitely should change the name. We could make it the `DaskParquetIOManager`, but I think it just `EpaCemsIOManager` might be better since it's pretty specific to CEMS. I also think it would be nice to expand it to handle writes as well instead of doing writes inside the asset itself, which is kinda confusing.

> EPA CEMS output location

Would it make sense to output the partitioned CEMS files in the `parquet/` directory and put the consolidated file at the top level?

zaneselvans:
I agree renaming `PandasParquetIOManager` makes sense. Should we do that now or get all this merged in and tweak?

Will we be able to get the output method inside an `EpaCemsIOManager` even though it doesn't work on single assets. It takes a whole year's worth of quarterly data, and slices it up into state-level annual data?

I guess writing the quarterly CEMS parquet files out to the same `parquet/` directory would be fine. It's like 100 files, but we already have 100+ files in there.

For distribution, we would want to make sure:
- `core_epacems__hourly_emissions.parquet` is available at the same path as now for public outputs.
- `core_epacems__hourly_emissions.parquet` is also made available inside the `parquet/` directory for the parquet-only distribution.
- The quarterly outputs which get consolidated are **not** distributed to `gs://parquet.catalyst.coop/nightly`
- The quarterly outputs **do** get saved in the `gs://builds.catalyst.coop/$BUILD_ID` bucket in case something goes wrong and we need to look at them.

src/pudl/metadata/classes.py
```diff
@@ -1436,40 +1441,42 @@ def format_df(self, df: pd.DataFrame | None = None, **kwargs: Any) -> pd.DataFra
...
-                        "categorical values in field enum constraint "
-                        f"and will be converted to nulls ({uncategorized})."
-                    )
-        df = (
-            # Reorder columns and insert missing columns
-            df.reindex(columns=dtypes.keys(), copy=False)
+
+        df = df.reindex(columns=dtypes.keys(), copy=False)
+        # Handle dtypes if not using pyarrow types
+        if not use_pyarrow_dtypes:
```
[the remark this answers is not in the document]
zschira: My original thought here was that we could potentially rely on the dtypes we get from reading the parquet files. The IO-manager is passing a schema with dtypes during reads, but running the full ETL like this ended up leading to different errors.

I just reverted that previous commit and changed the pyarrow dtype to remove the `UTC` timezone, which seems more consistent with how we normally deal with dtypes. I tested this and it fixed the problem you highlighted with loading the `core_ferc714__hourly_demand_pa` table. Re-running the full etl now to make sure nothing else breaks with this.

zschira:
I'll go ahead and rename the `PandasParquetIOManager`, and get the outputs in the right place. I think we could apply the `EpaCemsIOManager` to the `ops` that produce the per-year CEMS slices, but I'm not totally sure how easy this will be/if it will actually be cleaner. I'll think about making this change while rearranging where the outputs live, but I won't spend much time on it because I don't want this PR to languish.

src/pudl/etl/epacems_assets.py
```diff
@@ -107,30 +115,36 @@ def consolidate_partitions(context, partitions: list[YearPartitions]) -> None:
         context: dagster keyword that provides access to resources and config.
         partitions: Year and state combinations in the output database.
     """
-    partitioned_path = PudlPaths().output_dir / "core_epacems__hourly_emissions"
-    monolithic_path = PudlPaths().output_dir / "core_epacems__hourly_emissions.parquet"
+    partitioned_path = _partitioned_path()
+    # Create monolithic output in base output directory and parquet directory
+    monolithic_paths = [PudlPaths().output_dir, PudlPaths().output_dir / "parquet"]
```
zaneselvans, line 120: Why write incrementally to both of these locations instead of copying the whole file after the fact as part of the data deployment? Parquet is very fast so I imagine this doesn't add a lot of time, but I'd imagine copying a file is faster.

With the addition of parquet outputs for all of the tables, to me it seems like the `parquet/` directory becomes the canonical parquet output location, and pulling the consolidated file up to the top level for distribution is a special case that is necessitated by our decision not to distribute all of the parquet outputs right now, and the need to have a flat file hierarchy for Zenodo and Kaggle.

Storing two copies of the consolidated file in the nightly build outputs will also increase their size by 5GB, and since we keep 30 days of build outputs, maybe increase overall storage use by ~100GB.

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -367,7 +367,7 @@
 
 
 @asset(
-    io_manager_key="pudl_sqlite_io_manager",
+    io_manager_key="pudl_io_manager",
```
zaneselvans, line 370: While merging in `main` I found this one straggler reference to the old sqlite IO manager.

src/pudl/etl/epacems_assets.py
```diff
@@ -107,30 +115,36 @@ def consolidate_partitions(context, partitions: list[YearPartitions]) -> None:
...
+    for year_partition in partitions:
+        for state in EPACEMS_STATES:
+            for base_path in monolithic_paths:
+                monolithic_path = base_path / "core_epacems__hourly_emissions.parquet"
+                with pq.ParquetWriter(
+                    where=monolithic_path,
+                    schema=schema,
+                    compression="snappy",
+                    version="2.6",
+                ) as monolithic_writer:
```
zaneselvans, line 132: I ran the full ETL locally overnight and the quarterly outputs look as expected, but both copies of the consolidated outputs only contain a single state-year of data: Alaska, 2023 (through Q3 since that's as much data is available in 2023 currently). So I wonder if this isn't overwriting the file entirely rather than appending new row-groups. I think I remember this being why the `with pq.ParquetWriter` was the outermost element of this looping construct.

zschira: Reverted to keeping the `ParquetWriter` at the top level, and only writing to the `parquet/` directory.

src/pudl/etl/epacems_assets.py
zschira: That all makes sense, and I was also running the etl overnight and saw the overwrite errors, so I've gone to just writing to the `parquet/` directory. It seems like that's what we want?

zaneselvans: Yes, I think we can handle all the what-file-goes-where stuff in the "distribution" phase, and keep this focused on data production.

src/pudl/etl/epacems_assets.py
```diff
@@ -115,36 +115,32 @@ def consolidate_partitions(context, partitions: list[YearPartitions]) -> None:
         context: dagster keyword that provides access to resources and config.
         partitions: Year and state combinations in the output database.
     """
-    partitioned_path = _partitioned_path()
-    # Create monolithic output in base output directory and parquet directory
-    monolithic_paths = [PudlPaths().output_dir, PudlPaths().output_dir / "parquet"]
+    partitioned_path = PudlPaths().output_dir / "core_epacems__hourly_emissions"
```
zaneselvans, line 118: Oh I thought having this under the `parquet/` directory made good organizational sense.  Or even just writing all the quarterly files directly out to the `parquet/` directory without the subdirectory would be fine, since we iterate over them by filename, rather than by walking the directory and there's "only" ~100 of them.  That way any parquet file we generate can be found in the same way without needing to think about whether it's in the special partitioned category.

zschira: Oh yeah it was still writing the partitioned outputs to the `parquet/` directory, but was trying to read from the wrong place. Fixed now

docker/gcp_pudl_etl.sh
```diff
@@ -172,8 +172,10 @@ function update_nightly_branch() {
 function clean_up_outputs_for_distribution() {
     # Compress the SQLite DBs for easier distribution
     gzip --verbose "$PUDL_OUTPUT"/*.sqlite && \
-    # Remove redundant multi-file EPA CEMS outputs prior to distribution
-    rm -rf "$PUDL_OUTPUT/core_epacems__hourly_emissions/" && \
+    # Grab the consolidated EPA CEMS outputs for distribution
+    cp "$PUDL_OUTPUT/parquet/core_epacems__hourly_emissions.parquet" "$PUDL_OUTPUT" && \
+    # Remove all other parquet output, which we are not yet distributing.
+    rm -rf "$PUDL_OUTPUT/parquet" && \
```
zaneselvans, line 178: @zschira I updated this to work with the new output locations. Does this look correct to your?

zschira: That looks good to me! Anything else before this is ready to merge?

I did run the ETL locally, and had a weird issue that I think is unrelated. I'm investigating/re-running now.

zaneselvans: I think everything in here looks good to me, but let's wait until we get a good nightly build and can push a release out before we merge in this relatively large change.

zaneselvans (approved):

zaneselvans:
@bendnorman I think we're just waiting to see if your comments were all addressed.

bendnorman (approved):

MichaelTiemannOSC:
Awesome work!

PR #3297, opened 2024-01-25
Update conda lock; tell CodeCov to chill

zaneselvans:
# Overview

- Make sure that the version of `conda-lock` installed in the base conda environment will sort its outputs appropriately.
- Ask CodeCov to give us a little bit of slack vs. the base branch since there's some stochastic elements to our tests and it's sad for this never to pass.

```[tasklist]
# To-do list
```


.codecov.yml
```diff
@@ -2,7 +2,14 @@
...
   round: nearest
-  precision: 1
+  precision: 2
+  status:
+    project:
+      default:
+        target: auto
+        # Allow a little downward flexibility here because some code is only run
+        # sometimes (e.g. zenodo archive downloading)
+        threshold: 0.1%
```
jdangerx, line 12: Is 0.1% too  big of a threshold? Intuitively that seems like it would be easy to add untested code... I guess we can Remain Vigilant in PR review and see if codecov is too lax. But also I'm OK with just continuing to ignore the codecov ❌ when it seems bogus.

jdangerx (approved):
I think I'd lightly prefer to just ignore codecov when it's crying wolf, rather than giving it a threshold, but either works for me. The conda-lock changes seem totally benign.

.codecov.yml
zaneselvans: What I was thinking with the 0.1% is that's functionally been our threshold for caring since there was no visible percentage difference at that level with the previous settings. I doubt anybody goes looking if the coverage "fails" but the number displayed stays 92.7%, so this would make the automated system reflect our human thresholds better.

At the same time, to provide more visibility into changes below that threshold, I upped the precision of the numbers displayed to include two decimal points, so at least we can see what we've decided to ignore.

There's some natural variability in the test coverage we get because we're using the GitHub caching system for our input data, so the code that pulls from Zenodo directly only gets hit when the cache expires and needs to be refreshed (or we use a new archive that's not cached). I think that's more than a 0.1% difference though (we currently have 13k lines of code so 0.1% is 13 lines)

zaneselvans: I think having the test be "failing" so frequently also inures people to it, so it ends up more likely to get ignored, vs if it's only failing at times when we really do want to do further investigation, I think it's more likely we'll pay attention it and use it as a real signal.

PR #3299, opened 2024-01-26
Add retries, and use fsspec to handle GCS

### LINKED ISSUE #3298: Nightly build failure 2024-01-26
# Overview

Seems like a Zenodo upload failure again.

# Next steps

What next steps do we need to do to understand or remediate the issue?

It looks like an actual Zenodo file upload failed, instead of just the publication - we should check the actual Zenodo draft and see if everything made it. If not, we should:

```[tasklist]
- [x] add retry logic to `zenodo_data_release.py`
- [x] add `fsspec` logic to `zenodo_data_release.py`
- [x] tell `zenodo_data_release.py` to pull from the nightly build bucket
```

```[tasklist]
- [x] validation tests passed
- [x] [S3 distribution bucket](https://s3.console.aws.amazon.com/s3/buckets/pudl.catalyst.coop?region=us-west-2&bucketType=general&prefix=nightly/&showversions=false) was updated at the expected time
- [x] [GCP distribution bucket](https://console.cloud.google.com/storage/browser/pudl.catalyst.coop/nightly;tab=objects?project=catalyst-cooperative-pudl) was updated at the expected time
- [x] [GCP internal bucket](https://console.cloud.google.com/storage/browser/builds.catalyst.coop) was updated at the expected time
- [x] [Datasette PUDL version](https://data.catalyst.coop/pudl/core_pudl__codes_datasources) points at the same hash as [nightly](https://github.com/catalyst-cooperative/pudl/tree/nightly)
- [x] [Zenodo sandbox record](https://sandbox.zenodo.org/doi/10.5072/zenodo.5563) was updated to the record number in the logs (search for `zenodo_data_release.py` and `Using metadata from` in the logs, to see what the new record number should be!)
```

# Relevant logs
[link to build logs from internal distribution bucket](https://console.cloud.google.com/storage/browser/_details/builds.catalyst.coop/2024-01-26-0602-bb1a0c08b-main/2024-01-26-0602-bb1a0c08b-main-pudl-etl.log?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22)))

```
2024-01-26 12:46:25,321: INFO - Uploading file to https://sandbox.zenodo.org/api/files/e728d1ee-73e4-4364-af3b-b5e97ca21663/ferc2_dbf.sqlite.gz (zenodo_data_release.py:162)
  File "/home/mambauser/env/lib/python3.11/site-packages/urllib3/connectionpool.py", line 467, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/home/mambauser/env/lib/python3.11/site-packages/urllib3/connectionpool.py", line 462, in _make_request
    httplib_response = conn.getresponse()
                       ^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.11/http/client.py", line 1386, in getresponse
    response.begin()
  File "/home/mambauser/env/lib/python3.11/http/client.py", line 325, in begin
    version, status, reason = self._read_status()
                              ^^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.11/http/client.py", line 286, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.11/socket.py", line 706, in readinto
    return self._sock.recv_into(b)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.11/ssl.py", line 1315, in recv_into
    return self.read(nbytes, buffer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.11/ssl.py", line 1167, in read
    return self._sslobj.read(len, buffer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TimeoutError: The read operation timed out
```

jdangerx:
# Overview

Closes #3298 .

Zenodo sandbox has been a bit flaky, this adds retries to handle the flakiness & also lets us simply kick off manual retries by syncing directly from a cloud bucket.

# Testing

Ran the dang thing with both `gs://pudl.catalyst.coop/nightly` as well as a local directory, and saw that the Zenodo draft release had the correct files in it each time.

In terms of automated testing, as of now we basically test it every time we run a nightly build. If we want to test it in CI or something, I think we'd have to set up:

- a small GCS bucket with some stuff in it
- a sandbox Zenodo record to run tests in

And then we would have to try syncing from GCS to that record, verifying that the record has what we expect, and then doing the same with a local directory. We could move this into some sort of `pudl/scripts` module and then we could get coverage metrics for it and use its methods in pytest. 

```[tasklist]
# To-do list
- [x] manual testing
```

### DIFF
--- a/devtools/zenodo/zenodo_data_release.py
+++ b/devtools/zenodo/zenodo_data_release.py
@@ -7,7 +7,7 @@ import tempfile
 import time
 from dataclasses import dataclass
 from pathlib import Path
-from typing import BinaryIO
+from typing import IO
 
 import click
 import fsspec
@@ -92,15 +92,13 @@ class ZenodoClient:
         Passes method, url, and **kwargs to requests.request.
         """
         base_timeout = 2
-        try_num = 1
-        while try_num < max_tries:
+        for try_num in range(1, max_tries):
             try:
                 return requests.request(method=method, url=url, **kwargs)
             except requests.RequestException as e:
                 timeout = base_timeout**try_num
                 logger.warning(f"Attempt #{try_num} Got {e}, retrying in {timeout} s")
                 time.sleep(timeout)
-            try_num += 1
 
         # don't catch errors on the last try.
         return requests.request(method=method, url=url, **kwargs)
@@ -114,7 +112,7 @@ class ZenodoClient:
             method="GET",
             url=f"{self.base_url}/deposit/depositions/{deposition_id}",
             headers=self.auth_headers,
-            timeout=10,
+            timeout=5,
         )
         logger.debug(
             f"License from JSON for {deposition_id} is "
@@ -131,7 +129,7 @@ class ZenodoClient:
             method="GET",
             url=f"{self.base_url}/records/{record_id}",
             headers=self.auth_headers,
-            timeout=10,
+            timeout=5,
         )
         return _NewRecord(**response.json())
 
@@ -145,7 +143,7 @@ class ZenodoClient:
             method="POST",
             url=f"{self.base_url}/records/{record_id}/versions",
             headers=self.auth_headers,
-            timeout=10,
+            timeout=5,
         )
         return _NewRecord(**response.json())
 
@@ -161,7 +159,7 @@ class ZenodoClient:
         data = {"metadata": metadata.model_dump()}
         logger.debug(f"Setting metadata for {deposition_id} to {data}")
         response = self.retry_request(
-            method="PUT", url=url, json=data, headers=self.auth_headers, timeout=10
+            method="PUT", url=url, json=data, headers=self.auth_headers, timeout=5
         )
         return _LegacyDeposition(**response.json())
 
@@ -174,11 +172,11 @@ class ZenodoClient:
             method="DELETE",
             url=f"{self.base_url}/deposit/depositions/{deposition_id}/files/{file_id}",
             headers=self.auth_headers,
-            timeout=10,
+            timeout=5,
         )
 
     def create_bucket_file(
-        self, bucket_url: AnyHttpUrl, file_name: str, file_content: BinaryIO
+        self, bucket_url: AnyHttpUrl, file_name: str, file_content: IO[bytes]
     ) -> requests.Response:
         """LEGACY API: Upload a file to a deposition's file bucket.
 
@@ -188,14 +186,14 @@ class ZenodoClient:
         url = f"{bucket_url}/{file_name}"
         logger.info(f"Uploading file to {url}")
 
-        # don't worry about timeout=10s - that's "time for the server to
+        # don't worry about timeout=5s - that's "time for the server to
         # connect", not "time to upload whole file."
         response = self.retry_request(
             method="PUT",
             url=url,
             headers=self.auth_headers,
             data=file_content,
-            timeout=10,
+            timeout=5,
         )
         return response
 
@@ -205,7 +203,7 @@ class ZenodoClient:
             method="POST",
             url=f"{self.base_url}/deposit/depositions/{deposition_id}/actions/publish",
             headers=self.auth_headers,
-            timeout=10,
+            timeout=5,
         )
         return _LegacyDeposition(**response.json())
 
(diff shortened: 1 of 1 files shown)


devtools/zenodo/zenodo_data_release.py
```diff
@@ -259,11 +266,27 @@ def sync_directory(self, source_dir: SourceData) -> "ContentComplete":
...
+        files = fsspec.open_files(globbed, mode="rb")
+        if len(files) == 0:
+            raise ValueError(
+                f"No files found in {globbed}! Did you pass the right dirname?"
             )
-            logger.info(f"Upload to {bucket_url}/{name} got {response.text}")
+        for openable_file in files:
+            name = Path(openable_file.path).name
+            with openable_file.open() as of:
+                if "local" not in openable_file.fs.protocol:
```
jdangerx, line 277: Should clean this up a bit. `fsspec` *tries* to paper over the differences between remote/local, but sometimes they're actually important (like when we don't want to do a remote read() on the file every time we need to retry due to Zenodo).

So, I should extract this into some sort of "get the file that we want to upload - which will either be the local file itself, or a tempfile we got from a remote place."

devtools/zenodo/zenodo_data_release.py
```diff
@@ -3,12 +3,14 @@
 import datetime
 import logging
 import os
-from collections.abc import Iterable
+import tempfile
+import time
 from dataclasses import dataclass
 from pathlib import Path
-from typing import BinaryIO
+from typing import IO
```
jdangerx, line 10: See [this issue](https://github.com/python/typeshed/issues/7843) - properly type annotating "file-like-objects" in Python is apparently complicated, but the "blessed" option is `IO[bytes]`.

devtools/zenodo/zenodo_data_release.py
jdangerx, line 1: Conveniently, the first half of the diff is retries and the second half is fsspec!

devtools/zenodo/zenodo_data_release.py
```diff
@@ -331,22 +368,20 @@ def get_html_url(self):
...
     "--source-dir",
-    type=click.Path(
-        exists=True, file_okay=False, dir_okay=True, readable=True, path_type=Path
-    ),
+    type=str,
     required=True,
-    help="Path to a directory whose contents will be uploaded to Zenodo. Subdirectories are ignored.",
+    help="Path to a directory whose contents will be uploaded to Zenodo. "
+    "Subdirectories are ignored. Can get files from GCS as well - just prefix "
+    "with gs://.",
```
zaneselvans, line 375: Would this also work with `s3://` paths?

I guess there's no real benefit from that though, since we're pulling the data down and then pushing it up rather than copying between remote locations directly. So it'll probably be faster and less flaky to pull from GCS since we're running on teh google.

devtools/zenodo/zenodo_data_release.py
```diff
@@ -84,13 +86,31 @@ def __init__(self, env: str):
...
 
         Depositions can be published *or* unpublished.
         """
-        response = requests.get(
-            f"{self.base_url}/deposit/depositions/{deposition_id}",
+        response = self.retry_request(
+            method="GET",
+            url=f"{self.base_url}/deposit/depositions/{deposition_id}",
             headers=self.auth_headers,
             timeout=5,
```
zaneselvans, line 115: I don't think it really matters, but it seems funny that we start with a timeout of 5 seconds, and then we have to retry, we do with `2**1` (2) seconds, and then `2**2`(4) seconds.  Or does the `timeout` that's hiding inside of `**kwargs` not get overwritten when `timeout` is assigned in the `except` clause?

zaneselvans (approved):
Left a couple of questions for my personal understanding if you get a chance to respond.

PR #3300, opened 2024-01-26
Parameterize the FERC1 transform step what transfers quarterly filed data into annual columns

cmgosnell:
# Overview

a wrap up of #3234 bc i forgot to really turn on the fix in #3234, plus i found this same missing data in the assets table, so I made this a standard transform and applied it to both tables

# Testing

```[tasklist]
# To-do list
```


src/pudl/transform/ferc1.py
```diff
@@ -716,6 +716,48 @@ def combine_axis_columns_xbrl(
     return df
 
 
+class AssignQuarterlyFiledDataToAnnualDbf(TransformParams):
+    """Parameters for transfering quarterly reported data to annual columns."""
+
+    annual_cols: list[str] = []
+    quarterly_cols: list[str] = []
```
zaneselvans, line 723: There's an implicit 1:1 linkage between the individual elements of these two lists.  If they are in different orders such that `annual_cols[n]` does not correspond to `quarterly_cols[n]` then the result of the transformation will be incorrect.  A data structure that explicitly ties the quarterly and annual columns to each other feels like it would be more robust, like a dictionary that maps individual quarterly columns to their corresponding annual columns, and would also mean you don't need to validate that you've got the same number of columns in each attribute.

src/pudl/transform/ferc1.py
```diff
@@ -716,6 +716,48 @@ def combine_axis_columns_xbrl(
...
+
+
+def assign_quarterly_filed_data_to_annual_dbf(
+    df: pd.DataFrame, params: AssignQuarterlyFiledDataToAnnualDbf
+) -> pd.DataFrame:
+    """Transfer the quarterly reported data to the annual columns.
+
+    For some reason in the dbf data for this table reported all of the
+    balance data as quarterly data between specific years. We already choose
+    the end of the year in :meth:`select_annual_rows_dbf`.
```
zaneselvans, line 744:
```suggestion
    For some reason in the dbf data for this table reported all of the
    balance data as quarterly data between specific years. We already choose
    the end of the year in :meth:`select_annual_rows_dbf`. This ensures that
    by this point, any quarterly data remaining in the input dataframe pertains
    to the 4th quarter.
```

src/pudl/transform/ferc1.py
```diff
@@ -716,6 +716,48 @@ def combine_axis_columns_xbrl(
...
+                "There must be the same number of annual columns and quarterly "
+                f"columns but found: {self.annual_cols=} & {self.quarterly_cols=}"
+            )
+        return self
+
+
+def assign_quarterly_filed_data_to_annual_dbf(
+    df: pd.DataFrame, params: AssignQuarterlyFiledDataToAnnualDbf
+) -> pd.DataFrame:
+    """Transfer the quarterly reported data to the annual columns.
```
zaneselvans, line 740:
```suggestion
    """Transfer 4th quarter reported data to the annual columns.
```

zaneselvans (changes requested):
I think using a dictionary or other structure that explicitly maps the quarterly columns to their annual counterparts would be more robust.

src/pudl/transform/ferc1.py
cmgosnell: this is a good idea I'll convert these two guys to a dict

src/pudl/transform/ferc1.py
```diff
@@ -716,6 +716,41 @@ def combine_axis_columns_xbrl(
     return df
 
 
+class AssignQuarterlyDataToYearlyDbf(TransformParams):
+    """Parameters for transfering quarterly reported data to annual columns."""
+
+    quarterly_to_yearly_column_map: dict[str, str] = {}
+    quarterly_filed_years: list[int] = []
```
zaneselvans, line 723: There was a mix of "quarterly-to-yearly" and "yearly-to-quarterly" naming in the classes, functions, and parameters which was confusing. I made it all quarterly-to-early.

zaneselvans (approved):

src/pudl/transform/ferc1.py
```diff
@@ -1541,7 +1541,7 @@ class Ferc1TableTransformParams(TableTransformParams):
     merge_xbrl_metadata: MergeXbrlMetadata = MergeXbrlMetadata()
     align_row_numbers_dbf: AlignRowNumbersDbf = AlignRowNumbersDbf()
     drop_duplicate_rows_dbf: DropDuplicateRowsDbf = DropDuplicateRowsDbf()
-    assign_quarterly_filed_data_to_annual_dbf: AssignQuarterlyDataToYearlyDbf = (
+    assign_quarterly_data_to_yearly_dbf: AssignQuarterlyDataToYearlyDbf = (
```
zaneselvans, line 1544: D'oh! I fucked it up. I was editing in the GitHub online editor because I was running a full ETL in the background. Sorry!

PR #3302, opened 2024-01-28
Integrate splink model for FERC to EIA record linkage

katie-lamb:
# Overview
Integrate the `splink` model into PUDL to conduct the FERC to EIA entity matching.

Closes [32](https://github.com/catalyst-cooperative/ccai-entity-matching/issues/32) and [10](https://github.com/catalyst-cooperative/ccai-entity-matching/issues/110) but more generally is one of the last lingering things in that CCAI repo.

The bottleneck right now is actually the name cleaning. There's probably a way to speed this up?

Also, the devtools notebook that I added is the best way I could come up with to show a bunch of visuals about the model. I'm not sure what the best way to maintain this notebook is. Or if the visualizations could somehow be integrated into the experiment tracking.

After merging this we should be able to directly compare the output of the existing PUDL model that uses logistic regression to perform entity matching for FERC1 to EIA to this `splink` model. Then we can choose which one we want to primarily use (I think probably splink) based on what metrics we want to prioritize. I also have a call scheduled with RMI to get their input on comparing matches and metrics to prioritize.

[This comment](https://github.com/catalyst-cooperative/ccai-entity-matching/issues/110#issuecomment-1907398826) gives a first look at how the splink model compares to the PUDL results.

Is it okay to temporarily have both match results in the DB?

What did you change?

Integrated a new module with the `splink` model for entity matching.

# Testing

# Pros and Cons of Splink
**Pros**
- Model performance is less dependent on the quality of the training data
- Active discussion and development in the splink repo. A lot of new features seem to be going in.
- Lots of helpful visualizations and tools to make sure the model is doing what you think it should be doing and parameters are being weighted in an intuitive way
- The results from conducting an unsupervised match look similar
- Seems like they're headed in the direction of using labeling functions to create more training data that iteratively improves model results.
- We don't rely on the manually created and slightly fishy `plant_id_pudl` or `utility_id_pudl` for blocking

**Cons**
- If development stopped on `splink` then the structure is less transferrable to a more traditional linear model built with `sklearn`.
- There's maybe some weirdness with dependencies, but that's maybe always sort of the case when we add cool new dependencies to PUDL. 

```[tasklist]
# To-do list
- [x] Figure out a fix for the way `poetry` handles the `sqlalchemy` dependency in `splink` that makes it incompatible with `pudl` (see above discussion).
- [x] Add in a `devtools` notebook that can output all the helpful `splink` visualizations for this match as well as the unsupervised matching model
- [x] Track precision and recall
- [x] Improve performance of blocking rules - try soundex/dmetaphone, try faiss
- [x] Move metrics to experiment tracking infra
- [x] Do [a comparison of all the metrics](https://github.com/catalyst-cooperative/pudl/pull/3302#issuecomment-1965684234) in `prettyify_best_matches` between the splink and existing PUDL model.
- [x] Create spinoff issues to refactor the dataframe embedder factory vs cleaning factory and to move the model config to Dagster config so it can be set through the UI.
- [x] Go through remaining TODO comments
- [x] Actually persist matching results to DB once we're done tinkering with it
- [x] Remove old model and update asset names
- [x] Add more testing
```


zaneselvans:
I think the "right" solution here (if we can get it implemented) is probably for `splink` to remove the `sqlalchemy<2` constraint. If it's really only there to prevent issues with `pandas<2` that should probably be the responsibility of users who are depending on `splink` -- if `splink` allows any version of `sqlalchemy` then users who need to use `pandas<2` for some reason can pin their own dependencies. I [don't see any issues related to this](https://github.com/moj-analytical-services/splink/issues?q=is%3Aissue+sqlalchemy) in the `splink` repo, so maybe we should create one to discuss and see if there are other blocking issues for them. Given that SQLAlchemy 2.x has been out for a year, and pandas 2.0 has been out for 9 months, I don't think this would be an unreasonable request. Pandas 3.0 is [slated to come out this spring](https://github.com/pandas-dev/pandas/milestone/102).

As far as where the dependency restriction we're specifically running into is coming from, if we're installing from `conda-forge`, it'll be in the [recipe for the conda package](https://github.com/conda-forge/splink-feedstock/blob/main/recipe/meta.yaml):

```yaml
requirements:
  host:
    - python >=3.8,<4.0
    - pip
    - poetry >=0.12
  run:
    - python >=3.8,<4.0
    - jsonschema >=3.2,<5.0
    - python-duckdb >=0.8.0
    - pandas >=1.2.0
    - sqlglot >=7.0.0
    - altair >=5.0.1,<6.0
    - jinja2 >=3.0.3,<4.0.0
  run_constrained:
    - pyspark >=3.2.1
    - sqlalchemy >=1.4.0,<2.0.0
```

I had not previously encountered `run_constrained` but [here's the documentation from conda](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained).  The conda recipe dependencies are typically manually compiled, so this is probably an imperfect translation of the poetry dependency specs. Also I think one *could* probably implement the less restrictive dependencies just by changing the conda packaging, but that's kinda hacky. And who knows maybe there are some other real issues with sqlalchemy 2.

Just for testing, you might try manually installing splink and its dependencies into the `pudl-dev` environment with `pip` and see if everything works, or if it runs into real problems.

katie-lamb:
> I think the "right" solution here (if we can get it implemented) is probably for `splink` to remove the `sqlalchemy<2` constraint.

Ya I think that's probably best as well. I didn't notice that they're not on the latest pandas version. I'll make an issue, they're usually pretty responsive.

> Just for testing, you might try manually installing splink and its dependencies into the `pudl-dev` environment with `pip` and see if everything works, or if it runs into real problems.

Yep, that's what I've been doing for now.

zaneselvans:
See also this issue in the splink repo: https://github.com/moj-analytical-services/splink/issues/1906

src/pudl/analysis/record_linkage/eia_ferc1_splink_rule_definitions.py
```diff
@@ -0,0 +1,87 @@
+import splink.duckdb.comparison_level_library as cll
+import splink.duckdb.comparison_library as cl
+import splink.duckdb.comparison_template_library as ctl
+from splink.duckdb.blocking_rule_library import block_on
+
+blocking_rule_1 = "l.report_year = r.report_year and jaro_winkler_similarity(l.plant_name, r.plant_name) >= .7"
```
RobinL, line 6: This filter condition could have [negative performance consequences](https://moj-analytical-services.github.io/splink/topic_guides/blocking/performance.html#efficient-blocking-rules).   

(it's not a problem if it's running fast enough for your needs)

Two options for improving performance whilst retaining a 'similar' intention could be:
replacing 
`jaro_winkler_similarity(l.plant_name, r.plant_name) >= .7`
with  a substring like
`substr(l.plant_name,1,2) = substr(r.plant_name,1,2)`

or possibly pre-processing the input data to take a soundex/dmetaphone of `plant_name` and blocking on dmeta_plan_name

katie-lamb: Got it! I think replacing these `jaro_winkler_similarity` blocking rules makes sense.

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -108,6 +108,42 @@ def embed_dataframe_graph(df: pd.DataFrame) -> FeatureMatrix:
     return embed_dataframe_graph
 
 
+def dataframe_cleaner_factory(
```
katie-lamb, line 111: @zschira I made this "cleaner factory" which is very similar to the "embedding factory" except that it doesn't have `weights` and also the output of the transform and fit steps is a pandas dataframe instead of a numpy array or spare matrix. It might be cleaner to just adapt the embedding factory to work for both? Not sure. They serve different purposes - the embedding factory results in a matrix embedding versus the cleaner factory is only intended to clean up some columns. I could see how this can be confusing to a user.

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -108,6 +108,42 @@ def embed_dataframe_graph(df: pd.DataFrame) -> FeatureMatrix:
...
+
+    @op(name=f"{name_prefix}_train")
+    def train_dataframe_cleaner(df: pd.DataFrame):
+        """Train :class:`sklearn.compose.ColumnTransformer` on input."""
+        column_transformer = ColumnTransformer(
+            transformers=[
+                (name, column_transform.as_pipeline(), column_transform.columns)
+                for name, column_transform in vectorizers.items()
+            ],
+        ).set_output(transform="pandas")
```
katie-lamb, line 124: I kind of think this slows down performance but maybe I'm wrong and my computer is just moving slow. I could instead have it output a numpy array (default) and then create a dataframe from it with assigned column names.

src/pudl/analysis/record_linkage/embed_dataframe.py
```diff
@@ -190,6 +226,73 @@ def as_transformer(self):
...
+    df: pd.DataFrame, fuel_type_col: str, name_col: str
+) -> pd.DataFrame:
+    """Impute missing fuel type data from a name column.
+
+    If a missing fuel type code is contained in the plant name,
+    fill in the fuel type code PUDL for that record. E.g. "Washington Hydro"
+    """
+    if fuel_type_col not in df.columns:
+        raise AssertionError(f"{fuel_type_col} is not in dataframe columns.")
+    # TODO: dynamically get this list from core_eia__codes_energy_sources
```
katie-lamb, line 260: What is the best way to dynamically generate this fuel type list from `core_eia__codes_energy_sources`? I guess it could be a dagster op with that table as an input dependency, but that seems overly complicated. Generate it from the `fuel_type_code_pudl` column of the "core_eia__codes_energy_sources" table of `pudl.metadata.eia`? That seems cleaner. I guess it doesn't need to be dynamically generated but it would be nice if it was.

src/pudl/analysis/record_linkage/eia_ferc1_splink_model_config.py
```diff
@@ -0,0 +1,93 @@
+"""The model parameters for the FERC1 to EIA splink record linkage model.
```
katie-lamb, line 1: Does it make sense to have these model parameter definitions hang out in this separate module?

src/pudl/analysis/record_linkage/eia_ferc1_splink_model_config.py
```diff
@@ -0,0 +1,93 @@
...
+This module enumerates the blocking rules as well as the comparison levels
+for the matching columns that are used in the FERC1 to EIA record linkage
+model.
+"""
+import splink.duckdb.comparison_level_library as cll
+import splink.duckdb.comparison_library as cl
+import splink.duckdb.comparison_template_library as ctl
+from splink.duckdb.blocking_rule_library import block_on
+
+blocking_rule_1 = "l.report_year = r.report_year and jaro_winkler_similarity(l.plant_name, r.plant_name) >= .7"
```
katie-lamb, line 12: TODO for me: update these blocking rules to use an alternative to `jaro_winkler_similarity` to improve model performance.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage_splink_model.py
```diff
@@ -0,0 +1,246 @@
...
+        [eia_df, ferc_df],
+        input_table_aliases=["eia_df", "ferc_df"],
+        settings_dict=settings_dict,
+    )
+    linker.register_table(train_df, "training_labels", overwrite=True)
+    linker.estimate_u_using_random_sampling(max_pairs=1e7)
+    linker.estimate_m_from_pairwise_labels("training_labels")
+    # set a match probability threshold?
+    # do this at a bunch of different thresholds and print out stats
+    preds_df = linker.predict(threshold_match_probability=0.9)
```
katie-lamb, line 181: It would be nice if here we could get precision and recall stats for predictions at different match thresholds. Can we track this through experiment tracking infrastructure?

katie-lamb:
@zschira there's still a lengthy tasklist for this PR but I think you can give it a first look now and familiarize.

pyproject.toml
```diff
@@ -69,6 +69,7 @@ dependencies = [
     "scikit-learn>=1.3",
     "scipy>=1.11",
     "Shapely>=2",
+    "splink>=3.9.11",
```
zaneselvans, line 72: It looks like [splink put out a new 3.9.12 release](https://github.com/moj-analytical-services/splink/releases/tag/v3.9.12) with full SQLAlchemy 2.x support (not just a packaging change), which might be good to use as the minimum allowed version.

pyproject.toml
```diff
@@ -73,6 +73,7 @@ dependencies = [
     "scikit-learn>=1.3",
     "scipy>=1.11",
     "Shapely>=2",
+    "splink>=3.9.12",
```
zaneselvans, line 76: I'm surprised that `duckdb` isn't showing up anywhere in the dependencies. Is that expected?

src/pudl/analysis/record_linkage/__init__.py
```diff
@@ -2,6 +2,8 @@
 from . import (
     classify_plants_ferc1,
     eia_ferc1_record_linkage,
+    eia_ferc1_record_linkage_splink_model,
+    eia_ferc1_splink_model_config,
```
zaneselvans, line 6: These module names are quite long. Is the idea that their contents will get rolled into the more generic `eia_ferc1_record_linkage` module once splink is the only option? If not, can we simplify them? The fact that they're already in the `record_linkage` subpackage and that `splink` is dedicated to record linkage makes that bit of the name seem like it might be superfluous.  Maybe `eia_ferc1_splink_model` and `eia_ferc1_splink_config`?

katie-lamb: Yep, the idea is that it would fully replace the other module, in which case I think `eia_ferc1_model` would be fine.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage_splink_model.py
```diff
@@ -0,0 +1,388 @@
...
+    connected_df = add_null_overrides(
+        connected_df
+    )  # Override specified values with NA record_id_eia
+    return Resource.from_id(
+        "out_pudl__yearly_assn_eia_ferc1_plant_parts"
+    ).enforce_schema(connected_df)
+
+
+@models.pudl_model(
+    "out_pudl__yearly_assn_eia_ferc1_plant_parts_splink",
```
katie-lamb, line 339: This asset name will be updated once the existing PUDL model is replaced, it won't have the `splink` suffix.

katie-lamb:
@zschira I know this is a mega big PR so I think I left a comment everywhere that I had questions. Besides that, I think what would be most helpful is if you tried to create the environment and run the splink model, to make sure everything works for you. The idea is that this PR will fully replace the linear regression model with this splink model since results are comparable and this model is more adaptable and efficient. When this PR is ready to go, I'll take out the old model and rename the splink assets to be the old asset name.

src/pudl/analysis/record_linkage/embed_dataframe.py
zschira: It looks like you might be able to use `pudl.metadata.codes.CODE_METADATA["core_eia__codes_energy_sources"]`. This is used to set constraints from `fuel_type_code_pudl`, so I think all the information you need is there

zschira (commented):
The new model is running well on my computer! I do think it would be good to get the splink configuration integrated into Dagster's config, and to consolidate the name cleaner/dataframe embedding stuff, but maybe we can push that to another PR so we can get the splink model integrated sooner?

src/pudl/analysis/record_linkage/eia_ferc1_splink_model_config.py
katie-lamb: Ah good idea. I'll make an issue in SEC-EIA land so that we can do this for the next splink model we create.

src/pudl/analysis/record_linkage/embed_dataframe.py
katie-lamb: I'll make an issue in SEC-EIA land to refactor this.

zschira (approved):
Lets get it merged!

PR #3331, opened 2024-01-31
Make an EIA860m Changelog table

### LINKED ISSUE #3330: Make a eia80m changelog table
Our existing system can process monthly 860M data from July 2015 to present, but currently only includes the most recent result. We would need to alter it to include and combine all months. This will allow us to see trends over time instead of just a snapshot.

```[tasklist]
- [ ] Add a full date range for 860m in `working_partitions`
- [ ] Grab the most recent month for the standard/current eia860m extraction
- [ ] Minor update to data extraction process to read all data at once
- [ ] Tweak data cleaning code to handle all data at once
- [ ] Add database tables with these new versions
```

(this is for dbcp)

cmgosnell:
# Overview

Closes #3330.

What did you change?

# Testing

```[tasklist]
# To-do list
- [ ] Update the [release notes](../docs/release_notes.rst): reference the PR and related issues. ( i haven't added this bc 
```


src/pudl/package_data/eia860m/column_maps/generator_existing.csv
```diff
@@ -30,5 +30,5 @@ longitude,,,,,,longitude,longitude,longitude,longitude,longitude,longitude,longi
...
+energy_storage_capacity_mwh,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh,nameplate_energy_capacity_mwh
+net_capacity_mwdc,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw,dc_net_capacity_mw
```
cmgosnell, line 34: found two old missing columns when running the full extract.

src/pudl/extract/eia860m.py
```diff
@@ -85,3 +90,51 @@ def append_eia860m(eia860_raw_dfs, eia860m_raw_dfs):
             sort=True,
         )
     return eia860_raw_dfs
+
+
+@asset(
+    required_resource_keys={"datastore", "dataset_settings"},
+)
+def raw_eia860m__all_dfs(context):
```
cmgosnell, line 98: this guy is kinda a hack to mirror some of the behavior in the other excel extractors that use `raw_df_factory`

src/pudl/helpers.py
```diff
@@ -620,25 +620,21 @@ def expand_timeseries(
...
+            date_col
+        ] + pd.tseries.offsets.DateOffset(months=1)
         end_dates.loc[:, date_col] = pd.to_datetime(
             {
                 "year": end_dates[date_col].dt.year,
-                "month": end_dates[date_col].dt.month + 1,
+                "month": end_dates[date_col].dt.month,
                 "day": 1,
             }
         )
```
cmgosnell, line 632: this was wreaking havoc because if a month in end_dates was "12" it would get reassigned to 13 which is an invalid month. but i needed to reassign the day here back to 1 in case the max date's day was 31 and the `freq` was `D` because it would expect to expand the time_series through the end of the `fill_through_freq` which would require the end_date here to be the first day of the next month. Its a weird expectation but hey there are unit tests and this new setup conforms to that.

src/pudl/package_data/settings/etl_full.yml
```diff
@@ -239,7 +239,105 @@ datasets:
           2021,
           2022,
         ]
-      eia860m: true
+      eia860m_year_months:
```
cmgosnell, line 242: this is a looooot of year_months in the settings here... i could convert to using "all" like CEMS.

src/pudl/extract/eia860.py
```diff
@@ -106,10 +105,8 @@ def extract_eia860(context, raw_eia860__all_dfs):
...
 
     if eia_settings.eia860.eia860m:
-        eia860m_data_source = DataSource.from_id("eia860m")
-        eia860m_date = eia860m_data_source.working_partitions["year_month"]
         eia860m_raw_dfs = pudl.extract.eia860m.Extractor(ds).extract(
-            year_month=eia860m_date
+            year_month=eia_settings.eia860.eia860m_date
         )
         raw_eia860__all_dfs = pudl.extract.eia860m.append_eia860m(
             eia860_raw_dfs=raw_eia860__all_dfs, eia860m_raw_dfs=eia860m_raw_dfs
```
cmgosnell, line 112: instead of re-extracting the one-month 860m for the normal 860 assets, we could now rely on the raw 860m assets. But rn they are not generated iteratively, but rather in one big clump which takes a while because there are a loooot of `year_month`s. So making the normal raw 860 assets depend on the new raw 860m assets would mean effectively delaying when we run 860. So i didn't do this. also it wouldn't save that much time bc one year of 860m is nbd. but idk @bendnorman

src/pudl/metadata/resources/eia860m.py
```diff
@@ -0,0 +1,56 @@
+"""Definitions of data tables primarily coming from EIA-860m."""
+from typing import Any
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "core_eia861__monthly_generators_changelog": {
```
bendnorman, line 5: Is this suppose to be yearly or monthly? I'm getting an error that `core_eia860m__yearly_generators_changelog_eia860m` doesn't exist in the metadata.

src/pudl/package_data/settings/etl_fast.yml
```diff
@@ -42,7 +42,7 @@ datasets:
       years: [2020, 2023]
     eia860:
       years: [2020, 2022]
-      eia860m: true
+      eia860m_year_months: ["2023-06"]
```
bendnorman, line 45: Does it make sense produce a change log from one month of data? Should we load the last year or all of the data?

src/pudl/extract/eia860.py
bendnorman: That seems fine as long as we create an issue to track the extraction improvments.

src/pudl/metadata/resources/eia860m.py
cmgosnell: doh yea i junked up the name of the transform asset. (and i smashed the materialize all assets after converting this to a ready for review and got two errors including this one) i will fix.

src/pudl/settings.py
```diff
@@ -213,10 +213,12 @@ class Eia860Settings(GenericDatasetSettings):
 
     data_source: ClassVar[DataSource] = DataSource.from_id("eia860")
     eia860m_data_source: ClassVar[DataSource] = DataSource.from_id("eia860m")
-    eia860m_date: ClassVar[str] = eia860m_data_source.working_partitions["year_month"]
-
     years: list[int] = data_source.working_partitions["years"]
-    eia860m: bool = True
+    eia860m_year_months: list[str] = eia860m_data_source.working_partitions[
+        "year_months"
+    ]
```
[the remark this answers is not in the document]
cmgosnell: I was wondering that... i think it might make sense but they are intertwined so I hesitated. This old boolean was used to determine whether one month of eia860m was incorporated into the core 860 assets.

If we want to make a new settings class I think we'd need to keep the old eia860m boolean in the 860 settings class. I was thinking the 860 and 860m classes should talk to each other which imo would be messy but they can both go look at the working partitions from sources separately.

src/pudl/package_data/settings/etl_fast.yml
cmgosnell: yea this is a fair question.. One month is definitely not a change log. I think it would take a while to do a full year so maybe like a lot of the other fast settings we feed this two partitions. it'll be kinda a silly lil changelog but i think that is okay for the fast CI

src/pudl/extract/eia860.py
cmgosnell: okay new issue created #3366

src/pudl/package_data/eia860m/column_maps/generator_existing.csv
```diff
@@ -12,8 +12,8 @@ operational_status_code,status,status,status,status,status,status,status,status,
...
...,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year,operating_year
```
cmgosnell, line 16: @aesharpe / @e-belfer  i noticed during my changelog 860m building that the column names in this mapping was resulting in the 860m operating date probably not getting integrated because the column name was inconsistent with the rest of the 860 in the column maps.

src/pudl/package_data/settings/etl_fast.yml
bendnorman: Yeah!  I think a two-month changelog is sufficient for testing out any logic that expects multiple months of data.

src/pudl/helpers.py
cmgosnell: i went through and checked. this function was only being used in a yearly context. and I will add a december case to the unit tests. but it won't be the same behavior as before because a december month in the previous setup would have failed.

src/pudl/transform/eia860m.py
```diff
@@ -0,0 +1,91 @@
...
+import pandas as pd
+from dagster import asset
+
+import pudl
+
+logger = pudl.logging_helpers.get_logger(__name__)
+
+
+@asset(io_manager_key="pudl_io_manager")
+def core_eia860m__monthly_generators_changelog(
```
bendnorman, line 12: This is an interesting naming problem. This table is like our slowly changing dimension tables (`scd`) except it only contains reported records when attributes of an entity actually change and it tracks a `valid_till_date`. This is how [many people model slowly changing dimensions](https://en.wikipedia.org/wiki/Slowly_changing_dimension#Type_2:_add_new_row). Our SCD tables have a record for every time there is a report for an entity even when none of the attributes change.

We could call this asset `core_eia860m__scd_generators` but it wouldn't perfectly align with how we've modeled our other SCD tables. Given we probably aren't going to our current SCD tables' modeling or naming convention maybe we should just stick with "changelog" as the asset type. What do you think @cmgosnell ?

bendnorman (changes requested):
Looks good! Just a few requests and questions.

src/pudl/transform/eia860m.py
```diff
@@ -0,0 +1,91 @@
...
+    eia860m_all = (
+        pudl.transform.eia860._core_eia860__generators(
+            raw_eia860__generator_proposed=raw_eia860m__generator_proposed,
+            raw_eia860__generator_existing=raw_eia860m__generator_existing,
+            raw_eia860__generator_retired=raw_eia860m__generator_retired.assign(
+                operational_status_code=pd.NA
+            ),
+            raw_eia860__generator=pd.DataFrame(columns=list(eia860_columns)),
+        )
+        # drop all the non 86om cols
```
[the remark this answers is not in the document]
cmgosnell: b4by

src/pudl/transform/eia860m.py
```diff
@@ -0,0 +1,91 @@
...
+        ]
+    )
+    eia860m_all = (
+        pudl.transform.eia860._core_eia860__generators(
+            raw_eia860__generator_proposed=raw_eia860m__generator_proposed,
+            raw_eia860__generator_existing=raw_eia860m__generator_existing,
+            raw_eia860__generator_retired=raw_eia860m__generator_retired.assign(
+                operational_status_code=pd.NA
+            ),
+            raw_eia860__generator=pd.DataFrame(columns=list(eia860_columns)),
```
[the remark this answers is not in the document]
cmgosnell: in the pre-860m years, 860 had all of their generators in one big table. but there is no 860m `raw_eia860__generator` equivalent.

this empty df just enables us to use the standard eia860 transform function for generators. it's built with the list of all of the eia columns so that inside `_core_eia860__generators` we aren't missing any columns.

src/pudl/settings.py
```diff
@@ -232,18 +233,41 @@ def check_eia860m_date(cls, eia860m: bool) -> bool:
...
                 f"the eia860 years: {cls.data_source.working_partitions['years']}. "
                 f"Consider switching eia860m parameter to False."
             )
         return eia860m
 
 
+class Eia860mSettings(GenericDatasetSettings):
+    """An immutable pydantic model to validate EIA 860 settings.
+
+    This model also check 860m settings.
```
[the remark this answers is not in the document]
cmgosnell: ha yes this was a copy/paste oopsies

src/pudl/transform/eia860m.py
cmgosnell: yea i wondered this myself. They are somewhat similar but they act very differently in the context  of how the relate to the rest of the db. the `scd` tables have FK relationships to the rest of the db tables while this one definitely does not and should not. I think i would rather keep it clearly different because of this. But I'm open to changing it to `scd`. it definitely is a table of `scd`'s. i'm not sure here.

if we keep do keep `changelog`, it should at least be changed to `core_eia860m__changelog_generators` to be more consistent with the rest.

src/pudl/metadata/resources/eia.py
```diff
@@ -520,7 +520,12 @@
                 "fields": [["plant_id_eia", "generator_id"]],
                 # exclude core_epa__assn_eia_epacamd_subplant_ids bc there are generator ids in this
                 # glue table that come only from epacamd
-                "exclude": ["core_epa__assn_eia_epacamd_subplant_ids"],
+                # also exclude the 860 changelog table bc that table doesn't get harvested
+                # and therefor there are a few straggler generators that don't end up in this table
+                "exclude": [
+                    "core_epa__assn_eia_epacamd_subplant_ids",
+                    "core_eia860m__monthly_generators_changelog",
+                ],
```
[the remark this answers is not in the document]
bendnorman: That all makes sense! I wonder what percentage of generators this table can be linked to out `entity` or `scd` tables. If users want more information about a generator they find in this changelog the *might* be able to link it to our other data.

src/pudl/transform/eia860m.py
bendnorman: I think the fact it doesn't have FKs with the rest of the db tables and has slightly different modeling from our other SCD tables are important distinctions. Given that I say we go with `core_eia860m__changelog_generators`.

src/pudl/metadata/resources/eia.py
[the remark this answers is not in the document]
cmgosnell: there were less than 100 total generators that were not in the harvested tables! so not a gazillion for sure.

bendnorman (approved):

bendnorman:
Hello! What are the remaining pieces of work on this?

cmgosnell:
There are a handful of validation errors. I waited a few days to be able to get to this because there was a pr that got merged into main that I caught that had non-updated validation errors. I wanted to make sure I was trying to debug things from this pr only.. which are these:

```python
FAILED test/validate/eia_test.py::test_minmax_rows[eia_raw-bf_eia923-1607752-1607752-128252] - ValueError: bf_eia923: found 1537748 rows, expected 1607752. Off by -4.354%, allowed margin of 0.000%
FAILED test/validate/eia_test.py::test_minmax_rows[eia_raw-frc_eia923-650671-265784-25443] - ValueError: frc_eia923: found 637127 rows, expected 650671. Off by -2.082%, allowed margin of 0.000%
FAILED test/validate/eia_test.py::test_minmax_rows[eia_raw-gf_eia923-2918170-2918170-246324] - ValueError: gf_eia923: found 2881840 rows, expected 2918170. Off by -1.245%, allowed margin of 0.000%
FAILED test/validate/eia_test.py::test_minmax_rows[eia_monthly-bf_eia923-1607752-1607752-128252] - ValueError: bf_eia923: found 1537748 rows, expected 1607752. Off by -4.354%, allowed margin of 0.000%
FAILED test/validate/eia_test.py::test_minmax_rows[eia_monthly-frc_eia923-650671-265784-25443] - ValueError: frc_eia923: found 259743 rows, expected 265784. Off by -2.273%, allowed margin of 0.000%
FAILED test/validate/eia_test.py::test_minmax_rows[eia_monthly-gf_eia923-2918170-2918170-246324] - ValueError: gf_eia923: found 2881840 rows, expected 2918170. Off by -1.245%, allowed margin of 0.000%
```

i suspect this is from the expand timeseries munging but i'm investigating now

cmgosnell:
- [x] Is this loss of records in the core tables? No. I used the `sqlite-table-diff` and first checked the three core tables implicated and my local tables were all the exact same as the last nightly build tables.
 - [ ] Is this loss of records in the core tables?

cmgosnell:
okay i was jsut being a complete dufous with this validation tests. i just reran them everything is indeed fine.

PR #3332, opened 2024-01-31
Trigger CI on merge group; only trigger integration on merge group.

### LINKED ISSUE #3317: Use merge queue in PUDL repo
1. We require PRs to be up-to-date before merging, so we aren't running CI on some stale combination of `main` and your branch.
2. However, that means a lot of manual updating and re-running of CI.
3. [Merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) lets you put your PR in line. Then GH will run CI on the first PR in line. If it passes, it merges in and then the next PR gets updated to the latest version of the base branch, before then getting CI run on it. If CI fails, the PR is removed from the queue and GH moves on to the next one.

Practically, this means clicking "Merge when ready" when you want to merge something into `main`, instead of "Squash and Merge" or "Enable auto-merge."

Caveats/details:
* if we want this, _**we can't require RTD build before merge**_. We'll still trigger the build on the PR, and we can manually refuse to click "merge when ready" unless RTD succeeds. But RTD's builder doesn't support merge queues, yet.
* we are moving the integration tests to the merge queue - which means that they will no longer run until *after* you press "merge when ready." If they fail, you'll have to go back, fix it, and then re-queue it. But that also means that you don't have to worry about accidentally kicking of $$$ integration tests in your PR.
* there's also the concept of "merge groups" - so if, for example, 2 PRs get enqueued within 5 minutes of each other, we can actually run CI on the two of them combined. If that fails, we split that back up and re-run individually. This can save CI time also! 
* instead of running the CI one run at a time, we can run the CI with high concurrency - which means that if we have A, B, and C waiting in line, GH can run CI on (A + `main`), (B + (A + `main`)), and (C + (B + (A + `main`))) all at once. And if they all pass, all three get merged in order. If we only have concurrency of 1, then these three checks would run in series.

```[tasklist]
## Scope
- [ ] team knows how to work with merge queues: what happens in the normal path? what happens when CI fails?
- [x] current PR checks trigger on `merge_group:` event
- [x] `ci-integration` check gets pushed to `merge_group` only
- [ ] merge queues turned on in repo settings
```

jdangerx:
# Overview

Closes #3317.

This makes our PR checks also run on merge_group triggers, so that our required checks can actually pass when PRs are added to the merge queue.

# Testing

We'll have to turn on the merge queues & test it out manually.

```[tasklist]
# To-do list
```

### DIFF
--- a/.github/workflows/docker-build-test.yml
+++ b/.github/workflows/docker-build-test.yml
@@ -3,6 +3,9 @@ name: docker-build-test
 on:
   workflow_dispatch:
   pull_request:
+  merge_group:
+    types:
+      - checks_requested
 
 jobs:
   pudl_docker_build:
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -3,13 +3,10 @@ name: pytest
 
 on:
   pull_request:
-    types:
-      - opened
-      - synchronize
-      - ready_for_review
   merge_group:
     types:
       - checks_requested
+  workflow_dispatch:
 
 env:
   PUDL_OUTPUT: /home/runner/pudl-work/output/
@@ -110,7 +107,7 @@ jobs:
     runs-on:
       group: large-runner-group
       labels: ubuntu-22.04-4core
-    if: github.event.merge_group
+    if: github.event.merge_group || github.event.workflow_dispatch
     permissions:
       contents: read
       id-token: write
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -4,10 +4,12 @@ name: pytest
 on:
   pull_request:
     types:
-      - created
       - opened
       - synchronize
       - ready_for_review
+  merge_group:
+    types:
+      - checks_requested
 
 env:
   PUDL_OUTPUT: /home/runner/pudl-work/output/
@@ -108,7 +110,7 @@ jobs:
     runs-on:
       group: large-runner-group
       labels: ubuntu-22.04-4core
-    if: github.event.pull_request.draft == false
+    if: github.event.merge_group
     permissions:
       contents: read
       id-token: write


.github/workflows/pytest.yml
```diff
@@ -4,10 +4,12 @@ name: pytest
 on:
   pull_request:
     types:
-      - created
```
jdangerx, line 7: This isn't a valid event type, cleaning this up.

.github/workflows/pytest.yml
```diff
@@ -4,10 +4,12 @@ name: pytest
...
       - opened
       - synchronize
       - ready_for_review
```
jdangerx, line 9: I bet we can actually stop triggering on these specific events, though since everything *but* `ci-integration` is free I think it's probably fine to keep it for now.

zaneselvans: I think we only included `ready_for_review` so that when we switched out of draft mode the integration tests would get run, but since we don't need to do that in the new setup, it seems like it would be nice to skip it, since then you can potentially merge immediately if all of the other requirements have been satisfied instead of needing to wait 5 minutes for the faster CI steps to complete.  It looks like [the default types](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) are `opened`, `synchronize`, and `reopened` which seem right to me so I think removing the `types` section here makes sense.

.github/workflows/docker-build-test.yml
```diff
@@ -3,6 +3,9 @@ name: docker-build-test
 on:
   workflow_dispatch:
   pull_request:
+  merge_group:
+    types:
+      - checks_requested
```
zaneselvans, line 8: Should we also add the `release` check to the `merge_group` checks?

zaneselvans (changes requested):

.github/workflows/docker-build-test.yml
jdangerx: Right now, `release` only runs on `push`, not `pull_request`. I think we *could* run it on `merge_group`, though potentially this will try to release the same version of the code twice on the test PyPI - which might cause spurious failures? I would rather not change more configs now, get the merge queue stuff working with a close mirror of our existing `pull_request:` flows, and then add `release` to the flow as a separate experiment.

zaneselvans: Ah okay, sounds good.

zaneselvans (approved):

Issue #3316, opened 2024-01-29
Release v2024.02.05 checklist


jdangerx:
We'll do this once #3052 is closed.

Full documentation of process [here](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_a_release.html).

```[tasklist]
## Release Checklist
- [x] set a release date & notify team
- [ ] #3342
- [ ] #3344
- [ ] https://github.com/catalyst-cooperative/pudl/pull/3346
- [x] Push v2024.02.05 tag
- [x] Verify that the GitHub (software) release has been published **(multiple failures, see below)**
- [x] Verify RTD build for new version is up to date **(required manual activation of new version)**
- [x] Verify [`catalystcoop.pudl` PyPI (software) release](https://pypi.org/project/catalystcoop.pudl/)
- [x] Verify that [PUDL repo archive on Zenodo](https://zenodo.org/doi/10.5281/zenodo.3404014) has been updated w/ new version
- [x] Wait 6-10 hours for a successful build to complete **(multiple failures, see below)**
- [x] Verify `stable` and version tag point at same git ref **(do manually, see #3347)**
- [x] Verify that RTD docs build for `stable` branch gets updated
- [x] Verify `gs://pudl.catalyst.coop/vYYYY.MM.DD` has the right data.
- [x] Verify `gs://pudl.catalyst.coop/stable` has the right data. **Note: logfile is bigger than in `v2024.02.05`**
- [x] Verify `s3://pudl.catalyst.coop/vYYYY.MM.DD` has the right data.
- [x] Verify `s3://pudl.catalyst.coop/stable` has the right data.
- [x] Verify Zenodo draft deposition has expected data
- [x] Update Zenodo description and other metadata based on release notes / details
- [x] Manually publish the new Zenodo deposition
- [x] Update dependencies and merge the conda-forge feedstock PR: https://github.com/conda-forge/catalystcoop.pudl-feedstock/pull/18
- [x] Create a GitHub Announcement #3351 (should automate with `gh release create` command)
- [x] Update [release documentation](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_a_release.html) to better reflect the actual process
- [x] Create an issue template for doing a release with a task list
```

```[tasklist]
# Issues to Address
[...]

PR #3347, opened 2024-02-04
Standardize process for merging tagged commits into persistent branches automatically

zaneselvans:
# Overview

One of the many issues that came up in #3316 was that the `release.yml` workflow was unable to merge the tagged release into the `stable` branch.

This PR moves the branch merging logic into the nightly build script (where it really should be... so it can depend on whether the ETL succeeded!) and re-uses the logic that we ultimately came up with for merging the nightly tags into the nightly branch.

It also eliminates an accidental dependence on the 3rd party `regex` package, which really should have been just the built-in python `re` package.

```[tasklist]
# To-do list
```


jdangerx:
I agree that we want to wait until *after* the build finishes to update the `stable` branch - maybe it makes sense to use the same logic as `update_nightly_branch` in `gcp_pudl_etl.sh` as opposed to doing this in `release.yml`. Then, also, the logic gets tested every nightly build instead of every release cycle. What do you think @zaneselvans ?

zaneselvans:
@jdangerx Yeah, I think that's the right place for it to happen, and we've already got the logic over there.

jdangerx:
I can take a crack at that if you want, or I'm happy to review your crack at it.

zaneselvans:
I've had my fill of this particular crack!

src/pudl/extract/excel.py
```diff
@@ -1,12 +1,12 @@
 """Load excel metadata CSV files form a python data package."""
 import importlib.resources
 import pathlib
+import re
 from collections import defaultdict
 from io import BytesIO
 
 import dbfread
 import pandas as pd
-import regex as re
```
zaneselvans, line 9: This seems to have been an accident. The built-in `re` module does exactly the same thing.

pyproject.toml
```diff
@@ -66,7 +66,6 @@ dependencies = [
     "pyxlsb>=1.0.9", # pandas[excel]
     "pyyaml>=6",
     "readthedocs-sphinx-ext>=2",
-    "regex>=2023.12.25",
```
zaneselvans, line 69: Removing this since we only accidentally ended up depending on it directly.

docker/gcp_pudl_etl.sh
```diff
@@ -186,20 +186,22 @@ function notify_slack() {
     upload_file_to_slack "$LOGFILE" "$BUILD_ID logs:"
 }
 
-function update_nightly_branch() {
+function merge_tag_into_branch() {
```
zaneselvans, line 189: I think we're doing exactly the same thing in these two cases, just with a different tag and a different persistent branch that it needs to be merged into, so having a single function seems ideal. Also it'll get tested in the nightly builds rather than only when we run a release which is nice.

.github/workflows/release.yml
```diff
@@ -94,13 +95,6 @@ jobs:
           --title 'PUDL ${{ github.ref_name }}'
           --generate-notes
           --repo '${{ github.repository }}'
-      - name: Tag the current stable commit
-        run: |
-          git config user.email "pudl@catalyst.coop"
-          git config user.name "pudlbot"
-          git checkout stable
-          git merge --ff-only ${{ github.ref_name }}
-          git push
```
zaneselvans, line 103: All this logic now in the build script, where it can be conditional on the ETL actually succeeding.

docker/gcp_pudl_etl.sh
```diff
@@ -243,9 +245,13 @@ SAVE_OUTPUTS_SUCCESS=${PIPESTATUS[0]}
...
 if [[ $ETL_SUCCESS == 0 ]]; then
     if [[ "$GITHUB_ACTION_TRIGGER" == "schedule" ]]; then
-        update_nightly_branch 2>&1 | tee -a "$LOGFILE"
+        merge_tag_into_branch "$NIGHTLY_TAG" nightly 2>&1 | tee -a "$LOGFILE"
         UPDATE_NIGHTLY_SUCCESS=${PIPESTATUS[0]}
     fi
+    # If running a tagged release, merge the tag into the stable branch
+    if [[ "$GITHUB_ACTION_TRIGGER" == "push" && "$BUILD_REF" == v20* ]]; then
+        merge_tag_into_branch "$BUILD_REF" stable 2>&1 | tee -a "$LOGFILE"
+    fi
```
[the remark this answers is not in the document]
zaneselvans: Ah yep. Added it everywhere there was an `UPDATE_NIGHTLY_SUCCESS`

bendnorman (approved):

PR #3352, opened 2024-02-05
Add a simple test coverage check.

### LINKED ISSUE #3337: Capture and check test coverage in the merge queue
Now that we are using the merge queue in this repo #3317 it seems like we are no longer getting any feedback about our test coverage, since that can only be generated when the integration tests run, and the test coverage is not currently a required check.

```[tasklist]
- [x] Add `ci-coverage` to required checks to ensure we continue generating test coverage and uploading it to CodeCov.
- [x] Ensure that test coverage associated with `main` is updated. [Seems to be working](https://app.codecov.io/gh/catalyst-cooperative/pudl)
- [x] Add a simple fixed threshold for use in checking whether test coverage is good enough
- [x] Update docs & PR template to direct folks to run `make pytest-coverage` locally before submitting to the merge queue
```

zaneselvans:
# Overview

Add a simple test coverage check which can prevent PRs that reduce test coverage from getting merged into `main` in the merge queue.

Closes #3337 

# Testing

```[tasklist]
# To-do list
- [x] Try running `ci-integration` with a `workflow_dispatch` trigger and get it to run the tests + coverage.
- [ ] Try submitting to merge queue with 100% coverage requirement and get it rejected.
- [ ] Reset target coverage threshold to... 92.8% (currently [CodeCov says we're at 92.95%](https://app.codecov.io/gh/catalyst-cooperative/pudl))
```


.github/workflows/pytest.yml
```diff
@@ -97,11 +97,11 @@ jobs:
...
 
-      - name: Upload coverage
+      - name: Upload unit tests coverage artifact
         uses: actions/upload-artifact@v4
         with:
           name: coverage-unit
-          path: coverage.xml
+          path: ./*coverage*
 
   ci-integration:
```
zaneselvans, line 106: The integration tests are still skipped even with a manual `workflow_dispatch`, so it seems like this conditional isn't working correctly.

.github/workflows/pytest.yml
```diff
@@ -50,11 +50,11 @@ jobs:
           pip install --no-deps --editable .
           make docs-build
 
-      - name: Upload coverage
+      - name: Upload docs coverage artifact
         uses: actions/upload-artifact@v4
         with:
           name: coverage-docs
-          path: coverage.xml
+          path: ./*coverage*
```
zaneselvans, line 57: Upload both the `coverage.xml` report and the `.coverage` outputs so they're both available to the different coverage actions later.

.github/workflows/pytest.yml
```diff
@@ -198,11 +198,30 @@ jobs:
           path: coverage
       - name: List downloaded files
         run: |
-          ls -R
+          find coverage -type f
```
zaneselvans, line 201: More focused output -- just the downloaded files.

.github/workflows/pytest.yml
```diff
@@ -198,11 +198,30 @@ jobs:
...
+        run: |
+          micromamba run -n coverage coverage combine coverage/*/.coverage
+          micromamba run -n coverage coverage xml
+      - name: Check coverage
+        uses: orgoro/coverage@v3.1
+        with:
+          coverageFile: coverage.xml
+          token: ${{ secrets.GITHUB_TOKEN }}
+          # Set impossibly high to test failure condition in merge queue
+          thresholdAll: 1.0
```
zaneselvans, line 224: Install coverage, use it to combine the `.coverage` artifacts and convert them into a unified `coverage.xml` and then use the `orgoro/coverage` action to check whether we met the threshold value.

pudlbot (dismissed):

zaneselvans:
I'm going to try adding this to the merge queue to see what happens / what the real full reported coverage is. But it should fail because I set the threshold to 100%.

zaneselvans:
Booooo, [the action doesn't work](https://github.com/catalyst-cooperative/pudl/actions/runs/7779627007/job/21212786841) with the `merge_group` trigger.

jdangerx:
@zaneselvans should we close this and make a new PR in the future when we're ready to take a different coverage approach?

zaneselvans:
@jdangerx Probably this is not the way to do it, so yeah. But we need to create a new issue for getting at least a "do you meet the threshold" check implemented, and I think it's pretty high priority. We will lose coverage otherwise!

zaneselvans:
Actually on second thought I think there are some changes here that we should incorporate here, uploading / downloading all the coverage outputs for future use.   It was finnicky to get working and I'd rather not have to do it again.

.github/workflows/pytest.yml
```diff
@@ -198,11 +198,24 @@ jobs:
...
+          init-shell: bash
+          environment-name: coverage
+          create-args: >-
+            python=3.11
+            coverage>=7.4.1
+      - name: Combine coverage data and check that we have required coverage
+        run: |
+          micromamba run -n coverage coverage combine coverage/*/.coverage
+          micromamba run -n coverage coverage xml
+          micromamba run -n coverage coverage report --fail-under=100
```
zaneselvans, line 218: This should implement the simple fixed threshold check.  I want to test it without submitting to the merge queue, but the `workflow_dispatch` trigger isn't working.  Is there something obvious wrong with the conditional @jdangerx? I'm having trouble finding examples to compare against.

.github/workflows/pytest.yml
```diff
@@ -97,17 +97,17 @@ jobs:
...
           name: coverage-unit
-          path: coverage.xml
+          path: ./*coverage*
 
   ci-integration:
     runs-on:
       group: large-runner-group
       labels: ubuntu-22.04-4core
-    if: github.event.merge_group || github.event.workflow_dispatch
+    if: github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'merge_queue'))
```
zaneselvans, line 110: @jdangerx this seems to work but also feels super janky.

zaneselvans:
Would like to try submitting this to the merge queue and see if it fails as expected due to the 100% coverage requirement.

pudlbot (approved):

zaneselvans:
Well that didn't work as expected.

PR #3357, opened 2024-02-05
TEST: Remove integration tests - so codecov *should* fail on merge

jdangerx:
# Overview

This is related to #3337, but does not close it.

It looks like coverage is being properly reported to Codecov, so I'm running a test to see if removing all the integration tests will fail our test coverage:

1. Remove all integration tests, making our coverage plummet to ~30%
2. Shove this into the merge queue
3. See if codecov fails at the "coverage went down" step (good!) or at the backstop "force failure always" step (bad!)

@zaneselvans - obviously you would never really approve a PR that removed all of our integration tests unceremoniously. But. This is for science and I don't expect it to actually succeed in the merge queue. If it slips through somehow we can revert it.

# Testing

I made a temporary auto-fail step which would run in the PR and not just in the merge queue. It failed: https://github.com/catalyst-cooperative/pudl/actions/runs/7788331033/job/21237515254?pr=3357

```[tasklist]
# To-do list
- [x] verify that the auto-failing block does indeed auto-fail
```


zaneselvans (approved):
Just to test my intuition, my guess is that it won't fail in the `codecov` action because all that action does is upload the coverage report to the external service -- I don't think it checks anything related to the contents of the report. And the merge queue won't wait to hear back from the external CodeCov CI service, so we'll get to the forced failure.

jdangerx:
Sigh. Fine, I'll add a test.

jdangerx:
> Just to test my intuition, my guess is that it won't fail in the `codecov` action because all that action does is upload the coverage report to the external service -- I don't think it checks anything related to the contents of the report. And the merge queue won't wait to hear back from the external CodeCov CI service, so we'll get to the forced failure.

Yes, I fear that's what will happen too. Though I do sometimes see a red X on the Github Actions check status for the codecov action when the coverage goes down, which gives me hope that we will fail *before* the forced failure.

zaneselvans:
My guess is that unless CodeCov does some work to integrate with merge queues (which would be great) we'll need a super simple "did overall coverage meet our hard-coded threshold value" check that we can do via bash in the workflow.

zaneselvans:
I think we're in decent shape with the test coverage now. It's just a fixed threshold for now but all the mechanics seem to be working, so I'm going to close this.

Issue #3364, opened 2024-02-06
Nightly Build Failure 2024-02-06


zaneselvans:
# Overview

It looks like an error in pulling data from GCS, caught by Google via a checksum mismatch.

# Next steps

What next steps do we need to do to understand or remediate the issue?

# Verify that everything is fixed!

Once you've applied any necessary fixes, make sure that the nightly build outputs are all in their right places.

```[tasklist]
- [ ] [S3 distribution bucket](https://s3.console.aws.amazon.com/s3/buckets/pudl.catalyst.coop?region=us-west-2&bucketType=general&prefix=nightly/&showversions=false) was updated at the expected time
- [ ] [GCP distribution bucket](https://console.cloud.google.com/storage/browser/pudl.catalyst.coop/nightly;tab=objects?project=catalyst-cooperative-pudl) was updated at the expected time
- [ ] [GCP internal bucket](https://console.cloud.google.com/storage/browser/builds.catalyst.coop) was updated at the expected time
- [ ] [Datasette PUDL version](https://data.catalyst.coop/pudl/core_pudl__codes_datasources) points at the same hash as [nightly](https://github.com/catalyst-cooperative/pudl/tree/nightly)
- [ ] [Zenodo sandbox record](https://sandbox.zenodo.org/doi/10.5072/zenodo.5563) was updated to the record number in the logs (search for `zenodo_data_release.py` and `Draft` in the logs, to see what the new record number should be!)
```

# Relevant logs
[link to build logs from internal distribution bucket]( PLEASE FIND THE ACTUAL LINK AND FILL IN HERE )

```
2024-02-06 07:01:42 +0000 - dagster - ERROR - etl_job - eb4af868-6b94-4a47-b66a-13810f197ec1 - 1104 - RUN_FAILURE - Execution of run for "etl_job" failed. Steps failed: ['raw_eia923__all_dfs.extract_single_eia923_year[2003]'].
Traceback (most recent call last):
  File "/home/mambauser/env/bin/pudl_etl", line 8, in <module>
    sys.exit(pudl_etl())
             ^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
[...]

PR #3369, opened 2024-02-07
Use context managers for opening zipfiles

bendnorman:
# Overview

Hopefully, using context managers to close connections to zip files will resolve our sporadic `BadZipFile` error #3364.  

# Testing

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/convert/censusdp1tract_to_sqlite.py
+++ b/src/pudl/convert/censusdp1tract_to_sqlite.py
@@ -66,31 +66,31 @@ def raw_censusdp1tract__all_tables(context):
         # Use datastore to grab the Census DP1 zipfile
         tmpdir_path = Path(tmpdir)
         assert tmpdir_path.is_dir()
-        zip_ref = ds.get_zipfile_resource(
+        with ds.get_zipfile_resource(
             "censusdp1tract", year=context.op_config["year"]
-        )
-        extract_root = tmpdir_path / Path(zip_ref.filelist[0].filename)
-        out_dir = PudlPaths().output_dir
-        assert out_dir.is_dir()
-        out_path = PudlPaths().output_dir / "censusdp1tract.sqlite"
+        ) as zip_ref:
+            extract_root = tmpdir_path / Path(zip_ref.filelist[0].filename)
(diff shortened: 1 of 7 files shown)


jdangerx (changes requested):
Sweet, glad to see it's so straightforward :) There are a few other places where we use `get_zipfile_resource` outside of a context manager (`dbf.py`, `eia176.py`, `epacems.py`, `etl_test.py`, `glue_assets.py`) - any particular reason not to change those as well?

bendnorman:
Ugh I goofed this rebase. @jdangerx any idea how to resolve it?

bendnorman:
I went back and added context managers to where we use `get_zipfile_resource` except for `pudl.extract.dbf.FercDbfReader`. https://github.com/catalyst-cooperative/pudl/blob/3630dcaa7c4b667128903f6320d9cc2b905869eb/src/pudl/extract/dbf.py#L340-L350

It doesn't seem right to pass an open `ZipFile` object into an objection initialization because it's not super clear when it should be closed. I think it'd be better to pass a zipfile path or the arguments of `get_zipfile_resource` into this class so it's methods can open and close the file as needed. Given we haven't seen this bit of code produce the `BadZipError` I think it's fine to keep it as is for now. I can create an issue to redesign the classes that accept open zipfiles.

I'm also more optimistic adding context managers will resolve the `BadZipFile("File is not a zip file")` error. The `zipfile` docs say:

> You must call [close()](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.close) before exiting your program or essential records will not be written.

This makes me think the "essential records" weren't written and the PHMSA and FERC 714 extractors tried to reopen these corrupted zipfiles. However, I don't know what "essential records" would need to be written when you are reading the file 🤷

jdangerx:
Leaving the `FercDbfReader` as is seems fine! Maybe a comment in-line to leave a breadcrumb trail would be helpful for future Ben :)

Your comment about essential records led me down this rabbit hole: https://github.com/python/cpython/blob/3.12/Lib/zipfile/__init__.py#L1211

jdangerx (approved):
🤐

PR #3389, opened 2024-02-12
Extract raw 923 Schedule 8 deprecated tabs (2008-2011)

### LINKED ISSUE #3388: Map EIA 923 Schedule 8 older deprecated tabs (e.g. Boiler NOX operations)
As part of raw data integration of EIA 923 Schedule 8 data, we should bring in the various tables and parts of the form that have been deprecated over time. This includes:

- [x] Schedule 8C Nitrogen Oxide Emission Controls (2008-11)
- [x] Schedule 8E Flue Gas Particulate Collection Information (2008-11)

These tables have columns that get absorbed into the `emissions_control` tab in 2012, but we should bring them in separately and join them during the transformation process.

e-belfer:
# Overview

Closes #3388. Maps a few tabs from 2008-2011 that are integrated into the `emissions_control` tab after 2012.

Brings in the remaining raw data from EIA 923 Schedule 8.

What did you change?
Mapped the spreadsheets for two 923 Excel tabs.

# Testing

Generate `raw_eia923__boiler_nox_operation` and `raw_eia923__fgp_operation` in dagster.

```[tasklist]
# To-do list
```


aesharpe (commented):
The issue attached to this PR says that these two tables were combined in 2012 to create the emissions control table but should be integrated separately. Have we tried to see whether we could add this information into the emissions control table and just have one instead of three tables? 

Even if the tables as a whole aren't comparable enough to be combined pre and post 2012, we should make sure that any individual columns that have the same value pre and post 2012 have the same name.

e-belfer:
They are comparable and should be combined in the transform step. If it were just one table that had changed, I would certainly advocate for us to just join them in the raw extraction step. However we don't currently have a way to extract and join two tables, and it is arguably moving past the "raw" stage of the data to add a merge step into the extraction. So I suggest we extract both tables as they are, and then do the merging of the pre-2012 and post-2012 data into one table during the transform, as is consistent with our other tables.

aesharpe (approved):

Issue #3381, opened 2024-02-09
Nightly Build Failure 2024-02-09


jdangerx:
# Overview

What do you think happened, if anything?

Our coverage failure threshold is a little tetchy because it fails on that threshold even when we aren't running *all* the tests.

# Next steps

What next steps do we need to do to understand or remediate the issue?

* figure out why pytest is even running with coverage in `gcp_pudl_etl.sh` in the first place
* either stop pytest from running with coverage, or make the coverage cover docs/unit/integration instead of just unit/integration.

# Verify that everything is fixed!

Once you've applied any necessary fixes, make sure that the nightly build outputs are all in their right places.

```[tasklist]
- [ ] [S3 distribution bucket](https://s3.console.aws.amazon.com/s3/buckets/pudl.catalyst.coop?region=us-west-2&bucketType=general&prefix=nightly/&showversions=false) was updated at the expected time
- [ ] [GCP distribution bucket](https://console.cloud.google.com/storage/browser/pudl.catalyst.coop/nightly;tab=objects?project=catalyst-cooperative-pudl) was updated at the expected time
- [ ] [GCP internal bucket](https://console.cloud.google.com/storage/browser/builds.catalyst.coop) was updated at the expected time
- [ ] [Datasette PUDL version](https://data.catalyst.coop/pudl/core_pudl__codes_datasources) points at the same hash as [nightly](https://github.com/catalyst-cooperative/pudl/tree/nightly)
- [ ] [Zenodo sandbox record](https://sandbox.zenodo.org/doi/10.5072/zenodo.5563) was updated to the record number in the logs (search for `zenodo_data_release.py` and `Draft` in the logs, to see what the new record number should be!)
```

# Relevant logs
[link to build logs from internal distribution bucket](https://console.cloud.google.com/storage/browser/_details/builds.catalyst.coop/2024-02-09-0602-227119442-main/2024-02-09-0602-227119442-main-pudl-etl.log)

Only one test session started - so it only ran the unit/integration tests. And we don't run docs build on nightly.

```
FAIL Required test coverage of 92.7% not reached. Total coverage: 68.97%
= 1269 passed, 5 skipped, 15 xfailed, 2 xpassed, 22 warnings in 2994.44s (0:49:54) =```

PR #3391, opened 2024-02-12
Simplify pytest-cov configuration.

jdangerx:
# Overview

I was investigating #3381 when I ran into a slight weirdness in our `pytest-cov` configuration in `pyproject.toml`.

It sort of seems like this would fix it? @zaneselvans I know you were the one who was messing with this in the first place, did you already try this and have it not work?

See docs for v4.1.0 at
https://github.com/pytest-dev/pytest-cov/blob/2c9f2170d8575b21bafb6402eb30ca7de31e20b9/docs/config.rst
- particularly:

> If you use the --cov=something option (with a value) then coverage's
> source option will also get overridden. If you have multiple sources it
> might be easier to set those in .coveragerc and always use --cov
> (without a value) instead of having a long command line with --cov=pkg1
> --cov=pkg2 --cov=pkg3 ....

# Testing

I removed `.coverage`, then ran `make pytest-coverage` and saw that the coverage denominator was still 13450 statements.

```[tasklist]
# To-do list
```


pyproject.toml
```diff
@@ -301,11 +301,7 @@ filterwarnings = [
 ]
 
 markers = ["slow: marks tests as slow (deselect with '-m \"not slow\"')"]
-# Oddly, despite the use of --cov-config=pyproject.toml here, pytest does not seem to
-# pick up the source directories specified in the [tool.coverage.run] section below.
-# (though it *does* pick up the omit parameters!). This means we need to specify the
-# source directories we want to collect coverage twice, and keep the two in sync.
-addopts = "--verbose --pdbcls=IPython.terminal.debugger:TerminalPdb --cov-append --cov-config=pyproject.toml --cov-report=xml --cov=src/pudl --cov=test/integration --cov=test/unit"
+addopts = "--verbose --pdbcls=IPython.terminal.debugger:TerminalPdb --cov-append --cov-config=pyproject.toml --cov-report=xml --cov -n auto"
```
zaneselvans, line 304: I don't remember exactly what was happening here, but it seemed to have to do with coverage behaving differently when run directly, vs. run through pytest.  But maybe I just didn't realize that a blanket `--cov` would default to whatever `sources` were listed in the configuration file. Seems likely.

We also need to deal with `-n auto` crashing your machine when you try to run the validation tests since we end up with potentially several copies of our largest dataframes getting read into memory simultaneously which is too much. But this problem isn't limited to this line.

Did you look at the `Makefile`, `.pre-commit-config.yaml` and `docker/gcp_pudl_etl.sh` to see if the pytest / coverage commands being used in there need to be updated to work with these `addopts` and/or avoid duplicative specifications of the same arguments in multiple places?

zaneselvans (changes requested):
I think you're probably right about swapping in a `--cov` here.

There are a couple of other places where we're calling `pytest` that we might want to check to see if we're repeatedly specifying the same args unnecessarily or in conflict: the nightly build script, the Makefile, and the pre-commit configuration. (the pytest workflow uses make)

jdangerx:
@zaneselvans - went through the places you mentioned (`Makefile`, `gcp_pudl_etl.sh`, and pre-commit) and found that we were respecifying the gcs cache path everywhere. So I pulled that into the pyproject.toml.

Other than that, I removed the `-n auto` because of the memory issue, and everything else seems to be fine / non-duplicative.

zaneselvans (approved):

zaneselvans:
If there's a safe number of CPUs we could use by default (like 4?) it would nice to keep some amount of parallelism. It does speed things up.  Maybe we can turn it back on after we kill `PudlTabl` and get all the validations into the ETL / Pandera setup.

PR #3399, opened 2024-02-13
Update nightly build script to distribute parquet

### LINKED ISSUE #3362: Update experimental parquet outputs in nightly builds
Once we are successfully generating parquet outputs as part of the nightly builds (See #3296), start deploying & updating those outputs as part of the nightly or release builds in a "private beta" (non-public location) that we can share with some users gather feedback and usage data.

```[tasklist]
- [ ] Deploy nightly outputs to `gs://parquet.catalyst.coop/nightly/`
- [ ] Deploy tagged version outputs to `gs://parquet.catalyst.coop/vYYYY.MM.DD/`
- [ ] Tagged versions should also update `gs://parquet.catalyst.coop/stable/`
```

zaneselvans:
# Overview

Distribute parquet outputs to a private bucket using the same path conventions that we're using in the public buckets for the other outputs.

Closes #3362 

# Testing

I wish there were an easy way to test this besides running the nightly builds and having them fail repeatedly.

```[tasklist]
# To-do list
```

### DIFF
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -126,7 +126,7 @@ function distribute_parquet() {
 function copy_outputs_to_distribution_bucket() {
     # Only attempt to update outputs if we have a real value of BUILD_REF
     # This avoids accidentally blowing away the whole bucket if it's not set.
-    echo "Copying Parquet output to private distribution bucket"
+    echo "Copying outputs to distribution buckets"
     if [[ -n "$BUILD_REF" ]]; then
         if [[ "$GITHUB_ACTION_TRIGGER" == "schedule" ]]; then
             # If running nightly builds, copy outputs to the "nightly" bucket path
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -79,23 +79,54 @@ function upload_to_dist_path() {
     GCS_PATH="gs://pudl.catalyst.coop/$1/"
     AWS_PATH="s3://pudl.catalyst.coop/$1/"
 
-    # If the old outputs don't exist, these will exit with status 1, so we
-    # don't && them with the rest of the commands.
-    echo "Removing old outputs from $GCS_PATH."
-    gsutil -m -u "$GCP_BILLING_PROJECT" rm -r "$GCS_PATH"
-    echo "Removing old outputs from $AWS_PATH."
-    aws s3 rm "$AWS_PATH" --recursive
+    # Only attempt to update outputs if we have an argument
+    # This avoids accidentally blowing away the whole bucket if it's not set.
+    if [[ -n "$1" ]]; then
+        # If the old outputs don't exist, these will exit with status 1, so we
+        # don't && them with the rest of the commands.
+        echo "Removing old outputs from $GCS_PATH."
+        gsutil -m -u "$GCP_BILLING_PROJECT" rm -r "$GCS_PATH"
+        echo "Removing old outputs from $AWS_PATH."
+        aws s3 rm --recursive "$AWS_PATH"
 
-    echo "Copying outputs to $GCS_PATH:" && \
-    gsutil -m -u "$GCP_BILLING_PROJECT" cp -r "$PUDL_OUTPUT/*" "$GCS_PATH" && \
-    echo "Copying outputs to $AWS_PATH" && \
-    aws s3 cp "$PUDL_OUTPUT/" "$AWS_PATH" --recursive
+        echo "Copying outputs to $GCS_PATH:" && \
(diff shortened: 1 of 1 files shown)


docker/gcp_pudl_etl.sh
```diff
@@ -79,23 +79,54 @@ function upload_to_dist_path() {
...
-    echo "Removing old outputs from $AWS_PATH."
-    aws s3 rm "$AWS_PATH" --recursive
-
-    echo "Copying outputs to $GCS_PATH:" && \
-    gsutil -m -u "$GCP_BILLING_PROJECT" cp -r "$PUDL_OUTPUT/*" "$GCS_PATH" && \
-    echo "Copying outputs to $AWS_PATH" && \
-    aws s3 cp "$PUDL_OUTPUT/" "$AWS_PATH" --recursive
+    # Only attempt to update outputs if we have an argument
+    # This avoids accidentally blowing away the whole bucket if it's not set.
+    if [[ -n "$1" ]]; then
```
zaneselvans, line 84: I think if we called this function without an argument it would delete all of our previous releases which is bad, so I added this small margin of safety.  Maybe we should automatically make the versioned releases read-only as soon as we create them so this can't happen.

`stable` should always point at a versioned release, so replacing it if need be wouldn't be hard.  `nightly` is ephemeral so not a disaster if it accidentally gets wiped out. And the build outputs for recent `nightly` builds are still laying around so we could repopulate it manually.

docker/gcp_pudl_etl.sh
```diff
@@ -225,6 +257,9 @@ if [[ $ETL_SUCCESS == 0 ]]; then
     # should be moved to the triggering github action. Having it here feels fragmented.
     # Distribute outputs if branch is main or the build was triggered by tag push
     if [[ "$GITHUB_ACTION_TRIGGER" == "push" || "$BUILD_REF" == "main" ]]; then
+        # Distribute Parquet outputs to a private bucket
+        distribute_parquet 2>&1 | tee -a "$LOGFILE"
+        DISTRIBUTE_PARQUET=${PIPESTATUS[0]}
```
zaneselvans, line 262: This has to happen before we cleanup the outputs, because we remove all the `parquet/*.parquet` files there.

docker/gcp_pudl_etl.sh
```diff
@@ -79,23 +79,54 @@ function upload_to_dist_path() {
...
+        echo "Copying outputs to $PARQUET_BUCKET/$DIST_PATH" && \
+        gsutil -m -u "$GCP_BILLING_PROJECT" cp -r "$PUDL_OUTPUT/parquet/*" "$PARQUET_BUCKET/$DIST_PATH"
+
+        # If running a tagged release, ALSO update the stable distribution bucket path:
+        if [[ "$GITHUB_ACTION_TRIGGER" == "push" && "$BUILD_REF" == v20* ]]; then
+            echo "Copying outputs to $PARQUET_BUCKET/stable" && \
+            gsutil -m -u "$GCP_BILLING_PROJECT" cp -r "$PUDL_OUTPUT/parquet/*" "$PARQUET_BUCKET/stable"
+        fi
+    fi
 }
```
zaneselvans, line 124: First I tried moving the wholesale removal of the `parquet/` directory from the output cleanup into the deployment function, but this didn't work because of the need to deploy twice when we have a versioned release (once to the versioned path, and once to `stable`)

Then I tried more selectively deploying outputs in the existing deployment function, instead of just copying everything in the directory, but this didn't work because [the S3 CLI (incredibly) can't understand wildcards](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3/index.html#use-of-exclude-and-include-filters) and there was no simple way to specify the set of parquet vs. non-parquet files for separate distribution.

Finally I made this terrible cut-and-paste copy of the deployment function which gets called before the output cleanup.

Is there a simple better way? Short of re-writing the entire thing in Python and bailing on these CLIs?

bendnorman (approved):
Looks good! I gave the `deploy-pudl-vm-service-account` the Storage Admin role on the `gs://parquet.catalyst.coop` bucket so it can create and delete files.

docker/gcp_pudl_etl.sh
zaneselvans: Great. Hopefully this is easy?  I'd like to add this to the process / script for our next release.

docker/gcp_pudl_etl.sh
zaneselvans: I was referring to `copy_outputs_to_distribution_bucket()` which calls `upload_to_dist_path()` twice with different arguments if we're doing a versioned (aka `stable`) release.

Issue #3366, opened 2024-02-06
Generalize `pudl.extract.raw_df_factory` to extract any partition (not just `year`)


cmgosnell:
right now `pudl.extract.raw_df_factory` expects the partition of the extractor to be `year` which is hard coded in a few places.

See [comment](https://github.com/catalyst-cooperative/pudl/pull/3331#discussion_r1478958501) which spurred this issue.

```[tasklist]
- [ ] generalize `year_extractor_factory` -> `partition_extractor_factory`
- [ ] generalize `extract_single_year` -> `extract_single_partition`
- [ ] generalize `years_from_settings_factory` -> `partitions_from_settings_factory`
- [ ] integrate partition changes into `raw_df_factory` & `raw_df_factory.raw_dfs`
- [ ] test new setup with existing year-based use of excel extractor (eia860, eia923, phmsagas.... (i think that's all?)
- [ ] convert `pudl.extract.eia860m.raw_eia860m__all_dfs` to use `raw_df_factory`
- [ ] ...(i'm sure i missed something)?
```

NateWasTaken:
Hello can I take this assignment?

NateWasTaken:
Sorry - forgot to tag. @catalyst-cooperative/com-dev Can I take this assignment

NateWasTaken:
> Hi @NateWasTaken! Thanks for the message. This issue has actually been addressed in #3402 and should be closed, my mistake! I'll take a look through our existing issues and see what other candidates for good first issues I can find.

That would be great thank you e-belfer. My partner and I are working on a class project and hoping to contribute in any way possible!

PR #3402, opened 2024-02-16
Get multiple years of EIA 176/191/757A CSV data

### LINKED ISSUE #3339: Update `CsvExtractor` to handle new 176/191/757A data format
EIA 176 data was originally getting pulled from the bundled data, after investigation of the bulk API data proved it was not suitable. In https://github.com/catalyst-cooperative/pudl/issues/2603#issuecomment-1908880586 we discussed updating the data source to query NGQV directly. This has now been implemented in https://github.com/catalyst-cooperative/pudl-archiver/pull/267, producing 3 new archives with 1 zipped CSV per year ([176](https://zenodo.org/records/10607843), [191](https://zenodo.org/records/10607837), [757A](https://zenodo.org/records/10607839) archives). The advantage of this format is that we:
- don't have to hand-map line numbers to variables
- can easily extract a subset of years, as we do with other EIA data

However, the `CsvExtractor` written by @davidmudrauskas will have to be slightly updated to handle this new data format. The extractor should be able to take a subset of years from the ETL settings, pass it through `dataset_settings` to the extractor, and extract and concatenate the relevant zipped CSVs into one dagster asset. There should be three extractors with essentially identical implementations of the archiver - 176, 191 and 757A.

davidmudrauskas:
# Overview

Closes #3339.

This updates and implements `CsvExtractor` to extract data for EIA forms 176, 191 and 757A. The extractors can take a subset of years from the ETL settings.

As part of this we abstracted the `ExcelExtractor` and `CsvExtractor`, along with corresponding metadata objects into `GenericExtractor` and `GenericMetadata`. 

# Testing

See unit tests, results from local Dagster executions.

```[tasklist]
# To-do list
- [x] For significant ETL changes, ensure the full ETL runs locally
```


src/pudl/extract/extractor.py
```diff
@@ -0,0 +1,151 @@
...
+    DynamicOut,
+    DynamicOutput,
+    OpDefinition,
+    graph_asset,
+    op,
+)
+
+
+# TODO: Consider abstract class
+class GenericExtractor:
```
davidmudrauskas, line 16: I created this module to capture a number of top-level functions (below) that were no longer scoped to just Excel extractors, given how we want to pass multiple years to extract CSV data for. To get the inheritance hierarchy and typing represented so far to work, I created this stub base class and plan to look into the proper way to make this abstract, or Python's equivalent of an interface.

src/pudl/extract/csv.py
```diff
@@ -1,72 +1,121 @@
...
+        """
+        raw_dfs = {}
+        logger.info(f"Extracting {self._dataset_name} data.")
+
+        page = "data"
+        dfs = [
+            pd.DataFrame(),
+        ]
+        # TODO: Use the below commented-out code for consistency? I'm only expecting individual years at this point
+        # for partition in pudl.helpers.iterate_multivalue_dict(**partition):
```
davidmudrauskas, line 46: This method is still messy and I will definitely clean it up. Here and below there are multiple blocks of code from the Excel extractor that could be worth conforming to, depending on how defensive/sophisicated/conformist we want to be. It could largely get reused/"contracted" in a DRY manner if we decompose/refactor things. Any thoughts on how much to put into making methods/"contracts" conform across extractors of different types?

src/pudl/extract/csv.py
```diff
@@ -1,72 +1,121 @@
...
 
-    The extraction logic is invoked by calling extract() method of this class.
-    """
+    def extract(self, **partition):
+        """Extracts dataframes.
 
-    def __init__(self, zipfile: ZipFile, table_file_map: dict[str, str]):
-        """Create a new instance of CsvExtractor.
+        Returns dict where keys are page names and values are
+        DataFrames containing data across given years.
```
davidmudrauskas, line 32: I need to double check this. We'll need to add a `year` column at some point to distinguish the source data, which is time-variant, and it isn't clear to me if/where that is happening.

src/pudl/extract/csv.py
```diff
@@ -1,72 +1,121 @@
...
-        return list(self._table_file_map)
-
-    def extract_one(self, table_name: str) -> pd.DataFrame:
-        """Read the data from the CSV source file and return as a dataframe."""
-        logger.info(f"Extracting {table_name} from CSV into pandas DataFrame.")
-        filename = self._table_file_map[table_name]
-        with self._zipfile.open(filename) as f:
+        filename = self.csv_filename(partition["year"])
+
+        with self.ds.get_zipfile_resource(self._dataset_name) as zf, zf.open(
```
davidmudrauskas, line 105: Unlike the Excel extractor and the earlier incarnation of the CSV extractor, the source contents aren't zipped into a folder but instead into a top-level CSV file. I'll check on whether that makes a difference (i.e., in Dagster testing).

src/pudl/extract/csv.py
e-belfer: Right now the year is getting passed into the extract function as part of `**partition`. We've switched to unpacking this to accommodate a few datasets that require more than one partition to point to a particular file (e.g., `year` _and_ `form`). Most but not all datasets use `year` as the partition, so we should avoid baking `year` into the extraction where possible, although it's certainly baked in currently (see, e.g. https://github.com/catalyst-cooperative/pudl/issues/3366).

src/pudl/extract/extractor.py
zaneselvans: I think you've got a choice of using a `Protocol` (Python's non-binding answer to interfaces / structural typing) or an abstract base class (`abc.ABC`) and some `@abstractmethod` decorators. We've mostly gone with ABC elsewhere.

I liked this video exploring the two options:
https://www.youtube.com/watch?v=xvb5hGLoK0A

src/pudl/extract/eia176.py
```diff
@@ -1,26 +1,34 @@
...
 
+raw_eia176__all_dfs = raw_df_factory(CsvExtractor, name="eia176")
 
-@asset(required_resource_keys={"datastore"})
-def raw_eia176__company(context):
-    """Extract raw EIA company data from CSV sheets into dataframes.
 
-    Args:
-        context: dagster keyword that provides access to resources and config.
+@multi_asset(
```
e-belfer, line 14: Multi-assets get cranky when there's only one asset (they wind up making an asset for each letter of the string in `raw_table_names`. This might change a bit depending on how the extractor gets refactored, but it currently should look something closer to:

```suggestion
from dagster import AssetOut, Output, asset
from pudl.extract.csv import CsvExtractor
from pudl.extract.extractor import raw_df_factory


class Eia176Extractor(CsvExtractor):
    """Extractor for EIA form 176."""

    def __init__(self, *args, **kwargs):
        """Initialize the module.

        Args:
            ds (:class:datastore.Datastore): Initialized datastore.
        """
        self._dataset_name = "eia176"
        super().__init__(dataset_name= self._dataset_name, *args, **kwargs)

raw_eia176__all_dfs = raw_df_factory(Eia176Extractor, name="eia176")

@asset(
    required_resource_keys={"datastore", "dataset_settings"},
)
def raw_eia176__data(context, raw_eia176__all_dfs):
    """Extract raw EIA company data from CSV sheets into dataframes.

    Returns:
        An extracted EIA 176 dataframe.
    """
    return (
        Output(value=df)
        for df in raw_eia176__all_dfs.items()
    )
```

src/pudl/extract/csv.py
e-belfer: I think there's a lot of potential to condense and refactor things here, if you're up for it! I left some thoughts on how I would go about restructuring this below.

e-belfer:
Thanks for wading into our extractors!

`process_raw`, `process_renamed` and `process_final` are functions that we tend to use when something is strange in the original data that prevents us from effectively extracting it - tons of empty columns we don't need to carry around, column names with `.` in them, etc. I think as much as possible preserving this functionality in the GenericExtractor would be ideal.

I'd suggest setting up a generic extraction function that is subclassed for both Excel and Csv but is composed of the following steps:

1. `load` - opens the file. Will be completely different for both data sources. I think this should probably be an `@abstractmethod`
2. `process` - includes `simplify_columns`, `process_raw` and `process_renamed`, as well as the append step. Identical for both, so this can just get moved into the GenericExtractor.
3. `validate` - includes all the column checks, also identical for both.
4. `combine` - combines all the datasets, and includes `process_final_page`.

If that's too big of a project, at the least I'd suggest adding `self.cols_added` to the `GenericExtractor` and pulling all the column checks into `validate(self, df)` in the `GenericExtractor` to cut down on redundancy. Then we could drop the processing steps from the CSV Extractor and revisit that question when we need to. 

Some other small changes to fix problems Dagster will bark at you about (annoyingly, I can't make suggestions on files you haven't changed, so pardon the long list):

- In `pudl.metadata.sources`, add the `working_partitions` to the current `eia176` dictionary entry: `"working_partitions": {"years": sorted(set(range(1997, 2023)))}` - this is basically all the years of data we expect to have.
- You'll want to make EIA 176 a settings object in `pudl.settings.py` - see `Eia923Settings` as an example of what this should look like. This should also get added to `EiaSettings`
- In `pudl.package_data.settings` `eia176` should get added to both the fast and full ETL to mirror the other EIA years.

davidmudrauskas:
Thanks for all the input. I've taken a pass at reconciling the extractor classes and I think it'll be relatively straightforward to wrap that up and see how it does in Dagster. That's how I'll spend Monday.

I'm keen on delivering something consolidated and clean, but if this drags out at all, I think it will be relatively easy to fall back to a less consolidated version.

davidmudrauskas:
I think this is making good progress. I think I've wrapped up what we need for unit tests. Abstracting the functionality has created more to review but I expect to wrap up soon. 

Let me know if you need me to switch up the pace and/or approach. I still have some follow-ups to address from above, and then I will run some Dagster jobs locally to check behavior. I'm planning on tackling that tomorrow evening.

src/pudl/extract/extractor.py
```diff
@@ -11,18 +13,239 @@
...
+    @staticmethod
+    def _get_partition_selection(partition: dict[str, PartitionSelection]) -> str:
+        """Grab the partition key."""
+        partition_names = list(partition.keys())
+        if len(partition_names) != 1:
+            raise AssertionError(
+                f"Expecting exactly one attribute to define this partition (found: {partition})"
+            )
+
+        partition_name = partition_names[0]
```
e-belfer, line 73: A file might currently be defined by more than one partition (e.g., form _and_ year) - PHMSA is a definite edge case, if it's helpful to look at how the partitions are structured there: https://zenodo.org/records/10231982

e-belfer: Ignore me, I see you've handled this below!

e-belfer:
Looks like there's a tiny docs issue but otherwise this direction is looking great so far! I'm going to ping another infrastructurally-minded person to take a look this week and see if they have additional feedback.

davidmudrauskas:
I've got the DAG working locally for all years (screenshot). I thought I'd be able to preview the output in the UI but I'm not quite finding that; when I peek at the output location at least, I see eia176-looking data of some formatting/encoding.

I plan on filling out the PR description tomorrow evening, along with any comments that I think will be helpful in the formal PR. The documentation step has failed again, despite a hail-mary attempt at fixing it. I'll have to find a better way to track down the unexpected indentation it's complaining about.

Finally, I noticed our Zenodo records have [multiple versions](https://zenodo.org/records/10607843). Right now we're pointing to a concrete version (instead of the more generic pointer to the latest version). I'm curious how we handle newer version.

Screenshot 2024-02-20 at 11 07 45 PM

e-belfer (changes requested):
This looks really great! Looking at the raw table in `pudl.devtools.inspect_assets.ipynb`, everything looks as expected with one exception. We should add a `report_year` column in the extraction step before we concatenate the pages, so that we know which records come from what year's files when they're all merged together. 

This could just look like subclassing `process_raw` or `process_renamed` to add a `report_year` column in the EIA 176 extraction module, I think.

Made one suggestion to fix the pesky docs-build error.

> Finally, I noticed our Zenodo records have [multiple versions](https://zenodo.org/records/10607843). Right now we're pointing to a concrete version (instead of the more generic pointer to the latest version). I'm curious how we handle newer version.

If I'm not wrong, this is to make it possible to clearly identify which version of the dataset different PUDL releases are using, so that someone using an older version of the code could pull the corresponding raw data versions and produce a database that exactly matches the versioned published output. But maybe @zaneselvans has more insight into this behavior. On the `pudl-archiver` side we always refer to the concept DOI.

src/pudl/extract/extractor.py
```diff
@@ -0,0 +1,374 @@
...
+logger = pudl.logging_helpers.get_logger(__name__)
+
+
+class GenericMetadata:
+    """Load generic metadata from Python package data.
+
+    When metadata object is instantiated, it is given ${dataset} name and it
+    will attempt to load csv files from pudl.package_data.${dataset} package.
+
+    It expects the following kinds of files:
```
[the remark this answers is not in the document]
davidmudrauskas: Thanks! I can't find `pudl.docs._build.autoapi.pudl.extract.extractor.index` locally. Maybe I just need to dig into the build more.

src/pudl/extract/eia176.py
```diff
@@ -1,26 +1,39 @@
...
+        Args:
+            ds (:class:datastore.Datastore): Initialized datastore.
+        """
+        self.METADATA = GenericMetadata("eia176")
+        super().__init__(*args, **kwargs)
+
+    def get_page_cols(self, page: str, partition_key: str) -> list[str]:
+        """Get the columns for a particular page and partition key."""
+        # The page columns for EIA176 do not vary by year
+        return super().get_page_cols(page, "any_year")
```
davidmudrauskas, line 24: I thought maybe I'd be able to bypass the storing of column mappings in a separate CSV like we do for Excel extractors, but this lined up better with how we're using factories and just ends up being more consistent and flexible overall.

src/pudl/extract/excel.py
```diff
@@ -181,32 +135,28 @@ class GenericExtractor:
     this method should be overwritten in the dataset-specific extractor.
     """
 
-    METADATA: Metadata = None
-    """Instance of metadata object to use with this extractor."""
-
-    BLACKLISTED_PAGES = []
-    """List of supported pages that should not be extracted."""
+    METADATA: ExcelMetadata = None
```
davidmudrauskas, line 138: Once we subclassed `GenericMetadata` for this type of extractor, I had to specify its type here to get method shortcuts working in my IDE (PyCharm).

src/pudl/extract/csv.py
```diff
@@ -1,72 +1,52 @@
...
+    """Class for extracting dataframes from CSV files.
 
     The extraction logic is invoked by calling extract() method of this class.
     """
 
-    def __init__(self, zipfile: ZipFile, table_file_map: dict[str, str]):
-        """Create a new instance of CsvExtractor.
-
-        This can be used for retrieving data from CSV files.
+    def source_filename(self, page: str, **partition: PartitionSelection) -> str:
```
davidmudrauskas, line 16: Type hints for **kwargs are not intuitive to me (we're actually working with a `dict` in `**partitions`) but I found defining `PartitionSelection` useful for me to wrap my head around the intention and use of the variables here. Happy to remove it if it isn't helpful.

src/pudl/extract/extractor.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        ...
+
+    def process_raw(
+        self, df: pd.DataFrame, page: str, **partition: PartitionSelection
+    ) -> pd.DataFrame:
+        """Takes any special steps for processing raw data and renaming columns."""
+        return df
+
+    @staticmethod
```
davidmudrauskas, line 155: `@staticmethod` annotations were pulled from the original, less abstract implementation. I'm not clear on the original motivation. For `process_renamed` at least the method (including subclass implementations) is only ever called at a point where an extractor object would have been instantiated.

src/pudl/extract/extractor.py
```diff
@@ -0,0 +1,379 @@
...
+    def _load_csv(package: str, filename: str) -> pd.DataFrame:
+        """Load metadata from a filename that is found in a package."""
+        return pd.read_csv(
+            importlib.resources.files(package) / filename, index_col=0, comment="#"
+        )
+
+    @staticmethod
+    def _get_partition_selection(partition: dict[str, PartitionSelection]) -> str:
```
davidmudrauskas, line 70: `PartitionSelection` can potentially specify multiple attributes (e.g., `{'form': 'gas_distribution', 'year'='2020'}`)` but as far as I'm seeing, Excel and CSV extractors all expect a single attribute for getting the filename, validation, etc. I think if we need to we can override this in a subclass for more complex cases.

src/pudl/package_data/settings/etl_fast.yml
```diff
@@ -38,6 +38,8 @@ datasets:
   ferc714:
     years: [2019, 2020]
   eia:
+    eia176:
+      years: [2020, 2022]
```
davidmudrauskas, line 42: The `description` associated with this doesn't match up at all -- I can make it more generic if that would be helpful.

davidmudrauskas:
I still need to do the implementations for EIA 191 and 757A to close out the ticket. Then I need to go through the PR checklist.

src/pudl/extract/extractor.py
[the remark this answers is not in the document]
e-belfer: Ah, it should show up once you run `make docs-build` and it fails - this is the same docs test that is getting run on CI. For See here for more info on the various tests that one can run on PUDL: https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html

src/pudl/extract/eia176.py
e-belfer: Do you mind just spelling out in the docstring what you're doing here for future reference - e.g., "EIA 176 data has the same set of columns for all years, so regardless of the partition key provided we select the same columns here" or something like this?

src/pudl/extract/excel.py
e-belfer: That makes sense and seems correct to me, we want to be specifying type as much as possible.

src/pudl/package_data/settings/etl_fast.yml
e-belfer: Which `description` are you referring to?

src/pudl/extract/extractor.py
e-belfer: I'm not sure either, it doesn't make intuitive sense to me - maybe @zschira you have thoughts?

e-belfer:
This looks great to me! I think after the last few little changes, we're going to want to run first the fast ETL and then the full ETL + validations to make sure that this PR doesn't introduce any unexpected changes into all the other datasets. If the full ETL proves too hard on your local machine we can kick off a build on Batch, so just let me know if you want me to do that. 

I also took a look through our docs and it doesn't seem like any changes are required, so you should be all good to go there.

src/pudl/extract/extractor.py
```diff
@@ -0,0 +1,375 @@
...
+    extractor_cls: type[GenericExtractor], name: str
+) -> OpDefinition:
+    """Construct a Dagster op that extracts one year of data, given an extractor class.
+
+    Args:
+        extractor_cls: Class of type :class:`Extractor` used to extract the data.
+        name: Name of an Excel based dataset (e.g. "eia860").
+    """
+
+    def extract_single_year(context, year: int) -> dict[str, pd.DataFrame]:
```
zschira, line 294: This is definitely a nitpick, but my preference when using factories like this to produce `op`s and `asset`s is to just apply the decorator to the internal functions like:

```py
def year_extractor_factory(
    extractor_cls: type[GenericExtractor], name: str
) -> OpDefinition:
    @op(
        required_resource_keys={"datastore", "dataset_settings"},
        name=f"extract_single_{name}_year",
    )
    def extract_single_year(context, year: int) -> dict[str, pd.DataFrame]:
        ...

    return extract_single_year

**Review by zschira — APPROVED** [review]
This looks good to me! I left a couple minor comments, but overall I like the idea of moving towards a gneric interface for our extractors

**Comment by davidmudrauskas** [comment]
FYI I plan on wrapping things up here sometime this week.

File: src/pudl/package_data/settings/etl_fast.yml (hunk 13 above)
Comment (davidmudrauskas) [reply src/pudl/package_data/settings/etl_fast.yml:42]: In this file (`etl_fast.yml`), the key-value pair:

```
description: >
  FERC 1 data from 2020 and 2021, EIA 860/923 from 2020 and 2022 (output to SQLite) plus
  EPA CEMS hourly emissions data from 2020 and 2022 (output to Parquet).
```

**Comment by davidmudrauskas** [comment]
I was able to run `ferc_to_sqlite_fast` and `etl_fast` successfully locally. I started with a clean slate under the latest conditions, having run `make install-pudl`, reactivating by `pudl-dev` environment, blowing away preexisting SQLite DBs, and running `alembic upgrade head` to get the latest schema. One step did fail along the way, but I was able to patch the bad reference easily and rerun that step for a successful overall materialization (see screenshots).

I had a slightly gnarly merge conflict I was able to resolve, but I wanted to call out here for good measure. I captured the relevant changes from conflicts with https://github.com/catalyst-cooperative/pudl/pull/3272 -- namely using the `calamine` engine to read Excel files. I'm motivated to get this merged before another conflict (😅). I'll run the full ETL tonight and see how it does.

Screenshot 2024-02-29 at 10 58 57 PM
Screenshot 2024-02-29 at 10 07 24 PM

**Comment by davidmudrauskas** [comment]
`etl_full` had a couple issues (after rerunning some steps successfully following timeouts from Zenodo):

```
dagster._core.errors.DagsterExecutionStepExecutionError: Error occurred while executing op "core_ferc1__yearly_depreciation_changes_sched219":
AssertionError: Found errors while running tests on the calculations:
                                  error_frequency  tolerance_error_frequency  is_error_error_frequency  relative_error_magnitude  tolerance_relative_error_magnitude  is_error_relative_error_magnitude  null_calculated_value_frequency  tolerance_null_calculated_value_frequency is_error_null_calculated_value_frequency  null_reported_value_frequency  tolerance_null_reported_value_frequency  is_error_null_reported_value_frequency
group     table_name group_value                                                                                                                                                                                                                                                                                                                                                                                                            
ungrouped ungrouped  ungrouped           0.056631                      0.058                     False                  0.045143                               0.045                               True                         0.012066                                        0.7                                    False                       0.543337                                      1.0                                   False
```

```
dagster._core.errors.DagsterExecutionStepExecutionError: Error occurred while executing op "core_ferc1__yearly_operating_revenues_sched300":
AssertionError: Found errors while running tests on the calculations:
                                                                             error_frequency  tolerance_error_frequency  is_error_error_frequency  relative_error_magnitude  tolerance_relative_error_magnitude  is_error_relative_error_magnitude  null_calculated_value_frequency  tolerance_null_calculated_value_frequency is_error_null_calculated_value_frequency  null_reported_value_frequency  tolerance_null_reported_value_frequency  is_error_null_reported_value_frequency
group            table_name                                     group_value                                                                                                                                                                                                                                                                                                                                                                                                            
utility_id_ferc1 core_ferc1__yearly_operating_revenues_sched300 275                 0.033333                       0.17                     False                   0.02507                                0.02                               True                              0.5                                        1.0                                    False                            0.0                                      1.0                                   False
```

I think it had previously been suggested some tolerance tests were too sensitive? I can look further into it.

File: src/pudl/package_data/settings/etl_fast.yml (hunk 13 above)
Comment (e-belfer) [reply src/pudl/package_data/settings/etl_fast.yml:43]: Oh yes, good catch! That definitely needs to get updated. How about:

> A smaller subset of data to test the ETL. This includes the most recent years of XBRL and DBF data for FERC data, and corresponding years for EIA and EPA CEMS data. FERC and EIA data is output to SQLite, and EPA CEMS hourly emissions data is output to Parquet.

Comment (e-belfer) [reply src/pudl/package_data/settings/etl_fast.yml:43]: Missed your rewrite of the description, which looks just fine as well!

**Comment by e-belfer** [comment]
@davidmudrauskas I'm going to kick off a full build on this branch to see if I can reproduce the calc check failures you're seeing. The code changes you made _shouldn't_ touch the FERC data much at all, so I can look into this and see what else is going wrong.

File: src/pudl/package_data/eia191/column_maps/data.csv [hunk 15]
```diff
@@ -0,0 +1,18 @@
...
+gas_field_code,gas_field_code
+reservoir_code,reservoir_code
+company_name,company_name
+field_name,field_name
+reservoir_name,reservoir_name
+field_type,field_type
+county_name,county_name
+status,status
+base_gas,base_gas
+working_gas_capacity_(mcf),working_gas_capacity_
```
e-belfer, line 15: Should be changed as we actually drop the parentheses before we rename the columns.

```suggestion
working_gas_capacity_mcf,working_gas_capacity_mcf
total_field_capacity_mcf,total_field_capacity_mcf
maximum_daily_delivery_mcf,maximum_daily_delivery_mcf
```

e-belfer (changes requested):
@davidmudrauskas I'm not seeing this calculation issue when I run the full ETL on the branch (instead, it fails on another occasional error we see relating to file read-ins, which itself is also not indicative of a problem with the branch). One possibility is that you ran `ferc_to_sqlite_fast` but not `ferc_to_sqlite_full` and so you're only reading in a subset of the FERC data to the full ETL.

So, the last blocking change is just a tiny bit of clean-up on the column naming CSVs for EIA 191. See my suggested change there. Once that's updated feel free to dismiss my review and add this PR to the merge queue, or I can do it on Monday!

davidmudrauskas:
I caught some issues running integration tests (via `make pytest-coverage`):

```
=========================================================================================== short test summary info ============================================================================================
FAILED test/integration/etl_test.py::TestCsvExtractor::test_extract_eia176 - AttributeError: module 'pudl.extract.csv' has no attribute 'get_table_file_map'
FAILED test/integration/etl_test.py::TestExcelExtractor::test_extract_eia860 - AttributeError: 'DataFrame' object has no attribute 'sheet_names'
FAILED test/integration/etl_test.py::TestExcelExtractor::test_extract_eia923 - AttributeError: 'DataFrame' object has no attribute 'sheet_names'
========================================================== 3 failed, 81 passed, 4 skipped, 6 xfailed, 2 xpassed, 1843 warnings in 15398.31s (4:16:38) ==========================================================
```

I was able to patch them, add a couple more for the new extractors, and run the failed and new ones successfully. I'm getting through `make pytest-validate` now, and once anything from that is settled I'll look for a final review and approval to unblock merging.

e-belfer (approved):
Looks good to go once the tests pass, thanks for all of your work on this one!

davidmudrauskas:
Looks like I still have to execute `ferc_to_sqlite_full` locally and then re-execute the failed steps in my `etl_full` execution to get things to a proper state to run validations without failure. Working through that.

e-belfer:
@davidmudrauskas If it becomes difficult to do this locally or more trouble than it's worth, we can rerun the full build on Google Batch. No trouble at all, just let me know if you'd like me to kick off a build!

davidmudrauskas:
@e-belfer Yeah let's do that, thanks. I think I've run into some stale artifacts and running the full jobs/tests/validations would be a lift.

e-belfer:
@davidmudrauskas No trouble at all, I'll kick things off now and merge them in if everything passes.

zaneselvans:
The batch build from this afternoon just passed, so I've queued this to merge.

PR #3403, opened 2024-02-16
Clean EIA 860 and 923 FGD operation and maintenance data

### LINKED ISSUE #3392: Clean EIA-923 Schedule 8C FGD Operation and Maintenance
Scope: create `_core_eia923__fgd_operation_maintenance` table that is cleaned and ready for harvesting with 860 FGD data.

Tasklist:
- [ ] standardize NAs
- [ ] convert thousands of dollars to dollars
- [ ] ensure primary key (year, plant_id, so2_control_unit) is unique

e-belfer:
# Overview

Closes #3392 and #3394.

Cleans 923 FGD Operation and Maintenance data and 860 FGD Equipment data for preparation for harvesting. 

What did you change?
- Drop values with plant and control id values of NA.
- Replace . values with NA.
- Convert dollars to thousands of dollars.
- Fix datetimes for SO2 test dates.
- Ensure a unique primary key and drop some duplicated rows.
- Write asset checks to check for continuity, null columns, and cost totals equaling their components

Still to do:
- [x] Resolve merge conflicts with main
- [x] Add all fields to PUDL metadata and encode them (uncomment existing lines).
- [x] Update coding tables to actually encode the relevant columns
- [x] Fix asset checks for fast ETL
- [x] Add fields for 860 records, add io manager and write `_core` table to DB
- [x] Write remaining tests for 860
- [x] Update alembic migrations
- [x] Why aren't encoding tables encoding?
- [x] After PR reviewed, redo alembic migrations to only have one
- [x] Fix docs build

Questions for reviewer:
We're dealing with `pct` inconsistently in PUDL. I'm standardizing to 0-1 scale, instead of 0-100. What "unit" suffix makes most sense here? `_rate`, `_ratio`, `_proportion`, etc.?

# Testing

Materialize `_core_eia923_fgd_operation_maintenance` and `_core_eia860_fgd_equipment`

```[tasklist]
# To-do list
- [x] Update release notes
```


src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,96 @@ def _core_eia923__fuel_receipts_costs(
...
+def _core_eia923__fgd_operation_maintenance(
+    raw_eia923__fgd_operation_maintenance: pd.DataFrame,
+) -> pd.DataFrame:
+    """Transforms the _core_eia923__fgd_operation_maintenance table.
+
+    Transformations include:
+
+    * Drop values with plant and boiler id values of NA.
+    * Replace . values with NA.
+    * Convert dollars to thousands of dollars.
```
aesharpe, line 1238: I think this is backwards - should be thousands of dollars to dollars?

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,96 @@ def _core_eia923__fuel_receipts_costs(
...
+    fgd_df = fgd_df.drop_duplicates()
+
+    # Replace the EIA923 NA value ('.') with a real NA value.
+    fgd_df = pudl.helpers.fix_eia_na(fgd_df)
+
+    # Convert thousands of dollars to dollars
+    fgd_df.loc[:, fgd_df.columns.str.endswith("_1000_dollars")] = (
+        fgd_df.loc[:, fgd_df.columns.str.endswith("_1000_dollars")] * 1000
+    )
+    fgd_df.columns = fgd_df.columns.str.replace("_1000_dollars", "")  # Rename columns
```
aesharpe, line 1261: I do the same conversion in my cooling table PR, so we should probably pick one way and do it the same in both. No strong preference. I think mine uses a `for` loop so this one is probably better! I will edit mine.

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,96 @@ def _core_eia923__fuel_receipts_costs(
...
+
+    # Convert SO2 test date column to datetime
+    # First, rescue a few troublesome datetimes that look like 3/08 or 01/08
```
aesharpe, line 1264: By rescue do you mean convert/fix? Wording is a little unclear.

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,96 @@ def _core_eia923__fuel_receipts_costs(
...
+    )
+    logger.info(
+        f"Dropping SO2 test dates that were not parseable: {fgd_df[(test_datetime.isnull())&(fgd_df.so2_test_date.notnull())].so2_test_date.unique()}"
+    )
+    fgd_df.loc[:, "so2_test_date"] = test_datetime
+
+    # Handle mixed boolean types in control flag column
+    fgd_df["fgd_control_flag"] = fgd_df["fgd_control_flag"].replace(
+        ["Y", "N", 1.0, 0.0], [True, False, True, False]
+    )
```
aesharpe, line 1289: You can also use the helper function `convert_col_to_bool`

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,96 @@ def _core_eia923__fuel_receipts_costs(
...
+
+    fgd_df["fgd_operational_status"] = fgd_df.fgd_operational_status.map(
+        pudl.helpers.label_map(
+            CODE_METADATA["core_eia__codes_operational_status"]["df"],
+            from_col="code",
+            to_col="label",  # OR operational_status? To discuss!
+            null_value=pd.NA,
+        )
+    )
```
aesharpe, line 1298: So I thiiiink the code metadata mapping happens automatically with the "encoder" step. I don't understand how this works very well though.

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,96 @@ def _core_eia923__fuel_receipts_costs(
...
+    # There are two remaining duplicates from plant_id_eia 6016, one row with cost data
+    # and one without. Keep the row with data.
+    dup_filt = (
+        (fgd_df.plant_id_eia == 6016)
+        & (fgd_df.so2_control_id_eia == 1)
+        & (fgd_df.report_year.isin([2018, 2019]))
+    )
+    # Drop the rows where every value other than the IDs and data maturity are null
+    fgd_df = fgd_df.loc[
+        ~((dup_filt) & (fgd_df.isna().sum(axis=1) >= len(fgd_df.columns) - 4))
```
aesharpe, line 1309: What is the -4 for?

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,96 @@ def _core_eia923__fuel_receipts_costs(
...
+    ]
+
+    fgd_df = (
+        pudl.metadata.classes.Package.from_resource_ids()
+        .get_resource("core_eia__yearly_so2_control_equipment")
+        .encode(fgd_df)
+    )
```
aesharpe, line 1316: I think this is the step that I was talking about above that will run the encoder without your needing to directly reference the `CODE_METADATA` there could be a caveat that I'm not aware of though.

src/pudl/package_data/eia923/column_maps/fgd_operation_maintenance.csv
```diff
@@ -6,8 +6,7 @@ opex_fgd_feed_materials_chemical_1000_dollars,feed_material_chemicals,feed_mater
...
-opex_fgd_maintenance_material_other_1000_dollars,maintenance_material_other,maintenance_material_other,maintenance_material_other,maintenance_material_other,,,,,,,,,,,
+opex_fgd_maintenance_material_other_1000_dollars,maintenance_material_other,maintenance_material_other,maintenance_material_other,maintenance_material_other,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars,fgd_maintenance_material_and_other_costs_thousand_dollars
```
aesharpe, line 9: What is the reasoning behind this change?

src/pudl/metadata/fields.py
```diff
@@ -2202,6 +2225,36 @@
         "description": "Production expenses: maintenance, supervision, and engineering (USD).",
         "unit": "USD",
     },
+    "opex_fgd_feed_materials_chemical": {
+        "type": "integer",
+        "unit": "dollars",
+        "description": "Annual operation and maintenance expenditures for feed materials and chemicals for FGD equipment, excluding electricity.",
+    },
```
aesharpe, line 2232: What are feed materials in this context?

aesharpe: Also maybe change the last bit to excluding the cost of electricity?

src/pudl/metadata/fields.py
```diff
@@ -3047,6 +3100,18 @@
         "description": "Identifies whether the reported value of emissions was measured, calculated, or measured and substitute.",
         "constraints": {"enum": EPACEMS_MEASUREMENT_CODES},
     },
+    "so2_removal_efficiency_100pct_load": {
+        "type": "number",
+        "description": "Removal efficiency for sulfur dioxide (to the nearest 0.1 percent by weight) at tested rate.",
```
aesharpe, line 3105: What does "at tested rate" mean?

src/pudl/metadata/fields.py
```diff
@@ -3047,6 +3100,18 @@
...
+    },
+    "so2_removal_efficiency_annual": {
+        "type": "number",
+        "description": "Removal efficiency for sulfur dioxide (to the nearest 0.1 percent by weight) at annual operating factor.",
```
aesharpe, line 3109: What is the annual operating factor?

src/pudl/metadata/resources/eia.py
```diff
@@ -1045,6 +1045,35 @@
         "sources": ["eia860", "eia923"],
         "etl_group": "outputs",
     },
+    "core_eia__yearly_so2_control_equipment": {
```
aesharpe, line 1048: I think it may be too soon to add this here if we haven't added it to the harvesting process yet.

src/pudl/transform/eia923.py
e-belfer: Oh definitely, good catch

src/pudl/package_data/eia923/column_maps/fgd_operation_maintenance.csv
e-belfer: Maintenance material and other is a stable category over time, that was getting mapped to two different columns depending on the original data source. It seems to be an error in the extraction, and I've collapsed this.

src/pudl/metadata/resources/eia.py
e-belfer: I'm using it to encode the table in preparation for harvesting, rather than adding the `_core` table here as the `_core` table won't be persisted to the database. It's jumping the gun a bit but I don't want to finish cleaning here without standardizing dtypes. Thoughts?

e-belfer: Switched this up a bit to reflect our likely pause on harvesting work in the near future.

src/pudl/metadata/fields.py
e-belfer: These are taken directly from the EIA form (same for below)

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1224,90 @@ def _core_eia923__fuel_receipts_costs(
...
+    fgd_df = fgd_df.dropna(subset=["plant_id_eia", "so2_control_id_eia"])
+    fgd_df = fgd_df.drop_duplicates()
+
+    # Replace the EIA923 NA value ('.') with a real NA value.
+    fgd_df = pudl.helpers.fix_eia_na(fgd_df)
+
+    # Convert thousands of dollars to dollars
+    fgd_df.loc[:, fgd_df.columns.str.endswith("_1000_dollars")] = (
+        fgd_df.loc[:, fgd_df.columns.str.endswith("_1000_dollars")] * 1000
+    )
```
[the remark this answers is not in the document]
e-belfer: Amazing, I'll switch it!

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1224,90 @@ def _core_eia923__fuel_receipts_costs(
...
+        ~((dup_filt) & (fgd_df.isna().sum(axis=1) >= len(fgd_df.columns) - 4))
+    ]
+
+    fgd_df = (
+        pudl.metadata.classes.Package.from_resource_ids()
+        .get_resource("core_eia923__yearly_fgd_operation_maintenance")
+        .encode(fgd_df)
+    )
+
+    return fgd_df
```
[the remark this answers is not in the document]
e-belfer: I think this will require serious downstream changes to the harvesting process that most EIA tables fit into, so that's out of scope here.

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1224,90 @@ def _core_eia923__fuel_receipts_costs(
...
+    # There are two remaining duplicates from plant_id_eia 6016, one row with cost data
+    # and one without. Keep the row with data.
+    dup_filt = (
+        (fgd_df.plant_id_eia == 6016)
+        & (fgd_df.so2_control_id_eia == 1)
+        & (fgd_df.report_year.isin([2018, 2019]))
+    )
+    # Drop the rows where every value other than the IDs and data maturity are null
+    fgd_df = fgd_df.loc[
+        ~((dup_filt) & (fgd_df.isna().sum(axis=1) >= len(fgd_df.columns) - 4))
```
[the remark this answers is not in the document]
jdangerx: I think `first()` [takes first non-NA value](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.first.html)!

```python
>>> df
  group    v1    v2
0     a  <NA>   1.2
1     a   1.1  <NA>
2     b     2     2
3     b     3     3
>>> df.groupby("group").first()
        v1   v2
group          
a      1.1  1.2
b        2    2
```

src/pudl/metadata/resources/eia.py
```diff
@@ -381,7 +393,13 @@
             ],
             "primary_key": ["code"],
             "foreign_key_rules": {
-                "fields": [["operational_status_code"]],
+                "fields": [
+                    [
+                        "operational_status_code",
+                        "cooling_status",
+                        "fgd_operational_status",
```
e-belfer, line 400: Add `code` suffix

src/pudl/helpers.py
```diff
@@ -1441,6 +1446,67 @@ def dedupe_on_category(
...
+    Returns:
+        The standardized dataframe.
+    """
+    logger.info(f"Standardizing ratios and percentages for {mixed_cols}")
+    for col in mixed_cols:
+        if not pd.api.types.is_numeric_dtype(frac_df[col]):
+            raise AssertionError(
+                f"{col}: Standardization method requires numeric dtype."
+            )
+        frac_df.loc[(frac_df[col] > 1) & (frac_df[col] <= 100), col] /= 100
```
zaneselvans, line 1501: Is there any way we can guard against the case where the reported value is between 0 and 1, and it's supposed to represent a fraction of a percent? This is a modest issue if the values overall are evenly distributed across the whole range from 0 to 100%, but if the expected values are on the order of 1% it could be quite wrong.

In some EIA data, we've found that the reporting format varies by year, with some years using 0.0-1.0 and others using 0-100%, and have restricted this kind of reformatting to apply only to the years reporting percentages.  In years with mixed reporting formats, it might also be the case that within a given utility's data, the format is consistent.

src/pudl/metadata/resources/eia860.py
```diff
@@ -708,6 +708,70 @@
...
         "etl_group": "eia860",
     },
+    "_core_eia860__fgd_equipment": {
+        "description": (
+            "Information about FGD equipment at generation facilities, "
+            "from EIA-860 Schedule 6E. Note: This table has been cleaned, but not "
+            "harvested with other EIA 923 or 860 data. The same variables present in "
+            "this table may show up in other _core tables in other years. Once this "
+            "table has been harvested, it will be removed from the PUDL database."
+        ),
```
zaneselvans, line 718: I think we should spell out what FGD means in the table description at least. It's used in many other names but never explicitly expanded as far as I've seen.

zaneselvans (changes requested):
I'm apprehensive about the blanket fraction vs. percentage standardization function.

If the encoding is still giving you a headache I'm happy to look at it or chat. I made that thing and it's more opaque than it should be.

src/pudl/metadata/fields.py
```diff
@@ -1302,7 +1306,64 @@
     },
     "ferc_qualifying_facility": {
         "type": "boolean",
-        "description": "Indicatates whether or not a generator is a qualifying FERC cogeneation facility.",
+        "description": "Indicatates whether or not a generator is a qualifying FERC cogeneration facility.",
+    },
+    "fgd_control_flag": {
+        "type": "boolean",
+        "description": "Indicatates whether or not a plant has a FGD control unit.",
```
zaneselvans, line 1313: I think we should spell out "flue gas desulfurization" in the long descriptions for clarity, since it's used in an abbreviated form in so many shorter names and might be confusing to someone unfamiliar with emissions controls.

jdangerx:
That docs build error looks like the error I fixed with my own docs changes - I can take a look to debug before taking a deeper look in the actual review.

src/pudl/transform/eia923.py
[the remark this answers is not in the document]
jdangerx: Yeah, using our PKs as indices on the dataframes is definitely a big question that is out of scope.

src/pudl/metadata/fields.py
```diff
@@ -1302,7 +1306,64 @@
...
+        "description": "The actual or projected in-service datetime of this flue gas desulfurization system",
+        "type": "date",
+    },
+    "fgd_operational_status_code": {
+        "type": "string",
+        "description": "Operating status code for flue gas desulfurization equipment.",
+    },
+    "fgd_manufacturer": {
+        "type": "string",
+        "description": "Nname of flue gas desulfurization equipment manufacturer.",
```
jdangerx, line 1344:
```suggestion
        "description": "Name of flue gas desulfurization equipment manufacturer.",
```

src/pudl/metadata/fields.py
```diff
@@ -1341,6 +1402,24 @@
         ),
         "type": "string",
     },
+    "flue_gas_bypass_fgd": {
+        "type": "boolean",
+        "description": "Indicates whether flue gas can bypass the flue gas desulfurization unit.",
+    },
+    "flue_gas_entering_fgd_pct_of_total": {
```
jdangerx, line 1409: Not for this PR, but we should maybe make clear when a column is expected to be a ratio (0-1) and when it's expected to be a percent.

Then users can use that explicit signal instead of "are all the values between 0 and 1? then probably it's a ratio."

Two ways I could imagine doing this are:
- naming columns with `_ratio` instead of `pct_of_total`
- using the `"unit"` to clarify "percent" vs. "ratio"

src/pudl/transform/eia860.py
```diff
@@ -1157,3 +1157,172 @@ def cooling_equipment_continuity(cooling_equipment):
...
+    # uniqueness.
+    fgd_df.loc[
+        (fgd_df.plant_id_eia == 6016)
+        & (fgd_df.report_date == "2011-01-01")
+        & (fgd_df.fgd_operating_date == "2009-03-01"),
+        "so2_control_id_eia",
+    ] = "01"
+
+    # Add a manufacturer name from the code.
+    fgd_df["fgd_manufacturer"] = fgd_df.fgd_manufacturer_code.map(
```
jdangerx, line 1230: Hm, should I have done this for any of the code columns in the cooling equipment? Seems slightly denormalizey but also probably useful to have the manufacturer name in here so 🤷

src/pudl/transform/eia860.py
```diff
@@ -1157,3 +1157,172 @@ def cooling_equipment_continuity(cooling_equipment):
...
+        fgd_df.pond_landfill_requirements_acre_foot_per_year, errors="coerce"
+    )
+
+    fgd_df = (
+        pudl.metadata.classes.Package.from_resource_ids()
+        .get_resource("_core_eia860__fgd_equipment")
+        .encode(fgd_df)
+    )
+
+    return fgd_df.pipe(apply_pudl_dtypes, strict=True)
```
jdangerx, line 1262: tiny nit: you can return this without the assignment
```suggestion
    return (
        pudl.metadata.classes.Package.from_resource_ids()
        .get_resource("_core_eia860__fgd_equipment")
        .encode(fgd_df)
        .pipe(apply_pudl_dtypes, strict=True)
    )
```

Or, if you want to keep that assignment around so you can shove a breakpoint in there and see what is "about to get returned", you can also do this:

```suggestion
    fgd_df = (
        pudl.metadata.classes.Package.from_resource_ids()
        .get_resource("_core_eia860__fgd_equipment")
        .encode(fgd_df)
        .pipe(apply_pudl_dtypes, strict=True)
    )

    return fgd_df
```

src/pudl/transform/eia860.py
```diff
@@ -1157,3 +1157,172 @@ def cooling_equipment_continuity(cooling_equipment):
...
+    """Check that columns other than expected columns aren't null."""
+    fast_run_null_cols = {
+        "county",
+        "fgd_operational_status_code",
+        "operating_date",
+        "fgd_manufacturer_code",
+        "plant_summer_capacity_mw",
+        "water_source",
+    }
+    if fgd.report_date.min() >= pd.Timestamp("2011-01-01T00:00:00"):
```
jdangerx, line 1276: non-blocking: this seems like it will be a pretty common pattern going forward, someday we should figure out a neater way of doing this.

src/pudl/transform/eia860.py
```diff
@@ -1157,3 +1157,172 @@ def cooling_equipment_continuity(cooling_equipment):
...
+@asset_check(asset=_core_eia860__fgd_equipment, blocking=True)
+def fgd_cost_discrepancy_check(fgd):
+    """Costs should sum to cost_total.
+
+    To allow for *some* data quality errors we assert that costs ~=
+    total cost at least 99% of the time (with a 1% acceptable
+    discrepancy).
+    """
+
+    def sum_to_target_rate(sum_cols, target, threshold):
```
jdangerx, line 1293: non-blocking: similar to above - "these columns sum to this other column" seems like a nice thing to streamline, eventually.

src/pudl/transform/eia860.py
```diff
@@ -1157,3 +1157,172 @@ def cooling_equipment_continuity(cooling_equipment):
...
+    return pudl.validate.group_mean_continuity_check(
+        df=fgd,
+        thresholds={
+            "flue_gas_exit_rate_cubic_feet_per_minute": 0.1,
+            "flue_gas_exit_temperature_fahrenheit": 0.1,
+            "pond_landfill_requirements_acre_foot_per_year": 0.1,
+            "so2_removal_efficiency_design": 0.1,
+            "so2_emission_rate_lbs_per_hour": 0.1,
+            "specifications_of_coal_ash": 0.1,
+            "specifications_of_coal_sulfur": 7,  # To do: investigate
```
jdangerx, line 1322: nit: somehow by convention `TODO` is how you mark todo's in code, and various editors/tooling all look for that specific word. so maybe worth doing this:

```suggestion
            "specifications_of_coal_sulfur": 7,  # TODO (2024-03-14): investigate
```

src/pudl/transform/eia923.py
```diff
@@ -1390,3 +1391,148 @@ def cooling_system_information_continuity(csi):
...
+    fgd_df = convert_col_to_bool(
+        df=fgd_df,
+        col_name="fgd_control_flag",
+        true_values=["Y", 1.0],
+        false_values=["N", 0.0],
+    )
+
+    # Take the non-NA values from each column for duplicate rows
+    pkey = ["plant_id_eia", "so2_control_id_eia", "report_date"]
+    fgd_df = pudl.helpers.dedupe_and_drop_nas(fgd_df, primary_key_cols=pkey)
```
jdangerx, line 1465: TODO (for me): I'll use this method for the EIA923 cooling systems information stuff after this gets merged.

src/pudl/transform/eia923.py
```diff
@@ -1390,3 +1391,148 @@ def cooling_system_information_continuity(csi):
...
+
+    fgd_df = (
+        pudl.metadata.classes.Package.from_resource_ids()
+        .get_resource("_core_eia923__fgd_operation_maintenance")
+        .encode(fgd_df)
+    )
+
+    return fgd_df.pipe(apply_pudl_dtypes, strict=False)
```
jdangerx, line 1473: See the comment above about streamlining this return statement.

jdangerx (changes requested):
A bunch of tiny little suggestions that you can take or leave, stylistically. 

But, blocking concerns:
- [x] we should probably write a test for `dedupe_drop_na` which will clarify what sorts of deduping it does - what if there are two rows that both have *some* non-NA values and some NA values, but their non-NA values never conflict? Should we pick up values from both rows? I think so, but we should make sure that's clear...
- [x] probably worth addressing Zane's concern about the percentage conversion with some fresh docs at least.

Otherwise looks pretty good! Ping me again and I'll gladly re-review. Also I'm gonna kick off some materializations now but I assume you've been materializing successfully locally already...

src/pudl/helpers.py
```diff
@@ -1446,6 +1446,67 @@ def dedupe_on_category(
...
+        if frac_df[col].max() > 1:
+            logger.warn(
```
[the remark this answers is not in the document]
zaneselvans: If it's a warning we'll probably ignore it down the line.

If there are heuristic checks that would identify when we're incorrectly categorizing something as a percentage vs. a fraction, maybe that's a kind of data validation we should be running on any column that's supposed to end up being a fraction?

Besides the case where there's a lot of values around 1% (which we could maybe notice based on there being a bimodal distribution of values ~100x off from each other?) the case where there are a few values that just naturally happen to be less than 1% but which then get multiplied by 100x and thus wildly & systematically magnified could be pretty weird in a different way, and harder to notice, since the values would just get scattered throughout the rest of the spectrum of values between 0.01-1.0

src/pudl/transform/eia923.py
e-belfer: I can just do it in this PR, no trouble

src/pudl/transform/eia860.py
e-belfer: This column doesn't seem to exist in the 860 cooling tables, unless I'm missing something?

src/pudl/helpers.py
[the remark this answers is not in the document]
e-belfer: Fixed to raise an error and to add a date range for this function to be applied over.

src/pudl/transform/eia923.py
```diff
@@ -1301,16 +1302,13 @@ def _core_eia923__cooling_system_information(
...
-    unduplicated = csi_df.loc[~dupe_mask]
-
     resource = pudl.metadata.classes.Package.from_resource_ids().get_resource(
         "_core_eia923__cooling_system_information"
     )
+
+    primary_key = ["plant_id_eia", "report_date", "cooling_id_eia"]
     return (
-        pd.concat([unduplicated, deduplicated], ignore_index=True)
+        pudl.helpers.dedupe_and_drop_nas(csi_df, primary_key_cols=primary_key)
```
jdangerx, line 1311: Ha! Thanks :)

jdangerx (approved):
Looks good to me, let's smash that MF merge-when-ready button!

zaneselvans:
@e-belfer It looks like there were significant decreases in test coverage for `pudl.transform.eia923` and `pudl.transform.eia860` [based on the CodeCov report](https://app.codecov.io/gh/catalyst-cooperative/pudl/commit/8fc2ca69a0e6a02cecffa88ff3441945e9a5dd07). Is this expected?  It seems like maybe some of the new functions aren't actually getting called?

jdangerx:
Well, we know that the asset checks are actually running, because a failed check can fail the integration tests. But, for some reason codecov isn't actually marking them as covered...

e-belfer:
> but if that turns out to be some deep rabbit-hole we can mark the asset checks as #pragma: no cover and sort out the issue later.

This has indeed turned out to be a rabbit hole. I'm going to write up a new issue to keep track of some asset check improvements including this one, and `pragma: no cover` the asset checks for now.

Issue #1196, opened 2021-09-06
Debug foreign key constraints in direct-to-sqlite ETL


zaneselvans:
With foreign key constraint checking turned on, we get integrity errors during the main ETL. The first errors are related to tables where the foreign key and primary key have different names. E.g. the `fuel_type_code_aer` column in the `generation_fuel_eia923` table is a foreign key to the `abbr` column in the `fuel_type_aer_eia923` table. Not sure if this is related to the failure but it did get through half a dozen normal tables before having this problem.

## Tasks
* [ ] Debug failing foreign key constraints.
* [ ] Compare the ETL run time with foreign key constraint checking enabled vs. disabled to see whether it's worth trying to defer the check until after the DB has been completely loaded.

ezwelty:
That foreign key error is due to a foreign key error ;)
The problem is that `fuel_type_aer_eia923.fuel_type` references `fuel_type_eia923.abbr`, but the values of the latter are not present in the former (in fact, one is a code, the other a description). So I would assume this foreign key relationship is spurious, and exclude it by updating the foreign key rules of `fuel_type_eia923`.

```py
{
    "fuel_type_eia923": {
        "schema": {
            "fields": ["abbr", "fuel_type"],
            "primary_key": ["abbr"],
            "foreign_key_rules": {
                "fields": [["fuel_type"], ["fuel_type_code"]],
                "exclude": ["fuel_type_aer_eia923"]
            },
        },
        "sources": ["eia923"]
}
```

I'll work through these in the morning, unless you beat me to it.

ezwelty:
@zaneselvans Here are all the foreign key violations. You should review them to decide which are spurious and can be excluded (by modifying the foreign key rules or renaming columns) and which suggest problems with the data.

```
[{'resource': 'plants_small_ferc1',
  'foreign_key': {'fields': ['fuel_type'],
   'reference': {'resource': 'fuel_type_eia923', 'fields': ['abbr']}},
  'invalid':     fuel_type
  0            
  1         oil
  2         oil
  3         oil
  4     oil/gas
  ..        ...
  699          
  700          
  701          
  702          
  703          
  
  [704 rows x 1 columns]},
 {'resource': 'fuel_type_aer_eia923',
  'foreign_key': {'fields': ['fuel_type'],
   'reference': {'resource': 'fuel_type_eia923', 'fields': ['abbr']}},
  'invalid':                                           fuel_type
  0                              Solar PV and thermal
[...]

PR #3405, opened 2024-02-16
Transform EIA860 and EIA923 Cooling System Tables

### LINKED ISSUE #3393: Clean EIA-923 Schedule 8D Cooling System Information

aesharpe:
# Overview

Closes #3393 and #3395.

Overall structure:
- EIA-923 cooling systems information + EIA860 cooling equipment had some unit conversions, spot fixes, and deduplication
- defined resource schemas + fields for both CSI + CE; this entailed some renaming of raw data fields
- defined a bunch of new codes (cooling type, water source type, tower type, etc) and had to do some remapping for CSI
- added asset check infrastructure + asset checks
- did some refactoring of how we make sphinx refs for resources in docs - this lets us make refs to _* resources
- DB migration

# Testing

Generate `_core_eia923__cooling_system_information` and `_core_eia860_cooling_equipment` for both full and fast ETLs.

```[tasklist]
# To-do list
```


src/pudl/helpers.py
```diff
@@ -983,7 +983,8 @@ def convert_to_date(
         year_col: the name of the year column in the original table.
         month_col: the name of the month column in the original table.
         day_col: the name of the day column in the original table.
-        month_value: generated month if no month exists.
+        month_value: generated month if no month exists or if the month
+            value is NA.
```
aesharpe, line 987: I updated this function because there was a month column but it contained NA values for the years where it was annual vs. monthly. I wanted to fill those in with 1, but the previous `convert_to_date` function only fills in the month if there is no month column at all, which there is. So I edited it to fill in the month value if the month column contains NA values. 

I did not check whether this has any downstream effects in other tables, but my guess is that it doesn't because most tables are not combining annual and monthly data and therefore don't have NA values in the month column.

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,46 @@ def _core_eia923__fuel_receipts_costs(
...
+    csi_df = raw_eia923__cooling_system_information
+    # Generic cleaning
+    csi_df = csi_df.pipe(pudl.helpers.fix_eia_na).pipe(pudl.helpers.convert_to_date)
+
+    # Convert cooling_status from code to description
+    csi_df["cooling_status"] = csi_df.cooling_status.str.upper().map(
+        pudl.helpers.label_map(
+            CODE_METADATA["core_eia__codes_operational_status"]["df"],
+            from_col="code",
+            to_col="description",
```
aesharpe, line 1242: Not sure if this should be the description or the label or something else.

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,46 @@ def _core_eia923__fuel_receipts_costs(
...
+    # Combine the annual rates with the monthly rates
+    cols = {
+        "withdrawal_rate_gallons_per_minute": "annual_withdrawal_rate_tenth_cubic_feet_per_second",
+        "discharge_rate_gallons_per_minute": "annual_discharge_rate_tenth_cubic_feet_per_second",
+        "consumption_rate_gallons_per_minute": "annual_consumption_rate_tenth_cubic_feet_per_second",
+    }
+    for new_col, old_col in cols.items():
+        csi_df.loc[csi_df["report_date"].dt.year.isin([2008, 2009]), new_col] = csi_df[
+            old_col
+        ]
```
aesharpe, line 1256: The EIA923 cooling table was reported annually in 2008 and 2009 and monthly thereafter. Because all the values in the table are either static or rates (ex: water consumption in gallons/min averaged out over either the month or the year), I think it is probably ok to keep them combined. These rates to vary by month, but if there is only one month reported for the whole year, I suppose the user can assume that it's a yearly vs. a monthly average (we can also include this in the column description). I don't think it's worth separating these years out into another table.

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1223,46 @@ def _core_eia923__fuel_receipts_costs(
...
+
+    # In 2008 and 2009 the rate columns are clearly labeled as 0.1 cubic feet / second
+    # but that rate persists until 2013 when they change it to gallons/min.
+    # Convert tenth cubic feet/second columns to gallons/minute.
+    conversion = 10 / 60 * 7.48052
+
+    for new_col in cols:
+        csi_df.loc[csi_df["report_date"].dt.year.isin(range(2008, 2013)), new_col] = (
+            csi_df[new_col] * conversion
+        )
```
aesharpe, line 1266: Something weird happens between 2012 and 2013. The values for the rates go way up! The column also changes from 0.1 cubic feet / second to gallons per minutes, but even with the conversion in place the values spike in 2013. Here is the average rate for each year: 

Screen Shot 2024-02-16 at 12 36 01 PM

Screen Shot 2024-02-16 at 12 35 36 PM

Screen Shot 2024-02-16 at 12 35 21 PM

aesharpe:
I think it probably makes sense to keep these tables separate post harvesting just because there is so much information that would be duplicated from the annual 860 table to the monthly 923 table.

src/pudl/metadata/codes.py
```diff
@@ -2295,6 +2295,133 @@
...
         "ignored_codes": [],
     },
+    "core_eia__codes_cooling_tower_types": {
+        "df": pd.DataFrame(
+            columns=["code", "label", "description"],
+            data=[
+                (
+                    "MD",
+                    "mechanical_dry",
+                    "Mechanical draft, dry process",
```
aesharpe, line 2305: The docs build is failing because I need to add these to the `RESOURCE_METADATA` table in `metadata/resources/eia.py`. 

I added it, but it's causing dagster to fail, I'm not exactly sure why.

src/pudl/metadata/resources/eia.py
```diff
@@ -1045,6 +1045,56 @@
...
     },
+    # "core_eia__codes_cooling_tower_types": {
+    #     "description": "A coding table for cooling tower types.",
+    #     "schema": {
+    #         "fields": ["code", "label", "description"],
+    #         "primary_key": ["code"],
+    #         "foreign_key_rules": {
+    #             "fields": [
+    #                 ["tower_type_1", "tower_type_2", "tower_type_3", "tower_type_4"]
+    #             ]
```
aesharpe, line 1056: This is all commented out because it is somehow breaking the dagster launch. Need to dig more into why. If we wanted to leave all the encoding stuff out of this PR we could because it's more for the harvesting step anyways. I didn't realize how interconnected it all was. My initial thought was just to add the `CODE_METADATA` but the builds would fail without the coding tables depicted here (see [comment](https://github.com/catalyst-cooperative/pudl/pull/3405/files#r1497026657)). And then these cause dagster to fail. So need to figure out the chain of events.

e-belfer (changes requested):
Major questions about conversation rates and the monthly/annual munging. If you can give me more info about the dagster issue you're having I can try and help debug, but I'm not seeing any problem locally.

e-belfer:
> I think it probably makes sense to keep these tables separate post harvesting just because there is so much information that would be duplicated from the annual 860 table to the monthly 923 table.

I'm confused about this - to me the goal of harvesting is to combine duplicate info reported at different frequencies and reconcile it. Do you mean that the fields are overlapping, or what is meant here?

src/pudl/transform/eia923.py
aesharpe: I like the idea of creating an annual average column and averaging the monthly values there instead of combining the annual and the monthly. Is there anything else you're confused about here that I could clarify?

src/pudl/transform/eia923.py
aesharpe: I don't think this has to do with the annual vs monthly because that only applies for 2008 and 2009 and this issue goes through 2012. Also these values are the annual average.

src/pudl/package_data/eia860/column_maps/cooling_equipment.csv
```diff
@@ -36,10 +36,10 @@ tower_water_rate_100pct_gallons_per_minute,,,,,tower_water_rate_gallons_per_minu
...
-cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment,cost_chlorine_equipment,cost_chlorine_equipment,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars
+cooling_equipment_total_cost_thousand_dollars,cost_total,cost_total,cost_total,cost_total_thousand_dollars,cost_total_thousand_dollars,cost_total_thousand_dollars,cost_total_thousand_dollars,cost_total_thousand_dollars,cost_total_thousand_dollars,cost_total_thousand_dollars,cost_total_thousand_dollars,cost_total_thousand_dollars,cost_total_thousand_dollars,cost_total_thousand_dollars
```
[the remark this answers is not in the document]
aesharpe: given that it get's converted to dollars I'm not hugely motivated to change it

src/pudl/helpers.py
```diff
@@ -963,8 +963,8 @@ def convert_to_date(
     year_col: str = "report_year",
     month_col: str = "report_month",
     day_col: str = "report_day",
-    month_value: int = 1,
-    day_value: int = 1,
+    month_na_value: int = 1,
```
jdangerx, line 966: Renamed these to be a little clearer about what *type* of value this is.

src/pudl/transform/eia860.py
```diff
@@ -1021,3 +1021,88 @@ def _core_eia860__boiler_stack_flue(
...
+
+    - only completely null cols are tower type 3 and 4
+    - no measurement cols move by 80% or more over the course of one report period.
+    """
+    expected_null_cols = {"tower_type_3", "tower_type_4"}
+    pudl.validate.no_null_cols(
+        cooling_equipment,
+        cols=set(cooling_equipment.columns) - expected_null_cols,
+    )
+    smoothly_varying_cols = [
```
jdangerx, line 1082: Just took a stab at "stuff that seems like it should mostly move smoothly." We could easily extract this into a utility function, and even include different thresholds per column - but figured this was OK for a quick & dirty check.

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1224,59 @@ def _core_eia923__fuel_receipts_costs(
...
+
+    return csi_df.pipe(apply_pudl_dtypes, group="eia923")
+
+
+@asset_check(asset=_core_eia923__cooling_system_information, blocking=True)
+def cooling_system_information_check(csi):
+    """Check data quality.
+
+    - no completely null columns
+    - withdrawal == discharge + consumption at least 90% of the time
```
jdangerx, line 1269: This was a property that seems like it's always true in theory - based on a bunch of spot checks. But 🤷 

Hmm, I should probably pull out that 'slowly varying' logic from above and apply that to the various rates here, too. A TODO for tomorrow.

src/pudl/transform/eia860.py
e-belfer: This makes sense intuitively but I'm wary of defining completely ad hoc validation tests for each table, ideally we'd have uniform validation expectations across each dataset at least.

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1224,59 @@ def _core_eia923__fuel_receipts_costs(
...
+
+@asset
+def _core_eia923__cooling_system_information(
+    raw_eia923__cooling_system_information: pd.DataFrame,
+) -> pd.DataFrame:
+    """Transforms the eia923__cooling_system_information dataframe.
+
+    Applies typical NA conversion and date conversion.
+
+    As of 2024-02-28: 2008 and 2009 only have annual rates, but the
```
e-belfer, line 1237: Rather than dropping this data, I'd make an annual rate column for each of these and just move the 2008-9 data in there. There's no good reason to drop the values, even though they shouldn't be in the same column as the rest of the data for legibility.

src/pudl/transform/eia923.py
e-belfer: The description and the actual math are slightly different here, would be good to explain this isn't `==` but actually ~99% comparison.

src/pudl/transform/eia923.py
jdangerx: There's already an "annual" columns that only have non-null values for both of these years, but it's the raw cfs value instead of GPM - do you mean we should make an annual GPM column and remove the annual CFS column?

e-belfer: So the raw data reports values in both columns for 2008-9 data? Are they different than one another?
I think the ideal outcome would be one monthly and one annual column with the same units. Whether we fill the annual column with monthly data I feel less set on.

src/pudl/transform/eia860.py
```diff
@@ -1021,3 +1021,43 @@ def _core_eia860__boiler_stack_flue(
...
+        ce_df.loc[ce_df["report_date"].dt.year.isin(range(2009, 2013)), col_minutes] = (
+            ce_df[col_seconds] * conversion
+        )
+
+    # Convert thousands of dollars to dollars and remove suffix from column name
+    for col in ce_df:
+        if "thousand_dollars" in col:
+            ce_df.loc[:, col] = ce_df[col] * 1000
+    ce_df.columns = [col.replace("_thousand_dollars", "") for col in ce_df.columns]
+
```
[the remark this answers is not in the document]
jdangerx: Ah, I see, the fields haven't been defined with a type or description in `fields.py` - so then `apply_pudl_dtypes` doesn't do anything. I'll go through and add these field descriptions/types too.

src/pudl/transform/eia923.py
jdangerx: Raw data only reports annual data for 2008-09; and then it only reports non-annual data for 2010+

jdangerx:
TODO:
- [ ] Still have the EIA-860 fields to do, but the EIA-923 ones appear to be behaving themselves. 
- [ ] pull the slowly-varying-fields check into a `pudl.validate` utility method
- [ ] daydream about revamping our homegrown fields setup

src/pudl/package_data/eia923/column_maps/cooling_system_information.csv
```diff
@@ -6,26 +6,26 @@ cooling_id_eia,cooling_system_id,cooling_system_id,cooling_system_id,cooling_sys
...
-discharge_volume_million_gallons,,,,,discharge_volume_million_gallons,discharge_volume_million_gallons,discharge_volume_million_gallons,discharge_volume_million_gallons,discharge_volume_million_gallons,discharge_volume_million_gallons,discharge_volume_million_gallons,discharge_volume_million_gallons,discharge_volume_million_gallons,discharge_volume_million_gallons,discharge_volume_million_gallons
-consumption_volume_million_gallons,,,,,consumption_volume_million_gallons,consumption_volume_million_gallons,consumption_volume_million_gallons,consumption_volume_million_gallons,consumption_volume_million_gallons,consumption_volume_million_gallons,consumption_volume_million_gallons,consumption_volume_million_gallons,consumption_volume_million_gallons,consumption_volume_million_gallons,consumption_volume_million_gallons
-chlorine_annual_1000_lbs,annual_chlorine,annual_chlorine,,,,,,,,,,,,,
-annual_withdrawal_rate_cubic_feet_per_second,annual_withdrawal_rate,annual_withdrawal_rate,,,,,,,,,,,,,
```
e-belfer: PUDL standard is to minimize abbreviations unless absolutely ridiculous (e.g., multi-letter acronyms for leak detection protocols that are widely known), so these should actually stay `cubic_feet_per_second` and `fahrenheit` as Austen originally had them.

src/pudl/package_data/eia860/column_maps/cooling_equipment.csv
```diff
@@ -34,16 +34,15 @@ tower_type_3,,,,,tower_type_3,tower_type_3,tower_type_3,tower_type_3,tower_type_
...
-cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment,cost_chlorine_equipment,cost_chlorine_equipment,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars,cost_chlorine_equipment_thousand_dollars
+power_requirement_mw,power_requirement_kwh,power_requirement_kwh,power_requirement_kwh,power_requirement_mw,power_requirement_mw,power_requirement_mw,power_requirement_mw,power_requirement_mw,power_requirement_mw,power_requirement_mw,power_requirement_mw,power_requirement_mw,power_requirement_mw,power_requirement_mw
```
jdangerx, line 37: There's a weird thing here where, for a few years, the "power requirement (MW)" column was reported as `power_requirement_kwh`. I did some digging to confirm that this was an error and that `_kwh` should be mapped to `_mw` - is there somewhere I can document that digging for the future?

src/pudl/metadata/fields.py
jdangerx, line 1: I think there's a relatively straightforward approach to making our fields easier to work with, which this comment box is too small to contain.

src/pudl/package_data/eia923/column_maps/cooling_system_information.csv
```diff
@@ -6,26 +6,26 @@ cooling_id_eia,cooling_system_id,cooling_system_id,cooling_system_id,cooling_sys
...
-maximum_discharge_temperature_fahreneit,,,discharge_maximum_temperature,discharge_maximum_temperature,discharge_maximum_temperature_f,discharge_maximum_temperature_f,discharge_maximum_temperature_f,discharge_maximum_temperature_f,discharge_maximum_temperature_f,discharge_maximum_temperature_f,discharge_maximum_temperature_f,discharge_maximum_temperature_f,discharge_maximum_temperature_f,discharge_maximum_temperature_f,discharge_maximum_temperature_f
+monthly_average_intake_temperature_fahrenheit,,,intake_averagetemperature,intake_average_temperature,intake_average_temperature_f,intake_average_temperature_f,intake_average_temperature_f,intake_average_temperature_f,intake_average_temperature_f,intake_average_temperature_f,intake_average_temperature_f,intake_average_temperature_f,intake_average_temperature_f,intake_average_temperature_f,intake_average_temperature_f
```
jdangerx, line 15: @e-belfer I changed these back to their long forms!

jdangerx (commented):
@e-belfer I think this is properly ready for review:

- fields have dtypes associated
- code tables have been added to db schema
- fixed conversion rates
- renamed fields to clarify their content
- removed fields that didn't need to be around anymore
- added asset checks, including some continuity check tooling

src/pudl/metadata/fields.py
```diff
@@ -620,6 +616,15 @@
         "type": "integer",
         "description": "Year boiler was or is expected to be in compliance with federal, state and/or local regulations for sulfur dioxide emissions.",
     },
+    "cooling_id_eia": {
+        "description": (
+            "Identification Code for Cooling System (if multiple cooling "
```
e-belfer, line 621: Can we normalize the capitalization here to just say "Identification code reporting to EIA for cooling system (if multiple...."

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1224,126 @@ def _core_eia923__fuel_receipts_costs(
...
+@asset_check(asset=_core_eia923__cooling_system_information, blocking=True)
+def cooling_system_information_continuity(csi):
+    """Check to see if columns vary as slowly as expected."""
+    return pudl.validate.group_mean_continuity_check(
+        df=csi,
+        thresholds={
+            "annual_average_consumption_rate_gallons_per_minute": 0.1,
+            "annual_average_discharge_rate_gallons_per_minute": 0.1,
+            "annual_average_withdrawal_rate_gallons_per_minute": 0.1,
+            "monthly_average_consumption_rate_gallons_per_minute": 0.5,
```
e-belfer, line 1338: I wonder if we want to up `n_outliers_allowed` and drop this down as it seems super high, but that's probably out of scope for this PR.

src/pudl/metadata/codes.py
```diff
@@ -2318,6 +2318,160 @@
...
+                ),
+                (
+                    "BE",
+                    "reclaimed_water",
+                    "Reclaimed water (ex: treated wastewater effluent)",
+                ),
+                (
+                    "SA",
+                    "saline_water",
+                    "Saline Water",
```
e-belfer, line 2402: Very much a nit: Can we standardize to "water" instead of using a mix of "Water" and "water" across new coding tables?

src/pudl/etl/__init__.py
```diff
@@ -41,6 +42,9 @@
 
 logger = pudl.logging_helpers.get_logger(__name__)
 
+# Asset Checks are still Experimental, silence the warning since we use them
+# everywhere.
+warnings.filterwarnings("ignore", category=ExperimentalWarning)
```
e-belfer, line 47: I do still see this error when I `launch_dagster` just FYI, but it's probably helpful to see this warning somewhere.

e-belfer (changes requested):
Baby nits mostly left, plus some older comments I forgot.

One major question: I don't see the asset checks in the Dagster UI when I materialize them, but they do materialize successfully. I've updated my environment locally. Is that expected?

src/pudl/transform/eia923.py
jdangerx: I took a look at the `top` values for this and it was something like 1.4/0.49/0.48/0.46 etc - I think probably there's just pretty large month-to-month variation, but we could dig deeper into this once we have more time.

src/pudl/metadata/fields.py
```diff
@@ -3680,7 +3685,418 @@
             }
         },
     },
-    "eia": {"fuel_units": {"constraints": {"enum": sorted(FUEL_UNITS_EIA.keys())}}},
+    "eia": {
```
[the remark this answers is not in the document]
jdangerx: screed alert, sorry:

While our current system is "put all fields in one big bucket," I think we should move towards splitting our fields up into meaningful categories.

The benefit of putting everything in one general bucket is that any resource can use these fields without having to redefine them.

The downside is that it makes it harder to add new fields and asset schemas down the line:
- when adding a new field with the same name as an existing one, it's hard to know where the existing one came from & what contexts its used in, so it's hard to make an informed decision about what to do with the new field
- if you are thinking about what to name your column, and want to see what other related assets have named a similar column, you are stuck wading through *all* columns ever named

It's sort of like shoving all of your functions into one module.

We attempted to mitigate these downsides with our group/resource override system, which is what I'm using here, but that system is pretty janky...
* `Resource.dict_from_resource_descriptor` does a bunch of logic to figure out which override to use, but...
*`Field.from_id` doesn't understand overrides directly
*`get_pudl_dtypes` applies group overrides, but not resource overrides

My dream is to evolve our current system into one that follows industry practices such as

1. Bundling the field definitions *with* the schema definitions, like most other schema systems like Frictionless, Pandera, SQLAlchemy, Pydantic etc.
2. Allowing for re-use of fields by direct, explicit reference - also like those above systems.

And I think that shoving these fields into an `"eia"` group is how we stop digging this hole quite so quickly while we wait for time to make the system better. I had initially hoped to shove them into "eia923_cooling" and "eia860_cooling" groups, but our system doesn't support "multiple namespaces" like that...

src/pudl/transform/eia860.py
```diff
@@ -1021,3 +1021,95 @@ def _core_eia860__boiler_stack_flue(
...
+    # Convert thousands of dollars to dollars and remove suffix from column name
+    ce_df.loc[:, ce_df.columns.str.endswith("_thousand_dollars")] *= 1000
+    ce_df.columns = ce_df.columns.str.replace("_thousand_dollars", "")
+
+    return ce_df.pipe(apply_pudl_dtypes, group="eia", strict=True)
+
+
+@asset_check(asset=_core_eia860__cooling_equipment, blocking=True)
+def cooling_equipment_null_cols(cooling_equipment):
+    """The only null cols we expect are tower type 3 and 4."""
```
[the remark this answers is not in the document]
jdangerx: I think this check *should* be fine in fast ETL since we expect tower 3 and tower 4 in recent years. In other cases, I think we should probably get to a point where we're enforcing that the assets have the same schemas regardless of what years they were run for.

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
jdangerx: @zaneselvans @cmgosnell : Ella & I talked a bit more about this and have a couple bullet points:

* having One Big Honkin namespace does provide some structural defense against bad naming
* but, if we want to have field defs colocated with schema definitions at some point in the future (which makes defining schemas + resources much easier), we will have to abandon One Big Dictionary at some point

We're curious to see if there's other context about why we have the fields set up the way they are, and what you think about putting these fields in the general dictionary or keeping them separate.

src/pudl/transform/eia860.py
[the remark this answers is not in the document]
e-belfer: I more mean if there's a field like `annual_rate` that only exists for 2008-09, it'll always be entirely null for 2020/2021 runs.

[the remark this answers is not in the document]
jdangerx: Ah, yeah, I guess that will be the case, good thought!

I think the fast way around this is to have `cols_that_are_allowed_to_be_all_null` or some better name, and include the annual rate cols. WDYT?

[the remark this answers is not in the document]
e-belfer: Seems fine to me. Other alternative is to only run this check on the full ETL which would lose some information.

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
zaneselvans: I think the main motivation that we had when we set up the One Giant Namespace was avoiding redefining the same columns in many different places, and needing to keep those definitions synchronized, which of course never happens reliably -- this was a problem we had with the system prior to creating the current metadata setup.

I could imagine the opposite convention being fine, where by default each resource has to define its own fields, but it has the option of explicitly grabbing a definition from somewhere else to reuse it if it means the same thing / represents the same information, to avoid duplication and ensure that all the field definitions stay consistent. I do worry that we might not notice when we are giving different fields the same names even when they don't really have the same meaning. But also there are overly generic field names in the current system that should really be differentiated somehow.

I'm apprehensive about starting to do things differently but only with new fields, since it feels like added complexity / internal inconsistency, where we're doing the same thing in multiple different ways simultaneously. But I see that One Giant Namespace doesn't seem like it can scale gracefully to thousands of named columns.

src/pudl/transform/eia860.py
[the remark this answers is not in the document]
zaneselvans: I think this is going to be a pervasive issue with the data validation style asset checks -- we may have very different expectations about column contents if they contain one year or all the years of data.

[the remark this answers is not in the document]
e-belfer: ['annual_average_consumption_rate_gallons_per_minute', 'annual_average_discharge_rate_gallons_per_minute', 'annual_average_withdrawal_rate_gallons_per_minute', 'annual_maximum_intake_summer_temperature_fahrenheit', 'annual_maximum_intake_winter_temperature_fahrenheit', 'annual_maximum_outlet_summer_temperature_fahrenheit', 'annual_maximum_outlet_winter_temperature_fahrenheit', 'annual_total_chlorine_lbs'] 
are all the columns that fail on the fast ETL.

I wonder if we can use job ID or the years of the run to fine-tune which validation tests run for what, or define 'full ETL only' checks like our current validation tests. Possibly @bendnorman has thoughts here. But this is already a 1k line PR and I'm wondering if we just want to scope down here and leave this as a to-do in a separate issue.

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
jdangerx: > I think the main motivation that we had when we set up the One Giant Namespace was avoiding redefining the same columns in many different places

That was my understanding also, and I think the opposite convention will let us do this while also giving us the benefits of explicit references, namespacing, etc.

> I do worry that we might not notice when we are giving different fields the same names even when they don't really have the same meaning.

Yeah, Ella and I talked a bit about this issue - if a column name is the same across multiple tables, users might expect it to mean the same thing, even if they don't. Sort of like when you shadow a variable. We could run a check on pre-commit that checks for "shadowed" field names. In our schema definitions we could allow a "shadow_ok" flag that says, effectively, "Yeah, I know this shares the same name as something else and I don't care. This is both (a) different from the other thing and (b) not different enough that a different name makes sense"

> I'm apprehensive about starting to do things differently but only with new fields, since it feels like added complexity / internal inconsistency, where we're doing the same thing in multiple different ways simultaneously. But I see that One Giant Namespace doesn't seem like it can scale gracefully to thousands of named columns.

I think "migrating to a new system" here has to be done in one fell swoop, but "preparing to migrate to such a new system" doesn't. That being said, I think we can write a few scripts, re-do a medium of the resource + package logic, and get converted over to the new system without *too* much hassle... I've been meaning to write a little project pitch for that for a while.

src/pudl/transform/eia860.py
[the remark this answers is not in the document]
bendnorman: I bet we can access the dagster `context` object within the asset check and grab the etl configuration or job id. Let me check.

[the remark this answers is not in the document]
bendnorman: Ok! It looks like we can access `context` inside of an `@asset_check` (I think all Dagster concepts are just `@op`'s under the hood). With `context`, we can access the `dataset_settings` resource and the current `job_name`, which we could use to toggle between different sets of checks for a given asset and set of years.

However, IDK if `context.job_name` is the most reliable thing to use for determining what asset checks should be run. I think the name can get lost based on how you execute an asset. For example, while testing this out I executed a single asset in the `etl_fast` job and the `job_name` was "__ASSET_JOB". We're probably better off using the `dataset_settings` to toggle between different sets of checks for an asset.

For assets with metadata, we could probably store multiple sets of expected values given a set of years. I'm not sure how we would do this for assets without metadata though.

@jdangerx do you have a vision for storing/organizing asset check metadata?

e-belfer:
@jdangerx In addition to resolving this "where should our fields live" question, can you add the `dagster instance migrate` instructions (for minor version updates) somewhere to our docs? Otherwise I think this PR should be ready to merge today.

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
e-belfer: I'm worried about a lag between "preparing for the transition" and "having funding and time to do the transition" that looks like a mixed system for years, where we're using the group name spaces to mean two slightly different things and our docs don't clearly reflect what our practices are. Without the validation tests that make sure we don't have unexpected "shadowed" variables, I'm also hesitant to start doing this - plus, this PR is already so big that I don't want to start adding more metadata work in here, this feels like scope creep. So I'd lean towards preferring to stick with our current system for now, and moving these back to One Giant Namespace.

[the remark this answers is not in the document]
jdangerx: Yeah, I think that lag is certainly annoying/frustrating. I think that it's easier to combine namespaces than to split them, because the information of "where this field *would* belong" gets lost in the combination, but I'm OK with pushing things into One Giant Namespace (OGN).

Partially because so many things are already in the OGN that whatever we do to deal with it will have to figure out some way to split the namespace up semi-automatically anyway - so what's a few more fields to migrate in that world?

jdangerx:
By actually adding these schemas to our RESOURCE_METADATA & writing them out to DB, I hooked them up to a lot of data quality checks I hadn't before. Some changes:

- the cooling system types are actually *not* reported as codes, they're reported as human-facing that are roughly equivalent to the code descriptions (but not completely!)
- the EIA cooling system ID can be NA in rare cases - fill those to the string "NA" so that we can still use it as a primary key.
- there is one record that's duplicated in EIA923 - deduplicate that
- fields that EIA says are "whole fahrenheit degrees" are actually floats 🙃 

There still seems to be a doc build issue, which I should look into.

src/pudl/metadata/resources/eia.py
```diff
@@ -1046,6 +1058,59 @@
...
+    },
+    "core_eia__codes_cooling_system_types": {
+        "description": "A coding table for cooling system types.",
+        "schema": {
+            "fields": ["code", "label", "description"],
+            "primary_key": ["code"],
+            "foreign_key_rules": {
+                "fields": [
+                    [
+                        "cooling_type",
```
jdangerx, line 1100: Oops, I should make these fields also their own singleton lists, I suppose.

Makefile
jdangerx, line 1: This just defaults the `make docs-build` output to verbose, which gives a lot more information about the actual Python errors when things go wrong.

docs/templates/data_source_parent.rst.jinja
```diff
@@ -38,7 +38,7 @@ the names and descriptions of each of its fields.
    * - Data Dictionary
      - Browse Online
    {%- for resource in source_resources %}
-   * - :ref:`{{ resource.name }}`
+   * - :ref:`{{ resource.sphinx_ref_name }}`
```
jdangerx, line 41: I ran into a docs build error where Sphinx was unhappy about leading underscores in ref labels.

There were two ways around this:
* don't include the intermediate resources in this list
* figure out how to make legal ref labels with leading underscores

I figured, we went to all this effort to make these tables, we shouldn't hide them - and we already had a somewhat-working way of making legal ref labels (adding the `i` as a separator). So I went for the second option.

migrations/versions/d245fb7fb73e_add_cooling_equipment_codes_and_tables.py
jdangerx, line 1: I had a whole bunch of little migrations that I combined into this Big One. It adds the two cooling equipment data tables themselves, as well as several coding tables.

src/pudl/etl/__init__.py
```diff
@@ -108,6 +112,9 @@
     ),
 )
 
+asset_check_modules = [pudl.transform.eia860, pudl.transform.eia923]
```
jdangerx, line 115: In the future we probably want to do something like "define a list of all modules we would want to get asset checks *or* assets from, and then initialize `default_asset_checks` and `default_assets` from that single list." But. That is out of scope of this PR.

src/pudl/metadata/classes.py
```diff
@@ -1292,6 +1291,22 @@ class Resource(PudlMeta):
         "contributors", "keywords", "licenses", "sources", fn=_check_unique
     )
 
+    @property
+    def sphinx_ref_name(self):
```
jdangerx, line 1295: This basically moves the logic of the `format_resource_name_cross_ref` jinja filter to `Resource`.

src/pudl/metadata/codes.py
```diff
@@ -2318,6 +2318,212 @@
...
+                ),
+                (
+                    "OT",
+                    "other",
+                    "Other",
+                ),
+            ],
+        ),
+        "code_fixes": {
+            "RECIRCULATING WITH INDUCED DRAFT COOLING TOWER": "RI",
```
jdangerx, line 2510: EIA 923 has the long versions of these codes in the data files themselves.

They're not *exactly* the same as the descriptions above, and also have some whitespace/casing issues, so I canonicalize them to uppercase in the transform and use this `code_fixes` dictionary to map them correctly.

jdangerx (commented):
Oy, this got really big, but I hope these comments help guide this 🤞 final review.

Overall structure:

* EIA-923 cooling systems information + EIA860 cooling equipment had some unit conversions, spot fixes, and deduplication
* defined resource schemas + fields for both CSI + CE; this entailed some renaming of raw data fields
* defined a bunch of new codes (cooling type, water source type, tower type, etc) and had to do some remapping for CSI
* added asset check infrastructure + asset checks
* did some refactoring of how we make sphinx refs for resources in docs - this lets us make refs to `_*` resources
* migration

src/pudl/metadata/codes.py
```diff
@@ -2318,6 +2318,212 @@
...
+                ),
+                (
+                    "RC",
+                    "recirculating_pond",
+                    "Recirculating with cooling pond(s) or canal(s)",
+                ),
+                (
+                    "RF",
+                    "recirculating_forced_draft",
+                    "Recirculating with cooling pond(s) or canal(s)",
```
e-belfer, line 2485: Is this description accidentally copied from RC?

src/pudl/metadata/fields.py
```diff
@@ -882,6 +978,22 @@
             "different."
         ),
     },
+    "emission_control_id_eia": {
+        "type": "string",
+        "description": (
+            "The emission control ID used to collect so2, nox, particulate, "
```
e-belfer, line 984: Nit: I think this didn't even come from your branch, but it should be SO2 and NOx

src/pudl/metadata/fields.py
```diff
@@ -882,6 +978,22 @@
...
+            "with emissions_control_type as it's not guaranteed to be unique."
+        ),
+    },
+    "emission_control_id_pudl": {
+        "type": "number",
+        "description": "A PUDL-generated ID used to distinguish emission control units in the same report year and plant id. This ID should not be used to track units over time or between plants.",
+    },
+    "emission_control_id_type": {
+        "type": "string",
+        "description": "The type of emissions control id: so2, nox, particulate, or mercury.",
```
e-belfer, line 995: Same here

src/pudl/metadata/fields.py
```diff
@@ -1874,6 +2009,80 @@
             "min, equal to or less than 5 min, or some other way."
         ),
     },
+    "monthly_average_consumption_rate_gallons_per_minute": {
+        "description": "Monthly average consumption rate of cooling water",
+        "type": "number",
+        "unit": "gpm",
+    },
+    "monthly_average_discharge_rate_gallons_per_minute": {
+        "description": "Monthly average discharge rate of cooling water - available 2010+",
```
e-belfer, line 2018: I'd drop the 'available 2010+' as this is true for a bunch of other columns but only reported here, no?

src/pudl/metadata/fields.py
```diff
@@ -1874,6 +2009,80 @@
...
+        ),
+        "type": "number",
+        "unit": "lb",
+    },
+    "monthly_total_consumption_volume_gallons": {
+        "description": "Monthly volume of water consumed at consumption point (accurate to 0.1 million gal)",
+        "type": "number",
+        "unit": "gal",
+    },
+    "monthly_total_cooling_hours_in_service": {
```
e-belfer, line 2066: We use "hr" elsewhere and should stick to that for consistency

src/pudl/metadata/resources/eia860.py
```diff
@@ -642,6 +642,68 @@
...
     },
+    "_core_eia860__cooling_equipment": {
+        "description": (
+            "Information about cooling equipment at generation facilities, "
+            "from EIA-860 Schedule 6D. Currently in an intermediate state, "
+            "use at your own risk."
+        ),
+        "schema": {
+            "fields": [
+                "plant_id_eia",
```
e-belfer, line 653: Nice to have: organize columns roughly by content - e.g., put county and state next to one another, utility ID and name together, etc. `report_date` should be first to be consistent with other data sets.

src/pudl/metadata/resources/eia923.py
```diff
@@ -3,6 +3,15 @@
...
 
 TABLE_DESCRIPTIONS: dict[str, str] = {
+    "_core_eia923__cooling_system_information": (
+        """EIA-923 Cooling System Information, from EIA-923 Schedule 8D.
+
+Reports monthly information about cooling systems at generation facilities,
+mainly water volumes and temperatures. In 2008 and 2009, EIA only reports
+annual averages, but in later years all data is monthly.
+
+Currently in an intermediate state, use at your own risk."""
```
e-belfer, line 13: See the warning I wrote up in #3403, I think it's possibly clearer than "use at your own risk" :)

src/pudl/metadata/resources/eia923.py
```diff
@@ -654,6 +663,50 @@
         "sources": ["eia923"],
         "etl_group": "eia923",
     },
+    "_core_eia923__cooling_system_information": {
+        "description": TABLE_DESCRIPTIONS["_core_eia923__cooling_system_information"],
+        "schema": {
+            "fields": [
+                "plant_id_eia",
```
e-belfer, line 670: See above comment re order of columns

src/pudl/transform/eia860.py
```diff
@@ -1021,3 +1027,124 @@ def _core_eia860__boiler_stack_flue(
...
+    in MW. This is corroborated by the values for these plants matching up with
+    the stated MW values in later years. Additionally, the PDF form for those
+    years indicates that the value should be in MW, and KWh isn't even a power
+    measurement.
+
+    In 2009, we have two incorrectly entered ``cooling_type`` values of ``HR``,
+    for utility ID 14328, plant IDs 56532/56476, and cooling ID ACC1. This
+    corresponds to the Colusa and Gateway generating stations run by PG&E. In
+    all later years, these cooling facilities are marked as ``DC``, or "dry
+    cooling"; however, ``HR`` follows the coding pattern for hybrid systems. As
```
e-belfer, line 1054: Can you clarify what this means, HR follows the coding pattern for hybrid systems?

src/pudl/transform/eia860.py
```diff
@@ -1021,3 +1027,124 @@ def _core_eia860__boiler_stack_flue(
...
+    ] = 1971
+    ce_df.loc[ce_df["pond_operating_year"] == 0, "pond_operating_year"] = np.nan
+    ce_df.loc[ce_df["tower_operating_year"] == 974, "tower_operating_year"] = 1974
+    ce_df = (
+        ce_df.pipe(pudl.helpers.month_year_to_date)
+        .pipe(pudl.helpers.convert_to_date)
+        .rename(columns={"operating_date": "cooling_system_operating_date"})
+    )
+
+    # There's one row which has an NA cooling_id_eia, which we make "NA" to
```
e-belfer, line 1075: Would it be more consistent to convert this to "PLANT", or does the plant have more than one cooling system associated with it?

src/pudl/transform/eia923.py
```diff
@@ -1223,3 +1238,155 @@ def _core_eia923__fuel_receipts_costs(
...
+    so we interpret "0.1 cubic feet per second" to mean "cubic feet per
+    second, with precision of 0.1 cfs."
+    """
+    csi_df = raw_eia923__cooling_system_information
+    csi_df = csi_df.pipe(pudl.helpers.fix_eia_na).pipe(pudl.helpers.convert_to_date)
+
+    # cooling_id_eia is sometimes NA, but we also want to use it as a primary
+    # key. fortunately it's a string so we can just convert all NA values to
+    # "NA"
+    csi_df.cooling_id_eia = csi_df.cooling_id_eia.fillna("NA")
```
e-belfer, line 1269: Same question as with the 860 data re: NA vs PLANT.

e-belfer (changes requested):
:heavy_check_mark: `make docs-build` works and the tables show up in our docs as expected

:heavy_check_mark: assets materialize and look as expected in terms of order, encoding, and asset checks passing

All that's left are tiny nits re the descriptions, units and column ordering. Once those are fixed, feel free to dismiss this review and merge the PR in, or just bump me for approval.

src/pudl/metadata/fields.py
jdangerx: Hmm, I see `operating_time_hours` but I don't see `_hr` anywhere in the field names; in codes I see a lot more `_hour` than i see `_hr`.

e-belfer: I meant in the units of `operating_time_hours`, I think the unit listed is `hr`

jdangerx: Ahhh good eye thanks!

e-belfer (approved):
Presuming tests pass, merge it!

jdangerx:
After much debugging of the `_core_eia860__cooling_equipment` ↔️ `core_eia860__scd_plants` foreign key which seemed to be fine, I realized the actual FK error in the tests was with the *923* table instead, and I had run into https://github.com/catalyst-cooperative/pudl/issues/1196. After fixing the foreign key, my local checks all worked and I hope this is reflected in the GHA run!

Issue #3404, opened 2024-02-16
determine why rate base assets are smaller than liabilities in leafy exploded table


cmgosnell:
## possible weirdness?
is it the sub-totals not having corrections?
is it that the tags aren't full propagating?
should there be sign differences within the liabilities or asset tables?
did we do soemthing wrong/weird with the plant in service table calcs & sign changes?
Are we canceling out the calced value if the reported value is null but we have leafy/child data?

## ways to determine what's going wrong?
- look at a few utilities?
- check if the total liabilites (root) is the sum of the leafy liabilities? NO!
- check if *most* of the correction records are ending up in the outputs? NO! & they have different dimension values. looks like when we assigned dim="total" this isn't propagating into the correction records...

cmgosnell:
Okay so the vast majority of the problem here was [just from one hard-coded old table name](https://github.com/catalyst-cooperative/pudl/pull/3450/files#r1516781247)

![image](https://github.com/catalyst-cooperative/pudl/assets/25487998/c09c52c4-79b4-4777-9c63-5a7c807328f0)
![image](https://github.com/catalyst-cooperative/pudl/assets/25487998/ce544085-7eb3-4590-a5ab-eeaa839581aa)


There is still the 2021 problem but i'm addressing that over in #3309

Issue #3072, opened 2023-11-21
Make a transformer to add a column w/ a standard value


cmgosnell:
```[tasklist]
### Tasks
- [x] build standard transformer in `pudl.transform.classes` that adds colums(s) with standard value(s)
- [x] add this transform step into `pudl.transform.ferc1.Ferc1AbstractTableTransformer.transform_main`
- [x] add a metadata transform step as well
- [x] add a calculations table transform step as well
- [x] ensure this works for the `add_corrections` and corrections in table metadata & calculations
- [x] convert all of the current assign-ing of (mostly) single value columns into a transform params
```

this will be a bit of a weird transformer because one parameter will need to effect outcomes in many places across the transform. There will be a normal looking transform `TransformParams` & transform function & method that actually does the assignment of a new column. But this transform param will also need to be access in the methods/functions that muck with the table metadata and calculations.

cmgosnell:
Okay in service of #3404 i learned that the calculation components are not always getting the dimension tags when we have applied them uniformly across all tables.

A lil example:

```python
from pudl.output.ferc1 import NodeId
from pudl.etl import defs
_core_ferc1_xbrl__calculation_components = defs.load_asset_value("_core_ferc1_xbrl__calculation_components")
calc_idx = [c for c in list(NodeId._fields) if c != "table_name"] # remove tbl name bc long
parent_idx = [f"{c}_parent" for c in calc_idx]
_core_ferc1_xbrl__calculation_components[_core_ferc1_xbrl__calculation_components.xbrl_factoid_parent == "nuclear_fuel_net"][
      parent_idx + calc_idx
]
```

![Image](https://github.com/catalyst-cooperative/pudl/assets/25487998/41884ab6-da3c-4322-a04e-3800a8270872)

I believe this dimensions are being dropped and/or not being assigned in `add_calculation_corrections`. But in general I've had a hard time seeing where we've added these uniform dimensions across the board and i think fixing this `_correction` problem will be much easier to do with a parameterized transform for this adding of uniform dimension columns.

cmgosnell:
this was closed via #3409

Issue #3125, opened 2023-12-05
fix contact future warning


cmgosnell:
# the warning to squash
```python
FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
```
From a quick search it looks like this warning is popping up *122 times*
## one possible solution
i re-remembered to deal with this future warning that i believe is showing up in lots of places iirc

this warning only happens bc there is a column in the dfs being concat-ed where at least one of the dfs contains all null values and the dtype for that column are not the same in the dfs being concat-ed. my solution (which i of course found on some stack overflow something and now can’t find is:
```python
pd.concat([df1, df2.astype(df1.dtypes)])
```
(if the columns in df1 and df2 are not the same you need to make sure the columns your feeding to `astype` all show up in `df2`, but that's pretty straightforward)

we could add a helper concat function to do this in many places if we wanted. that feels a little ridiculous though.

## another solution
drop all null columns before concating. this may result in some unintended consequences though.

PR #3409, opened 2024-02-19
Parametermize adding a column in the FERC1 transform & ensure `_correction` records end up in the calculation compoent table

cmgosnell:
# Overview

Closes #[3072](https://github.com/catalyst-cooperative/pudl/issues/3072). Working on [#3404](https://github.com/catalyst-cooperative/pudl/issues/3404).

- it was previously pretty hard to keep track of when we added dimension columns into the ferc1 transform like `utility_type="electric"
- we weren't getting very many `_correction` records into the `detailed` and rate base tables because the correction records in the `_core_ferc1_xbrl__calculation_components` table were not showing up with the same dimensions

What did you change?
- see task list in #3072 for parameterization
- added in `_correction` records to the `_core_ferc1__table_dimensions` table even if there aren't correction records in the data so we can have them in the `_core_ferc1_xbrl__calculation_components` with the same exact dimensions as their parents
- ensure the correction records don't get grabbed as total -> sub-total records
- moved the big `explosion_args` into a global variable. this is entirely for notebook-able testing. it's been a pain ergonomically for a while to have that dictionary inside the factory instead of accessible.
- light changes to the `forest_as_table`: keep the columns with fully null dimensions so that it is much easier to set indexes on a particular layer and select records using the nodes (which have all dimensions even the fully null ones) 

# Testing

```[tasklist]
# To-do list
```


src/pudl/transform/ferc1.py
```diff
@@ -6175,6 +6122,20 @@ def _core_ferc1__table_dimensions(**kwargs) -> pd.DataFrame:
...
+    dimensions = pd.concat(
+        [
+            dimensions[non_correction_mask],
+            (
+                dimensions[non_correction_mask].assign(
+                    xbrl_factoid=lambda x: x.xbrl_factoid + "_correction"
+                )
+            ),
+        ]
+    )
```
cmgosnell, line 6138: move into the `infer_intra_factoid_totals`

src/pudl/output/ferc1.py
```diff
@@ -1747,6 +1749,18 @@ def calc_idx(self: Self) -> list[str]:
         """Primary key columns for calculations in this explosion."""
         return [col for col in list(NodeId._fields) if col in self.exploded_pks]
 
+    def prep_table_to_explode(
+        self: Self, table_name: str, table_df: pd.DataFrame
+    ) -> pd.DataFrame:
+        """Prep an core input table for explosion."""
```
e-belfer, line 1755: Can you make this slightly more descriptive? "Assign table name and rename factoid column in preparation for explosion" or something similar?

src/pudl/output/ferc1.py
```diff
@@ -2885,7 +2889,10 @@ def _propagate_tags_to_corrections(annotated_forest: nx.DiGraph) -> nx.DiGraph:
         # for every correction node, we assume that that nodes parent tags can apply
         parents = list(annotated_forest.predecessors(correction_node))
         # all correction records shoul have a parent and only one
-        assert len(parents) == 1
+        if len(parents) != 1:
+            raise AssertionError(
+                f"Found more than one parent node for {correction_node=}\n{parents=}"
```
e-belfer, line 2894: Great, this seems super useful

src/pudl/transform/ferc1.py
```diff
@@ -751,6 +751,35 @@ def assign_quarterly_data_to_yearly_dbf(
...
 
 
+class AddColumnWithUniformValue(TransformParams):
+    """Parameters for adding a column to a table with a signle value."""
+
+    column_value: Any = None
+    is_dimension: bool = False
+
+
+class AddColumsWithUniformValue(TransformParams):
```
e-belfer, line 761: Should be `Columns` not `Colums` I think? As discussed making `Value` into `Values` might also help visually differentiate these

src/pudl/transform/ferc1.py
```diff
@@ -751,6 +751,35 @@ def assign_quarterly_data_to_yearly_dbf(
...
+    """Parameters for adding columns to a table with a signle value."""
```
e-belfer, line 762: Should be single? Here and above.

src/pudl/transform/ferc1.py
```diff
@@ -1520,7 +1549,9 @@ def add_corrections(
         f"{num_records} original records."
     )
 
-    return pd.concat([calculated_df, corrections], axis="index")
+    return pd.concat(
+        [calculated_df.convert_dtypes(), corrections.convert_dtypes()], axis="index"
```
e-belfer, line 1553: Is there a reason to do dtypes conversions here and not further upstream?

src/pudl/transform/ferc1.py
```diff
@@ -1880,6 +1929,7 @@ def transform_main(self, df: pd.DataFrame) -> pd.DataFrame:
                 .encode
             )
             .pipe(self.merge_xbrl_metadata)
+            .pipe(self.add_columns_with_uniform_value)
```
e-belfer, line 1932: So much neater, thanks for this!

src/pudl/transform/ferc1.py
```diff
@@ -2161,11 +2215,6 @@ def add_calculation_corrections(
                 weight=1,
             )
         )
-        # for every calc component, also make the parent-only version
```
e-belfer, line 2164: Can you explain why this gets dropped?

src/pudl/transform/ferc1.py
```diff
@@ -6475,6 +6422,22 @@ def make_calculation_dimensions_explicit(
             the parental dimensions or the child dimensions.
     """
     logger.info(f"Adding {dimensions=} into calculation component table.")
+    # even though we don't actually get correction records for all of the
+    # factiods, we are still going to force them to exist here so all of the
```
e-belfer, line 6426: Should this be for all factoids, or all calculated factoids? If factoids only ever exist as components I'm not sure why we would expect a correction record for them?

test/unit/transform/ferc1_test.py
```diff
@@ -415,6 +420,65 @@ def test_unstack_balances_to_report_year_instant_xbrl():
...
+
+    Will ``dimension_columns`` return only one column if the dimension column from
+    :class:`AddColumsWithUniformValue` and :class:`ReconcileTableCalculations` are the same?
+    """
+    add_columns_with_uniform_value = AddColumsWithUniformValue(
+        columns_to_add={
+            "added_dim": AddColumnWithUniformValue(
+                column_value="i'm dim", is_dimension=True
+            ),
+            "not_a_dim": AddColumnWithUniformValue(
```
e-belfer, line 456: Do you want to assert that this column got added to the dataframe, but not as a dimension column?

e-belfer (changes requested):
I have a few clarifying questions and some minor nits. Otherwise mostly looks good to go and everything generates as expected.

src/pudl/transform/ferc1.py
cmgosnell: lol good catch

src/pudl/transform/ferc1.py
cmgosnell: this has to do with #3125 but i actually didn't even do this right

src/pudl/transform/ferc1.py
cmgosnell: lol again good catch!

src/pudl/transform/ferc1.py
cmgosnell: this is a very good point. this is part of why i wanted to move this into this function instead of do this in `_core_ferc1__table_dimensions`. because these `_correction` records are only ever generated here & they are only used to as the left df in a left merge down below when defining `calc_comps_w_implied_dims`, if they don't show up `null_dim` (which is made out of the calculation components) they will just be correctly lost in the merge.

but that feels pretty opaque. so I'll add another part of the mask to grab only table dim factoids that show up as parents (== calculated values) in the calculation components

test/unit/transform/ferc1_test.py
cmgosnell: in here i'm testing the `dimension_columns` parameter class but not i'm not actually doing the transform at all.

in the `test_add_columns_with_uniform_value` above, we're actually doing the transform. in that context i didn't set the `is_dimension` bool so by default they are both `False` and are being added. I could make that explicit above and test it up there or add a new test that does that mixed `is_dimension` transform.

src/pudl/transform/ferc1.py
cmgosnell: hm this is less straightforward than i had thought. because this function should reallly be called `make_xbrl_factoid_dimensions_explicit` because this is being used to make dimension explict for dfs that are not the calc comps table. I think it is possible to only add corrections for calculated values but it will be a little complicated. it seems simpler to keep the existing behavior and just explain the left merge drop here.

test/unit/transform/ferc1_test.py
cmgosnell: okay i added a second step in `test_add_columns_with_uniform_value`

e-belfer (approved):
Thanks for addressing my comments.

PR #3434, opened 2024-02-27
Add RMI beta access to builds.catalyst.coop

jdangerx:
# Overview

We have some beta Parquet outputs that we'd like RMI to be able to access on a provisional basis.

The simplest way to do this was to set up a service account for that access, so we made a service account with terraform.

# Testing

Ran `tf plan`, saw that it added the right things, and then ran `tf apply`.

```[tasklist]
# To-do list
```


terraform/main.tf
```diff
@@ -68,3 +68,16 @@ module "gh_oidc" {
     }
   }
 }
+
+resource "google_service_account" "service_account" {
```
jdangerx, line 72: I didn't add any private key credentials here, even though they *can* be managed via terraform, because the creds would be stored in plaintext in the terraform state. Seemed nicer to just handle that separately.

terraform/main.tf
```diff
@@ -68,3 +68,16 @@ module "gh_oidc" {
...
+  account_id   = "rmi-beta-access"
+  display_name = "RMI Beta Access"
+}
+
+resource "google_storage_bucket_iam_binding" "binding" {
+  bucket = "builds.catalyst.coop"
+  role   = "roles/storage.objectViewer"
+  members = [
+    "serviceAccount:rmi-beta-access@catalyst-cooperative-pudl.iam.gserviceaccount.com",
```
jdangerx, line 81: Once we run through the auth with RMI folks, we should verify that the SA actually has the correct roles before merging :)

terraform/main.tf
```diff
@@ -68,3 +68,16 @@ module "gh_oidc" {
     }
   }
 }
+
+resource "google_service_account" "service_account" {
+  account_id   = "rmi-beta-access"
+  display_name = "RMI Beta Access"
+}
+
+resource "google_storage_bucket_iam_binding" "binding" {
```
jdangerx, line 77: Chose `binding` instead of `policy` so we don't accidentally lock the rest of us out of the `builds.catalyst.coop` bucket. Though eventually we should pull those permissions *into* terraform too, seemed out of scope for this small RMI request.

zaneselvans:
Do they really want to be looking at the nightly build outputs, rather than the data that's published to `gs://parquet.catalyst.coop/nightly` after the builds succeed?

jdangerx:
Ah good call. Yeah, `parquet.catalyst.coop` makes a lot more sense, I'll swap that around.

zaneselvans:
The stable / versioned parquet outputs are also up and running now, in case those are more appropriate

```console
 gsutil ls -lh gs://parquet.catalyst.coop
    gs://parquet.catalyst.coop/nightly/
    gs://parquet.catalyst.coop/stable/
    gs://parquet.catalyst.coop/v2024.2.6/
```

zaneselvans (approved):
I have no familiarity with the Terraform stuff so... not sure I can provide any useful review here.

If the bucket and the service account already exist, and the service account already has access to the bucket, what does this setup do?  When does it get deployed / applied? Does it basically overwrite whatever settings have been manually input?

Note that I also created a `private-beta@catalyst.coop` permissions group and added Alex's RMI and Gmail addresses to it. Do we think we'll want to go with per-org or per-user service accounts and credentials instead? What should become of that group?

Could we use terraform to apply an object lock to the versioned releases that we've got up on S3 so we can't accidentally delete them?

PR #3442, opened 2024-03-05
Add BA codes and EIA sector IDs to EIA-860M changelog table

### LINKED ISSUE #3437: include `balancing_authority_code_eia` in the `core_eia860m__changelog_generators` table
I've been exploring the `core_eia860m__changelog_generators` table since we'd like to use it as our source for 860m data but our current uses of 860m data require BA codes. I like the changelog approach here (I'll admit to originally be skeptical), but I'd like to see it include all the fields in the source, which would mean adding back BA codes and sectors.

Include balancing authority codes (and sectors) as reported in `core_eia860m__changelog_generators` so the table records the changes in these fields, to the extent they exist, like it records the changes in the values of other columns. 

**Describe alternatives you've considered**
We could merge these columns on from the plants table but that is not equivalent because the source is different, and IIRC, the values in the plants table have been through additional cleaning, harvesting, and conforming processes.

zaneselvans:
# Overview

* Add type hints and revise docstring for eia860m extract function
* Rename EIA consolidated sector appropriate in eia860m column mappings
* Add balancing_authority_code_eia and sector_id_eia to the core_eia860m__changelog_generators table. This required translating the categorical strings used in the EIA-860M to identify sectors to the canonical IDs that are associated with the sectors, defined in core_eia__codes_sector_consolidated.
* Add an alembic schema migration to acommodate the new fields.
* Re-lock conda depdencies to avoid yanked pytest 8.1.0 version.

Closes #3437

# Testing

* Ran the full ETL locally and checked FKs
* Ran `make pytest-coverage` locally

```[tasklist]
# To-do list
- [x] For significant ETL changes, ensure the full ETL runs locally (all good except for FKs)
- [x] Ensure docs build, unit & integration tests, and test coverage pass locally (all good except FKs)
- [x] Clean up BA codes.
```


src/pudl/extract/eia860m.py
```diff
@@ -64,21 +65,20 @@ def get_dtypes(page, **partition):
         }
 
 
-def append_eia860m(eia860_raw_dfs, eia860m_raw_dfs):
+def append_eia860m(
+    eia860_raw_dfs: dict[str, pd.DataFrame], eia860m_raw_dfs: dict[str, pd.DataFrame]
+) -> dict[str, pd.DataFrame]:
```
zaneselvans, line 70: No substantive changes in this module. Just type hints / docstrings.

src/pudl/package_data/eia860m/column_maps/generator_existing.csv
```diff
@@ -15,7 +15,7 @@ winter_capacity_mw,,,,,,,,,,,,,net_winter_capacity_mw,net_winter_capacity_mw,net
...
+sector_name_eia,sector_name,sector_name,sector_name,sector_name,sector_name,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector,sector
```
zaneselvans, line 18: Switched to using the canonical name that we've defined elsewhere.

src/pudl/transform/eia860m.py
```diff
@@ -56,8 +58,27 @@ def core_eia860m__changelog_generators(
...
+                "Electric Utility": 1,
+                "IPP Non-CHP": 2,
+                "IPP CHP": 3,
+                "Commercial Non-CHP": 4,
+                "Commercial CHP": 5,
+                "Industrial Non-CHP": 6,
+                "Industrial CHP": 7,
+            }
+        )
+    )
```
zaneselvans, line 78: I added this translation early, because of the column removal that happens below.

We don't store a mapping of sector names to sector codes anywhere else, because the names are not consistently used across tables, and everywhere else the code is directly reported.  Really we should probably replace the sector names everywhere else with our human-readable labels (or just drop them from the core tables, and merge in the human readable labels for output tables)

src/pudl/transform/eia860m.py
```diff
@@ -69,8 +90,10 @@ def core_eia860m__changelog_generators(
     # there is one plant/gen that has duplicate values
     gens_idx = ["plant_id_eia", "generator_id", "report_date"]
     dupe_mask = (eia860m_all.plant_id_eia == 56032) & (eia860m_all.generator_id == "1")
-    deduped = eia860m_all[dupe_mask].drop_duplicates(subset=gens_idx, keep="first")
-    without_known_dupes = eia860m_all[~dupe_mask]
+    deduped: pd.DataFrame = eia860m_all[dupe_mask].drop_duplicates(
+        subset=gens_idx, keep="first"
+    )
+    without_known_dupes: pd.DataFrame = eia860m_all[~dupe_mask]
```
zaneselvans, line 96: I was getting some weird linting error about subsequent code being unreachable, which adding these type hints fixed. Not sure what was going on.

zaneselvans:
It turns out that the BA codes in this table are a mess. Cleaning them up now.

src/pudl/package_data/eia860m/column_maps/generator_existing.csv
cmgosnell: good good! This also needs to be updated in the other two `generator_*` column maps/

src/pudl/transform/eia860m.py
cmgosnell: hm that's really weird. I don't love that tbh

environments/conda-linux-64.lock.yml
cmgosnell, line 1: why are these files showing up in here?

docs/release_notes.rst
```diff
@@ -15,6 +15,12 @@ Data Cleaning
   :issue:`3340` and PR :pr:`3419`. This change also fixed a bug that was preventing
   other columns harvested with a special process from being saved.
 
+Schema Changes
+^^^^^^^^^^^^^^
+* Added ``balancing_authority_code_eia`` and ``sector_id_eia`` into the
+  :ref:`core_eia860m__changelog_generators` table. This required some cleanup of the BA
+  codes, which were not standardized. See issue :issue:`3437` and PR :pr:`3442`.
```
cmgosnell, line 22: do you mean that the BA codes in 860m were not standardized in this table? If so i'm not sure that addition is super helpful in release notes because its not like they existed in the table beforehand. totally a nit and not a hill i will die on.

cmgosnell (approved):
this looks good to me. nits but nothing is blocking.

src/pudl/transform/eia860m.py
cmgosnell: dang ya it would obviously be better to make an encoder table and use it across the board but that feels very oos for this pr so this seems fine.

cmgosnell (changes requested):
@zaneselvans  sorry i hit approve before bc i forgot there is one real request in here in the column maps.

src/pudl/package_data/eia860m/column_maps/generator_existing.csv
zaneselvans: Ooh, whooops! Good catch.

src/pudl/transform/eia860m.py
zaneselvans: I had to physically restrain myself...

docs/release_notes.rst
zaneselvans: I meant that the BA codes in the original EIA860M were not standardized. They were significantly messier than the BA codes reported elsewhere.  Like there were as many new funky codes in just this one table as we had collected from all the other tables.

src/pudl/package_data/eia860m/column_maps/generator_existing.csv
zaneselvans: Maybe this will improve coverage actually.

environments/conda-linux-64.lock.yml
zaneselvans: There was a problem with the previous lockfiles -- pytest v8.1.0 was in there, but it was yanked shortly after being released, and so it was impossible to install.

docs/release_notes.rst
zaneselvans: I'll clarify the note.

src/pudl/transform/eia860m.py
zaneselvans: Bug seems to have been fixed with some VS Code updates. Not sure what that was all about.

docs/release_notes.rst
cmgosnell: the clarification is nice! i still don't think this is a helpful release note (if we had been including this column but not encoding it and then transitioned to doing the cleanup then i think it would make more sense). but again not a hill i will die on.

src/pudl/transform/eia860m.py
cmgosnell: good jerb

cmgosnell (approved):

Issue #3309, opened 2024-01-29
why is basically nothing in the rate base table from `core_ferc1__yearly_depreciation_by_function_sched219`in 2021?


cmgosnell:
[see this comment that id-ed the problem](https://github.com/catalyst-cooperative/pudl/pull/3214#issuecomment-1912467242). track it down & fix it. presumably its from the og table transform.

![image](https://github.com/catalyst-cooperative/pudl/assets/25487998/7afb290e-4f26-4a76-857f-bb03be2fa55d)

```[tasklist]
### Tasks
- [x] add standard subtotal_corrections
- [ ] add unit test for multi-dimension corrections
- [ ] rn the merge validation fails in `calculate_values_from_components`: can we just parameterize the validation? or do we need to selectively drop calc component records?
- [ ] make sure total (parent) -> null subtotal (child) corrections get made
- [ ] make sure null calculated values and/or null reported values actually get corrections
- [ ] do we need to iteratively make corrections (i.e. correct leafs, calculate parents w/ corrections, calculate grandparents /w corrections)
- [ ] do we want to make corrections on two dimensions or just choose one? (probably just choose one)
```

e-belfer:
@cmgosnell:
Out of 205 unique entities in the raw XBRL instant data, there are only 4 utilities that report data in the leafy "accumulated provision for depreciation of X" columns of the XBRL instant data in 2021. The rest report data in what is ultimately the "total" plant function of the "accumulated depreciation" factoid. For some reason, the corrections from subtotal to total in the `plant_function` sub-dimension are not getting propagated to the rate base table.

To see the values reported in this table:

`out_ferc1__yearly_rate_base[(out_ferc1__yearly_rate_base.table_name == "core_ferc1__yearly_depreciation_by_function_sched219")&(out_ferc1__yearly_rate_base.report_year==2021)&(out_ferc1__yearly_rate_base.ending_balance.notnull())]`

These are the same utilities observed in the transformed table:
```
from dagster import AssetKey

from pudl.etl import defs

asset_key = "core_ferc1__yearly_depreciation_by_function_sched219"
df = defs.load_asset_value(AssetKey(asset_key))

df[(df.report_year==2021)&(df.plant_function!="total")&(df.ending_balance.notnull())]
```

cmgosnell:
[...]

PR #3450, opened 2024-03-07
Add total -> subtotal calculation correction & fix hard-coded plant-in-service table name

### LINKED ISSUE #3404: determine why rate base assets are smaller than liabilities in leafy exploded table

cmgosnell:
# Overview

Closes #3404 and closes on #3309.

### What did you change?
- add total -> sub-dimension corrections
- rename `subtotal` to `subdimension` to keep it consistent
- consolidated how we are reconciling table calcs via `reconcile_one_type_of_table_calculations`
- enable multiple dims in the sub-dimension calculation reconciliation: did this by adding `subdimension_merge_validation` to enable a `many_to_many` merge for more than one dimension column
- add unit tests to explode multi-dimension correction records
- enable reported parent w/ # -> null calculation component correction
   - this required a bunch of increases in the calculation metric checking badness
- explicitly instead of implicitly add the standard `_correction` and the new `_subdimension_correction` records to the calculation components table

#### for the `depreciation_by_function` table subtotal corrections are not in `_core_ferc1_xbrl__calculation_components`: SOLVED
- we are ending up with ~200 correction records almost all for 2021 in the core table which is exactly what we'd expect.
- but table has one factoid. and no inherent calculations with one factoid. the calculations are all total -> subtotals. rn there are total -> subtotal calculations happening for this table in its transform. 
- the way we compile the parts for the calc components table in each table transform is slightly different than how we do it in `_core_ferc1_xbrl__calculation_components` and we need some way to make sure the subtotal ->total calcs and either corrections end up in the final table:
  - option: standardize moar between the transforms and the core calc component table! this feels hard bc there are different needs/inputs into the transforms calc components and the core all table calc component
  - option: cache and reuse each transformed table-level calc component and then compile them together and deal with the inter-table calcs then.
  - option (preferred): be more explicit about adding corrections in the core calc components *at the end*. I think rn we are missing these corrections bc they are semi-haphazardly being added during `make_xbrl_factoid_dimensions_explicit`, but this table has so calc comps until `infer_intra_factoid_totals`
#### Does the current setup for many:many result in double counting in the rate base table??
- i looks like the answer here is no. I've checked several data points (grabbing a utility/year and checking through the process), but i'd like to add some output unit tests for it as well that are the follow up to the transform unit tests.
- Why? well because `infer_intra_factoid_totals` has no "total"s in the children.... that means when there are mutli-dimensions, the calculation for the total/total (parents) record is connected to sub-dims/sub-dims (children). there are total/sub-dim & sub-dim/total calculations, but those are not actually connected to the total/total calculation are are thus pruned from the tree.

## remaining todos:
```[tasklist]
 - [x] convert param notes to TODOS (I decided to convert them all to `# worst {thing}: {worst one}`, bc its not realllly a todo, its mostly a helpful lil "hey if your looking into this maybe start w/ this worst guy"/"why is 2021 consistently bad across the tables"
 - [x] check to see if there are null total/total records with non-null total/sub-dim records (& thus are we loosing data)
- [x] convert the one unit test into many tests
- [x] add unit test to ensure dropping the sub-dim/total calcs (re:double counting!)
 - [ ] make an issue about standardizing `row_type_xbrl` instead of assigning `is_calc`
```

# Testing

```[tasklist]
# To-do list
```


src/pudl/transform/ferc1.py
```diff
@@ -4593,7 +4644,7 @@ def convert_xbrl_metadata_json_to_df(
             {
                 "name": "electric_plant_in_service_and_completed_construction_not_classified_electric",
                 "weight": 1.0,
-                "source_tables": ["plant_in_service_ferc1"],
+                "source_tables": ["core_ferc1__yearly_plant_in_service_sched204"],
                 "utility_type": "electric",
             }
```
cmgosnell, line 4649: this old table name was resulting in the vast majority of the $s from this table not showing up in the rate base table

src/pudl/transform/ferc1.py
```diff
@@ -963,7 +966,7 @@ def reconcile_table_calculations(
         and ``rel_diff``.
     """
     # If we don't have this value, we aren't doing any calculation checking:
-    if params.column_to_check is None or calculation_components.empty:
+    if params.column_to_check is None:
```
cmgosnell, line 969: previously the `depreciation_by_function` table was entirely getting skipped over because it has no calc components prior to inferring the total to sub-dim calcs via `_add_intra_table_calculation_dimensions`

src/pudl/transform/ferc1.py
```diff
@@ -1054,32 +1105,6 @@ def _calculation_components_subtotal_calculations(
     return calc_comps_w_totals
 
 
-def _check_subtotal_calculations(
```
cmgosnell, line 1057: now both the "standard" calcs and the subtotal calcs are being checked via `reconcile_one_type_of_table_calculations`

src/pudl/transform/ferc1.py
```diff
@@ -1173,18 +1201,19 @@ def calculate_values_from_components(
             pd.merge(
                 calculation_components,
                 data,
-                validate="one_to_many",
+                validate=calc_to_data_merge_validation,
```
cmgosnell, line 1204: we were not able to test the `depreciation_by_function` table because it has two dimension columns and child calc components *necessarily & correctly* are part of more than one parent calculation.

src/pudl/transform/ferc1.py
```diff
@@ -1475,7 +1515,7 @@ def apply_metric(self: Self, df: pd.DataFrame) -> pd.Series:
         return (
             df[df.row_type_xbrl == "calculated_value"]
             .groupby(self.groupby_cols())
-            .apply(self.metric)
+            .apply(self.metric, include_groups=False)
```
cmgosnell, line 1518: this was to remove a future warning and preserve existing behavior

src/pudl/transform/ferc1.py
```diff
@@ -4546,7 +4623,7 @@ def convert_xbrl_metadata_json_to_df(
             }
             for new_fact in ["megawatt_hours_purchased", "purchased_mwh"]
         ]
-        new_facts = pd.DataFrame(facts_to_add).convert_dtypes()
+        new_facts = pd.DataFrame(facts_to_add).astype(tbl_meta.dtypes, errors="ignore")
```
cmgosnell, line 4626: these were to remove future warnings about concat-ing fully null columns

src/pudl/transform/ferc1.py
```diff
@@ -6428,26 +6510,8 @@ def make_xbrl_factoid_dimensions_explicit(
...
-    table_dimensions_ferc1 = pd.concat(
-        [
-            table_dimensions_ferc1[non_correction_mask],
-            (
-                table_dimensions_ferc1[non_correction_mask].assign(
-                    xbrl_factoid=lambda x: x.xbrl_factoid + "_correction"
-                )
-            ),
-        ]
-    )
```
cmgosnell, line 6449: this is now all happening in `add_calculation_component_corrections`

src/pudl/transform/ferc1.py
```diff
@@ -6677,6 +6739,55 @@ def infer_intra_factoid_totals(
     return calcs_with_totals
 
 
+def add_calculation_component_corrections(
```
cmgosnell, line 6742: me q: rn this is only being used to compile the big mutli-table calc component table. can i use this in the table transforms as well?

src/pudl/transform/params/ferc1.py
cmgosnell, line 1: many of the calculation metrics got worse because we are now including calculations when there is a null on either the calculated or reported side of the calculations.

test/unit/transform/ferc1_test.py
```diff
@@ -548,6 +556,207 @@ def test_calculate_values_from_components():
     )
 
 
+def test_many_to_many_subdimensions_calculate_values_from_components():
```
cmgosnell, line 559: there are effectively 4 different unit tests in here all using the same lil helper `get_reconcilded_data` to build and poop out the corrected calculations. This could easily be broken up into 4 unit tests. i kinda think it is easier to read all in one place bc it is 4 iterations on the same idea. but very open to rejiggering.

src/pudl/transform/params/ferc1.py
```diff
@@ -3551,11 +3573,16 @@
...
+            "subdimension_column": "plant_function",
             "group_metric_checks": {
                 "group_metric_tolerances": {
                     "ungrouped": {"null_calculated_value_frequency": 0.84},
                     "report_year": {"null_calculated_value_frequency": 1.0},
+                    "utility_id_ferc1": {
+                        "error_frequency": 0.23,
+                        "relative_error_magnitude": 1.0,
+                        "null_calculated_value_frequency": 1.0,
+                    },  # 331
```
e-belfer, line 3585: `TODO: Investigate utility 331`

test/unit/transform/ferc1_test.py
```diff
@@ -548,6 +556,207 @@ def test_calculate_values_from_components():
...
+electric,in_service,70
+gas,future,5
+gas,in_service,5
+total,total,{total_total_value}
+    """
+        )
+    )
+    out2 = get_reconcilded_data(data2)
+    assert (
+        out2.loc[out2.row_type_xbrl == "subdimension_correction", "value"].to_numpy()[0]
```
cmgosnell, line 670: add another assertion for only one correction record

src/pudl/transform/ferc1.py
```diff
@@ -5713,6 +5713,63 @@ def transform_main(self, df: pd.DataFrame) -> pd.DataFrame:
         )
         return df
 
+    def transform_end(self, df: pd.DataFrame) -> pd.DataFrame:
+        """Run standard :meth:`Ferc1AbstractTableTransformer.transform_end` plus a data validation step.
```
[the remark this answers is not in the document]
cmgosnell: this was much more clear! i made one tweak to this which was in your first e.g.: "(e.g. `plant_status == "total"`)" instead of "in_service" there.

cmgosnell:
great catch on the `subdimension_correction` not in the enum! I added `subtotal_correction` into the enum in the schema and then i up and changed it to subdimension bc that was more consistent w/ the code. will change now.

cmgosnell:
i just re-ran the fast etl which i hadn't done in a minute and noticed there are a few tables that need small adjustments to their calc metric tolerances. i'll update and push momentarily

e-belfer (approved):
Everything ran as expected and looks good to me.

PR #3482, opened 2024-03-19
eia860 solar: extract

### LINKED ISSUE #3469: Compile file, page, and column maps for the EIA860 solar generator attributes (exists in 2013-2022 data)
[See docs about mapping EIA forms](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#a-eia-forms)

```[tasklist]
- [x] `{data_source}/file_map.csv`: Paths (within the annual zip archive) to the files we parse.
- [ ] `{data_source}/page_map.csv`: Mapping between the named spreadsheet pages we refer to across years, and the numerical index of that page within the workbook.
- [ ] `{data_source}/skiprows.csv`: A per-page, per-year number of rows that should be skipped when reading the spreadsheet.
- [ ] `{data_source}/skipfooter.csv`: A per-page, per-year number of rows that should be ignored at the end of the page when reading the spreadsheet.
- [ ] `{data_source}/column_maps/${page_name}.csv`: A mapping from annual spreadsheet columns to consistent inter-year column names that we refer to in the raw dataframes during the extract step. The spreadsheet columns can be referred to either by their simplified snake_case column header (in eia860, eia860m, and eia923) or numerical column index (eia861).
```

### LINKED ISSUE #3470: Extract EIA860 solar: generator attributes into several `raw_eia860__solar_generator_*` assets.

cmgosnell:
# Overview

Closes #3469 & Closes #3470.

- the fact that we don't have the eia 860 solar tables!

What did you change?
- mapped all the extraction metadata for the operating and retired solar tabs

Notes/Question:
- there are a ton of boolean (Y/N) columns. I used the prefix `uses_` for all of those. Normally i like `is_` prefix for boolean but they are all columns that indicate whether the solar generator _uses_ a particular technology. i could go either way on this one though. definitely curious what folks think

# Testing

```[tasklist]
# To-do list
- [x] Run the full raw extract 860 assets and mannually check that these two new assets exist and look reasonable on first blush
- [x] For minor ETL changes or data additions, once `make pytest-coverage` passes, make sure you have a fresh full PUDL DB downloaded locally, materialize new/changed assets and all their downstream assets and [run relevant data validation tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#data-validation) using `pytest` and `--live-dbs`.
```


zaneselvans:
On the boolean column naming I've also seen `has_` as a prefix alongside `is_` for descriptive attributes.

src/pudl/extract/eia860.py
```diff
@@ -77,6 +77,8 @@ def get_dtypes(page, **partition):
     "raw_eia860__generator_existing",
     "raw_eia860__generator_proposed",
     "raw_eia860__generator_retired",
+    "raw_eia860__generator_solar_existing",
+    "raw_eia860__generator_solar_retired",
```
aesharpe, line 81: Two quick things here: 
1) What is the reasoning behind including `generator` in the table name?
2) In the raw csv, the tab for the retired table says "retired or canceled" -- this was the same for the storage table, so I gave it the suffix `_retired_canceled`. It's a little clunky, but I think it's more informative. Whatever we choose it should probably be the same for both tables.

src/pudl/package_data/eia860/column_maps/generator_solar_existing.csv
```diff
@@ -0,0 +1,42 @@
...
+uses_east_west_fixed_tilt,,,,east_west_fixed_tilt,east_west_fixed_tilt,east_west_fixed_tilt,east_west_fixed_tilt,east_west_fixed_tilt,east_west_fixed_tilt,east_west_fixed_tilt
+uses_parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough
+uses_linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel
+uses_power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower
+uses_dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine
+uses_other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology
+uses_azimuth_angle,,,,,azimuth_angle,azimuth_angle,azimuth_angle,azimuth_angle,azimuth_angle,azimuth_angle
```
aesharpe, line 29: I don't think the `azimuth angle` column is a boolean. Should probably remove the `uses` prefix.

src/pudl/package_data/eia860/column_maps/generator_solar_existing.csv
```diff
@@ -0,0 +1,42 @@
...
+uses_tilt_angle,,,,tilt_angle,tilt_angle,tilt_angle,tilt_angle,tilt_angle,tilt_angle,tilt_angle
```
aesharpe, line 30: Same with the `tilt angle` -- not a boolean column.

src/pudl/package_data/eia860/column_maps/generator_solar_retired.csv
```diff
@@ -0,0 +1,44 @@
...
+uses_parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough,parabolic_trough
+uses_linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel,linear_fresnel
+uses_power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower,power_tower
+uses_dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine,dish_engine
+uses_other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology,other_solar_technology
+uses_azimuth_angle,,,,,azimuth_angle,azimuth_angle,azimuth_angle,azimuth_angle,azimuth_angle,azimuth_angle
+uses_tilt_angle,,,,tilt_angle,tilt_angle,tilt_angle,tilt_angle,tilt_angle,tilt_angle,tilt_angle
```
aesharpe, line 32: Same comment as above about cols not being boolean.

aesharpe (changes requested):
Ran the `debug-column-mapping` notebook on the branch with newly materialized assets and everything went smoothly! 

Just a few questions about names and found some columns that aren't booleans and should have the verb prefixes removed.

aesharpe:
> On the boolean column naming I've also seen `has_` as a prefix alongside `is_` for descriptive attributes.

Just to record our decision made in slack - we'll stick to any simple, present tense verb suffixes for boolean columns and not try to hyper standardize which verbs we use.

src/pudl/extract/eia860.py
cmgosnell: I used `generator` because there are generator IDs. They are just a special solar case of generators. I could be convinced that we should do that but i personally think it makes sense to make it explicit and consistent with the other generator tabs.

And I used `retired` only because that is what we use for the standard retired/canceled generators. I'm slightly partial to this shorter version partially bc shorter but also just because i think we should be consistent and if we add both suffixes it would be good to add them to the current generators tabs as well.

src/pudl/package_data/eia860/column_maps/generator_solar_existing.csv
cmgosnell: great catch!

src/pudl/extract/eia860.py
aesharpe: Ah ok that makes sense! Is it duplicate data from what is in the generator table? If it's unique, maybe we could rename the generator table to be more specific too. Like generator_steam_existing or something like that...idk.

I also hate the clunkiness of `retired_canceled` but to your point in the generators table, it is more explicit which I think is important. Feels like small beans though. Maybe let's get a third opinion @e-belfer what do you think?

cmgosnell: From my understanding the current generators includes all of the generators and these two new solar and storage tables are special subsets of the generators that have specific solar and storage attributes. Because of that, I personally do not think it makes sense to rename the current generators tabs.

yea i don't disagree w/ `retired_canceled`. i just don't think its a big deal bc its entirely internal and if we name this one `retired_canceled` we should definitely also rename the existing ones which will be fine but slightly more involved.

aesharpe (approved):

PR #3489, opened 2024-03-20
Add manual GridPath RA Toolkit renewable profile data source.

zaneselvans:
# Overview

Adds necessary Zenodo DOI and data source metadata for the manually compiled GridPath RA Toolkit data archive.

Closes https://github.com/catalyst-cooperative/pudl-archiver/issues/296

# Testing

I validated the manually compiled `datapackage.json` using `frictionless validate` (before switching file paths for URLs -- this command would fail or be very slow now, since it would try and download the remote files)

```console
frictionless validate datapackage.json
```

* Downloaded the new archive with `pudl_datastore`
* Validated the new archive with `pudl_datastore`
* Successfully accessed all files in the new archive using a `Datastore` instance in a notebook.

```[tasklist]
# To-do list
- [x] Ensure docs build locally
```


src/pudl/workspace/datastore.py
```diff
@@ -60,8 +60,12 @@ def __init__(self, datapackage_json: dict, dataset: str, doi: ZenodoDoi):
...
         """Returns zenodo url that holds contents of given named resource."""
         res = self._get_resource_metadata(name)
-        # remote_url is sometimes set on the local cached version of datapackage.json
-        # so we should be using that if it exists.
+        # In older cached archives, "remote_url" was used to refer to the original path
+        # to the file, while the canonical "path" field was updated by the datastore
+        # to refer to the local path to the associated file relative to the location of
+        # datapackage.json. This behavior is deprecated and no longer used, but we need
+        # to retain this logic to support older cached archives, e.g. censusdp1tract
+        # which hasn't changed since 2020.
```
zaneselvans, line 68: A clarification that goes along with the changes in https://github.com/catalyst-cooperative/pudl-archiver/pull/303

src/pudl/metadata/sources.py
```diff
@@ -676,6 +676,43 @@
...
+                "weather",
+                "capacity factor",
+                "hourly",
+                "united states",
+                "usa",
+            }
+        ),
+        "license_raw": LICENSES["cc-by-4.0"],
+        "license_pudl": LICENSES["cc-by-4.0"],
+    },
```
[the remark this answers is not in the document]
zaneselvans: Also @e-belfer if you want to inspect the archive:

https://zenodo.org/records/10844662

It sounds like there might be some additional useful open data in the upstream source that should be added, so I will probably end up making another version, and we could change the way the data is currently partitioned if it makes sense... There are kind of 3 phases:
- project level (generator / plant)
- aggregated
- aggregated + extended

And in terms of the inputs, there's a wind vs. solar dimension.

Then there are some additional files that fit into that framework, like the daily weather stats, the files defining the aggregations of wind and solar data, and potentially hourly capacity factor derates for thermal plants, which I didn't realize used only open inputs (still need to verify that)

src/pudl/metadata/sources.py
```diff
@@ -676,6 +676,43 @@
...
+            "some problematic individual generator profiles modified such that they "
+            "match the overall production curve of the balancing authority they are "
+            "within. This data also contains some daily weather data from several "
+            "sites across the western US and tables describing the way in which "
+            "individual wind and solar projects were aggregated up to the level of "
+            "balancing authority or transmission zone."
+        ),
+        "keywords": sorted(
+            {
+                "solar",
```
e-belfer, line 700: Non-blocking: Do you want to add `gridpath` and `resource adequacy` here?

e-belfer (approved):
Works as expected for me. One non-blocking suggestion re: keywords that might make sense to add here.

src/pudl/metadata/sources.py
zaneselvans: Sure, that's reasonable. But really I'll need to add them by hand into the archive itself. None of this metadata is used anywhere.

PR #3491, opened 2024-03-20
Fix nightly build FK failure

e-belfer:
# Overview

Fixes nightly build failure.

What did you change?
Adds `_core_eia923_fgd_operation_maintenance` to FK exclusions for harvested plants table.

# Testing

Run `alembic upgrade head`, then `pudl_check_fks`

```[tasklist]
# To-do list
```


src/pudl/metadata/resources/eia860.py
```diff
@@ -373,6 +373,7 @@
                 # See: https://github.com/catalyst-cooperative/pudl/issues/1196
                 "exclude": [
                     "_core_eia923__cooling_system_information",
+                    "_core_eia923__fgd_operation_maintenance",
```
jdangerx, line 376: Oops, sorry about that... during review, I had actually thought "hm, I ran into some FK stuff with this specific table in the EIA 923 data on *my* PR, I wonder if it will be a problem here?"

But then I looked at the data & somehow thought "eh, it seems annual-only, guess that's fine." 🤷

jdangerx (approved):

src/pudl/metadata/resources/eia860.py
e-belfer: That's also what I thought, but we aren't yet harvesting any plant IDs present in this table so there may be some new values here that aren't represented in the harvested data, I think. To be resolved when we harvest!

Issue #3453, opened 2024-03-08
Nightly Build Failure 2024-03-08


jdangerx:
# Overview

What do you think happened, if anything?

Looks like we are looking for the individual PHMSA Excel files as top-level resource names instead of trying to match to the right partition. Why, though, and why now?

# Next steps

What next steps do we need to do to understand or remediate the issue?

1. Reproduce this error by reproducing the `.get_unique_resource()` call:
```
>>> from pudl.workspace.datastore import Datastore
>>> ds = Datastore()
>>> ur = ds.get_unique_resource("phmsagas", name='annual_gas_distribution_1992.xlsx')
2024-03-08 09:12:29 [    INFO] catalystcoop.pudl.workspace.datastore:263 Retrieving https://zenodo.org/api/records/10493790/files from zenodo
2024-03-08 09:12:30 [    INFO] catalystcoop.pudl.workspace.datastore:263 Retrieving https://zenodo.org/api/records/10493790/files/datapackage.json/content from zenodo
Traceback (most recent call last):
  File "/Users/dazhong-catalyst/work/pudl/src/pudl/workspace/datastore.py", line 405, in get_unique_resource
    _, content = next(res)
                 ^^^^^^^^^
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dazhong-catalyst/work/pudl/src/pudl/workspace/datastore.py", line 407, in get_unique_resource
    raise KeyError(f"No resources found for {dataset}: {filters}") from err
KeyError: "No resources found for phmsagas: {'name': 'annual_gas_distribution_1992.xlsx'}"
```

2. See why we're making that incorrect get_unique_resource call

# Verify that everything is fixed!

Once you've applied any necessary fixes, make sure that the nightly build outputs are all in their right places.

```[tasklist]
- [ ] [S3 distribution bucket](https://s3.console.aws.amazon.com/s3/buckets/pudl.catalyst.coop?region=us-west-2&bucketType=general&prefix=nightly/&showversions=false) was updated at the expected time
- [ ] [GCP distribution bucket](https://console.cloud.google.com/storage/browser/pudl.catalyst.coop/nightly;tab=objects?project=catalyst-cooperative-pudl) was updated at the expected time
- [ ] [GCP internal bucket](https://console.cloud.google.com/storage/browser/builds.catalyst.coop) was updated at the expected time
[...]

PR #3493, opened 2024-03-21
Add logline that tells us more about BadZipFile.

jdangerx:
# Overview

Related to #3453 .

We keep getting these BadZipFile errors but don't really have much to go on - so we should log out some information about the ZipFiles.

What did you change?

Add a logline.

# Testing

Ran `raw_phmsagas__all_dfs` asset in dagster and verified that these logs showed up in the terminal:

```
2024-03-21 16:45:22 [    INFO] catalystcoop.pudl.workspace.datastore:425 Got resource dataset='phmsagas', filters={'year': 2008, 'form': 'gas_distribution'}, md5sum='4ef276c35808fd8ea0feddf6e2881511', 8513846 bytes; turning into ZipFile    
```

### DIFF
--- a/src/pudl/workspace/datastore.py
+++ b/src/pudl/workspace/datastore.py
@@ -419,7 +419,14 @@ class Datastore:
 
     def get_zipfile_resource(self, dataset: str, **filters: Any) -> zipfile.ZipFile:
         """Retrieves unique resource and opens it as a ZipFile."""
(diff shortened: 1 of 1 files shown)


e-belfer (approved):
Works as expected. Eventually would be nice to turn this into a debug or possibly run the nightly builds with debug logs, but seems good for now!

PR #3497, opened 2024-03-25
Extract EIA 930 data, refactor extractors to handle different date partitions

### LINKED ISSUE #3486: Make a CSV Extractor for EIA 930 files
Extract EIA 930 data into 3 raw dataframe assets: raw_eia930__balance, raw_eia930__subregion, and raw_eia930__interchange.

To do this, this PR does the following:
- refactor the extractor to grab any specified partition, not just `year` (incidentally, this will make #3451 much simpler)
- map columns for three tables
- add new `assets` to `eia930.py`

Outstanding questions for the reviewer:
- Dagster makes specifying `str|int` very complex, and we use both string and int partition types. For now, I'm getting around this by specifying a `dict` type without any particular types inside it in `extract_single_partition`, but this doesn't seem ideal. Thoughts?
- Use of assets in EIA 930 to support easier subsetting, instead of multi-assets?
- We don't typically specify dtypes on import, but these are large datasets. Should we?

```[tasklist]
### Tasks
- [x] Add DOI to `datastore.py` and run `pudl_datastore --dataset eia930`
- [x] Update working partitions to use `half` instead of `h`
- [x] Refactor extractor to take `partition` rather than `year` as the unit for parallelization of extraction
- [x] Map column name changes over time for each table
- [x] Handle `subregion` only existing 2018-2024
- [x] Handle `half_year` vs `half_years`
- [x] Fix `extractor_test.py`
- [ ] Handle mixed dtypes `str|int` in partitions and test on other extracted datasets
- [ ] Add EIA 930 form PDFs to `docs.data_sources`
```

e-belfer:
# Overview

Closes #3486. Progress, design and questions for the reviewer are tracked in the issue.

# Questions for the reviewer
This seems like a slightly clunky way of avoiding Dagster typing challenges and of handling inconsistencies in our pluralization of partitions, that was avoided earlier by just hardcoding the parameter as `year` or `years`. Additional notes on these challenges in the issue. Feedback welcome on design decisions here in the base extractor class.

# Testing

Extract all raw archives, or run the fast ETL.

```[tasklist]
# To-do list
- [x] Review the PR yourself and call out any questions or issues you have (see issue)
```


jdangerx (changes requested):
> Dagster makes specifying str|int very complex, and we use both string and int partition types. For now, I'm getting around this by specifying a dict type without any particular types inside it in extract_single_partition, but this doesn't seem ideal. Thoughts?

I was able to add this code and make Dagster's type inference happy - specifically it seems to not like having parametrized types like `dict` or `list` that then also have union types inside them. It's not *that* pretty but also not *that* awful. I think it'd be nice to put it in.

```diff
diff --git a/src/pudl/extract/extractor.py b/src/pudl/extract/extractor.py
index 55fb846ae..a76be685d 100644
--- a/src/pudl/extract/extractor.py
+++ b/src/pudl/extract/extractor.py
@@ -3,13 +3,17 @@
 import importlib.resources
 from abc import ABC, abstractmethod
 from collections import defaultdict
+from typing import Any
 
 import pandas as pd
 from dagster import (
     AssetsDefinition,
+    DagsterType,
     DynamicOut,
     DynamicOutput,
+    In,
     OpDefinition,
+    TypeCheckContext,
     graph_asset,
     op,
 )
@@ -290,6 +294,22 @@ def concat_pages(paged_dfs: list[dict[str, pd.DataFrame]]) -> dict[str, pd.DataF
     return all_data
 
 
+def _is_dict_str_strint(_context: TypeCheckContext, x: Any) -> bool:
+    if not isinstance(x, dict):
+        return False
+    for key, value in x.items():
+        if not isinstance(key, str):
+            return False
+        if not isinstance(value, str | int):
+            return False
+    return True
+
+
+dagster_dict_str_strint = DagsterType(
+    name="dict[str, str | int]", type_check_fn=_is_dict_str_strint
+)
+
+
 def partition_extractor_factory(
     extractor_cls: type[GenericExtractor], name: str
 ) -> OpDefinition:
@@ -303,10 +323,11 @@ def partition_extractor_factory(
     @op(
         required_resource_keys={"datastore", "dataset_settings"},
         name=f"extract_single_{name}_partition",
+        ins={"part_dict": In(dagster_type=dagster_dict_str_strint)},
     )
     def extract_single_partition(
         context,
-        part_dict: dict,  # TODO: Handle str|int type errors
+        part_dict: dict[str, str | int],
     ) -> dict[str, pd.DataFrame]:
         """A function that extracts a year of spreadsheet data from an Excel file.
 
```

We could pull this into a separate utility `custom_dagster_types` module if we wanted to, also.

In the future I suppose we'll want to refactor to use specific `Partition` classes, which will end up playing nicer with `DagsterType`. But I think that's *certainly* out of scope of this PR.

> Use of assets in EIA 930 to support easier subsetting, instead of multi-assets?

You mean for `balance`, `interchange`, and `subregion`? I like it!

> We don't typically specify dtypes on import, but these are large datasets. Should we?

Right now we don't have a mechanism for checking this without promoting this all the way to a resource in the `RESOURCE_METADATA`.

We could in theory just do some pandas dtype conversions so we don't just have `object` everywhere. The risk is that it will start to expose problems with the underlying data (e.g. "this string can't be converted to a float!") and involve actual data cleaning. This hurts me, as Mr. I Love Types, but I think defining dtypes at this point might not be worth it...

src/pudl/package_data/eia930/__init__.py
```diff
@@ -0,0 +1 @@
+"""CSV file extraction maps for EIA 176."""
```
e-belfer, line 1: This should be 930, not 176. Same for the file below. Will fix the copy-pasta.

src/pudl/extract/extractor.py
```diff
@@ -290,10 +294,26 @@ def concat_pages(paged_dfs: list[dict[str, pd.DataFrame]]) -> dict[str, pd.DataF
...
+        return False
+    for key, value in x.items():
+        if not isinstance(key, str):
+            return False
+        if not isinstance(value, str | int):
+            return False
+    return True
+
+
+dagster_dict_str_strint = DagsterType(
```
jdangerx, line 308:
```suggestion

# 2024-03-27: Dagster can't automatically convert union types within
# parametrized types; we have to write our own custom DagsterType for now.
dagster_dict_str_strint = DagsterType(
```

src/pudl/extract/extractor.py
```diff
@@ -333,27 +356,46 @@ def years_from_settings_factory(name: str) -> OpDefinition:
...
+            var
+            for var in vars(data_settings)
+            if any(
+                date_partition in var
+                for date_partition in ["years", "half_years", "year_quarters"]
+            )
+        ]
+        assert (
+            len(partition) == 1
+        ), f"Only one working partition is supported: {partition}."
```
jdangerx, line 389: just for my clarification: this is so we don't try to get "years" and also "half_years"?

jdangerx (approved):
Looks good, modulo those docstring changes you wanted to make! Just wanted to add a little comment for the dagster type stuff since that's kind of weird.

PR #3498, opened 2024-03-25
NREL ATB axtraction

### LINKED ISSUE #3468: Extract the NREL ATB data to a raw dataframe

cmgosnell:
# Overview

Closes #3468.

What did you change?

# Testing

```[tasklist]
# To-do list
```


cmgosnell:
hm i accidentally made this branch off of my unfinished eia extraction branch... i think it's okay now but i probably should have done something besides merge in main...

src/pudl/extract/excel.py
```diff
@@ -85,17 +85,6 @@ def get_file_name(self, page, **partition):
...
-        """Return dictionary for renaming columns in a given partition and page."""
-        return {
-            v: k
-            for k, v in self._column_map[page]
-            .T.loc[str(self._get_partition_selection(partition))]
-            .to_dict()
-            .items()
-            if v != -1
-        }
-
```
cmgosnell, line 98: i moved this from the excel extractor to the generic extractor with the general idea that `process_raw` should actually do any renaming. This was one of several little migrations that needed to happen for that.

src/pudl/extract/extractor.py
```diff
@@ -145,7 +156,7 @@ def process_raw(
         self, df: pd.DataFrame, page: str, **partition: PartitionSelection
     ) -> pd.DataFrame:
         """Takes any special steps for processing raw data and renaming columns."""
-        return df
+        return df.rename(columns=self._metadata.get_column_map(page, **partition))
```
cmgosnell, line 159: previously only the excel extractor actually did the rename by default. it seems like most of the new csv extracted raw inputs were not trying to rename any of the columns, but still have column_maps w/ 1:1 existing name : pudl name. So as far as i can tell nothing has changed.

src/pudl/extract/parquet.py
```diff
@@ -0,0 +1,51 @@
...
+            page: pudl name for the dataset contents, eg "boiler_generator_assn" or
+                "data"
+            partition: partition to load. Examples:
+                {'year': 2009}
+
+        Returns:
+            string name of the parquet file
+        """
+        partition_selection = self._metadata._get_partition_selection(partition)
+        return f"{self._dataset_name}-{partition_selection}.parquet"
```
cmgosnell, line 32: this actually isn't needed for this extractor in its current form.

src/pudl/extract/parquet.py
```diff
@@ -0,0 +1,51 @@
...
+                "data"
+            partition: partition to load. Examples:
+                {'year': 2009}
+                {'year_month': '2020-08'}
+
+        Returns:
+            pd.DataFrame instance containing CSV data
+        """
+        res = self.ds.get_unique_resource(self._dataset_name, **partition)
+        df = pd.read_parquet(io.BytesIO(res))
```
cmgosnell, line 50: is there a better way to do this?

test/unit/extract/csv_test.py
```diff
@@ -55,12 +55,17 @@ def test_load_source(mock_pd):
 
 
 def test_extract():
-    extractor = FakeExtractor()
+    class SuperFakeExtractor(FakeExtractor):
+        def process_raw(self, df, page, **partition):
+            return df
```
cmgosnell, line 60: i removed the rename here. i think it would be easier overall if this test could either be ripped from the 176 extractor which would require all of the metadata bits necessary to really make an test extractor run.

src/pudl/package_data/nrelatb/column_maps/data.csv
```diff
@@ -0,0 +1,21 @@
+year_index,2019,2020,2021,2022,2023
+report_year,atb_year,atb_year,atb_year,atb_year,atb_year
+core_metric_key,core_metric_key,core_metric_key,core_metric_key,core_metric_key,core_metric_key
+core_metric_parameter,core_metric_parameter,core_metric_parameter,core_metric_parameter,core_metric_parameter,core_metric_parameter
+core_metric_case,core_metric_case,core_metric_case,core_metric_case,core_metric_case,core_metric_case
+cost_recovery_period_years,crpyears,crpyears,crpyears,crpyears,crpyears
+technology_description,technology,technology,technology,technology,technology
+technology_alias,,,technology_alias,technology_alias,technology_alias
+display_name,,,display_name,display_name,display_name
+is_default,,,default,default,default
```
cmgosnell, line 10: i couldn't find documentation about what this field actually is. There are only ever nulls 0's and 1's.... it looks like a quasi boolean but i can't tell.

src/pudl/package_data/nrelatb/column_maps/data.csv
```diff
@@ -0,0 +1,21 @@
+year_index,2019,2020,2021,2022,2023
+report_year,atb_year,atb_year,atb_year,atb_year,atb_year
+core_metric_key,core_metric_key,core_metric_key,core_metric_key,core_metric_key,core_metric_key
+core_metric_parameter,core_metric_parameter,core_metric_parameter,core_metric_parameter,core_metric_parameter,core_metric_parameter
+core_metric_case,core_metric_case,core_metric_case,core_metric_case,core_metric_case,core_metric_case
+cost_recovery_period_years,crpyears,crpyears,crpyears,crpyears,crpyears
+technology_description,technology,technology,technology,technology,technology
+technology_alias,,,technology_alias,technology_alias,technology_alias
```
cmgosnell, line 8: the description and the alias fields were very similar in content and i could not find a reference to this field in the documentation.

src/pudl/extract/excel.py
e-belfer: Oops, perils of parallelism. We're doing similar things in slightly different ways, as I ran into the same issue in #3497.

e-belfer: But I think this looks neat and yours will get merged in first, so I'll handle the conflict!

src/pudl/extract/extractor.py
e-belfer: Tested to make sure that if we delete the column mapping CSV for datasets where it actually isn't necessary everything works just fine in the extraction. So this seems like a good and fine decision.

src/pudl/package_data/nrelatb/column_maps/data.csv
e-belfer: Blocking: can you add the docs you're referencing here to `pudl.docs.data_sources.nrelatb`?

test/unit/extract/csv_test.py
```diff
@@ -55,12 +55,17 @@ def test_load_source(mock_pd):
...
+    extractor = SuperFakeExtractor()
     # Create a sample of data we could expect from an EIA CSV
     company_field = "company"
     company_data = "Total of All Companies"
     df = pd.DataFrame([company_data])
     df.columns = [company_field]
+
     # TODO: Once FakeExtractor is independent of eia176, mock out populating _column_map for PARTITION_SELECTION;
     #  Also include negative tests, i.e., for partition selections not in the _column_map
     with (
```
e-belfer, line 71: Instead of dropping the rename, let's actually test it.

```suggestion
    with (
        patch.object(CsvExtractor, "load_source", return_value=df),
        patch.object(
            # Testing the rename
            GenericMetadata,
            "get_column_map",
            return_value={company_field: 'company_rename'},
        ),
        patch.object(
            # Transposing the df here to get the orientation we expect get_page_cols to return
            CsvExtractor,
            "get_page_cols",
            return_value=df.T.index,
        ),
    ):
        res = extractor.extract(**PARTITION)
    assert len(res) == 1 # Assert only one page extracted
    assert list(res.keys()) == [PAGE] # Assert it is named correctly
    assert res[PAGE]['company_rename'][0] == company_data # Assert that column correctly renamed and data is there.

```

e-belfer (changes requested):
Two smallish blocking changes:

1. Please add any ATB docs that you're referencing into our `data_sources` folder.
2. Suggested changes to the extraction test to actually test the rename, rather than skipping it.

src/pudl/extract/parquet.py
e-belfer: This should be fine, I think!

src/pudl/package_data/nrelatb/column_maps/data.csv
cmgosnell: I was about to go add em but then got turned around because the docs i went digging into are a [series of web pages](https://atb.nrel.gov/electricity/2023/definitions). these docs are mostly pdfs with some html of the empty forms. Does that still seem reasonable? and if so should I just extract the html from the few docs pages per year that I've been pawing through? this feels a little off brand from the rest of these docs.

e-belfer: Hm... this is the thing we would automatically link in our ReadtheDocs page for that data source, and I see some value in having an archived link of the dataset documentation in case it all gets taken down, particularly for older datasets. As long as the file name reflects what it contains (e.g., 2023 ATB Definitions, 2023 ATB Scenarios) I think archiving them should make sense so long as they aren't prohibitively large. I would imagine these pages aren't changing more than once per year. But it's not a hill I'll die on, if you don't think this makes sense don't worry about it.

cmgosnell: hm well i think that actually the info on the read the docs page is just linked to the info in `metadata.sources`. i believe the docs pdfs/html form are just for internal use. I could be wrong about that though.

e-belfer (approved):
Looks good to go!

PR #3514, opened 2024-03-26
WIP: GridPath RA Toolkit wind and solar generation profiles

### LINKED ISSUE #3509: Extract raw aggregated extended wind and solar generation profiles
```[tasklist]
- [x] Decide whether to use existing CSV extraction tooling, or something bespoke
- [x] Create raw asset for the most processed wind data (aggregated, extended)
- [x] Create raw asset for the most processed solar data (aggregated, extended)
- [x] Generalize per-directory raw profile creation into an asset factory
```

### LINKED ISSUE #3510: Transform wind and solar generation profiles into core tables
```[tasklist]
- [x] Build a datetime column and adjust to UTC
- [x] Reshape to tidy data format
- [x] Decide how to combine wind and solar profiles into one table
- [x] Define field & resource metadata / schema
```

### LINKED ISSUE #3511: Extract raw wind and solar aggregation tables
```[tasklist]
- [x] Extract wind aggregation table
- [x] Extract solar aggregation table
- [x] Verify that hybrid profiles are defined in aggregation tables
```

### LINKED ISSUE #3515: Transform wind and solar aggregations into core tables
```[tasklist]
- [x] Split composite aggregation keys into plant & generator IDs
- [x] Rename columns using PUDL naming conventions
- [x] Set column dtypes
- [x] Create field and resource definitions.
- [x] Add FK constraints linking aggregations to EIA plant & generator tables
- [x] Debug `plant_id_eia` and `generator_id` FK constraint violations
- [x] ~Fix generator IDs with leading zeroes at the beginning of the transform~ (unnecessary)
- [x] ~Replace `plant_id_eia` 64272 with 64481 at the beginning of the transform~ (unnecessary)
- [x] Use an `@asset_check()` to ensure that all aggregation keys appear in aggregation table
- [x] Ask Ana / Elaine whether `Hybrid_Solar_64481` should include both generators `AEC1` and `AEC2`.
- [ ] Check hour starting vs. ending conventions in our hourly tables.
```

zaneselvans:
# Overview

A first draft of extracting and transforming the most processed version of the hourly wind and solar profiles from the GridPath Resource Adequacy Toolkit.

Closes #3509 
Closes #3510 
Closes #3511 
Closes #3515 

# Testing

- Ran the new extract and transform assets locally and reviewed them in a notebook.
- Checked that the new table and parquet outputs are being generated.

```[tasklist]
# To-do list
- [x] Get feedback on draft
- [x] Document extract step better
- [x] Rename all `utc_datetime` to `datetime_utc`
- [x] Update the release notes. Reference the PR and related issues.
- [x] Run `make pytest-coverage` (otherwise the merge queue may reject your PR)
- [x] Move partition choosing logic from extract into settings class
- [x] Pull private extract functions out of factory
- [x] Rename `aggregation_key` -> `aggregation_group`
- [x] Rename aggregation table to `core_gridpathratoolkit__assn_generator_aggregation_group`
- [x] Rename capacity factor table to `out_gridpathratoolkit__hourly_available_capacity_factor`
- [x] Reference the table that's compiled by aggregation in aggregation group table description
```


src/pudl/extract/gridpathratk.py
```diff
@@ -0,0 +1,85 @@
...
+
+# TODO: Reorganize the archive to use asset names for partitions directly
+GRIDPATHRATK_PARTS = {
+    "hourly-solar-zonal-capacity-synth": {
+        "table_name": "aggregated_extended_solar",
+    },
+    "hourly-wind-zonal-capacity-synth": {
+        "table_name": "aggregated_extended_wind",
+    },
+}
```
zaneselvans, line 44: I can change the name of the archive files to match whatever we want to call these so there's no mapping. Hopefully we'll also need to update it to include the thermal plant derates.  I also want to get rid of the top level directory inside the zipfiles. It might also be possible to break the archive down along several real partitions (aggregation / extended or not) but it's not a long series like with the years or states so I kinda don't think it's worth it, and just addressing the individual zipfiles as a unit is fine.

src/pudl/extract/gridpathratk.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def raw_gridpathratk_profile_asset_factory(part: str) -> AssetsDefinition:
```
zaneselvans, line 47: Should probably standardize on `profile` or `capacity_factors` or some other name that can remain consistent throughout the whole process.

zaneselvans: The structure of the thermal derate data is the same, but it's not aggregated, so there's ~100M records spread across ~1000 files, and it won't fit in memory, so I think it'll need to be done in pieces as with the CEMS. But luckily the files are each small, so we can parallelize massively and gracefully if we bring them in. But the factory would need to be modified to allow that kind of processing.

src/pudl/extract/gridpathratk.py
```diff
@@ -0,0 +1,85 @@
...
+    This factory works on the processed hourly profiles that store one capacity factor
+    time series per file with a the time index stored in a separate timestamps.csv. The
+    stems of the filenames are used to identify the timeseries for later processing.
+    """
+
+    @asset(
+        name=f"raw_gridpathratk__{GRIDPATHRATK_PARTS[part]["table_name"]}",
+        required_resource_keys={"datastore", "dataset_settings"},
+    )
+    def _extract_raw_gridpathratk(context):
```
zaneselvans, line 59: I kept the processing here to the bare minimum required to get all the data into a well structured dataframe.

zaneselvans (commented):
This is a draft, but it does get the data into the database & parquet outputs. I've highlighted some things I'm not entirely happy with, and would love feedback before I revise.

src/pudl/metadata/resources/gridpathratk.py
```diff
@@ -0,0 +1,29 @@
...
+            "individual plants, where the individual plants are hybrid wind or solar "
+            "plus electricity storage facilities. For the hybrid facilities the "
+            "capacity factor represents the available output of only the renewable "
+            "generators. Estimates of the generation profiles have been extended "
+            "across a longer range of dates than the original data. See the Appendix "
+            "the GridPath Resource Adequacy Toolkit report for more details."
+        ),
+        "schema": {
+            "fields": [
+                "utc_datetime",
```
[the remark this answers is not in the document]
e-belfer: I agree, should be `datetime_utc`.

src/pudl/metadata/resources/gridpathratk.py
```diff
@@ -0,0 +1,29 @@
...
+    "core_gridpathratk__aggregated_extended_profiles": {
+        "description": (
+            "Hourly capacity factors defining the generation profiles for solar and "
+            "wind. Contains a mix of profiles representing whole regions and "
+            "individual plants, where the individual plants are hybrid wind or solar "
+            "plus electricity storage facilities. For the hybrid facilities the "
+            "capacity factor represents the available output of only the renewable "
+            "generators. Estimates of the generation profiles have been extended "
+            "across a longer range of dates than the original data. See the Appendix "
+            "the GridPath Resource Adequacy Toolkit report for more details."
```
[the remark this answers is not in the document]
e-belfer:
```suggestion
            "Hourly capacity factors defining the generation profiles for solar and "
            "wind. Contains a mix of profiles representing whole regions and "
            "individual plants, where the individual plants are hybrid wind or solar "
            "plus electricity storage facilities. For the hybrid facilities the "
            "capacity factor represents the available output of only the renewable "
            "generators. Estimates of the generation profiles have been extended "
            "across a longer range of dates than the original data. See the Appendix "
            "of the GridPath Resource Adequacy Toolkit report for more details."
```

src/pudl/extract/gridpathratk.py
e-belfer: I don't have an opinion here, guessing @cmgosnell might. But agree that renaming the files during the download process to match whatever we want here and get rid of this second dictionary seems like a high priority. I could see having partitions that point to level of processing/aggregation (e.g., `generator`, `zonal`, `zonal_extended`), but agreed we can revisit this as we add in more tables.

src/pudl/metadata/resources/gridpathratk.py
```diff
@@ -0,0 +1,29 @@
+"""Table definitions for data coming from the GridPath Resource Adequacy Toolkit."""
+
+from typing import Any
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "core_gridpathratk__aggregated_extended_profiles": {
```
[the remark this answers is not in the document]
e-belfer: I don't have an opinion here, guessing @cmgosnell might.

src/pudl/extract/gridpathratk.py
e-belfer: Is your suggestion to have some gridpath data written to parquet, but not all tables? I'd suggest standardizing whatever the approach you wind up with is, even if the other tables could just be handled in memory. I'm wary of winding up with three different extraction streams for different tables, doesn't seem optimal. Ideally we'd subclass `Extractor` to make a `Parquet` extraction class that writes CSVs to Parquet files, but perhaps this is out of scope for the budget here.

src/pudl/metadata/resources/gridpathratk.py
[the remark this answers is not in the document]
e-belfer: But, I don't think this matches our existing naming conventions. There should be a `yearly` if the data is yearly, or some other indication of data frequency.

e-belfer (commented):
Left some comments, mostly around anticipating the parquet extraction. Everything ran fine for me.

src/pudl/metadata/resources/gridpathratk.py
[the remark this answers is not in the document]
zaneselvans: Ah yes, oops. Should be hourly.

src/pudl/extract/gridpathratk.py
zaneselvans: This would be on the output side, not the input.  On the inputs, it's a big zipfile filled with thousands of CSVs that each have ~200K rows, and can easily be processed in parallel without ever using too much memory, but we can't write 300M records to a table in SQLite.

src/pudl/metadata/resources/gridpathratk.py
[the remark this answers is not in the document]
cmgosnell: +1 to adding a frequency. I also think that `gridpathratk` as the dataset name doesn't conform to the rest of our data sources. I know that `gp` is less recognizable than `eia` in our little acronym corner but in this case I think `gp` is the "provider" and perhaps `ratk` is the form.

agreed that its hard to think about the main table-y part of the name... these are definitely going to be generation profiles. are all of the possible `ratk` tables "extended" in dates? if so I would probably take that out because its less what it is. but its not a hill i will die on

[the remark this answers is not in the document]
zaneselvans: @cmgosnell Can you say where you're thinking this distinction between the provider and the "form" needs to be made?  Here `core_gridpathratk` would be analogous to e.g. `core_eia860` which includes both the agency and the form.

src/pudl/metadata/resources/gridpathratk.py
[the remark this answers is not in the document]
zaneselvans: Are y'all saying that it should be changed now and also I should change the existing `utc_datetime` to be `datetime_utc` or that it should be changed now and I shouldn't change that other column, which would mean we'd end up with both `datetime_utc` and `utc_datetime` being defined now?

src/pudl/transform/gridpathratk.py
```diff
@@ -0,0 +1,57 @@
...
+        pd.DatetimeIndex(pd.to_datetime(profiles.loc[:, datetime_cols])) - utc_offset
+    )
+    profiles = (
+        profiles.drop(columns=datetime_cols)
+        .set_index("utc_datetime")
+        .stack()
+        .reset_index()
+    )
+    profiles.columns = ["utc_datetime", "aggregation_key", "capacity_factor"]
+    profiles = profiles.astype({"aggregation_key": "string"})
```
cmgosnell, line 31: why a string astype here but a categorical post concat-ing?

src/pudl/transform/gridpathratk.py
```diff
@@ -0,0 +1,57 @@
...
+    profiles["utc_datetime"] = (
+        pd.DatetimeIndex(pd.to_datetime(profiles.loc[:, datetime_cols])) - utc_offset
+    )
+    profiles = (
+        profiles.drop(columns=datetime_cols)
+        .set_index("utc_datetime")
+        .stack()
+        .reset_index()
+    )
+    profiles.columns = ["utc_datetime", "aggregation_key", "capacity_factor"]
```
cmgosnell, line 30: are you ensuring these columns are in this order? this seems more brittle than a rename.

src/pudl/extract/gridpathratk.py
cmgosnell: i think this looks mostly fine. a little more docs about why your grabbing the timestamps first, adding the aggregation key and concat-ing em together.

src/pudl/extract/gridpathratk.py
cmgosnell: i am torn on the naming of `generation_profiles` or `capacity_factors`.... It seems like colloquially people call these profiles but they are just a collection of capacity factors. I lean slightly towards capacity factors. and plus the asset name will have `hourly` in it so it'll be pretty clear what it is

cmgosnell (commented):
this looks surprisingly simple all in all.

to me, it looks like the archive has two partitions: `technology_type` and `aggregation_level`. iirc the ds is already equip with the option to grab multiple parts.

you probably know this but once you settle on some inputs you'll need to add settings.

src/pudl/transform/gridpathratk.py
zaneselvans: If you concatenate columns with categorical dtypes, but they have different categories, they revert to the underlying dtype So there's no point in making the columns categorical until after the concatenation has happened.

src/pudl/transform/gridpathratk.py
zaneselvans: Unfortunately the columns have no names until this assignment happens.

src/pudl/extract/gridpathratk.py
zaneselvans: I think I agree with you @cmgosnell . Something like `raw_gridpathratk__hourly_solar_capacity_factor` is legible and accurate.

src/pudl/metadata/resources/gridpathratk.py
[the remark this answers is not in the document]
cmgosnell: I'm speaking specifically of the `gridpathratk` part of this name. I'm mostly nothing that all of our other data sources are acronyms of "agencies" and then numbers or acronyms of "forms" and this one is  `gridpath` (agency) and `ratk` (form).

I think `gprat` is more in line. Or idk the GP model is actually published by GridLab... Should it be `gl` and the `gpra` or `gprat`?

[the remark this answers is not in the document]
zaneselvans: I feel like we can get away with using short acronyms for the familiar agencies like `ferc`, `eia`, and `epa` because in this context it's clear what they refer to, but `gp` or `gl` doesn't communicate who/where the data is coming from.

I'm pretty sure the GridPath model is independently developed by Ana / Blue Marble Analytics, and the GridPath Resource Adequacy Toolkit was a study supported by GridLab which used the GridPath model. I think the data was largely compiled by Elaine / Moment Energy Insights. So I don't know that there's a single clear "agency" like source to refer to.  I also suspect that this will be a one-off. Maybe just using `gridpathratoolkit` which is how they refer to it everywhere would be clearest.

> The GridPath RA Toolkit refers to both the datasets and the algorithms that were developed to support the 2026 Western US case study, which can be downloaded from this site.

zaneselvans:
Ah, @cmgosnell thanks for the reminder about needing to create ETL settings.

I think I am not going to try and split the data up into partitions along the technology type and level-of-processing dimensions, just because there are so few files, and also there are some other files which don't neatly fit into that organization. E.g. the historical daily weather data.

src/pudl/extract/gridpathratk.py
zaneselvans: Got rid of this, and renamed the files to reflect the things that are needed to distinguish between the different versions of the data so e.g.:
- `original_solar_capacity.zip`
- `aggregated_solar_capacity.zip`
- `aggregated_extended_solar_capacity.zip`
- `solar_capacity_aggregations.csv`

src/pudl/metadata/resources/gridpathratk.py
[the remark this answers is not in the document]
zaneselvans: I went ahead and made this adjustment to the FERC-714 data. For reasons I don't understand this made the alembic migrations very unhappy (somehow the new `datetime_utc` column ended up being left out of the primary key for the historical hourly demand table) so I had to wipe the migrations out and start over.  But the names are all the same now.

src/pudl/extract/gridpathratoolkit.py
```diff
@@ -0,0 +1,128 @@
...
+                "condition": "wind" in gpratk_settings.technology_types
+                and (
+                    "extended" in gpratk_settings.processing_levels
+                    or "aggregated" in gpratk_settings.processing_levels
+                ),
+            },
+        }
+
+        if part not in parts:
+            raise ValueError(f"Unable to process {part}!")
```
zaneselvans, line 111: This feels a bit messy to me, but it works, and the interface it provides via the settings yaml file is simple.

src/pudl/transform/gridpathratoolkit.py
```diff
@@ -0,0 +1,156 @@
...
+    return pd.concat(
+        [
+            _transform_capacity_factors(
+                capacity_factors=raw_df, utc_offset=pacific_standard_time
+            )
+            for raw_df in [
+                raw_gridpathratoolkit__aggregated_extended_solar_capacity,
+                raw_gridpathratoolkit__aggregated_extended_wind_capacity,
+            ]
+            if not raw_df.empty
```
zaneselvans, line 63: This check for an empty dataframe is required in case any of the technology types isn't being processed.

src/pudl/transform/gridpathratoolkit.py
```diff
@@ -0,0 +1,156 @@
...
+        .stack()
+        .reset_index()
+    )
+    capacity_factors.columns = ["datetime_utc", "aggregation_key", "capacity_factor"]
+    capacity_factors = capacity_factors.astype({"aggregation_key": "string"})
+    return capacity_factors
+
+
+@asset(io_manager_key="pudl_io_manager")
+def core_gridpathratoolkit__hourly_aggregated_extended_capacity_factors(
```
zaneselvans, line 44: This table name is a mouthful, and without the specificity of the wind / solar / thermal technology it feels awfully vague. Are there better options?

If we think about how we'd want to name the tables in the event that we've got all of them in or we're doing the aggregations and extensions ourselves, what would they look like?  Would the aggregation and extension be part of the "output" stage? At what point do the wind and solar profiles come together? Is that only after aggregation?

Is it wrong that the table name is plural? I think I might have carried that over from "profiles" accidentally.

Other options?

* `core_gridpathratoolkit__hourly_solar_capacity_factor`
* `out_gridpathratoolkit__hourly_weather_dependent_capacity_factors` (wind, solar, thermal derates)

src/pudl/extract/gridpathratoolkit.py
cmgosnell: hm i feel like this whole `parts` dictionary (maybe with the exception of the `func` should live in the settings classes. and then you can do what most of the setting classes do an verify the parts based on the working partitions.

cmgosnell (changes requested):
this looks pretty good overall! sprinkling of comments. most feel clarifying. my main request is actually migrating the parts dict to the settings class. but lmk if that feels off base.

src/pudl/extract/gridpathratoolkit.py
```diff
@@ -0,0 +1,128 @@
...
+        Args:
+            context: dagster keyword that provides access to resources and config.
+        """
+        gpratk_settings = context.resources.dataset_settings.gridpathratoolkit
+        ds = context.resources.datastore
+
+        parts = {
+            "daily_weather": {
+                "func": _extract_csv,
+                "condition": gpratk_settings.daily_weather,
```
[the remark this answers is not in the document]
zaneselvans: I'm not sure what you mean. It does live in the settings class, and this the place where a decision is made based on that setting.  If it's true, it gets processed. If it's false, it doesn't get processed.

src/pudl/metadata/fields.py
```diff
@@ -1776,6 +1784,10 @@
             "customers used to determine SAIDI and SAIFI."
         ),
     },
+    "include_generator": {
+        "type": "boolean",
+        "description": "Indicates whether a generator is included in the generation profile aggregation.",
+    },
```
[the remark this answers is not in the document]
zaneselvans: Every row in the aggregation file describes a single generator.  Groups of rows with the same `aggregation_key` are combined using a weighted average to produce an aggregate generation profile. For reasons that I don't fully understand, sometimes generators are excluded from that aggregation process, and this column is what determines whether a generator is excluded.

src/pudl/extract/gridpathratoolkit.py
zaneselvans: I'm not sure what you're thinking of here. What additional work are you saying should be done in the settings class? I'm not trying to verify that we've gotten valid settings here, I'm trying to control what processing happens based on the values found in the settings.

src/pudl/metadata/fields.py
```diff
@@ -64,6 +64,10 @@
             "capable of recording and transmitting instantaneous data."
         ),
     },
+    "aggregation_key": {
+        "type": "string",
+        "description": "A key that indicates on what basis the associated data was aggregated.",
+    },
```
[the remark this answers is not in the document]
zaneselvans: These values are defined in the aggregation files and they can really be anything -- it's just the label used for the group of generators that get aggregated together.  Right now it takes on values like `Wind_PSCO` or `Hybrid_Solar_62284` which happen to be descriptive.

src/pudl/metadata/resources/gridpathratoolkit.py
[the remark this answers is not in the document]
zaneselvans: We don't need it right now, and it was going to take a little bit of work to actually transform it into something tidy and normalized, and we're overtime already on this dataset, so I wanted to defer it until later.

src/pudl/metadata/resources/gridpathratoolkit.py
```diff
@@ -0,0 +1,52 @@
+"""Table definitions for data coming from the GridPath Resource Adequacy Toolkit."""
+
+from typing import Any
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "core_gridpathratoolkit__hourly_aggregated_extended_capacity_factors": {
```
cmgosnell, line 6: zane and I chatted about this name and came to `out_gridpathratoolkit__hourly_potential_capacity_factors` (potential or available either one! or some other more standard name for generation profiles).

because this really is a processed output table (we just happen to be building it in a lightweight ETL) but it is processed by GP.

src/pudl/metadata/resources/gridpathratoolkit.py
```diff
@@ -0,0 +1,52 @@
...
+            "This table defines which individual generator profiles are combined when "
+            "creating aggregated capacity factors / generation profiles. Generator "
+            "capacity is used to weight the contribution of each generator in the "
+            "resulting aggregated profiles."
+        ),
+        "schema": {
+            "fields": [
+                "plant_id_eia",
+                "generator_id",
+                "aggregation_key",
```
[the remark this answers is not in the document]
zaneselvans: Yes, I think that's right, and also probably not something we'd like to do.

src/pudl/metadata/resources/gridpathratoolkit.py
```diff
@@ -0,0 +1,52 @@
...
+                "capacity_factor",
+            ],
+            "primary_key": ["datetime_utc", "aggregation_key"],
+        },
+        "sources": ["gridpathratoolkit"],
+        "field_namespace": "gridpathratoolkit",
+        "etl_group": "gridpathratoolkit",
+    },
+    "core_gridpathratoolkit__capacity_factor_aggregations": {
```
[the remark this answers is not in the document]
cmgosnell: `core_gridpathratoolkit__assn_generator_aggregation_group` (with a renamed `aggregation_key` -> `aggregation_group`

src/pudl/extract/gridpathratoolkit.py
cmgosnell: - migrate logic about parts dict's `condition` into settings class. class will the compile the list of parts and verify that all of those parts are `working_partitions`.
- `gpratk_settings` in this function will only need to access `parts` (which is to say the parts you want to process not the working partitions)
- in `raw_gridpathratoolkit_assets` you'll still loop over the working_partitions, but in this factory, you will check in `gpratk_settings` if you are processing that part, if yes: process. if no, return empty df
- in this function, compile the list of parts to process via `_extract_csv` vs `_exctract_X`

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
cmgosnell: Can you add this information into the description?

src/pudl/settings.py
```diff
@@ -348,6 +347,60 @@ class GlueSettings(FrozenBaseModel):
...
+    # AGGREGATED = auto()
+    # ORIGINAL = auto()
+
+
+class GridPathRAToolkitSettings(GenericDatasetSettings):
+    """An immutable pydantic model to validate GridPath RA Toolkit settings."""
+
+    data_source: ClassVar[DataSource] = DataSource.from_id("gridpathratoolkit")
+    technology_types: list[str] = ["wind", "solar"]
+    processing_levels: list[str] = ["extended"]
```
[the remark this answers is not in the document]
bendnorman: Hmm I haven't run into this whitelisted error. Can you share the full stack trace?

src/pudl/transform/gridpathratoolkit.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    # Get rid of non-alphanumeric characters, and convert to lowercase
+    agg = pudl.helpers.simplify_columns(raw_agg)
+    # Split the eia_uniqueid into plant and generator IDs
+    # Should always be an integer (plant id) followed by an underscore, with anything
+    # after the underscore being the generator ID (a string).
+    separate_ids = agg["eia_uniqueid"].str.split("_", n=1, expand=True)
+    separate_ids.columns = ["plant_id_eia", "generator_id"]
+    separate_ids = separate_ids.astype({"plant_id_eia": int, "generator_id": "string"})
+    agg = pd.concat([agg, separate_ids], axis="columns")
```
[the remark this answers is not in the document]
zaneselvans: I agree that it'd be more readable, but it didn't work, and the issue wasn't immediately obvious, so I didn't put more time into it.

src/pudl/settings.py
```diff
@@ -348,6 +348,88 @@ class GlueSettings(FrozenBaseModel):
...
+    daily_weather: bool = True
+    parts: list[str] = []
```
cmgosnell, line 378: I think our default here across the other setting classes is to include all of the working partitions.

src/pudl/settings.py
```diff
@@ -348,6 +348,88 @@ class GlueSettings(FrozenBaseModel):
...
+        if "solar" in info.data["technology_types"] and (
+            "extended" in info.data["processing_levels"]
+            or "aggregated" in info.data["processing_levels"]
+        ):
+            parts.append("solar_capacity_aggregations")
+        if "wind" in info.data["technology_types"] and (
+            "extended" in info.data["processing_levels"]
+            or "aggregated" in info.data["processing_levels"]
+        ):
+            parts.append("wind_capacity_aggregations")
```
cmgosnell, line 429: this is not blocking I think it would be good to condense this logic a little bit so we don't need to add a new if block for every new technology and/or processing level. it might only be worth is as we add more but i think it'd be helpful and i'd be happy to suggest something if you'd like

src/pudl/transform/gridpathratoolkit.py
```diff
@@ -0,0 +1,161 @@
...
+    """Basic transformations that can be applied to many profiles.
+
+    - Construct a datetime column and adjust it to be in UTC.
+    - Reshape the table from wide to tidy format.
+    - Name columns appropriately.
+    """
+    # Convert timestamp to UTC before stacking!
+    # TODO: Ensure that we're using the same hour starting vs. ending convention
+    # here as we use in the FERC-714 and EPA CEMS data. Should this be reflected in
+    # the column names here and elsewhere, whatever the convention is?
```
cmgosnell, line 31: is this todo still relevant?

src/pudl/transform/gridpathratoolkit.py
```diff
@@ -0,0 +1,161 @@
...
+        .stack()
+        .reset_index()
+    )
+    capacity_factors.columns = ["datetime_utc", "aggregation_group", "capacity_factor"]
+    capacity_factors = capacity_factors.astype({"aggregation_group": "string"})
+    return capacity_factors
+
+
+@asset(io_manager_key="pudl_io_manager")
+def out_gridpathratoolkit__hourly_available_capacity_factor(
```
cmgosnell, line 49: can you add a note in here about why this is an `out_` asset even though it is in the transform module?

src/pudl/metadata/resources/gridpathratoolkit.py
```diff
@@ -0,0 +1,56 @@
...
+        "etl_group": "gridpathratoolkit",
+    },
+    "core_gridpathratoolkit__assn_generator_aggregation_group": {
+        "description": (
+            "This table defines which individual generator profiles are combined when "
+            "creating aggregated capacity factors / generation profiles. Generator "
+            "capacity is used to weight the contribution of each generator in the "
+            "resulting aggregated profiles, and is made available in this table for "
+            "convenience and legibility. The resulting aggregated profiles are stored "
+            "in :ref:`out_gridpathratoolkit__hourly_available_capacity_factor`."
```
cmgosnell, line 40: does :ref:`table_name` make sense in all of the contexts of where this table description is going to go? i don't think we are doing this anywhere else

cmgosnell (changes requested):
smol nits but this is looking really good! my only real blocking comment is about adding a comment about why we're building an `out_` asset in a transform module

src/pudl/transform/gridpathratoolkit.py
zaneselvans: Unfortunately it is still relevant but I punted on it here because it means digging into wtf is going on in our other hourly data which felt OOS for this PR & the GridLab work.

src/pudl/settings.py
zaneselvans: If you've got a clear suggestion sure!  Unfortunately the various combinations of technology & processing don't behave uniformly across the board.  E.g. there's no aggregated version of the thermal derates, but there is an extended version (it's just every single generator gets an extended timeseries), so it jumps straight from original to extended.  Are you thinking having 2 layers of nesting, one for tech type and one for processing level?

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
zaneselvans: Okay done.

src/pudl/metadata/resources/gridpathratoolkit.py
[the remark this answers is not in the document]
zaneselvans: Renamed!

src/pudl/metadata/resources/gridpathratoolkit.py
```diff
@@ -0,0 +1,52 @@
...
+        "sources": ["gridpathratoolkit"],
+        "field_namespace": "gridpathratoolkit",
+        "etl_group": "gridpathratoolkit",
+    },
+    "core_gridpathratoolkit__capacity_factor_aggregations": {
+        "description": (
+            "This table defines which individual generator profiles are combined when "
+            "creating aggregated capacity factors / generation profiles. Generator "
+            "capacity is used to weight the contribution of each generator in the "
+            "resulting aggregated profiles."
```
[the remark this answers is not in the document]
zaneselvans: Mentioned!

src/pudl/settings.py
zaneselvans: I guess that's true.

src/pudl/metadata/resources/gridpathratoolkit.py
zaneselvans: It'll show up as a link in the data dictionary, but not in the Datasette metadata. It looks like we've got a couple of references like this in the MCOE tables.  One advantage of making it a ref is that if the table name changes, the docs build will fail, and we'll be reminded to update the reference. But yes it will look a bit wonky in the datasette metadata.

src/pudl/transform/gridpathratoolkit.py
zaneselvans: Sure, I'll move it down here from the module docstring.

src/pudl/settings.py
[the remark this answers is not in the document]
zaneselvans: Here's what it looks like:

```
File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/dagster/_serdes/serdes.py", line 805, in _pack_value
    _pack_value(item, whitelist_map, f"{descent_path}[{idx}]")
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/dagster/_serdes/serdes.py", line 827, in _pack_value
    raise SerializationError(
dagster._serdes.errors.SerializationError: Can only serialize whitelisted Enums, received GPRATKTechType.
Descent path: <root:dict>.dataset_settings.config.gridpathratoolkit.technology_types[0]

The above exception was the direct cause of the following exception:

  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/dagster/_grpc/server.py", line 408, in __init__
    self._loaded_repositories: Optional[LoadedRepositories] = LoadedRepositories(
                                                              ^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/dagster/_grpc/server.py", line 257, in __init__
    with user_code_error_boundary(
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/dagster/_core/errors.py", line 297, in user_code_error_boundary
    raise new_error from e
dagster._core.errors.DagsterUserCodeLoadError: Error occurred during the loading of Dagster definitions in from pudl.etl import defs
/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/dagster/_core/workspace/context.py:622: UserWarning: Error loading repository location pudl.etl:dagster._core.errors.DagsterUserCodeLoadError: Error occurred during the loading of Dagster definitions in from pudl.etl import defs

Stack Trace:
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/dagster/_grpc/server.py", line 408, in __init__
    self._loaded_repositories: Optional[LoadedRepositories] = LoadedRepositories(
                                                              ^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/dagster/_grpc/server.py", line 257, in __init__
    with user_code_error_boundary(
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/dagster/_core/errors.py", line 297, in user_code_error_boundary
    raise new_error from e

dagster._serdes.errors.SerializationError: Can only serialize whitelisted Enums, received GPRATKTechType.
Descent path: <root:dict>.dataset_settings.config.gridpathratoolkit.technology_types[0]
```

src/pudl/metadata/resources/gridpathratoolkit.py
cmgosnell: okay yea thats a little weird in non-docs contexts but alas i do like that the docs build would fail :-/

src/pudl/transform/gridpathratoolkit.py
cmgosnell: okay... that's not idea but is definitely oos

cmgosnell (approved):
🎉 🤞🏻

src/pudl/settings.py
zaneselvans: Okay, gonna punt this until we add more partitions.

PR #3516, opened 2024-03-27
Extract EIA923 energy storage table

### LINKED ISSUE #3512: Compile column maps and other required metadata for the EIA923 energy storage operations table (exists in 2014-2023 data)

### LINKED ISSUE #3513: Extract EIA923 energy storage operations into a `raw_eia923__energy_storage` asset

aesharpe:
# Overview

Closes #3512 and closes #3513 

* Extracted the EIA923 Energy Storage tables that are archived but not yet integrated into PUDL

What did you change?

* Mapped all the extraction metadata for the operating and retired/canceled eia923 energy storage tables
* The `energy_storage` table was already in some of the mapping spreadsheets but not correctly, so I updated those values.

Notes/Question:

* It feels too soon to update the release notes seeing as we will do the rest of the integration tasks so soon.

# Testing

```[tasklist]
# To-do list
- [x] Run the `debug-column-mapping` notebook
```


src/pudl/package_data/eia923/column_maps/energy_storage.csv
```diff
@@ -0,0 +1,71 @@
+﻿year_index,2014,2015,2016,2017,2018,2019,2020,2021,2022,2023
+plant_id_eia,plant_id,plant_id,plant_id,plant_id,plant_id,plant_id,plant_id,plant_id,plant_id,plant_id
+is_combined_heat_power,combined_heat_and_power_plant,combined_heat_and_power_plant,combined_heat_and_power_plant,combined_heat_and_power_plant,combined_heat_and_power_plant,combined_heat_and_power_plant,combined_heat_and_power_plant,combined_heat_and_power_plant,combined_heat_and_power_plant,combined_heat_and_power_plant
```
aesharpe, line 3: I mention this in #3512, but this column shows up in other tables as just `combined_heat_power`. Based on how we've treated booleans in the previous table updates I think it makes sense for this to have the `is_` prefix. 

If you agree, I'll update it in the other table as well.

src/pudl/package_data/eia923/column_maps/energy_storage.csv
```diff
@@ -0,0 +1,71 @@
...
+electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity
+gross_generation_mwh,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours
+net_generation_mwh,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours
```
aesharpe, line 70: I mention this in #3512, but the `generation_fuel` table contains this same column with `megawatthours` written out instead of abbreviated. I think it should be abbreviated in both.

If you agree, I'll go ahead and change it in the `generation_fuel` table.

src/pudl/package_data/eia923/column_maps/energy_storage.csv
cmgosnell: yes i think that makes sense! although iirc there are several hard-coded column names in the transform step  and in the harvestable entity columns so if you change it be sure to make sure you go track down all of the threads. (also there are many boolean columns that don't conform w/ the `is_` prefix in other EIA tables... i'd love to change em all but that feels very oos for this)

src/pudl/package_data/eia923/column_maps/energy_storage.csv
cmgosnell: the `net_generation_megawatthours` is a weird holdover. the column is immediately dropped in `_core_eia_923__generation_fuel_eia923`. in both these tables, this column is the total for the whole year, where the `net_generation_mwh_{month}` is the monthly data. we convert the monthly columns into the mwh column via [_yearly_to_monthly_records](https://github.com/catalyst-cooperative/pudl/blob/b57dfad8cb7427815656e0bfb90d35bae012fc26/src/pudl/transform/eia923.py#L386).

I think if you name this `net_generation_mwh` I think it'll junk up the yearly to monthly process. Maybe we should name this `net_generation_mwh_total`? But regardless this guy is gonna be dropped.

cmgosnell (changes requested):
this looks grand! your two column name comments have follow-up requests but everything else looks great. I ran the extractor it all looked good to me

src/pudl/package_data/eia923/column_maps/energy_storage.csv
aesharpe: Yeah, I figure for now we should just fix 'em as they come up

aesharpe: It looks like there is a field called `associated_combined_heat_power` in the EIA860 generator tables. This field feels very similar, the only difference being that the EIA860 tables are reported at the `generator_id` level and the EIA923 tables are reported at the `plant_id_eia` level. This makes sense because generators are *associated* with combined heat and power power plants whereas plants *are* combined heat and power. I'm wondering if we should give these columns the same name.

aesharpe: Huh, it also seems like the `combined_heat_power` field doesn't get passed from the raw table to any of the transformed `generation_fuel` tables in PUDL. Not sure what to make of that.

src/pudl/package_data/eia923/column_maps/energy_storage.csv
aesharpe: Ok I like the idea of `mwh_total`. Do we vet the sum of the months with the total at all or do we just drop the total in favor of the summed months?

src/pudl/package_data/eia923/column_maps/generation_fuel.csv
```diff
@@ -93,6 +93,6 @@ total_fuel_consumption_quantity,total_fuel_consumption_quantity,total_fuel_consu
...
+net_generation_mwh_total,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours,net_generation_megawatthours
```
zaneselvans, line 96: We're not actually keeping this column in the end, right?  If we are, I think the units should be the last part of the name. based on how we've named other columns and [our naming conventions](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/naming_conventions.html#columns-and-field-names)

> Append units to field names where applicable (e.g. net_generation_mwh). This includes “per unit” signifiers (e.g. _pct for percent, _ppm for parts per million, or a generic _per_unit when the type of unit varies, as in columns containing a heterogeneous collection of fuels)

src/pudl/package_data/eia923/column_maps/energy_storage.csv
cmgosnell: we do not use the total column to check the monthly data. it would be a good idea overall but we definitely drop the annual total. checking would be good but oos for this for sure.

src/pudl/package_data/eia923/column_maps/energy_storage.csv
cmgosnell: it looks like it was renamed to `associated_combined_heat_power` at some point and the column from generation_fuel never got renamed. I think that's a fine name all in all and will be minimally invasive to name this new column `associated_combined_heat_power` and fix the gen fuel name

src/pudl/package_data/eia923/column_maps/generation_fuel.csv
cmgosnell: we are not keeping this column. but your right our conventions would imply `total_net_generation_mwh`. i think that makes sense.

src/pudl/package_data/eia923/column_maps/energy_storage.csv
aesharpe: agreed! good to know though

src/pudl/package_data/eia923/column_maps/generation_fuel.csv
aesharpe: just going to flag that we should also hypothetically convert all of the columns that say `net_generation_mwh_<MONTH>` by that same logic, but def OOS for this project (also they get combined into `net_generation_mwh` anyways.

For now I will change the column to `total_net_generation_mwh`

cmgosnell (approved):
looks good!

PR #3526, opened 2024-03-30
Transform and harvesting the eia860 Energy Storage table

### LINKED ISSUE #3506: Transform raw extracted EIA860 energy storage attributes into a clean `_core_eia860__energy_storage` asset for harvesting.

aesharpe:
# Overview

Closes #3506

* Adds a transformer for the EIA860 energy storage table
* Puts this table through the harvesting process

What did you change?

* Added a transformer for the storage table
* Added new fields for the storage table
* Added coding metadata for `core_eia__codes_storage_technology_types` and `core_eia__codes_storage_enclosure_types`
* Updated some column names to fix typos, better reflect column contents, naming protocols, and existing column names
* Add the `core_eia860__yearly_generators_energy_storage` to the list of finished EIA assets and resource metadata

Notes

* I noticed that some of the columns that get dropped from the specific generator tables have values that aren't in the general generators table. We should probably do a comparison before we drop them, but that is OOS.
* Looks like the boolean columns are coming out as 1/0 values in the `core` tables, need to figure out where to add the dtype fix to stop this.

# Testing

* Still need to run all the tests

```[tasklist]
# To-do list
```


src/pudl/metadata/fields.py
```diff
@@ -3060,6 +3070,11 @@
         "description": "Reactive Power Output (MVAr)",
         "unit": "MVAr",
     },
+    "reactive_power_rating": {
+        "type": "number",
+        "description": "The nameplate reactive power rating for the energy storage device.",
+        "unit": "MVAr",
+    },
```
aesharpe, line 3077: I think this is the same as the existing `reactive_power_output_mvar` column, but I'm not 100% sure.

cmgosnell (changes requested):
It looks like you missed one code step, which is to define those tables as db tables. there are many examples of that in `metadata.resources.eia`. This could be the cause of your pydantic error.

You also need to add the _core table in the tables to be harvested (see comment in `transform.eia`

>     I noticed that some of the columns that get dropped from the specific generator tables have values that aren't in the general generators table. We should probably do a comparison before we drop them, but that is OOS.

I'm not sure what you mean here. This maybe could be related to the previous harvest comment.

>     Looks like the boolean columns are coming out as 1/0 values in the core tables, need to figure out where to add the dtype fix to stop this.

Are you seeing this in the `_core` table or in the `core` table? the `core` table is definitely getting dtyped with the dypes defined in the schema.

(i haven't tried to build these assets yet)

src/pudl/transform/eia860.py
```diff
@@ -356,6 +356,70 @@ def _core_eia860__generators(
...
+        "served_co_located_renewable_firming",
+        "served_frequency_regulation_applications",
+        "served_load_following_applications",
+        "served_load_management_applications",
+        "served_ramping_spinning_reserve_applications",
+        "served_system_peak_shaving_applications",
+        "served_transmission_and_distribution_deferral_applications",
+        "served_voltage_or_reactive_power_support_applications",
+        "stored_excess_wind_and_solar_generation",
+    ]
```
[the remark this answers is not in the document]
aesharpe: I was torn about this because I sort of thought the opposite - that if a column name changed it would be easier to flag if the column name was fully written out (i.e. searchable in the code base) as opposed to found via `filter` and prefix codes. But I don't feel super strongly here. Better to be consistent.

src/pudl/transform/eia.py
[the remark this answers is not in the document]
aesharpe: good catch!

src/pudl/transform/eia860.py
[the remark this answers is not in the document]
aesharpe: It's definitely prettier not to have them written out

aesharpe:
> Are you seeing this in the `_core` table or in the `core` table? the `core` table is definitely getting dtyped with the dypes defined in the schema.

Yeah, both tables just have 0 or 1 values instead of True/False

src/pudl/metadata/fields.py
```diff
@@ -3329,6 +3339,46 @@
...
+        "type": "boolean",
+        "description": "Whether the energy storage device served arbitrage applications during the reporting year",
+    },
+    "served_backup_power_applications": {
+        "type": "boolean",
+        "description": "Whether the energy storage device served backup power applications during the reporting year.",
+    },
+    "served_co_located_renewable_firming": {
+        "type": "boolean",
+        "description": "Whether the energy storage device served renewable firming applications during the reporting year.",
```
zaneselvans, line 3352: Why does this boolean column not end in `_applications` while all the others do?

If they *all* end in `_applications` then is that suffix adding useful information to the column name, or could we drop it and have shorter names?

src/pudl/metadata/resources/eia860.py
```diff
@@ -802,6 +802,47 @@
         "sources": ["eia860"],
         "etl_group": "eia860",
     },
+    "core_eia860__yearly_generators_energy_storage": {
```
zaneselvans, line 805: This is another table that seems like it might currently be a mix of slowly changing dimensions and yearly data. Do the boolean values really change every year in general? Or are they stable descriptors of what role the battery plays longer term?

src/pudl/transform/eia860.py
```diff
@@ -356,6 +356,57 @@ def _core_eia860__generators(
...
+    * A lot of the values that are repeated in the generators table are dropped post
+    harvesting. It would be nice to compare the values before dropping them to make sure
+    these specific generator tables don't contain information that the big generator
+    tables does not.
+    """
+    storage_ex = raw_eia860__generator_energy_storage_existing
+    storage_re = raw_eia860__generator_energy_storage_retired
+
+    # every boolean column in the raw storage tables has a served_ or stored_ prefix
+    boolean_columns_to_fix = list(storage_ex.filter(regex="served_|stored_"))
```
zaneselvans, line 376: To ensure that it only matches at the beginning of the column name

```suggestion
    boolean_columns_to_fix = list(storage_ex.filter(regex="^(served_|stored_)"))
```

aesharpe: oh great fix

src/pudl/metadata/resources/eia860.py
aesharpe: I just did a little review and it looks like the boolean values are very stable SCD values. Should I turn this into two tables? an SCD and a yearly table?

aesharpe: It looks like the other non-boolean columns are also mostly stable, but not as stable as the booleans. What is the threshold for SCD?

src/pudl/metadata/fields.py
aesharpe: hmm I think this one might have slipped through the cracks. The reason I added `application` as a suffix was mostly for readability / comprehension. The columns are kind of weird and I figured it makes more sense to have _verb, adjective, noun_ than just _verb, adjective_. The verb `served` isn't ideal for clarity, but I can't really think of a better one here.

src/pudl/metadata/resources/eia860.py
cmgosnell: we don't have a set threshold but if most of the columns are mostly stable i'd say go for it on the scd name

aesharpe:
> Yeah, both tables just have 0 or 1 values instead of True/False

Realized that this is because I was viewing them in SQL where the dtypes were dropped

src/pudl/transform/eia860.py
```diff
@@ -410,6 +410,52 @@ def _core_eia860__generators_solar(
...
+    raw_eia860__generator_energy_storage_existing: pd.DataFrame,
+    raw_eia860__generator_energy_storage_retired: pd.DataFrame,
+) -> pd.DataFrame:
+    """Transform the energy storage specific generators table.
+
+    Notes for possible future cleaning:
+    * A lot of the values that are repeated in the generators table are dropped post
+    harvesting. It would be nice to compare the values before dropping them to make sure
+    these specific generator tables don't contain information that the big generator
+    tables does not.
```
cmgosnell, line 424: the harvested columns are integrated into the generators/plants table which is subtly different than being dropped. They don't end up in the storage table, but that is because they are slurped up and associated with that generator. If there was substantial difference between in the values in this table and the other tables, there would be a failure in the harvesting step.

cmgosnell (approved):
I second zane's suggestion to remove the `_application` suffix but that is not blocking for me. also one docs comment that is also not blocking.

zaneselvans:
Yeah, SQLite stores booleans as integers. It's generally kinda lazy about dtypes.

If you want to see rich dtypes, you can always read from the parquet files, which is what you'll get if you use  something like 

```py
raw_weather = defs.load_asset_value(AssetKey("raw_gridpathratoolkit__daily_weather"))
```

aesharpe:
> I second zane's suggestion to remove the `_application` suffix but that is not blocking for me. also one docs comment that is also not blocking.

Happy to change if it's 2:1

Issue #423, opened 2019-09-23
generators_entity_eia operating_date field blank when op dates disagree


gschivley:
### Describe the bug
I created a PUDL sqlite database with EIA860 and 923 data from 2011-2017. At least one plant (55177) has a NULL `operating_date` value in the for all generators in `generators_entity_eia` table. All years show that generators came online in 2001 but the files split (3 to 3) about if it was in March or June. 

### Bug Severity
How badly is this bug affecting you?
 - **Medium:** With some effort, I can work around the bug.

I use `operating_date` to determine a retirement year for power plants. Null values mean that plants don't have a retirement date, so I need to go back and add them to my settings file by hand.

### To Reproduce
Build the datapackages with 860 data from 2011-2017.

### Expected behavior
I'm not quite sure what the expected behavior should be. In this specific case the year of operation is consistent across time despite the changing month. Maybe use the most recent value instead of NULL?

zaneselvans:
Right now, this is the expected behavior -- when any of the supposedly constant fields in 860 is too inconsistent, it gets marked NA/Null, since we want to have a single value associated with these long-lived entities (plants, generators, etc.) but it's not clear what that value should be, in the general case. Is there any reason to think that the most recent value is more likely to be correct than the older values?

In this particular case, it seems like as long as the dates are close to each other they should be considered consistent (like we do with lat/lon, where they're rounded to the nearest 0.01 degrees or something). This would happen in the harvesting functions that @cmgosnell set up.

gschivley:
I agree that it’s difficult to generalize rules for these disagreements. The fact that it changes and is then consistent makes me trust the new values a bit more. 

I’ll take a look to see which other plants have values that change over time.

grgmiller:
I just wanted to re-flag this issue. I'm noticing missing operating dates for a specific plant (plant 136, generator 2). In the 2022 EIA-860 data, this has an operating date of December 1984, but as far as I can tell this date is missing from pudl. It looks like back in 2005 this generator had an operating date of January 1985, and in more recent versions of 860, it has a reported operating date of December 1984. 

Regardless of the ability to create a single static generator_operating_date, it would be nice if there was a way to access any data that exists in the original raw EIA spreadsheets through PUDL, rather than that data being deleted. (Maybe this does exist in one of the output tables and I just haven't been able to find it yet). 

[...]

PR #3529, opened 2024-04-02
Schema cleanup

zaneselvans:
# Overview

Some miscellaneous column mapping cleanup, mostly for the entity-specific columns which get turned into `operating_date` values. Things I came across while looking into a better fix for #423.

* Map the `operating_year` and `operating_month` columns in the EIA860M retired generators consistently -- these were getting silently lost because  the other EIA860 generator operating date columns are name `generator_operating_date`.
* Ensure all `*_(operating|retired)_(year|month)` columns have specific names in column  mapping metadata CSVs, and remove the manual renames that were happening elsewhere  to ensure that there aren't a bunch of different operating_date and retirement_date  columns floating around in the transforms which actually apply to different entities. This change affected the operating month/year/date columns for cooling_system, emission_control, fgd, and fgp.
* Put the `convert_to_date()` and `month_year_to_date()` helper functions next to each other in their module... they really need to be combined!
* Rename `obstacle_num_faa` to `obstacle_id_faa` and update field description to provide more context on what those IDs refer to.
* Rename `core_eia860__yearly_generators_wind` to `core_eia860__scd_generators_wind`, as it's not a data table, but a generator attribute table.
* Add a new alembic migration to deal with the wind table and FAA column renames.

# Testing

* I ran the full ETL locally and didn't get any failures due to bad / missing column or table names.
* I ran `make pytest-validate` on the fully populated database and it passed.
* I ran `make pytest-integration-full` on the fully populated database and it passed.

```[tasklist]
# To-do list
- [x] Update the release notes: reference the PR and related issues.
```


cmgosnell (approved):
📜 🧼 👍🏻

PR #3532, opened 2024-04-03
Breakdown total `utility_type` and partial `in_rate_base` in rate base table

### LINKED ISSUE #3530: break down of `utility_type=="total"` and null/partial in_rate_base records
## Goal:
- Have all records feeding into the rate base table be broken down into specific `utility_type`'s based on the annual utility report break down. 
- Have all partial (or null) `tags_in_rate_base` records be broken down into either in rate base or not in rate base based on the annual utility break down.

## Questions
The first question is about ordering of these breakdowns: Should we split up the null in_rate_base records first and then split the total utility_type records but do the latter using only the records that are in rate base (including the null records that got broken down)? Or should we do these breakdowns independently (i.e. calculate a in/out of rate base ratio and then calculate a utilty_type ratio and apply that)? I set it up to do the `utility_type` breakdown first and then the `in_rate_base` breakdown. Intuitively I don't think it should matter what order this happens in but it was just something I wanted to flag.

Negative values. A lot of the weirdness here resulted from negative values for the utility/year grouped `ending_balance`. I was able to get to a process that felt reasonable via `get_split_col_ratio` but I'd like to just double check that that methodology is what is desired here.

## Remaining TODOS -> now both done
- DONE - I expect the new `ending_balance` to be the same pre and post this treatment but it's off... only by a small amount but that could be illuminating a real problem:`New ending balance is not the same as the old ending balance: new_balance=2842690360221.375, old_balance=2842635943306.375`. fixed by fillna(1) on the null ratios (some util/years only have unlabeled records)
- DONE - Doing the breakdown was junking up 2021 (total in rate base $$ spiked). i suspect its from something from the 2021 subtotal correction tags: this did indeed get fixed.

### LINKED ISSUE #3534: Second Review of Rate Base
Hi again @jrea-rmi  let's do this review again! Follow up #3479

### This that are still in progress:
- #3309
- #3530 (the files below have a first draft of breaking down the utility type but not the null rate base tags.. progress in #3532)

Jon, if you'd rather wait until those two things are wrapped up just let us know!

### Files:
- [out_ferc1__yearly_rate_base.pkl.zip](https://github.com/catalyst-cooperative/pudl/files/14856137/out_ferc1__yearly_rate_base.pkl.zip)
- [out_ferc1__detailed_balance_sheet_assets.pkl.zip](https://github.com/catalyst-cooperative/pudl/files/14856162/out_ferc1__detailed_balance_sheet_assets.pkl.zip)
- [out_ferc1__detailed_balance_sheet_liabilities.pkl.zip](https://github.com/catalyst-cooperative/pudl/files/14856164/out_ferc1__detailed_balance_sheet_liabilities.pkl.zip)
- [forest_as_table_core_ferc1__yearly_balance_sheet_assets_sched110.pkl.zip](https://github.com/catalyst-cooperative/pudl/files/14856165/forest_as_table_core_ferc1__yearly_balance_sheet_assets_sched110.pkl.zip)
- [forest_as_table_core_ferc1__yearly_income_statements_sched114.pkl.zip](https://github.com/catalyst-cooperative/pudl/files/14856166/forest_as_table_core_ferc1__yearly_income_statements_sched114.pkl.zip)

### Plotz
image
image
by input table & utility
image
image
image
image
image

cmgosnell:
# Overview

Closes #3530.

Also closes #3534:
- added more `in_rate_base` and `rate_base_category` tags
- began propagating the `rate_base_category` tag

Extra:
- there were lots of warnings and error messages in the logs that are now no longer popping up bc of simple reordering of things or silencing things when its not really a problem any more.

# Testing

```[tasklist]
# To-do list
```


src/pudl/output/ferc1.py
```diff
@@ -3012,15 +3012,55 @@ def out_ferc1__yearly_rate_base(
...
+    in_rate_base = rate_base[rate_base.tags_in_rate_base == "yes"]
+    # note: we need the `tags_in_rate_base` column for these checks
+    check_tag_propagation_compared_to_compiled_tags(
+        in_rate_base, "in_rate_base", _out_ferc1__detailed_tags
+    )
+    check_for_correction_xbrl_factoids_with_tag(in_rate_base, "in_rate_base")
+    return in_rate_base
+
+
+def prep_cash_working_capital(
```
cmgosnell, line 3053: i moved this stuff out of the asset function mostly for readability now that we are doing more than this + concat.

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3084,112 @@ def out_ferc1__yearly_rate_base(
...
+    )
+    return rate_base_broken_down.convert_dtypes()
+
+
+def get_split_col_ratio(rate_base_labeled, ratio_idx, split_col) -> pd.DataFrame:
+    """Calculate the percentage of the total labeled ``ending_balance`` within each tag group.
+
+    Make ratio column with a 0-1 value of the sum of ``ending_balance`` in each labled
+    ``split_col`` within ``ratio_idx``.
+    """
```
cmgosnell, line 3128: i had a hard time explaining these steps..

src/pudl/output/ferc1.py
```diff
@@ -3012,15 +3012,55 @@ def out_ferc1__yearly_rate_base(
...
+    core_ferc1__yearly_operating_expenses_sched320,
+) -> pd.DataFrame:
+    """Extract a new ``cash_working_capital`` xbrl_factoid for the rate base table.
+
+    In standard ratemaking processes, utilities are enabled to include working
+    capital - sometimes referred to as cash on hand or cash reverves. A standard
+    ratemaking process is to consider the available rate-baseable working capital to
+    be one eigth of the average operations and maintenance expense. This function
+    grabs that expense and in preparation to concatenate it with the rest of the
+    assets and liabilities from the granular exploded data.
```
aesharpe, line 3063: A couple little readability notes:
- "are enabled to include" sounds a little funny, maybe "are allowed to include"
- _Where_ are they able to include the cash reserves, the rate base? Would be nice to add that to the end of the first sentence.
- "is to consider" feels a little clunky, maybe just "considers"
- "in preparation" -- should this be "prepares it" ?

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3084,112 @@ def out_ferc1__yearly_rate_base(
         # the assets/liabilites both use ending_balance for its main $$ column
         .rename(columns={"dollar_value": "ending_balance"})
     )
-    # then select only the leafy exploded records that are in rate base and concat
-    in_rate_base = pd.concat(
+    return cash_working_capital
+
+
+def breakdown_unlabeled(
+    rate_base: pd.DataFrame,
```
aesharpe, line 3091: This should probably have the `_df` suffix. Also applicable elsewhere/more broadly

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3084,112 @@ def out_ferc1__yearly_rate_base(
...
+    split_col: str,
+) -> pd.DataFrame:
+    """Breakdown rate base records with an unlabeled split_col.
+
+    Args:
+        rate_base: full table of rate base data.
+        split_col: column with the label that contains unlabeled values
+            (ex: null or total).
+    """
+    # the works for both the util type and the in rate base tags because
```
aesharpe, line 3101: Should this say "this works"? Also, should probably be `in_rate_base` and `util_type`

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3084,112 @@ def out_ferc1__yearly_rate_base(
...
+    # for both tags: unlabeled is represented as a null and but just util type:
```
aesharpe, line 3102: "and but" doesn't make sense, maybe a typo?

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3084,112 @@ def out_ferc1__yearly_rate_base(
...
+    unlabeled_mask = (rate_base[split_col] == "total") | rate_base[split_col].isnull()
+    ratio_idx = ["report_year", "utility_id_ferc1"]
+    ratio_df = get_split_col_ratio(
+        # remove the unlabeled records because total is the value
+        # we want to breakdown so we can't have it in the columns to sum up
+        rate_base[~unlabeled_mask],
+        ratio_idx=ratio_idx,
+        split_col=split_col,
+    )
+    rate_base_broken_down = apply_ratio_to_breakdown_unlabeled(
```
aesharpe, line 3113: Should have `_df` suffix

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3084,112 @@ def out_ferc1__yearly_rate_base(
...
+
+def breakdown_unlabeled(
+    rate_base: pd.DataFrame,
+    split_col: str,
+) -> pd.DataFrame:
+    """Breakdown rate base records with an unlabeled split_col.
+
+    Args:
+        rate_base: full table of rate base data.
+        split_col: column with the label that contains unlabeled values
```
aesharpe, line 3098: Not sure what you mean by "column with the label that contains unlabeled values". I guess consequently unsure why it's called split_col. Might be nice to include the word split in the definition to help explain its significance.

src/pudl/output/ferc1.py
aesharpe: It might be helpful to explain more of the context in the description?

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3084,112 @@ def out_ferc1__yearly_rate_base(
...
+def apply_ratio_to_breakdown_unlabeled(
+    rate_base: pd.DataFrame,
+    ratio_df: pd.DataFrame,
+    unlabeled_mask: pd.Series,
+    ratio_idx: list[str],
+    split_col: str,
+):
+    """For the unlabeled tag records, broadcast labeled tags and allocate the ``ending_balance``.
+
+    Apply ratios from :func:`get_split_col_ratio` to ``ending_balance``.
```
aesharpe, line 3160: Does this mean that you're allocating `ending_balance` based on the ratio assigned to that row?

src/pudl/output/ferc1.py
```diff
@@ -3012,15 +3012,55 @@ def out_ferc1__yearly_rate_base(
...
-    the available rate-baseable working capital to be one eigth of the average
-    operations and maintenance expense. This function grabs that expense and
-    concatenates it with the rest of the assets and liabilities from the granular
-    exploded data.
+    This rate base table also contains one new "cash_working_capital" xbrl_factoid
+    from :ref:`core_ferc1__yearly_operating_expenses_sched320` via
+    :func:`prep_cash_working_capital`.
 
+    We also break down records that have nulls or totals in two of the key tag
+    columns: ``tags_aggregatable_utility_type`` and ``tags_in_rate_base`` via
```
aesharpe, line 3020: I don't totally understand what you mean by "break down" here

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3084,112 @@ def out_ferc1__yearly_rate_base(
...
-    # then select only the leafy exploded records that are in rate base and concat
-    in_rate_base = pd.concat(
+    return cash_working_capital
+
+
+def breakdown_unlabeled(
+    rate_base: pd.DataFrame,
+    split_col: str,
+) -> pd.DataFrame:
+    """Breakdown rate base records with an unlabeled split_col.
```
aesharpe, line 3094: I think this should be "break down" not breakdown (also note for function name, maybe should be `break_down_unlabeled`: 

> with breakdown, you are describing an analysis or report, while when using break down, you are describing a collapse or sudden failure.

src/pudl/output/ferc1.py
```diff
@@ -3012,15 +3012,55 @@ def out_ferc1__yearly_rate_base(
...
 
-    This rate base table also contains one specific addition from
-    :ref:`core_ferc1__yearly_operating_expenses_sched320`. In standard ratemaking
-    processes, utilities are enabled to include working capital - sometimes referred
-    to as cash on hand or cash reverves. A standard ratemaking process is to consider
-    the available rate-baseable working capital to be one eigth of the average
-    operations and maintenance expense. This function grabs that expense and
-    concatenates it with the rest of the assets and liabilities from the granular
-    exploded data.
+    This rate base table also contains one new "cash_working_capital" xbrl_factoid
```
aesharpe, line 3015: Be consistent in how you reference `rate base` I saw it written as `rate-base` somewhere else in the code (line 3006)

aesharpe: Also, would be nice to explain either here or in the `prep_cash_working_capital` function doc string why you need to add a new factoid. Is it not there in the first place? Are you deriving it from other values? Why isn't it there?

aesharpe (commented):
For right now it's mostly comments on the concept / explanation vs. the code itself. I can go through and walk through each of the code steps if that's helpful.

High level thoughts:
- use `_df` suffix
- use `break down` instead of `breakdown` unless I'm understanding it wrong

src/pudl/output/ferc1.py
cmgosnell: so fair!.. i think the correct word here is actually "disaggregate". We have records with totals and/or nulls for these key tags which we want to separate those records into component parts.

src/pudl/output/ferc1.py
cmgosnell: this is great ty!!

src/pudl/output/ferc1.py
cmgosnell: okay i added a comment about the fact that it isn't reported. but i don't really know why it isn't reported! so imma leave that bit out.

src/pudl/output/ferc1.py
cmgosnell: i changed this to disaggregate bc i think that is more specific than break down

src/pudl/output/ferc1.py
cmgosnell: yes! I ended up removing this whole function and moving this content up into the main function above and explaining it above.

src/pudl/output/ferc1.py
cmgosnell: yes your questions and suggestions were very helpful to nudge me to clarify what is going on in here. I updated all of the docs.

src/pudl/output/ferc1.py
```diff
@@ -2288,10 +2288,12 @@ def propagate_node_attributes(self: Self, annotated_forest: nx.DiGraph):
...
         Propagate tags leafwards, rootward &  to the _correction nodes.
         """
+        tags_to_propagate = ["in_rate_base", "rate_base_category"]
         ## Leafwards propagation
-        annotated_forest = _propagate_tags_leafward(annotated_forest, ["in_rate_base"])
+        annotated_forest = _propagate_tags_leafward(annotated_forest, tags_to_propagate)
         # Rootward propagation
-        annotated_forest = _propagate_tag_rootward(annotated_forest, "in_rate_base")
+        for tag in tags_to_propagate:
+            annotated_forest = _propagate_tag_rootward(annotated_forest, tag)
```
cmgosnell, line 2296: this adds the `rate_base_category` into the set of tags that we are propagating up and down the forest.

src/pudl/package_data/ferc1/xbrl_factoid_rate_base_category_tags.csv
cmgosnell, line 1: two additions of tags thanks for jon from #3534 (in this file and next tag file)

src/pudl/output/ferc1.py
```diff
@@ -2274,36 +2274,45 @@ def annotated_forest(self: Self) -> nx.DiGraph:
...
         nx.set_edge_attributes(annotated_forest, self.edge_attrs)
-        annotated_forest = self.propagate_node_attributes(annotated_forest)
 
         logger.info("Checking whether any pruned nodes were also tagged.")
         self.check_lost_tags(lost_nodes=self.pruned)
         logger.info("Checking whether any orphaned nodes were also tagged.")
         self.check_lost_tags(lost_nodes=self.orphans)
-        self.check_conflicting_tags(annotated_forest)
+
+        annotated_forest = self.propagate_node_attributes(annotated_forest)
```
cmgosnell, line 2283: i moved the propagation of tags to be after the checking of lost tags because that check doesn't make sense after the tags have been propagated. i added a note in the `check_lost_tags` method about why.

src/pudl/output/ferc1.py
```diff
@@ -2274,36 +2274,45 @@ def annotated_forest(self: Self) -> nx.DiGraph:
...
-        ## Leafwards propagation
-        annotated_forest = _propagate_tags_leafward(annotated_forest, ["in_rate_base"])
-        # Rootward propagation
-        annotated_forest = _propagate_tag_rootward(annotated_forest, "in_rate_base")
+        tags_to_propagate = ["in_rate_base", "rate_base_category"]
+        # Root-ward propagation
+        for tag in tags_to_propagate:
+            annotated_forest = _propagate_tag_rootward(annotated_forest, tag)
+        ## Leaf-wards propagation
+        annotated_forest = _propagate_tags_leafward(annotated_forest, tags_to_propagate)
```
cmgosnell, line 2299: this change does two things:
* adds another tag that we are now propagating
* sends the tag from the leaves towards the roots first (see the docs above for why)

src/pudl/output/ferc1.py
```diff
@@ -2274,36 +2274,45 @@ def annotated_forest(self: Self) -> nx.DiGraph:
...
+        It is not necessarily a problem if there are "lost" tags. This is mostly
+        here as a debugging tool.
+        """
         if lost_nodes:
             lost = pd.DataFrame(lost_nodes).set_index(self.calc_cols)
             tagged = self.tags.set_index(self.calc_cols)
             lost_tagged = tagged.index.intersection(lost.index)
             if not lost_tagged.empty:
-                logger.warning(
+                logger.debug(
```
cmgosnell, line 2315: this change is just to silence this. its not a bad check overall but it definitely isn't an error and it prints out a whole df which is helpful for debugging only

src/pudl/output/ferc1.py
```diff
@@ -2274,36 +2274,45 @@ def annotated_forest(self: Self) -> nx.DiGraph:
...
         return annotated_forest
 
     def propagate_node_attributes(self: Self, annotated_forest: nx.DiGraph):
         """Propagate tags.
 
-        Propagate tags leafwards, rootward &  to the _correction nodes.
+        Propagate tags root-ward, leaf-wards &  to the _correction nodes. We
+        propagate the tags root-ward first because we primarily manually
+        compiled tags for the leaf nodes, so we want to send those leafy tags
+        root-ward first before trying to send tags back up the graph.
```
aesharpe, line 2292: are root-ward and leaf-ward a play on "forward" and "backward"? It's a little confusing. Maybe say "starting from X" or "from X to Y". I'm not sure I understand the sentences with those phrases.

aesharpe: It looks like these are commonly used terms in the module. Maybe it's not worth changing all of them? As long as there is a clear definition somewhere.

aesharpe: A clarifying question - why are manually compiled tags prioritized?

src/pudl/output/ferc1.py
```diff
@@ -3012,15 +3016,64 @@ def out_ferc1__yearly_rate_base(
...
+    assets = _out_ferc1__detailed_balance_sheet_assets
+    liabilities = _out_ferc1__detailed_balance_sheet_liabilities.assign(
+        ending_balance=lambda x: -x.ending_balance
+    )
+    cash_working_capital = prep_cash_working_capital(
+        core_ferc1__yearly_operating_expenses_sched320
+    )
+
+    # concat then select only the leafy exploded records that are in rate base
+    rate_base = (
```
aesharpe, line 3036: add `_df` suffix

src/pudl/output/ferc1.py
```diff
@@ -3012,15 +3016,64 @@ def out_ferc1__yearly_rate_base(
...
+    )
+
+    in_rate_base = rate_base[rate_base.tags_in_rate_base == "yes"]
+    # note: we need the `tags_in_rate_base` column for these checks
+    check_tag_propagation_compared_to_compiled_tags(
+        in_rate_base, "in_rate_base", _out_ferc1__detailed_tags
+    )
+    check_for_correction_xbrl_factoids_with_tag(in_rate_base, "in_rate_base")
+    return in_rate_base.dropna(subset=["ending_balance"])
+
```
aesharpe, line 3061: Could these checks become asset checks?? I know nothing about this....

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3098,109 @@ def out_ferc1__yearly_rate_base(
         # the assets/liabilites both use ending_balance for its main $$ column
         .rename(columns={"dollar_value": "ending_balance"})
     )
-    # then select only the leafy exploded records that are in rate base and concat
-    in_rate_base = pd.concat(
+    return cash_working_capital
```
aesharpe, line 3101: Add `_df` suffix

src/pudl/output/ferc1.py
```diff
@@ -3036,6 +3089,7 @@ def out_ferc1__yearly_rate_base(
         .assign(
             dollar_value=lambda x: x.dollar_value.divide(8),
             xbrl_factoid="cash_working_capital",  # newly definied (do we need to add it anywhere?)
+            tags_in_rate_base="yes",
```
aesharpe, line 3092: Can we get rid of this comment?

src/pudl/output/ferc1.py
```diff
@@ -3036,6 +3089,7 @@ def out_ferc1__yearly_rate_base(
         .assign(
             dollar_value=lambda x: x.dollar_value.divide(8),
```
aesharpe, line 3090: Could we just name this `ending_balance` here so we don't have to use `.rename(columns={"dollar_value": "ending_balance"}` a few lines down?

aesharpe (commented):
I read through the relevant parts of`pudl/output/ferc1.py` to wrap my head around the rate base table, and I have some additional comments that don't pertain directly to code edited in this PR. Here are those thoughts: 

**1. Tags**

Is there a place where the term “tag” is more explicitly defined? It doesn't fit into the “tree” metaphor so it’s harder to discern what it is. In theory, tag = a label for something, but that seems pretty vague. It would help to clarify the meaning of "tag" somewhere in the docs. Maybe in `_out_ferc1__detailed_tags` or `_get_tags`

It seems like sometimes you use the term “label” and sometimes you use the term “tag” and I can’t tell if they are supposed to be interchangeable or not.

**2. The Tree Metaphor**

(Starting line 2051)
```
Several individual values from the higher level
tables will appear as root nodes at the top of each hierarchy, and the leaves in
the underlying tree structure are the individually reported non-calculated values
that make them up.
```
The way this passage uses the tree metaphor is confusing. It refers to roots “at the top” and leaves as “underlying structure” — is that an accidental flip-flop? Roots are at the bottom and leaves are at the top. Are “high-level” tables leafy tables? Or are “high-level” tables root tables? The language here is playing tricks on my brain. It would be ideal to use terms that relate to or evoke the part of the tree that each table is referring to. The word “granular” feels like it should be low to the ground like “roots” but I think it's used in reference to the leaves (because it’s the least aggregated data?). But maybe I’m getting that wrong. The whole pipeline seems like a backwards tree…lots of roots that lead to one big chunky tree trunk? Or maybe it’s a few core roots leading to a zillion leaves? The roots-to-leaves metaphor is hard because there are often many of both of them. Is that intended? More clarification would be helpful.

I’m also having trouble grasping the “seed” and “forest” references. What is the difference between a seed and a root? It feels a little too abstracted.

**3. Unit Tests**

Do you want/plan to write any unit tests for this table?

src/pudl/output/ferc1.py
cmgosnell: yes it is. the `_propagate_tag_rootward` tags one `tag` while the `_propagate_tags_leafward` tags a list of tags

src/pudl/output/ferc1.py
cmgosnell: yes the tree, root, leaf terminology is more or less grafted directly from [networkx](https://networkx.org/documentation/stable/reference/algorithms/tree.html) which is the primary tool we are using to grab the most granular components of the financial tables to feed into the rate base table.

the tags are entirely manually compiled but we don't manually compile the tags for ever single `xbrl_factoid` across the detailed tables because that would be *alot*. We use this propagation step to spread the tags we have compiled.

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3098,109 @@ def out_ferc1__yearly_rate_base(
...
+
+
+def disaggregate_unlabeled_or_total_tag(
+    rate_base_df: pd.DataFrame,
+    tag_col: str,
+) -> pd.DataFrame:
+    """Disaggregate records with an null or total in the ``tag_col``.
+
+    We have records in the rate base table with totals and/or nulls for key tag
+    columns which we want to separate into component parts. This is done in two
```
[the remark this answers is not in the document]
cmgosnell: the whole point of identifying the leaves of the trees (which is all work done to make these `out_ferc1_detailed_*` assets that are inputs into the rate base table) is to get only the granular, unique xbrl_factoids. So any total record that is comprised of other non-total sub-components has already been removed when we selected the leaves of the tree.

src/pudl/output/ferc1.py
```diff
@@ -3012,15 +3016,64 @@ def out_ferc1__yearly_rate_base(
...
+    core_ferc1__yearly_operating_expenses_sched320,
+) -> pd.DataFrame:
+    """Extract a new ``cash_working_capital`` ``xbrl_factoid`` for the rate base table.
+
+    In standard ratemaking processes, utilities are allowed to include working
+    capital - sometimes referred to as cash on hand or cash reverves - in their rate
+    base. A standard ratemaking process considers the available rate-baseable working
+    capital to be one eigth of the average operations and maintenance expense. This
+    function grabs that expense and prepares to concatenate it with the rest of the
+    assets and liabilities from the detailed rate base data.
```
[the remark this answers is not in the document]
cmgosnell: this function calculates the cash working capital by calculating the "operating & maintenance" * 1/8 = "cash working capital" so it can be added to the rest of the rate base data.

this function preps this new `xrbl_factoid` so we can add it with the other rate base data

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3098,109 @@ def out_ferc1__yearly_rate_base(
...
+    * use this ratio to disaggregate the ``ending_balance`` from records with total
+      and null tags across the component tags.
+
+    Args:
+        rate_base_df: full table of rate base data.
+        tag_col: column with the tags that contains unlabeled values
+            (ex: null or total).
+    """
+    # this works for both the utility_type and in_rate_base tags because
+    # for both tags unlabeled is represented as a null and/or a "total"
```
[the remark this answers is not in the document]
cmgosnell: null != total, but these are the records we want to disaggregate. i've somewhat off-handedly called them "unlabeled" but as you pointed out that is a little confusing.

null & total records are the ones we want to disaggregate. I personally think the doc string explains this better than this note does

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3098,109 @@ def out_ferc1__yearly_rate_base(
...
+            rate_base_df[unlabeled_mask],
+            ratio_df,
+            on=ratio_idx,
+            how="left",
+            validate="m:m",
+            suffixes=("_unlabeled", ""),
+        )
+        .assign(
+            ending_balance=lambda x: x[f"ratio_{tag_col}"].fillna(1) * x.ending_balance,
+        )
```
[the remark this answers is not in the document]
cmgosnell: ah good question. I'll add a comment about this. na values from this ratio tag column do in effect = 100%. because the ratio is of the non-total or non-null tags. But occasionally there are no non-null or non-null tags. Which results in nulls in these ratio columns during the left merge. We want to preserve the ending balance's for them even if there are no non-null/non-total tags to disaggregate.

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3098,109 @@ def out_ferc1__yearly_rate_base(
...
-    return in_rate_base
+    if not np.isclose(
+        new_balance := rate_base_broken_down_df.ending_balance.sum(),
+        old_balance := rate_base_df.ending_balance.sum(),
+    ):
+        logger.warning(
+            f"{tag_col}: New ending balance is not the same as the old ending balance: "
+            f"{old_balance=}, {new_balance=}"
+        )
+    return rate_base_broken_down_df.convert_dtypes()
```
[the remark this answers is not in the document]
cmgosnell: good point on the log level. i think i converted it to a warning while debugging. i will swap out.

this could only happen before the concatenation if we were testing this on only the records that were unlabeled vs the disaggregated. i am partial to keeping it after the concatenation bc it answers the question "is the total ending balance the same before and after we disaggregated some of the records?"

src/pudl/output/ferc1.py
cmgosnell: i'll remove the () portion but i'd like to keep the newly defined bit

src/pudl/output/ferc1.py
cmgosnell: we could but then we'd have two $ columns in the data with very different values. we'd need to drop `dollar_value` below. is this blocking?

src/pudl/output/ferc1.py
cmgosnell: i believe they could! that's not a bad idea i'll try to migrate them to an asset check

src/pudl/output/ferc1.py
aesharpe: no

src/pudl/output/ferc1.py
cmgosnell: moved to asset checks

src/pudl/output/ferc1.py
```diff
@@ -2043,27 +2051,27 @@ def in_explosion_tables(table_name: str, in_explosion_table_names: list[str]) ->
...
 
-    We expect that the facts reported in high-level FERC tables like
+    We expect that the facts reported in less granular FERC tables like
     :ref:`core_ferc1__yearly_income_statements_sched114` and
     :ref:`core_ferc1__yearly_balance_sheet_assets_sched110` should be
     calculable from many individually reported granular values, based on the
     calculations encoded in the XBRL Metadata, and that these relationships should have
-    a hierarchical tree structure. Several individual values from the higher level
+    a hierarchical tree structure. Several individual values from the less granular
     tables will appear as root nodes at the top of each hierarchy, and the leaves in
```
aesharpe, line 2060: maybe say "at the base of each hierarchy" instead of "top" so it's more "rooty"

aesharpe: The word "several" and "appear" here are confusing. Which roots we use is intentional, right?
```
Several individual values from the less granular
tables will appear as root nodes at the top of each hierarchy
```

cmgosnell: yea i went back and forth on this one. i think "top of hierarchy" is more analogous to "root of tree" than "base of hierarchy". bc base of a hierarchy implies that its a wide (i.e. granular) base. and the top of a hierarchy is usually one of more single points (see) from which the rest of the hierarchy flows down (towards leaves)

cmgosnell: if you keep reading below here you'll see that a list of seed nodes can be given to this forest builder, which is the way we intentionally say what root(s) to start building from

src/pudl/output/ferc1.py
```diff
@@ -2043,27 +2051,27 @@ def in_explosion_tables(table_name: str, in_explosion_table_names: list[str]) ->
...
     the underlying tree structure are the individually reported non-calculated values
-    that make them up. Because the top-level tables have several distinct values in
-    them, composed of disjunct sets of reported values, we have a forest (a group of
-    several trees) rather than a single tree.
+    that make them up (i.e. the most granular values). Because the less granular tables
+    have several distinct values in them, composed of disjunct sets of reported values,
+    we have a forest (a group of several trees) rather than a single tree.
```
aesharpe, line 2064: This phrasing is still a little confusing. Are you saying that values from composite "less granular" tables can be both roots and leaves?

The word "underlying" still sounds too "rooty" for a "leaf" could take it out entirely and say something like: "Composite values (root nodes) from less granular tables form the base of the hierarchy from which the most granular calculation components (leaves) arise. Because the less granular tables have several distinct values in them and are composed of a disjunct sets of reported values, our result is a forest (a group of several trees) rather than a single tree."

src/pudl/output/ferc1.py
```diff
@@ -2043,27 +2051,27 @@ def in_explosion_tables(table_name: str, in_explosion_table_names: list[str]) ->
...
 
     The information required to build a calculation forest is most readily found in the
     :meth:`Exploder.exploded_calcs`  A list of seed nodes can also be supplied,
-    indicating which nodes must be present in the resulting forest. This can be used to
-    prune irrelevant portions of the overall forest out of the exploded metadata. If no
-    seeds are provided, then all of the nodes referenced in the exploded_calcs input
-    dataframe will be used as seeds.
+    indicating which nodes the should be the root(s) of the tree(s) we want to built.
+    This can be used to prune irrelevant portions of the overall forest out of the
```
aesharpe, line 2069: Maybe say: "Seed nodes are used to indicate which nodes should be the......."

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3125,127 @@ def out_ferc1__yearly_rate_base(
...
+def disaggregate_null_or_total_tag(
+    rate_base_df: pd.DataFrame,
+    tag_col: str,
+) -> pd.DataFrame:
+    """Disaggregate records with an null or total value in the ``tag_col``.
+
+    We have records in the rate base table with total and/or null values for
+    key tag columns which we want to separate into component parts because the
+    null or total values does not convery a level of detail we want for the
+    rate base table. This is done in two steps:
```
aesharpe, line 3140: yay! this is much more clear. Little grammar thing with  does vs. do

src/pudl/output/ferc1.py
```diff
@@ -3044,25 +3125,127 @@ def out_ferc1__yearly_rate_base(
...
+
+    * :func:`get_tag_col_ratio` : for each ``report_year`` and ``utility_id_ferc1``,
+      get a ratio of all of the ``ending_balance`` for all of the non-null and
+      non-total tags.
+    * use this ratio to disaggregate the ``ending_balance`` from records with total
+      and null tags across the component tags.
+
```
aesharpe, line 3147: I still like the idea of saying "tag value" instead of just "tag" but this is non blocking

src/pudl/output/ferc1.py
```diff
@@ -2274,36 +2282,45 @@ def annotated_forest(self: Self) -> nx.DiGraph:
...
         return annotated_forest
 
     def propagate_node_attributes(self: Self, annotated_forest: nx.DiGraph):
         """Propagate tags.
 
-        Propagate tags leafwards, rootward &  to the _correction nodes.
+        Propagate tags root-ward, leaf-wards &  to the _correction nodes. We
+        propagate the tags root-ward first because we primarily manually
+        compiled tags for the leaf nodes, so we want to send those leafy tags
+        root-ward first before trying to send tags leaf-ward.
```
aesharpe, line 2300: non blocking but would be nice to distinguish between tag columns / groups and tag values

src/pudl/output/ferc1.py
cmgosnell: agreed that underlying is not a great word here. I don't think your suggestion here is totally correct. A ton of the nodes on the tree are composite values - everything except the leaves are composite notes. and as i mentioned in the other thread i don't think "base of hierarchy" is accurate.

also, a deep re-write of this doc is something we just said on the phone was oos for this pr. my edits rn on this are simply to remove "underlying"

aesharpe: could say "form the base of the tree" instead of the hierarchy. I see why it's weird to say "base of the hierarchy"

aesharpe: I don't think this necessarily negates the fact that non-leaf values can be calculated values. But you're right, there's no need to get stuck here. You can keep it as is if you're more comfortable with that!!

src/pudl/output/ferc1.py
cmgosnell: i updated!

src/pudl/output/ferc1.py
aesharpe: I think we did discuss the importance of having a sentence that conveys the relationship between seed/root/leaves/tree/forest relationship (in addition to linking to the networkx docs). Def don't want to re-write all the tree language, just clarify it once.

src/pudl/output/ferc1.py
aesharpe: I think that's why I was confused about the word "appear" in the section I highlighted. The language portrays something that happens coincidentally when in reality it is planned. But this is too nuanced spend time on rly.

src/pudl/output/ferc1.py
cmgosnell: okay im going to remove the hierarchy bit altogether. i don't think it adds anything and any other way besides "root" and "less granular" is complicating.

src/pudl/output/ferc1.py
cmgosnell: they do appear as roots until we "prune" all of the nodes that are not connected to the seed node (this stuff is a little weedsy and is documented in the many class methods

aesharpe (approved):

PR #3538, opened 2024-04-03
Extract AEO Table 54, with bonus 13/15/20.

### LINKED ISSUE #3368: Extract AEO Table 54 into a dataframe
EIA archives a bulk JSON version of the Annual Energy Outlook (AEO), but does not link to these archives from their [bulk data download page](https://www.eia.gov/opendata/v1/bulkfiles.php). The URLs take the form:

`https://www.eia.gov/opendata/bulk/AEOYYY.zip`

And they are available from 2014 to 2023.  There will be no 2024 AEO, while EIA updates their energy modeling system.

So - each record in this pile is a data series column, basically. There's a "data" field which has indices + corresponding data. Plus some metadata.

```json
{
  "series_id": "AEO.2023.LNG_HP.CAP_NA_ENUS_NA_TOT_NA_NA_GW.A",
  "name": "Renewable Energy : End-Use Generators : Net Summer Capacity : Total, High LNG Price, AEO2023",
  "units": "GW",
  "f": "A",
  "start": "2021",
  "end": "2050",
  "lastHistoricalPeriod": "2021",
  "last_updated": "2023-05-04T13:04:45-04:00",
  "data": [
    [
      "2050",
      230.576385
    ],
    ...
    [
      "2022",
      55.064106
    ]
  ]
}
```

And there is some taxonomy data as well - "categories" which don't have a 1:1 mapping to tables. There's over 14,000 of these! 

```json
{
  "category_id": "5048336",
  "parent_category_id": "5047484",
  "name": "Table 67.  Employment and Shipments by Industry, and Income and Employment by Region",
  "notes": "",
  "childseries": [
    "AEO.2023.LNG_LP.ECI_VOS_IDAL_NA_NA_NA_NA_BLNY09DLR.A",
    "AEO.2023.LNG_LP.ECI_VOS_MANF_NA_NA_NA_NA_BLNY09DLR.A",
    "AEO.2023.LNG_LP.ECI_VOS_NA_NA_NA_NA_NA_BLNY09DLR.A",
    "AEO.2023.LNG_LP.ECI_VOS_NIND_NA_NA_NA_NA_BLNY09DLR.A",
    "AEO.2023.LNG_LP.ECI_VOS_NMFG_NA_NA_NA_NA_BLNY09DLR.A"
  ]
}
```

If you dig in to the tree of categories, you find:

* one category per case - "Fast Builds Plus High LNG Price", "High LNG Price", "High Oil Price", etc. There are 20 cases.
* one category per subject, per case - "Electric Power Sector", "Macroeconomic", etc. 17 subjects * 20 cases = 340 subject/case combos.
* one category per table per subject/case combo that is relevant to the table; some tables have extra dimensions that shows up in the table name. There are 68 tables. Some examples:
  * Table 21 doesn't have another dimension:
    * 17x "Table 21.  Residential Sector Equipment Stock and Efficiency, and Distributed Generation"
  * Table 2 - there are two other dimensions: unit + geography
    * 170x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), East North Central"                                                                          
    * 170x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), East South Central"                                                                          
    * 170x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), Middle Atlantic"                                                                             
    * 170x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), Mountain"                                                                                    
    * 170x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), New England"                                                                                 
    * 170x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), Pacific"                                                                                     
    * 170x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), South Atlantic"                                                                              
    * 200x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), United States"                                                                               
    * 170x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), West North Central"                                                                          
    * 170x "Table 2.  Energy Consumption by Sector and Source (quadrillion Btu, unless otherwise noted), West South Central"

I think what we do is:

* create a taxonomy that links the individual category records to:
  * the tables they belong to
  * the case, subject, and other dimensions that should be applied to all of its child series

* for each category that has child series, turn the child series into a series which is indexed by case/subject/datetime/etc.:

  Something like...

  start_dt | end_dt | last_updated | case | subject | factoid | value | unit | geography
  ---------|--------|--------------|------|---------|---------|-------|------|----------
  2022-01-01T00:00:00 | 2023-01-01T00:00:00 | 2024-02-06T13:17:14| High LNG Price | Coal Supply and Prices | very_long_factoid_name | 123.456 | GW | West South Central
  
  And add those rows into the table.
  
* Write those tables out! Hooray. 

First one seems like it is the tricky part. If we get the data structures representing the tables, categories, and additional dimensions right, the rest should follow. I think this would probably take 25h of work or so - a bunch of string mapping, some tree munging. Then turning the child series into data + writing it out should be ~10h. With some contingency for unforeseen weirdness, we could say the whole project takes 40-50h.

If we cut scope to only integrate a few of the tables, that could save some time - 10-20 hours. We will still need to write the code to turn the manual string mappings into a taxonomy, and something that can turn the taxonomy + data into a set of tables. We may be able to skip some of the more complicated additional dimension work, depending on the tables.
```[tasklist]
### Tasks
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3473
- [ ] https://github.com/catalyst-cooperative/pudl-archiver/issues/293
```

jdangerx:
# Overview

Closes #3368.

Grabs 2023 AEO and pulls a single dataframe out of it for table 54, which includes *all* of the reference & side cases. Doesn't do any significant data cleaning.

~~Additionally, refactors how we load assets and asset checks into the modules - this makes adding modules slightly easier IMO - no need to add the module in both the asset modules & the check modules lists. Actually I should probably pull that into a separate PR.~~ See #3539 

Still some clean-up to do, detailed below, but otherwise this is getting quite close.

# Testing

* materialized these assets
* asset checks ran & passed

```[tasklist]
# To-do list
- [x] use a pandera schema to define types
- [x] make better errors than `assert X`
- [x] fix docs build: `/home/runner/work/pudl/pudl/docs/autoapi/pudl/extract/eia_aeo/index.rst:164:Bullet list ends without a blank line; unexpected unindent.`
- [ ] Use dataset settings instead of hardcoding 2023
```


src/pudl/extract/__init__.py
```diff
@@ -17,6 +17,7 @@
     eia861,
     eia923,
     eia930,
+    eia_aeo,
```
zaneselvans, line 20: Existing data source codes don't use an underscore to separate the "agency" and the "form", including the EIA AEO archive / datastore entry, so I think we probably should stick with `eiaaeo` unless we want to update them all to use underscores. `eia_bulk_elec` is the only outlier.

src/pudl/extract/eia_aeo.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+@asset_check(asset="raw_eia_aeo__electric_power_projections_regional")
+def raw_table_54_invariants(df: pd.DataFrame) -> AssetCheckResult:
+    """Check that the AEO Table 54 raw data conforms to *some* assumptions.
+
+    * all values are non-null - i.e. every fact has date, fact name, category
+    name, case, and unit
+    * we have values from every electricity market module region
+    * covers 20 cases and 26 electricity market module regions (25 regions + 1 national)
```
zaneselvans, line 218: I think the lack of a blank line after this bulleted list is what's breaking the docs.

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,252 @@
...
+        return AEOTable(pd.DataFrame.from_records(series_records))
+
+
+@multi_asset(
+    outs={
+        "raw_eiaaeo__electric_power_projections_regional": AssetOut(is_required=False),
+        "raw_eiaaeo__renewable_energy_regional": AssetOut(is_required=False),
+    },
+    can_subset=True,
+    required_resource_keys={"datastore"},
```
zaneselvans, line 200: Do we want to add ETL settings to the required resource keys and define classes for describing them in `pudl.settings` to allow us to control what data is being extracted, rather than hard-coding 2023?

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,252 @@
...
+    those right now.
+
+    The series each have their own timeseries data, as well as some metadata
+    such as a series name and units. Many different dimensions can be inferred
+    from the series names, but the data is somewhat heterogeneous so we do not
+    try to infer those here and leave that to the transformation step.
+    """
+    name_to_number = {
+        "raw_eiaaeo__electric_power_projections_regional": 54,
+        "raw_eiaaeo__renewable_energy_regional": 56,
```
zaneselvans, line 224: Are the renewable energy data not projections?

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,252 @@
...
+        """Load AEO JSON records into a graph datastructure.
+
+        Args:
+            records: the strings that contain an AEO report to parse.
+
+        """
+        categories, series = self.__load_records(records)
+        self.graph = self.__generate_graph(categories, series)
+        root = 4949903
+        self.__cases = set(self.graph.successors(root))
```
zaneselvans, line 71: So there's a structural requirement that the root node never has any child series, and that the top level categories always represent the different cases / scenarios?

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,252 @@
...
+        return graph
+
+    def __sanitize(self, s: str) -> str:
+        return re.sub(self.__sanitize_re, "_", s.lower().strip().replace(" : ", "__"))
+
+    def __series_to_records(
+        self, series_id: str, potential_parents: set[int]
+    ) -> pd.DataFrame:
+        # Use the graph info to figure out what case + category name this
+        # series belongs to
```
zaneselvans, line 124: I realize these are all private methods, but is there a reason not to use docstrings rather than comments to document them? Then the documentation will pop up automatically in editors & notebooks for our own reference.

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,252 @@
...
+        graph = nx.DiGraph(incoming_graph_data=edges)
+
+        nx.set_node_attributes(graph, categories | series)
+        return graph
+
+    def __sanitize(self, s: str) -> str:
+        return re.sub(self.__sanitize_re, "_", s.lower().strip().replace(" : ", "__"))
+
+    def __series_to_records(
+        self, series_id: str, potential_parents: set[int]
```
zaneselvans, line 121: If we have the series ID and the whole graph of connections between various series and categories, why is `potential_parents` necessary?

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,252 @@
...
+        """Get a specific table number as a DataFrame."""
+        matching_category_ids = {
+            n_id
+            for n_id in self.graph
+            if self.__sanitize(self.graph.nodes.get(n_id).get("name", "")).startswith(
+                f"table_{table_number}_"
+            )
+        }
+
+        # many series belong to more than one category, hence turning this into a set
```
zaneselvans, line 179: AH, so it's not a tree? The categories are like hierarchical, overlapping tags?

An explanation of the expected overall structure at the top of the module would be helpful for navigating this.

zaneselvans: So turning the list of matching series into a set avoids ending up with the same series in the dataframe multiple times, by virtue of it being accessible via multiple different categories?

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,252 @@
...
+            if a_id in self.__cases
+        ]
+        if len(case_names) != 1:
+            raise ValueError(
+                f"Found multiple AEO cases for series {series_id}: {case_names}"
+            )
+        case_name = case_names[0]
+
+        # whereas parent_names is a set, because we expect many of the
+        # predecessors to share a name
```
zaneselvans, line 141: I'm not really understanding the reasoning behind list above vs. set here.

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,252 @@
...
+    with ds.get_zipfile_resource("eiaaeo", year=year).open(
+        filename, mode="r"
+    ) as aeo_raw:
+        taxonomy = AEOTaxonomy(io.TextIOWrapper(aeo_raw))
+
+    selected = context.op_execution_context.selected_output_names
+    for asset_name in selected:
+        yield Output(
+            value=taxonomy.get_table(name_to_number[asset_name]), output_name=asset_name
+        )
```
zaneselvans, line 239: Is this generator construct to enable subsetting of the multi-asset?

zaneselvans (changes requested):
This seems clean and compact, but I had a hard time reading through it and understanding what was going on without more context about the expected structure of the data being extracted, which is probably valuable to capture with a little bit of narrative, as well as defensive checks.

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,252 @@
...
+    def __init__(self, records: Iterable[str]):
+        """Load AEO JSON records into a graph datastructure.
+
+        Args:
+            records: the strings that contain an AEO report to parse.
+
+        """
+        categories, series = self.__load_records(records)
+        self.graph = self.__generate_graph(categories, series)
+        root = 4949903
```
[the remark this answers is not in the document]
jdangerx: This appears to be a meaningless number that they assigned to the "root" category which has no name.

src/pudl/extract/eiaaeo.py
jdangerx: Yep!

src/pudl/extract/eiaaeo.py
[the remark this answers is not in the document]
zaneselvans: I would have gone with like, 0 or 1 personally but okay!

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,453 @@
...
+
+    In the first four generations we see a strictly branching tree, but many
+    leaf categories can point at the same data series so the whole taxonomy is
+    a DAG. This is because of two reasons:
+
+    * the subject tag doesn't affect data values, but because of the tree
+    structure, each leaf category is repeated once for each subject, leading to
+    multiple *duplicated* leaf categories pointing at the same data series.
+    * some data series are relevant to multiple different tables - so multiple
+    *different* leaf categories point at the same data series. In this case we
```
zaneselvans, line 110:
```suggestion
    different leaf categories point at the same data series. In this case we
```

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,453 @@
...
+    leaf categories can point at the same data series so the whole taxonomy is
+    a DAG. This is because of two reasons:
+
+    * the subject tag doesn't affect data values, but because of the tree
+    structure, each leaf category is repeated once for each subject, leading to
+    multiple *duplicated* leaf categories pointing at the same data series.
+    * some data series are relevant to multiple different tables - so multiple
+    different leaf categories point at the same data series. In this case we
+    would expect the names of the leaf category to reflect their different
+    identities.
```
zaneselvans, line 112:
```suggestion
    * the subject tag doesn't affect data values, but because of the tree
      structure, each leaf category is repeated once for each subject, leading to
      multiple *duplicated* leaf categories pointing at the same data series.
    * some data series are relevant to multiple different tables - so multiple
      different leaf categories point at the same data series. In this case we
      would expect the names of the leaf category to reflect their different
      identities.
```

src/pudl/extract/eiaaeo.py
```diff
@@ -0,0 +1,453 @@
...
+
+        edges = itertools.chain.from_iterable(
+            get_all_edges(category) for category in categories.values()
+        )
+        graph = nx.DiGraph(incoming_graph_data=edges)
+
+        nx.set_node_attributes(graph, categories | series)
+        return graph
+
+    def __generation_invariants(self) -> list:  # noqa: C901
```
zaneselvans, line 195: I feel like there are a bunch of error checking patterns encapsulated in this method that we could learn from and deploy elsewhere.

zaneselvans (approved):
So much clearer and more explicit! Thank you for making it legible to our future selves.  I would love to get better at the kind of error checking and compilation you added in here.

zaneselvans:
@jdangerx Also, the docs formatting issue was that you have to indent multi-line bullets. E.g.

This is valid:
```
* this is the first line of a
  multi-line bulleted list item.
* This is the second thing in my list
  and it also has more than one line
```

But this is not:

```
* this is the first line of a
multi-line bulleted list item.
* This is the second thing in my list
and it also has more than one line
```

jdangerx:
OMG thank you for finding the doc build problem 🙇 !

Maybe for cliffs notes club this week we can just go over my not-so-beautiful-but-maybe-worth-discussing graph invariant checker :)

PR #3539, opened 2024-04-03
Refactor etl/__init__.py to make adding new modules easier.

jdangerx:
# Overview

This changes how we add new modules to the ETL so we only have to add them in one place. Also it organizes them slightly.

Finally, there's a tiny opportunistic change to the docs to remove a reference to `dev`.

# Testing

* Checked to see that all modules on the left do indeed show up on the right.
* Loaded up Dagster UI and saw that the ETL appears to still be there

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -46,86 +46,83 @@ logger = pudl.logging_helpers.get_logger(__name__)
 # everywhere.
 warnings.filterwarnings("ignore", category=ExperimentalWarning)
 
-default_assets = (
-    *load_assets_from_modules([eia_bulk_elec_assets], group_name="core_eia_bulk_elec"),
-    *load_assets_from_modules([epacems_assets], group_name="core_epacems"),
-    *load_assets_from_modules([pudl.extract.eia176], group_name="raw_eia176"),
-    *load_assets_from_modules([pudl.extract.eia191], group_name="raw_eia191"),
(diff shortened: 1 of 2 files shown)


src/pudl/etl/__init__.py
```diff
@@ -46,86 +46,83 @@
...
+    "out_derived_gen_attributes": [pudl.analysis.mcoe],
+    "out_eia": [
+        pudl.output.eia,
+        pudl.output.eia860,
+        pudl.output.eia923,
+        pudl.output.eia_bulk_elec,
+    ],
+    "out_ferc1": [
+        pudl.output.ferc1,
+        pudl.analysis.record_linkage.classify_plants_ferc1,
```
jdangerx, line 100: Note that these two used to be loaded separately.

zaneselvans (approved):

zaneselvans:
The main duplication that's being removed here is that we're now using the same dictionary of modules to load assets and also to load asset checks, right?  This definitely tripped me up for a bit when I was trying to add an asset check to the GridPath RA Toolkit data.

jdangerx:
> The main duplication that's being removed here is that we're now using the same dictionary of modules to load assets and also to load asset checks, right? This definitely tripped me up for a bit when I was trying to add an asset check to the GridPath RA Toolkit data.

Yep, exactly!

Issue #3161, opened 2023-12-15
Remove `pudl_etl` and `ferc_to_sqlite` commands in favor of `dagster job execute`


bendnorman:

jdangerx:
This would be nice! What's the difference between the nightly build versions of the jobs & the current `etl_full` and `ferc_to_sqlite_full` jobs?

PR #3541, opened 2024-04-04
Attempt to limit `_out_ferc714__hourly_demand_matrix` concurrency

### LINKED ISSUE #3533: Nightly Build Failure 2024-04-03
# Overview

The `_out_ferc714__hourly_demand_matrix` job ran into an OOM kill.

From the logs we can see that it started at 07:10:43 UTC and got the kill at 07:11:46 UTC.

[here](https://console.cloud.google.com/monitoring/metrics-explorer;endTime=2024-04-03T07:20:00.000Z;startTime=2024-04-03T07:00:00.000Z?pageState=%7B%22domainObjectDeprecationId%22:%229341DDB9-6FE8-44A6-BB0E-430C82185495%22,%22title%22:%22Untitled%22,%22xyChart%22:%7B%22constantLines%22:%5B%5D,%22dataSets%22:%5B%7B%22plotType%22:%22LINE%22,%22targetAxis%22:%22Y1%22,%22timeSeriesFilter%22:%7B%22aggregations%22:%5B%7B%22crossSeriesReducer%22:%22REDUCE_SUM%22,%22groupByFields%22:%5B%5D,%22perSeriesAligner%22:%22ALIGN_MEAN%22%7D%5D,%22apiSource%22:%22DEFAULT_CLOUD%22,%22crossSeriesReducer%22:%22REDUCE_SUM%22,%22filter%22:%22metric.type%3D%5C%22compute.googleapis.com%2Finstance%2Fmemory%2Fballoon%2Fram_used%5C%22%20resource.type%3D%5C%22gce_instance%5C%22%22,%22groupByFields%22:%5B%5D,%22minAlignmentPeriod%22:%2260s%22,%22perSeriesAligner%22:%22ALIGN_MEAN%22%7D%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22y1Axis%22:%7B%22label%22:%22%22,%22scale%22:%22LINEAR%22%7D%7D%7D&project=catalyst-cooperative-pudl) are the VM instance RAM metrics over the 07:00 - 07:20 timeframe, you can see that when the job started we were using ~35GB of RAM already and then we quickly jumped up to 60+ before getting killed at 07:12.

image

We also saw this same issue 2024-04-02, but didn't make an issue for it. 

# Next steps

What next steps do we need to do to understand or remediate the issue?

We could maybe force this asset to run non-concurrently. 

We could maybe reduce the RAM usage of this asset.

```[tasklist]
```

# Relevant logs
[link to build logs from internal distribution bucket](https://console.cloud.google.com/storage/browser/builds.catalyst.coop/2024-04-03-0602-c28abb2be-main)

```
2024-04-03 07:10:03 +0000 - dagster - DEBUG - etl_job - be4a0899-56fa-429f-831a-93cd28c6648b - 1779 - _out_ferc714__hourly_demand_matrix - STEP_WORKER_STARTING - Launching subprocess for "_out_ferc714__hourly_demand_matrix".
2024-04-03 07:10:40 +0000 - dagster - DEBUG - etl_job - be4a0899-56fa-429f-831a-93cd28c6648b - 23579 - _out_ferc714__hourly_demand_matrix - STEP_WORKER_STARTED - Executing step "_out_ferc714__hourly_demand_matrix" in subprocess.
2024-04-03 07:10:42 +0000 - dagster - DEBUG - etl_job - be4a0899-56fa-429f-831a-93cd28c6648b - 23579 - _out_ferc714__hourly_demand_matrix - RESOURCE_INIT_STARTED - Starting initialization of resources [io_manager].
2024-04-03 07:10:42 +0000 - dagster - DEBUG - etl_job - be4a0899-56fa-429f-831a-93cd28c6648b - 23579 - _out_ferc714__hourly_demand_matrix - RESOURCE_INIT_SUCCESS - Finished initialization of resources [io_manager].
2024-04-03 07:10:43 +0000 - dagster - DEBUG - etl_job - be4a0899-56fa-429f-831a-93cd28c6648b - 23579 - _out_ferc714__hourly_demand_matrix - STEP_START - Started execution of step "_out_ferc714__hourly_demand_matrix".
2024-04-03 07:11:46 +0000 - dagster - DEBUG - etl_job - be4a0899-56fa-429f-831a-93cd28c6648b - 1779 - _out_ferc714__hourly_demand_matrix - ENGINE_EVENT - Multiprocess executor: child process for step _out_ferc714__hourly_demand_matrix was terminated by signal 9 (SIGKILL). This usually indicates that the process was killed by the operating system due to running out of memory. Possible solutions include increasing the amount of memory available to the run, reducing the amount of memory used by the ops in the run, or configuring the executor to run fewer ops concurrently.
```

There are several other concurrent tasks running during these two minutes:

```
> ag --no-numbers '07:1[01].*STEP' 2024-04-03-0602-c28abb2be-main-pudl-etl.log | cut -d ' ' -f15 | sort | uniq 
_core_eia860__boilers
_core_eia860__ownership
_core_eia860__plants
_core_eia860__utilities
_core_eia923__boiler_fuel
_core_eia923__coalmine
_core_eia923__cooling_system_information
_core_eia923__fgd_operation_maintenance
_out_ferc1__yearly_steam_plants_sched402_with_plant_ids.ferc_to_ferc.merge_steam_fuel_dfs
_out_ferc714__hourly_demand_matrix
core_demand_side_management_eia861
core_eia861__yearly_balancing_authority
core_eia861__yearly_dynamic_pricing
extract_phmsagas
raw_eia930__balance
raw_eia930__interchange
raw_eia930__subregion
```

<details>
<summary>all the dagster steps</summary>
```
> ag --no-numbers '07:1[01].*STEP' 2024-04-03-0602-c28abb2be-main-pudl-etl.log | cut -d ' ' -f1 -f2 -f15 -f17
2024-04-03 07:07:11 core_demand_side_management_eia861 STEP_WORKER_STARTED
2024-04-03 07:07:11 core_eia861__yearly_dynamic_pricing STEP_WORKER_STARTING
2024-04-03 07:10:00 extract_phmsagas STEP_WORKER_STARTING
2024-04-03 07:10:00 _core_eia860__ownership STEP_WORKER_STARTED
2024-04-03 07:10:02 _core_eia860__plants STEP_WORKER_STARTED
2024-04-03 07:10:03 _out_ferc714__hourly_demand_matrix STEP_WORKER_STARTING
2024-04-03 07:10:03 _core_eia860__ownership STEP_START
2024-04-03 07:10:03 _core_eia860__ownership STEP_INPUT
2024-04-03 07:10:04 _core_eia860__utilities STEP_WORKER_STARTED
2024-04-03 07:10:04 _core_eia860__plants STEP_START
2024-04-03 07:10:05 _core_eia860__plants STEP_INPUT
2024-04-03 07:10:06 _core_eia860__utilities STEP_START
2024-04-03 07:10:07 _core_eia860__utilities STEP_INPUT
2024-04-03 07:10:11 _core_eia860__utilities STEP_OUTPUT
2024-04-03 07:10:17 _out_ferc1__yearly_steam_plants_sched402_with_plant_ids.ferc_to_ferc.merge_steam_fuel_dfs STEP_WORKER_STARTING
2024-04-03 07:10:28 _core_eia860__plants STEP_OUTPUT
2024-04-03 07:10:29 _core_eia860__ownership STEP_OUTPUT
2024-04-03 07:10:29 _core_eia860__ownership STEP_SUCCESS
2024-04-03 07:10:30 extract_phmsagas STEP_WORKER_STARTED
2024-04-03 07:10:31 _core_eia860__plants STEP_SUCCESS
2024-04-03 07:10:36 raw_eia930__balance STEP_WORKER_STARTING
2024-04-03 07:10:37 raw_eia930__interchange STEP_WORKER_STARTING
2024-04-03 07:10:37 extract_phmsagas STEP_START
2024-04-03 07:10:40 _out_ferc714__hourly_demand_matrix STEP_WORKER_STARTED
2024-04-03 07:10:41 extract_phmsagas STEP_INPUT
2024-04-03 07:10:41 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:42 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:42 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:43 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:43 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:43 _core_eia860__boilers STEP_OUTPUT
2024-04-03 07:10:43 _out_ferc714__hourly_demand_matrix STEP_START
2024-04-03 07:10:43 _out_ferc714__hourly_demand_matrix STEP_INPUT
2024-04-03 07:10:43 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:43 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:44 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:44 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:44 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:45 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:45 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:45 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:45 extract_phmsagas STEP_OUTPUT
2024-04-03 07:10:47 raw_eia930__subregion STEP_WORKER_STARTING
2024-04-03 07:10:50 _out_ferc1__yearly_steam_plants_sched402_with_plant_ids.ferc_to_ferc.merge_steam_fuel_dfs STEP_WORKER_STARTED
2024-04-03 07:10:51 core_eia861__yearly_balancing_authority STEP_WORKER_STARTING
2024-04-03 07:10:57 raw_eia930__balance STEP_WORKER_STARTED
2024-04-03 07:10:57 raw_eia930__interchange STEP_WORKER_STARTED
2024-04-03 07:10:59 raw_eia930__balance STEP_START
2024-04-03 07:10:59 raw_eia930__interchange STEP_START
2024-04-03 07:11:00 _core_eia923__boiler_fuel STEP_WORKER_STARTING
2024-04-03 07:11:05 raw_eia930__subregion STEP_WORKER_STARTED
2024-04-03 07:11:08 raw_eia930__subregion STEP_START
2024-04-03 07:11:11 core_eia861__yearly_balancing_authority STEP_WORKER_STARTED
2024-04-03 07:11:18 core_eia861__yearly_balancing_authority STEP_START
2024-04-03 07:11:19 core_eia861__yearly_balancing_authority STEP_INPUT
2024-04-03 07:11:22 _core_eia923__boiler_fuel STEP_WORKER_STARTED
2024-04-03 07:11:23 _core_eia923__coalmine STEP_WORKER_STARTING
2024-04-03 07:11:24 _core_eia923__boiler_fuel STEP_START
2024-04-03 07:11:26 _core_eia923__boiler_fuel STEP_INPUT
2024-04-03 07:11:46 _out_ferc714__hourly_demand_matrix STEP_FAILURE
2024-04-03 07:11:46 _core_eia923__cooling_system_information STEP_WORKER_STARTING
2024-04-03 07:11:52 raw_eia930__interchange STEP_INPUT
2024-04-03 07:11:52 raw_eia930__interchange STEP_OUTPUT
2024-04-03 07:11:52 _core_eia923__boiler_fuel STEP_OUTPUT
2024-04-03 07:11:53 _core_eia923__coalmine STEP_WORKER_STARTED
2024-04-03 07:11:53 raw_eia930__balance STEP_INPUT
2024-04-03 07:11:53 raw_eia930__balance STEP_OUTPUT
2024-04-03 07:11:55 _core_eia923__coalmine STEP_START
2024-04-03 07:11:57 raw_eia930__subregion STEP_INPUT
2024-04-03 07:11:57 raw_eia930__subregion STEP_OUTPUT
2024-04-03 07:11:57 raw_eia930__subregion STEP_SUCCESS
2024-04-03 07:11:57 _core_eia923__fgd_operation_maintenance STEP_WORKER_STARTING
2024-04-03 07:11:58 _core_eia923__coalmine STEP_INPUT
```
</details>

bendnorman:
# Overview

Closes #3533.

Attempts to limit max concurrency to 1 when the `_out_ferc714__hourly_demand_matrix` asset is running in an attempt to reduce memory usage on the nightly build VM. 

I'm still a little confused about how dagster's concurrency limiting with tags works, see me [comment here](https://github.com/catalyst-cooperative/pudl/issues/3533#issuecomment-2035791127).

Also, if we keep needing to limit concurrency when certain assets are run because of memory issues, we should come up with a more generalization method so we can restrict the concurrency for any asset. This PR just adds some hard coded values to the an asset decorator and the ETL scripts.

# Testing

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/analysis/state_demand.py
+++ b/src/pudl/analysis/state_demand.py
@@ -473,10 +473,7 @@ def _out_ferc714__hourly_demand_matrix(
     return df
 
 
-@asset(
-    compute_kind="Python",
-    op_tags={"memory-use": "high"},
-)
+@asset(compute_kind="Python")
 def _out_ferc714__hourly_imputed_demand(
     _out_ferc714__hourly_demand_matrix: pd.DataFrame,
     _out_ferc714__utc_offset: pd.DataFrame,
--- a/src/pudl/analysis/state_demand.py
+++ b/src/pudl/analysis/state_demand.py
@@ -449,7 +449,7 @@ def melt_ferc714_hourly_demand_matrix(
             ),
         ),
     },
-    op_tags={"datasource": "ferc714"},
+    op_tags={"memory-use": "high"},
 )
 def _out_ferc714__hourly_demand_matrix(
     context, _out_ferc714__hourly_pivoted_demand_matrix: pd.DataFrame
@@ -473,7 +473,10 @@ def _out_ferc714__hourly_demand_matrix(
     return df
 
 
-@asset(compute_kind="Python")
+@asset(
+    compute_kind="Python",
+    op_tags={"memory-use": "high"},
+)
 def _out_ferc714__hourly_imputed_demand(
     _out_ferc714__hourly_demand_matrix: pd.DataFrame,
     _out_ferc714__utc_offset: pd.DataFrame,
--- a/src/pudl/analysis/state_demand.py
+++ b/src/pudl/analysis/state_demand.py
@@ -449,6 +449,7 @@ def melt_ferc714_hourly_demand_matrix(
             ),
         ),
     },
+    op_tags={"datasource": "ferc714"},
 )
 def _out_ferc714__hourly_demand_matrix(
     context, _out_ferc714__hourly_pivoted_demand_matrix: pd.DataFrame
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -206,17 +206,12 @@ default_resources = {
     "epacems_io_manager": epacems_io_manager,
 }
 
-# By default, limit CEMS and FERC 714 processing concurrency to prevent memory overload.
+# Limit the number of concurrent workers when launch assets that use a lot of memory.
 default_tag_concurrency_limits = [
     {
-        "key": "datasource",
-        "value": "epacems",
+        "key": "memory-use",
+        "value": "high",
         "limit": 2,
-    },
-    {
-        "key": "datasource",
-        "value": "ferc714",
-        "limit": 1,
     },
 ]
 default_config = pudl.helpers.get_dagster_execution_config(
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -206,13 +206,18 @@ default_resources = {
     "epacems_io_manager": epacems_io_manager,
 }
 
-# By default, limit CEMS year processing concurrency to prevent memory overload.
+# By default, limit CEMS and FERC 714 processing concurrency to prevent memory overload.
 default_tag_concurrency_limits = [
     {
         "key": "datasource",
         "value": "epacems",
         "limit": 2,
-    }
+    },
+    {
+        "key": "datasource",
+        "value": "ferc714",
+        "limit": 1,
+    },
 ]
 default_config = pudl.helpers.get_dagster_execution_config(
     tag_concurrency_limits=default_tag_concurrency_limits
--- a/src/pudl/etl/cli.py
+++ b/src/pudl/etl/cli.py
@@ -78,15 +78,6 @@ def pudl_etl_job_factory(
     help="Max number of processes Dagster can launch. Defaults to the number of CPUs.",
 )
 @click.option(
-    "--epacems-workers",
-    default=2,
-    type=int,
-    help=(
-        "Max number of processes Dagster can launch for EPA CEMS assets. Defaults "
-        "to max number of processes our typical local machines can handle."
-    ),
-)
-@click.option(
     "--gcs-cache-path",
     type=str,
     default="",
@@ -118,7 +109,6 @@ def pudl_etl_job_factory(
 def pudl_etl(
     etl_settings_yml: pathlib.Path,
     dagster_workers: int,
-    epacems_workers: int,
     gcs_cache_path: str,
     logfile: pathlib.Path,
     loglevel: str,
@@ -158,12 +148,18 @@ def pudl_etl(
         },
     }
 
+    # By default, limit CEMS and FERC 714 processing concurrency to prevent memory overload.
     tag_concurrency_limits = [
         {
             "key": "datasource",
             "value": "epacems",
-            "limit": epacems_workers,
-        }
+            "limit": 2,
+        },
+        {
+            "key": "datasource",
+            "value": "ferc714",
+            "limit": 1,
+        },
     ]
 
     run_config.update(
--- a/src/pudl/etl/cli.py
+++ b/src/pudl/etl/cli.py
@@ -148,17 +148,12 @@ def pudl_etl(
         },
     }
 
-    # By default, limit CEMS and FERC 714 processing concurrency to prevent memory overload.
+    # Limit the number of concurrent workers when launch assets that use a lot of memory.
     tag_concurrency_limits = [
         {
-            "key": "datasource",
-            "value": "epacems",
+            "key": "memory-use",
+            "value": "high",
             "limit": 2,
-        },
-        {
-            "key": "datasource",
-            "value": "ferc714",
-            "limit": 1,
         },
     ]
 
--- a/src/pudl/analysis/plant_parts_eia.py
+++ b/src/pudl/analysis/plant_parts_eia.py
@@ -368,6 +368,7 @@ PPE_COLS = [
 @asset(
     io_manager_key="pudl_io_manager",
     compute_kind="Python",
+    op_tags={"memory-use": "high"},
 )
 def out_eia__yearly_generators_by_ownership(
     out_eia__yearly_generators: pd.DataFrame, out_eia860__yearly_ownership: pd.DataFrame
@@ -382,6 +383,7 @@ def out_eia__yearly_generators_by_ownership(
 @asset(
     io_manager_key="pudl_io_manager",
     compute_kind="Python",
+    op_tags={"memory-use": "high"},
 )
 def out_eia__yearly_plant_parts(
     out_eia__yearly_generators_by_ownership: pd.DataFrame,
--- a/src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
+++ b/src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
@@ -289,7 +289,7 @@ def get_best_matches(
         "out_pudl__yearly_assn_eia_ferc1_plant_parts": Out(
             io_manager_key="pudl_io_manager"
         )
-    }
+    },
 )
 def get_full_records_with_overrides(best_match_df, inputs, experiment_tracker):
     """Join full dataframe onto matches to make usable and get stats.
--- a/src/pudl/etl/epacems_assets.py
+++ b/src/pudl/etl/epacems_assets.py
@@ -58,7 +58,7 @@ def get_years_from_settings(context):
 
 @op(
     required_resource_keys={"datastore", "dataset_settings"},
-    tags={"datasource": "epacems"},
+    tags={"memory-use": "high"},
 )
 def process_single_year(
     context,
--- a/src/pudl/extract/gridpathratoolkit.py
+++ b/src/pudl/extract/gridpathratoolkit.py
@@ -60,11 +60,15 @@ def raw_gridpathratoolkit_asset_factory(part: str) -> AssetsDefinition:
     into the ``aggregation_group`` field, indicating which generators were aggregated to
     produce the time series based on the wind and solar capacity aggregation tables.
     """
+    asset_kwargs = {
+        "name": f"raw_gridpathratoolkit__{part}",
+        "required_resource_keys": {"datastore", "dataset_settings"},
+        "compute_kind": "Python",
+    }
+    if part == "aggregated_extended_solar_capacity":
+        asset_kwargs["op_tags"] = {"memory-use": "high"}
 
-    @asset(
-        name=f"raw_gridpathratoolkit__{part}",
-        required_resource_keys={"datastore", "dataset_settings"},
-    )
+    @asset(**asset_kwargs)
     def _extract(context):
         """Extract raw GridPath RA Toolkit renewable energy generation profiles.
 
--- a/src/pudl/output/ferc714.py
+++ b/src/pudl/output/ferc714.py
@@ -566,7 +566,10 @@ def out_ferc714__respondents_with_fips(
     return fipsified
 
 
-@asset(compute_kind="Python")
+@asset(
+    compute_kind="Python",
+    op_tags={"memory-use": "high"},
+)
 def _out_ferc714__georeferenced_counties(
     out_ferc714__respondents_with_fips: pd.DataFrame,
     _core_censusdp1tract__counties: gpd.GeoDataFrame,
@@ -619,7 +622,11 @@ def _out_ferc714__georeferenced_respondents(
     return respondents_gdf
 
 
-@asset(compute_kind="Python", io_manager_key="pudl_io_manager")
+@asset(
+    compute_kind="Python",
+    io_manager_key="pudl_io_manager",
+    op_tags={"memory-use": "high"},
+)
 def out_ferc714__summarized_demand(
     _out_ferc714__annualized_respondents: pd.DataFrame,
     out_ferc714__hourly_planning_area_demand: pd.DataFrame,
--- a/src/pudl/transform/eia860m.py
+++ b/src/pudl/transform/eia860m.py
@@ -8,7 +8,11 @@ import pudl
 logger = pudl.logging_helpers.get_logger(__name__)
 
 
-@asset(io_manager_key="pudl_io_manager")
+@asset(
+    io_manager_key="pudl_io_manager",
+    compute_kind="Python",
+    op_tags={"memory-use": "high"},
+)
 def core_eia860m__changelog_generators(
     raw_eia860m__generator_proposed,
     raw_eia860m__generator_existing,
--- a/src/pudl/transform/gridpathratoolkit.py
+++ b/src/pudl/transform/gridpathratoolkit.py
@@ -40,7 +40,11 @@ def _transform_capacity_factors(
     return capacity_factors
 
 
-@asset(io_manager_key="pudl_io_manager")
+@asset(
+    io_manager_key="pudl_io_manager",
+    compute_kind="Python",
+    op_tags={"memory-use": "high"},
+)
 def out_gridpathratoolkit__hourly_available_capacity_factor(
     raw_gridpathratoolkit__aggregated_extended_solar_capacity: pd.DataFrame,
     raw_gridpathratoolkit__aggregated_extended_wind_capacity: pd.DataFrame,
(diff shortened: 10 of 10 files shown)


src/pudl/etl/cli.py
```diff
@@ -77,15 +77,6 @@ def get_pudl_etl_job():
...
 )
-@click.option(
-    "--epacems-workers",
-    default=2,
-    type=int,
-    help=(
-        "Max number of processes Dagster can launch for EPA CEMS assets. Defaults "
-        "to max number of processes our typical local machines can handle."
-    ),
-)
```
bendnorman, line 88: I removed this because I dont' think we are actually changing the number of workers using the command option. We could add an argument that can accept an arbitrary number of asset_name and concurrency limits:

```
pudl_etl src/pudl/package_data/settings/etl_full.yml --tag-concurrency-limits a 1 --tag-concurrency-limits b 2
```

bendnorman: In the long term I'd like to just remove the `pudl_etl` command in favor of preconfigured dagster jobs executed using [`dagster job execute`](https://github.com/catalyst-cooperative/pudl/issues/3161) (sorry I still need to flesh out this issue) .

src/pudl/etl/__init__.py
```diff
@@ -209,13 +209,18 @@ def _get_keys_from_assets(
...
         "key": "datasource",
         "value": "epacems",
         "limit": 2,
-    }
+    },
+    {
+        "key": "datasource",
+        "value": "ferc714",
+        "limit": 1,
+    },
```
zaneselvans, line 223: I don't *think* this will have any effect. The FERC-714 job is parallelized internally, not by virtue of the Dagster infrastructure. Right now it will use all available CPUs no matter what. I think it's the function deep down inside that needs to be told to do something different. By contrast the CEMS work units *are* being managed by Dagster.

src/pudl/etl/cli.py
zaneselvans: Agree that we aren't really using this option, and switching to the dagster native CLI would be good.

src/pudl/etl/cli.py
```diff
@@ -158,12 +148,18 @@ def pudl_etl(
...
             "value": "epacems",
-            "limit": epacems_workers,
-        }
+            "limit": 2,
+        },
+        {
+            "key": "datasource",
+            "value": "ferc714",
+            "limit": 1,
+        },
```
zaneselvans, line 162: As above, I don't think changed Dagster's configuration will have any impact on what happens inside the imputation function, which is doing a bunch of tensor math that's internally parallelized.

zaneselvans (changes requested):
I don't think this asset's CPU usage can be affected by the Dagster configuration.

src/pudl/etl/__init__.py
zaneselvans: * If I run `_out_ferc714__hourly_demand_matrix` by itself, it uses 1 CPU and ~4GB of memory, and takes ~3 minutes to complete.
* And then `_out_ferc714__hourly_imputed_demand` is the asset that tries to use lots of cores, but all within a single asset (the parallelism has nothing to do with Dagster). It's often running at 4-600% (on my laptop) with one process that seems to need ~1-2 GB of memory.

zaneselvans: Ah okay, it's the `numpy.linalg.svd` and `numpy.einsum` functions which rely on multithreaded BLAS and LAPACK libraries, whose behavior is controlled by environment variables or other libraries...

https://numpy.org/doc/stable/reference/routines.linalg.html

zaneselvans: But again the uses-lots-of-cores asset doesn't seem to be a uses-lots-of-memory asset so I'm confused.

bendnorman: `_out_ferc714__hourly_imputed_demand` wasn't running when we got the OOM error though. Assuming `_out_ferc714__hourly_demand_matrix` pushed us over the memory edge I thought we could limit the number of processes running when the asset is run. Maybe this is just playing wack a mole though.

I should probably do a full memory profile locally to understand what the culprit is.

zaneselvans: Oh is this a *global* concurrency limit? Like if this process is running, then no other Dagster ops can be running?

zaneselvans: Or are we also applying a global concurrency limit on the CEMS assets? It always seemed like other stuff was running, but only 2 CEMS assets would ever be materializing at a time.  If we have a global concurrency limit on CEMS we could probably run more than 2 at a time. I think each of them takes ~2GB of memory peak.

jdangerx: I also think that these are *tag* level concurrency and that @e-belfer is right. I think we have a couple options:

1. make a `high-mem` op_tag, set its tag concurrency to $LOW_NUM
2. set `max_concurrent` (which is the *global* concurrency limit) to $MEDIUM_NUM

We could combine these, also - say our highest memory usage asset uses 30GB memory, and our `high-mem` threshold is ~8GB. Then, if the `high-mem` concurrency limit is 1, we could set the *global* concurrency limit to 5 and still expect our peak memory usage to be less than (30 * 1 + 8 * 4) = 62 GB. 

We don't have to have an exhaustively-correct `high-mem` tag either - I think just picking off a few culprits will improve our reliability here.

Finally, it's maybe worth looking at some of [these recommendations](https://github.com/dagster-io/dagster/discussions/17616) for reducing memory usage - but that seems like more effort than the `high-mem` bandaid.

zaneselvans: I like the generic nature of the `high-mem` tag, at least as a short-term fix.

Probably a lot of our timeseries processing could be parallelized thoughtfully with Dask, and it seems like there could be a lot more timeseries in our future. But it'll take much more specific effort.

If/When we switch to using a Dask cluster as the executor (or anything other than One Big Machine) is there a way to use something like the `high-mem` tag to make sure that the node that gets that job has enough memory, without requiring *every* node to have a lot of memory?

bendnorman: I also like the `high-mem` tag idea. Would it protect us against a scenario where we have one `high-mem` asset running and a bunch of other medium - low memory assets running?

We could try just using the `high-mem` tag and if it doesn't work then we can limit global concurrency.

jdangerx: > If/When we switch to using a Dask cluster as the executor (or anything other than One Big Machine) is there a way to use something like the `high-mem` tag to make sure that the node that gets that job has enough memory, without requiring _every_ node to have a lot of memory?

Yeah, it looks like Dask supports differently-resourced worker types/pools: https://blog.dask.org/2022/02/17/helm-multiple-worker-groups

> I also like the high-mem tag idea. Would it protect us against a scenario where we have one high-mem asset running and a bunch of other medium - low memory assets running?
 
Not really, if we have 64 GB and our 30GB task is running, there's nothing to stop dagster from scheduling 1000 1GB tasks also. So the global limit is necessary to fully protect us.

src/pudl/analysis/state_demand.py
```diff
@@ -472,7 +473,10 @@ def _out_ferc714__hourly_demand_matrix(
     return df
 
 
-@asset(compute_kind="Python")
+@asset(
+    compute_kind="Python",
+    op_tags={"memory-use": "high"},
+)
 def _out_ferc714__hourly_imputed_demand(
```
bendnorman, line 480: I added a tag to this asset because it is parallelized internally and gobbles up a lot of memory and cpus.

src/pudl/etl/cli.py
```diff
@@ -158,12 +148,13 @@ def pudl_etl(
...
     tag_concurrency_limits = [
         {
-            "key": "datasource",
-            "value": "epacems",
-            "limit": epacems_workers,
-        }
+            "key": "memory-use",
+            "value": "high",
+            "limit": 2,
+        },
```
bendnorman, line 157: So this means no more than two high-memory processes launched by dagster will be running at the same time.  Hopefully this is enough to keep our nightly builds chugging along. If it doesn't work we can limit the concurrency globally and/or add the tag to more high memory assets.

src/pudl/analysis/state_demand.py
zaneselvans: I'll try running the full ETL locally and see what this does to the overall run time.  I'm worried that because this process is running for almost the entire duration of the ETL, having it tagged as high memory usage will mean that the EPA CEMS ends up taking 2x as long as it does now.

When I ran this process in isolation locally it used 4-6 CPUs, but only 1-2 GB of memory, so I think we might not need for it to be held out as a high memory use asset.

bendnorman: Ah ok. It also wasn't running when the VM failed so I'm down to not restrict it's concurrency.

src/pudl/etl/cli.py
bendnorman: Maybe we should add some tags to other high memory assets that were running when the VM failed?

```
_core_eia860__boilers
_core_eia860__ownership
_core_eia860__plants
_core_eia860__utilities
_core_eia923__boiler_fuel
_core_eia923__coalmine
_core_eia923__cooling_system_information
_core_eia923__fgd_operation_maintenance
_out_ferc1__yearly_steam_plants_sched402_with_plant_ids.ferc_to_ferc.merge_steam_fuel_dfs
_out_ferc714__hourly_demand_matrix
core_demand_side_management_eia861
core_eia861__yearly_balancing_authority
core_eia861__yearly_dynamic_pricing
extract_phmsagas
raw_eia930__balance
raw_eia930__interchange
raw_eia930__subregion
```

Given CEMS and `_out_ferc714__hourly_demand_matrix` are the only tagged assets it's unlikely they will be executed at the same time and limit concurrency.

src/pudl/analysis/state_demand.py
zaneselvans: I'm hunting down a bunch of other high memory assets now. There are some doozies! The EIA930 tops out at 11GB, and the EIA860M changelog table at 8GB.

zaneselvans:
I ran a bunch of assets in isolation locally, which I thought were likely to be high memory use, and added the memory-use: high tag when appropriate. I used 4GB as my threshold, since the nightly build machine has 4GB per CPU. Unfortunately, there are some graph-backed assets that I could not use the `op_tags` argument on, and one of them is the biggest memory hog of all:

- [ ] 11.0 GB, 100%, raw_eia930__all_dfs.concat_pages
- [x] 7.9 GB, 100%, core_eia860m__generators
- [x] 7.2 GB, 100%, _out_ferc714__georeferenced_counties
- [x] 7.2 GB, 100%, out_ferc714__summarized_demand
- [x] 6.0 GB, 100%, out_ferc714__hourly_planning_area_demand
- [x] 5.7 GB, 100%, out_eia__yearly_plant_parts
- [x] 5.0 GB, 100%, out_eia__yearly_generators_by_ownership
- [x] 4.8 GB, 100%, raw_gridpathratoolkit__aggregated_extended_solar_capacity
- [x] 4.7 GB, 100%, out_gridpathratoolkit__hourly_available_capacity_factor
- [ ] 4.6 GB, 100%, out_pudl__yearly_assn_eia_ferc1_plant_parts
- [ ] 4.6 GB, 100%, _out_ferc1__yearly_steam_plants_with_plant_ids
- [x] 4.0 GB, 100%, _out_ferc714__hourly_demand_matrix
- [x] 2.0 GB, 100%, core_epacems__hourly_emissions.process_single_year

zaneselvans (approved):

PR #3543, opened 2024-04-05
Manage concurrency of high-memory processes

zaneselvans:
# Overview

This PR follows on #3541 

* Tag additional high memory (>4GB peak) processes
* Dial down the parallelism of Numpy's linear algebra
* Increase the allowed concurrency of high memory processes, since we can be more confident about limiting their overall number with more complete tagging.

## Limiting concurrency of high memory ops

The nightly build machine has no swap, and 4GB of memory per CPU (16 CPUs, 64 GB of memory, as on all "highmem" instances). Some of the assets in our DAG use significantly more memory that that, and sometimes a few of them stack on top of each other and they run out of memory, crashing the process and causing the ETL to fail.

After a lot of time running individual assets and watching the per-process memory consumption while the ETL was running locally, it seems like there aren't *that* many assets / ops using significantly more than 4GB of memory peak. Here's a checklist of the ones I identified, with the highest memory usage I observed. The ones checked off have been tagged as high memory usage:

- [x] 11.0 GB, `raw_eia930__all_dfs.concat_pages` (just the concatenation is big!)
- [x] 7.9 GB, `core_eia860m__generators`
- [x] 7.2 GB, `_out_ferc714__georeferenced_counties`
- [x] 7.2 GB, `out_ferc714__summarized_demand`
- [x] 7.2 GB,  `compute_distance_with_year_penalty` (in FERC to FERC matching)
- [x] 6.0 GB, `out_ferc714__hourly_planning_area_demand`
- [x] 5.7 GB, `out_eia__yearly_plant_parts`
- [x] 5.0 GB, `out_eia__yearly_generators_by_ownership`
- [x] 4.8 GB, `raw_gridpathratoolkit__aggregated_extended_solar_capacity`
- [x] 4.7 GB, `out_gridpathratoolkit__hourly_available_capacity_factor`
- [x] 4.7 GB, `out_ferc714__hourly_estimated_state_demand`
- [x] 4.6 GB,  `out_pudl__yearly_assn_eia_ferc1_plant_parts`
- [x] 4.6 GB, `match_orphaned_records` (in FERC to FERC matching)
- [x] 4.0 GB, `_out_ferc714__hourly_demand_matrix`
- [x] 4.0 GB, `core_epacems__hourly_emissions.process_single_year`

The peak memory usage of the EIA930 extraction is very high, but also very brief -- it's only high during the concatenation of the dataframes after they've all been extracted. But that op is buried inside the `raw_df_factory` that's used for many of our CSV / Excel extractors, so we would need a way to selectively pass in the highmem tag, and ideally be able to apply it only to the concatenation step if appropriate.

## Numpy CPU contention

The other resource contention that we're maybe dealing with happens when we're imputing the hourly demand timeseries in the FERC-714 data. This process makes heavy use of Numpy's linear algebra functionality, which in turn depend on the BLAS and LAPACK libraries, which are multi-threaded and by default use all available CPUs. This behavior can be modulated by setting an environment variable `$OMP_NUM_THREADS` which might reduce the contention between that asset and all the other things running at the same time. Note that the linear algebra is not memory intensive, so this is a separate issue from the above high memory usage.

For reasons I don't understand, setting $OMP_NUM_THREADS=4` locally seems to cut the time to impute the hourly demand in half! At least when run in isolation. It cuts the observed CPU usage for that op from 400-600% down to 150-200%. I'll run the full ETL with this limit changed and see what overall impact it has on the runtime.

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/extract/extractor.py
+++ b/src/pudl/extract/extractor.py
@@ -262,13 +262,18 @@ class GenericExtractor(ABC):
         return all_page_dfs
 
 
-@op
+@op(tags={"memory-use": "high"})
 def concat_pages(paged_dfs: list[dict[str, pd.DataFrame]]) -> dict[str, pd.DataFrame]:
     """Concatenate similar pages of data from different years into single dataframes.
 
     Transform a list of dictionaries of dataframes into a single dictionary of
     dataframes, where each dataframe is the concatenation of dataframes with identical
     keys from the input list.
+
+    For the relatively large EIA930 dataset this is a very memory-intensive operation,
+    so the op is tagged with a high memory-use tag. For all the other datasets which use
+    this op, the time spent concatenating pages is very brief, so this tag should not
+    impact the overall concurrency of the DAG much.
 
     Args:
         paged_dfs: A list of dictionaries whose keys are page names, and values are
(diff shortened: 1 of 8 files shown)


src/pudl/extract/extractor.py
```diff
@@ -262,14 +262,19 @@ def extract(self, **partitions: PartitionSelection) -> dict[str, pd.DataFrame]:
         return all_page_dfs
 
 
-@op
+@op(tags={"memory-use": "high"})
```
zaneselvans, line 265: Only the EIA-930 concatenation is truly high memory (11GB!) but the other concatenations are extremely brief, so adding this concurrency limit to them incidentally should have very minor impacts on the overall DAG.

jdangerx (approved):
🧯

Issue #3542, opened 2024-04-04
Faulty encoding in the EIA transformers


aesharpe:
### Describe the bug
Some of the transform functions for EIA tables (specifically in EIA923, but potentially 860 as well) are running the `encode` step using resource metadata for normalized, post-harvest tables that don't have key encodable columns in their schema. This means that encodable columns in the denormalized tables are not getting encoded before being passed on to the harvesting process. The main culprit columns getting missed by this encoding error are `prime_mover_code` and `ba_code`.

**See branch `test-eia-transform-encoders`** for notebook explanation / code example

### Bug Severity
**Low** - it's not a blocking error and the changes of there being bad prime mover or balancing authority codes is relatively low. Moreover, the harvesting process itself will probably weed them out.

**However**, if there's an encodable column in a table we should be _making sure_ that it gets encoded via some sort of check rather than just relying on the person writing the transform function to pick the right table to pass to the encoder step. There is a world where the harvesting process gets messed up due to bad data bypassing the encoding step.

### To Reproduce
- Checkout the `test-eia-transform-encoders` branch 
- Open the notebook `test_encoder_in_eia923_transform.ipynb` and read the description
-  Play with the inputs to the `_core_eia923__generation` function encoder step to see how it changes depending on which table you pass in.

### Expected behavior
Passing a table to the encoder that does not have an encodable column (such as `prime_mover_code`) in its schema means that that column will not get encoded in the `core` table. 

In it's current state, I expect the following encodable columns to bypass the encoding step because the table being passed into the encode function does not have them in their schema (there could be more, this is just a cursory check of eia923 -- should check eia860 too): 

- `_core_eia923__fuel_receipts_costs`
   - `ba_code`
- `_core_eia923__cooling_system_information`
   - `ba_code`
- `_core_eia923__generation_fuel`
   - `ba_code`
- `_core_eia923__generation_fuel_nuclear`
   - `ba_code` 
- `_core_eia923__fgd_operation_maintenance`
   - `ba_code`
- `_core_eia923__generation`
   - `ba_code`
   - `prime_mover_code` 

### Suggested Fix
- We should make sure that all encodable columns are getting encoded in the transform step by passing appropriate tables to the encoder in each transform.
- We should also probably develop a _check_ that makes sure encodable columns are getting encoded.

Nancy9ice:
Hello, 
I would like to work on this issue. Do I need to work directly on the test-eia-transform-encoders branch or on the main branch?

[...]

PR #3546, opened 2024-04-05
EIA 923 energy storage transform

### LINKED ISSUE #3535: Transform raw extracted EIA923 energy storage attributes into a clean `_core_eia923__energy_storage` asset for harvesting.

### LINKED ISSUE #3537: Define a normalized `core_eia923__energy_storage_???` table
* Identify which other tables it should refer to with FK relationships.
* Might include new EIA860 energy storage equipment / facility tables if they aren't treated just as generators.

aesharpe:
# Overview

Closes #3535 and closes #3537

- Adds `core_eia923__monthly_energy_storage` table to the PUDL db.

**What did you change?**
- Made a `_core_eia923__energy_storage` transform function in the eia923 transform module.
   - Makes the data monthly
   - A lot of encoding steps to make sure encodable columns are encoded (haha). Related to this issue #3542.
- Move the `fix_boolean_columns` function from the `transform/eia860` to the `helpers` module.
   - There is already a boolean fixer column in the helper module: `convert_col_to_bool` that should probably be reconciled with this function, but I didn't want to spend time on it now.
- Added release notes.

# Testing

```[tasklist]
# To-do list
```


src/pudl/helpers.py
```diff
@@ -1779,6 +1779,25 @@ def convert_col_to_bool(
...
+    convert the columns to datatype Boolean.
+    """
+    fillna_cols = {col: pd.NA for col in boolean_columns_to_fix}
+    boolean_replace_cols = {
+        col: {"Y": True, "N": False, "X": False, "U": pd.NA}
+        for col in boolean_columns_to_fix
+    }
+    return df.fillna(fillna_cols).replace(to_replace=boolean_replace_cols)
+
+
```
aesharpe, line 1800: I moved this function over to the helpers module because I wanted to use it in the 923 transforms too. Some thoughts: 
- Maybe it should live in the `transform/eia` module instead?
- There is another boolean-transforming helper function already (`convert_col_to_bool`) that works slightly differently. Should probably either have one or the other. Having both in the `helpers` module seems weird. Could suffice to have one in eia and one in helpers. Ideally could merge them together. Idk if it's worth doing as part of this scope.

src/pudl/metadata/fields.py
```diff
@@ -1139,6 +1139,7 @@
             "A 2-3 letter code indicating the energy source (e.g. fuel type) "
             "associated with the record."
         ),
+        # Should this have an enum reference to the core_eia__codes_energy_sources table??
     },
     "energy_source_code_num": {
         "type": "string",
```
aesharpe, line 1145: I noticed that none of the `energy_source_code` fields (ex: `energy_source_code_num`, `energy_source_code_1`, `energy_source_code`) have `enums` established. There is a coding table for energy source code: `core_eia__codes_energy_sources` -- should we be referencing that in the field enum? 

Will values get encoded if there is no enum established in the field?

src/pudl/metadata/resources/eia860.py
```diff
@@ -242,6 +243,7 @@
                     "_out_eia__monthly_derived_generator_attributes",
                     "out_eia__monthly_generators",
                     "core_eia860m__changelog_generators",
+                    "core_eia923__monthly_energy_storage",
```
aesharpe, line 246: I have a strong desire to alphabetize all these table lists but have resisted. Is there a tool that can do this for us automatically in lots of places? Just a little nit that bothers me but is OOS.

src/pudl/metadata/resources/eia923.py
```diff
@@ -754,6 +757,32 @@
...
     },
+    "core_eia923__monthly_energy_storage": {
+        "description": TABLE_DESCRIPTIONS["core_eia923__monthly_energy_storage"],
+        "schema": {
+            "fields": [
+                "plant_id_eia",
+                "report_date",
+                "prime_mover_code",
+                "energy_source_code",
+                "data_maturity",
```
aesharpe, line 768: Wasn't sure if this was one of the columns to keep or drop

src/pudl/package_data/eia923/column_maps/energy_storage.csv
```diff
@@ -15,7 +15,7 @@ energy_source_code,reported_fuel_type_code,reported_fuel_type_code,reported_fuel
 fuel_type_code_aer,aer_fuel_type_code,aer_fuel_type_code,aer_fuel_type_code,aer_fuel_type_code,aer_fuel_type_code,aer_fuel_type_code,aer_fuel_type_code,aer_fuel_type_code,aer_fuel_type_code,aer_fuel_type_code
 balancing_authority_code_eia,,,,,balancing_authority_code,balancing_authority_code,balancing_authority_code,balancing_authority_code,balancing_authority_code,balancing_authority_code
 reporting_frequency_code,,,,,respondent_frequency,respondent_frequency,respondent_frequency,respondent_frequency,respondent_frequency,
-fuel_unit,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label
+fuel_units,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label,physical_unit_label
```
aesharpe, line 18: Changed this because there was already a `fuel_units` field defined.

src/pudl/package_data/eia923/column_maps/energy_storage.csv
```diff
@@ -64,8 +64,8 @@ net_generation_mwh_september,netgen_september,netgen_september,netgen_september,
...
+total_fuel_consumed_units,total_fuel_consumption_quantity,total_fuel_consumption_quantity,total_fuel_consumption_quantity,total_fuel_consumption_quantity,total_fuel_consumption_quantity,total_fuel_consumption_quantity,total_fuel_consumption_quantity,total_fuel_consumption_quantity,total_fuel_consumption_quantity,total_fuel_consumption_quantity
+total_fuel_consumed_for_electricity_units,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity,electric_fuel_consumption_quantity
+total_gross_generation_mwh,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours,gross_generation_megawatthours
```
aesharpe, line 69: Changed these to more closely mirror the name of the monthly reported values. But these columns get dropped anyways.

src/pudl/transform/eia.py
```diff
@@ -1247,10 +1252,6 @@ def harvested_entity_asset_factory(
         "_core_eia860__ownership",
         "_core_eia860__plants",
         "_core_eia860__utilities",
-        "_core_eia860__emissions_control_equipment",
-        "_core_eia860__boiler_emissions_control_equipment_assn",
-        "_core_eia860__boiler_cooling",
-        "_core_eia860__boiler_stack_flue",
```
aesharpe, line 1253: Just did some alphabetizing here....I caved just this one time.

src/pudl/transform/eia923.py
```diff
@@ -1534,3 +1534,62 @@ def fgd_continuity_check(fgd):  # pragma: no cover
...
+        pudl.metadata.classes.Package.from_resource_ids()
+        .get_resource("core_eia923__monthly_generation_fuel")
+        .encode(es_df)
+    )
+    # To encode the energy_source_code missed above
+    es_df = (
+        pudl.metadata.classes.Package.from_resource_ids()
+        .get_resource("core_eia923__monthly_energy_storage")
+        .encode(es_df)
+    )
```
aesharpe, line 1593: This feels like a lot of encoding steps for one table....related to issue in #3542

zaneselvans:
You can always click the link next to the failed check to get more information.  In this case the RTD build is failing because it can't solve the `pudl-dev` environment. Why RTD is having that problem but GitHub is not, I don't know.

aesharpe:
> You can always click the link next to the failed check to get more information.

I did this but I didn't fully understand what was going on / wasn't sure if it had to do with issues in `main`

src/pudl/transform/eia923.py
```diff
@@ -1534,3 +1534,62 @@ def fgd_continuity_check(fgd):  # pragma: no cover
...
+        pudl.metadata.classes.Package.from_resource_ids()
+        .get_resource("core_eia860__scd_generators")
+        .encode(es_df)
+    )
+    # To encode the fuel_type_code_aer values
+    es_df = (
+        pudl.metadata.classes.Package.from_resource_ids()
+        .get_resource("core_eia923__monthly_generation_fuel")
+        .encode(es_df)
+    )
```
zaneselvans, line 1587: Huh, this one is random. I guess the `fuel_type_code_aer` field is kind of random too. I'm a little surprised it's not part of the `core_eia860__scd_generators` table.

zaneselvans (changes requested):
Mostly little things, plus it looks like we're losing most of the values in the `fuel_units` column because they aren't being forced to line up with the ENUM constraint in the transform.

zaneselvans:
It looks like the version of that dependency that we added to the lockfile yesterday got yanked from `conda-forge` [because it was broken](https://github.com/conda-forge/jaraco.context-feedstock/pull/8). I'll update the lockfile on `main`!

zaneselvans:
I was poking at these little fixes locally while reviewing the PR last night and decided to just go ahead and push them since they seemed to work.

zaneselvans (approved):

src/pudl/metadata/fields.py
aesharpe: ahhh ok thanks for the clarification. Isn't the whole point of the encoding process to weed out erroneous values? Is it not duplicative to have the encoding step in the transformer and the enums defined in the field metadata?

src/pudl/metadata/resources/eia923.py
aesharpe: I guess I was thinking that an output table could pull plant name data from a harvested EIA entity table and `data_maturity` should not be a harvested value (at least between 923 and 860). I'm not sure where the `data_maturity` source of truth lives for each data source. I can look into it.

src/pudl/metadata/fields.py
zaneselvans: Maybe I'm not understanding what you're saying or thinking here, but this field does not indicate a fuel type. It indicates which of the several available `energy_source_code` slots some other column's fuel type was reported in, so we can identify which one is most important (N=1), second most important (N=2) etc.

src/pudl/metadata/resources/eia923.py
zaneselvans: `data_maturity` is inserted by us on a record-by-record basis when we read the data in from the spreadsheets, based on what kind of spreadsheet we're reading (final, early release, monthly updates, etc)

src/pudl/metadata/resources/eia923.py
```diff
@@ -123,6 +123,9 @@
 generation unit level up to the plant prime mover level, so as to be compatible with
 fossil fuel generation data."""
     ),
+    "core_eia923__monthly_energy_storage": (
+        """EIA-923 Monthly generation and fuel consumption for energy storage power plants. From EIA-923 Schedule 3."""
```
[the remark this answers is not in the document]
aesharpe: I think that's a good idea. That way if there are bad fuel_units they will get flagged and we can spot fix them.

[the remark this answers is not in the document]
aesharpe: Decided not to reference the `fuel_units` established in the `core_eia__codes_energy_sources` table because it's different for energy storage. In the context of energy storage, energy sources like pumped storage are considered to report in `mwh`. However, outside of the context of energy storage that's not the case.

src/pudl/transform/eia923.py
```diff
@@ -1534,3 +1534,80 @@ def fgd_continuity_check(fgd):  # pragma: no cover
...
+        .assign(
+            fuel_units=lambda x: x.fuel_units.map(
+                {
+                    "megawatthours": "mwh",
+                    "barrels": "barrels",
+                    "mcf": "mcf",
+                    "short tons": "short_tons",
+                }
+            )
+        )
```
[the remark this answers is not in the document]
aesharpe: Is there any chance is should just be `mwh`? pumped storage is inherently a form of storage so it makes sense that the energy_storage table would be the source of truth for what units it is reported in?

src/pudl/metadata/resources/eia923.py
aesharpe: It looks like `data_maturity` is in all the `core` table schemas so I'm going to leave it there

src/pudl/transform/eia923.py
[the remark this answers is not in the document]
zaneselvans: I don't understand what you're saying.

zaneselvans (approved):
Let's get this merged in.

src/pudl/transform/eia923.py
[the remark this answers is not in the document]
zaneselvans: In the energy source codes table, the fuel unit is linked to the energy source, and for renewables (WND, WAT, SUN) there is no fuel -- and so it has an NA value.  Except that in a storage specific context, pumped storage (which also has energy source WAT) **does** have a "fuel" which is MWH, but that's not something you can infer from the energy source -- it takes a combination of the energy source and the pumped storage context, which we don't have in the energy source code table.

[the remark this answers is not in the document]
aesharpe: This = the official `fuel_units` value for pumped storage stored in the `core_eia__codes_energy_sources` table.

PR #3548, opened 2024-04-06
Tag additional assets as high memory usage

zaneselvans:
# Overview

We're still getting an out-of-memory error even with the concurrency limits on high memory assets added in #3541 & #3543. Looking at the memory usage of last night's failed build and the previous night's successful one, we are still havin regular but transient spikes to >60GB of memory usage. So it seems like maybe we just got lucky the night before. 😞 Maybe there are additional high memory assets we need to tag?

Or maybe the peak memory usage of the ones that are tagged is much higher than it appeared during visual inspection due to low-frequency sampling?  I re-ran  the FERC-714 demand matrix with 100ms sampling instead of 2000ms, and the max was 9.2GB, which was significantly higher than I'd seen before.

This PR adds a [`RetryPolicy`](https://docs.dagster.io/_apidocs/ops#dagster.RetryPolicy) to the highest memory assets, so that if they fail due to temporarily high memory usage, they get another chance to run. (assuming Dagster catches the "the OS killed your op" exception and that triggers the retry)

## Successful 2024-04-05
image

## Failed 2024-04-06
image

```[tasklist]
# To-do list
- [ ] Attempt running a branch build and see if the retries get exercised / work as expected
```

### DIFF
--- a/src/pudl/analysis/state_demand.py
+++ b/src/pudl/analysis/state_demand.py
@@ -24,7 +24,7 @@ from typing import Any
 import geopandas as gpd
 import numpy as np
 import pandas as pd
-from dagster import AssetOut, Field, asset, multi_asset
+from dagster import AssetOut, Backoff, Field, Jitter, RetryPolicy, asset, multi_asset
 
 import pudl.analysis.timeseries_cleaning
 import pudl.logging_helpers
@@ -270,7 +270,8 @@ def load_hourly_demand_matrix_ferc714(
     """Read and format FERC 714 hourly demand into matrix form.
 
     Args:
-        out_ferc714__hourly_planning_area_demand: FERC 714 hourly demand time series by planning area.
+        out_ferc714__hourly_planning_area_demand: FERC 714 hourly demand time series by
+            planning area.
 
     Returns:
         Hourly demand as a matrix with a `datetime` row index
@@ -449,6 +450,12 @@ def melt_ferc714_hourly_demand_matrix(
             ),
         ),
     },
+    retry_policy=RetryPolicy(
+        max_retries=3,
+        delay=60,  # 1 minute
+        backoff=Backoff.EXPONENTIAL,
+        jitter=Jitter.PLUS_MINUS,
+    ),
     op_tags={"memory-use": "high"},
 )
 def _out_ferc714__hourly_demand_matrix(
--- a/src/pudl/analysis/state_demand.py
+++ b/src/pudl/analysis/state_demand.py
@@ -24,7 +24,7 @@ from typing import Any
 import geopandas as gpd
 import numpy as np
 import pandas as pd
-from dagster import AssetOut, Backoff, Field, Jitter, RetryPolicy, asset, multi_asset
+from dagster import AssetOut, Field, asset, multi_asset
 
 import pudl.analysis.timeseries_cleaning
 import pudl.logging_helpers
@@ -434,7 +434,7 @@ def melt_ferc714_hourly_demand_matrix(
 
 
 @asset(
-    compute_kind="Python",
+    compute_kind="pandas",
     config_schema={
         "min_data": Field(
             int,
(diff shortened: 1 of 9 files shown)


zaneselvans:
Ah. Unfortunately the RetryPolicy will not catch this kind of exception right now, though there is [a feature request](https://github.com/dagster-io/dagster/issues/16168) open.

src/pudl/extract/ferc714.py
```diff
@@ -64,13 +64,14 @@
 """Dictionary mapping PUDL tables to FERC-714 filenames and character encodings."""
 
 
-def generate_raw_ferc714_asset(table_name: str) -> AssetsDefinition:
+def raw_ferc714_asset_factory(table_name: str) -> AssetsDefinition:
```
zaneselvans, line 67: Standardize naming to match other asset factories.

src/pudl/extract/eia930.py
```diff
@@ -47,40 +47,34 @@ def process_raw(
...
+        op_tags={"memory-use": "high"},
+        retry_policy=RetryPolicy(
+            max_retries=3,
+            delay=60,  # 1 minute
+            backoff=Backoff.EXPONENTIAL,
+            jitter=Jitter.PLUS_MINUS,
+        ),
+        compute_kind="pandas",
+    )
+    def _extract_raw_eia930(
```
zaneselvans, line 64: Replaces 3 individual assets using the same pattern with an asset factory (though the diff is a bit messy).

Also, tag these 3 raw EIA-930 as high memory since they can very briefly use 9-8GB of memory, and they seem to be running around the same time as the FERC-714 asset that is crashing out the nightly builds, and there are 3 of them that all run simultaneously.

src/pudl/extract/eia930.py
```diff
@@ -47,40 +47,28 @@ def process_raw(
...
+    """Asset factory for individual raw EIA 930 dataframes."""
 
-@asset(
-    required_resource_keys={"datastore", "dataset_settings"},
-)
-def raw_eia930__interchange(raw_eia930__all_dfs):
-    """Extract raw EIA 930 interchange data from CSV sheets into dataframes.
+    @asset(
+        name=f"raw_eia930__{page}",
+        op_tags={"memory-use": "high"},
```
zaneselvans, line 55: The loading of the individual EIA-930 dataframes appears to be the proximate cause of our recent failures.  These assets just read the pickled dictionary of dataframes from disk, and then immediately write the dataframes back to disk as individual pickled files. Which means each of them can spike to ~10GB of memory, and they all happen at exactly the same time, and this frequently happens to coincide with the (also) high memory FERC-714 demand matrix asset.

zaneselvans:
Between these two runs I added more highmem tags, and some of the spikes to 100% memory utilization definitely went away, so this does seem to be addressing the problem. But also, there are still at least half a dozen instances where we get a memory spike, so I think there are more high-memory assets stacking on top of each other, but for some reason not crashing the processes all the time. If we start getting SIGKILL crashes again, I guess we can do another round of high memory asset hunting.

## FAILED
image

## PASSED
image

pudlbot (approved):

src/pudl/analysis/state_demand.py
```diff
@@ -433,7 +434,7 @@ def melt_ferc714_hourly_demand_matrix(
 
 
 @asset(
-    compute_kind="Python",
+    compute_kind="pandas",
```
bendnorman, line 437: I started adding these `compute_kind` tags to the `out` assets when I thought we were going to have some SQL views mixed in. I don't know how we want to use them going forward, given 99% of our assets use pandas.

src/pudl/extract/eia930.py
bendnorman: I haven't dug into the 930 code but would it be appropriate to use dask here so we aren't bringing 10 GB of ram into memory?

PR #3552, opened 2024-04-08
Rename BA & Utility service territory tables to use conventions

zaneselvans:
# Overview

I came across these tables which didn't really follow our naming conventions when doing the highmem tagging and decided to fix them.

# Testing

* Ran the full ETL locally.

```[tasklist]
# To-do list
```


src/pudl/analysis/service_territory.py
```diff
@@ -30,8 +30,6 @@
 MAP_CRS = "EPSG:3857"  # For mapping w/ OSM baselayer tiles
 CALC_CRS = "ESRI:102003"  # For accurate area calculations
 
-ENTITY_TYPE = {"ba": "balancing_authorities", "util": "utilities"}
-
```
zaneselvans, line 34: It was simpler to just avoid having this mapping. Maybe I should have made an `Enum`?

src/pudl/output/pudltabl.py
```diff
@@ -219,8 +219,8 @@ def _register_output_methods(self: Self):
             "out_ferc714__respondents_with_fips": "fipsified_respondents_ferc714",
             "out_ferc714__summarized_demand": "summarized_demand_ferc714",
             # service territory
-            "out_eia861__compiled_geometry_balancing_authorities": "compiled_geometry_balancing_authority_eia861",
-            "out_eia861__compiled_geometry_utilities": "compiled_geometry_utility_eia861",
+            "out_eia861__yearly_balancing_authority_service_territory": "compiled_geometry_balancing_authority_eia861",
+            "out_eia861__yearly_utility_service_territory": "compiled_geometry_utility_eia861",
```
zaneselvans, line 223: I left the `PudlTabl` interface unchanged.

src/pudl/analysis/service_territory.py
```diff
@@ -384,14 +383,14 @@ def _limit_years(df: pd.DataFrame) -> pd.DataFrame:
     return geom
 
 
-def compiled_geoms_asset_factory(
-    entity_type: Literal["ba", "util"],
+def service_territory_asset_factory(
+    entity_type: Literal["balancing_authority", "utility"],
```
zaneselvans, line 387: We aren't actually storing any geometry information in the table (though we could, if we wanted to use GeoParquet and/or SpatiaLite) so naming these things "geometry" seemed misleading.

src/pudl/analysis/service_territory.py
```diff
@@ -450,15 +449,14 @@ def dagster_compile_geoms(
             save_format=save_format,
         )
 
-    return [dagster_compile_geoms]
```
e-belfer, line 453: Is there a reason this was being stored in a list before and isn't any more?

e-belfer (approved):
One little question, otherwise looks good and makes sense to update!

src/pudl/analysis/service_territory.py
zaneselvans: I also found this confusing, but it seemed unnecessary, and it resulted in a second layer of iteration when generating the assets which also seemed unnecessary, and I removed it kind of expecting to break something, but then nothing broke, so it really did seem to be unnecessary. 🤷🏼

Issue #3147, opened 2023-12-11
Consolidate ferc1 outputs using Dagster asset factories


zaneselvans:
The top portion of the `pudl.output.ferc1` module contains a number of individual asset definitions for denormalized / output tables with very similar structures, which could be consolidated into a small number of asset factories using the pattern adopted in e.g. `pudl.extract.ferc714` (after PR #3123). See Dagster's blog post [Factory Patterns in Python](https://dagster.io/blog/python-factory-patterns) for some more background on the factory design pattern, and its application to Dagster assets.

Note that the calls to `pudl.helpers.organize_cols()` found in the current FERC 1 output asset definitions are no longer required, as the ordering of columns in the database is determined by the resource definitions / database schema now. These calls are leftover from when we were producing dataframes for users on request rather than writing these tables to the database.

Note that some of these assets currently create new columns containing derived values, and those would need to be preserved, either with their own asset definitions, or some way of keeping track of which calculations should be done for what tables inside the asset factory.

hfireborn:
@catalyst-cooperative/com-dev Is this still open? I'd like to work on this as a first time contributor

zaneselvans:
Hey there! Yes, this is still open. I was thinking about this as a good one after getting your office hours signup. There are lots of other examples of asset factories floating around that you could use as a guide. If you have a chance to get the PUDL / Dagster local development environment running, this should be a pretty easy thing to test out locally.

PR #3557, opened 2024-04-11
Create an asset factory to generate FERC1 output tables

hfireborn:
draft of asset factories for #3147.

Based on the tutorial in https://dagster.io/blog/python-factory-patterns, creates 3 assets using a factory for: out_ferc1__yearly_purchased_power_and_exchanges_sched326
out_ferc1__yearly_plant_in_service_sched204
out_ferc1__yearly_balance_sheet_assets_sched110

all of which just merge and then organize columns.

a few questions: 
1. Is this the type of solution you were looking for? 
2. double checking: Is it ok to omit the organize_cols parts of those functions?
3. We are currently omitting the pd.DataFrame input type specification in these assets, is this ok?


src/pudl/output/ferc1.py
```diff
@@ -228,6 +228,55 @@ def _out_ferc1__yearly_plants_utilities(
...
+            #     pudl.helpers.organize_cols,
+            #     [
+            #         "report_year",
+            #         "utility_id_ferc1",
+            #         "utility_id_pudl",
+            #         "utility_name_ferc1",
+            #         "seller_name",
+            #         "record_id",
+            #     ],
+            # )
```
zaneselvans, line 270: Yes, the `organize_cols` can and should be removed from *all* of the asset definitions in this module, since the database schema is now what determines the ordering of the columns.

src/pudl/output/ferc1.py
```diff
@@ -228,6 +228,55 @@ def _out_ferc1__yearly_plants_utilities(
...
+        "df2": "core_pudl__assn_ferc1_pudl_utilities",
+        "mg": "utility_id_ferc1",
+    },
+]
+
+
+# draft asset_factory
+def generate_asset_factory(spec):
+    @asset(io_manager_key="pudl_io_manager", compute_kind="Python")
+    def _asset() -> pd.DataFrame:
```
zaneselvans, line 257: The function which is used to define the asset needs to have arguments, because that's how Dagster infers the dependencies between assets. For more complex cases they can also be defined with arguments in the asset decorator, but I don't think that's necessary here. You can't just add them inside the inner function definition.

Search for "_asset_factory" in the code base to find other examples of how we're using this pattern.

src/pudl/output/ferc1.py
```diff
@@ -228,6 +228,55 @@ def _out_ferc1__yearly_plants_utilities(
     )
 
 
+## example/draft factory pattern
+specs = [
+    {
+        "name": "out_ferc1__yearly_purchased_power_and_exchanges_sched326",
+        "df1": "core_ferc1__yearly_purchased_power_and_exchanges_sched326",
+        "df2": "core_pudl__assn_ferc1_pudl_utilities",
```
zaneselvans, line 236: Note that in all cases, `df2` is the same. So it doesn't need to be part of these inputs. It can be hard-coded as one of the input assets that are used by the asset definition (I think it's used in all of the FERC 1 output assets)

src/pudl/output/ferc1.py
```diff
@@ -228,6 +228,55 @@ def _out_ferc1__yearly_plants_utilities(
     )
 
 
+## example/draft factory pattern
+specs = [
+    {
+        "name": "out_ferc1__yearly_purchased_power_and_exchanges_sched326",
```
zaneselvans, line 234: Note that in all cases, there's an easy way to construct the input (core) and output asset names, given just a single string, (`purchased_power_and_exchanges_sched326` in this case) -- they all have the same prefix structure. So you only need to store that string.

src/pudl/output/ferc1.py
```diff
@@ -228,6 +228,55 @@ def _out_ferc1__yearly_plants_utilities(
...
+        "df1": "core_ferc1__yearly_purchased_power_and_exchanges_sched326",
+        "df2": "core_pudl__assn_ferc1_pudl_utilities",
+        "mg": "utility_id_ferc1",
```
zaneselvans, line 237: Similarly, they all seem to get merged on the same column, so this doesn't need to be part of the specification -- it can just be hard-coded inside the inner factory function definition (unless there are some tables that need a different merge key)

zaneselvans (changes requested):
This is generally the shape of the solution we're looking for, but see my inline comments in the PR.

It looks like you don't have the pre-commit hooks installed locally, or the linters set up in VS Code, as the pre-commit checks and unit tests are failing on the PR, so you'll want to get that set up. The documentation build is also failing. It looks like you've got an undefined variable.

You'll definitely need to try running the ETL via Dagster locally to know whether your changes are working, and debug issues. You'll probably need to run it many many times, but you can just run the portion of the DAG that you're working on, in this case the FERC 1 related parts.

src/pudl/output/ferc1.py
hfireborn: @zaneselvans If all of the helpers.organize_cols can be removed, is it necessary to make an asset factory for this? It looks to me like all of the functions are essentially doing the same thing then; they take two DataFrames as inputs, merge them on something hard-coded and the same across all the functions, and then return.

Is there a reason why we can't we just simplify them all into one function and change where the functions are called?

zaneselvans: We do need an asset factory that pumps out the function definitions, because those `@asset` decorated function definitions and their inputs and outputs are what define the dependency graph in Dagster. It might help to read through the [core Dagster concepts](https://docs.dagster.io/concepts) or some of their examples, and also to try running the ETL through the Dagster UI locally if you haven't already.

zaneselvans: Of course each of the assets (functions) defined by the factory will be very similar (which is why we want to use a factory) but it's the additional information from the decorator or the arguments and return value that stitch the functions and the data flowing through them together.

hfireborn:
> We've created an asset factory that appeared to work locally on dagster. When trying to push our code, we ran into some errors getting our codebase up to date that we are currently resolving.

@zaneselvans  For some reason we are failing unit tests when we try to commit that seem unrelated to the code in output/ferc1.py we were working with.

=========================== short test summary info ============================
FAILED test/unit/workspace/datastore_test.py::TestDatapackageDescriptor::test_get_partition_filters - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestDatapackageDescriptor::test_get_resource_path - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestDatapackageDescriptor::test_get_resources_filtering - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestDatapackageDescriptor::test_json_string_representation - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestDatapackageDescriptor::test_modernize_zenodo_legacy_api_url - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestZenodoFetcher::test_doi_format_is_correct - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestZenodoFetcher::test_doi_of_prod_epacems_matches - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestZenodoFetcher::test_get_descriptor_http_calls - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestZenodoFetcher::test_get_known_datasets - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestZenodoFetcher::test_get_resource - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestZenodoFetcher::test_get_resource_with_invalid_checksum - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestZenodoFetcher::test_get_resource_with_nonexistent_resource_fails - TypeError: Package.metadata_validate() missing 1 required positional argume...
FAILED test/unit/workspace/datastore_test.py::TestZenodoFetcher::test_get_unknown_dataset - TypeError: Package.metadata_validate() missing 1 required positional argume...
= 13 failed, 1367 passed, 1 skipped, 3 deselected, 9 xfailed, 44 warnings in 32.77s =

conda-lock...........................................(no files to check)Skipped


We thought it was a package update issue, guessing that we were pulling code from the remote but not updating the packages that accompany them, so we tried:

make conda-clean
make conda-lock.yml
make install-pudl

but we are still receiving these errors. Not sure where to go from here, any help is appreciated!

zaneselvans:
Those unit test errors look like they result from having an older version of the `frictionless` package installed in your environment. We recently updated from v4 to v5.

I would remove any local changes to the conda lockfiles, remove your existing pudl-dev environment, make sure your local branch has all the most recent changes from `main` upstream, and rebuild the conda environment, which should look something like:

```console
git checkout -- environments
mamba deactivate
mamba env remove -n pudl-dev
git pull
make install-pudl
mamba activate pudl-dev
```

Given the changes you're making, you shouldn't need to run `make conda-clean` or `make conda-lock.yml`, which will update/change the locked dependencies.

src/pudl/output/ferc1.py
```diff
@@ -422,6 +445,53 @@ def out_ferc1__yearly_steam_plants_fuel_sched402(
...
+        core_: pd.DataFrame,
+        core_pudl__assn_ferc1_pudl_utilities: pd.DataFrame,
+    ) -> pd.DataFrame:
+        """Generate a dataframe for {} asset specification.""".format(spec["name"])
+        return_df = core_.merge(
+            core_pudl__assn_ferc1_pudl_utilities, on="utility_id_ferc1"
+        )
+        return return_df
+
+    return _asset
```
zaneselvans, line 476: A more specific name would be helpful, and I think this can be as simple as just taking a string as input:

```suggestion
def ferc1_output_asset_factory(table_name: str) -> AssetsDefinition:
    @asset(
        name=f"out_ferc1__{table_name}",
        ins={
            core_pudl__assn_ferc1_pudl_utilities: AssetIn(),
            f"core_ferc1__{table_name}": AssetIn(),
        },
        io_manager_key="pudl_io_manager",
        compute_kind="Python",
    )
    def _ferc1_output_asset(**kwargs) -> pd.DataFrame:
        f"""Generate a dataframe for out_ferc1__{table_name} asset specification."""
        return kwargs[core_ferc1__{table_name}].merge(
            core_pudl__assn_ferc1_pudl_utilities, on="utility_id_ferc1"
        )

    return _ferc1_output_asset
```

src/pudl/output/ferc1.py
```diff
@@ -422,6 +445,53 @@ def out_ferc1__yearly_steam_plants_fuel_sched402(
...
+    """Create a list of generated FERC Form 1 assets.
+
+    Returns:
+        A list of :class:`AssetsDefinitions` where each asset is an generated FERC Form 1
+        table.
+    """
+    return [generate_asset_factory(**kwargs) for kwargs in specs]
+
+
+exploded_ferc1_assets = create_generated_assets()
```
zaneselvans, line 492: I don't think a whole separate function is needed here, we can generate the `AssetsDefinitions` for dagster to pick up on module import with something like:

```py
ferc1_output_assets = [
    ferc1_output_asset_factory(table_name) for table name in
    [
        "yearly_purchased_power_and_exchanges_sched326",
        "yearly_plant_in_service_sched204",
        "yearly_balance_sheet_assets_sched110",
    ]
]
```

zaneselvans (changes requested):

hfireborn:
@zaneselvans our most recent commit from right before our meeting has some updated changes to it as well as fixing the merge conflicts, if you are able to take a look at that.

src/pudl/output/ferc1.py
```diff
@@ -422,6 +445,53 @@ def out_ferc1__yearly_steam_plants_fuel_sched402(
     return fuel_df
 
 
+## example/draft factory pattern
+specs = [
+    {"name": "ferc1__yearly_purchased_power_and_exchanges_sched326"},
+    {"name": "ferc1__yearly_plant_in_service_sched204"},
+    {"name": "ferc1__yearly_balance_sheet_assets_sched110"},
+]
```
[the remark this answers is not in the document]
zaneselvans: Please remove the `specs` variable and just list a list of the table names with no prefix in a list comprehension to generate the assets.

src/pudl/output/ferc1.py
```diff
@@ -422,73 +422,53 @@ def out_ferc1__yearly_steam_plants_fuel_sched402(
...
+        name=f"out_{spec['name']}",
+        io_manager_key="pudl_io_manager",
+        compute_kind="Python",
+        ins=ins,
     )
-    return pis_df
+    def _asset(
+        **kwargs: dict[str, pd.DataFrame],
+    ) -> pd.DataFrame:
+        """Generate a dataframe for {} asset specification.""".format(spec["name"])
```
zaneselvans, line 452: Please use f-string formatting here.

src/pudl/output/ferc1.py
```diff
@@ -422,73 +422,53 @@ def out_ferc1__yearly_steam_plants_fuel_sched402(
...
-                "utility_id_pudl",
-                "utility_name_ferc1",
-                "record_id",
-                "asset_type",
-            ],
-        )
-    )
-    return out_ferc1__yearly_balance_sheet_assets_sched110
+
+def create_generated_assets() -> list[AssetsDefinition]:
```
zaneselvans, line 461: Please remove this function and use a list comprehension that iterates of the table names with no prefix to call the asset factory function.

src/pudl/output/ferc1.py
```diff
@@ -422,73 +422,53 @@ def out_ferc1__yearly_steam_plants_fuel_sched402(
...
-        pudl.helpers.organize_cols,
-        [
-            "report_year",
-            "utility_id_ferc1",
-            "utility_id_pudl",
-            "utility_name_ferc1",
-            "record_id",
-        ],
+# draft asset_factory
+def generate_asset_factory(spec) -> AssetsDefinition:
```
zaneselvans, line 434: Please rename to `ferc1_output_asset_factory` and have the factory depend on a string, which is the table name with no prefix, rather than `spec`.

zaneselvans (changes requested):

zaneselvans:
I think the semester ended for the students who were working on this issue / PR so it's probably not going to get merged unless we can rope someone else into working on it... @seeess1 would you be interested by chance? Mostly it's just some Dagster asset factory stuff.  Once we get your environment back to working!

seeess1:
> I think the semester ended for the students who were working on this issue / PR so it's probably not going to get merged unless we can rope someone else into working on it... @seeess1 would you be interested by chance? Mostly it's just some Dagster asset factory stuff. Once we get your environment back to working!

Yep would be happy to take a look. Hoping to get to the point where I can load assets in a notebook. We can try and resolve that first in Slack and then I can take a look here.

e-belfer:
Will wrap this up in #3883, closing this PR.

PR #3558, opened 2024-04-12
Draft of package-level field encoding, applied to EIA

zaneselvans:
# Overview

This PR adds a `Package.encode()` method, which can be used on any dataframe, and which will use our coding tables to fix any column that has an encoder associated with it in the Package. This can be used to avoid the problem identified in #3542, where columns which are not part of a defined resource / database table schema need to be cleaned up before they're fed into the harvesting process.

I apply this Package level encoding to the inputs into the harvesting process, and to the core EIA tables (which aren't generated by the harvesting process) across the board, in the asset factories which define both sets of tables.

# Testing

Ran the EIA harvesting process locally, and also all of the core (not harvested) EIA tables.

```[tasklist]
# To-do list
- [x] Create a single `pudl.metadata.classes.PUDL_PKG` and reference it everyewhere else.
- [x] Remove / replace existing `encode()` calls with the package-level encoder.
- [x] Figure out how conveniently ensure encoding happens to ALL EIA tables
- [x] Remove redundant encoding from `harvest_entity_tables()`
- [x] Fix errors resulting from more complete encoding
- [x] Make an encoder unit test
- [x] Ensure docs build, unit & integration tests, and test coverage pass locally with `make pytest-coverage`
```


src/pudl/metadata/classes.py
```diff
@@ -526,6 +527,22 @@ def to_rst(
         )
         return rendered
 
+    def generate_encodable_data(self: Self, size: int = 10) -> pd.Series:
```
zaneselvans, line 530: I created this for testing, and it was useful, so I thought I would leave it.  Maybe there's  a relevant unit test that should be created for the coding process?

src/pudl/transform/eia.py
```diff
@@ -37,14 +37,15 @@
...
 from pudl.helpers import convert_cols_dtypes
-from pudl.metadata.classes import Package
 from pudl.metadata.enums import APPROXIMATE_TIMEZONES
 from pudl.metadata.fields import apply_pudl_dtypes, get_pudl_dtypes
 from pudl.metadata.resources import ENTITIES
 from pudl.settings import EiaSettings
 
 logger = pudl.logging_helpers.get_logger(__name__)
 
+PUDL_PKG = pudl.metadata.classes.Package.from_resource_ids()
```
zaneselvans, line 47: Compiling the dictionary of encoders only takes 50ms (1/20th of a second) but it does seem silly to do it over and over again, and in general the pattern we have now of recreating the Package from scratch everywhere many times seems kind of silly. This doesn't have to be defined here, but as an example it works.

src/pudl/transform/eia.py
```diff
@@ -1278,6 +1274,9 @@ def harvested_entity(context, **clean_dfs):
         logger.info(f"Harvesting IDs & consistent static attributes for EIA {entity}")
         eia_settings = context.resources.dataset_settings.eia
         debug = context.op_config["debug"]
+        clean_dfs = {
+            df_name: PUDL_PKG.encode(clean_dfs[df_name]) for df_name in clean_dfs
+        }
```
zaneselvans, line 1279: This ensures that every dataframe being fed into the harvesting process has been encoded -- including columns that aren't part of any schema.

src/pudl/transform/eia.py
```diff
@@ -1318,9 +1317,12 @@ def finished_eia_asset_factory(
     )
     def finished_eia_asset(**kwargs) -> pd.DataFrame:
         """Enforce PUDL DB schema on a cleaned EIA dataframe."""
-        df = convert_cols_dtypes(kwargs[_core_table_name], data_source="eia")
-        res = Package.from_resource_ids().get_resource(table_name)
-        return res.enforce_schema(df)
+        res = PUDL_PKG.get_resource(table_name)
+        return (
+            PUDL_PKG.encode(kwargs[_core_table_name])
```
zaneselvans, line 1322: Ensures that all of the EIA core data tables which aren't produced by the harvesting process also get encoded.

src/pudl/transform/eia.py
cmgosnell: I don't think this package actually compiles the dictionary. If i understand correctly the `cached_property` only compiles the dictionary the first time the property is accessed

cmgosnell: if there is no variation in all of the gazillion places where we are currently recreating the Package.... should we just define this in `metadata.classes` or somewhere in metadata and use that across the board?

src/pudl/transform/eia.py
```diff
@@ -661,11 +658,8 @@ def harvest_entity_tables(  # noqa: C901
     logger.info(f"Average consistency of static {entity.value} values is {mcs:.2%}")
 
     # Apply standard PUDL data types to the new entity tables:
-    pkg = Package.from_resource_ids()
-    entity_res = pkg.get_resource(f"core_eia__entity_{entity.value}")
-    entity_df = apply_pudl_dtypes(entity_df, group="eia").pipe(entity_res.encode)
-    annual_res = pkg.get_resource(f"core_eia860__scd_{entity.value}")
-    annual_df = apply_pudl_dtypes(annual_df, group="eia").pipe(annual_res.encode)
+    entity_df = PUDL_PKG.encode(apply_pudl_dtypes(entity_df, group="eia"))
+    annual_df = PUDL_PKG.encode(apply_pudl_dtypes(annual_df, group="eia"))
```
cmgosnell, line 662: this seems fine to do but also in theory wouldn't this be duplicative if all of the inputs to the harvesting process are already encoded?

src/pudl/metadata/classes.py
```diff
@@ -2028,6 +2045,41 @@ def sort_resource_names(resource: Resource):
...
+            if field.encoder
+        ]
+        encoders: dict[SnakeCase, Encoder] = {}
+        for field in encoded_fields:
+            if field.name not in encoders:
+                encoders[field.name] = field.encoder
+            else:
+                pd.testing.assert_frame_equal(encoders[field.name].df, field.encoder.df)
+                assert encoders[field.name].code_fixes == field.encoder.code_fixes
+                assert encoders[field.name].ignored_codes == field.encoder.ignored_codes
```
cmgosnell, line 2068: this is all just ensuring that if the field exists more than once in `encoded_fields` the encoding components are exactly the same? Can you add a comment in here about that? it took me a second to understand.

src/pudl/metadata/classes.py
```diff
@@ -2028,6 +2045,41 @@ def sort_resource_names(resource: Resource):
...
+        """Clean up all coded columns in a dataframe based on PUDL coding tables.
+
+        Returns:
+            A modified copy of the input dataframe.
+        """
+        encoded_df = df.copy()
+        for col in df.columns:
+            if col in self.encoders:
+                encoded_df[col] = self.encoders[col].encode(df[col])
+        return encoded_df
```
cmgosnell, line 2081: complete nit but it avoids a layer of nesting:
```suggestion
        encoded_df = df.copy()
        encoder_df_cols = [col for col in df if col in self.encoders]
        for encode_col in encoder_df_cols:
            encoded_df[col] = self.encoders[col].encode(df[col])
        return encoded_df
```

also are you meaning to copy to df as `encoded_df` but keep using `df` (in the column loop and in the `encode(df[col]`?

src/pudl/metadata/classes.py
zaneselvans: Yes, that's right. I'll add a comment.

src/pudl/metadata/classes.py
zaneselvans: I did intentionally leave `df` in the "source" role down below but that's being a little too clever. I'll swap it out.

src/pudl/transform/eia.py
zaneselvans: There's like 50 places where we create the full Package and they're always exactly the same and imutable. I'd love to put this in the `metadata.classes` and just reference it there from everywhere. It'll have to be at the bottom after the `Package` has been defined but that's fine.

src/pudl/transform/eia.py
zaneselvans: Definitely. Where's the best place to inject the encoding for all the tables? We don't have a `transform_end()` step here like we do for FERC 1, but it would be nice to ensure that every table gets encoded by default before it comes out of the `_core_eia*` step.

src/pudl/transform/eia.py
```diff
@@ -424,11 +425,7 @@ def _compile_all_entity_records(
     # most columns become objects (ack!), so assign types
     compiled_df = apply_pudl_dtypes(compiled_df, group="eia")
     # encode the compiled options!
-    compiled_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource(f"core_eia860__scd_{entity.value}")
-        .encode(compiled_df)
-    )
+    compiled_df = PUDL_PKG.encode(compiled_df)
```
zaneselvans, line 428: @cmgosnell so is this encoding also redundant now?

zaneselvans: Yes, yes it is.

src/pudl/metadata/classes.py
zaneselvans: I created a unit test that makes use of this function to generate encodable data, and then encode it for each encoder.

src/pudl/metadata/classes.py
```diff
@@ -2028,6 +2045,48 @@ def sort_resource_names(resource: Resource):
...
+        encoded_df = df.copy()
+        for col in encoded_df.columns:
+            if col in self.encoders:
+                encoded_df[col] = self.encoders[col].encode(
+                    encoded_df[col], dtype=Field.from_id(col).to_pandas_dtype()
+                )
+        return encoded_df
+
+
+PUDL_PACKAGE = Package.from_resource_ids()
```
zaneselvans, line 2088: This has to happen after the definition of the `Package` class, and I did it here rather in `pudl/metadata/__init__.py` because it's used below a couple of times, but I import it at the metadata package level to keep the path shorter and kind of unbury it from this middle of the module definition.

src/pudl/transform/eia.py
```diff
@@ -1278,6 +1268,9 @@ def harvested_entity(context, **clean_dfs):
         logger.info(f"Harvesting IDs & consistent static attributes for EIA {entity}")
         eia_settings = context.resources.dataset_settings.eia
         debug = context.op_config["debug"]
+        clean_dfs = {
+            df_name: PUDL_PACKAGE.encode(clean_dfs[df_name]) for df_name in clean_dfs
+        }
```
zaneselvans, line 1273: Ensures that all dataframes headed into the harvesting process are encoded, if they weren't in their transforms already.

src/pudl/transform/eia.py
```diff
@@ -423,12 +424,6 @@ def _compile_all_entity_records(
     logger.debug("    Casting harvested IDs to correct data types")
     # most columns become objects (ack!), so assign types
     compiled_df = apply_pudl_dtypes(compiled_df, group="eia")
-    # encode the compiled options!
-    compiled_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource(f"core_eia860__scd_{entity.value}")
-        .encode(compiled_df)
-    )
```
zaneselvans, line 431: No longer required, as encoding happens earlier.

src/pudl/transform/eia860.py
```diff
@@ -156,11 +157,7 @@ def _core_eia860__ownership(raw_eia860__ownership: pd.DataFrame) -> pd.DataFrame
         own_df.operator_utility_id_eia == own_df.owner_utility_id_eia
     ) & (own_df.fraction_owned == 1.0)
     own_df.loc[single_owner_operator, "operator_utility_id_eia"] = pd.NA
-    own_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia860__scd_ownership")
-        .encode(own_df)
-    )
+    own_df = PUDL_PACKAGE.encode(own_df)
```
zaneselvans, line 160: Needs to happen early because we're working with the state/country codes below.

src/pudl/transform/eia860.py
```diff
@@ -312,12 +309,20 @@ def _core_eia860__generators(
...
+    # PA -> PACW in Oregon
+    gens_df.loc[
+        (gens_df.state == "OR") & (gens_df.balancing_authority_code_eia == "PA"),
+        "balancing_authority_code_eia",
+    ] = "PACW"
+    # PA -> PACE in Utah
+    gens_df.loc[
+        (gens_df.state == "UT") & (gens_df.balancing_authority_code_eia == "PA"),
+        "balancing_authority_code_eia",
+    ] = "PACE"
```
zaneselvans, line 324: Relocated from the `eia860m` specific transform to allow the spot fix to happen before the encoding in all cases.

src/pudl/transform/eia860.py
```diff
@@ -568,14 +556,8 @@ def _core_eia860__plants(raw_eia860__plant: pd.DataFrame) -> pd.DataFrame:
...
-    p_df = pudl.helpers.convert_to_date(p_df)
-
-    p_df = clean_nerc(p_df, idx_cols=["plant_id_eia", "report_date", "nerc_region"])
-
-    p_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia860__scd_plants")
-        .encode(p_df)
+    p_df = pudl.helpers.convert_to_date(p_df).pipe(
+        clean_nerc, idx_cols=["plant_id_eia", "report_date", "nerc_region"]
```
zaneselvans, line 560: Removed the encoding step because it's the last part of the transform, and now happens across the board in the harvesting process.

src/pudl/transform/eia860.py
```diff
@@ -593,24 +575,18 @@ def _core_eia860__boiler_generator_assn(
...
-    b_g_df = pudl.helpers.convert_cols_dtypes(df=b_g_df, data_source="eia")
-    b_g_df = b_g_df.drop_duplicates()
-
     b_g_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia860__assn_boiler_generator")
-        .encode(b_g_df)
+        pudl.helpers.convert_to_date(raw_eia860__boiler_generator_assn)
+        .pipe(pudl.helpers.convert_cols_dtypes, data_source="eia")
+        .drop_duplicates()
```
zaneselvans, line 589: Removed the encoding step because it's the last part of the transform, and now happens in the harvesting process for this table.

src/pudl/transform/eia860.py
```diff
@@ -897,12 +873,6 @@ def _core_eia860__boilers(
         b_df.loc[b_df.report_date.dt.year < 2012, "efficiency_50pct_load"] / 100
     )
 
-    b_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia860__scd_boilers")
-        .encode(b_df)
-    )
-
```
zaneselvans, line 905: Removed the encoding step because it's the last part of the transform, and now happens in the harvesting process for this table.

src/pudl/transform/eia860.py
```diff
@@ -1023,12 +993,6 @@ def _core_eia860__emissions_control_equipment(
         1000.0 * emce_df["emission_control_equipment_cost"]
     )
 
-    emce_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia860__scd_emissions_control_equipment")
-        .encode(emce_df)
-    )
-
```
zaneselvans, line 1031: Removed the encoding step because it's the last part of the transform, and now happens in the harvesting process for this table.

src/pudl/transform/eia860.py
```diff
@@ -1250,10 +1214,8 @@ def _core_eia860__cooling_equipment(
     ce_df.loc[:, ce_df.columns.str.endswith("_thousand_dollars")] *= 1000
     ce_df.columns = ce_df.columns.str.replace("_thousand_dollars", "")
 
-    resource = pudl.metadata.classes.Package.from_resource_ids().get_resource(
-        "_core_eia860__cooling_equipment"
-    )
-    return ce_df.pipe(apply_pudl_dtypes, group="eia", strict=True).pipe(resource.encode)
+    # Encoding is required here because this table is not yet getting harvested.
+    return apply_pudl_dtypes(ce_df, group="eia", strict=True).pipe(PUDL_PACKAGE.encode)
```
zaneselvans, line 1218: Left the encoding step here because this table does not currently get passed into the harvesting process.

**Question:** Based on the name, this table shouldn't be getting written to the database, but it is, because it's the only version of this data that's available right now.  Is that how we want to do this? It's a little confusing.

src/pudl/transform/eia860m.py
```diff
@@ -81,26 +81,6 @@ def core_eia860m__changelog_generators(
...
-        (eia860m_all.state == "UT")
-        & (eia860m_all.balancing_authority_code_eia == "PA"),
-        "balancing_authority_code_eia",
-    ] = "PACE"
-
-    # Standard encoding of categorical columns.
-    eia860m_all = pudl.metadata.classes.Resource.from_id(
-        "core_eia860m__changelog_generators"
-    ).encode(eia860m_all)
-
```
zaneselvans, line 103: Relocated to the generic `eia860` transform.

src/pudl/transform/eia861.py
```diff
@@ -2261,12 +2259,7 @@ def core_eia861__yearly_reliability(
         # Drop duplicate entries for utilities 13027, 3408 and 9697
         .pipe(_drop_dupes, df_name="Reliability", subset=idx_cols)
         .pipe(_post_process)
-    )
-
-    transformed_r = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia861__yearly_reliability")
-        .encode(transformed_r)
+        .pipe(PUDL_PACKAGE.encode)
```
zaneselvans, line 2262: This is the only table that was getting encoded in the EIA-861.  I **could** put this step in the `_post_process()` function, but there may be columns in the (only partially processed) EIA-861 which match the FK relationship rules, but that haven't been cleaned up enough to encode yet.

src/pudl/transform/eia923.py
```diff
@@ -613,7 +611,7 @@ def gen_fuel_nuclear(gen_fuel_nuke: pd.DataFrame) -> pd.DataFrame:
         "_core_eia923__generation_fuel_nuclear": AssetOut(),
     },
 )
-def _core_eia_923__generation_fuel_eia923(raw_eia923__generation_fuel: pd.DataFrame):
+def _core_eia923__pre_generation_fuel(raw_eia923__generation_fuel: pd.DataFrame):
```
zaneselvans, line 614: Just fixing some weird naming.

src/pudl/transform/eia923.py
```diff
@@ -678,13 +676,7 @@ def _core_eia_923__generation_fuel_eia923(raw_eia923__generation_fuel: pd.DataFr
...
 
     gen_fuel = _clean_gen_fuel_energy_sources(gen_fuel)
-
-    gen_fuel = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia923__monthly_generation_fuel")
-        .encode(gen_fuel)
-    )
-
+    gen_fuel = PUDL_PACKAGE.encode(gen_fuel)
```
zaneselvans, line 679: Left this here because of the labeling step below.

src/pudl/transform/eia923.py
```diff
@@ -882,11 +874,7 @@ def _core_eia923__boiler_fuel(raw_eia923__boiler_fuel: pd.DataFrame) -> pd.DataF
 
     bf_df = remove_duplicate_pks_boiler_fuel_eia923(bf_df)
 
-    bf_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia923__monthly_boiler_fuel")
-        .encode(bf_df)
-    )
+    bf_df = PUDL_PACKAGE.encode(bf_df)
```
zaneselvans, line 877: Left this here because of the labeling step below.

src/pudl/transform/eia923.py
```diff
@@ -1010,12 +996,6 @@ def _core_eia923__generation(raw_eia923__generator: pd.DataFrame) -> pd.DataFram
     dupes = gen_df[gen_df.duplicated(subset=unique_subset, keep=False)]
     gen_df = gen_df.drop(dupes.net_generation_mwh.isna().index)
 
-    gen_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia923__monthly_generation")
-        .encode(gen_df)
-    )
-
```
zaneselvans, line 1018: Removed because it's done in the harvesting step now.

src/pudl/transform/eia923.py
```diff
@@ -1091,12 +1071,6 @@ def _core_eia923__coalmine(
     # then make the id index a column for simpler transferability
     cmi_df = cmi_df.reset_index()
 
-    cmi_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia923__entity_coalmine")
-        .encode(cmi_df)
-    )
-
```
zaneselvans, line 1099: Removed because it's done in the harvesting step now.

src/pudl/transform/eia923.py
```diff
@@ -1217,11 +1191,7 @@ def _core_eia923__fuel_receipts_costs(
             unmapped=pd.NA,
         )
     )
-    frc_df = (
-        pudl.metadata.classes.Package.from_resource_ids()
-        .get_resource("core_eia923__monthly_fuel_receipts_costs")
-        .encode(frc_df)
-    )
+    frc_df = PUDL_PACKAGE.encode(frc_df)
```
zaneselvans, line 1194: Left this here because of the labeling step below.

src/pudl/transform/eia923.py
```diff
@@ -1302,15 +1272,9 @@ def _core_eia923__cooling_system_information(
...
-        "_core_eia923__cooling_system_information"
-    )
-
     primary_key = ["plant_id_eia", "report_date", "cooling_id_eia"]
-    return (
-        pudl.helpers.dedupe_and_drop_nas(csi_df, primary_key_cols=primary_key)
-        .pipe(apply_pudl_dtypes, group="eia", strict=False)
-        .pipe(resource.encode)
+    return pudl.helpers.dedupe_and_drop_nas(csi_df, primary_key_cols=primary_key).pipe(
+        apply_pudl_dtypes, group="eia", strict=False
```
zaneselvans, line 1277: Removed encoding because it happens in Harvesting step now.

src/pudl/transform/ferc714.py
```diff
@@ -321,7 +321,7 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
     Returns:
         The post-processed dataframe.
     """
-    return Package.from_resource_ids().get_resource(table_name).enforce_schema(df)
+    return PUDL_PACKAGE.get_resource(table_name).enforce_schema(df)
```
zaneselvans, line 324: Could add an encoding step here and have it apply to all FERC-714 tables. I don't think there are currently any encoded values though.

src/pudl/transform/ferc1.py
```diff
@@ -1969,11 +1970,7 @@ def transform_main(self, df: pd.DataFrame) -> pd.DataFrame:
             .pipe(self.nullify_outliers)
             .pipe(self.replace_with_na)
             .pipe(self.drop_invalid_rows)
-            .pipe(
-                pudl.metadata.classes.Package.from_resource_ids()
-                .get_resource(self.table_id.value)
-                .encode
-            )
+            .pipe(PUDL_PACKAGE.encode)
```
zaneselvans, line 1973: So encoding is applied to all FERC 1 tables already.

test/unit/io_managers_test.py
```diff
@@ -283,8 +284,7 @@ def test_migrations_match_metadata(tmp_path, monkeypatch):
     # run all the migrations on a fresh DB at tmp_path/pudl.sqlite
     alembic.config.main(["upgrade", "head"])
 
-    pkg = Package.from_resource_ids()
-    PudlSQLiteIOManager(base_dir=tmp_path, db_name="pudl", package=pkg)
+    PudlSQLiteIOManager(base_dir=tmp_path, db_name="pudl", package=PUDL_PACKAGE)
```
zaneselvans, line 287: I ultimately didn't try adding the encoding step to the IO Manager, because I think there are probably some EIA-861 tables with names that match the encoding tables FKs, but which aren't ready to be encoded yet, and because I saw there was already an encoding step happening in the FERC transforms.

test/unit/metadata_test.py
```diff
@@ -59,6 +62,14 @@ def test_pyarrow_schemas(resource_name: str):
...
 
 
+@pytest.mark.parametrize("encoder_name", sorted(PUDL_ENCODERS.keys()))
+def test_encoders(encoder_name: SnakeCase):
+    """Verify that Encoders work on the kinds of values they're supposed to."""
+    encoder = PUDL_ENCODERS[encoder_name]
+    test_data = encoder.generate_encodable_data(size=100)
+    _ = encoder.encode(test_data)
+
+
```
zaneselvans, line 72: This adds 112 test cases, which take 0.2 seconds to run.

src/pudl/transform/eia.py
```diff
@@ -1029,12 +1019,17 @@ def core_eia860__assn_boiler_generator(context, **clean_dfs) -> pd.DataFrame:
...
+    bga_out = (
+        pd.merge(
+            left=bga_out,
+            right=bga_w_units,
+            how="left",
+            on=["plant_id_eia", "generator_id", "boiler_id"],
+        )
+        .astype({"unit_id_pudl": pd.Int64Dtype()})
+        .pipe(apply_pudl_dtypes, group="eia")
+        .pipe(PUDL_PACKAGE.encode)
```
zaneselvans, line 1031: I think the BGA table was previously getting encoded by one of those calls inside the harvesting process, and it doesn't get passed in in the same way as all the other tables, so we ended up with some FK constraint errors, and so I'm adding it in here.

src/pudl/metadata/classes.py
cmgosnell: that sounds very handy overall! and it seems like this code should live over in the unit test.

src/pudl/metadata/classes.py
cmgosnell: maybe add this as a comment so someone later doesn't try to move it and get all janked up with imports

src/pudl/transform/eia.py
cmgosnell: yes that makes sense. the bga asset is entirely separate from harvesting or the _core -> core assets

src/pudl/transform/eia.py
cmgosnell: inside of the harvesting process you removed both the encoding and the `apply_pudl_dtypes`. did you intend to add `apply_pudl_dtypes` here?

src/pudl/transform/eia860.py
cmgosnell: ah okay. so this was previously happening for 860m transform only because this code is not in the gens resource and therefor was not previously getting encoded. cool!

the first line of the comment does not make sense in this different context (the deduplication is an 860m only thing) and should be removed.

src/pudl/transform/eia860.py
cmgosnell: hm well but is it at all helpful for the eia bga/unit_id_pudl assignment?

src/pudl/transform/eia860.py
cmgosnell: same question as above (re the `eia.core_eia860__assn_boiler_generator`

src/pudl/transform/eia860.py
```diff
@@ -376,11 +381,7 @@ def _core_eia860__generators_solar(
         .pipe(pudl.helpers.fix_boolean_columns, boolean_columns_to_fix)
         .pipe(pudl.helpers.month_year_to_date)
         .pipe(pudl.helpers.convert_to_date)
-        .pipe(
-            pudl.metadata.classes.Package.from_resource_ids()
-            .get_resource("core_eia860__scd_generators")
-            .encode
-        )
+        .pipe(PUDL_PACKAGE.encode)
```
cmgosnell, line 384: why did you leave in all of the `__generators_*` encoding steps? all of them are both going through harvesting and getting encoded in the _core -> core asset factory. Is it for the `operational_status` column? this can totally be oos for this pr but to me it would make way more sense to assign that column post harvesting.

src/pudl/transform/ferc1.py
cmgosnell: this is an instance where i personally think it is fine to use the resource specific encoder bc we definitely know the resource name. but i don't feel very strongly about that so this is fine

cmgosnell (changes requested):
one docs nit & one non-blocking request to move testing code to testing module + some non-blocking comments.

src/pudl/metadata/classes.py
zaneselvans: I could turn it into a function that takes an Encoder in the test modules, but unfortunately then it can't be imported for use in development or to mess around in a notebook.

src/pudl/transform/eia.py
zaneselvans: I think I removed it because the tables are immediately written to the DB after harvesting and `enforce_schema` applies dtypes. Actually looking back there now I see that `enforce_schema()` is called within the harvesting function, which might even be superfluous, given that the dataframes are written to the database pretty much immediately after being returned I think?

src/pudl/transform/eia860.py
zaneselvans: Yes, right now the `operational_status` assignment below depends on having a clean version of the operational status code, and I didn't want to deal with that level of reorganization.

src/pudl/transform/eia860.py
zaneselvans: Ah, maybe this is why I had to add an explicit encoding to the BGA function?

src/pudl/transform/eia860.py
zaneselvans: But it's not harvested and normalized like the other `core_` tables, so that also seems a little bit off.

src/pudl/transform/eia860.py
zaneselvans: How are you identifying this as something that might need to be independently encoded?

src/pudl/transform/eia.py
cmgosnell: yes good point!

src/pudl/transform/eia.py
cmgosnell: remove this from the bottom and add it the the `clean_dfs` at the top of this function

src/pudl/transform/eia860.py
cmgosnell: yes. okay i added a comment over in the transfrom.eia module about this so i'm resolving this thread

cmgosnell (approved):

zaneselvans:
Okay, I ran `make pytest-coverage` locally and everything passed soooo, I think this is good to merge.

PR #3561, opened 2024-04-13
Update conda lockfile and other versions.

zaneselvans:


src/pudl/metadata/classes.py
```diff
@@ -68,7 +68,7 @@
 # to define an inconvenient alias for it.
 warnings.filterwarnings(
     action="ignore",
-    message='Field name "schema" shadows an attribute in parent "PudlMeta"',
+    message='Field name "schema" in "Resource" shadows an attribute in parent "PudlMeta"',
```
zaneselvans, line 71: Warning got more detailed in Pydantic 2.7 so had to update this message.

pudlbot (approved):

PR #3570, opened 2024-04-15
Transform NREL ATB

### LINKED ISSUE #3507: Create a transform that reshapes the NREL ATB data into a well normalized core table.
```[tasklist]
### Tasks
- [x] Remove duplicate records in 2021
- [x] Standardize nulls: nulls in their PKs are actually generally `*`'s meaning they pertain to all (these mostly got wiped out during the table unstacking
- [x] figure out what the first letter of the `core_metric_key` is (currently being called `mystery_code`): de-scoped see #3576
- [x] clean up `core_metric_parameter` (which will end up being the column names)
- [x] Pivot table to include one column per `core_metric_parameter` -> three separate tbales
- [x] standardize any `core_metric_parameter` that has a difference schema over the years -> `broadcast_fcr_param_across_tech_detail`
- [x] build a lil extra table(s) with the non-IDX/non-value columns from the raw data
- [x] clean up codes and types and booleans
- [x] confirm units and add them directly into the column names. maybe add a defensive check from the `units` colunm
- [x] convert draft tables from `_core` -> `core` and add db schema
```

## Known data quirks:
* The data here is relatively generic and best used for forward-looking generic resources, but could also be used to fill data gaps.
* Both CSVs and Parquet files in the S3 bucket contain summary data. The detailed data and assumptions used in the calculations are published in a large Excel workbook with many tabs and formulas.
* Technical limitations are documented by NREL [here](https://atb.nrel.gov/electricity/2023/technical_limitations).
* Data is easily accessible via Parquet in a cloud bucket, but it resembles a lookup table more than a database table.
* “Core_metric_key” field has changed over time and is blank for some records in 2022 and earlier records. It is also not a unique identifier (due to the detailed technology column).
* are used as NAs in some columns.

cmgosnell:
# Overview

Closes #3507.

(almost) adds normalized NREL ATB tables.

What did you change?
tracked in the issue tasklist

# Remaining Questions/Next steps
- [x] table name help plz! the smallest table rn is just three different rates so it feels easy to name it something about rates. but the other two have much more/varied info in them so I named them based on the additional unique PK... but that doesn't feel right
- [x] is `unstacker_of_tables` okay (both the name and the thing it does)? it feels hacky to have it just be a place that compiles the unstacking params for all of the tables but it felt very easy to use
- [x] I haven't yet made these assets into an asset factory bc i assume there is going to be unique cleaning to be done even though rn they are all

# Testing

```[tasklist]
# To-do list
```


src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,300 @@
...
+        params_all = (
+            self.rate_table.params
+            + self.scenario_table.params
+            + self.tech_detail_table.params
+        )
+        assert not {x for x in params_all if params_all.count(x) > 1}
+        return params_all
+
+
+def unstacker_of_tables() -> Unstacker:
```
cmgosnell, line 75: this is a compilation of parameters for how to unstack all of the three tables we want to break out from the raw rnel table. It feels kinda silly to make this into a function that just compiles parameters and use it for each of the individual tables, but it is helpful to run a few checks with the all three together (both pydantic validations and checks w/ the  `_core_nrelatb__transform_start` data).

this could be a global variable but I find it pretty frustrating ergonomically when trying to tweak parameters that get loaded & validated on import.

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,300 @@
...
+        """
+        return [c for c in self.idx if c != "core_metric_parameter"]
+
+
+class Unstacker(BaseModel):
+    """Class that defines how to unstack the :func:`transform_start` table into all of the tidy NREL tables."""
+
+    rate_table: TableUnstacker
+    scenario_table: TableUnstacker
+    tech_detail_table: TableUnstacker
```
cmgosnell, line 57: these are 1000% draft names...

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,300 @@
...
+        )
+    return nrelatb_unstacked
+
+
+@asset
+def _core_nrelatb__yearly_rates(_core_nrelatb__transform_start) -> pd.DataFrame:
+    """Transform the yearly rates table.
+
+    Right now, this just unstacks the table.
+    """
```
cmgosnell, line 214: right now these three normalized assets (this one and the two below) all do the same thing w/ different unstacker parameters. i almost made an asset factory but i assume there will be more bespoke cleaning.

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,300 @@
...
+            level="core_metric_parameter",
+        )
+    )
+    nrelatb_unstacked.columns = nrelatb_unstacked.columns.droplevel()
+    # this next step regarding the fcr param is only applicable to
+    # the table w/ the fcr param:
+    if "fcr" in nrelatb_unstacked:
+        nrelatb_unstacked = broadcast_fcr_param_across_tech_detail(
+            nrelatb_unstacked, table_unstacker.idx_unstacked
+        )
```
cmgosnell, line 205: this could be done in `_core_nrelatb__yearly_technology_detail` bc it only is done on that table but it felt more reasonable to put it here bc it is a required second-step for unstacking

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,300 @@
...
+        tech_detail_table=unstacker_detail_table,
+    )
+    return unstack_tables
+
+
+@asset
+def _core_nrelatb__transform_start(raw_nrelatb__data):
+    """Transform raw NREL ATB data into semi."""
+    ## EARLY CLEANING/DEDUPING
+    raw_nrelatb__data["mystery_code"] = raw_nrelatb__data.core_metric_key.str[0]
```
cmgosnell, line 146: `mystery_code`... still is a mystery to me. we need to figure out what this code means. it is either `M` or `R` but that does not seem to correspond to Market/R&D in the `core_metric_case` column.

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,300 @@
...
+            "cost_recovery_period_years",
+            "technology_description",
+            "core_metric_parameter",
+        ],
+        params=[
+            "inflation rate",
+            "interest during construction - nominal",
+            "tax rate (federal and state)",
+        ],
+    )
```
cmgosnell, line 95: i figured out which primary keys + parameters would result in the least sparse tables by trial and error in a notebook. i could add more how i did that and at least one more defensive check in `transform_unstack` about the nulls in the output columns.

src/pudl/transform/nrelatb.py
cmgosnell: i made this moot by removed two of the oldest years of data #3576

src/pudl/transform/nrelatb.py
cmgosnell: removed this in favor of storing this info in `Unstacker` per @jdangerx 's suggestion

src/pudl/transform/nrelatb.py
cmgosnell: this is how i've been swapping params:

```python
for tech_varying_param in Unstacker().tech_detail_table.params:
    try:
        transform_unstack(
            _core_nrelatb__transform_start,
            TableUnstacker(
                idx=Unstacker().scenario_table.idx,
                params=Unstacker().scenario_table.params + [tech_varying_param]
            )
        )
        print(f"WHOAAA you can add {tech_varying_param}")
    except ValueError:
        print(f"don't add {tech_varying_param}")
```

src/pudl/helpers.py
```diff
@@ -1792,7 +1792,7 @@ def fix_boolean_columns(
     """
     fillna_cols = {col: pd.NA for col in boolean_columns_to_fix}
     boolean_replace_cols = {
-        col: {"Y": True, "N": False, "X": False, "U": pd.NA}
+        col: {"Y": True, "N": False, "X": False, "U": pd.NA, 1: True, 0: False}
```
jdangerx, line 1795: Huh! First time we've had to replace 0's and 1's with their boolean counterparts 🎉 

I'm slightly surprised that pandas doesn't coerce these automatically - one of my favorite Weird Python Things is that `True` and `False` are literally the integers `1` and `0`... but I guess that makes sense.

```
>>> isinstance(True, int)
True
>>> True + True
2
>>> (True + True) * 5
10
>>> 1 == True
True
>>> 0 == False
True
```

src/pudl/metadata/sources.py
```diff
@@ -753,7 +753,9 @@
             "source_format": "Parquet",
         },
         "working_partitions": {
-            "years": sorted(set(range(2019, 2024))),
+            "years": sorted(
```
jdangerx, line 756: Any reason to not use `list(range(2021, 2024))`? Or just `[2021, 2022, 2023]`?

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+class TableNormalizer(BaseModel):
+    """Info needed to convert a selection of a the raw NREL table into a normalized table."""
+
+    idx: list[str]
+    """Primary key columns of normalized subset table."""
+    columns: list[str]
+    """Columns reported in raw NREL table that are unique to the :attr:`idx`."""
+
+
+def transform_normlize(nrelatb: pd.DataFrame, normalizer: TableNormalizer):
```
jdangerx, line 37:
```suggestion
def transform_normalize(nrelatb: pd.DataFrame, normalizer: TableNormalizer):
```

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+    "core_metric_parameter",
+]
+"""Expected primary key columns for the raw nrelatb data.
+
+The normalized core tables we are trying to build will result in a subset of these columns.
+"""
+
+
+class TableNormalizer(BaseModel):
+    """Info needed to convert a selection of a the raw NREL table into a normalized table."""
```
jdangerx, line 29: Now that thousands of GSoD applicants are looking at our every line of code, I'm gonna be nitpicky about typos 😉 

```suggestion
    """Info needed to convert a selection of the raw NREL table into a normalized table."""
```

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+
+def transform_normlize(nrelatb: pd.DataFrame, normalizer: TableNormalizer):
+    """Normalize a subset of the NREL ATB data into a small tables."""
+    smol_table = nrelatb[normalizer.idx + normalizer.columns].drop_duplicates()
+    assert smol_table[smol_table.duplicated(normalizer.idx, keep=False)].empty
+    return smol_table.dropna(how="all", subset=normalizer.columns)
+
+
+class Normalizer(BaseModel):
+    """Class that defines how to normalize all of the NREL tables that get the :func:`transform_normlize` treatment."""
```
jdangerx, line 45:
```suggestion
    """Class that defines how to normalize all of the NREL tables that get the :func:`transform_normalize` treatment."""
```

src/pudl/transform/nrelatb.py
jdangerx: Separately from the typo, it's probably useful to describe somewhere the funky dissection you're doing where certain data columns correspond with certain index columns and certain `metric`s...

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+    """Info needed to convert a selection of a the raw NREL table into a normalized table."""
+
+    idx: list[str]
+    """Primary key columns of normalized subset table."""
+    columns: list[str]
+    """Columns reported in raw NREL table that are unique to the :attr:`idx`."""
+
+
+def transform_normlize(nrelatb: pd.DataFrame, normalizer: TableNormalizer):
+    """Normalize a subset of the NREL ATB data into a small tables."""
```
jdangerx, line 38:
```suggestion
    """Normalize a subset of the NREL ATB data into a small table."""
```

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+        nrelatb[param_mask]
+        .set_index(table_unstacker.idx)
+        .sort_index()[["value"]]
+        .drop_duplicates()
+        .unstack(level="core_metric_parameter")
+    )
+    nrelatb_unstacked.columns = nrelatb_unstacked.columns.droplevel()
+    # this next step regarding the fcr param is only applicable to
+    # the table w/ the fcr param:
+    if "fcr" in nrelatb_unstacked:
```
jdangerx, line 106: Instead of having a special case within this general function, it might make sense to only call this function in the scenario table. Then we're not checking "is this the scenario table?" for all the other tables - and we also make the scenario table logic a little more explicit!


```python
def transform_unstack(
    nrelatb: pd.DataFrame, table_unstacker: TableUnstacker
) -> pd.DataFrame:
    """Generic unstacking function to convert ATB data from skinny to wider."""
    param_mask = nrelatb.core_metric_parameter.isin(table_unstacker.params)
    # assert not nrelatb[param_mask].duplicated(subset=table_unstacker.idx).any()

    nrelatb_unstacked = (
        nrelatb[param_mask]
        .set_index(table_unstacker.idx)
        .sort_index()[["value"]]
        .drop_duplicates()
        .unstack(level="core_metric_parameter")
    )
    nrelatb_unstacked.columns = nrelatb_unstacked.columns.droplevel()
    return nrelatb_unstacked


@asset
def _core_nrelatb__yearly_projections_by_scenario(
    _core_nrelatb__transform_start,
) -> pd.DataFrame:
    """Transform the yearly NREL ATB scenario table.

    Right now, this just unstacks the table.
    """
    unstacker = Unstacker()
    df = (
        transform_unstack(_core_nrelatb__transform_start, unstacker.scenario_table)
        .pipe(broadcast_fcr_param_across_tech_detail, unstacker.idx_unstacked)
    )
    return df

```

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+
+def broadcast_fcr_param_across_tech_detail(
+    nrelatb_unstacked: pd.DataFrame, idx_unstacked: list[str]
+) -> pd.DataFrame:
+    """For older years, broadcast the FCR parameter across the technical detail columns.
+
+    We want to table schema to be consistent for all years of ATB data. Mostly the parameters
+    have the same primary keys across all of the years. But the ``fcr`` parameter is the
+    only exception. For the older years (pre-2023), the FCR parameter is not variable
+    based on tech detail so we are going to broadcast the pre-2023 fcr values acorss the
```
jdangerx, line 193:
```suggestion
    based on tech detail so we are going to broadcast the pre-2023 fcr values across the
```

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+        # bc these records are not sparse, we do not want to broadcast these records
+        # we use the columns of nrelatb_unstacked bc those of the params names
+        & (
+            nrelatb_unstacked_fcr[[c for c in nrelatb_unstacked if c != "fcr"]]
+            .isna()
+            .all(axis=1)
+        )
+    )
+    # there are a small number of records that have "Nuclear" tech detail col
+    # we are going to drop this detail bc it is always the same as the tech description
```
jdangerx, line 210: Maybe I'm missing something... where does this happen?

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+    based on tech detail so we are going to broadcast the pre-2023 fcr values acorss the
+    tech details that exist in the data.
+    """
+    nrelatb_unstacked_fcr = nrelatb_unstacked.reset_index()
+    mask_fcr = (
+        nrelatb_unstacked_fcr.fcr.notnull()
+        & (nrelatb_unstacked_fcr.report_year < 2023)
+        # there are weirdly some fcr nuclear records that have a bunch of data
+        # bc these records are not sparse, we do not want to broadcast these records
+        # we use the columns of nrelatb_unstacked bc those of the params names
```
jdangerx, line 202: I know this is still WIP, but just a flag here to not forget to complete this thought!

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+        raw_nrelatb__data.replace(["*", ""], pd.NA)
+        .pipe(
+            helpers.fix_boolean_columns,
+            boolean_columns_to_fix=list(raw_nrelatb__data.filter(regex=r"^is_")),
+        )
+        ## clean & rename values in core_metric_parameters
+        .pipe(helpers.simplify_strings, ["core_metric_parameter"])
+        .pipe(helpers.cleanstrings_snake, cols=["core_metric_parameter"])
+        .replace({"core_metric_parameter": core_metric_parameters_rename})
+        # we are droping dropping records that are completely dupes - not just
```
jdangerx, line 285:
```suggestion
        # we are dropping records that are completely dupes - not just
```

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+        _core_nrelatb__transform_start, Unstacker().tech_detail_table
+    )
+    return df
+
+
+@asset
+def _core_nrelatb__yearly_revisions(
+    _core_nrelatb__transform_start: pd.DataFrame,
+) -> pd.DataFrame:
+    """Transform small table with the r."""
```
jdangerx, line 347: Another flag for "don't forget to finish this thought!"

jdangerx (commented):
This seems like it's going in a good direction! Slight structural suggestion about the special case for the FCR broadcast. Also I flagged a bunch of small typos, inspired by our recent flurry of typo fix PRs.

I also thought about whether `TableNormalizer` and `TableUnstacker` should be the same type, since they are very similar.

I think they are *conceptually* different even if the _implementation_ is currently very similar. So I think it's a case of "two things that *are* actually different" and we should keep them separate. Might be good to write a little bit more in the documentation about why we have these two different approaches to extracting small subsets of data out of the ATB.

src/pudl/transform/nrelatb.py
cmgosnell: lol ty!

src/pudl/transform/nrelatb.py
cmgosnell: i was debating doing this and i agree i do think it makes more sense to do it in the specific table's transform

src/pudl/transform/nrelatb.py
cmgosnell: lol yes ty!!

src/pudl/transform/nrelatb.py
cmgosnell: ah lol good catch! i was dropping them before adding the `.isna().all(axis=1)` portion of the `mask_frc` above. So this is no longer happening so i can remove this big assert.

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,368 @@
...
+            self.rate_table.params
+            + self.scenario_table.params
+            + self.tech_detail_table.params
+        )
+        assert not {x for x in params_all if params_all.count(x) > 1}
+        return params_all
+
+
+def broadcast_fcr_param_across_tech_detail(
+    nrelatb_unstacked: pd.DataFrame, idx_unstacked: list[str]
```
cmgosnell, line 186: now that this is only being used in `_core_nrelatb__yearly_projections_by_scenario`, i'd like to move this down in the module. but i'll keep it here for easier viewing of diffs/following comments for now.

src/pudl/metadata/sources.py
cmgosnell: I believe either of those would work! But I was just repeating the pattern for a range of years across the other sources in here.

I personally like the `range` because it makes sure you don't skip years. it feels a lil silly for three years but i think it is good practice. when i was messing around in a notebook and I was trying the 2019-2023 range i accedenitally skipped 2020.

src/pudl/helpers.py
cmgosnell: oh you are totally right here. I'll remove this. I forgot that I haven't made the schema for these tables yet and therefor can't yet apply the schema dtypes and so of course I am not yet getting the proper bool outputs.

src/pudl/metadata/sources.py
jdangerx: Yeah, I can see that - I think `list(range(...))` seems more concise / least surprising - `sorted(set(range(...)))` says "make this range, throw away the order, then re-order it" vs. "make this range, and turn it into a list."

cmgosnell: that makes a lo of sense! i changed this one. i am a lil tempted to just go change em all but i will refrain

jdangerx: Maybe as a follow-up PR 😉

src/pudl/transform/nrelatb.py
cmgosnell: okay i think i've added this context into the appropriate class/function docs.

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,415 @@
...
+    There are several columns in the raw ATB data that are not a part of the primary keys
+    for the tables that get the :func:`transform_unstack` treatment. This class helps us
+    build these smaller tables which have a smaller subset of primary key columns.
+    """
+
+    # There is info about revisions in the old data, but not in the 2021 -> current data
+    # revisions: TableNormalizer = TableNormalizer(
+    #     idx=["report_year"],
+    #     columns=["revision_num", "update_date"],
+    # )
```
cmgosnell, line 71: it probably makes more sense to take this out entirely. maybe i'll add this context into the issue that i made which would enable the older years of data.

src/pudl/metadata/enums.py
```diff
@@ -190,3 +190,28 @@
...
+    "Nuclear",
+    "OffShoreWind",
+    "Pumped Storage Hydropower",
+    "ResPV",
+    "Residential Battery Storage",
+    "Utility-Scale Battery Storage",
+    "Utility-Scale PV-Plus-Battery",
+    "UtilityPV",
+}
+"""NREL ATB technology descriptions."""
```
cmgosnell, line 217: I considered doing either encoding or otherwise renaming of these guys. I don't love that they are a mix of formats.

cmgosnell:
okay i think this is ready for a re-review. my remaining wonderings are:
- [ ] are the tables names okay? the tables ended up being pretty clean/clear about what type of info is in what type of table
- [ ] should i squish the `additional` parameter into the non-additional param? see comment... i think yes.
- [ ] i copied and pasted a lot of the field descriptions from nrel's extensive docs. i think they are descriptive and reasonable

src/pudl/metadata/resources/nrelatb.py
```diff
@@ -0,0 +1,104 @@
...
+                "wacc_nominal",
+                "wacc_real",
+            ],
+        },
+        "sources": ["nrelatb"],
+        "etl_group": "nrelatb",
+        "field_namespace": "nrelatb",
+    },
+    "core_nrelatb__yearly_projections_by_technology_detail": {
+        "description": "Projections of costs and performance for various technologies.",
```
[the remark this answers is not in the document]
zaneselvans: I'd suggest `core_nrelatb__yearly_projected_cost_performance` -- yearly future data about costs.  I think having "projected" is important in differentiating between past and future data.

migrations/versions/3e08dd518f9c_add_my_cool_nrel_atb_tables.py
```diff
@@ -0,0 +1,88 @@
+"""Add my cool NREL ATB tables
```
jdangerx, line 1: 😎

src/pudl/metadata/resources/nrelatb.py
```diff
@@ -0,0 +1,104 @@
...
+from typing import Any
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = {  # }
+    "core_nrelatb__yearly_financial_cases": {
+        "description": "Finanial assumptions for each core_metric_case and technology_description.",
+        "schema": {
+            "fields": [
+                "report_year",
+                "core_metric_case",
+                "core_metric_variable_year",
```
jdangerx, line 12: IIRC you wanted to rename this to `projection_year` also.

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,417 @@
...
+    "cost_recovery_period_years",
+    "scenario_atb",
+    "technology_description",
+    "technology_description_detail_1",
+    "technology_description_detail_2",
+    "core_metric_parameter",
+]
+"""Expected primary key columns for the raw nrelatb data.
+
+The normalized core tables we are trying to build will have primayr keys which
```
jdangerx, line 24:
```suggestion
The normalized core tables we are trying to build will have primary keys which
```

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,417 @@
...
+
+
+def transform_normalize(nrelatb: pd.DataFrame, normalizer: TableNormalizer):
+    """Normalize a subset of the NREL ATB data into a small table.
+
+    Given a :class:`TableNormalizer` with a set of primary keys (``idx``) and a
+    list of columns (``columns``), build a table with just those primary key and
+    data columns from the larger ATB semi-processed table (output of
+    :func:`_core_nrelatb__transform_start`). Ensure that the output table is
+    unquie absed on the primary keys.
```
jdangerx, line 45:
```suggestion
    unique based on the primary keys.
```
I just installed [this spell checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker). Seems like there's a bunch of false positives for things like "zenodo," but it's relatively easy to add stuff to the user dictionary 🤷 idk, might be useful?

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,417 @@
...
+
+
+def transform_unstack(
+    nrelatb: pd.DataFrame, table_unstacker: TableUnstacker
+) -> pd.DataFrame:
+    """Generic unstacking function to convert ATB data from a skinny to wider format.
+
+    This function applies :func:`pandas.unstack` to a subset of values
+    for ``core_metric_parameter`` (via :attr:`TableUnstacker.core_metric_parameters`)
+    with differnt primary keys (via :attr:`TableUnstacker.idx`). If the set of given
```
jdangerx, line 117:
```suggestion
    with different primary keys (via :attr:`TableUnstacker.idx`). If the set of given
```

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,417 @@
...
+        # this only applies to technology_description's Coal_Retrofits & NaturalGas_Retrofits (maybe we should combine)
+        "additional_occ": "capex_overnight_additional_per_kw",
+        "wacc_nominal": "wacc_nominal",
+        "wacc_real": "wacc_real",
+        "cfc": "capex_construction_finance_factor",
+        "lcoe": "levelized_cost_of_energy_per_mwh",
+        "heat_rate_penalty": "heat_rate_penalty",
+    }
+    nrelatb = (
+        raw_nrelatb__data.replace(["*", ""], pd.NA)
```
jdangerx, line 272: You mentioned that `*` is "wildcard", which is probably different from "no value at all." My understanding of why it's OK to merge these into the same `pd.NA` is:

* `*` appears in index columns when they mean "wildcard" but doesn't appear in data columns
* `""` appears in data columns when they mean "no value" but doesn't appear in index columns

Is that right?

src/pudl/transform/nrelatb.py
```diff
@@ -0,0 +1,417 @@
...
+    )
+    return df
+
+
+def broadcast_fixed_charge_rate_across_tech_detail(
+    nrelatb_unstacked: pd.DataFrame, idx_unstacked: list[str]
+) -> pd.DataFrame:
+    """For older years, broadcast the ``fixed_charge_rate`` parameter across the technical detail columns.
+
+    We want to table schema to be consistent for all years of ATB data. Mostly the parameters
```
jdangerx, line 336:
```suggestion
    We want the table schema to be consistent for all years of ATB data. Mostly the parameters
```

src/pudl/transform/nrelatb.py
cmgosnell: omigosh i've needed this for so long great suggestion i cannot spell v well

src/pudl/transform/nrelatb.py
cmgosnell: yes this is exactly right. but i appreciate you for commenting about this bc it made me go see if I could remove the * from this null-ification. because conceptually what is happening with mostly of the *'s in the index columns is that those columns are being dropped for the records that have any *'s in them during the unstacking (i.e. when there is a *'s in the index it **almost** always means that record should be a part of a table which has an idx w/o the column containing *'s). this is true for almost all of the cases except for the some of the `cost_recovery_period_years` `core_nrelatb__yearly_financial_cases_by_scenario` (which rn has some nulls in it so i can't add this column to the pks of the table).

imma go see if i miscategorized a parameter or if i should broadcast these records across the existing `cost_recovery_period_years`.

cmgosnell: okay i did not miscategorize a param (which is to say the values are non-unique in the for these params in the fin cases tbl w/o the `cost_recovery_period_years` in the idk).

so i did a quick broadcasting of the `*` (mirroring what i did for the fcr) and it worked well but it resulted in records which still have a nulls in that column. because some of the records didn't match the idx values (they were `left_only` records in the broadcast merge). i'm sure i could find a defensible way to broadcast values into that column but i am going to lay this down for now for time sake.

src/pudl/metadata/fields.py
```diff
@@ -4443,6 +4508,106 @@
...
+        ),
+    },
+    "fixed_charge_rate": {
+        "type": "number",
+        "description": (
+            "Amount of revenue per dollar of investment required that must be collected annually "
+            "from customers to pay the carrying charges on that investment."
+        ),
+    },
+    "wacc_nominal": {
```
jdangerx, line 4603: clarification: Is the difference between these two rates that `nominal rate = real rate * expected inflation`?

cmgosnell: I believe so!

PR #3578, opened 2024-04-19
Ensure correct dtypes in EIA-930 CSV extract.

zaneselvans:
# Overview

Previously almost all numerical columns in the EIA-930 extraction were appearing as type `object` because in the CSV they show up as quoted strings with commas separating every 3 digits.

This PR ensures that those columns are parsed as numbers, and the datetime columns are parsed as datetimes when the CSVs are first read in, dramatically reducing overall memory usage, and avoiding the need to deal with enormous untyped dataframes in the transform steps.

However, this is kind of a hack, which overrides the entire `load_source()` method for the `CsvExtractor` class, and which really needs to be able to pass different arguments in to `read_csv()` depending on which of the several dataframes is being extracted. What's a better way to do that?

# Testing

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/extract/eia930.py
+++ b/src/pudl/extract/eia930.py
@@ -46,6 +46,41 @@ class Extractor(CsvExtractor):
         """Transforms raw dataframe and rename columns."""
         return df.rename(columns=self.METADATA.get_column_map(page, **partition))
 
+    def load_source(self, page: str, **partition: PartitionSelection) -> pd.DataFrame:
+        """Produce the dataframe object for the given partition.
+
+        Args:
+            page: pudl name for the dataset contents, eg "boiler_generator_assn" or
+                "data"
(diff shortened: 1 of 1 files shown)


e-belfer:
This is absolutely the right and intended place to do this, in my opinion.

zaneselvans:
The problem I'm seeing here is that we potentially need to send different args to`read_csv()` depending on *which* of the (in this case 3) dataframes are being extracted, so just overriding the method at the level of EIA-930 isn't specific enough.  What I've got there now works for the `balance` table, but it doesn't do the right thing for the other two tables.

e-belfer:
Does the `parse_dates` parameter want to be a function of either `filename` or the `**partition`, then?

zaneselvans:
Closing this because it's been pulled into #3579

PR #3582, opened 2024-04-20
Extract Net Summer Electricity Generating Capacity from AEO Table 54

### LINKED ISSUE #3581: Transform high-priority AEO Table 54 data into normalized assets
Table 54 has heterogeneous data with a variety of different units, so we'll have to pull those out into multiple different assets.

The first set of data we want to pull out of Table 54 is all the data that's in gigawatts. That will be two assets, one with electricity sector data and one with end-use sector data.

Electricity sales data is also high-priority.

For each, we need to subset the raw table 54 data, transform it into a table with one column for each variable, and then persist it to the database.

These are the high priority resources we need to create:

```[tasklist]
# Resources
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3596
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3597
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3607
```

jdangerx:
# Overview

Closes #3581.

What did you change?

# Testing

```[tasklist]
# To-do list
```


src/pudl/extract/eiaaeo.py
jdangerx, line 1: Just renaming things to match the schema we decided on.

src/pudl/metadata/enums.py
```diff
@@ -215,3 +266,45 @@
...
+
+
+TECH_DESCRIPTIONS_EIAAEO: list[str] = [
+    "coal",
+    "combined_cycle",
+    "combustion_turbine_diesel",
+    "distributed_generation",
+    "diurnal_storage",
+    "fuel_cells",
+    "natural_gas",
```
jdangerx, line 278: Added `natural_gas`, `petroleum`, and `pumped_storage_other` which show up in the "generation by fuel type". 

I think the dimension for the "generation_by_fuel_type" is still a "tech description" and not a "fuel type", though:

```
    Coal
    Petroleum
    Natural Gas
    Nuclear
    Pumped Storage and Other 13/
    Renewable Sources 14/
    Distributed Generation 6/
 ```
So I'm ok shoving those into the same bucket.

File: src/pudl/metadata/resources/eiaaeo.py
Comment (jdangerx) [inline src/pudl/metadata/resources/eiaaeo.py:1]: Just moving stuff around to promote the resource.

**Review by jdangerx — COMMENTED** [review]
Hopefully I've worked out enough kinks through this process that the other resources are easier to add!

File: src/pudl/transform/eiaaeo.py [hunk 2]
```diff
@@ -0,0 +1,324 @@
...
+
+import numpy as np
+import pandas as pd
+from dagster import asset
+
+from pudl.helpers import apply_pudl_dtypes
+
+
+def __sanitize_string(series: pd.Series) -> pd.Series:
+    return series.str.lower().str.strip().str.replace(r"\W+", "_", regex=True)
```
cmgosnell, line 21: not required to swap or anything but there is a `pudl.helper.cleanstrings_snake` which i does this to a list of columns not just one series.

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+def get_series_info(series_name: pd.Series) -> pd.DataFrame:
```
cmgosnell, line 24: what is a `series_name`? is it like a weird series of a kind of index/pk of whatever table? ah i see this what they have named this column.

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+            0: "variable_fields",
+            1: "region_series",
+            2: "case_series",
+            3: "report_year_series",
+        }
+    )
+    variable_fields = fields.variable_fields.str.split(" : ", expand=True).rename(
+        columns={0: "topic", 1: "subtopic", 2: "variable_name", 3: "dimension"}
+    )
+    return fields.join(variable_fields)
```
cmgosnell, line 45: ow this is total personal preference but i don't love that you can't tell `join` to join on the index like you can with `merge`. its just nice when the how of the squish is really explicit

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+    sanitized = enriched.assign(
+        **{
+            colname: __sanitize_string(enriched[colname])
+            for colname in enriched.columns
+            if colname not in {"projection_year", "value"}
+        }
+    ).pipe(
+        lambda df: df.assign(
+            report_year_series=df.report_year_series.str.replace("aeo", ""),
+            value=pd.to_numeric(df.value.str.replace("- -", "")),
```
cmgosnell, line 95: is "- -" their null?

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+    ).all(), "Series and category must agree on region!"
+    assert (
+        sanitized.case_series == sanitized.model_case_eiaaeo
+    ).all(), "Series and taxonomy must agree on case!"
+
+    # TODO 2024-04-20: one day we'll include report year in the extract step -
+    # at that point we'll have to compare that with the report year extracted
+    # from the series name.
+    return sanitized.drop(columns=["region_series", "case_series"]).rename(
+        columns={"region_category": "region", "report_year_series": "report_year"}
```
cmgosnell, line 110: two questions:
* could you just name the series in `get_category_info` `region` if you are going to just rename this down here?
* is this the emm region? why not name it with its long/pudl name as soon as you get it out of the category col?

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+    )
+
+
+def unstack(df: pd.DataFrame, eventual_pk: list[str]):
+    """Unstack the values by the various variable names provided.
+
+    2024-04-27: NB it's important that `projection_year` is the last level of
+    index, so that when we sort and diff() we are actually subtracting year
+    over year.
```
cmgosnell, line 119: what is NB? also should this be converted into a `Raises:` section since you added the assert in here?

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+    return unstacked
+
+
+@asset(io_manager_key="pudl_io_manager")
+def core_eiaaeo__yearly_projected_generation_in_electric_sector_by_technology(
+    raw_eiaaeo__electric_power_projections_regional,
+):
+    """Projected net summer generation capacity and additions/retirements."""
+
+    def _drop_aggregated_generation_data(df: pd.DataFrame) -> pd.DataFrame:
```
cmgosnell, line 140: i could imagine some of these next few internal functions being generalized a bit and used for other assets given what we talked about in the schema process. but maybe not depending on the other tables! which is to say this looks cool and i'm sure you'll keep an eye towards generalizing when helpful while iterating over the other tables

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+                "generation_for_own_use",
+            }
+        )
+        return df.loc[~(generation_mask & is_aggregated_mask)]
+
+    def _collect_totals(df: pd.DataFrame) -> pd.DataFrame:
+        """Various total columns are named differently - they don't have to be."""
+        return df.assign(
+            dimension=df.dimension.str.replace(r"^total.*$", "total", regex=True)
+        )
```
cmgosnell, line 159: wait are the columns named differently or the values in this dimension column named differently?

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+        """De-accumulate the cumulative columns so we get yearly values.
+
+        Cumulative values are less useful because they require the user to know
+        what year the accumulation starts.
+        """
+        new_columns = {}
+        for decumulate_column in decumulate_columns:
+            cumulative_colname = f"cumulative_{decumulate_column}"
+            column = df[cumulative_colname]
+            new_columns[decumulate_column] = column.diff().fillna(column)
```
cmgosnell, line 171: nice! i really like the accumulation of a dict for the assign.

okay so this assumes/expects that the index is sorted and have that date col as the last col in the mutli-index? this might be personal preference, but I think that assertion in `unstack` might make more sense at the start of this function with a direct reference to the fact that the index is set in `unstack`. I just think without that the assumptions about what the data looks like when it comes into this function feel pretty important and unstated.

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+    ratio_close_additions_to_total, ratio_close_reported_calculated = (
+        check_totals_add_up(unstacked)
+    )
+    assert (
+        0.999 < ratio_close_additions_to_total <= 1.0
+    ), f"planned + unplanned vs. total: {ratio_close_additions_to_total}"
+    assert (
+        0.999 < ratio_close_reported_calculated <= 1.0
+    ), f"reported vs. calculated totals: {ratio_close_reported_calculated}"
+    # Now that the totals add up, we can drop the total_additions col
```
cmgosnell, line 225: light nit but i think it would be cleaner to put these asserts into `check_totals_add_up` unless you really need these two totals for later.

src/pudl/transform/eiaaeo.py
```diff
@@ -0,0 +1,324 @@
...
+                "capacity": "summer_capacity_mw",
+                "unplanned_additions": "summer_capacity_unplanned_additions_mw",
+                "planned_additions": "summer_capacity_planned_additions_mw",
+                "retirements": "summer_capacity_retirements_mw",
+                "generation": "gross_generation_mwh",
+                "region": "electricity_market_module_region_eiaaeo",
+                "dimension": "technology_description_eiaaeo",
+            }
+        )
+        .pipe(apply_pudl_dtypes, group="eiaaeo")
```
cmgosnell, line 261: i believe that the types of the tables will already be applied bc you are using `pudl_io_manager` (truly everyone does this bc its so helpful to make sure dtypes are type-able b4 trying to poop it out into the db but it is redundant bc types get applied during `enforce_schema`)

cmgosnell (changes requested):
overall this looks really good! and it definitely seems like a lot of this setup will be super helpful in subsequent tables. I mostly added clarifying questions and relatively minor nits.

Also, it might be good to try to use the `pudl.transform.classes.convert_units` and/or the `pudl.transform.classes.AbstractTableTransformer.convert_units` for these unit converstions.

src/pudl/extract/eiaaeo.py
```diff
@@ -57,12 +56,12 @@ class AEOSeries(BaseModel):
...
     """Data schema for a raw AEO table."""
 
-    date: Timestamp = Field(coerce=True)
+    projection_year: int = Field(coerce=True)
     value: str = Field(coerce=True)
     units: str = Field(coerce=True)
     series_name: str = Field(coerce=True)
     category_name: str = Field(coerce=True)
-    case: str = Field(coerce=True)
+    model_case_eiaaeo: str = Field(coerce=True)
```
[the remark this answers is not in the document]
jdangerx: I have to name the fields *something* so I might as well name them things that will mean things down the road. Except `series_name` and `category_name` which get dissected and sold for parts pretty much immediately.

src/pudl/metadata/enums.py
```diff
@@ -26,6 +26,8 @@
 https://en.wikipedia.org/wiki/Time_in_Canada#IANA_time_zone_database
 """
 
+AEO_CASES: list[str] = ["reference"]
+
```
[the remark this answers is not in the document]
jdangerx: Ooh yeah merge issue. Good catch1

src/pudl/transform/eiaaeo.py
jdangerx: Yeah, we're combining a bunch of "series" and the series name contains a bunch of useful metadata that they "could" have just put in separate JSON fields but 🤷

src/pudl/transform/eiaaeo.py
jdangerx: Yep!

src/pudl/transform/eiaaeo.py
jdangerx: two answers:
* I compare the region from the category name & the series name to make sure they match - so I wanted to keep that origin suffix
* I found it a little nicer to have One Big Rename that renames everything to the final PUDL name rather than Several Small Renames when I was writing the code 🤷 Is this blocking?

src/pudl/transform/eiaaeo.py
jdangerx: ["Nota bene," Latin for "note well"](https://en.wikipedia.org/wiki/Nota_bene) - IDK why I chose to use this instead of the more normal word "Note," maybe I was spending too much time reading legalese in leases.

src/pudl/transform/eiaaeo.py
jdangerx: Yeah, I figured it would be good to encapsulate them within the function now and then move them out if necessary.

src/pudl/transform/eiaaeo.py
jdangerx: The values in the dimension column are named differently *and* the columns are named differently. So we had "cumulative_unplanned_additions|total_additions" and then "cumulative_planned_additions|total_planned_additions". Which was very silly since those `total_*` dimensions would just have null values for the non-applicable columns. Hence combining them into just `total`.

src/pudl/transform/eiaaeo.py
jdangerx: It was nice to return the totals in development, and IMO "figure out how close the totals are" could be a pretty generic function - if we want to call it in different places we'd need to pass in different thresholds, error messages, etc. So I figured I'd keep the *called* function simpler and put the error handling stuff in the caller. This will end up getting refactored a bit in #3597 since that has some total-summing too - let's discuss best structure there once this is merged!

src/pudl/transform/eiaaeo.py
jdangerx: I believe that too, I can take it out (but it was very helpful during development!)

src/pudl/transform/eiaaeo.py
jdangerx: Yeah, that makes a lot of sense. Moving

src/pudl/transform/eiaaeo.py
cmgosnell: grand grand this makes sense and this is definitely not blocking! i personally like early renames for easier searching but that is 1000% a personal preference.

src/pudl/transform/eiaaeo.py
cmgosnell: lolol 📜

src/pudl/transform/eiaaeo.py
jdangerx: Turns out this broke a bunch of stuff by coercing `pd.NA` to `"none"` as well as not snaking out the `/`s in the various names. So I kept `__sanitize_string` around.

src/pudl/metadata/classes.py
```diff
@@ -1644,9 +1644,9 @@ def enforce_schema(self, df: pd.DataFrame) -> pd.DataFrame:
 
         df = self.format_df(df)
         pk = self.schema.primary_key
-        if pk and not df[df.duplicated(subset=pk)].empty:
+        if pk and not (dupes := df[df.duplicated(subset=pk)]).empty:
             raise ValueError(
-                f"{self.name} Duplicate primary keys when enforcing schema."
+                f"{self.name} {len(dupes)}/{len(df)} duplicate primary keys ({pk=}) when enforcing schema."
```
cmgosnell, line 1649: grand!

cmgosnell (approved):
this looks great!! and I'm excited to see how this evolves with the new round of tables

PR #3584, opened 2024-04-20
EIA-930 initial transform

### LINKED ISSUE #3505: Initial transform of EIA930 tables
Create a basic set of `core_eia930` assets sufficient for the GridLab MVP. 

```[tasklist]
- [x] Switch to using April `eia930` DOI in datastore.
- [x] Add `core_eia930__hourly_balancing_authority_demand` asset
- [x] Add `core_eia930__hourly_balancing_authority_net_generation` asset
- [x] Add `core_eia930__hourly_balancing_authority_interchange` asset
- [x] Add `core_eia930__hourly_subregion_demand` asset
- [x] Add `core_eia930__assn_balancing_authority_subregion`
- [x] Add `core_eia930__assn_balancing_authority_region` asset
- [x] Standardize column names in interchange table column maps
- [x] Define schemas for each of these tables
- [x] Write EIA-930 tables to SQLite (~18min; adds 50% / 8GB to pudl.sqlite)
- [x] If SQLite output too slow/big, try Parquet only (90sec; 260MB)
- [x] DATA: Evaluate consistency of BA codes found in EIA860, EIA923, & EIA930
- [x] DOC: Are `generation`/`demand`/`interchange` really ~MW~? Or MWh?
- [x] DOC: Clarify whether these hourly timestamps are Hour Ending or Hour Starting
- [x] DOC: Clarify meanings of `reported`, `adjusted`, `imputed` and `forecast` demand/interchange/generation
- [x] DOC: Verify that `demand_reported_mw` for BA and subregions actually mean the same thing
- [x] DOC: Flesh out real table descriptions
- [x] ENG: Address access/database consequences of only writing hourly EIA-930 data to Parquet
- [x] ENG: copy the EIA-930 parquet outputs out of `parquet` folder for distribution in nightly builds
- [ ] Figure out if day ahead demand forecast timestamps are when the forecast was made, or the forecast time
```

```[tasklist]
# Out of Scope
- [x] DATA: Remove local timestamps and record the timezone associated with each BA's reporting
- [x] ENG: Define allowed `region_code_eia` values in the BA coding table
- [ ] DATA: [Evaluate discrepancy between reported total & per-fuel generation](https://github.com/catalyst-cooperative/pudl/pull/3584#issuecomment-2067908462) (it is not small!)
- [ ] DATA: [Evaluate consistency of aggregated subregion demand and BA demand](https://github.com/catalyst-cooperative/pudl/pull/3584#issuecomment-2067908462)
- [ ] ENG/DATA: Define additional asset checks to validate the new tables
- [ ] ENG: Harvest BA codes from EIA-860, EIA-861, EIA-923, and EIA-930
- [ ] DATA: Create a coding table for cleaning and documenting BA subregion codes
```

zaneselvans:
# Overview

- Add several core PUDL tables derived from the EIA-930 balancing authority, interchange, and subregion inputs.
- Add or augment BA region & subregion coding / association tables so that information about the BA regions & subregions lives somewhere clear in the DB.
- Remove all hourly tables from the SQLite outputs and (as with the EIA-930 hourly tables) output them only as Parquet files. This shrinks the PUDL DB, speeds up the ETL, and makes the tables more usable.
- Clean up the data source documentation page Jinja templates, and ensure that links to Datasette aren't created for the Parquet-only outputs. 
- Create an EIA-930 data source documentation page

Closes #3505

# Out of Scope
- There are a number of data consistency issues we should investigate, but that will have to come later. See #3505 
- We should probably also define additional asset checks for EIA-930, maybe along the lines of those consistency issues.

# Testing

See tasklist below

```[tasklist]
# To-do list
- [x] Run the full ETL locally
- [x] Run `make pytest-integration-full`
- [x] Add `adjacent_balancing_authority_code_eia` to list of BA FK columns
- [x] Run `make pytest-validate`
- [x] Fix FK constraint violations in `core_eia930__assn_balancing_authority_region`
- [x] Fix NA values for CAN/MEX `region_code_eia`
- [x] Run `make pytest-coverage`
- [x] Update the release notes and reference the PR / related issues.
- [x] Run a branch deployment to make sure we don't have big memory/concurrency issues
- [x] NAME: rename `adjacent_{sub}region_code_eia` to `{sub}region_code_adjacent_eia`
- [x] Use `energy_source` rather than integer index level when reshaping for legibility
- [x] DATA: omit `region_code_eia` from `core_eia930__hourly_balancing_authority_demand`
- [x] NAME/DOC: rename `energy_source` to... `generation_energy_source` and document why it's not coded
- [x] NAME: rename `report_datetime_utc` to `datetime_utc` for uniformity with other hourly tables
- [x] Condense all alembic migrations into a single file before merging
- [x] DATA: verify that pandera schema asset checks require joint uniqueness of PK tuples.
- [x] DOC: Explain `energysource_processingstatus` naming of net generation columns pre-stack
- [x] Verify definitions of various data fields. See #3505
- [x] Flesh out EIA-930 field descriptions
- [x] Compare reported vs. calculated demand visually
- [x] Flesh out EIA-930 operations table description
- [x] Create EIA-930 data source documentation page
- [x] Avoid dead links for Parquet only outputs in the PUDL data dictionary
- [x] Manage EIA-930 parquet outputs in nightly builds
- [x] Move gzip instructions down in nightly build output access page
- [x] rename EIA930_GENERATION_ENERGY_SOURCES to put EIA930 at the end
- [x] Rename `region_eia` to `balancing_authority_region_name_eia`
- [x] and `region_code_eia` to `balancing_authority_region_code_eia`
- [x] Remove TODO by the note that there are big differences in reported & calculated totals.
- [x] fix weird extra space in service territory test docstring
- [x] add defensive check to ensure there are never underscores in generation energy sources.
- [x] In resource descriptions use EIA-930 not "Electric Grid Monitor"
- [x] Clarify description of interchange in BA operations table.
- [x] Restore `core_ferc__codes_accounts` table
- [ ] Clarify meaning of day ahead forecast timestamps
```


zaneselvans:
# Investigating Data Consistency

## Total vs. Aggregated Net Generation

* EIA-930 reports net generation by fuel, and also as a total.
* In theory, it seems like if you add up all the individual fuel specific generation numbers you should get the total.
* Note that interchange doesn't seem to be included in the totals.
* However, there's a fair bit of dispersion in the correlation between the calculated and reported totals. See the **log-log** plot below.
* Most of the dispersion seems to be in the samples representing smaller values, but sometimes they're off by like an order of magnitude.
* Maybe there's a way of combining the `reported`, `adjusted` and `imputed` values to get something closer to the total?  But that doesn't make a lot of sense since I'm comparing the `reported` versions of both the per-fuel and total generation numbers.

```py
total_mask = core_eia930__hourly_balancing_authority_net_generation.energy_source == "total"
netgen_total = (
    core_eia930__hourly_balancing_authority_net_generation[total_mask]
    .drop(columns=["energy_source", "report_datetime_local"])
    .set_index(
        [
            "balancing_authority_code_eia",
            "report_datetime_utc",
        ]
    )
    .sort_index()
)
netgen_byfuel = core_eia930__hourly_balancing_authority_net_generation[
    ~total_mask
    & (core_eia930__hourly_balancing_authority_net_generation.energy_source != "interchange")
]
calculated_totals = (
    netgen_byfuel.groupby(
        [
            "balancing_authority_code_eia",
            "report_datetime_utc",
        ],
        observed=True)
    .agg(
        {
            "net_generation_adjusted_mw": "sum",
            "net_generation_imputed_mw": "sum",
            "net_generation_reported_mw": "sum",
        }
    )
    .sort_index()
)
fig, ax = plt.subplots(figsize=(10, 10))

ax.scatter(
    calculated_totals.net_generation_reported_mw,
    netgen_total.net_generation_reported_mw,
    s=1, alpha=0.05,
)
ax.set_xlim(1, 1e6)
ax.set_ylim(1, 1e6)
ax.loglog()
ax.set_aspect("equal")
ax.grid()
plt.xlabel("Sum of generation by all fuels [MW]")
plt.ylabel("Reported total generation [MW]")
plt.title("Net Generation by Balancing Authority")
plt.show()
```

image

## Reported BA vs. Aggregated Subregion Demand

* The correlation between whole-BA demand, and aggregations of individual subregions under a BA seems better than the net generation above.
* However, note that the demand which is reported by subregion is just much, much smaller (like a factor of 10 or more) than that which is reported by the whole BAs. Not sure if that's just because some BA's don't really have subregions?

```py
aggregated_ba_demand = core_eia930__hourly_subregion_demand.groupby(["report_datetime_utc", "balancing_authority_code_eia"], observed=True).demand_reported_mw.sum()
ba_demand = core_eia930__hourly_balancing_authority_demand.set_index(["report_datetime_utc", "balancing_authority_code_eia"]).demand_reported_mw
shared_index = aggregated_ba_demand.index.intersection(ba_demand.index)

fig, ax = plt.subplots(figsize=(10, 10))

ax.scatter(
    aggregated_ba_demand.loc[shared_index],
    ba_demand.loc[shared_index],
    s=1, alpha=0.05,
)
ax.set_xlim(1e2, 1e6)
ax.set_ylim(1e2, 1e6)
ax.loglog()
ax.set_aspect("equal")
ax.grid()
plt.xlabel("Subregion demand aggregated by BA [MW]")
plt.ylabel("Reported BA Demand [MW]")
plt.title("Demand by Balancing Authority")
plt.show()
```

image

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+
+    # Prepare a multi-index for the columns so that we can stack cleanly
+    netgen_renamed.columns = pd.MultiIndex.from_tuples(
+        [x.split("_") for x in netgen_renamed.columns], names=["energy_source", None]
+    )
+    netgen_stacked = (
+        netgen_renamed.stack(level=0, future_stack=True)
+        .rename(columns=lambda x: f"net_generation_{x}_mw")
+        .reset_index()
+        .astype({"energy_source": "string"})
```
zaneselvans, line 63: `energy_source` has to be explicitly typed because it gets `object` after the `stack()` puts it in the index.

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+        .astype(
+            {
+                "balancing_authority_code_eia": pd.CategoricalDtype(),
+                "energy_source": pd.CategoricalDtype(),
+            }
+        )
```
zaneselvans, line 69: Setting these to be `category` saves lots of memory usage, but this should be happening automatically now for `energy_source` as it's getting written out to Parquet and back in through the IO Manager.  The BA ID has a FK relationship with the BA code table, so not sure that it gets to be a category (it's not ENUM constrained, but FK constrained).

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+    # TODO[zaneselvans] 2024-04-20: Verify that sum of net generation from all fuels
+    # adds up to the total And then drop the total rows.
+    # NOTE: currently there are some BIG differences between the calculated total and
+    # the reported total.
+
+    # netgen_stacked = netgen_stacked[netgen_stacked["energy_source"] != "total"
+
+    demand = raw_eia930__balance[
+        qual_cols + list(raw_eia930__balance.filter(like="demand"))
+    ].astype({"balancing_authority_code_eia": pd.CategoricalDtype()})
```
zaneselvans, line 80: A whole second asset, but much simpler.

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+logger = pudl.logging_helpers.get_logger(__name__)
+
+
+@multi_asset(
+    outs={
+        "core_eia930__hourly_balancing_authority_net_generation": AssetOut(
+            io_manager_key="parquet_io_manager"
+        ),
+        "core_eia930__hourly_balancing_authority_demand": AssetOut(
+            io_manager_key="parquet_io_manager"
```
zaneselvans, line 17: Note that right now, all the hourly EIA-930 data is getting written to Parquet only, because otherwise it takes 20 minutes to write the tables out, and they take up 8GB of space in SQLite, while the Parquet outputs take 90 seconds to write out and 260MB on disk.

However, this means that we've got empty tables inside the DB with the table schema (which is weird) and also that we will need to specifically pull the parquet outputs up out of the `parquet` folder for distribution.

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+                "demand_reported_mw",
+            ],
+        ]
+    )
+    assn = (
+        demand.groupby("balancing_authority_code_eia")["subregion_code_eia"]
+        .unique()
+        .explode()
+        .to_frame()
+        .reset_index()
```
zaneselvans, line 133: This is kind of a lazy harvesting step.  It just grabs all the values and their associations and assumes they're consistent.  I guess we should check that no subregion is part of more than one BA.

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+    )
+
+
+@multi_asset(
+    outs={
+        "core_eia930__hourly_balancing_authority_interchange": AssetOut(
+            io_manager_key="parquet_io_manager"
+        ),
+        "core_eia930__assn_balancing_authority_region": AssetOut(
+            io_manager_key="pudl_io_manager"
```
zaneselvans, line 155: Writing the association tables to the DB is useful because it ends up with a BA code FK relationship, and that identified some new BAs that needed to be added to the coding table (Mexico & Canada BAs)

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+        .adjacent_balancing_authority_code_eia.unique()
+        .explode()
+        .to_frame()
+        .reset_index()
+        .astype(
+            {
+                "adjacent_region_code_eia": pd.CategoricalDtype(),
+                "adjacent_balancing_authority_code_eia": pd.CategoricalDtype(),
+            }
+        )
```
zaneselvans, line 207: As I said above, I think this `astype()` is probably no longer necessary, since we're writing the data out with the IO Manager.

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+        }
+    ).loc[
+        :,
+        [
+            "report_datetime_local",
+            "report_datetime_utc",
+            "balancing_authority_code_eia",
+            "adjacent_balancing_authority_code_eia",
+            "region_code_eia",
+            "adjacent_region_code_eia",
```
zaneselvans, line 192: The region codes are technically superfluous because they're implied by the BA codes, if we wanted a more normalized version of this table.

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+    * region_code_eia and adjacent_region_code_eia are from the same set of values, but
+      adjacent_region_code_eia also includes "CAN" and "MEX" because foreign countries
+      can be adjacent, but are not reporting directly.
+    * similarly adjacent_balancing_authority_and balancing_authority_code come from the
+      same pool of values, but the adjancent_balancing_authority_code_eia includes some
+      codes that hail from Canada and Mexico.
+    * There's an implied set of associations between the regions and the balancing
+      authorities.
+    * Need to check for consistency with BA codes mentioned here and elsewhere in
+      PUDL (plants table, EIA861 BA table).
```
zaneselvans, line 174: Need to remove this notes as they're out of date.

src/pudl/metadata/fields.py
```diff
@@ -1133,6 +1165,26 @@
         ),
         "unit": "MWh",
     },
+    # TODO[zaneselvans] 2024-04-20: Flesh out description
+    "energy_source": {
```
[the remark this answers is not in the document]
zaneselvans: Note also that I shortened some of these to match other categorizations that we have (and also to make this more compact):
- `all_petroleum_products` -> `oil`
- `natural_gas` -> `gas`
- `other_fuel_sources` -> `other`
- `hydro_and_pumped_storage` -> `hydro`
- `unknown_fuel_source` -> `unknown`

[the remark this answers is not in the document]
cmgosnell: not blocking but it might be good to make this an encoded table so you can add any description for any of these guys.

src/pudl/metadata/fields.py
```diff
@@ -3139,6 +3215,31 @@
         "type": "string",
         "description": "Identifier indicating original FERC Form 1 source record. format: {table_name}_{report_year}_{report_prd}_{respondent_id}_{spplmnt_num}_{row_number}. Unique within FERC Form 1 DB tables which are not row-mapped.",
     },
+    # TODO[zaneselvans]: 2024-04-20 research the real definition of this column and
+    # create a coding table that explains it.
+    "region_code_eia": {
```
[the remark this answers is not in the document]
cmgosnell: is there a reasonable assn table(s) that would make this relationship clear? also yea this definition needs a real definition bc i am confused about what this field means

src/pudl/transform/eia930.py
cmgosnell: oh my i didn't really conceive of the fact that not exporting these tables to the db mean we don't get the db checks. that seems like a real problem actually. are these any schema checks done with the `parquet_io_manager`?

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+
+    Extract the net generation information from the balance table and extract fuel type
+    from the net generation columns.
+    """
+    qual_cols = [
+        "report_datetime_local",
+        "report_datetime_utc",
+        "balancing_authority_code_eia",
+        # report_date
+        # eia_region_code,
```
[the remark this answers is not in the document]
cmgosnell: if we are not loosing it entirely (i.e. we can merge it back in later in `out_`) i don't think they should be here

migrations/versions/350144ddf348_remove_storage_enum_constraints.py
cmgosnell, line 1: total nit but i think it is nice to downgrade schema to the last (pre-pr) schema and then wipe all the pr changes and make one migration just so its easier to follow

src/pudl/metadata/enums.py
```diff
@@ -73,6 +73,111 @@
...
+    "NE",
+    "NW",
+    "NY",
+    "SE",
+    "SW",
+    "TEN",
+    "TEX",
+]
+
+EIA_SUBREGIONS: list[str] = [
```
cmgosnell, line 94: sigh why are these (string) numbers and letters?

src/pudl/metadata/fields.py
```diff
@@ -48,6 +50,15 @@
...
     "address_2": {"type": "string", "description": "Second line of the address."},
+    "adjacent_balancing_authority_code_eia": {
+        "type": "string",
+        "description": "EIA short code for the other adjacent balancing authority, with which interchange is occuring. Includes Canadian and Mexican BAs.",
+    },
+    "adjacent_region_code_eia": {
+        "type": "string",
+        "description": "EIA code for the other adjacent region, with which interchange is occuring.",
+        "constraints": {"enum": EIA_REGIONS},
+    },
```
cmgosnell, line 61: nit but i strongly prefer putting a qualifiers to a field name at or towards the end of a field name so the fields are easy to round up together. these would be `balancing_authority_code_adjacent_eia` and `region_code_adjacent_eia` bc we always put source at the very end

src/pudl/metadata/fields.py
```diff
@@ -3704,47 +3798,22 @@
     "storage_enclosure_code": {
         "type": "string",
         "description": "A code representing the enclosure type that best describes where the generator is located.",
-        "constraints": {
-            "enum": set(
-                CODE_METADATA["core_eia__codes_storage_enclosure_types"]["df"].code
-            )
-        },
     },
```
cmgosnell, line 3801: why are all of these enums gone?

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,179 @@
...
+)
+def core_eia930__hourly_balancing_authority_assets(
+    raw_eia930__balance: pd.DataFrame,
+):
+    """Transforms raw_eia923__balance dataframe.
+
+    Extract the net generation information from the balance table and extract fuel type
+    from the net generation columns.
+    """
+    qual_cols = [
```
cmgosnell, line 30: what does `qual` mean?

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,224 @@
...
+    # is much faster.
+    netgen_renamed = netgen.rename(
+        lambda col: col.removeprefix("net_generation_").removesuffix("_mw"),
+        axis="columns",
+    ).set_index(qual_cols)
+
+    # Prepare a multi-index for the columns so that we can stack cleanly
+    netgen_renamed.columns = pd.MultiIndex.from_tuples(
+        [x.split("_") for x in netgen_renamed.columns], names=["energy_source", None]
+    )
```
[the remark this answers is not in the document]
cmgosnell: it would be nice to add a description of the column format somewhere. without that in here this is a little hard to follow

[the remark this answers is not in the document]
cmgosnell: what happens with the interchange from the filtering of columns above in all of this? or does interchange end up as a energy source?

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,179 @@
...
+    netgen_stacked = (
+        netgen_renamed.stack(level=0, future_stack=True)
```
cmgosnell, line 60: can you swap from level 0 to the name? mostly for readability. i assume it is `"energy_source"`?

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,179 @@
...
+        .rename(columns=lambda x: f"net_generation_{x}_mw")
```
cmgosnell, line 61: wait your adding back in the energy source into the column names?! i'm confused in general where this started and where you are trying to take this.... was this all to stack the energy_source but keep the report-type or whatever (imputed/reported etc)? again i think a description of the column structure of the inputs and desired outputs would go along way to making this more readable for me.

src/pudl/transform/eia930.py
cmgosnell: is this necessary bc `enforce_schema`? (i get the urge bc you were probably notebooking about it)

src/pudl/transform/eia930.py
cmgosnell: i think that would be an easy and valuable little defensive check to add right here

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,179 @@
...
+        .explode()
+        .to_frame()
+        .reset_index()
+    )
+    return (
+        Output(value=demand, output_name="core_eia930__hourly_subregion_demand"),
+        Output(
+            value=assn, output_name="core_eia930__assn_balancing_authority_subregion"
+        ),
+    )
```
cmgosnell, line 125: instead of returning two assets here you could just make `core_eia930__hourly_subregion_demand` and then make `core_eia930__assn_balancing_authority_subregion` take `core_eia930__hourly_subregion_demand` as an input. I think that would be slightly more readable.

cmgosnell (commented):
it seems like there a lot of todo's and questions left in here. I can't tell which ones you are suggesting to do later because of resource constraints and which ones you want to address before merging. or which comments were notes for you or you want help figuring out.

but structurally this looks good. In the transform module i would suggest breaking out the multi assets into individual assets unless there is some reason to keep them together.

src/pudl/extract/eia930.py
```diff
@@ -17,6 +20,22 @@ def __init__(self, *args, **kwargs):
...
+            ],
+            "dtype": {
+                "Balancing Authority": "string",
+                "Region": "string",
+                "DIBA_Region": "string",
+                "Directly Interconnected Balancing Authority": "string",
+                "Sub-Region": "string",
+            },
+            "cache_dates": True,
+        }
```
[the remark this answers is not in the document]
zaneselvans: It's a mix. The date parsing columns all overlap, and `read_csv()` will ignore missing columns in the `dtype` argument.

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
zaneselvans: I think we probably want to avoid encoding many millions of rows if we can, and these values are extracted directly from the renamed column headers so we have complete control over what values they end up containing. But agree we need to document this. Probably this should be a less generic field name, with a description that talks specifically about what's going on in the EIA-930.

src/pudl/metadata/fields.py
```diff
@@ -3164,6 +3265,14 @@
...
     "report_date": {"type": "date", "description": "Date reported."},
+    "report_datetime_local": {
+        "type": "datetime",
+        "description": "Local date and time of the report.",
+    },
+    "report_datetime_utc": {
+        "type": "datetime",
+        "description": "UTC date and time of the report.",
+    },
```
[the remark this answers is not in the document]
zaneselvans: Yes, it's all hourly. In CEMS we have `operating_datetime_utc`, which I also don't like. There's an important hour ending vs. hour beginning aspect that we need to communicate which isn't getting across right now. I think all these hourly timestamp columns might need some naming thinking.

src/pudl/transform/eia930.py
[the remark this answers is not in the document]
zaneselvans: Right now we're doing different things in different tables.  If we do want it to be in the table we're presenting to the public, I think we should probably just leave it in here for now since we aren't making any separate output tables (and also having both core and output tables for tens of millions of rows is going to be a lot of not very helpful bloat)

src/pudl/transform/eia930.py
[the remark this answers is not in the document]
zaneselvans: The interchange has to be dealt with separately in the column filtering, but yes it ends up looking like an energy source because it also comes in "reported", "adjusted", and "imputed" flavors.

src/pudl/transform/eia930.py
zaneselvans: All the same schema checks do get done regardless of whether it's written out to Parquet or SQLite. The only thing that I think we'll miss out on is FK validity, but the association tables which are built from the data tables here DO have FK relationships (for the BA codes), and they're being satisfied so I think that's okay.

migrations/versions/350144ddf348_remove_storage_enum_constraints.py
zaneselvans: Yeah I was thinking of doing that.

src/pudl/metadata/enums.py
zaneselvans: Because that's what they reported. The field has to be a string because most of them are strings, but I think these subregion IDs are just whatever tf the ISO felt like using.

src/pudl/metadata/fields.py
zaneselvans: These columns are naturally constrained by virtue of having a FK relationship with the coding tables. The `enum` is duplicative.  There are some other cases where we are using some *other* column that's not part of the FK relationship as a constraint, and those do need to be `enum`.

src/pudl/transform/eia930.py
zaneselvans: Not sure, this was already there.

src/pudl/transform/eia930.py
zaneselvans: Hmm, maybe? I haven't done that before but can give it a try.

src/pudl/transform/eia930.py
zaneselvans: Hmm, thinking about this more, I'm not sure this is valid in general. E.g. you could imagine more than one ISO having a subregion `0001` even if they don't at the moment.

src/pudl/transform/eia930.py
zaneselvans: We could, but they're really both derived from the one input, and creating them at the same time means that we don't have to read the other table back in again after it's been produced.

src/pudl/transform/eia930.py
zaneselvans: Here `x` is `reported`/`imputed`/`adjusted` not the energy source.

src/pudl/extract/eia930.py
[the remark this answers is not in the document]
cmgosnell: can you add this in as a comment

src/pudl/transform/eia930.py
zaneselvans: Verify that the pandera asset checks require joint uniqueness of the PK tuple.

src/pudl/extract/eia930.py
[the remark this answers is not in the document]
zaneselvans: Okay done.

src/pudl/transform/eia930.py
zaneselvans: Removed this typing.

migrations/versions/350144ddf348_remove_storage_enum_constraints.py
zaneselvans: Added this to the PR task list. Will do it at the very end just in case the schemas keep changing.

src/pudl/transform/eia930.py
zaneselvans: Switched to `nondata_cols`

src/pudl/metadata/fields.py
```diff
@@ -1878,6 +1930,12 @@
         "type": "number",
         "unit": "gpm",
     },
+    # TODO[zaneselvans]: 2024-04-20 Research real definition of this column.
+    "interchange_mw": {
+        "type": "number",
+        "description": "Interchange in MW.",
+        "unit": "MW",
+    },
```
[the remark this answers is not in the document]
zaneselvans: Looking at the [EIA-930 data docs](https://www.eia.gov/electricity/gridmonitor/about) it sure seems like these should be MWh:

> BAs report all data as hourly integrated values, meaning they are time-weighted hourly averaged values.

zaneselvans:
Thoughts / decisions while renaming some things and documenting the EIA-930 tables/columns. See [their data docs](https://www.eia.gov/electricity/gridmonitor/about)

## Local Time

The reported local time is complicated. It includes daylight savings time gaps and duplicate hours. Some BAs span multiple timezones. Some BAs have chosen surprising timezones that don't correspond to their physical location. UTC is unambiguous and monotonically increasing, and can be localized intentionally by users based on their analytical needs, so I've removed the local time columns from the core tables, and added a `report_timezone` column to the BA codes table, which can be used for (re-)localization.

## Other BA attributes

While adding the `report_timezone` column to the BA codes table I also added the `region_eia`, `region_code_eia`, and `balancing_authority_retirement_date` columns that EIA had compiled, and removed the `region_code_eia` from the (very long) core tables.

## MW vs. MWh?

In EIA's CSVs, the demand, net generation, and interchange columns all end with `_mw` as if they are power, not energy. But in their description of the quantities they say:

> BAs report all data as hourly integrated values, meaning they are time-weighted hourly averaged values.

And in their own plots displaying these quantities, they use megawatthours:

image

and everywhere else that we have net generation and demand numbers, they're in units of energy (MWh) not power (MW) so I have gone ahead and changed all of the demand / interchange / net generation columns to be MWh.

## Demand Forecasts

EIA says that the demand forecast is a day ahead forecast, but it's not clear to me whether the timestamp associated with the data indicates the time at which the forecast was made, or the time at which the forecast is attempting to predict demand.

## Change to table normalizations

After gaining a better understanding of how net generation, interchange, and demand relate to each other, I went ahead and re-organized the main balancing authority table breakdown into:
* `core_eia930__hourly_balancing_authority_operations` (which contains total net generation, demand, and interchange) and
* `core_eia930__hourly_balancing_authority_net_generation_by_energy_source` (which breaks down the net generation into several categories)

zaneselvans:
# Visual Sanity Checks

Definitely some wonky stuff going on in the original data.

## Total vs. Aggregated Net Generation (all energy sources)
```python
netgen_reported = (
    core_eia930__hourly_balancing_authority_operations
    .set_index(
        [
            "balancing_authority_code_eia",
            "datetime_utc",
        ]
    )
    .loc[:, ["net_generation_reported_mwh"]]
).sort_index()

netgen_calculated = core_eia930__hourly_balancing_authority_net_generation_by_energy_source.groupby(
        [
            "balancing_authority_code_eia",
            "datetime_utc",
        ]
).net_generation_reported_mwh.sum().sort_index().to_frame()

fig, ax = plt.subplots(figsize=(10, 10))
ax.scatter(
    netgen_reported.net_generation_reported_mwh,
    netgen_calculated.net_generation_reported_mwh,
    s=1, alpha=0.05,
)
ax.set_xlim(1, 1e6)
ax.set_ylim(1, 1e6)
ax.loglog()
ax.set_aspect("equal")
plt.xlabel("Reported total generation [MWh]")
plt.ylabel("Sum of generation by all fuels [MWh]")
plt.title("Net Generation by Balancing Authority")
plt.grid()
ticks = np.logspace(0, 6, 13)
plt.xticks(ticks)
plt.yticks(ticks)
plt.show()
```
image

## Reported BA vs. Aggregated Subregion Demand
```python
aggregated_ba_demand = core_eia930__hourly_balancing_authority_subregion_demand.groupby(
    [
        "balancing_authority_code_eia",
        "datetime_utc",
    ],
    observed=True,
).demand_reported_mwh.sum()
reported_ba_demand = core_eia930__hourly_balancing_authority_operations.set_index(
    [
        "balancing_authority_code_eia",
        "datetime_utc",
    ]
).demand_reported_mwh
shared_index = aggregated_ba_demand.index.intersection(reported_ba_demand.index)

fig, ax = plt.subplots(figsize=(10, 10))
ax.scatter(
    reported_ba_demand.loc[shared_index],
    aggregated_ba_demand.loc[shared_index],
    s=1,
    alpha=0.05,
)
ax.set_xlim(1e2, 1e6)
ax.set_ylim(1e2, 1e6)
ax.loglog()
ax.set_aspect("equal")
ticks = np.logspace(2, 6, 9)
plt.xticks(ticks)
plt.yticks(ticks)
ax.grid()
plt.xlabel("Reported BA Demand [MWh]")
plt.ylabel("Subregion demand aggregated by BA [MWh]")
plt.title("Demand by Balancing Authority")
plt.show()
```
image

## Calculated vs. Reported Demand
```python
status = "reported"
fig, ax = plt.subplots(figsize=(10, 10))
ba_ops = core_eia930__hourly_balancing_authority_operations
ax.scatter(
    ba_ops[f"demand_{status}_mwh"],
    ba_ops[f"net_generation_{status}_mwh"] - ba_ops[f"interchange_{status}_mwh"],
    s=1,
    alpha=0.05,
)
ax.set_xlim(1e2, 1e6)
ax.set_ylim(1e2, 1e6)
ax.loglog()
ax.set_aspect("equal")
ticks = np.logspace(0, 6, 13)
plt.xticks(ticks)
plt.yticks(ticks)
ax.grid()
plt.xlabel(f"{status.title()} Demand [MWh]")
plt.ylabel("Net Generation minus Interchange [MWh]")
plt.title(f"{status.title()} vs. Calculated BA Demand")
plt.show()
```
image

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+@multi_asset(
+    outs={
+        "core_eia930__hourly_balancing_authority_net_generation_by_energy_source": AssetOut(
+            io_manager_key="parquet_io_manager"
+        ),
+        "core_eia930__hourly_balancing_authority_operations": AssetOut(
+            io_manager_key="parquet_io_manager"
+        ),
```
zaneselvans, line 19: I went ahead and re-organized these tables into net generation by energy source and totals for generation, interchange, and demand as we discussed.

Makefile
```diff
@@ -79,7 +88,7 @@ docs-clean:
 .PHONY: docs-build
 docs-build: docs-clean
 	doc8 docs/ README.rst
-	coverage run ${covargs} -- ${CONDA_PREFIX}/bin/sphinx-build -v -W -b html docs docs/_build/html
+	coverage run ${covargs} -- ${CONDA_PREFIX}/bin/sphinx-build --jobs auto -v -W -b html docs docs/_build/html
```
zaneselvans, line 91: Parallelize the docs build to speed it up.

Makefile
```diff
@@ -70,6 +70,15 @@ install-pudl: pudl-dev
...
 	rm -rf docs/_build
+	rm -rf docs/autoapi
+	rm -f docs/data_dictionaries/pudl_db.rst
+	rm -f docs/data_dictionaries/codes_and_labels.rst
+	rm -rf docs/data_dictionaries/code_csvs
+	rm -f docs/data_sources/eia*.rst
+	rm -f docs/data_sources/epacems*.rst
+	rm -f docs/data_sources/ferc*.rst
+	rm -f docs/data_sources/gridpathratoolkit*.rst
+	rm -f docs/data_sources/phmsagas*.rst
```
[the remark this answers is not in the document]
cmgosnell: hm i don't have a bunch of cram in these places. but maybe mine hasn't barfed in a while

docs/templates/eia923_child.rst.jinja
```diff
@@ -66,26 +64,30 @@ both the Form and the Form Instructions files.
 
 {% block notable_irregularities %}
 
-**File Naming Conventions**
+File Naming Conventions
+-----------------------
```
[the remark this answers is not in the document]
cmgosnell: i'm a lil surprised to see this and some other little docs cleanup stuff in this pr. its all good but ⏲️

src/pudl/metadata/dfs.py
```diff
@@ -4,234 +4,95 @@
 
 import pandas as pd
 
-FERC_ACCOUNTS: pd.DataFrame = pd.DataFrame(
```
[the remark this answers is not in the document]
cmgosnell: i think you added this a while back in hopes that we could link it to all of the FERC1 tables that do have some ferc account info in them. I personally think it's helpful and certainly not harming anything being here.

src/pudl/metadata/fields.py
```diff
@@ -895,6 +923,28 @@
...
+        "description": "Electricity demand adjusted by EIA to reflect non-physical commercial transfers through pseudo-ties and dynamic scheduling.",
+        "unit": "MWh",
+    },
+    # TODO[zaneselvans] 2024-04-20: Is the timestamp when the forecast was made? Or the
+    # time at which the forecast is trying to predict demand?
+    "demand_forecast_mwh": {
+        "type": "number",
+        "description": "Day ahead demand forecast.",
+        "unit": "MWh",
+    },
```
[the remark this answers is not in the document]
cmgosnell: why not put out a call to users who actually play with this value? i'd personally say keep this description as is for now but see if folks generally know more about this.

src/pudl/metadata/resources/ferc1.py
```diff
@@ -311,16 +311,6 @@
...
-    "core_ferc__codes_accounts": {
-        "description": "Account numbers from the FERC Uniform System of Accounts for Electric Plant, which is defined in Code of Federal Regulations (CFR) Title 18, Chapter I, Subchapter C, Part 101. (See e.g. https://www.law.cornell.edu/cfr/text/18/part-101).",
-        "schema": {
-            "fields": ["ferc_account_id", "ferc_account_description"],
-            "primary_key": ["ferc_account_id"],
-        },
-        "sources": ["ferc1"],
-        "etl_group": "static_ferc1",
-        "field_namespace": "ferc1",
-    },
```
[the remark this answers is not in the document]
cmgosnell: I personally think this table could be helpful and think it should stay in.

test/integration/output_test.py
```diff
@@ -186,17 +186,3 @@ def test_service_territory_outputs(pudl_engine, df_name):
...
-    [
-        "out_ferc714__hourly_estimated_state_demand",
-    ],
-)
-def test_state_demand_outputs(pudl_engine, df_name):
-    """Test state demand analysis methods."""
-    df = pd.read_sql(df_name, pudl_engine)
-    assert isinstance(df, pd.DataFrame), f"{df_name} is {type(df)} not DataFrame!"
-    logger.info(f"Found {len(df)} rows in {df_name}")
-    assert not df.empty, f"{df_name} is empty!"
```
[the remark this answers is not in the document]
cmgosnell: if the number of rows is being tested then this is empty test does seem duplicative and not worth replicating for the parquet format

docs/templates/eia923_child.rst.jinja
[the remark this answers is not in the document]
zaneselvans: I didn't bill for a lot of this peripheral stuff. It just came up with trying to get rid of the dead links for tables that aren't written to the DB because of the move to more parquet-only tables.

Makefile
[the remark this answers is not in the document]
zaneselvans: Usually the build script cleans up after itself. I just got tired of having to remove the transient .rst files by hand.

src/pudl/metadata/resources/ferc1.py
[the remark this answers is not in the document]
zaneselvans: I agree having a table defining the FERC accounts would be good, it just feels like it should probably list all the FERC accounts, and either vary by year, or represent the most recent version of the accounts, whereas this one is hand compiled, linked to a particular year of DBF data, and only covers the accounts that show up in the Electric Plant in Service table.

I imagine a complete table could be built from information in the FERC Form 1 XBRL taxonomy / metadata.

But I can add the old table back in if you feel like it's valuable in the interim.

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
zaneselvans: Yeah, I'll ask GregM and Jacques on the main issue.

test/validate/service_territory_test.py
```diff
@@ -67,18 +60,18 @@ def test_minmax_rows_and_year_in_ferc714_hourly_planning_area_demand(
...
     if not live_dbs:
         pytest.skip("Data validation only works with a live PUDL DB.")
-    hpad_ferc714 = pudl_out_orig.__getattribute__(df_name)()
-    _ = hpad_ferc714.pipe(
-        pv.check_min_rows, expected_rows=expected_rows, margin=0.0, df_name=df_name
-    ).pipe(pv.check_max_rows, expected_rows=expected_rows, margin=0.0, df_name=df_name)
+    parquet_path = PudlPaths().parquet_path(resource_id)
+    meta = pq.read_metadata(parquet_path)
+    assert meta.num_rows == expected_rows
+    hpad_ferc714 = pd.read_parquet(parquet_path, dtype_backend="pyarrow")
```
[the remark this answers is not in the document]
cmgosnell: Not required but it'd be great to add this first portion as a min max row test for all the parquet outputs

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,134 @@
...
+            ]
+            + ["net_generation_unknown_reported_mwh"]
+        ]
+        .rename(
+            # Rename columns so that they contain only the energy source and the level
+            # of processing with the pattern: energysource_levelofprocessing so the
+            # column name can be split on "_" to build a MultiIndex before stacking.
+            lambda col: col.removeprefix("net_generation_").removesuffix("_mwh"),
+            axis="columns",
+        )
```
cmgosnell, line 66: okay this means that the `EIA930_GENERATION_ENERGY_SOURCES` cannot have any `_` in those strings. That seems like an important implied detail. Can you add a little check for that somewhere that yells if this is ever not the case?

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,134 @@
...
+    netgen_by_source = (
+        netgen_by_source.stack(level="generation_energy_source", future_stack=True)
+        .rename(columns=lambda x: f"net_generation_{x}_mwh")
+        .reset_index()
+    )
+
+    # TODO[zaneselvans] 2024-04-20: Verify that sum of net generation from all fuels
+    # adds up to the reported total
+    # NOTE: currently there are some BIG differences between the calculated total and
+    # the reported total.
```
cmgosnell, line 82: is this still a todo if you've verified that there are big differences? You've already added a note about this in the function docs so i would suggest removing this.

test/validate/service_territory_test.py
```diff
@@ -30,34 +32,25 @@ def test_minmax_rows(
...
-        expected_rows: Expected number of rows that the dataframe should
-            contain when all data is loaded and is output without aggregation.
-        df_name: Shorthand name identifying the dataframe, corresponding
-            to the name of the function used to pull it from the PudlTabl
-            output object.
+        expected_rows: Expected number of rows that the dataframe should contain when
+            all data is loaded and is output without aggregation.
+
+        df_name: Shorthand name identifying the dataframe, corresponding to the name of
+            the function used to pull it from the PudlTabl output object.
```
cmgosnell, line 39: weird space in here!

test/validate/service_territory_test.py
```diff
@@ -30,34 +32,25 @@ def test_minmax_rows(
...
-    _ = (
-        pudl_out_orig.__getattribute__(df_name)()
-        .pipe(
-            pv.check_min_rows, expected_rows=expected_rows, margin=0.0, df_name=df_name
-        )
-        .pipe(
-            pv.check_max_rows, expected_rows=expected_rows, margin=0.0, df_name=df_name
-        )
-    )
+    assert pudl_out_orig.__getattribute__(df_name)().shape[0] == expected_rows
```
cmgosnell, line 43: this seems totally fine.. but why not keep doing the standard thing? i get that we've long transitioned to always using a 0% margin and therefor this test for a long long time has really been a `test_expected_rows`... but still. absolutely a nit, but it always feels bad to slowly diverge away from something in a way that makes it harder to change it in a systematic way.

test/validate/state_demand_test.py
```diff
@@ -2,37 +2,29 @@
...
+from pudl.workspace.setup import PudlPaths
 
 logger = logging.getLogger(__name__)
 
 
 @pytest.mark.parametrize(
-    "df_name,expected_rows",
-    [("predicted_state_hourly_demand", 6_706_318)],
+    "resource_id,expected_rows",
+    [("out_ferc714__hourly_estimated_state_demand", 6_706_318)],
```
cmgosnell, line 15: resource_id because its called a resource in our metadata? not asset_name or table_name?

cmgosnell (changes requested):
I have a few mostly minor name suggestions and questions but the mechanics of all of this look great! this seems very close imo!!

On the table names: does it add anything to have `_balancing_authority_` in all of these table names? If it does I'd suggest moving it to the end so that the beginning of the unique table name is the "what" of the table is at the front (same feels as [this comment](https://github.com/catalyst-cooperative/pudl/pull/3588#pullrequestreview-2022642660))

i skimmed the docs changes in `docs` folder in this pr just bc of season of the docs upcoming overall and the overall time we have for this. the mechanics of not pooping out the links to the parquet only tables seem good good.

test/validate/state_demand_test.py
zaneselvans: I used `resource_id` because that's what we call these strings in the context of our metadata classes, like `Package.from_resource_ids()`

test/validate/service_territory_test.py
zaneselvans: I used the shape attribute because the normal way that we do it counts up the number of rows twice, (once for min, another for max) but `df.shape` doesn't require any calculation, and this dataframe has millions of rows, and I think we'll do something more like this going forward (as with reading the number of rows from the Parquet metadata, where it can be found without doing any computation.

src/pudl/metadata/fields.py
```diff
@@ -4736,24 +4814,36 @@
...
+        "datetime_utc": {
+            "description": "Timestamp at the end of the hour for which the data is reported."
+        },
+    },
+    "core_eia930__hourly_balancing_authority_subregion_demand": {
+        "datetime_utc": {
+            "description": "Timestamp at the end of the hour for which the data is reported."
+        },
+        "demand_reported_mwh": {
+            "description": "Originally reported electricity demand for the balancing area subregion. Method of calculation may vary by BA.",
```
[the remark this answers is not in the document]
zaneselvans: Internally the BAs have different ways of calculating "demand" as generation minus interchange, because some of them use physical metered values, and others use virtual transactions that "move" generation economically but not physically. So the way they're allocating demand within their boundaries isn't necessarily the same for different BAs.

src/pudl/metadata/fields.py
```diff
@@ -3196,6 +3277,21 @@
...
+                CODE_METADATA["core_eia__codes_balancing_authorities"]["df"][
+                    "region_code_eia"
+                ].dropna()
+            )
+        },
+    },
+    "region_eia": {
+        "type": "string",
+        "description": "Human-readable name of the EIA balancing region.",
+    },
```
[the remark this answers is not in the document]
zaneselvans: I was kinda hoping they were general EIA regions bug if they are not (because of course they are not) then I will make it clear that they pertain only to the eia930. The documentation I found didn't give much indication as to the scope or meaning of the regions. Just codes and associated timezones.

src/pudl/metadata/fields.py
```diff
@@ -1916,13 +1963,32 @@
...
         "type": "boolean",
         "description": (
             "Indicates whether the associated state reports data within the EPA's "
             "Continuous Emissions Monitoring System."
         ),
     },
+    "is_generation_only": {
+        "type": "boolean",
+        "description": "Indicates whether the balancing authority is generation-only, meaning it does not serve retail customers and thus reports only net generation and interchange, but not demand.",
+    },
```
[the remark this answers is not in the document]
zaneselvans: Not 100% certain. Many of them seem to be renewable projects. I think this really belongs in a BA entity / SCD table, but we don't have one yet, and I wanted to capture it somewhere, and this was the only obviously place at the moment.

zaneselvans:
> On the table names: does it add anything to have `_balancing_authority_` in all of these table names? If it does I'd suggest moving it to the end so that the beginning of the unique table name is the "what" of the table is at the front.

@cmgosnell here's my thought process on the table names. Although it is long, I think we probably do want `balancing_authority` in the table names because `subregion_demand` and `operations` both beg the question "subregion/operations of... what?" and listing out all of demand, net generation, and interchange for BA level totals is too verbose (and EIA also calls it the BA operations table in their docs).

We also already have some hourly demand tables, which include what entity / geography the demand pertains to, and I think having `balancing_authority` fit that same pattern helps make it clear these are all conceptually similar tables:

* `out_ferc714__hourly_estimated_state_demand`
* `out_ferc714__hourly_planning_area_demand`

The current BA table naming matches the pattern that's established with those tables of "hourly" + descriptive entity + reported variable (here demand, interchange or generation). This pattern also reads like something you'd actually say, which makes it easier to remember without needing to look it up every time, unlike "hourly demand estimated state" (or was it hourly demand state estimated?) and avoids the need for a connector word like "hourly estimated demand by state" which can work okay when there's just one dimension, but as we've seen in the net generation tables, gets unwieldy quickly when you have 2 or 3. Also `hourly_operations_by_balancing_authority` didn't really seem to make sense and `hourly_interchange_by_balancing_authority` also seemed inaccurate since it's really by pairs of balancing authorities which `hourly_balancing_authority_interchange` allows in English.

test/validate/service_territory_test.py
zaneselvans: Fixed.

src/pudl/transform/eia930.py
zaneselvans: TODO TODONE

docs/data_access.rst
```diff
@@ -109,57 +109,77 @@ version of Datasette (see above). These nightly build outputs can be accessed us
...
+decompress them locally, at the command line on Linux, MacOS, or Windows you can use the
+``gunzip`` command. (Git for Windows installs ``gzip`` / ``gunzip`` by default, and it
+can also be installed using the conda package manager).
+
+.. code-block:: console
+
+   $ gunzip *.sqlite.gz
+
+If you're not familiar with using Unix command line tools in Windows you can also use a
+3rd party tool like `7zip <https://www.7-zip.org/download.html>`__.
```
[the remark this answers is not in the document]
zaneselvans: This didn't seem ideal. There's a bunch of additional `.sqlite.gz` files below. Maybe I'll just put it back at the bottom of the section.

src/pudl/metadata/enums.py
```diff
@@ -73,6 +73,28 @@
...
+    "America/Anchorage",
+    "America/Chicago",
+    "America/Denver",
+    "America/Los_Angeles",
+    "America/New_York",
+    "America/Phoenix",
+    "Pacific/Honolulu",
+]
+
+EIA930_GENERATION_ENERGY_SOURCES = [
```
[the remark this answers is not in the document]
zaneselvans: Swapped

test/validate/service_territory_test.py
zaneselvans: Space == bad. Removed.

src/pudl/transform/eia930.py
zaneselvans: I've added an assertion where the global constant is defined, but I'm not sure that's what you had in mind here.

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
zaneselvans: I went with:

* `balancing_authority_region_name_eia`
* `balancing_authority_region_code_eia`

Which matched the existing:

* `balancing_authority_name_eia`
* `balancing_authority_code_eia`
* `balancing_authority_subregion_name_eia`
* `balancing_authority_subregion_code_eia`

And did not result in weird inconsistencies with other things then wanting to also have `_eia930` suffixes.

src/pudl/metadata/resources/ferc1.py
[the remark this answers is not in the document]
zaneselvans: Restored the table.

src/pudl/metadata/dfs.py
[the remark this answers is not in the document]
zaneselvans: I definitely added it way back when, before we really understood the FERC data well and long before XBRL made its debut on the scene. Something like this seems useful, this table is just a bad attempt at it right now.

src/pudl/metadata/fields.py
```diff
@@ -895,6 +938,28 @@
         "description": "Gross megawatt-hours delivered in power exchanges and used as the basis for settlement.",
         "unit": "MWh",
     },
+    "demand_adjusted_mwh": {
+        "type": "number",
+        "description": "Electricity demand adjusted by EIA to reflect non-physical commercial transfers through pseudo-ties and dynamic scheduling.",
+        "unit": "MWh",
+    },
+    # TODO[zaneselvans] 2024-04-20: Is the timestamp when the forecast was made? Or the
+    # time at which the forecast is trying to predict demand?
```
cmgosnell, line 947: just noting this remaining todo (which seems truly still undone)

src/pudl/transform/eia930.py
cmgosnell: that is fine! not blocking at all but adding a comment in this function about this implied assumption would be good.

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
cmgosnell: okay these seem much better to me! if these `balancing_authority_region_*`s are indeed 930 specific (which i expect that they are, i would lean towards wanting the full `_eia930` suffix but that is not a hill i will die on and does not feel blocking to me

test/validate/service_territory_test.py
cmgosnell: yea this seems like a much better method. i like what you are doing here, it is just a divergence from the standard process which imo is confusing to our future selves and more error prone/harder to track down future fixes. can you make a new smol issue which is a cleaning up of the min/max row validation tests with this as the suggested methodology and a note that when that happens we should make sure that these more one-off versions should be swooped up into that new methodology?

src/pudl/transform/eia930.py
```diff
@@ -0,0 +1,132 @@
...
+)
+def core_eia930__hourly_balancing_authority_assets(
+    raw_eia930__balance: pd.DataFrame,
+):
+    """Separate raw_eia930__balance into net generation and demand tables.
+
+    Energy source starts out in the column names, but is stacked into a categorical
+    column. For structural purposes "interchange" is also treated as an "energy source"
+    and stacked into the same column. For the moment "total" (sum of all energy sources)
+    is also included, because the reported and calculated totals acrss all energy
```
cmgosnell, line 31:
```suggestion
    is also included, because the reported and calculated totals across all energy
```

cmgosnell (changes requested):
tiny requests!

I feel pretty strongly about not leading with the entity of a table unless that table is full of scd or static entity type data. so many of our tables have implied entities (plants! generators! utilities! etc etc) that we don't put in the table name. but i honestly don't have a good solution for the `operations` table bc yea i think that name would be too short.

src/pudl/metadata/fields.py
zaneselvans: Yep, still out there. Will post Q more broadly.

src/pudl/transform/eia930.py
zaneselvans: fixed

cmgosnell (approved):
🎉 ⏳

PR #3588, opened 2024-04-24
Enumerate all AEO table 54 schemas.

### LINKED ISSUE #3581: Transform high-priority AEO Table 54 data into normalized assets

jdangerx:
# Overview

doesn't really close #3581, but sure helps!

Table 54 has a bunch of confusingly heterogeneous data, here I'm sorting them and organizing them in some way.

Also, I'm introducing a new pattern of defining schemas in a 'staging area' that can be reviewed + checked in, but don't get published or expect there to be an asset involved. This enables the following flow:

1. figure out some resource designs ahead of time - often we'll want to design multiple resources at once due to interdependencies
2. merge that
3. work on new assets that generate those resources, maybe even in parallel!
4. promote and merge new assets as they become complete, so we don't have to wait for *all* assets to be done to use *any* assets

# Testing

Nothing seems to have broken because the `RESOURCE_METADATA` dictionary is actually empty right now. Once we start promoting assets, then we'll have to make sure *those* assets actually get pulled in, etc.

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/metadata/resources/eiaaeo.py
+++ b/src/pudl/metadata/resources/eiaaeo.py
@@ -13,7 +13,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "technology_description_eiaaeo",
@@ -25,7 +25,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "technology_description_eiaaeo",
@@ -50,7 +50,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "fuel_type_eiaaeo",
@@ -59,7 +59,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "fuel_type_eiaaeo",
@@ -77,7 +77,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "gross_generation_mwh",
@@ -86,7 +86,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
             ],
@@ -110,7 +110,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "gross_generation_mwh",
@@ -119,7 +119,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
             ],
@@ -133,7 +133,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "customer_class",
@@ -141,7 +141,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "customer_class",
@@ -156,15 +156,15 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "generation_source_or_sink_type_eiaaeo",
-                "net_energy_for_load_mwh",
+                "net_load_mwh",
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "generation_source_or_sink_type_eiaaeo",
@@ -179,7 +179,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "customer_class",
@@ -188,7 +188,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "customer_class",
@@ -205,7 +205,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "service_category_eiaaeo",
@@ -214,7 +214,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "service_category_eiaaeo",
@@ -233,7 +233,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "fuel_type_eiaaeo",
@@ -241,7 +241,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "fuel_type_eiaaeo",
@@ -260,7 +260,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "fuel_type_eiaaeo",
@@ -269,7 +269,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "fuel_type_eiaaeo",
@@ -288,7 +288,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
         "schema": {
             "fields": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
                 "carbon_mass_tons",
@@ -299,7 +299,7 @@ _STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
             ],
             "primary_key": [
                 "report_year",
-                "emm_region_eiaaeo",
+                "electricity_market_module_region_eiaaeo",
                 "model_case_eiaaeo",
                 "projection_year",
             ],
@@ -342,7 +342,7 @@ _STAGING_FIELD_METADATA: dict[str, dict[str, Any]] = {
         "description": "Total carbon emissions in short tons.",
         "unit": "short_tons",
     },
-    "emm_region_eiaaeo": {
+    "electricity_market_module_region_eiaaeo": {
         "type": "string",
         "description": "AEO projection region.",
         "constraints": {
--- a/src/pudl/metadata/resources/eiaaeo.py
+++ b/src/pudl/metadata/resources/eiaaeo.py
@@ -1,0 +1,514 @@
+"""Table definitions for EIA AEO resources."""
+
+from typing import Any
+
+# 2024-04-24: workaround so we can *define* a bunch of schemas without
+# documenting them or expecting assets to exist
+_STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "core_eiaaeo__generation_electric_sector": {
+        "description": (
+            "Projected generation capacity & total generation in the electric "
+            "sector."
+        ),
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region",
+                "aeo_case",
+                "projection_date",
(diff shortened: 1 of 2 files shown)


src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
+"""Table definitions for EIA AEO resources."""
+
+from typing import Any
+
+# 2024-04-24: workaround so we can *define* a bunch of schemas without
+# documenting them or expecting assets to exist
+_STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "core_eiaaeo__generation_electric_sector": {
+        "description": (
+            "Projected generation capacity & total generation in the electric "
```
jdangerx, line 10: It's not clear to me that generation and capacity *should* live as separate columns in the same table, especially since the "fuel type" subdimension is slightly different between the two data types - there are extra "sales to customers" and "generation for own use" values there which apply to the total generation but not the capacity.

I wonder if there should be "end-use/electricity-sector generation by fuel type" tables, and then also "end-use/electricity-sector generation *not by fuel type*" tables that include the "total across all fuel types" and "sales to customer/grid" and "generation for own use" columns.

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+                "aeo_case",
+                "projection_date",
+                "aeo_electricity_sector_fuel_type",
+                "net_summer_capacity_mw",  # should this be capacity_mw_eia?
+                # should we include "first modeled year" as a field, to help interpret all the cumulative fields?
+                "net_summer_capacity_cumulative_planned_additions_mw",
+                "net_summer_capacity_cumulative_unplanned_additions_mw",
+                "net_summer_capacity_cumulative_total_additions_mw",  # this is just planned + unplanned - do we even care about publishing this?
+                "net_summer_capacity_cumulative_retirements_mw",
+                "gross_generation_mwh",  # chose gross bc the numbers add up to "total generation" which then equals "sales to customers + generation for own use" - but... is that right?
```
jdangerx, line 26: i.e., the "renewable" + "coal" + "whatever" + ... all add up to the "total generation" projected for the year. Which then is equal to "sales + generation for own use" - which sounds to me like gross, not net, generation.

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+    },
+    "core_eiaaeo__electricity_sales": {
+        "description": "Projected electricity sales.",
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region",
+                "aeo_case",
+                "projection_date",
+                "aeo_sales_sector",
```
jdangerx, line 82: We could unstack this since there's only residential/industrial/commercial/transportation/total - should we?

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+    },
+    "core_eiaaeo__net_energy_for_load": {
+        "description": "Projected net energy for load.",
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region",
+                "aeo_case",
+                "projection_date",
+                "aeo_net_energy_for_load_sector",
```
jdangerx, line 105: We *could* unstack this, it's:

```
                "gross_international_imports",
                "gross_international_exports",
                "gross_interregional_electricity_imports",
                "gross_interregional_electricty_exports",
                "purchases_from_combined_heat_and_power",
                "electric_power_sector_generation_for_customer",
                "total", 
```

... should we?

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+    },
+    "core_eiaaeo__end_use_prices_by_sector": {
+        "description": "Projected electricity cost to the end user by sector.",
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region",
+                "aeo_case",
+                "projection_date",
+                "aeo_sales_sector",
```
jdangerx, line 128: If we unstack the res/ind/com/trans stuff above we should unstack it here too. And if not, we shouldn't.

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+        "description": (
+            "Projected electricity cost to the end user by service category."
+        ),
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region",
+                "aeo_case",
+                "projection_date",
+                "aeo_service_category",
```
jdangerx, line 155: This is just generation/transmission/distribution - should we unstack it?

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+            "including electricity-only and combined-heat-and-power plants "
+            "that have a regulatory status."
+        ),
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region",
+                "aeo_case",
+                "projection_date",
+                "aeo_fuel_consumption_type",
```
jdangerx, line 184: Yadda yadda should we unstack this? coal/natural gas/oil

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+            "electricity-only and combined-heat-and-power plants that have a "
+            "regulatory status."
+        ),
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region",
+                "aeo_case",
+                "projection_date",
+                "aeo_fuel_cost_type",
```
jdangerx, line 211: coal/gas/two types of oil - unstack?

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+            "regulatory status."
+        ),
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region",
+                "aeo_case",
+                "projection_date",
+                "carbon_mass_tons",
+                "co2_mass_tons",
```
jdangerx, line 241: I chose to unstack this because I already saw a bunch of these fields lying around. :shrug:

jdangerx (commented):
OK, broadly there are:
* questions about whether we should stack/unstack certain fields
* questions about whether certain fields belong in the same table
* I think the names are mostly awful, though I guess if we decide to unstack a bunch of stuff then we are freed from the burden of coming up with names.

My overall pitch is, I think:
* unstack all the places where I've flagged that we might want to
* have generation-by-fuel and generation-in-general tables for electricity sector + end-use sector, so the fuel-type type is clean and we don't have to mess around with munging various total columns.

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+        "type": "date",
+        "description": "Date at which this data is projected to be true.",
+    },
+    "real_value_base_year": {
+        "type": "integer",
+        "description": "Nominal monetary values are converted to values from this year.",
+    },
+    "electricity_market_module_region": {
+        "type": "string",
+        "description": "AEO projection region. Maybe the same as NERC region?",
```
jdangerx, line 277: As mentioned in standup, here's a fun question... this seems similar to NERC region, are they the same **concept** or do they just happen to have similar values?

cmgosnell (commented):
just a smattering of comments/responses:

i rlly like the staging resource! very helpful

is it easy to check if the totals that are sprinkled all around calculable with the sub-components? and if so can we just get rid of them alllll?

on the unstacking: i think i am lightly partial to not unstacking. especially if we can get rid of the totals it seems more tidy to have the categorical-like columns as columns when the "data" column(s) all contain the same type of value. especially if we remove the totals!

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+            "``real_value_base_year``)."
+        ),
+        "unit": "USD/MWh",
+    },
+    "sales_mwh": {
+        "type": "number",
+        "description": "ALREADY EXISTS, JUST HERE FOR CONVENIENCE",
+        "unit": "MWh",
+    },
+    "net_summer_capacity_mw": {
```
jdangerx, line 471:
```suggestion
    "summer_capacity_mw": {
```

already exists!

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+        "type": "number",
+        "description": "The steady hourly output that generating equipment is expected to apply to system load.",
+        "unit": "mw",
+    },  # should this be capacity_mw_eia?
+    "net_summer_capacity_cumulative_planned_additions_mw": {
```
jdangerx, line 476:
```suggestion
    "summer_capacity_cumulative_planned_additions_mw": {
```

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+        "type": "number",
+        "description": (
+            "The total planned additions to generating capacity since the "
+            "beginning of the AEO period."
+        ),
+        "unit": "mw",
+    },
+    "net_summer_capacity_cumulative_unplanned_additions_mw": {
```
jdangerx, line 484:
```suggestion
    "summer_capacity_cumulative_unplanned_additions_mw": {
```

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,514 @@
...
+        "type": "number",
+        "description": (
+            "The total unplanned additions to generating capacity since the "
+            "beginning of the AEO period."
+        ),
+        "unit": "mw",
+    },
+    "net_summer_capacity_cumulative_total_additions_mw": {  # this is just planned + unplanned - do we even care about publishing this?
```
jdangerx, line 492:
```suggestion
    "summer_capacity_cumulative_total_additions_mw": {  # this is just planned + unplanned - do we even care about publishing this?
```

jdangerx (commented):
We should split the generation tables into:

* `electric_power_sector_generation_by_technology`
* `end_use_sector_generation_by_fuel_type`
* `electric_power_sector_generation` (including total, sales, for own use)
* `end_use_sector_generation` (inc. total, sales, for own use)

We should rename the fields as commented.

We should remove calculable fields.

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,513 @@
+"""Table definitions for EIA AEO resources."""
+
+from typing import Any
+
+# 2024-04-24: workaround so we can *define* a bunch of schemas without
+# documenting them or expecting assets to exist
+_STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "core_eiaaeo__electric_sector_generation_by_technology": {
```
jdangerx, line 8: @cmgosnell I see you named your ATB tables `core_nrelatb__yearly_projection_*`, should I add `yearly_projection` to my table names too?

jdangerx: I did it, it's an important distinction and otherwise we'd rely on people knowing what AEO is.

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,524 @@
+"""Table definitions for EIA AEO resources."""
+
+from typing import Any
+
+# 2024-04-24: workaround so we can *define* a bunch of schemas without
+# documenting them or expecting assets to exist
+_STAGING_RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "core_eiaaeo__yearly_projection_electric_sector_generation_by_technology": {
```
cmgosnell, line 8: since our prs are twins rn, there is much overlap. @zaneselvans zane over there [suggested using `projected`](https://github.com/catalyst-cooperative/pudl/pull/3570#discussion_r1578269762) instead of `projection`. not actually sure why? zane why? is it bc it is verb-y? i like the ring of projected better

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,524 @@
...
+        "description": (
+            "Projected generation capacity & total generation in the electric "
+            "sector, broken out by technology."
+        ),
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region_aeo",
+                "modeling_case_aeo",
+                "projection_year",
```
cmgosnell, line 18: nice

cmgosnell: hm i know we literally came up with `modeling_case_aeo` yesterday, but a) it should almost certainly have the full `_eiaaeo` (🧑‍🌾) suffix and maybe for a light attempt at more brevity it would be `model_case_eiaaeo`? 

also the full `_eiaaeo` (🧑‍🌾) suffix note i believe should be applied to many of these fields.

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,524 @@
...
+            "generating systems in the residential, commercial, and industrial "
+            "sectors used primarily for own-use generation, but which may also "
+            "sell some power to the grid. Solar photovoltaic capacity portion of "
+            "Renewable Sources in gigawatts direct current; other technologies "
+            "in gigawatts alternating current."
+        ),
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region_aeo",
```
cmgosnell, line 53: i wonder if this guy should just be `region_eiaaeo`? that seems clear enough and distinct and less of a mouthful

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,524 @@
...
+    "core_eiaaeo__yearly_projection_net_energy_for_load": {
+        "description": "Projected net energy for load.",
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region_aeo",
+                "modeling_case_aeo",
+                "projection_year",
+                "generation_source_or_sink_type_aeo",
+                "net_energy_for_load_mwh",
```
cmgosnell, line 163: could this guy be `net_generation_for_load_mwh`? that would be more consistent w/ our other net gen names. alternatively we do have a `net_load_mwh` field which looks applicable

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,524 @@
...
+    "core_eiaaeo__yearly_projection_end_use_prices_by_sector": {
+        "description": "Projected electricity cost to the end user by sector.",
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region_aeo",
+                "modeling_case_aeo",
+                "projection_year",
+                "customer_class",
+                "nominal_price_per_mwh",
```
cmgosnell, line 186: i wonder if this guy should just be `price_per_mwh`? bc i think by default the $s across our tables are nominal and the real-ness is the exception. that's a lil weirder in the context of projections so i think there is a fine argument for keeping nominal. (also my always suggestion to move the "what" to the front of the name and qualifiers later for easier finding/sorting despite being slightly less colloquial:`price_nominal_per_mwh` and `price_real_per_mwh`)

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -0,0 +1,524 @@
...
+        ),
+        "schema": {
+            "fields": [
+                "report_year",
+                "electricity_market_module_region_aeo",
+                "modeling_case_aeo",
+                "projection_year",
+                "fuel_type_aeo",
+                "fuel_cost_per_mmbtu",
+                "fuel_cost_real_per_mmbtu_aeo",
```
cmgosnell, line 268: did you add the source suffix here bc the real-ness is source dependent? that seems realonable ;-)

cmgosnell (changes requested):
so many almost tables! this staging pattern is like getting an ultrasound for the soon to be born tables

i have some suggestions about the table names. which is mostly to bring the main "what" of the table as far to the forefront as is possible in our table format. here are my suggestions for each of the end bit of the table name (i don't love the `_in_` portion, but idk how else to do this while making it semi-colloquial)

-core_eiaaeo__yearly_projection_electric_sector_generation_by_technology: generation_in_electric_sector_by_technology
-core_eiaaeo__yearly_projection_end_use_sectors_generation_by_fuel_type: generation_in_end_use_sectors_by_fuel_type
-core_eiaaeo__yearly_projection_electric_sector_generation: generation_in_electric_sector
-core_eiaaeo__yearly_projection_end_use_sectors_generation: generation_in_end_use_sectors
-core_eiaaeo__yearly_projection_electricity_sales: sales_electricity/generation_sales/generation_electric_sales
-core_eiaaeo__yearly_projection_net_energy_for_load: net_energy_for_load
-core_eiaaeo__yearly_projection_end_use_prices_by_sector: end_use_prices_by_sector
-core_eiaaeo__yearly_projection_end_use_prices_by_service_category: end_use_prices_by_service_category
-core_eiaaeo__yearly_projection_fuel_consumption_by_type: fuel_consumption_by_type
-core_eiaaeo__yearly_projection_electric_sector_fuel_cost_by_type: fuel_cost_in_electric_sector_by_type
-core_eiaaeo__yearly_projection_electric_power_sector_emissions: emissions_in_electric_sector

src/pudl/metadata/resources/eiaaeo.py
jdangerx: Sometimes they refer to it as "EMM region," maybe we can go with that. There are other AEO datasets that are cut up by different regions e.g. 
image

src/pudl/metadata/resources/eiaaeo.py
jdangerx: Yeah, it just seemed funny to have "real" as its own field divorced from context - if it's AEO at least we know that the associated year is `report_year-1`. Outside of an AEO context this just doesn't really mean anything...

src/pudl/metadata/resources/eiaaeo.py
cmgosnell: ah yes i see. I think i would be partial to using the existing field `net_load_mwh`. I think right now it shows up in the context of pumped hydro plants in ferc1 tables but that seems relevant in a source/sink context. but this is not a hill i will die on i think what you have here is fine.

src/pudl/metadata/resources/eiaaeo.py
jdangerx: By "existing name" do you mean `emm_region_eiaaeo` or `electricity_market_module_region_eiaaeo`?

src/pudl/metadata/resources/eiaaeo.py
jdangerx: I guess the table name also provides context. Yeah, let's go for `net_load_mwh`.

cmgosnell (approved):
wahoo!

PR #3591, opened 2024-04-25
Speed up nb-output-clear step in pre-commit

jdangerx:
# Overview

I finally got fed up waiting several seconds for `nb-output-clear` to do its thing and made it only clear ipynbs that actually changed.

# Testing

1. made some trivial changes to ipynbs in devtools, set `verbose: true` on the hook
2. staged one ipynb change for commit
3. committed, saw that only the changed ipynb was passed through `nbconvert`
4. `git reset HEAD^`
5. only stage the precommit hook
6. commit, see that we skip the `nb-output-clear` step

```[tasklist]
# To-do list
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,11 +68,11 @@ repos:
         stages: [commit]
         language: system
         verbose: false
-        pass_filenames: false
-        always_run: true
-        entry:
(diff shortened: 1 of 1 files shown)


e-belfer (approved):
Works as expected on my end!

PR #3598, opened 2024-04-29
End use sectors generation by fuel type.

### LINKED ISSUE #3597: `core_eiaaeo__yearly_projected_generation_in_end_use_sectors_by_fuel_type`

jdangerx:
# Overview

Closes #3597 .

This adds the end use sector generation by fuel type.

What did you change?

Promoted the resource definition & fields, and then made a lil asset!

# Testing

Materialized things in Dagster.

```[tasklist]
# To-do list
```


src/pudl/transform/eiaaeo.py
```diff
@@ -193,9 +291,14 @@ def _decumulate(df: pd.DataFrame, decumulate_columns: list[str]) -> pd.DataFrame
         .pipe(_drop_aggregated_generation_data)
         .pipe(_collect_totals)
     )
+    # 1 BKWh = 1e12 Wh = 1e6 MWh
+    sanitized.loc[sanitized.units == "bkwh", "value"] *= 1e6
```
jdangerx, line 295: The existing `convert_units` method seems more tailored to wide tables, and we're converting in the narrow-table format here, so I'm keeping it like this for now. Probably worth pulling out into a separate function at some point, and maybe putting some default conversions into the `filter_enrich_sanitize` method.

src/pudl/transform/eiaaeo.py
```diff
@@ -55,6 +55,67 @@ def get_category_info(category_name: pd.Series) -> pd.Series:
     return fields
 
 
+def subtotals_match_reported_totals_ratio(
```
jdangerx, line 58: I split the old `check_totals_add_up` function into the two separate operations so that we could use them in different places, and then moved the original function into the body of the asset that it was specifically designed for.

src/pudl/workspace/datastore.py
```diff
@@ -208,7 +208,9 @@ class ZenodoDoiSettings(BaseSettings):
     phmsagas: ZenodoDoi = "10.5281/zenodo.10493790"
     nrelatb: ZenodoDoi = "10.5281/zenodo.10839268"
 
-    model_config = SettingsConfigDict(env_prefix="pudl_zenodo_doi_", env_file=".env")
+    model_config = SettingsConfigDict(
+        env_prefix="pudl_zenodo_doi_", env_file=".env", extra="ignore"
```
jdangerx, line 212: This lets you put settings for multiple different settings classes in the `.env` without blowing stuff up.

jdangerx (commented):
I feel like these various helper functions *could* live in some sort of shared place but I'm not quite sure about (a) how general they really are and (b) what is our strategy for discovering utility functions?

src/pudl/transform/eiaaeo.py
```diff
@@ -358,20 +358,115 @@ class AeoCheckSpec:
...
+
+    assert set(sanitized.topic.unique()) == {"electricity"}
+    assert set(sanitized.subtopic.unique()) == {"end_use_sectors"}
+    assert set(sanitized.units.unique()) == {"bkwh", "gw"}
+
+    trimmed = sanitized.drop(
+        columns=[
+            "topic",
+            "subtopic",
+            "units",
```
cmgosnell, line 423: not blocking: this seems unlikely to not be the case but it would be good to make sure that the `units == "gw"` always corresponds to `capacity` and `units == "bkwh"` always corresponds to `generation` .

src/pudl/transform/eiaaeo.py
```diff
@@ -358,20 +358,115 @@ class AeoCheckSpec:
...
 
+@asset(io_manager_key="pudl_io_manager")
+def core_eiaaeo__yearly_projected_generation_in_end_use_sectors_by_fuel_type(
+    raw_eiaaeo__electric_power_projections_regional,
+):
+    """Projected generation capacity + gross generation in end-use sectors.
+
+    This includes data that's reported by fuel type and ignores data that's
+    only reported at the system-wide level, such as total generation, sales to
+    grid, and generation for own use.
```
cmgosnell, line 400: why are you removing sales to grid and generation for own use? maybe bc these are ending up in a differnt table bc different pks? i half remember that we talked about this in the schema design but it would be good to add a note here about it.

cmgosnell (changes requested):
one request for a lil more explanation but this look so good!

cmgosnell (approved):
the merge conflicts made this round slightly harder to see but it looks like you added that one lil comment and unit assertion so I feel good about this!

Issue #3593, opened 2024-04-27
Nightly Build Failure 2024-04-27


zaneselvans:
# Overview

* The `etl_full.yml` settings for the NREL ATB included the years 2019 and 2020, which aren't yet working (See #3576) and Pydantic's validation of the settings correctly failed. I went ahead and removed those years from the settings file on `main` so we can get another attempted build tonight.
* However, this failure doesn't happen locally when I try to run the full ETL with the old settings, which is weird.
* While investigating this I was confused by the NREL ATB extraction, which doesn't seem to make any use of the settings or datastore. So maybe this is only working because it's relying on defaults that aren't informed by the ETL settings at all?
* The `raw_nrelatb__data` asset claims to require the `datastore` and `dataset_settings` resources, but doesn't actually make use of them.
* The NREL ATB `Extractor` claims to require a `Datastore` as input, but doesn't receive one.
* But looking at the other tabular extractors, they also don't seem to use any resources (even though they obviously must) so maybe there is a bunch of magic happening in the background? Can we document what is going on?
* It looks like there's a bit of stale documentation in the extraction system, with a mix of references to Excel and CSV files in places where they are not appropriate.

```py
class Extractor(ParquetExtractor):
    """Extractor for NREL ATB."""

    def __init__(self, *args, **kwargs):
        """Initialize the module.

        Args:
            ds (:class:datastore.Datastore): Initialized datastore.
        """
        self.METADATA = GenericMetadata("nrelatb")
        super().__init__(*args, **kwargs)


raw_nrelatb__all_dfs = raw_df_factory(Extractor, name="nrelatb")


@asset(
    required_resource_keys={"datastore", "dataset_settings"},
)
def raw_nrelatb__data(raw_nrelatb__all_dfs):
    """Extract raw NREL ATB data from annual parquet files to one dataframe.

    Returns:
        An extracted NREL ATB dataframe.
    """
    return Output(value=raw_nrelatb__all_dfs["data"])
```

cmgosnell:
ty for catching the non-working partitions in the full settings!

A lot of the magic is happening via `extract.extractor.raw_df_factory` which runs `extract.extractor.partition_extractor_factory` which uses the datastore and the dataset_settings. I was mirroring the eia 176 extract which required those two as inputs into the asset but doesn't pass them around - but instead accesses them within `raw_df_factory`

jdangerx:
Tangible outcome here is:
[...]

PR #3603, opened 2024-05-02
Remove unused resource keys from asset definitions

zaneselvans:
# Overview

When debugging the 2024-04-27 nightly build failure #3593 I noticed that there were a bunch of asset definitions annotated with required resource keys that were not being used. My guess is that this was due to cut-and-paste without understanding how Dagster resources work, but maybe I'm the one misunderstanding them?  If you don't actually use a resource within the asset being defined, are there good reasons to nevertheless supply it with these resources?

# Testing

```[tasklist]
# To-do list
- [x] Run `make pytest-integration` locally
```

### DIFF
--- a/src/pudl/extract/eia176.py
+++ b/src/pudl/extract/eia176.py
@@ -39,9 +39,7 @@ class Extractor(CsvExtractor):
 raw_eia176__all_dfs = raw_df_factory(Extractor, name="eia176")
 
 
-@asset(
-    required_resource_keys={"datastore", "dataset_settings"},
-)
+@asset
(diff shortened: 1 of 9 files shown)


bendnorman (approved):
I think we probably forgot to remove these resource keys in the [176 extractor PR](https://github.com/catalyst-cooperative/pudl/pull/3402). The dependencies are now specified in this extractor `@op`: https://github.com/catalyst-cooperative/pudl/blob/3a6619e158f2a72deb5681741a410426a5daeb08/src/pudl/extract/extractor.py#L330-L334

PR #3604, opened 2024-05-03
Unpin GDAL dependency

zaneselvans:
# Overview

* A while ago I pinned `gdal==3.8.4` because `3.8.5` was causing problems in the `pudl-archiver` repository due to upstream packaging / dependency issues. They've been resolved now.
* Bumped minimum versions of some other dependencies to reduce the environment solving search space.
* At some point recently the way our github badges are rendered changed, putting each on a separate line, which takes up a bunch of space. Tried to fix that.
* Fix a column naming issue in the EIA AEO asset checks

# Testing

I re-locked the conda dependencies and ran the `pudl-archiver` tests locally with it installing PUDL from this branch, and all went normally.

```[tasklist]
# To-do list
```


src/pudl/extract/eiaaeo.py
```diff
@@ -459,6 +459,6 @@ def raw_table_54_invariants(df: pd.DataFrame) -> AssetCheckResult:
     assert not df.empty
     assert df.notna().all().all()
     # covers 20 cases and 26 electricity market module regions (25 regions + 1 national)
-    assert len(df.case.value_counts()) == 20
+    assert len(df.model_case_eiaaeo.value_counts()) == 20
```
zaneselvans, line 462: Column name needed to be updated. Was causing asset check failure.

jdangerx (changes requested):
Eep, sorry about that. Let's add `blocking=True` to that asset check - then feel free to dismiss & merge or just ping me again.

jdangerx (approved):

PR #3606, opened 2024-05-03
Refactor EIA AEO totals checks.

jdangerx:
# Overview

As part of dealing with #3597 I wanted to pull the totals checks apart so I could use *only* the "sums within one fact make sense" logic. 

When making #3598 I realized I should probably take that logic out of that PR so that both PRs are a little easier to read / all other assets can be based on this branch instead of stacking asset PRs on top of each other.

# Testing

unit tests + materializing the asset

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/transform/eiaaeo.py
+++ b/src/pudl/transform/eiaaeo.py
@@ -63,6 +63,13 @@ def subtotals_match_reported_totals_ratio(
     Group by some key, then check that within each group the non-``"total"``
     values sum up to the corresponding ``"total"`` value.
 
+    Checks the list of fact columns to in aggregate, but if you want to check
+    that *each* column sums up correctly, individually, you can call this
+    function once per column.
+
+    TODO 2024-05-06: it may make sense to pass the threshold into this
+    function, which would clean up the call sites.
+
     Args:
         df: the dataframe to investigate
         pk: the key to group facts by
@@ -358,7 +365,7 @@ class AeoCheckSpec:
 
     name: str
     asset: str
-    num_rows_by_report_year: dict[str, int]
+    num_rows_by_report_year: dict[int, int]
     num_in_categories: dict[str, int]
 
 


src/pudl/transform/eiaaeo.py
```diff
@@ -193,9 +291,14 @@ def _decumulate(df: pd.DataFrame, decumulate_columns: list[str]) -> pd.DataFrame
         .pipe(_drop_aggregated_generation_data)
         .pipe(_collect_totals)
     )
+    # 1 BKWh = 1e12 Wh = 1e6 MWh
```
jdangerx, line 294: This unit conversion cleanup was just... there for the taking. Not really related to the rest of this PR.

src/pudl/transform/eiaaeo.py
```diff
@@ -218,15 +321,7 @@ def _decumulate(df: pd.DataFrame, decumulate_columns: list[str]) -> pd.DataFrame
...
-    ratio_close_additions_to_total, ratio_close_reported_calculated = (
-        check_totals_add_up(unstacked)
-    )
-    assert (
-        0.999 < ratio_close_additions_to_total <= 1.0
-    ), f"planned + unplanned vs. total: {ratio_close_additions_to_total}"
-    assert (
-        0.999 < ratio_close_reported_calculated <= 1.0
-    ), f"reported vs. calculated totals: {ratio_close_reported_calculated}"
+    _check_totals_add_up(unstacked)
```
jdangerx, line 324: Now that the useful functionality of "do the summing" is encapsulated, it seemed much easier to have a separate "make some assertions" function to do all the asset-specific assertions.

src/pudl/transform/eiaaeo.py
```diff
@@ -178,6 +241,41 @@ def _decumulate(df: pd.DataFrame, decumulate_columns: list[str]) -> pd.DataFrame
             columns=[f"cumulative_{c}" for c in decumulate_columns]
         )
 
+    def _check_totals_add_up(unstacked) -> None:
```
jdangerx, line 244: I pulled this function into the asset scope since the thresholds + specific columns are all very asset-specific.

src/pudl/workspace/datastore.py
```diff
@@ -208,7 +208,9 @@ class ZenodoDoiSettings(BaseSettings):
     phmsagas: ZenodoDoi = "10.5281/zenodo.10493790"
     nrelatb: ZenodoDoi = "10.5281/zenodo.10839268"
 
-    model_config = SettingsConfigDict(env_prefix="pudl_zenodo_doi_", env_file=".env")
+    model_config = SettingsConfigDict(
```
jdangerx, line 211: This makes it so that having `.env` values for other settings classes doesn't break the validation here.

src/pudl/transform/eiaaeo.py
```diff
@@ -55,6 +55,67 @@ def get_category_info(category_name: pd.Series) -> pd.Series:
...
+        df.loc[df[dimension_column] != "total", pk + fact_columns].groupby(pk).sum()
+    )
+    joined = calculated_totals.join(
+        reported_totals, lsuffix="_calc", rsuffix="_reported", how="inner"
+    )
+    close = np.isclose(
+        joined.loc[:, [f"{c}_calc" for c in fact_columns]],
+        joined.loc[:, [f"{c}_reported" for c in fact_columns]],
+    )
+    return close.sum() / reported_totals.size
```
cmgosnell, line 96: the only think i don't love about this implementation is that it will be hard to see which column in `fact_columns` is the source of the problem. and because you put the assertion outside of this function once you've hit the assertion you no longer have the detail that would enable you to figure out which fact column is the source of the problem.

cmgosnell (approved):
I'm smashing approve because my two comments are not blocking and this is clearly a great improvement!!

src/pudl/transform/eiaaeo.py
jdangerx: Now that you point this out, I definitely ran into this but since I was deep in the guts of the calculation already I just... looked. I'll throw a TODO in here to keep this note somewhere and maybe we'll figure out a better pattern later.

src/pudl/transform/eiaaeo.py
```diff
@@ -178,6 +241,41 @@ def _decumulate(df: pd.DataFrame, decumulate_columns: list[str]) -> pd.DataFrame
...
+            0.999 < ratio_close_reported_calculated <= 1.0
+        ), f"reported vs. calculated totals: {ratio_close_reported_calculated}"
+
+        totals = unstacked.xs("total", level="dimension")
+        ratio_close_additions_to_total = series_sum_ratio(
+            summands=totals.loc[
+                :, ["cumulative_planned_additions", "cumulative_unplanned_additions"]
+            ],
+            total=totals.cumulative_total_additions,
+        )
```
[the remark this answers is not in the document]
jdangerx: I thought about that - since we don't actually know what the future call sites look like I leaned towards keeping this thing generic. Once we start seeing "oh, this setup is a pain" we can easily pull that into `series_sum_ratio` - but IME it's harder to say "oh, actually we want different setup here vs. in the other place, now we have to refactor `series_sum_ratio`."

Issue #3612, opened 2024-05-06
Issues setting up PUDL dev environment


denimalpaca:
### Describe the issues
Here's a list of issues I had setting up the development environment from [this guide](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/dev_setup.html), and why I think I had them:
- Using miniforge to install mamba did not result in a conda environment with python=3.12, which is needed for certain PUDL packages

In [Running the ETL Pipeline](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/run_the_etl.html) doc:
- Running the alembic command errored when it needed to run a PUDL file; there is no indication in the doc that PUDL has to be installed to the local python env as a package (how I fixed the issue) or where/how to run the command if it is not installed.

When I ran the `ferc_to_sqlite_fast` DAG, I got the following task failure/error:
```
dagster._core.errors.DagsterExecutionStepExecutionError: Error occurred while executing op "ferc1_xbrl":

  File "/Users/denimalpaca/miniforge3/lib/python3.12/site-packages/dagster/_core/execution/plan/execute_plan.py", line 282, in dagster_event_sequence_for_step
    for step_event in check.generator(step_events):
  File "/Users/denimalpaca/miniforge3/lib/python3.12/site-packages/dagster/_core/execution/plan/execute_step.py", line 523, in core_dagster_event_sequence_for_step
    for user_event in _step_output_error_checked_user_event_sequence(
  File "/Users/denimalpaca/miniforge3/lib/python3.12/site-packages/dagster/_core/execution/plan/execute_step.py", line 202, in _step_output_error_checked_user_event_sequence
    for user_event in user_event_sequence:
  File "/Users/denimalpaca/miniforge3/lib/python3.12/site-packages/dagster/_core/execution/plan/execute_step.py", line 100, in _process_asset_results_to_events
    for user_event in user_event_sequence:
  File "/Users/denimalpaca/miniforge3/lib/python3.12/site-packages/dagster/_core/execution/plan/compute.py", line 208, in execute_core_compute
    for step_output in _yield_compute_results(step_context, inputs, compute_fn, compute_context):
  File "/Users/denimalpaca/miniforge3/lib/python3.12/site-packages/dagster/_core/execution/plan/compute.py", line 177, in _yield_compute_results
    for event in iterate_with_context(
[...]

PR #3622, opened 2024-05-07
Always clobber existing outputs in FERC to SQLite conversions

zaneselvans:
# Overview

Always clobber the FERC SQLite DBs when re-running the `ferc_to_sqlite` jobs, and remove the `--clobber` option from the CLI. Partly in response to confusion that came up in #3612 

# Testing

* With existing FERC databases (and JSON outputs) I re-ran both the fast and full FERC to SQLite jobs in the Dagster UI.
* Also checked that the CLI using `ferc_to_sqlite src/pudl/package_data/settings/etl_full.yml` with existing outputs, and it ran with no complaints.

```[tasklist]
# To-do list
- [x] Ensure docs build, unit & integration tests, and test coverage pass locally with `make pytest-coverage`
```


docs/dev/run_the_etl.rst
```diff
@@ -222,19 +222,7 @@ the job. The bottom part of the window contains dagster logs. You can
 view logs from the ``pudl`` package in the CLI window the ``dagster-webserver`` process
 is running in.
 
-If you need to set op configurations, such as the ``clobber`` setting, you can
-add them in the Launchpad tab of the job like so::
```
zaneselvans, line 226: Is there another example that we should put in here instead?

docs/dev/run_the_etl.rst
```diff
@@ -471,15 +459,8 @@ what the script is doing, and file outputs the directory you specified via the
...
 
-If you need to re-run ``ferc_to_sqlite`` and want to overwrite their previous outputs
-you can add ``--clobber`` (run ``ferc_to_sqlite --clobber``).  All of the PUDL scripts
-also have help messages if you want additional information (run ``script_name --help``).
-
-.. note::
-
-  The ``pudl_etl`` command does not have a ``--clobber`` option because each etl run
-  uses the same database file to read and write tables.  This facilitates re-running
-  small portions of the ETL using Dagster.
```
zaneselvans, line 482: This no longer seemed necessary once the other discussion of clobber was gone.

test/unit/extract/xbrl_test.py
```diff
@@ -135,68 +134,6 @@ def test_xbrl2sqlite(settings, forms, mocker, tmp_path):
...
-                xbrl_num_workers=10,
-            ),
-        },
-    )
-
-    assert ferc1_sqlite_path.exists()
-    xbrl2sqlite_op_factory(XbrlFormNumber.FORM1)(context)
-    assert not ferc1_sqlite_path.exists()
-
-
```
zaneselvans, line 199: My interpretation of these is tests is that they were only checking the clobber/no-clobber behavior, which should no longer be necessary.

docs/dev/dev_setup.rst
```diff
@@ -44,6 +44,13 @@ with the following commands:
     $ mamba update mamba
     $ conda config --set channel_priority strict
 
+.. note::
+
+   Make sure that after you've installed ``conda`` or ``mamba`` that you can activate
+   and deactivate the ``base`` environment (which is created by default). Try creating
+   a simple new environment and activating it to make sure everything is working and
+   your shell has been correctly configured to use the new package manager.
+
```
zaneselvans, line 53: @denimalpaca would this have helped avoid the conda environment initialization problem you ran into?

test/unit/extract/xbrl_test.py
jdangerx: Yeah, and they also occasionally screwed up nightly builds for complicated reasons. 🔥

docs/dev/dev_setup.rst
```diff
@@ -97,11 +104,12 @@ Makefiles work, check out `this excellent Makefile tutorial
 <https://makefiletutorial.com/>`__
 
 To create the ``pudl-dev`` environment and install the local PUDL package using
-``make``, run:
+``make``, and then activate the newly created environment run:
```
jdangerx, line 107: Non-blocking: something like this reads a little better to me.

```
To create the ``pudl-dev`` environment, install the local PUDL package, and activate the newly created environment, run:
```

docs/dev/dev_setup.rst
```diff
@@ -245,7 +253,10 @@ stuff should go. We call this a "PUDL workspace".
 
 First, create a directory to store local caches of raw PUDL data. You can put this
 anywhere, but we put this in ``~/pudl_input`` in the documentation.  Then create an
-environment variable called ``PUDL_INPUT`` to store the path to this new directory:
+environment variable called ``PUDL_INPUT`` to store the path to this new directory and
+make sure that it is set whenever you start up a new shell. These shorthand commands
+will append a line to the end of your shell initialization file. If you need to change
+it later you'll want to edit those files directly:
```
jdangerx, line 259: Non-blocking: while we're here, we can tell them 

```
Please replace ``/absolute/path/to/pudl_input`` with the actual path to the directory.
```

I've answered several environment questions from people who've copy-pasted without noticing that that's a fake path.

docs/dev/run_the_etl.rst
jdangerx: I feel like the only times I ever used this were for `clobber` - nothing comes to mind.

docs/release_notes.rst
```diff
@@ -126,6 +126,13 @@ Bug Fixes
   used on any dataframe, with every column encoded based on the field definitions and
   FK constraints associated with the column name.
 
+CLI Changes
+^^^^^^^^^^^
+* Removed the ``--clobber`` option from the ``ferc_to_sqlite`` command and associated
+  assets. We rebuild these databases infrequently, and needing either edit the runtime
```
jdangerx, line 132:
```suggestion
  assets. We rebuild these databases infrequently, and needing to either edit the runtime
```

jdangerx (approved):
Seems good, just a couple small suggestions for the docs that are non blocking.

PR #3625, opened 2024-05-07
Q1 2024 eia860m eia923

### LINKED ISSUE #3617: Q1-2024 Updates to EIA860m

Based on the [Existing Data Updates Docs](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html)

- [x] [Obtain fresh data](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#obtain-fresh-data)
- [x] [Map the structure of the new data](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#map-the-structure-of-the-new-data)
- [x] [Test data extraction](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#test-data-extraction)
- [x] [Update table and column transformations](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-table-column-transformations)
- [x] [Update the PUDL db schema](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-pudl-db-schema)
- [x] [Connect datasets](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#connect-datasets)
- [x] [Run the ETL](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#run-the-etl)
- [x] [Update the output routines and run full tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-output-routines-and-run-full-tests)
- [ ] [Run and update data validations](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#run-and-update-data-validations)
- [ ] [Update the documentation](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-documentation)

### LINKED ISSUE #3618: Q1-2024 Updates to EIA923

Based on the [Existing Data Updates Docs](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html)

- [ ] [Obtain fresh data](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#obtain-fresh-data)
- [ ] [Map the structure of the new data](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#map-the-structure-of-the-new-data)
- [ ] [Test data extraction](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#test-data-extraction)
- [ ] [Update table and column transformations](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-table-column-transformations)
- [ ] [Update the PUDL db schema](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-pudl-db-schema)
- [ ] [Connect datasets](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#connect-datasets)
- [ ] [Run the ETL](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#run-the-etl)
- [ ] [Update the output routines and run full tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-output-routines-and-run-full-tests)
- [ ] [Run and update data validations](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#run-and-update-data-validations)
- [ ] [Update the documentation](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-documentation)

aesharpe:
# Overview

Closes #3617
Closes #3618

Adds Q1 2024 data for EIA860m and EIA923

**What did you change?**
Added new DOIs, partitions, and package_data mapping

**What's still left to do**
There is an issue with the 860m data integration. We normally have EIA860 data from at least one year prior to the EIA860m data. Right now we do not have that. Early Release EIA860 data comes out in June. This means that we are relying on EIA860m for both 2023 and 2024 data. This is causing some issues. For example, 2023 data is getting dropped from the bf_eia923 table.

Another issue is the Eia860Settings function check_eia860m_year_month. I had to comment this out to get the dagster assets to load.

# Testing

```[tasklist]
# To-do list
```


src/pudl/settings.py
```diff
@@ -248,36 +248,24 @@ class Eia860Settings(GenericDatasetSettings):
...
-        if eia860m and (eia860m_year != expected_year):
-            raise AssertionError(
-                f"Attempting to integrate an eia860m year "
-                f"({eia860m_year}) from {cls.eia860m_year_month} not immediately following"
-                f"the eia860 years: {cls.data_source.working_partitions['years']}. "
-                f"Consider switching eia860m parameter to False."
-            )
-        return eia860m
+    @property
+    def eia860m_year_month(self) -> list[str]:
```
cmgosnell, line 254: two things:
- because this can now be a list of `year_months` I would suggest making this plural!
- this is not necessary/blocking at all but it'd be nice to use the `eia860m` bool to set or not this guy. We use the bool in the extract step later but it'd be nice to just wipe the `year_months` if we are not doing 860m at all

test/unit/settings_test.py
```diff
@@ -131,15 +131,15 @@ class TestEIA860Settings:
...
-        settings_cls.eia860m_year_month = original_eia80m_year_month
+    # def test_860m(self: Self):
+    #     """Test validation error is raised when eia860m date is within 860 years."""
+    #     settings_cls = Eia860Settings
+    #     original_eia80m_year_month = settings_cls.eia860m_year_month
+    #     settings_cls.eia860m_year_month = ["2019-11"]
+
+    #     with pytest.raises(ValidationError):
+    #         settings_cls(eia860m=True)
+    #     settings_cls.eia860m_year_month = original_eia80m_year_month
```
cmgosnell, line 142: I am nearly sure you needed to remove this because you removed the ability to set the `eia860m_year_month` on initialization as i commented about above. If you add that back in you could keep this test and even add another that checks whether or not you new "add extra in between 860 and 860m years" functionality.

cmgosnell (changes requested):
it looks like this is definitely going in the right direction! I have some suggestions mostly in the settings class!

(i haven't looked at any of the column maps yet but i wanted to give you an early review on the settings cram)

src/pudl/settings.py
```diff
@@ -248,36 +248,24 @@ class Eia860Settings(GenericDatasetSettings):
 
     data_source: ClassVar[DataSource] = DataSource.from_id("eia860")
     years: list[int] = data_source.working_partitions["years"]
-
     eia860m: bool = True
-    eia860m_year_month: ClassVar[str] = max(
-        DataSource.from_id("eia860m").working_partitions["year_months"]
-    )
```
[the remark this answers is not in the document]
aesharpe: I tried something like that before, and it didn't seem like I could add anything to the `eia860m_year_months` value, but I will look into that.

test/validate/eia_test.py
```diff
@@ -46,17 +46,17 @@ def test_no_null_cols_eia(pudl_out_eia, live_dbs, cols, df_name):
...
+        ("bf_eia923", 1_626_501, 1_626_501, 134_619),
         ("bga_eia860", 142_391, 142_391, 142_391),
-        ("boil_eia860", 83_418, 83_418, 83_418),
-        ("frc_eia923", 656_388, 268_352, 25_443),
-        ("gen_eia923", None, 5_179_478, 433_336),
-        ("gens_eia860", 557_369, 557_369, 557_369),
-        ("gf_eia923", 2_933_836, 2_933_836, 246_324),
+        ("boil_eia860", 86_675, 86_675, 86_675),
+        ("frc_eia923", 661_976, 270_947, 26_470),
+        ("gen_eia923", None, 5_179_478, 433_327),
```
zaneselvans, line 53: Do we know what's up with `gen_eia923`? Why would it have no change to the monthly rows, and lose 9 annual rows?

src/pudl/settings.py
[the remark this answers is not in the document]
cmgosnell: i think pydantic is not super intuitive with these things so if you wanna pair about it lmk!

src/pudl/settings.py
```diff
@@ -243,41 +244,32 @@ class Eia860Settings(GenericDatasetSettings):
...
+            years_eia860m = {date.split("-")[0] for date in all_eia860m_dates}
+            extra_years_eia860m = [
+                year for year in years_eia860m if int(year) not in info.data["years"]
+            ]
+            eia860m_year_months = [
+                max(date for date in all_eia860m_dates if date.startswith(year))
+                for year in extra_years_eia860m
+            ]
+            return eia860m_year_months
+        return v
```
cmgosnell, line 272: I think the way this is setup no matter what you feed into the settings class the same list of year_months will be spit out. which is not ideal.

imo the right conceptual process for this would be very similar to what you are doing but instead of always assigning the same `eia860m_year_months`, you'd check what years are in `v` (which is whatever you assign `eia860m_year_months` when instantiating the class). If there are any gap years between the last `info.data["years"]` and the years in `v`, then add the max 860m date.

test/validate/eia_test.py
aesharpe: yeah that was indeed a little weird. I didn't look into it because in the past we have overlooked these tiny row fluctuations. If you think it's worth a poke I will dig more!

src/pudl/settings.py
```diff
@@ -243,41 +244,71 @@ class Eia860Settings(GenericDatasetSettings):
...
     data_source: ClassVar[DataSource] = DataSource.from_id("eia860")
     years: list[int] = data_source.working_partitions["years"]
-
     eia860m: bool = True
-    eia860m_year_month: ClassVar[str] = max(
-        DataSource.from_id("eia860m").working_partitions["year_months"]
-    )
+    all_eia860m_year_months: list[str] = DataSource.from_id(
+        "eia860m"
+    ).working_partitions["year_months"]
```
aesharpe, line 255: I tried making `all_eia860m_year_months` a private variable by adding an `_` in front (because it shouldn't ever be changed/altered/used as anything but a reference within the class functions). However, when I added the underscore, I could no longer reference it in the class functions (ex: `info.data["_all_eia860m_year_months]`). So for now I removed the `_`. If anyone knows how to do this successfully lmk!

test/unit/settings_test.py
```diff
@@ -126,20 +126,17 @@ def test_none_quarters_raise(self: Self):
...
-        settings_cls.eia860m_year_month = "2019-11"
+        settings_cls = Eia860Settings()
+        original_eia80m_year_month = settings_cls.eia860m_year_months
+        settings_cls.eia860m_year_months = ["2019-11"]
 
-        with pytest.raises(ValidationError):
+        with pytest.raises(ValueError):
             settings_cls(eia860m=True)
-        settings_cls.eia860m_year_month = original_eia80m_year_month
+        settings_cls.eia860m_year_months = original_eia80m_year_month
```
aesharpe, line 139: For some reason when I run this test `pytest test/unit/settings_test.py::TestEIA860Settings::test_860m` I get the following error: 

```
name = 'eia860m_year_months', value = ['2019-11']

    def _check_frozen(self, name: str, value: Any) -> None:
        if self.model_config.get('frozen', None):
            typ = 'frozen_instance'
        elif getattr(self.model_fields.get(name), 'frozen', False):
            typ = 'frozen_field'
        else:
            return
        error: pydantic_core.InitErrorDetails = {
            'type': typ,
            'loc': (name,),
            'input': value,
        }
>       raise pydantic_core.ValidationError.from_exception_data(self.__class__.__name__, [error])
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for Eia860Settings
E       eia860m_year_months
E         Instance is frozen [type=frozen_instance, input_value=['2019-11'], input_type=list]
E           For further information visit https://errors.pydantic.dev/2.7/v/frozen_instance

../../../mambaforge/envs/pudl-dev/lib/python3.12/site-packages/pydantic/main.py:894: ValidationError
```

I don't know why this error wasn't happening before -- the test was still replacing the class variable `eia860m_year_month`. The only difference was that it was a single string value rather than a list of strings.

aesharpe: Oh also, I changed it from a `Validation` to a `Value` error because that's what error the `only_years_not_in_eia860` function should throw. 

...on a separate note, I'd love to better understand the difference between validation and value errors

test/unit/settings_test.py
aesharpe: Tried restoring this and ran into issues, see [below](https://github.com/catalyst-cooperative/pudl/pull/3625/files#r1596100838)

test/validate/eia_test.py
zaneselvans: I was more wondering why it isn't changing much more and in both the yearly and monthly cases, in line with the ~5% difference that's seen in the other tables.

src/pudl/settings.py
zaneselvans: Our Pydantic settings classes are immutable, meaning that none of their attributes can be changed once the class has been instantiated. This is something Pydantic implements, not Python (Python doesn't natively let you make class attributes private, really). I suspect that Pydantic is taking the "private" hint from the leading underscore and purposefully not making that attribute available in the dictionary of values in the validator.

test/unit/settings_test.py
zaneselvans: `ValidationError` is [defined by Pydantic](https://docs.pydantic.dev/latest/errors/validation_errors/), while `ValueError` is [a built-in error type](https://docs.python.org/3/library/exceptions.html#ValueError) that's part of Python:

> Raised when an operation or function receives an argument that has the right type but an inappropriate value

Pydantic gathers together information from all of the different validation failures and tries to present it together for use in debugging. [Pydantic says](https://docs.pydantic.dev/latest/concepts/validators/#handling-errors-in-validators) validators should throw either `ValueError` or `AssertionError` when they fail, so I assume the validation machinery is specifically trying to catch those exceptions in compiling the more extensive `ValidationError` output.

cmgosnell: @aesharpe  I will test this locally but I believe this is happening because you are instantiating the object in this first line (`settings_cls = Eia860Settings()`) when it was previously `settings_cls = Eia860Settings`

src/pudl/settings.py
```diff
@@ -243,41 +244,82 @@ class Eia860Settings(GenericDatasetSettings):
...
-            raise AssertionError(
-                f"Attempting to integrate an eia860m year "
-                f"({eia860m_year}) from {cls.eia860m_year_month} not immediately following"
-                f"the eia860 years: {cls.data_source.working_partitions['years']}. "
-                f"Consider switching eia860m parameter to False."
-            )
-        return eia860m
+        if info.data["eia860m"]:
+            all_eia860m_years = {
+                date.split("-")[0] for date in info.data["all_eia860m_year_months"]
```
zaneselvans, line 271: Rather than doing this string-splitting to extract the year in multiple places, might it make sense to parse the values once when the object is instantiated? We might also want to impose a regex constraint on the string to ensure that it's guaranteed to be a valid `YYYY-MM` date.

src/pudl/settings.py
```diff
@@ -243,41 +244,82 @@ class Eia860Settings(GenericDatasetSettings):
...
+            extra_eia860m_year_months = [
+                max(
+                    date
+                    for date in info.data["all_eia860m_year_months"]
+                    if date.startswith(year)
+                )
+                for year in (extra_eia860m_years - years_in_v)
+            ]
+            return v + extra_eia860m_year_months
+        return v
```
zaneselvans, line 293: To avoid having multiple return statements, could you always return
```
v + extra_eia860m_year_months
```
by initializing `extra_eia860m_year_months` to be an empty list?

src/pudl/settings.py
```diff
@@ -243,41 +244,82 @@ class Eia860Settings(GenericDatasetSettings):
...
+        if info.data["eia860m"]:
+            all_eia860m_years = {
+                date.split("-")[0] for date in info.data["all_eia860m_year_months"]
+            }
+            # The years in 860m that are not in 860
+            extra_eia860m_years = {
+                year
+                for year in all_eia860m_years
+                if int(year) not in info.data["years"]
+            }
```
zaneselvans, line 278: Something I think this might permit, but which we probably wouldn't want is if someone put in an early EIA-860M year-month, and then only processed a single year of annual EIA-860 data from after that early EIA-860M date, I think it would try to add in EIA-860M dates from before the included EIA-860 data, as well as after it.

test/unit/settings_test.py
cmgosnell: but if you don't instantiate the class you no longer get access to the default `eia860m_year_months`. This attribute was a `ClassVar` it was accessible w/o instantiating the class. But you can still grab the default value is you just instantiate a version of the class. i thiiink to get this tests as is to run properly you'll need to do something like this:

```python
settings_cls = Eia860Settings
original_eia80m_year_month = Eia860Settings().eia860m_year_months
settings_cls.eia860m_year_months = ["2019-11"]

with pytest.raises(ValueError):
    settings_cls(eia860m=True)
settings_cls.eia860m_year_months = original_eia80m_year_month
```

buuuutttt this test is testing the previous classes functionality. its testing the previous field validation of the `eia860m` boolean. Because you have added several good new validations, it would be good to add lil tests to those failure modes.

cmgosnell: I would add something like this:

```python
class TestEIA860Settings:
    """Test EIA860 setting validation."""
    def test_860m_year_is_one_more_than_860_max_year(self: Self):
        """Test whether or not the first EIA860m year is the next year after the max 860 year."""
        settings_eia860= Eia860Settings(eia860m_year_months=["2024-01"])
        assert max(settings_eia860.years)+1 == int(min(settings_eia860.eia860m_year_months).split("-")[0])
```

cmgosnell: also something weird is happening bc:

```python
settings_eia860= Eia860Settings(years=[2022], eia860m_year_months=["2019-01"])
min(pd.to_datetime(settings_eia860.eia860m_year_months).year)
>> 2015
```

src/pudl/settings.py
cmgosnell: I think it would be much more straightforward to use the date tools our there:

```python
all_eia860m_years = pd.to_datetime(info.data["all_eia860m_year_months"]).year
```

test/unit/settings_test.py
aesharpe: > ```python
> settings_cls = Eia860Settings
> original_eia80m_year_month = Eia860Settings().eia860m_year_months
> settings_cls.eia860m_year_months = ["2019-11"]
> 
> with pytest.raises(ValueError):
>     settings_cls(eia860m=True)
> settings_cls.eia860m_year_months = original_eia80m_year_month
> ```

I made this suggested edit, and it doesn't seem to be working. The test fails because it doesn't raise a `ValueError`. I added a `breakpoint()` to the `only_years_not_in_eia860()` function in `Eia860Settings` that it should be testing, and it doesn't stop when I run the tests. This leads me to believe that the `only_years_not_in_eia860` function isn't getting called in the test?

aesharpe: I think the problem is that you can't have `field_validators` for `ClassVar`s??

zaneselvans: One thing that's happening here is you can't assign a new value to `settings_cls.eia860m_year_months` because the settings class is immutable.  That assignment by itself is causing an error:

```py
from pudl.settings import Eia860Settings
settings_cls = Eia860Settings()
settings_cls.eia860m_year_months = ["2019-01"]
```

I think what you're trying to test here is that if we attempt to construct `Eia860Settings` where both the EIA860 and the EIA860M dates include the same year, a `ValidationError` will be raised. I think that might look like:

```py
    def test_eia860_years_overlap_eia860m_years(self: Self):
        """Test validation error is raised when eia860m date is within eia860 years."""
        # Identify the last valid EIA-860 year:
        max_eia860_year = max(Eia860Settings().years)
        # Use that year to construct an EIA-860M year that overlaps the EIA-860 years:
        bad_eia860m_year_month = f"{max_eia860_year}-01"

        # Attempt to construct an EIA-860 settings object with an EIA-860M year that
        # overlaps the EIA-860 years, which should result in a ValidationError:
        with pytest.raises(ValidationError):
            _ = Eia860Settings(
                eia860m=True,
                years=[max_eia860_year],
                eia860m_year_months=[bad_eia860m_year_month],
            )
```

src/pudl/settings.py
zaneselvans: In playing with the example test for overlapping years in the EIA-860M and EIA-860 settings, this bug was exercised -- it added **all** of the EIA-860M years before the test EIA-860 year, as well as after.

test/unit/settings_test.py
zaneselvans: This test as I've written it doesn't currently fail because of the missing return value that Christina highlighted in one of the validators. (`eia860m_year_months` is always `None` no matter what).  If you fix that return value, it still doesn't fail, and the settings object that's created looks like:

```py
Eia860Settings(
    disabled=False,
    years=[2022],
    eia860m=True,
    all_eia860m_year_months=[
        "2015-07",
        "2015-08",
        "2015-09",
        "2015-10",
        "2015-11",
        "2015-12",
        # etc. ...
        "2023-10",
        "2023-11",
        "2023-12",
        "2024-01",
        "2024-02",
        "2024-03",
    ],
    eia860m_year_months=[
        "2022-01",
        "2020-12",
        "2019-12",
        "2018-12",
        "2021-12",
        "2023-12",
        "2024-03",
        "2016-12",
        "2015-12",
        "2017-12",
    ],
)
```

src/pudl/settings.py
```diff
@@ -243,41 +244,82 @@ class Eia860Settings(GenericDatasetSettings):
...
+        "eia860m"
+    ).working_partitions["year_months"]
+    eia860m_year_months: list[str] = [max(all_eia860m_year_months)]
 
-    @field_validator("eia860m")
+    @field_validator("eia860m_year_months")
     @classmethod
-    def check_eia860m_year_month(cls, eia860m: bool) -> bool:
-        """Check 860m date-year is exactly one year after most recent working 860 year.
+    def add_other_860m_years(cls, v, info: FieldValidationInfo) -> list[str]:
```
zaneselvans, line 260: I think this might do the right thing:

```py
    @field_validator("eia860m_year_months")
    @classmethod
    def add_other_860m_years(cls, v, info: FieldValidationInfo) -> list[str]:
        """Find extra years from EIA860m if applicable.

        There's a gap in reporting (after the new year but before the EIA 860 early
        release data in June) when we rely on two years worth of EIA860m data instead
        of just one. This function adds the last available month_year values for each
        year of 860m data that is not yet available in 860.

        """
        extra_eia860m_year_months = []
        if info.data["eia860m"]:
            all_eia860m_years = set(
                pd.to_datetime(info.data["all_eia860m_year_months"]).year.unique()
            )
            # The years in 860m that are not in 860
            extra_eia860m_years = {
                year for year in all_eia860m_years if year not in info.data["years"]
            }
            # The years already listed as variables in eia860m_year_months
            years_in_v = set(pd.to_datetime(v).year.unique())
            # The max year_month values available in 860m for each year not
            # covered by EIA860 (and not already listed in the eia860m_year_months
            # variable)
            extra_eia860m_year_months = [
                max(
                    date
                    for date in pd.to_datetime(info.data["all_eia860m_year_months"])
                    if date.year == year
                )
                for year in (extra_eia860m_years - years_in_v)
                if year > max(info.data["years"])
            ]
        return v + list(pd.Series(extra_eia860m_year_months).dt.strftime("%Y-%m"))
```

zaneselvans: But I'm not sure what the expected / desired behavior is for adding "extra" EIA-860M years. I think the **only** years we ever want to populated withe EIA-860M data are the ones that are not available from EIA-860 at all, rather than just any that weren't included in the arguments to the settings.

test/validate/eia_test.py
zaneselvans: I think what we want to pay attention to is how big the deviation is from the expected value. Often we don't expect to see any change, so a tiny change is fine. But when we expect to see a significant change (5% more records) and we see no change or a tiny change, that seems suspicious.

zaneselvans: Here we don't expect the BGA or ownership tables or emissions equipment tables from EIA-860 to change (and they don't), because they only draw from the full EIA-860 data, but I kind of do expect `gen_eia923` to change, since we're adding several new months of EIA-923 data.

src/pudl/settings.py
```diff
@@ -243,41 +244,84 @@ class Eia860Settings(GenericDatasetSettings):
...
+        eia860m_year_months ClassVar[str]: The 860m year-months to incorporate.
     """
 
     data_source: ClassVar[DataSource] = DataSource.from_id("eia860")
     years: list[int] = data_source.working_partitions["years"]
+    eia860m: bool = Field(validate_default=True, default=True)
+    all_eia860m_year_months: list[str] = DataSource.from_id(
+        "eia860m"
+    ).working_partitions["year_months"]
+    eia860m_year_months: list[str] = Field(validate_default=True, default=[])
```
aesharpe, line 256: I made this default value empty again because it felt weird to have a non-empty default option for `eia860_year_months` when you could specify `eia860m = False`.

src/pudl/settings.py
```diff
@@ -243,41 +244,84 @@ class Eia860Settings(GenericDatasetSettings):
...
+            all_eia860m_years = set(
+                pd.to_datetime(info.data["all_eia860m_year_months"]).year
             )
-        return eia860m
+            # The years in 860m that are not in 860
+            extra_eia860m_years = {
+                year
+                for year in all_eia860m_years
+                if year not in DataSource.from_id("eia860").working_partitions["years"]
+            }
```
aesharpe, line 279: I changed this from `info.data["years"]` so that it only populates the `extra_eia860m_years` list with years that aren't available in 860 (even if they aren't specified in `years`)

test/unit/settings_test.py
```diff
@@ -145,6 +145,26 @@ def test_eia860_years_overlap_eia860m_years(self: Self):
...
             )
 
+    def test_eia860m_years_overlap_eia860m_years(self: Self):
+        """Test validation error is raised when eia860m years overlap."""
+        max_eia860_year = max(Eia860Settings().years)
+        acceptable_eia860m_year = max_eia860_year + 1
+        bad_eia860m_year_months = [
+            f"{acceptable_eia860m_year}-01",
+            f"{acceptable_eia860m_year}-05",
+        ]
```
zaneselvans, line 156: I think there will be times when these inputs fail for the wrong reason, not because they overlap, but because there's no EIA-860M data available for May yet in the `max(eia860.years) + 1` year.  If you use Jan and Feb I think you won't have that problem, since if the new "early release" data for EIA-860 for 2023 comes out in June, there *should* be Jan and Feb 2024 data available by then.

test/unit/settings_test.py
```diff
@@ -145,6 +145,26 @@ def test_eia860_years_overlap_eia860m_years(self: Self):
...
+        with pytest.raises(ValidationError):
+            _ = Eia860Settings(
+                eia860m=True,
+                years=[max_eia860_year],
+                eia860m_year_months=bad_eia860m_year_months,
+            )
+
+    def test_eia860m(self: Self):
+        """Test creation of eia860m_year_month values when eia860m is True."""
+        eia860_settings = Eia860Settings()
```
zaneselvans, line 182: I think it's more robust and readable to explicitly set `eia860m=True` in the arguments here, rather than relying implicitly on the default value.

test/validate/eia_test.py
aesharpe: I tried digging into this, but I can't seem to get any answers. It might be helpful to pair on this?

src/pudl/settings.py
zaneselvans: So the default is `[]` but if `eia860m=True` and the default value is used, the correct final validated value for `eia860m_year_months` will be dynamically determined by the validator below?

test/unit/settings_test.py
zaneselvans, line 182:
```suggestion
        eia860_settings = Eia860Settings(eia860m=True)
```

test/unit/settings_test.py
```diff
@@ -145,6 +145,26 @@ def test_eia860_years_overlap_eia860m_years(self: Self):
                 eia860m_year_months=[bad_eia860m_year_month],
             )
 
+    def test_eia860m_years_overlap_eia860m_years(self: Self):
+        """Test validation error is raised when eia860m years overlap."""
+        max_eia860_year = max(Eia860Settings().years)
+        acceptable_eia860m_year = max_eia860_year + 1
+        bad_eia860m_year_months = [
+            f"{acceptable_eia860m_year}-01",
+            f"{acceptable_eia860m_year}-05",
```
zaneselvans, line 155:
```suggestion
            f"{acceptable_eia860m_year}-02",
```

zaneselvans (changes requested):

src/pudl/settings.py
aesharpe: Yes, and the `settings_test` `test_eia860m_after_eia860` function should verify that

test/validate/eia_test.py
zaneselvans: Looking at the latest dates that show up in the tables which can be returned by `PudlTabl.gen_eia923()`

```py
gen_eia923_ms = pd.read_sql("out_eia923__monthly_generation", pudl_engine)
gen_eia923_ys = pd.read_sql("out_eia923__yearly_generation", pudl_engine)
gf_by_gen_eia923_ms = pd.read_sql("out_eia923__monthly_generation_fuel_by_generator", pudl_engine)
gf_by_gen_eia923_ys = pd.read_sql("out_eia923__yearly_generation_fuel_by_generator", pudl_engine)
frc_eia923 = pd.read_sql("out_eia923__monthly_fuel_receipts_costs", pudl_engine)

print(f"gen MS: {gen_eia923_ms.report_date.max()}")
print(f"gen YS: {gen_eia923_ys.report_date.max()}")
print(f"gen fuel by gen MS: {gf_by_gen_eia923_ms.report_date.max()}")
print(f"gen fuel by gen YS: {gf_by_gen_eia923_ys.report_date.max()}")
print(f"frc MS: {frc_eia923.report_date.max()}")

# gen MS: 2023-12-01 00:00:00
# gen YS: 2022-01-01 00:00:00
# gen fuel by gen MS: 2022-12-01 00:00:00
# gen fuel by gen YS: 2022-01-01 00:00:00
# frc MS: 2023-11-01 00:00:00
```

zaneselvans (approved):

PR #3628, opened 2024-05-11
Tweak eia860m settings validations and tests.

zaneselvans:
The comments and suggestions on #3625 were getting a bit hard to disentangle, so here is a little PR that can be compared more legibly maybe.

This fixes the failing unit test and some other minor issues, but I think there are still issues to be resolved in the validation logic and additional tests which should be added to make sure this somewhat complicated setting is really doing what we want it to be doing.

### DIFF
--- a/test/unit/settings_test.py
+++ b/test/unit/settings_test.py
@@ -38,8 +38,9 @@ class TestGenericDatasetSettings:
             working_tables = ["table"]
 
             class Test(GenericDatasetSettings):
-                data_source: DataSource(
-                    working_partitions=working_partitions, working_tables=working_tables
+                data_source: DataSource = DataSource(
+                    working_partitions=working_partitions,
+                    working_tables=working_tables,
                 )
 
             Test()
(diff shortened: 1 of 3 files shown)


test/unit/settings_test.py
```diff
@@ -125,18 +126,24 @@ def test_none_quarters_raise(self: Self):
...
 
-class TestEIA860Settings:
+class TestEia860Settings:
     """Test EIA860 setting validation."""
 
-    def test_860m(self: Self):
-        """Test validation error is raised when eia860m date is within 860 years."""
-        settings_cls = Eia860Settings()
-        original_eia80m_year_month = settings_cls.eia860m_year_months
-        settings_cls.eia860m_year_months = ["2019-11"]
```
zaneselvans, line 135: This assignment is invalid, because the class is immutable.  It was "working" before because this was `ClassVar` but it was actually bad -- the assignment was clobbering the value built into the class.

A simpler / less invasive way to check that we get an error when there's overlap between the EIA-860 and EIA-860M years is to explicitly give the class constructor dates which we know will clash, as is happening below now.

PR #3630, opened 2024-05-13
NREL ATB - Stop dropping duplicate values before unstacking

cmgosnell:
# Overview

make sure drop duplicates includes the idx values!

GridPath Ana mentioned that there were some columns in the `core_nrelatb__yearly_projected_cost_performance` table that looked much too sparse. ([example here w/ `heat_rate_mmbtu_per_mwh`](https://data.catalyst.coop/pudl?sql=select+*%0D%0Afrom+core_nrelatb__yearly_projected_cost_performance%0D%0Awhere+technology_description_detail_2+like+%27%25CT%25%27))

I added two kinds of asset_check on the `core_nrelatb__yearly_projected_cost_performance` asset. One for total nulls in the table and one outlining specific columns that only have data in them for some `technology_description`s. This setup could be generalized to the other two tables that get the `unstack` treatment but this table continues to be the most null-filled so i started here.

I also renamed `fuel_cost` to `fuel_cost_per_mwh` per @zaneselvans 's comment below

# Testing

I re-ran all of the nrel atb assets and checked several examples like the `heat_rate_mmbtu_per_mwh` example Ana gave

I also figured out how to actually run the asset_checks (ty @jdangerx 😎 )

I re-ran the atb assets for both the full and fast etl

```[tasklist]
# To-do list
```


src/pudl/transform/nrelatb.py
```diff
@@ -121,12 +121,13 @@ def transform_unstack(
...
-            nrelatb.core_metric_parameter.isin(table_unstacker.core_metric_parameters)
+        nrelatb.loc[
+            nrelatb.core_metric_parameter.isin(table_unstacker.core_metric_parameters),
+            table_unstacker.idx + ["value"],
         ]
-        .set_index(table_unstacker.idx)
-        .sort_index()[["value"]]
         .drop_duplicates()
+        .set_index(table_unstacker.idx)
+        .sort_index()
```
cmgosnell, line 130: i was previously applying the `drop_duplicates` after the setting of the index so it was dropping any duplicate `value`. so now i'm selecting only the idx + value columns first, then drop_dup-ing, then index-ing.

jdangerx: Ahh makes sense. Should we have some sort of "there's X% non-null values" check for this table?

jdangerx (approved):
LGTM and makes sense - it might make sense to add some checks that would catch this specific type of problem before merging, but that's non-blocking.

cmgosnell:
@jdangerx  yes i total null check sounds good. i thought about it before but some of the columns in this table only have values in 2023 so i was hesitant but it would have caught this error because i added the drop_duplicates late in the process.

I will add a little check now

zaneselvans:
for these variables that really do only pertain to one or a few technology types, and should be null everywhere else, it might be good to make those expectations explicit with an `asset_check` so we and others can see that it's supposed to be that way.

cmgosnell:
> for these variables that really do only pertain to one or a few technology types, and should be null everywhere else, it might be good to make those expectations explicit with an asset_check so we and others can see that it's supposed to be that way.

i am actively adding these now

PR #3638, opened 2024-05-15
Update expected MCOE row counts

zaneselvans:
# Overview

We lost 9 rows in the annual capacity factor and MCOE outputs, which is almost certainly because we lost 9 rows in the annual `gen_eia923` outputs... which we already knew -- not sure why this didn't trip the validations up before and get included in the initial row count updates. 🤦🏼 

I also changed the `pytest-minmax-rows` make target to run on 8 CPUs instead of one to make doing this check faster and more convenient in the future. It used less than 10 GB of memory at peak, so this should not freeze anybody's computer up.

# Testing

I ran `make pytest-minmax-rows` locally and they all passed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ pytest-jupyter:
 # whose name contains "minmax_rows" so it's important to follow that naming convention.
 .PHONY: pytest-minmax-rows
 pytest-minmax-rows:
-	pytest --live-dbs test/validate -k minmax_rows
+	pytest -n 8 --no-cov --live-dbs test/validate -k minmax_rows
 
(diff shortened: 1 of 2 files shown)


cmgosnell (approved):

cmgosnell:
@aesharpe  when you were checking the validations did you see this at all? where you testing them via pytest and only doing testing a few of the validation test modules or something? i'm just curious how this slipped through when i know you did update the validations.

zaneselvans:
@cmgosnell is it possible that the changes in the MCOE tables weren't present before because the EIA plant ID mapping hadn't happened yet?

PR #3640, opened 2024-05-15
Add template that includes overview/success criteria/tasks

jdangerx:
# Overview

It's kind of annoying to keep typing the same sections over and over again when creating generic issues.

What did you change?

Added a template.

# Testing

Copied the template text into an issue and hit the 'preview' tab.

### DIFF
--- a/.github/ISSUE_TEMPLATE/task.md
+++ b/.github/ISSUE_TEMPLATE/task.md
@@ -18,9 +18,7 @@ How will we know that we're done?
 * [ ] ...
 
 
-### Tasks
-What are the next steps to take?
-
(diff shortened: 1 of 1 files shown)


aesharpe (commented):
This looks good, but I like how the other template had a testing section too. Maybe it's overkill though. Less is probably better!

aesharpe (approved):

jdangerx:
I think you're thinking of the PR template, which does and will continue to have a testing section. To some extent, the success criteria of an issue is sort of like the testing plan for a PR - both are sort of "how do we know this thing is actually working?"

PR #3652, opened 2024-05-23
Organize the large new data section of release notes

zaneselvans:
# Overview

There's so much new data in the upcoming release that the release notes section for new data really needed to be broken down by data source. Otherwise it was an unreadable blob of text with different changes all mixed together.

# Testing

I built the docs locally and they worked.


jdangerx (approved):
Thanks for cleaning this up!

PR #3656, opened 2024-05-27
Add EIA AEO fuel cost projections

### LINKED ISSUE #3649: `core_eiaaeo__yearly_projected_fuel_cost_in_electric_sector_by_type`
### Overview
A few people have asked us about this set of data from AEO table 54, so we should pull it in.

### Success criteria
* [ ] regional fuel cost projections are in the PUDL database under `core_eiaaeo__yearly_projected_fuel_cost_in_electric_sector_by_type`

```[tasklist]
### Next steps
- [x] figure out which series names correspond to this dataset
- [x] update the resource schema to include the real $ base year
- [x] extract the dang data
- [x] graph the outputs to see if they seem reasonable
- [x] write asset checks to assert things about the shape of the data
```

jdangerx:
# Overview

Closes #3649.

* promotes `core_eiaaeo__yearly_projected_fuel_cost_in_electric_sector_by_type` resource + updates fields
* adds `real_cost_basis_year` field so we know what the real cost actually means

# Testing

Materialized assets via Dagster, asset checks passed, and made some graphs as seen in #3649 .

```[tasklist]
# To-do list
- [x] If updating analyses or data processing functions: make sure to update or write data validation tests (e.g.,  `test_minmax_rows()`)
```


src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -408,8 +400,9 @@
     for key, value in _STAGING_RESOURCE_METADATA.items()
     if key
     in {
+        "core_eiaaeo__yearly_projected_electric_sales",
```
jdangerx, line 403: Just realphabetizing.

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -266,6 +266,7 @@
                 "fuel_type_eiaaeo",
                 "fuel_cost_per_mmbtu",
                 "fuel_cost_real_per_mmbtu_eiaaeo",
+                "real_cost_basis_year",
```
jdangerx, line 269: Added this so we know what year the real cost refers to.

src/pudl/metadata/resources/eiaaeo.py
```diff
@@ -342,15 +343,6 @@
         "description": "Total carbon emissions in short tons.",
         "unit": "short_tons",
     },
-    "fuel_cost_real_per_mmbtu_eiaaeo": {
```
jdangerx, line 345: moved to the actual `fields` dictionary.

src/pudl/transform/eiaaeo.py
```diff
@@ -431,52 +431,6 @@ def core_eiaaeo__yearly_projected_electric_sales(
     return renamed_for_pudl
 
 
-@dataclass
```
jdangerx, line 434: Moved these down so they're not interspersed between assets.

src/pudl/transform/eiaaeo.py
```diff
@@ -560,6 +514,114 @@ def core_eiaaeo__yearly_projected_generation_in_end_use_sectors_by_fuel_type(
     return renamed_for_pudl
 
 
+@asset(io_manager_key="pudl_io_manager")
+def core_eiaaeo__yearly_projected_fuel_cost_in_electric_sector_by_type(
+    raw_eiaaeo__electric_power_projections_regional,
+):
+    """Projected generation capacity + gross generation in end-use sectors.
```
jdangerx, line 521: TODO: fix this docstring -_-

jdangerx:
FYI @anamileva and @gschivley, two people who have been curious about this.

src/pudl/metadata/fields.py
```diff
@@ -1644,6 +1644,15 @@
         "description": "Average cost of fuel delivered in the report year per reported fuel unit (USD).",
         "unit": "USD",
     },
+    "fuel_cost_real_per_mmbtu_eiaaeo": {
```
aesharpe, line 1647: why do we need the suffix `eiaaeo`?

aesharpe: Also this is a nit and not blocking but here we say: `cost_real` but in the other column we say `real_cost`. Maybe we should pick one?

aesharpe (changes requested):
Generally looks good!

It looks like you added the resource metadata for this table in another PR? I have one comment about the table description (in `metadata/resources/eiaaeo.py`) containing the phrase: `Projected fuel prices to the electric power sector` -- what does "to the electric power sector" mean -- should this be "for"? It also doesn't quite match the description you put in the release notes that says "generation and capacity" and leaves out cost.

src/pudl/metadata/fields.py
jdangerx: We wanted to add the suffix `eiaaeo` because it's possible that other datasets mean something different by "real fuel cost" - maybe it is always based in the same year, for example. It's easier to remove a suffix and merge two fields into one than it is to have to tease apart the different use cases later.

I think this column would be named "Fuel Cost (Real 2022 USD/MMbtu)" and the other would be named "Real Cost Basis Year" (not "Cost (real) Basis Year") in English - which is why I put them that way. I think it's slightly less consistent but more readable the way it is, but happy to switch out if you feel strongly about it.

jdangerx:
> It looks like you added the resource metadata for this table in another PR?

Yeah, see https://github.com/catalyst-cooperative/pudl/pull/3588 !

These resource schemas are pretty closely coupled to each other, so I wanted to design them all at the same time, but also I didn't want to have to implement all the resources at once.

I added the metadatas to a "staging" area, which let @cmgosnell review the schemas & fields on their own before I did any more implementation.

Now, when implementing a new AEO asset, I just promote the resource definition from the staging area.

I think it's a nice pattern to follow, it avoids huge PRs as well as re-work when we realize the asset design was wrong.

> I have one comment about the table description (in `metadata/resources/eiaaeo.py`) containing the phrase: `Projected fuel prices to the electric power sector` -- what does "to the electric power sector" mean -- should this be "for"? It also doesn't quite match the description you put in the release notes that says "generation and capacity" and leaves out cost.

Yeah, good catch :) let me fix the docs!

src/pudl/metadata/fields.py
aesharpe: mmm yeah there is no great way to deal with this -- could say `real_cost_fuel`? but that is inconsistent with how we report `fuel_cost` in other columns too 🤷. I'll leave this up to you.

src/pudl/transform/eiaaeo.py
```diff
@@ -560,6 +514,115 @@ def core_eiaaeo__yearly_projected_generation_in_end_use_sectors_by_fuel_type(
...
+    "electricity_market_module_region_eiaaeo": 26,
+}
+check_specs = [
+    AeoCheckSpec(
+        name="gen_in_electric_sector_by_tech",
+        asset="core_eiaaeo__yearly_projected_generation_in_electric_sector_by_technology",
+        num_rows_by_report_year={2023: 166972},
+        category_counts=BASE_AEO_CATEGORIES
+        | {
+            "technology_description_eiaaeo": 13,
```
aesharpe, line 593: Is there a reason these aren't in test modules?

aesharpe (approved):
Nothing is blocking, so I'm approving this, but I have some questions!

src/pudl/transform/eiaaeo.py
zaneselvans: These are for defining asset checks, which get run during the ETL after an asset has been materialized.

PR #3658, opened 2024-05-28
Add retry logic to datastore.get_zipfile_resource.

### LINKED ISSUE #3657: Nightly Build Failure 2024-05-28
# Overview

Looks like we had some sort of transient zipfile error - the same `BadZipFile` error we've run into a few times in the past #3453 #3358 #3356 #3338 #3271

This time it's for ferc714, not phmsa - so it's not related to the specific dataset.

Last time we ran into this problem, we added a log line to see what bytes we were actually getting and trying to understand as a ZipFile. It occurs in a few places, each with the same MD5 sum.

The latest instance of that occurs at 06:44 UTC:

```
2024-05-28 06:43:00 [    INFO] catalystcoop.pudl.workspace.datastore:427 Got resource dataset='ferc714', filters={'name': 'ferc714.zip'}, md5sum='4ed329eaee51a685858f45b9d5ceb8b1', 57293468 bytes; turning into ZipFile
2024-05-28 06:43:04 [    INFO] catalystcoop.pudl.workspace.datastore:427 Got resource dataset='ferc714', filters={'name': 'ferc714.zip'}, md5sum='4ed329eaee51a685858f45b9d5ceb8b1', 57293468 bytes; turning into ZipFile
2024-05-28 06:43:30 [    INFO] catalystcoop.pudl.workspace.datastore:427 Got resource dataset='ferc714', filters={'name': 'ferc714.zip'}, md5sum='4ed329eaee51a685858f45b9d5ceb8b1', 57293468 bytes; turning into ZipFile
2024-05-28 06:43:41 [    INFO] catalystcoop.pudl.workspace.datastore:427 Got resource dataset='ferc714', filters={'name': 'ferc714.zip'}, md5sum='4ed329eaee51a685858f45b9d5ceb8b1', 57293468 bytes; turning into ZipFile
2024-05-28 06:43:42 [    INFO] catalystcoop.pudl.workspace.datastore:427 Got resource dataset='ferc714', filters={'name': 'ferc714.zip'}, md5sum='4ed329eaee51a685858f45b9d5ceb8b1', 57293468 bytes; turning into ZipFile
2024-05-28 06:43:42 [    INFO] catalystcoop.pudl.workspace.datastore:427 Got resource dataset='ferc714', filters={'name': 'ferc714.zip'}, md5sum='4ed329eaee51a685858f45b9d5ceb8b1', 57293468 bytes; turning into ZipFile
2024-05-28 06:44:17 [    INFO] catalystcoop.pudl.workspace.datastore:427 Got resource dataset='ferc714', filters={'name': 'ferc714.zip'}, md5sum='4ed329eaee51a685858f45b9d5ceb8b1', 57293468 bytes; turning into ZipFile
2024-05-28 06:44:19 [    INFO] catalystcoop.pudl.workspace.datastore:427 Got resource dataset='ferc714', filters={'name': 'ferc714.zip'}, md5sum='4ed329eaee51a685858f45b9d5ceb8b1', 57293468 bytes; turning into ZipFile
```

And that MD5 sum matches the valid ZIP file I have locally for FERC 714 - which means the bytes we're passing in to `zipfile.ZipFile` are legit.

The failure happens hours later, at 10:13:

```
2024-05-28 10:13:39 +0000 - dagster - ERROR - etl_job - 79bc45f2-b011-4188-8288-eb6fc6a6b8fb - 1381 - RUN_FAILURE - Execution of run for "etl_job" failed. Steps failed: ['raw_ferc714__hourly_planning_area_demand'].
...
Stack Trace:
  File "/home/mambauser/env/lib/python3.12/site-packages/dagster/_core/execution/plan/utils.py", line 54, in op_execution_error_boundary
    yield
  File "/home/mambauser/env/lib/python3.12/site-packages/dagster/_utils/__init__.py", line 468, in iterate_with_context
    next_output = next(iterator)
                  ^^^^^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.12/site-packages/dagster/_core/execution/plan/compute_generator.py", line 141, in _coerc
e_op_compute_fn_to_iterator
    result = invoke_compute_fn(
             ^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.12/site-packages/dagster/_core/execution/plan/compute_generator.py", line 129, in invoke_compute_fn
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/pudl/src/pudl/extract/ferc714.py", line 90, in _extract_raw_ferc714
    ds.get_zipfile_resource("ferc714", name="ferc714.zip") as zf,
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/pudl/src/pudl/workspace/datastore.py", line 431, in get_zipfile_resource
    return zipfile.ZipFile(resource)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mambauser/env/lib/python3.12/zipfile/__init__.py", line 1349, in __init__
    self._RealGetContents()
  File "/home/mambauser/env/lib/python3.12/zipfile/__init__.py", line 1416, in _RealGetContents
    raise BadZipFile("File is not a zip file")
```

We're using Python 3.12.3 in the environment, so that points at the `if not endrec` condition in https://github.com/python/cpython/blob/v3.12.3/Lib/zipfile/__init__.py#L1411-L1416:

```python
        try:
            endrec = _EndRecData(fp)
        except OSError:
            raise BadZipFile("File is not a zip file")
        if not endrec:
            raise BadZipFile("File is not a zip file")
```

Unfortunately, there are many ways in which `_EndRecData` might return something falsey... most of them would be because of some sort of actual content issue with the file, though, and since the hashes are the same I suspect that it's an OSError that's getting swallowed.

# Next steps

I think we should just wrap this whole "make a ZipFile" thing in a retry - if the file *truly* isn't a ZIP file, it doesn't take a lot of time to re-learn that - and if it is just some weird OSError that we don't know about / can't know about, we should hopefully get over it with a retry.

Also, we should do a manual retry to see if that works.

```[tasklist]
```

# Relevant logs
[link to build logs from internal distribution bucket](https://console.cloud.google.com/storage/browser/_details/builds.catalyst.coop/2024-05-28-0602-4ba111aa8-main/2024-05-28-0602-4ba111aa8-main-pudl-etl.log)

See above for excerpts.

jdangerx:
# Overview

Closes #3657.

One more step towards figuring out that weird BadZipFile error that shows up sporadically.

# Testing

Bunch of unit tests!

Suppose I could run the pytest-coverage, too, but it might not run into those `BadZipFile` errors that seem to only pop up very sporadically.

```[tasklist]
# To-do list
```


src/pudl/utils.py
```diff
@@ -0,0 +1,37 @@
+"""A place for helper functions that *don't* import PUDL and thus can be imported anywhere."""
```
jdangerx, line 1: I did this because I started running into circular import problems.

However, *most* of the functions in `pudl.helpers` don't import PUDL. Only these four do:

* `create_datasette_metadata_yaml`
* `check_tables_have_metadata`
* `expand_timeseries`
* `convert_cols_dtypes`

Should we move *those* to a separate file instead of making a fresh file for all the functions where we definitely won't run into circular imports?

test/unit/workspace/datastore_test.py
```diff
@@ -335,4 +338,99 @@ def test_get_resource_with_nonexistent_resource_fails(self):
...
 
+def test_get_zipfile_resource_retry(mocker):
+    ds = datastore.Datastore()
+    not_a_zipfile = b"aaa"
+    ds.get_unique_resource = mocker.MagicMock(return_value=not_a_zipfile)
+    sleep_mock = mocker.MagicMock()
+    with mocker.patch("time.sleep", sleep_mock), pytest.raises(zipfile.BadZipFile):
+        ds.get_zipfile_resource("test_dataset")
+
+    sleep_mock.assert_has_calls([mocker.call(2**x) for x in range(5)])
```
jdangerx, line 349: These assertions of "retried several times" seem like they're just re-testing the retry logic we've already tested - I'm happy to take them out.

test/unit/workspace/datastore_test.py
```diff
@@ -335,4 +338,99 @@ def test_get_resource_with_nonexistent_resource_fails(self):
         self.assertRaises(KeyError, self.fetcher.get_resource, res)
 
 
+def test_get_zipfile_resource_retry(mocker):
+    ds = datastore.Datastore()
+    not_a_zipfile = b"aaa"
+    ds.get_unique_resource = mocker.MagicMock(return_value=not_a_zipfile)
```
jdangerx, line 344: Potentially better to patch the `zipfile.ZipFile` method so we can easily test the "error, then succeed" flow.

src/pudl/utils.py
e-belfer: Where is the import error coming up? Is it possible that the call to `import pudl` plus the multiple `from` calls in `pudl.workspace.datastore.py` is the source of the import error and could be fixed through removing the first call, rather than splitting helpers into two not clearly defined separate files?

If not, my preference would be to do what you suggest (move these four into their own file for helpers that rely on other PUDL functions), rather than start a new file with a single function and expect us to migrate over time.

test/unit/utils_test.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_retry(mocker):
+    func = mocker.MagicMock(side_effect=[RuntimeError, RuntimeError, RuntimeError, 1])
+    sleep_mock = mocker.MagicMock()
+    with mocker.patch("time.sleep", sleep_mock):
+        assert retry(func=func, retry_on=(RuntimeError,)) == 1
+
+    assert sleep_mock.call_count == 3
+    sleep_mock.assert_has_calls([mocker.call(2**x) for x in range(3)])
```
e-belfer, line 11: Will `MagicMock` add a 7 second delay to our unit tests here? Doesn't seem to be the case when I run unit tests, but maybe I'm missing something. If so, we could reduce the number of `RuntimeErrror` calls by one to gain a bit of speed back here if this becomes a noticeable lag.

e-belfer (changes requested):
Some questions and a request re: the reorganization of the `pudl.helpers.py` file. Otherwise, all looks good on my end.

src/pudl/utils.py
jdangerx: Huh. The error was:

```
pytest test/unit/helpers_test.py                                 
ImportError while loading conftest '/Users/dazhong-catalyst/work/pudl/test/conftest.py'.                                        
test/conftest.py:15: in <module>                                                                                                
    import pudl                                                                                                                 
src/pudl/__init__.py:5: in <module>                                                                                             
    from . import (                                                                                                             
src/pudl/analysis/__init__.py:9: in <module>                                                                                    
    from . import (                                                                                                             
src/pudl/analysis/allocate_gen_fuel.py:145: in <module>                                                                         
    from pudl.metadata.fields import apply_pudl_dtypes          
src/pudl/metadata/__init__.py:3: in <module>                    
    from . import (             
src/pudl/metadata/classes.py:63: in <module>                    
    from pudl.workspace.datastore import Datastore, ZenodoDoi   
src/pudl/workspace/__init__.py:11: in <module>                  
    from . import datastore, resource_cache, setup              
src/pudl/workspace/datastore.py:26: in <module>                 
    from pudl.helpers import retry                              
src/pudl/helpers.py:34: in <module>                             
    from pudl.metadata.classes import DatasetteMetadata         
E   ImportError: cannot import name 'DatasetteMetadata' from partially initialized module 'pudl.metadata.classes' (most likely d
ue to a circular import) (/Users/dazhong-catalyst/work/pudl/src/pudl/metadata/classes.py)
```

And for some reason this diff fixed it - currently reading through the import system docs to try to figure out what the heck was going on.

```diff
diff --git a/src/pudl/helpers.py b/src/pudl/helpers.py
index 87d9fc41a..f97a27c10 100644
--- a/src/pudl/helpers.py
+++ b/src/pudl/helpers.py
@@ -30,7 +31,7 @@ from dagster import AssetKey, AssetsDefinition, AssetSelection, SourceAsset
 from pandas._libs.missing import NAType
 
 import pudl.logging_helpers
-from pudl.metadata.classes import DatasetteMetadata
+import pudl.metadata.classes
 from pudl.metadata.fields import apply_pudl_dtypes, get_pudl_dtypes
 from pudl.workspace.setup import PudlPaths
 
@@ -2029,7 +2030,7 @@ def create_datasette_metadata_yaml() -> str:
         derived from their datapackage.json as YAML.
     """
     pudl_output = PudlPaths().pudl_output
-    metadata = DatasetteMetadata.from_data_source_ids(pudl_output)
+    metadata = pudl.metadata.classes.DatasetteMetadata.from_data_source_ids(pudl_output)
     return metadata.to_yaml()
```

jdangerx: I guess my crude mental model is that:

At this specific import time, `DatasetteMetadata` hasn't been defined in `pudl.metadata.classes` yet, since `pudl.metadata.classes` imports `pudl.workspace.datastore` which imports `pudl.helpers`. But we can import the `pudl.metadata.classes` *module* directly since the `import` statement doesn't actually check that `DatasetteMetadata` is defined in the module.

The reason that we can import attributes from `pudl.metadata.fields` and `pudl.workspace.setup` is because they don't circularly depend on `pudl.helpers`.

I think long term we should probably aim to have "truly independent helpers" and "helpers that actually need the rest of PUDL" as separate files, to avoid  but (a) I'm not sure what to name the two files and (b) this works - so I'm happy to punt on that until the next circular import happens >_<

e-belfer: Ah, ok. Glad you managed to figure it out, this seems like the much preferable solution.

Issue #3664, opened 2024-06-10
Get basic user metrics we technically have access to


jdangerx:
### Overview

We want to be able to see our basic usage metrics:
* how many different IPs are accessing our data via Datasette?
* how many different IPs are accessing our data on S3?

Let's write a script that figures that out and provides metrics!

fly.io currently doesn't retain logs for a long time so we need to use the https://github.com/superfly/fly-log-shipper fly log shipper to send logs to S3.

### Success Criteria

* [ ] timestamp, IP address, and accessed resource are available for S3 logs
* [ ] timestamp, IP address, and accessed resource available for Datasette logs

```[tasklist]
### Next steps
* [ ] get S3 permissions to update IAM credentials so CLI/boto can work
* [ ] set up S3 bucket for fly-log-shipper to ship logs to
* [ ] set up some sort of application access for fly-log-shipper so it can send logs to S3
* [ ] deploy datasette behind nginx within the docker instance, and configure it to log the X-Forwarded-For or X-Real-IP headers... need to determine which one of these the Fly.IO load balancer uses. Do this with a tiny little datasette instance on free tier app first.
* [ ] write a script to parse logs into timestamp/IP/resource records
* [ ] compare & contrast
```

jdangerx:
We should timebox this to 5h and prioritize getting S3 parquet logs because of the possibility of replacing datasette altogether.

bendnorman:
I think revamping the `pudl-usage-metrics` repo will take some work. Maybe we can simplify the task by "disabling" the current metrics in the ETL:
- old datasette logs from the Cloud Run days which are probably still helpful for us
- intake catalog logs which we never really utilized 

and integrate just the s3 logs since those are the highest value / most relevant rn. I[ opened a PR](https://github.com/catalyst-cooperative/pudl-usage-metrics/pull/115) with my janky s3 log download script and notebook.

The ETl generally works like this:
1. Pull some logs from GCS
2. Does some cleaning with pandas and dagster
3. Load the cleaned logs into Cloud SQL postgres

I have a github action that processes the latest logs and loads them to Cloud SQL. Cloud SQL is kind of expensive so it might make more sense to use BQ.

I think it makes sense to create a quick design doc for the usage metrics revamp, given there is a lot we could do.

PR #3666, opened 2024-06-12
Add staging environment

jdangerx:
# Overview

Relates to #3664 

I wanted to set up nginx to get actual IPs from people.

But that seemed sort of risky to do on the live production app, so I wanted to get a staging environment set up so that we can actually test changes to the application setup before having them go live on production.

And also, deploying the whole damn database is quite slow, so I wanted a nice way to deploy only a small SQLite file while I'm futzing with the infrastructure. I had been commenting things out but decided it was time to make a proper CLI option.

So basically it's a yak-shave. But not a super long one.

# Testing

Deployed just ferc1/ferc2 xbrls to staging:

```
$ python publish.py --staging --only ferc1_xbrl.sqlite --only ferc2_xbrl.sqlite 
```

Also ran with the full database list, but passed `--build-only` since staging environment isn't sized to handle all the dbs:
```
$ python publish.py --staging -- --build-only
```

And saw that the Dockerfile included the right databases:

```
ENV DATABASES '/data/pudl.sqlite /data/ferc1_dbf.sqlite /data/ferc1_xbrl.sqlite /data/ferc2_dbf.sqlite /data/ferc2_xbrl.sqlite /data/ferc60_dbf.sqlite /data/ferc60_xbrl.sqlite /data/ferc6_dbf.sqlite /data/ferc6_xbrl.sqlite /data/ferc714_xbrl.sqlite /data/censusdp1tract.sqlite'
```

So that seems fine!

```[tasklist]
# To-do list
```


devtools/datasette/fly/run.sh
```diff
@@ -4,7 +4,8 @@ set -eux
 shopt -s nullglob
 
 find /data/ -name '*.sqlite' -delete
+ls
```
jdangerx, line 7: This is helpful for debugging and I figured I'd leave it in.

devtools/datasette/fly/run.sh
```diff
@@ -4,7 +4,8 @@ set -eux
...
 mv all_dbs.tar.zst /data
 zstd -f -d /data/all_dbs.tar.zst -o /data/all_dbs.tar
 tar -xf /data/all_dbs.tar --directory /data
-datasette serve --host 0.0.0.0 /data/pudl.sqlite /data/ferc*.sqlite /data/censusdp1tract.sqlite --cors --inspect-file inspect-data.json --metadata metadata.yml --setting sql_time_limit_ms 5000 --port $PORT
+datasette serve --host 0.0.0.0 ${DATABASES} --cors --inspect-file inspect-data.json --metadata metadata.yml --setting sql_time_limit_ms 5000 --port $PORT
```
jdangerx, line 11: This lets us configure the databases & their ordering via env var in Python. Sort of janky, but 🤷

devtools/datasette/publish.py
```diff
@@ -27,7 +27,8 @@
 
 import click
 
-from pudl.helpers import check_tables_have_metadata, create_datasette_metadata_yaml
+from pudl.helpers import check_tables_have_metadata
+from pudl.metadata.classes import DatasetteMetadata
```
jdangerx, line 31: @e-belfer Turns out that `create_datasette_metadata_yaml` function that was using this import was:
* only used in one place
* 3 lines long

So I decided to inline it in the one place it's used. One fewer circular import to deal with!

devtools/datasette/publish.py
```diff
@@ -85,11 +90,17 @@ def inspect_data(datasets: list[str], pudl_output: Path) -> str:
 
 @click.command(context_settings={"help_option_names": ["-h", "--help"]})
 @click.option(
-    "--fly",
-    "-f",
+    "--production",
```
jdangerx, line 93: Changed `--fly` to `--production` now that we have multiple fly environments.

devtools/datasette/publish.py
```diff
@@ -85,12 +90,16 @@ def inspect_data(datasets: list[str], pudl_output: Path) -> str:
...
     "deploy",
-    flag_value="fly",
-    help="Deploy Datasette to fly.io.",
-    default=True,
```
jdangerx, line 93: Got rid of the default here to make it harder to accidentally deploy to production.

devtools/datasette/publish.py
zaneselvans: There's a reference to it in `test/integration/datasette_metadata_test.py` (about how we're not using it there) which should get updated/removed.

devtools/datasette/fly/staging.toml
```diff
@@ -0,0 +1,27 @@
...
+[http_service]
+  internal_port = 8080
+  force_https = true
+  auto_stop_machines = true
+  auto_start_machines = true
+  min_machines_running = 0
+  processes = ["app"]
+
+[[vm]]
+  memory = "1gb"
```
zaneselvans, line 22: I guess 1GB is enough since it's not going to load the whole SQLite DBs into memory, but each of the DBF and XBRL derived Form 1 DBs are close to 1GB (which is kind of wild since one of them covers 27 years, and the other covers 2 years)

zaneselvans (approved):

devtools/datasette/publish.py
jdangerx: Ah, you mean in the comment? I can remove that :)

devtools/datasette/fly/staging.toml
jdangerx: It's easy enough to change if necessary, but I figured most of the staging stuff is gonna be with small subsets of the data anyways.

jdangerx: Slash, it totally worked for my weird little nginx journey today.

PR #3669, opened 2024-06-12
Log datasette access IPs

jdangerx:
# Overview

Relates to #3664

Right now, our datasette instance only logs out the IP of the load balancer it sits behind. Useless!

If we shove datasette behind nginx, we can then use nginx's ["real IP" module](https://nginx.org/en/docs/http/ngx_http_realip_module.html) to grab the IPs of the clients that the load balancer is forwarding.

This PR does that, *and*:
* reorganizes the deploy logic slightly so there is less duplication between the various deploy targets: "metadata" is an early exit, but everything else should share all the logic except for "how do we take this Dockerfile and deploy it?"
* make `--local` target actually use Docker - this makes the deploy slightly slower but also much closer to what is *actually* getting deployed on fly.io

# Testing

How did you make sure this worked?

I did a fast deploy to our staging server:

```
$ python publish.py --staging --only ferc2_xbrl.sqlite --only ferc6_xbrl.sqlite
```

Then I went to the staging url: https://catalyst-coop-pudl-staging.fly.dev/

And I saw my IP show up in the fly.io live logs:

https://fly.io/apps/catalyst-coop-pudl-staging/monitoring

Which I verified with a quick `curl icanhazip.com` / asking other people to do the same.

```[tasklist]
# To-do list
```


devtools/datasette/fly/50-mod-http-realip.conf
```diff
@@ -0,0 +1 @@
+load_module modules/ngx_http_realip_module.so;
```
jdangerx, line 1: This could technically also live in `nginx.conf` but this is more idiomatic.

devtools/datasette/fly/run.sh
```diff
@@ -8,4 +8,5 @@ ls
 mv all_dbs.tar.zst /data
 zstd -f -d /data/all_dbs.tar.zst -o /data/all_dbs.tar
 tar -xf /data/all_dbs.tar --directory /data
-datasette serve --host 0.0.0.0 ${DATABASES} --cors --inspect-file inspect-data.json --metadata metadata.yml --setting sql_time_limit_ms 5000 --port $PORT
+datasette serve --host 0.0.0.0 ${DATABASES} --cors --inspect-file inspect-data.json --metadata metadata.yml --setting sql_time_limit_ms 5000 --port $DATASETTE_PORT > /dev/null &
```
jdangerx, line 11: We want to suppress the `datasette serve` logs since we get the same information in the nginx access logs.

devtools/datasette/publish.py
```diff
@@ -35,18 +35,27 @@
...
-EXPOSE 8080
 
+COPY nginx.conf /usr/share/nginx/nginx.conf
+COPY 50-mod-http-realip.conf /etc/nginx/modules-enabled/
+
+RUN mkdir /data \
+    && ln -sf /dev/stdout /var/log/nginx/access.log \
+    && ln -sf /dev/stderr /var/log/nginx/error.log
+
+COPY . /app
```
jdangerx, line 54: Moved this down to the bottom since all the other setup changes much less frequently than the contents of the Docker context - this helps with build caching.

src/pudl/metadata/classes.py
```diff
@@ -2216,7 +2216,7 @@ def from_data_source_ids(
             xbrl_resources=xbrl_resources,
         )
 
-    def to_yaml(self) -> None:
+    def to_yaml(self) -> str:
```
jdangerx, line 2219: mypy was complaining about this.

devtools/datasette/fly/nginx.conf
```diff
@@ -0,0 +1,17 @@
+daemon off;
+
+events {
+  worker_connections  1024;
+}
+http {
+  server {
+    listen 8080;
```
jdangerx, line 8: In theory we could turn this (and the Dockerfile) into a jinja template, which might be cleaner, but it felt like gilding the lily to me.

devtools/datasette/publish.py
```diff
@@ -138,48 +150,48 @@ def deploy_datasette(
...
+    metadata_path = fly_dir / "metadata.yml"
+    metadata_yml = DatasetteMetadata.from_data_source_ids(pudl_output).to_yaml()
+
+    logging.info(f"Writing Datasette metadata to: {metadata_path}")
+    with metadata_path.open("w") as f:
+        f.write(metadata_yml)
+    check_tables_have_metadata(metadata_yml, databases)
+
+    if deploy == "metadata":
+        logging.info("Only writing metadata. Aborting now.")
```
bendnorman, line 175: You maybe forgot to abort the script here?

bendnorman (changes requested):
This all mostly makes sense to me! On the staging server I get a 502 but when I reload the page it works.

Also, how do we transfer the logs into a bucket?

devtools/datasette/publish.py
jdangerx: Sure did, fixing! Fortunately we only *actually deploy* if we actively choose a valid deployment target :sweat_smile:

jdangerx:
Yeah, the 502 is happening because nginx will start up and become available before datasette is fully initialized. Since the staging server shuts down with inactivity, you'll see a 502 first every time the VM comes out of inactivity.

In terms of how to transfer logs to a bucket: the official solution is fly-log-shipper: https://github.com/superfly/fly-log-shipper.

bendnorman:
We can't declare victory on tracking logs until we set up the sink right? Any idea how the fly-log-shipper actually runs? Also, looks like there isn't a GCS sink :(

jdangerx:
We'd have to set up the sink to declare victory, yes. The [quick-start in their docs](https://github.com/superfly/fly-log-shipper?tab=readme-ov-file#quick-start) seems reasonable.

The confusing thing about the GCS sink is that it's clearly supported by [Vector](https://vector.dev/), the library that `fly-log-shipper` uses... but the `fly-log-shipper` docs seem to omit GCS from the documentation. So we could *try* configuring Vector for GCS, or we can shove the logs into S3 and then pull them into GCS separately.

bendnorman (approved):
Sounds good! Looks like those next steps are tracked in the issue 👍

PR #3686, opened 2024-06-21
Update docs to link to .zip not .gz S3 archives

zaneselvans:
# Overview

The nightly builds are passing and generating the new ZIP rather than gzip archives but... I forgot to update the links in the docs when I changed those file paths.

```[tasklist]
# To-do list
- [x] Check that the RTD docs build passed and try clicking the new ZIP links.
```

### DIFF
--- a/docs/data_access.rst
+++ b/docs/data_access.rst
@@ -112,8 +112,8 @@ HTTPS using the following links:
 Fully Processed SQLite Databases
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-* `Main PUDL Database <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/pudl.sqlite.gz>`__
-* `US Census DP1 Database (2010) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/censusdp1tract.sqlite.gz>`__
+* `Main PUDL Database <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/pudl.sqlite.zip>`__
+* `US Census DP1 Database (2010) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/censusdp1tract.sqlite.zip>`__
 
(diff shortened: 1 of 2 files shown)


docs/data_access.rst
```diff
@@ -136,48 +136,38 @@ Raw FERC DBF & XBRL data converted to SQLite
...
   * `FERC-60 XBRL Taxonomy Metadata as JSON (2021) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc60_xbrl_taxonomy_metadata.json>`__
 
 * FERC Form 714:
 
-  * `FERC-714 SQLite derived from XBRL (2021-2022) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc714_xbrl.sqlite.gz>`__
+  * `FERC-714 SQLite derived from XBRL (2021-2022) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc714_xbrl.sqlite.zip>`__
   * `FERC-714 Datapackage (JSON) describing SQLite derived from XBRL <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc714_xbrl_datapackage.json>`__
   * `FERC-714 XBRL Taxonomy Metadata as JSON (2021-2022) <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/ferc714_xbrl_taxonomy_metadata.json>`__
 
-.. note::
```
jdangerx, line 171: Honestly a great benefit from switching to `.zip` from `.gz`.

jdangerx (approved):

docs/data_access.rst
zaneselvans: Sadly it turns out that Kaggle treats the `.gz` and `.zip` archives differently and creates a whole directory for every `.zip` when it's extracted regardless of whether there is any directory path in the archive which is very annoying. 🙄

zaneselvans: And results in this nonsense:

image

PR #3687, opened 2024-06-21
Make asset checks matter to integration tests

### LINKED ISSUE #3705: Asset check failures do not fail integration tests
### Overview

@cmgosnell found out that asset check failures appear to not fail integration tests, even if they have `blocking=True` set - she also found that there are no `fgd_equipment_null_cols` or `cooling_equipment_null_cols` checks even being run in the CI logs here: https://github.com/catalyst-cooperative/pudl/actions/runs/9570342720/job/26384880686.

We should probably fix this.

### Success Criteria
* [ ] if an asset check fails in `etl_fast` it should stop integration tests from passing

```[tasklist]
### Tasks
- [x] make a bunch of asset checks fail and see that it doesn't blow up the integration tests
- [x] ... fix it
- [x] make a bunch of asset checks fail and see that it *does* blow up the integration tests
```

jdangerx:
# Overview

Closes #3705 .

What did you change?
* we weren't running asset checks in our ETL when running in integration tests. Instead of making a new, different ETL job for integration tests, we now just run the same job we use everywhere else, but with whatever dataset configuration we have defined for the tests.
* we also weren't asserting that the ETL execution actually succeeded. Asset check failures currently cause `Exception`s, so we don't technically need this, but it seems prudent to assert success anyways.

# Testing

* verified that the `fgd_equipment_null_check` asset check passes, by running the raw_eia860 assets and then the `_core_eia860__fgd_equipment` asset in Dagster
* ran integration tests with `make pytest-coverage` and saw the integration tests pass
* break the `fgd_equipment_null_check` by making it `return AssetCheckResult(passed=False)`
* re-run the `_core_eia860__fgd_equipment` asset in Dagster and see the check fail
* re-run integration tests with `make pytest-coverage` and see that the integration test fails with a bunch of errors about `dagster._core.errors.DagsterAssetCheckFailedError: Blocking check 'fgd_equipment_null_c
heck' for asset '_core_eia860__fgd_e...`

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/etl/cli.py
+++ b/src/pudl/etl/cli.py
@@ -2,16 +2,12 @@
 
 import pathlib
 import sys
-from collections.abc import Callable
 
 import click
 import fsspec
 from dagster import (
     DagsterInstance,
-    Definitions,
-    JobDefinition,
     build_reconstructable_job,
-    define_asset_job,
     execute_job,
 )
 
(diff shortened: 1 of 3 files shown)


jdangerx:
Frustratingly, running `make pytest-coverage` on this branch has caused a bevy of apparently unrelated errors in the integration tests so I will need to poke around more to validate that this indeed causes our tests to fail on failed asset checks.

src/pudl/etl/cli.py
```diff
@@ -23,42 +19,6 @@
 logger = pudl.logging_helpers.get_logger(__name__)
 
 
-def pudl_etl_job_factory(
```
jdangerx, line 26: This was only being used in `conftest` so I moved it over to avoid some import issues.

test/conftest.py
```diff
@@ -272,6 +279,29 @@ def ferc1_xbrl_taxonomy_metadata(ferc1_engine_xbrl: sa.Engine):
     return result.output_for_node("raw_ferc1_xbrl__metadata_json")
 
 
+def _pudl_etl_job_factory(
```
jdangerx, line 282: I removed one layer of function nesting here because it seemed unnecessary based on the [reconstructable jobs docs](https://docs.dagster.io/_apidocs/jobs#reconstructable-jobs).

Since we're only using `execute_in_process` right now anyways, we don't even need a reconstructable job, but it seems fine to leave this because the logic is pretty straightforward now.

test/conftest.py
```diff
@@ -272,6 +279,29 @@ def ferc1_xbrl_taxonomy_metadata(ferc1_engine_xbrl: sa.Engine):
...
+    Args:
+        loglevel: The log level for the job's execution.
+        logfile: Path to a log file for the job's execution.
+        process_epacems: Include EPA CEMS assets in the job execution.
+
+    Returns:
+        The job definition to be executed.
+    """
+    pudl.logging_helpers.configure_root_logger(logfile=logfile, loglevel=loglevel)
+    if not process_epacems:
```
jdangerx, line 300: Instead of making whole new `JobDefinitions` I just return the ones we *actually use* in other builds - that should make test behavior hew more closely to production behavior.

zaneselvans (approved):

Issue #3340, opened 2024-02-03
Operating generators missing 'generator operating date' value in out_eia__monthly_generators and out_eia__yearly_generators pudl tables


mariacastillo21:
### Describe the bug
About 5 GW of generator capacity — 59 operating generators across 26 plants — from 2001 to 2023 have missing generator operating dates in both monthly and annual eia 860 pudl tables.  All of these generators have an operating date reported for them per the latest (Dec 2023) 860m. 

### Bug Severity
How badly is this bug affecting you?
 - **Medium:** We could manually enter in these operating dates from raw EIA 860m, but would prefer to have them in the PUDL pipeline.

### To Reproduce
Steps to reproduce the behavior -- ideally including a code snippet that causes the error to appear.
_Source_ : Pudl_ sqlite downloaded from AWS build on 2/2/23.
_Tables_ :_out_eia__monthly_generators and out_eia__yearly_generators
_Comparison source_ : https://www.eia.gov/electricity/data/eia860m/xls/december_generator2023.xlsx

[Here](https://docs.google.com/spreadsheets/d/15pgWjtUox4EcvDkKA1eZbNyLBCX5jxZG/edit?usp=sharing&ouid=117717579251451091826&rtpof=true&sd=true) is a spreadsheet containing the plant and generator ids, annual report dates (where gen operating date is missing), and operational status column (for verification that this generator was marked as operating at the time).

### Expected behavior
Generator operating date column to be filled in with value from raw EIA 860m

### Software Environment?
 - Operating System: MacOS Ventura 13.2.1
 - Python version and distribution: Anaconda Python 3.10.6

### Additional context
Three generators with non-expected dates filled in:
- two generators (plant id 75, generator id D1 and D2) have operating month marked as 0
- one generator (plant id 1298, generator idGT1) has operating month marked at 88

cc'ing @arengel for visibility

zaneselvans:
Off the top of my head I would guess that this is the result of multiple inconsistent operating dates being reported in different years/months of the EIA data, such that none of them was the clear choice for the correct value, but we will look into it Monday!

e-belfer:
@mariacastillo21 Thanks for the report, Maria! Just wanted to give you a status update, we're about to merge two PRs that will affect this table and its output (the latest quarter of 860M data and #3331), so once they're merged I'll attempt to reproduce this behavior and debug it, and then we can discuss next steps if any changes are required to fix this bug.

e-belfer:
[...]

PR #3697, opened 2024-06-25
Take the most recently reported generator operating date when there's no 70%+ consistently reported date

### LINKED ISSUE #423: generators_entity_eia operating_date field blank when op dates disagree

e-belfer:
# Overview

Closes #423.

Inconsistently reported generator operating dates have been a persistent concern for downstream users, as they wind up being reported as nulls in our entity table after harvesting (see #423, #3340, e.g.). We attempted earlier to adopt a 1 year threshold for rounding and harvesting these dates, but it still left us with a small number of nulls. This PR imposes a more consistent and complete treatment of generator operating dates when they are inconsistently reported.

What did you change?
When generator operating dates are inconsistently reported in 860 and 860M, we default to taking the most recently reported generator operating date. Of the small number of generator operating dates that we could confirm were incorrectly reported, the most recently reported dates wound up being more factually correct (see #423). Rather than having a mix of methods for filling in generator operating dates, let's just adopt a blanket treatment here and assume that the most recent date is an acceptable substitute for a NAN.

# Testing

Run the debug_harvesting notebook, and compare the outputs with the generators identified in #423.

```[tasklist]
# To-do list
```


src/pudl/transform/eia.py
```diff
@@ -540,9 +545,9 @@ def harvest_entity_tables(  # noqa: C901
     entity_df = entity_id_df.copy()
     annual_df = annual_id_df.copy()
     special_case_cols = {
-        "latitude": [_lat_long, 1],
-        "longitude": [_lat_long, 1],
-        "generator_operating_date": [_round_operating_date, "Y"],
+        "latitude": {"method": _lat_long, "round_to": 1},
```
e-belfer, line 548: To do: confirm that I'm not changing the rounding behavior by defaulting to 2 instead of 1

src/pudl/transform/eia.py
```diff
@@ -260,37 +258,44 @@ def _round_operating_date(
...
+    logger.info(f"Rescued dates for {col} records: {len(op_df)}")
     logger.info(
-        f"Rescued rounded {col} for the following units ({entity_idx}): "
+        f"Rescued last {col} for the following units ({entity_idx}): "
         f"{sorted(op_df[entity_idx].apply(lambda row: '_'.join(row.to_numpy().astype(str)), axis=1))}"
     )
     # add the newly cleaned records
     op_clean_df = pd.concat([op_clean_df, op_df])
+    # assert all generator operating dates are not null
+    assert len(op_clean_df[op_clean_df.generator_operating_date.isnull()]) == 0
```
zaneselvans, line 298: This is a minor non-blocking nit, but it feels more readable to use `any()`/`all()` for this kind of test. E.g.
```suggestion
    assert op_clean_df.generator_operating_date.notnull().all()
```

src/pudl/transform/eia.py
zaneselvans: 0.1 degrees can be as much as 7 miles. Does that feel like it's close enough to consider consistent?

src/pudl/transform/eia.py
```diff
@@ -587,14 +592,14 @@ def harvest_entity_tables(  # noqa: C901
         dirty_df = col_df.merge(clean_df[clean_df[col].isnull()][id_cols])
 
         if col in special_case_cols:
-            clean_df = special_case_cols[col][0](
+            clean_df = special_case_cols[col]["method"](
```
zaneselvans, line 595: Gosh I hate when we use indices like this. Thanks for making it more readable.

zaneselvans (approved):
Nothing blocking but:

* I made one small suggestion to simplify the "there's no nulls" test.
* Rounding to 0.1 degrees (up to 7 miles) feels like it could be too aggressive?

e-belfer:
> Nothing blocking but:
> 
>     * Rounding to 0.1 degrees (up to 7 miles) feels like it could be too aggressive?

That decision long pre-dates this PR, and the only changes I'm making here are to just change the format of how we're storing the number. So that question feels out of scope here.

src/pudl/transform/eia.py
e-belfer: I'm not changing the number here, just reformatting how it is fed into the harvesting. If that's a decision we feel unsure about, we should probably take it up in another PR.

src/pudl/metadata/resources/eia.py
```diff
@@ -601,7 +601,12 @@
             ],
             "primary_key": ["plant_id_eia", "generator_id"],
             "foreign_key_rules": {
-                "fields": [["plant_id_eia", "generator_id"]],
+                "fields": [
```
e-belfer, line 604: Oops, this snuck in from another PR!

src/pudl/transform/eia.py
e-belfer: Verified this works as expected

zaneselvans: Ah, sorry I misread this somehow as going from rounding to 2 decimals to 1. Disregard!

zaneselvans:
The [offending version of twine](https://github.com/pypa/twine/issues/1125) has been yanked so this is working now. I went ahead and added it to the merge queue so it can make it into the builds tonight.

Issue #3700, opened 2024-06-26
Integrate 2023 FERC1 Data


cmgosnell:
### New year of data integration check-list:

After spending time on #3544, we have an updated 2023 FERC1 integration task list:

```[tasklist]
- [x] [Obtain fresh data](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#obtain-fresh-data)
- [x] [Test data extraction](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#test-data-extraction)
- [x] Extract: Investigate the apply-diffs failure in the `core_ferc1__yearly_plant_in_service_sched204` extraction from #3698
- [x] Transform: Fix and test the three easy transformer issues from #3698
- [x] [Connect datasets](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#connect-datasets)
- [x] [Run the Full ETL](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#run-the-etl)
- [x] [Run the Fast ETL](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#run-the-etl) - (many of the defensive validations fail more frequently with the fast etl years)
- [x] [Update the output routines and run full tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-output-routines-and-run-full-tests)
- [x] [Run and update data validations](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#run-and-update-data-validations)
- [x] [Update the documentation](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-documentation)
```

## validation test failures
The following min/max row failures happened but all seem to me to be within expected parameters when adding a new year (1/29 years ~3.4%)

```[tasklist]
### min/max row's to investigate
- [x] fbp_ferc1: found 26947 rows, expected 26188. Off by 2.898%
- [x] plants_all_ferc1: found 58520 rows, expected 56407. Off by 3.746%
- [x] plants_hydro_ferc1: found 7202 rows, expected 6979. Off by 3.195%
- [x] plant_in_service_ferc1: found 355918 rows, expected 335756. Off by 6.005%
- [x] plants_small_ferc1: found 17763 rows, expected 16987. Off by 4.568%
[...]

Issue #3698, opened 2024-06-25
Try to run transformation of new year of FERC 1 data and document results


cmgosnell:
```[tasklist]
### Tasks
- [x] Run new FERC 2023 sqlite db & new metadata through transform
- [x] Document any failures
- [x] Document any good things as well!
```

# Failures
## 🔴  `core_ferc1__yearly_hydroelectric_plants_sched406` -> Easy 🟢 
Stage: enforce_schema (last step!)

Error:
```python
2024-06-25 15:58:00 [    INFO] catalystcoop.pudl.transform.classes:1361 core_ferc1__yearly_hydroelectric_plants_sched406: Enforcing database schema on dataframe.
2024-06-25 15:58:00 -0400 - dagster - ERROR - etl_full - 17627cb0-37f2-41fa-99f8-980cec7dac31 - 46621 - core_ferc1__yearly_hydroelectric_plants_sched406 - STEP_FAILURE - Execution of step "core_ferc1__yearly_hydroelectric_plants_sched406" failed.

dagster._core.errors.DagsterExecutionStepExecutionError: Error occurred while executing op "core_ferc1__yearly_hydroelectric_plants_sched406"::

ValueError: invalid literal for int() with base 10: '.2175': Error while type casting for column 'project_num'
```
Why: There is a decimal in the `project_num` column and we have this column dtyped as an `int`

Solution: Easy
1. If we think decimal is fine, change dtype
2. If we think decimal is type, bespoke replace with non-decimal
CG tried option 2 locally and... there are more than one of these.. I used removeprefix and it worked locally:
```python
    def transform_main(self, df):
        """Add bespoke removal of duplicate record after standard transform_main."""
        df = super().transform_main(df).pipe(self.targeted_drop_duplicates)
        df.project_num = df.project_num.str.removeprefix(".")
        return df
```

## 🔴 `core_ferc1__yearly_operating_revenues_sched300` -> Easy 🟢 
Stage: `apply_xbrl_calculation_fixes` (early)

```python
2024-06-25 15:58:37 [    INFO] catalystcoop.pudl.transform.ferc1:2027 core_ferc1__yearly_operating_revenues_sched300: Processing XBRL metadata.
[...]

PR #3701, opened 2024-06-27
Integrate 2023 FERC1 Update

cmgosnell:
# Overview

Working on #3700.
Most of the transform changes are documented over in this scoping issue #3698

2024/07/12 Update: Right now there is one [integration test failure](https://github.com/catalyst-cooperative/pudl/pull/3710#issuecomment-2225592825) coming from `ferc_xbrl_updates`, which @zschira is tackling. Locally I've run the full and fast ETL. With a full db I've run the full validation tests. (I haven't been able to run the `build-deploy-pudl` because the integration test fails.. I'm xfail-ing the one test locally to test everything in one go).

# Testing

```[tasklist]
# To-do list
- [x] If updating analyses or data processing functions: make sure to update or write data validation tests (e.g.,  `test_minmax_rows()`)
```


src/pudl/io_managers.py
```diff
@@ -759,7 +759,7 @@ def __compare_dedupe_methodologies(
 
             # 2024-04-10: this threshold set by looking at existing values for FERC
             # <=2022.
-            threshold_pct = 0.3
+            threshold_pct = 0.32
```
cmgosnell, line 762: this is a hack! we need to investigate this to see if its really a problem

cmgosnell: After some investigation, I do not think this is a problem. For the table that caused this value error in the first place (`electric_plant_in_service_204_duration`), there are 33 excess non-null records out of a possible 10472. Which means that if there were 3 fewer non-null records this test would pass fine np.

After reviewing some of the 2023 excess non-nulls with the apply_diffs methodology, its pretty ambiguous what was supposed to be a record that was updating a previously non-null record into a null vs just that they didn't report anything when they were updating other values. Honestly the fluxxuation in the non-null values from submission to submission was slightly alarming (not all of the values changed by any means! but some were just super different from submission to submission...).

cmgosnell: I'm bumping this up again for the fast tests. This is the same # of excess non-nulls, but with one less year involved.

```
ValueError: Found more than expected excess non-null values using the currently  implemented apply_diffs methodology (#7285) as compared to the best_snapshot methodology (#7254). We expected the apply_diffs methodology to result in no more than 100.32% non-null records but found 100.43%.
```

src/pudl/io_managers.py
cmgosnell, line 1: see [this thread](https://github.com/catalyst-cooperative/pudl/pull/3701#discussion_r1674758622) for what's going on in here

src/pudl/package_data/settings/etl_fast.yml
```diff
@@ -35,7 +35,7 @@ description: >
 version: 0.1.0
 datasets:
   ferc1:
-    years: [2020, 2021, 2022]
+    years: [2020, 2021, 2023]
```
zaneselvans, line 38: Do we expect having a gap in the test years here to be totally fine?

src/pudl/transform/ferc1.py
```diff
@@ -3406,8 +3406,12 @@ class HydroelectricPlantsTableTransformer(Ferc1AbstractTableTransformer):
     table_id: TableIdFerc1 = TableIdFerc1.HYDROELECTRIC_PLANTS
 
     def transform_main(self, df):
-        """Add bespoke removal of duplicate record after standard transform_main."""
-        return super().transform_main(df).pipe(self.targeted_drop_duplicates)
+        """Standard transform_main, bespoke remove duplicate record & remove ``.`` from project_num column."""
+        df = super().transform_main(df).pipe(self.targeted_drop_duplicates)
+        # project_num is an integer column but in 2023 some of them have .'s
+        # as prefixes
+        df.project_num = df.project_num.str.removeprefix(".")
```
zaneselvans, line 3413: At this point it'll be a string column and you're relying on `enforce_schema()` to turn it back into an integer?

zaneselvans (approved):
It sounds like the missing new-table metadata issue appears in the branch this is merging into so this looks good to go. I asked a couple of non-blocking questions.

src/pudl/transform/ferc1.py
cmgosnell: exactly

zaneselvans (approved):

PR #3710, opened 2024-07-03
Update xbrl extraction to use new version

### LINKED ISSUE #3650: Update XBRL extraction to handle multiple taxonomies per year
### Overview

Once catalyst-cooperative/pudl-archiver#349 is done, we have to actually use all that fancy metadata to extract the filings correctly!

### Success criteria
* [ ] filings extracted using the correct taxonomies
* [ ] all FERC 1 transforms execute without errors, though they probably don't pass validation checks.

```[tasklist]
### Next steps
* [ ] read taxonomy and filename from the metadata
* [ ] chunk extraction based on which taxonomy is being used
```

zschira:
# Overview

Closes #3650.

This PR updates PUDL to use the new version of the `ferc-xbrl-extractor`, which handles a new layout of taxonomies in raw archives to accommodate data from one year using multiple taxonomies. Most of this work happened on the archiver/extractor side of things, so this simply bumps the version number and slightly modifies the interface between PUDL and the extractor.

This PR also makes some minor adjustments to settings to enable extraction of the new year of data. Specifically, it adds 2023 to the XBRL portion of the `ferc_to_sqlite_settings` for all forms, and removes references to taxonomies from the settings objects. While working on the settings, I also made a quality of life change by making the `etl_full` settings come from yaml instead of the class defaults. This also results in the settings showing up in the dagster Launchpad, which previously they didn't by default.


src/pudl/extract/xbrl.py
```diff
@@ -24,19 +23,14 @@ def __init__(self, datastore: Datastore):
         """Instantiate datastore wrapper for ferc1 resources."""
         self.datastore = datastore
 
-    def get_taxonomy(self, year: int, form: XbrlFormNumber) -> tuple[io.BytesIO, str]:
+    def get_taxonomy(self, form: XbrlFormNumber) -> tuple[io.BytesIO, str]:
```
zschira, line 26: All taxonomies are in 1 zip, so `year` is no longer necessary

src/pudl/extract/xbrl.py
```diff
@@ -116,23 +110,24 @@ def convert_form(
...
-            loglevel="INFO",
-            logfile=None,
-        )
+    filings_archives = [
+        datastore.get_filings(year, form) for year in form_settings.years
+    ]
+    # if we set clobber=True, clobbers on *every* call to run_main;
+    # we already delete the existing base on `clobber=True` in `xbrl2sqlite`
+    run_main(
+        filings=filings_archives,
```
zschira, line 122: Because all taxonomies are bundled together, the extractor can now just take all of the filings at once.

src/pudl/workspace/datastore.py
```diff
@@ -200,11 +200,11 @@ class ZenodoDoiSettings(BaseSettings):
...
-    ferc1: ZenodoDoi = "10.5281/zenodo.8326634"
-    ferc2: ZenodoDoi = "10.5281/zenodo.8326697"
-    ferc6: ZenodoDoi = "10.5281/zenodo.8326696"
-    ferc60: ZenodoDoi = "10.5281/zenodo.8326695"
-    ferc714: ZenodoDoi = "10.5281/zenodo.8326694"
+    ferc1: ZenodoDoi = "10.5281/zenodo.12549172"
+    ferc2: ZenodoDoi = "10.5281/zenodo.11408175"
+    ferc6: ZenodoDoi = "10.5281/zenodo.11408164"
+    ferc60: ZenodoDoi = "10.5281/zenodo.12021231"
+    ferc714: ZenodoDoi = "10.5281/zenodo.12021228"
```
zaneselvans, line 207: I'm guessing these are the ones that include the 2023 data, and any other updates that have happened since we last updated the FERC archives. Do both the fast and full ETL actually work with the new archives if they're restricted to just extracting the 2021-2022 data?

Because using the new extractor necessitates using new input data, would it make sense for this update to be bundled with #3650 / #3710 instead of happening on its own?

src/pudl/extract/xbrl.py
zaneselvans: The main change int he API here seems to be that since all the taxonomies are included, the extractor can dynamically choose which taxonomy to use based on what year of data is being extracted internally?

zaneselvans (commented):
* Have you run the fast & full ETLs locally to see if using the newer archives creates any issues even when they're restricted to the 2021-2022 XBRL data?
* Might it make sense to roll this update in with #3650 / #3710 instead?

src/pudl/workspace/datastore.py
zschira: I've tested the fast ETL with no problems, and running the full ETL now. Also, I had talked to Christina about bundling them, and we decided to just get this in first, but I don't think either of us felt too strongly, and we don't get any new functionality until #3700 is complete anyway, so I'm fine holding off.

src/pudl/extract/xbrl.py
zschira: Yeah now the archiver will look inside each filing and save metadata mapping the filing to the taxonomy it used, then the extractor will use that metadata to decide which taxonomy to use when parsing.

src/pudl/workspace/datastore.py
zaneselvans: This doesn't feel great to me since we know they go back and change / update old data without saying what changed, and we don't have an easy way to view or document the differences between the two outputs so I'm inclined to use this branch as the basis for the new 2023 data integration and bring in the new data and the new parsing that's required to read the new data at the same time, at which point it'll be more reasonable to expect there to be potentially significant changes to the outputs.

zaneselvans:
@zschira maybe just adding the update to the `FercXbrlToSqlite` settings here (removing the hard-coded taxonomies) would be easiest?

src/pudl/etl/__init__.py
```diff
@@ -262,7 +262,14 @@ def load_dataset_settings_from_file(setting_filename: str) -> dict:
         define_asset_job(
             name="etl_full",
             description="This job executes all years of all assets.",
-            config=default_config,
+            config=default_config
```
zschira, line 265: This will make the `etl_full` pull settings from the yaml file. Explicitly defining settings like this also makes the settings appear in the launchpad by default.

src/pudl/workspace/datastore.py
zschira: That all sounds reasonable to me!

cmgosnell:
hey @zschira ! I'm getting this one error saying that the datasette metadata don't have two of the xbrl tables: 

- 'energy_storage_operations_small_plants_419_duration'
- 'energy_storage_operations_small_plants_totals_419_duration'

This failed for me both locally with the CI test and when i kicked off a runner. I think these are the two new raw tables that got added in the 2023 schema. Is this something that rings a bell in terms of knowing how to add these two tables into the metadata?

I thiiiiiiink that's the last remaining error on these two joint branches. But this fails during the integration tests so its slightly hard to ensure (but i've run the full etl and the validation tests locally).

```
FAILED test/integration/datasette_metadata_test.py::test_database_metadata - AssertionError: These tables are missing datasette metadata: {'pudl': set(), 'ferc1_xbrl': {'energy_storage_operations_small_plants_totals_419_duration', 'energy_storage_operations_small_plants_419_duratio...
```

```
E       AssertionError: These tables are missing datasette metadata: {'pudl': set(), 'ferc1_xbrl': {'energy_storage_operations_small_plants_419_duration', 'energy_storage_operations_small_plants_totals_419_duration'}, 'ferc2_xbrl': set(), 'ferc60_xbrl': set(), 'ferc6_xbrl': set(), 'ferc714_xbrl': set()}
```

zaneselvans:
I ran the full ETL locally and then was able to reproduce this error with 

```
pytest --live-dbs test/integration/datasette_metadata_test.py::test_database_metadata
```

@zschira I wonder if this is happening because we're not using the most recent taxonomy to generate the `ferc1_xbrl_taxonomy_metadata.json`? Now that there are multiple taxonomies to look at, does it make sense to just use the most recent one and hope that (like with the old DBF database schemas) it ends up being able to accommodate all the older data structures as well? Or do we need to munge them together somehow?

zaneselvans (approved):

PR #3715, opened 2024-07-11
Superset deployment

bendnorman:
# Overview

This PR contains superset configuration and cloud deployment changes for our data exploration tool. 

# Testing

```[tasklist]
# To-do list
```


superset/superset_config.py
```diff
@@ -0,0 +1,69 @@
...
+            "client_id": AUTH0_CLIENT_ID,  # Client Id (Identify Superset application)
+            "client_secret": AUTH0_CLIENT_SECRET,  # Secret for this Client Id (Identify Superset application)
+            "client_kwargs": {"scope": "openid profile email groups"},
+            "server_metadata_url": f"https://{AUTH0_DOMAIN}/.well-known/openid-configuration",
+        },
+    }
+]
+
+AUTH_USER_REGISTRATION = True
+AUTH_USER_REGISTRATION_ROLE = "Gamma"
```
jdangerx, line 31: Should people get SQL Lab access by default?

terraform/main.tf
```diff
@@ -98,3 +98,255 @@ resource "google_storage_bucket_iam_binding" "binding" {
...
+        container_port = 8088
+      }
+      resources {
+        limits = {
+          cpu    = "4"
+          memory = "2048Mi"
+        }
+      }
+    }
+    volumes {
```
jdangerx, line 210: Did something come up with the GCS volume mount stuff, or is that just not pushed yet?

jdangerx (commented):
Driveby comments since I was looking at a bunch of tickets. Ping me whenever you want a more in-depth review!

superset/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# hadolint ignore=DL3006
+FROM apache/superset
```
bendnorman, line 2: @jdangerx should we pin the base image version? If so, can we set up dependabot to keep the base image and the `requirements.txt` file up to date?

superset/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+docker compose up
+```
+
+If this is the first time running superset locally or you recently ran `docker compose down` you'll need to run the commands in `setup.sh`.
+
+## Making changes to the production deployment
+TODO: instructions on how to connect to Cloud SQL
+
+## Deploy to Cloud Run
```
bendnorman, line 39: What should the deployment flow be for superset? I can think of a few reasons to trigger a new deployment:
When do we want to update superset
  * package in requirements.txt is updated
  * the base image is updated
  * we make a change to `superset_config.py`
  * I think we’ll probably want to redeploy superset when there is new data because changes to tables might break shared dashboards. Also, should superset point at `nightly` or `stable`?

superset/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+      SUPERSET_DB_HOST: ${SUPERSET_DB_HOST}
+      SUPERSET_DB_PORT: ${SUPERSET_DB_PORT}
+      SUPERSET_DB_USER: ${SUPERSET_DB_USER}
+      SUPERSET_DB_PASS: ${SUPERSET_DB_PASS}
+      SUPERSET_DB_NAME: ${SUPERSET_DB_NAME}
+    ports:
+      - "8080:8088"
+    volumes:
+      # - {path to your pudl.duckdb file}:/app/pudl.duckdb
+      - ./roles.json:/app/roles.json
```
bendnorman, line 21: Do we want to try to track the role definitions in git so our local deployments can use the roles we're using in production?

superset/superset_config.py
```diff
@@ -0,0 +1,74 @@
...
+
+def FLASK_APP_MUTATOR(app: Flask) -> None:  # noqa: N802
+    """Superset function that allows you to configure the Flask app.
+
+    Args:
+        app: The Flask app instance
+    """
+    app.config.update(
+        PREFERRED_URL_SCHEME="https",
+    )
```
bendnorman, line 74: Unfortunately, this didn't resolve the HTTP redirect issue. I think the issue might be related to the [auth0 Oauth development tokens](https://auth0.com/docs/authenticate/identity-providers/social-identity-providers/devkeys#limitations-of-developer-keys-when-using-universal-login) we're using.

terraform/main.tf
bendnorman: I hadn't pushed it!

terraform/main.tf
```diff
@@ -102,3 +102,297 @@ resource "google_storage_bucket_iam_binding" "binding" {
     "user:aengel@rmi.org",
   ]
 }
+
```
bendnorman, line 105: I'd like to learn more about how to structure terraform projects.  It'd be helpful to isolate the superset setup in a separate file or something so we can easily reuse it for other projects like the pudl usage metrics repo.

terraform/main.tf
```diff
@@ -102,3 +102,297 @@ resource "google_storage_bucket_iam_binding" "binding" {
...
+      password_change_interval    = "30s"
+      enable_password_policy      = true
+    }
+  }
+  # set `deletion_protection` to true, will ensure that one cannot accidentally delete this instance by
+  # use of Terraform whereas `deletion_protection_enabled` flag protects this instance at the GCP level.
+  deletion_protection = true
+}
+
+resource "google_secret_manager_secret" "superset_database_username" {
```
jdangerx, line 271: Hooray! I love that you're using the secret manager to manage secrets, it makes my shriveled little security heart sing.

It's a little heavyweight to define *all this boilerplate* just to say "there are these secrets, here are their names, and this service account has access to them." We can use the [`for_each`](https://developer.hashicorp.com/terraform/language/meta-arguments/for_each) syntax of Terraform to drastically reduce that.

I think we should do that *before* this gets merged, since it will make Terraform re-create all the secrets, which requires a little manual intervention to re-populate the secrets.

terraform/main.tf
```diff
@@ -102,3 +102,297 @@ resource "google_storage_bucket_iam_binding" "binding" {
...
+  secret_id = "superset-database-connection-name"
+  replication {
+    auto {}
+  }
+}
+
+resource "google_secret_manager_secret_iam_member" "superset_database_username_compute_iam" {
+  secret_id = google_secret_manager_secret.superset_database_username.id
+  role      = "roles/secretmanager.secretAccessor"
+  member    = "serviceAccount:345950277072-compute@developer.gserviceaccount.com"
```
jdangerx, line 300: I think it makes sense to make a *new* service account that's only for Superset instead of using the default Compute Engine SA. Then it's easier to restrict access to the prod database, which makes it harder to accidentally blow it up.

jdangerx (changes requested):
Thanks for wrangling all this! There's definitely bits to change but also it seems like it works.

The main blocking thing is just pinning the docker image - once that's done I'm happy to merge to `main`.

It's probably also a good idea to get people to run their own Auth0 apps instead of using the production one for development.

Some improvements/follow-ups that could be separate PRs:
* triggering redeploys on nightly build
* define a new SA specifically for Superset and give *that* all the accesses it needs - probably refactoring to use `for_each` along the way. 
* get a docker-compose setup for pointing at production

superset/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    environment:
+      AUTH0_CLIENT_ID: ${AUTH0_CLIENT_ID}
+      AUTH0_CLIENT_SECRET: ${AUTH0_CLIENT_SECRET}
+      AUTH0_DOMAIN: ${AUTH0_DOMAIN}
+      SUPERSET_SECRET_KEY: ${SUPERSET_SECRET_KEY}
+      SUPERSET_DB_HOST: ${SUPERSET_DB_HOST}
+      SUPERSET_DB_PORT: ${SUPERSET_DB_PORT}
+      SUPERSET_DB_USER: ${SUPERSET_DB_USER}
+      SUPERSET_DB_PASS: ${SUPERSET_DB_PASS}
+      SUPERSET_DB_NAME: ${SUPERSET_DB_NAME}
```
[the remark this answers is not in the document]
bendnorman: Good idea! I'll add some defaults for none secrety things

superset/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    ports:
+      - "8080:8088"
+    volumes:
+      # - {path to your pudl.duckdb file}:/app/pudl.duckdb
```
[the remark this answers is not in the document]
bendnorman: How about `$PUDL_OUTPUT/pudl.duckdb` or should it be a separate env var?

terraform/main.tf
```diff
@@ -102,3 +102,297 @@ resource "google_storage_bucket_iam_binding" "binding" {
...
+}
+
+resource "google_secret_manager_secret" "superset_secret_key" {
+  secret_id = "superset-secret-key"
+  replication {
+    auto {}
+  }
+}
+
+resource "google_sql_database_instance" "postgres_pvp_instance_name" {
```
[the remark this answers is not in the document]
bendnorman: I grabbed this terraform code from the [google docs ](https://cloud.google.com/sql/docs/postgres/create-instance#create-2nd-gen)and forgot to rename the terraform resource. Is it too late to rename it?

terraform/main.tf
bendnorman: Ah good to know! Do you think it's worth changing now?

terraform/main.tf
bendnorman: Ah yes it sure does. I'll create an issue for this.

bendnorman:
Thanks for all the great feedback! I pinned the docker base image version, updated auth0 env var instructions and set docker compose env var defaults. I also created some draft issues that I'll flesh out tomorrow.

terraform/main.tf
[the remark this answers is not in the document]
jdangerx: If you rename it, it will delete the resource and create a new one. So it sort of depends on how hard it is to re-initialize the state.

jdangerx (approved):

jdangerx:
You're hitting a bunch of this error in CI:
```
ERROR test/integration/glue_test.py::test_unmapped_utils_eia - TypeError: ForwardRef._evaluate() missing 1 required keyword-only argument: 'recursive_guard'
```
Which is related to dagster / python version incompatibilities: https://github.com/dagster-io/dagster/issues/22985

PR #3717, opened 2024-07-15
Update conda lockfile for week of 2024-07-15

pudlbot:


e-belfer:
Docs build issue was due to a problem in `sphinx` that was addressed a few hours ago (see [here](https://github.com/sphinx-doc/sphinx/issues/12585)). Will try to update the lockfile again and see if that fixes the issue.

e-belfer (approved):
Looks like the sphinx bump fixed our docs issue, should be good to go.

e-belfer (approved):
Addressing the dagster issue should hopefully fix the docker problem. I've verified that dagster deploys locally and runs as expected with the pinned version.

PR #3723, opened 2024-07-18
Update allowable not water limited capacity ratio.

zaneselvans:
# Overview

Changed the lower bound on the FERC Form 1 steam plants "not water limited ratio" validation test.

# Testing

I ran the full data validations against a fresh database and they all passed.


e-belfer (approved):
:+1:

PR #3670, opened 2024-06-14
Transform FERC-714 load forecast table

seeess1:
# Overview

NOTE: This is still a WIP. Just putting up a PR for now to make sure I'm on the right track and to get input.

Closes #3519.

What problem does this address?

This creates a core asset out of the `raw_ferc714__yearly_planning_area_forecast_demand` data.

What did you change?

(Still in progress)

# Testing

(Still in progress)

How did you make sure this worked? How can a reviewer verify this?

- I ran a local version of this and checked the dataframe output.

```[tasklist]
# To-do list
- [ ] If updating analyses or data processing functions: make sure to update or write data validation tests (e.g.,  `test_minmax_rows()`)
- [ ] Update the [release notes](../docs/release_notes.rst): reference the PR and related issues.
- [ ] Ensure docs build, unit & integration tests, and test coverage pass locally with `make pytest-coverage` (otherwise the merge queue may reject your PR)
- [ ] Review the PR yourself and call out any questions or issues you have
- [ ] For minor ETL changes or data additions, once `make pytest-coverage` passes, make sure you have a fresh full PUDL DB downloaded locally, materialize new/changed assets and all their downstream assets and [run relevant data validation tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#data-validation) using `pytest` and `--live-dbs`.
- [ ] For significant ETL, data coverage or analysis changes, once `make pytest-coverage` passes, ensure the full ETL runs locally and [run data validation tests](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/testing.html#data-validation) using `make pytest-validate` (a ~10 hour run). If you can't run this locally, run the `build-deploy-pudl` GitHub Action (or ask someone with permissions to). Then, check the logs on the `#pudl-deployments` Slack channel or `gs://builds.catalyst.coop`.
```

PR #3727, opened 2024-07-22
If branch is a fork, skip the `release.yml` workflow

e-belfer:
# Overview

Our `release.yml` workflow can't be run on forks without us taking over the PR (see #3670 for latest problems). This test doesn't often fail, and is typically only critical for release PRs, which we don't anticipate being external PRs. So, let's disable this test for PRs on forks.

What did you change?
Added an `if:` conditional to the first job, that checks whether it is a fork. The other jobs depend on the first, so if it is skipped the rest of the jobs will also be skipped.

# Testing

Check the test still runs on this branch. Once it's merged, open a fork and see if the check still attempts to run?

```[tasklist]
# To-do list
```


zaneselvans (commented):
Does skipping all of the jobs in the workflow count as "passing" in the context of a required check? I tried making a similar change on the FERC-714 PR (both on the PR branch and on main) and it didn't have the outcome I expected.

How can we most easily test that this does what we want? I imagine it needs to be on both `main` and the branch/fork that the PR is coming from, and then we'll need to make the `release` workflow a required check again and make a PR from a fork on a personal account.

e-belfer:
> Does skipping all of the jobs in the workflow count as "passing" in the context of a required check? I tried making a similar change on the FERC-714 PR (both on the PR branch and on main) and it didn't have the outcome I expected.

Tested by making the repository wrong here, and yes, this works as expected!

zaneselvans (approved):

zaneselvans:
Should I re-enable the `release` check as required now? It seems like we expect it to skip and have that be okay.

zaneselvans:
I've set the package build back to being required.

e-belfer:
[Based on this documentation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks) I think that should work.

PR #3740, opened 2024-07-26
Update all documentation URLs to point at nightly not latest

zaneselvans:
# Overview

* Looking at the Google Analytics stats from our RTD site, I noticed a mishmash of `latest` vs. `nightly` URLs in the results.  We use `nightly` as the default version on RTD, so I updated all of the URLs in our README, issue templates, etc. to point at `nightly` as well, so folks don't end up jumping between versions inadvertently, and so our analytics output is cleaner.
* I also added a couple of orgs to the PUDL users list in the README.
* Fixed the broken link to our release notes in our PR template and made it ore concise.

# Testing

I rebuilt the docs locally and tested all of the organizational links in the PUDL Users section manually.

```[tasklist]
# To-do list
- [ ] Ensure docs build, unit & integration tests, and test coverage pass locally with `make pytest-coverage`.
```


bendnorman (approved):
Thanks!

PR #3751, opened 2024-07-29
Add `generator_operating_date` to 860M changelog table

### LINKED ISSUE #3722: Include `generator_operating_date` in the `core_eia860m__changelog_generators` table
In some of our models we use monthly 860 data to define the set of generators since it has all the columns we need and reflects the best available data on planned retirements and additions. We'd like to be able to use `core_eia860m__changelog_generators`  for this purpose but it is missing the `generator_operating_date` which makes this inconvenient.

I'd like to have the `generator_operating_date` as reported in 860m included in the `core_eia860m__changelog_generators` table.

**Describe alternatives you've considered**
We could merge on that single column from `_out_eia__yearly_generators` but that's less convenient and it's not clear if those dates are actually the ones, or the same as the ones reported in 860m.

e-belfer:
# Overview

Closes #3722.

Adds `generator_operating_date` from 860M to EIA 860M changelog table.

What did you change?
Added `generator_operating_date` to the table schema and made a new alembic migration.

Note that this table has no downstream assets and does not get harvested, so this change will not affect any of the other EIA tables. Whereas elsewhere the `generator_operating_date` is treated as a static and harvested field, here it will originate from the referent 860M data record. This is consistent with the data in the rest of this table, which originates from 860M data only, and other attributes that are harvested from 860M data and treated as static entities in core EIA tables were already in this table (e.g., latitude and longitude). Thus, this seems like a small and straightforward fix.

# Testing

Generate the table, and make sure the `generator_operating_date` column is there as expected.

```[tasklist]
# To-do list
```


zaneselvans (approved):

src/pudl/metadata/resources/eia860m.py
```diff
@@ -32,6 +32,7 @@
                 "energy_source_code_1",
                 "energy_storage_capacity_mwh",
                 "fuel_type_code_pudl",
+                "generator_operating_date",
```
zaneselvans, line 35: Given that this is the only change, I'm assuming that the column is present in the dataframe that generates this asset, and it was getting dropped when we enforce the schema?

PR #3753, opened 2024-07-31
Add mozilla repo to wif pool

zschira:
# Overview

This PR updates our terraform config to include the mozilla development repo in our gcloud WIF pool.

## Questions
* My understanding is that I just need this update to `main.tf`, then update the workflow file to use the workload identity pool with the appropriate service account. Is there anything else we need to do?
* When I run `terraform plan` there's a bunch of new superset resources that would be deleted because they haven't been migrated into terraform yet. Is there an easy/clean way to avoid that?

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -66,6 +66,10 @@ module "gh_oidc" {
       sa_name   = "projects/${var.project_id}/serviceAccounts/gce-github-action-test@catalyst-cooperative-pudl.iam.gserviceaccount.com"
       attribute = "attribute.repository/catalyst-cooperative/gce-build-test"
     }
+    "sec-extraction-test-github-action" = {
+      sa_name   = "projects/catalyst-cooperative-mozilla/serviceAccounts/mozilla-dev-sa@catalyst-cooperative-mozilla.iam.gserviceaccount.com"
+      attribute = "attribute.repository/catalyst-cooperative/mozilla-sec-eia"
+    }
   }
 }
 


jdangerx:
> * My understanding is that I just need this update to `main.tf`, then update the workflow file to use the workload identity pool with the appropriate service account. Is there anything else we need to do?

I think that's right. The workflow has to come from the `mozilla-sec-eia` repo, though, because of:

```terraform
      attribute = "attribute.repository/catalyst-cooperative/mozilla-sec-eia"
```

> * When I run `terraform plan` there's a bunch of new superset resources that would be deleted because they haven't been migrated into terraform yet. Is there an easy/clean way to avoid that?

Not as far as I know. You can rebase your branch onto Ben's branch, or coordinate with @bendnorman  to just nuke out his superset resources for now - they're all recreatable via TF at this point anyways.

jdangerx (approved):
Looks good! Approved, pending the workarounds for `terraform plan` gunk caused by the superset resources.

zschira:
I tried rebasing off of the superset branch, but there were a couple resources that are a bit out of sync with what's on terraform still. I didn't want to screw anything up if Ben has been doing some prototyping in the cloud console or something that still needs to get translated to terraform, so I didn't use `terraform apply`. Instead, I just manually updated the resources in the console, but it should be exactly the same as what terraform would've done, and I think we can get it all sorted once superset is all finalized.

PR #3754, opened 2024-08-02
Update some dataset lists and date ranges.

zaneselvans:
# Overview

- Trenton noticed that the date ranges associated with some of our datasets were stale in the documentation.
- While looking into fixing that, I noticed that we also hadn't updated some of our lists of what data is available, e.g. in our README.
- There were also some outdated references to Sloan work that has either been finished or descoped.
- Within the current documentation structure it's a little hard to have a single authoritative source, because we don't have `data_source` pages for all of our recently datasets (e.g. Census DP1, NREL ATB, EIA AEO) and a number of datasets are only partially processed, but it would  still be good to highlight them (e.g. EIA-176, PHMSA, FERC 2) and provide background on them, since we're asking contributors to work on them, and in the case of the FERC data, they're probably the most accessible version of the data available on the web.
- To at least temporarily patch this up while the bigger docs refactor happens I created two separate pages for future target datasets, and data that has already been integrated into PUDL (at least partially) but doesn't have a full-on data source documentation page.
- I also updated the docs build environment that we use on ReadTheDocs and pinned `sphinx-autoapi<3.2` because of this issue: https://github.com/readthedocs/sphinx-autoapi/issues/476 (which necessetated rebuilding the conda environment).

# Testing

I rebuilt the docs locally and inspected them manually.

```[tasklist]
# To-do list
- [x] Run `make docs-build` locally to ensure that the changes haven't broken anything.
```


docs/data_sources/wip_future.rst
```diff
@@ -36,16 +32,6 @@ integration, don't hesitate to `open an issue on Github
 <https://github.com/catalyst-cooperative/pudl/issues>`__ requesting the data and
 explaining why it would be useful.
 
-.. _data-censusdp1tract:
-
-Census DP1
-^^^^^^^^^^
-The `US Census Demographic Profile 1 (DP1) <https://www.census.gov/geographies/mapping-files/2010/geo/tiger-data.html>`__
```
e-belfer, line 43: I think we want this to live somewhere, rather than getting rid of it completely - it's the only place where we explain our source of census data, and census data source is an often annoying technicality that users might actually care about.

e-belfer (changes requested):
Everything looks ok to me, but we shouldn't just delete the only mention of our Census data source from our documents. Can we make this its own little data page, or move it elsewhere?

docs/data_sources/wip_future.rst
zaneselvans: I was not thrilled about this move (on my part) either, it just didn't seem to fit in anywhere else.

In general I hope we can get to where every data source has its own dedicated documentation page, but the current template / framework makes that kind of janky and hard to update.

Should I make a non-templated page for it do you think?

pyproject.toml
```diff
@@ -75,7 +75,7 @@ dependencies = [
     "Shapely>=2",
     "splink>=3.9.14,<4", # Need to update PUDL to use new Splink v4 API. See issue #3735
     "sphinx>=7.4.4",
-    "sphinx-autoapi>=3",
+    "sphinx-autoapi>=3,<3.2",
```
zaneselvans, line 78: Required because of what seems like a bug in newer `sphinx-autoapi` which I reported here: https://github.com/readthedocs/sphinx-autoapi/issues/476

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -1415,7 +1415,7 @@ def make_parts_to_ids_dict():
     plant_id_eia column is always first, so we're going to grab the last column.
 
     Returns:
-        dictionary: plant-part names (keys) cooresponding to the main ID column
+        dictionary: plant-part names (keys) corresponding to the main ID column
```
zaneselvans, line 1418: Did a search-and-replace on this misspelling, since there were a bunch of them.

.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
   # Formatters: hooks that re-write Python & documentation files
   ####################################################################################
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.5.6
+    rev: v0.5.7
```
zaneselvans, line 32: Updated to stay in sync with what's in `pyproject.toml`

docs/conf.py
```diff
@@ -69,10 +69,12 @@
 # Redirects to keep folks from hitting 404 errors:
 redirects = {
     "data_dictionary": "data_dictionaries/pudl_db.html",
+    "data_sources/wip_future.html": "data_sources/future_data.html",
 }
 
 # Automatically generate API documentation during the doc build:
 autoapi_type = "python"
+autoapi_keep_files = False  # Set to True to debug auto-generated RST files
```
zaneselvans, line 77: Very helpful for debugging AutoAPI issues!!!

docs/conf.py
```diff
@@ -69,10 +69,12 @@
 # Redirects to keep folks from hitting 404 errors:
 redirects = {
     "data_dictionary": "data_dictionaries/pudl_db.html",
+    "data_sources/wip_future.html": "data_sources/future_data.html",
```
zaneselvans: To avoid any 404 errors due to the changed URL.

.readthedocs.yaml
```diff
@@ -8,9 +8,9 @@ version: 2
 
 # Set the version of Python and other tools you might need
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   tools:
-    python: mambaforge-22.9
+    python: mambaforge-23.11
```
zaneselvans, line 13: Update to the most recent documentation build environments. These should be closer to the environments we use locally, and in the nightly builds.

docs/conf.py
```diff
@@ -69,10 +69,12 @@
 # Redirects to keep folks from hitting 404 errors:
 redirects = {
     "data_dictionary": "data_dictionaries/pudl_db.html",
+    "data_sources/wip_future": "future_data.html",
```
zaneselvans, line 72: To avoid broken links to the old page URL.

zaneselvans (commented):
The main substantive changes here are:
- Remove specific dates from the README
- Add new datasets to the README
- Create two separate pages for future target datasets, and data that has already been integrated into PUDL (at least partially) but doesn't have a full-on data source documentation page.
- Also updated some docs build environment settings.

pudlbot (approved):

PR #3774, opened 2024-08-08
Make an EIA plant part association table with generators

### LINKED ISSUE #3773: Build an EIA plant part to generator association table

cmgosnell:
# Overview

Closes #3773.
I did have a question regarding what to name the plant_gen column... with @aesharpe  help we landed on just using the a `_plant_gen` suffix. It was a little tedious bc now there is a bunch of nearly duplicate column names. But it felt like the best solution because it avoid the possibilities of users picking up this table and now knowing whether these columns are associated with the ppe record of the `record_id_eia` or the plant_gen `record_id_eia_plant_gen`. So a handful of new columns got added in here with the `_plant_gen` suffix.

What did you change?

# Testing

```[tasklist]
# To-do list
```


src/pudl/analysis/plant_parts_eia.py
```diff
@@ -336,34 +336,6 @@
     "appro_part_label",
 ]
 
-PPE_COLS = [
```
cmgosnell, line 339: this just wasn't being used anywhere

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -713,12 +685,8 @@ def add_one_to_many(
...
             part_name="plant_gen",
             cols_to_keep=["plant_part"],
-        )[["record_id_eia_og", "record_id_eia", "plant_part_og"]].rename(
-            columns={
-                "record_id_eia": "gen_id",
-                "record_id_eia_og": "record_id_eia",
-                "plant_part_og": "plant_part",
-            }
+        )[["record_id_eia", "record_id_eia_plant_gen", "plant_part"]].rename(
+            columns={"record_id_eia_plant_gen": "gen_id"}
```
cmgosnell, line 689: i changed the `match_to_single_plant_part` function just a little to add a suffix to the plant_gen side of the merge instead of adding the suffix of `_og` to the `multi_gran_df` side

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -1119,12 +1087,8 @@ def execute(self, ppl):
...
             cols_to_keep=["plant_part"],
             one_to_many=True,
-        )[["record_id_eia_og", "record_id_eia", "plant_part_og"]].rename(
-            columns={
-                "record_id_eia": "gen_id",
-                "record_id_eia_og": "record_id_eia",
-                "plant_part_og": "plant_part",
-            },
+        )[["record_id_eia", "record_id_eia_plant_gen", "plant_part"]].rename(
+            columns={"record_id_eia_plant_gen": "gen_id"},
```
cmgosnell, line 1091: this is another instance of a change in which side of the `match_to_single_plant_part` merge gets a suffix

src/pudl/metadata/fields.py
```diff
@@ -1290,6 +1290,10 @@
         "type": "string",
         "description": "The code representing the most predominant type of energy that fuels the generator.",
     },
+    "energy_source_code_1_plant_gen": {
+        "type": "string",
+        "description": "Associated plant_gens' code representing the most predominant type of energy that fuels the generator.",
+    },
```
aesharpe, line 1296: Is "Associated plant_gens" a reference to the association table? It's a little bit confusing if so. Maybe we could add a phrase at the end saying something like "Describes the `record_id_eia_plant_gen` value in `out_eia__yearly_assn_plant_parts_plant_gen` instead of this phrase at the beginning.

src/pudl/metadata/fields.py
```diff
@@ -3388,6 +3426,10 @@
         "description": "Indicates whether the plant is regulated or non-regulated.",
     },
     "report_date": {"type": "date", "description": "Date reported."},
+    "report_date_plant_gen": {
+        "type": "date",
+        "description": "Associated plant_gen's date reported.",
+    },
```
aesharpe, line 3432: I feel like report date doesn't need the `plant_gen` suffix?

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -124,6 +127,34 @@
         "etl_group": "outputs",
         "field_namespace": "eia",
     },
+    "out_eia__yearly_assn_plant_parts_plant_gen": {
+        "description": """In order to easily determine what generator records are associated with every
+plant part record, we made this association table. This table associates every plant part
+record (identified as ``record_id_eia``) from the ``out_eia__yearly_plant_parts`` table to the associated
+'plant_gen' records (identified as ``record_id_eia_plant_gen``). All of the columns in this table
+besides ``record_id_eia`` are attributes of the plant_gen record and have a suffix of '_plant_gen'.""",
```
aesharpe, line 135: I think you could add a little bit of clarity here about the different record id columns being derived from the same column in the PPE table. Also the 1:M nature of the values and the fact that some values will have the same record id in each column because some of the plant parts are single generators.

I would also say "are attributes of the `record_id_eia_plant_gen` record..." for consistency.

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -124,6 +127,34 @@
...
+                "ferc1_generator_agg_id_plant_gen",
+                "prime_mover_code_plant_gen",
+                "unit_id_pudl_plant_gen",
+                "technology_description_plant_gen",
+                "ferc_acct_name_plant_gen",
+                "plant_id_eia_plant_gen",
+                "generator_operating_year_plant_gen",
+                "generator_id_plant_gen",
+            ],
+            "primary_key": ["record_id_eia", "record_id_eia_plant_gen"],
```
aesharpe, line 152: just verifying that report date doesn't need to be a PK?

aesharpe (commented):
Spell check queen! Just finishing materializing the assets to make sure they work. Added a few, mostly non-blocking comments.

**UPDATE**: materialization worked! I also have an additional question. Does the `gen_owned` vs. `gen_total` make a difference here? It looks like you are mapping to both `gen_owned` and `gen_total` generation record ids. Will that entail some overlap? Its that desired?

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -710,15 +682,11 @@ def add_one_to_many(
         # Get the 'm' generator IDs 1:m
         one_to_many_single = match_to_single_plant_part(
             multi_gran_df=plant_parts_eia_filt,
-            ppl=plant_parts_eia,
+            ppe=plant_parts_eia,
```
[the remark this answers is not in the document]
aesharpe: (most if not all of them are in the documentation not the code itself)

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -127,33 +127,34 @@
...
-    #             "generator_operating_year_plant_gen",
-    #             "generator_id_plant_gen",
-    #         ],
-    #         "primary_key": ["record_id_eia", "record_id_eia_plant_gen"],
-    #         "sources": ["eia860", "eia923"],
-    #         "etl_group": "outputs",
-    #         "field_namespace": "eia",
-    #     },
-    # },
+    "out_eia__yearly_plant_parts_plant_gen_assn": {
```
zaneselvans: Should this be an `out_eia__` or an `out_pudl__` table?  The FERC-EIA plant parts association is currently `out_pudl__`

For naming consistency I think we want to move the `assn` to the beginning after the double underscore as an indication of what kind of table it is. Though sadly when there's a time frequency we haven't be consistent about whether it comes first or second so there are both `__assn_yearly_` and `__yearly_assn_` tables:

```
core_eia860__assn_boiler_cooling
core_eia860__assn_boiler_generator
core_eia860__assn_boiler_stack_flue
core_eia860__assn_yearly_boiler_emissions_control_equipment
core_eia861__assn_balancing_authority
core_eia861__assn_utility
core_epa__assn_eia_epacamd
core_epa__assn_eia_epacamd_subplant_ids
core_pudl__assn_eia_pudl_plants
core_pudl__assn_eia_pudl_utilities
core_pudl__assn_ferc1_dbf_pudl_utilities
core_pudl__assn_ferc1_pudl_plants
core_pudl__assn_ferc1_pudl_utilities
core_pudl__assn_ferc1_xbrl_pudl_utilities
out_pudl__yearly_assn_eia_ferc1_plant_parts
```

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -124,6 +127,34 @@
         "etl_group": "outputs",
         "field_namespace": "eia",
     },
+    "out_eia__yearly_assn_plant_parts_plant_gen": {
```
zaneselvans, line 130: Is it clear whether this should be an `out_eia__` or an `out_pudl__` table?  The FERC-EIA plant parts association is currently `out_pudl__` but the plant parts table itself is `out_eia__`.  The fields on both sides of the association here are ultimately somehow derived from EIA, but the association itself feels like a PUDL construction.

```
out_eia__yearly_plant_parts
out_pudl__yearly_assn_eia_ferc1_plant_parts
```

src/pudl/metadata/fields.py
cmgosnell: all of these new columns only show up in this association table. So this description will only ever show up with the `out_eia__yearly_assn_plant_parts_plant_gen` table. So i'm not sure adding the table name here is helpful - we don't do that anywhere else.

I am not giddy about the description. I do like your suggestion here and otherwise to use `record_id_eia_plant_gen` instead of `plant_gen` for consistency. Is it the word "associated" that is throwing you off here? Are these better?

1. `record_id_eia_plant_gen`'s code representing the most predominant type of energy that fuels the generator.
2. Code representing the most predominant type of energy that fuels the `record_id_eia_plant_gen`'s generator.

src/pudl/metadata/fields.py
cmgosnell: your not wrong here... both the `report_date` column and the `plant_id_eia` are always consistent across ppe records within a given plant/year. I personally lean towards making all of the columns consistent and clear in that they are columns from the plant_gen record, even though this is a little extraneous. I mostly just think being inconsistent about the suffix will be harder to explain.

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
cmgosnell: good question! but no the report date doesn't need to be in there because there is a year element in the record_id_eia*

src/pudl/analysis/plant_parts_eia.py
cmgosnell: its just a merge suffix change within `match_to_single_plant_part`. I've made sure all of the use cases of `match_to_single_plant_part` all have the same names coming out of it, which was simple because its only being used two other times and they both already had renames here. The renames are now more simple bc as you can see we were removing the `_og` suffix before.

cmgosnell:
Also idk where @zaneselvans 's comment about the order of assn & yearly went. but i took the convention cue from `out_pudl__yearly_assn_eia_ferc1_plant_parts`

I was a little surprised to see the yearly before the assn tbh. I will use this `assn_yearly` convention for this table but I do not want to change `out_pudl__yearly_assn_eia_ferc1_plant_parts` without some warning to users using this table name and that feels oos here.

cmgosnell:
@aesharpe on your owned vs total question. Within `match_to_single_plant_part`, we use the `IDX_OWN_TO_ADD` (which is `["utility_id_eia", "ownership_record_type"]` to make these associations, but yea right now I didn't keep these two columns in the table. It might be good to keep them in there just to be explicit.

and maybe then if we have these two columns plus `report_date` and `plant_id_eia`, which are all columns that are necessarily the same values for any given `record_id_eia` and `record_id_eia_plant_gen`. So perhaps it does make sense to not add the `plant_gen` suffix to all four of those columns

zaneselvans:
I removed my comment because it was inspired by an older commit, in which `_assn` was at the end of the table name. I agree that fixing the mixed ordering of yearly / assn across all the tables is out of scope.

Not adding the suffixes to the columns which must always be the same between the two sides of the association seems good to me.

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
cmgosnell: i've updated the description a bunch!

src/pudl/metadata/fields.py
aesharpe: I like number 2!

Yeah, the word associated was throwing me off a bit. If you wanted to make it even simpler, you could just have the same definition as the non `plant_gen` column and then add the sentence: "Associated with the `record_di_eia_plant_gen` record at the end?

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -124,6 +127,41 @@
...
+        "description": """In order to easily determine what generator records are associated with every
+plant part record, we made this association table. This table associates every plant part
+record (identified as ``record_id_eia``) from the ``out_eia__yearly_plant_parts`` table to the associated
+'plant_gen' records (identified as ``record_id_eia_plant_gen``). The plant part records have a one to many
```
aesharpe, line 134: Maybe just add "from the same table" at the end of the sentence.

aesharpe (approved):

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -702,23 +674,14 @@ def add_one_to_many(
         plant_parts_eia_filt = plant_parts_eia.loc[
             plant_parts_eia.record_id_eia.isin(one_to_many.record_id_eia)
         ]
-
-        if plant_parts_eia_filt.empty:  # If 1:m matches not in plant_part subset
-            PLANT_PARTS.pop("plant_match_ferc1")  # Remove from rest of workflow
-            return plant_parts_eia
-
```
cmgosnell, line 709: this pop was the culprit of my lost column woes. I was previously using a function to grab the `id_cols` from `PLANT_PARTS`.

This is not a necessary step at all. I traced the following steps of building the ppe and rebuilt the fast and full version of this ppe and this new assn table.

cmgosnell:
This failed in the merge queue yesterday. it looks like it may be a memory constraint problem..... but I am not sure so I'm running it again to see if it was a one-off (the ci tests all passed locally)

zaneselvans:
Sadly, it seems reproducible. I'll try bumping the large runner to the next size up and kicking it off with a manual `workflow_dispatch` just to confirm that this is really the problem.  If we did that across the board, it would increase our CI costs from ~$50/mo to ~$100/mo.

zaneselvans:
Yep, [the test run with an 8CPU runner passed](https://github.com/catalyst-cooperative/pudl/actions/runs/10531990266)

cmgosnell:
okay since this is failing right at `fast_out_annual.plant_parts_eia`, I went and checked if there was any substantive tests happening on this ppe table. its only checking if it is non-null. This `PudlTabl` based test is a type of test that we want to rip out entirely and its expensive to grab this giant table for this little check. So I am trying cutting it from this test and turning the runner size back down.

PR #3780, opened 2024-08-10
added description of Data Source Heading

Nancy9ice:
# Overview

Closes one of the Google Season of Docs Tasks

This adds a description for the Data source heading in the readthedocs to aid clarity for users.

# Testing

I ran the command `sphinx-build -b html docs docs/_build/html` and checked if my changes reflected correctly locally

@aesharpe @zaneselvans


.gitignore
```diff
@@ -43,6 +43,7 @@ terraform/.terraform/*
 .env
 .hypothesis/
 
+*.rst
```
aesharpe, line 46: I don't think we want this in the final version?

docs/data_sources/index.rst
```diff
@@ -3,6 +3,8 @@
 Data Sources
 ============
 
+The following data sources serve as the foundation for our data pipelines.
```
aesharpe, line 6: Right now it's just one data pipeline, so I would keep "pipelines" singular.

.gitignore
zaneselvans: I agree with @aesharpe here -- we have many RST files which are checked into git and should be tracked.

Nancy9ice: You might have forgotten @aesharpe but when you, Ella and I had a conversation about how about 150 rst files were pushed when I made my first changes on pudl, I suggested that maybe we could add it to the gitignore file. but since @zaneselvans has already confirmed that many .rst files still need to be tracked. I'll remove that and push the changes

aesharpe (approved):

PR #3786, opened 2024-08-13
edited dev setup docs

Nancy9ice:
# Overview

On the development setup, Indicate that users should install ‘pre-commit’ before they use the ‘pre-commit install’ command

This helps users know the command to run if their `pre-commit install` command doesn't work

# Testing

I ran the command `sphinx-build -b html docs docs/_build/html` and checked if my changes reflected correctly locally

@aesharpe @zaneselvans


e-belfer (changes requested):
@Nancy9ice If `pre-commit` isn't installed by following the set-up instructions above, this is either a documentation problem, a local environment issue or a problem with how we've configured `conda-lock.yml`. Either way, we shouldn't instruct people to fix the problem by pip installing missing packages.

Are you able to reproduce the steps you took that led to you not having pre-commit as expected when you first set up your environment?

e-belfer:
Here's what I tried to replicate this problem:
I verified I did not have `pre-commit` installed locally. I ran `pre-commit uninstall` on the `pudl-dev` environment, removed my `pudl-dev` environment, re-ran `make install-pudl`. As expected, I was able to access the `pre-commit` package directly after environment installation.

It's possible that this is an OS-specific issue, or that there's some unclear documentation here. But either way, we'd want to fix those problems, rather than instructing users to install the package manually. So, I'm closing this PR. @Nancy9ice if you're able to reproduce the issue locally or have more information for me, let me know and we can re-open the PR as needed.

PR #3788, opened 2024-08-13
Fix `release.yml` fork behaviors

e-belfer:
# Overview

Our `release.yml` workflow doesn't currently trigger from PRs merging in for a fork.

What did you change?
- Updated `on` condition to include `pull_request`
- Updated conditional to correctly identify forks vs non-forked repos

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,13 +1,15 @@
 ---
 name: release
 
-on: push
+on:
+  pull_request:
+  push:
 
 jobs:
   build-distribution:
     name: Build catalystcoop.pudl distribution for PyPI
     runs-on: ubuntu-latest
-    if: github.repository == 'catalyst-cooperative/pudl'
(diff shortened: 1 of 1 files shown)


e-belfer:
The `github.repository` behavior works as expected here. But we still see "expected - waiting for status to be reported". Which means that the issue here likely doesn't have to do with the .yml file itself.

jdangerx (approved):
Whee! It seems to work!

cc @zaneselvans - I remember there being *some* reason that we wanted this to be `on: push` instead of `on: pull_request` but I don't remember - do you?

(and we have this as `on: pull_request` here because otherwise the workflow only runs on the head repo, not the base repo - so then the job never triggers on our side / the required check never reports status, and hangs in "expected")

zaneselvans:
Hmm. I don't remember why we would want it to be `on: push` instead of `on: pull_request`. If this works then let's go with it!

Issue #3576, opened 2024-04-17
NREL ATB data non-unique in 2019 (mostly) and 2020


cmgosnell:
In trying to integrate NREL ATB, I ran into an oddity that made it difficult to integrate the data from 2019 and 2020. Because of this we are not integrating these years of ATB. All of this exploration was based on this semi-cleaned asset: [_core_nrelatb__transform_start.pkl.zip](https://github.com/catalyst-cooperative/pudl/files/15016920/_core_nrelatb__transform_start.pkl.zip)

There is a column called `core_metric_case` which is always either "Market" or "R&D". Then there is another column called `core_metric_key` which is a composite (semi-)primary key column containing codes representing info stored in other columns in the data. The first character of the `core_metric_key` is always either `R`or `M`. We've called this first letter of the `core_metric_key` the `mystery_code`. We and other collaborators thought this corresponded to the `core_metric_case`. It does the ~75% of the time:

```python
>>> _core_nrelatb__transform_start[["mystery_code", "core_metric_case"]].value_counts()
mystery_code  core_metric_case
R             R&D                 728007
M             Market              453201
R             Market              330483
M             R&D                  55677
Name: count, dtype: int64
```

This would be fine if the values in the `value` column did not vary based on the `mystery_code` (we could drop fully duplicate records w/o this `mystery_code` or the `core_metric_key`). But the data does seem to value truly different values. Of the three data tables which are derived from the info in the `value` column, two tables have real variability in `value` based on the `mystery_code`. The records that are variable by `mystery_code` make up 12% of the `core_nrelatb__yearly_projections_by_scenario` table and 5% of the `core_nrelatb__yearly_rates_projections` table.

![image](https://github.com/catalyst-cooperative/pudl/assets/25487998/9e8001a5-434f-4fbf-ad9d-f3a8a172a13c)
![image](https://github.com/catalyst-cooperative/pudl/assets/25487998/2bd26e5d-7da7-4337-968c-f83f74b4f916)

PR #3791, opened 2024-08-15
Raise error if FERC1 column renames don't match expected data

### LINKED ISSUE #3756: Make FERC column mapping warnings into errors
In the processing of the FERC Form 1 there are a number of column renames where the column that's being renamed doesn't exist, probably because there have been name changes in the FERC Form 1 data or our column renames. This results in null values... and right now we typically overlook the warnings.

We should validate two types of column renames:

- The columns we're attempting to rename actually exist
- The columns we don't explicitly rename don't change names over time (something we capture explicitly in spreadsheet extraction by mapping every column we extract)

There are a number of places in the `pudl.extract.classes` module where we currently have `logger.warning()` but might actually want to use `raise ValueError()`. The clearest is probably the case where we are attempting to rename columns that do not exist in the dataframe.

Places where that currently results in errors (see the [`ferc1-column-errors`](https://github.com/catalyst-cooperative/pudl/tree/ferc1-column-errors) branch)

### `core_ferc1__yearly_sales_by_rate_schedules_sched304` (missing columns):
* `sales_axis`

### `core_ferc1__yearly_utility_plant_summary_sched200` (missing columns):
* `amortization_of_underground_storage_land_and_land_rights_utility_plant_in_service`
* Note that a column named `amortization_of_underground_storage_land_and_land_rightsutility_plant_in_service` does appear in the dataframe.

### `core_ferc1__yearly_balance_sheet_assets_sched110` (missing columns):
* `preliminary_natural_gas_survey_and_investigation_charges_and_other_preliminary_survey_and_investigation_charges`

e-belfer:
# Overview

Closes #3756 

Adds two validations into the FERC transforms:

1. If we're attempting to rename a column, that column should exist in the raw data.
2. All expected factoids are added as ENUMS in `FIELD_METADATA_BY_RESOURCE` for their respective tables. (Note that enums also log a warning if there are values that aren't in the constraints, but as all these values are FK values they'll raise a primary key error when `wide_to_tidy` is run).

What did you change?
Added one validation assertions. Debugged some renaming issues outlined in #3576. Constrained the factoid column for each FERC table.

# Testing

Generate all `core_ferc1` assets and anything downstream.

```[tasklist]
# To-do list
```


src/pudl/transform/classes.py
```diff
@@ -260,6 +260,10 @@ class RenameColumns(TransformParams):
     """
 
     columns: dict[str, str] = {}
+    # A dictionary of columns to be renamed.
+    not_renamed_columns: list[str] = []
+    # A list of raw columns which are expected not to be renamed. Any other
+    # columns in the raw data which fail to be renamed will raise an error.
```
zaneselvans, line 266: If we make these comments into triple-quoted strings they'll show up in the documentation as docstrings on these attributes, which would be wonderful.

src/pudl/transform/classes.py
```diff
@@ -1230,15 +1234,32 @@ def rename_columns(
...
+                .difference(set(params.columns))
+                .difference(set(params.not_renamed_columns))
             )
+            logger.warn(f"unmapped_cols is {unmapped_cols}")
+            if unmapped_cols:
+                raise ValueError(
+                    f"{self.table_id.value}: The following columns are present in the"
+                    " in the dataframe but not renamed.\n"
+                    f"Unmapped columns: {unmapped_cols}\nExisting Columns: {df.columns}"
+                )
```
zaneselvans, line 1262: Does this check necessitate a bunch of new transformation metadata? If so it'd be great to catch both extra and missing columns. I noticed that we check for both kinds of errors in the EIA data but only one kind of error here, but it seemed like it might be a hassle to compile and verify the necessary metadata.

```suggestion
            if unmapped_cols:
                raise ValueError(
                    f"{self.table_id.value}: The following columns are present in the"
                    " dataframe but are not renamed.\n"
                    f"Unmapped columns: {unmapped_cols}\nExisting Columns: {df.columns}"
                )
```

src/pudl/transform/params/ferc1.py
```diff
@@ -2623,7 +2623,93 @@
                     "adjustments": "adjustments",
                     "transfers": "transfers",
                     "ending_balance": "ending_balance",
-                }
+                },
+                "not_renamed_columns": [
```
zaneselvans, line 2627: Lol okay it does necessitate a bunch of new metadata.

I assume that you've programmatically generated these lists of not renamed columns. Are they just all of the columns that weren't getting renamed? If so, do we need to worry that we might be codifying errors in specifying them now? Like, are any of these columns that we weren't renaming actually columns that we *should* have been renaming, but weren't resulting in their contents getting lost when we eventually `enforce_schema()` at the end of the transform step?

e-belfer: Yes, these are all generated programatically. I might suggest sorting the columns in the error message, to make it easier to spot near-duplicates/typos.

However, the overwhelming majority of these columns are factoid names, which we're collapsing into the `xbrl_factoid` column, so they wouldn't be affected by `enforce_schema`. Several others are not and should probably be checked.

e-belfer:
@cmgosnell I still want to kick off a full build once the coverage issues are fixed to ensure I haven't introduced any downstream changes, but I implemented the changes discussed so it should be ready for review otherwise!

src/pudl/metadata/fields.py
```diff
@@ -196,6 +202,25 @@
     "asset_type": {
         "type": "string",
         "description": "Type of asset being reported to the core_ferc1__yearly_balance_sheet_assets_sched110 table.",
+        "constraints": {
+            "enum": ASSET_TYPES_FERC1.extend(
+                # Add all possible correction records into enum
```
cmgosnell, line 207: not blocking at all: for all of these corrections, it am tempted to suggest adding a correction for all of them. i know not all factoids get corrections because not all factoids are calculated (and even within those not all of them get corrections bc some calculated fields are surprisingly clean). but i see why you wouldn't do this and don't feel strongly about it either way.

```python
ASSET_TYPES_FERC1 + ["{t}_correction" for t in ASSET_TYPES_FERC1]
```

cmgosnell (approved):
its a little funny that my suggestion to swap the not-being-renamed columns to enums results in more lines of code being changed.... but lol i think this is still a much better solution and the thing we really care about.

I made one small idek if its a good suggestion definitely not blocking feel free to ignore my tricks to reduce lines but not actually reduce complexity.

Issue #3757, opened 2024-08-05
Make Excel & CSV column mapping warnings into errors


zaneselvans:
Occasionally spreadsheet headers change in the EIA inputs, and currently this generates a warning, but not an error... but it really needs to be an error, because if we don't address it, we end up dropping a bunch of data that actually exists.  Sometimes we catch these issues when updating the data, but sometimes we do not!

See the [`xlsx-column-errors`](https://github.com/catalyst-cooperative/pudl/tree/xlsx-column-errors) branch where these errors are highlighted.

* The change from logging a warning to raising an error can be made in the base Extractor class that we use for all of our spreadsheet based data sources.
* There will need to be some (per data source?) special cases to ignore -- e.g. in the EIA-923 there's an "Early Release" warning that shows up as a column header, but is really just meant to flag that the whole spreadsheet is Early Release data.
* Will potentially affect EIA 176, 860, 860M, 861, and 923, as well as PHSMA data.

# Identified Errors

## EIA-176/191/757a
* All good!

## EIA-860/860M
* All good!

## EIA-861
* All good!

## EIA-923
* byproduct_expenses_and_revenues/2009:
  *  Extra columns: `unnamed_30`
* boiler_fuel/2023
  * Extra columns: `early_release_data_july_2024_not_fully_edited_use_with_caution_do_not_aggregate_to_state_regional_or_national_totals`

## PHMSA Natural Gas
* yearly_transmission_gathering_preparer_certification/2000: Expected columns not found in extracted table:
  * `preparer_telephone`
  * `preparer_name`
* yearly_distribution/(2004-2007): Extra columns found in extracted table:
  * `unnamed_227`
* yearly_distribution/2009: Extra columns found in extracted table:
[...]

PR #3820, opened 2024-08-27
Fix column mapping errors

cmgosnell:
# Overview

Closes #[3757](https://github.com/catalyst-cooperative/pudl/issues/3757).

What did you change?

# Testing

```[tasklist]
# To-do list
- [x] Re-ran the full etl locally
```


src/pudl/extract/eia923.py
```diff
@@ -60,7 +60,6 @@ def process_raw(self, df, page, **partition):
                 f"column with {partition['year']}"
             )
             df.loc[mask, "report_year"] = partition["year"]
-        df = self.add_data_maturity(df, page, **partition)
```
cmgosnell, line 63: this was also added later on a few lines above

src/pudl/extract/phmsagas.py
```diff
@@ -52,6 +52,32 @@ def process_renamed(self, newdata: pd.DataFrame, page: str, **partition):
...
+        unnamed_columns = newdata.filter(like="unnamed").columns
+        if (page in unnamed_page_years) and (
+            int(partition["year"]) in unnamed_page_years[page]
+        ):
+            newdata = newdata.drop(columns=unnamed_columns)
+        elif not unnamed_columns.empty:
+            logger.warning(
+                "We found some unnamed columns that are probably not expected. "
+                f"Consider dropping them. Columns found: {unnamed_columns}"
+            )
```
cmgosnell, line 80: there were several instances of `unnamed_X` columns in PHMSA. I considered adding them all into the column maps but that would just mean definitely keeping track of trash columns. In 2009 there were a ton of extra columns seemingly from two records that had a multi-line comment that shifted the rest of the cells over. The options were to map a bunch of bad columns and maybe save a few records in a really bespoke way or just drop em all. this seemed like the easiest and most discrete way to drop em all rather than map these all in the column maps.

src/pudl/package_data/phmsagas/column_maps/yearly_gathering_pipe_miles_by_nps.csv
cmgosnell, line 1: weirdly most of the phmsa errors came from these `part{xxx}additional` columns that appear in the pdf instructions but not in the data themselves. these usually coorespond to other sizes of pipe.

src/pudl/extract/phmsagas.py
```diff
@@ -52,6 +52,32 @@ def process_renamed(self, newdata: pd.DataFrame, page: str, **partition):
                     f"\n{to_drop}"
                 )
                 newdata = newdata.drop(columns=to_drop, errors="ignore")
+        # there is an annoying middling number of columns in phmsa raw data that are unnamed
+        # and have a smattering of random values. we want to drop these guys. but we are going
+        # to enumerate what we expect to need to
```
zaneselvans, line 57: It seems like this is the only special case that you ended up needing to deal with all the current errors, which is great! Do you want to finish this thought that got left off mid-sentence in the comment?

test/unit/extract/csv_test.py
```diff
@@ -70,7 +70,7 @@ def test_extract():
             # Testing the rename
             GenericMetadata,
             "get_column_map",
-            return_value={company_field: "company_rename"},
+            return_value={"company_rename": company_field},
```
zaneselvans, line 73: I'm confused about what's going on with the swaparoo in this test.

zaneselvans (approved):
Left a couple of small non-blocking requests for clarification.

test/unit/extract/csv_test.py
cmgosnell: ah yes. the swaparoo needed to happen in this test because it was in the wrong order and it was making the unit tests fail bc this test actually runs an extraction validation which now has the column map failure.

src/pudl/extract/phmsagas.py
cmgosnell: ah yes lol my classic mid sentence walk away

cmgosnell: updated

zaneselvans:
Coverage dropped below the threshold partly because now the code where the expected and actual columns don't match is never hit, and partly because the special case for PHMSA is not exercised in the fast test.

Can we easily add a unit test that exercises the "columns don't match" behavior?

zaneselvans:
Yes we can. I think.

PR #3821, opened 2024-08-29
added primary key column in data dictionary page

Nancy9ice:
Overview
Include the primary keys of each table in the data dictionary page

The addition of a primary key column to the data dictionary page helps users to know the column(s) in a table that is/are primary key(s). That way, it will be easy for them to understand how to create relationships between tables.

Testing

I ran the command make docs-build and checked if my changes reflected correctly locally

@aesharpe @zaneselvans


e-belfer (changes requested):
I think that rather than adding a column that has a boolean for the primary keys, it would be easier to interpret a list of primary keys after the ``resource.description`` section.

Probably something like: "This table has the following primary keys: [get list from resource.schema.primary_key]" if there are primary keys, otherwise "This table doesn't have defined primary keys".

Were you intending to add foreign keys to this PR as well? I see it in the description of the PR, but not in the changes made. Totally fine if you want to split this into 2 PRs, just wanted to check! I'd suggest adding these in the same way, rather than in a table.

Nancy9ice:
> I think that rather than adding a column that has a boolean for the primary keys, it would be easier to interpret a list of primary keys after the `resource.description` section.
> 
> Probably something like: "This table has the following primary keys: [get list from resource.schema.primary_key]" if there are primary keys, otherwise "This table doesn't have defined primary keys".
> 
> Were you intending to add foreign keys to this PR as well? I see it in the description of the PR, but not in the changes made. Totally fine if you want to split this into 2 PRs, just wanted to check! I'd suggest adding these in the same way, rather than in a table.

Alright @e-belfer, I'll work on your comments. Yes, I'll add the foreign key to a different PR

docs/templates/resource.rst.jinja
```diff
@@ -4,6 +4,15 @@
 
 {{ resource.description | wordwrap(78) if resource.description else 'No table description available.' }}
 
+{% if resource.schema.primary_key -%}
+**The table has the following primary keys:**
+{% for key in resource.schema.primary_key %}
+  - {{ key }}
```
e-belfer, line 10:
```suggestion
* {{ key }}
```

docs/templates/resource.rst.jinja
```diff
@@ -4,6 +4,15 @@
 
 {{ resource.description | wordwrap(78) if resource.description else 'No table description available.' }}
 
+{% if resource.schema.primary_key -%}
+**The table has the following primary keys:**
```
e-belfer, line 8:
```suggestion
**The table has the following primary key columns:**
```

e-belfer (changes requested):
I'd suggest using the * instead of the - to avoid rendering a large grey box for each primary key bullet list. Otherwise, looks great!

aesharpe (approved):

aesharpe:
The pytest ci-coverage tests failed when we tried to integrate this, @zaneselvans any idea why? This PR has nothing to do with coverage as far as I know.

cmgosnell:
@aesharpe it looks like it was not a code coverage failure but rather a failure in [the process](https://github.com/catalyst-cooperative/pudl/actions/runs/10672940961/job/29582752826). i'm curious if it was just a fluke so i kicked off the test again

zaneselvans:
It looks like the coverage action expects to find `coverage/*/.coverage` files (which IIRC are the default output format, and are SQLite DBs) but the coverage outputs that seem to be present are `coverage/*/coverage.xml` which we generate to upload to CodeCov.  I'm not sure why this would have suddenly changed. The coverage library hasn't had a major version lately and neither has the github artifact downloading action that's used to bring the docs, unit, and integration test coverage files together in the coverage workflow.

I would try changing this line near the end of `.github/workflows/pytest.yml`:

```yml
          micromamba run -n coverage coverage combine coverage/*/.coverage
```
to

```yml
          micromamba run -n coverage coverage combine coverage/*/coverage.xml
```

and see if it behaves any differently.

zaneselvans:
Whatever is causing this issue seems to apply across the board, rather than to just this PR. I'm applying the change I suggested above in #3823 and re-submitting to the queue, so we'll see if it works.

PR #3822, opened 2024-08-30
Extract 714 xbrl

### LINKED ISSUE #3813: Extract the FERC 714 XBRL annual load forecast table
* In the CSVs exported from DBF is `Part 3 Schedule 2 - Planning Area Forecast Demand.csv`
* In the XBRL data this table is `planning_area_hourly_demand_and_forecast_summer_and_winter_peak_demand_and_annual_net_energy_for_load_table_03_2_duration`
* In PUDL this table is `core_ferc714__yearly_planning_area_demand_forecast`

There are a couple of other tables which also come from Part 3 Schedule 2, which is confusing.

### LINKED ISSUE #3814: Extract the FERC 714 XBRL hourly historic load table
* In PUDL we call this table `out_ferc714__hourly_planning_area_demand`
* In the CSV exported from DBF this table is found in "Part 3 Schedule 2 - Planning Area Hourly Demand.csv"
* In the FERC-714 XBRL data the relevant tables are:
  * `planning_area_hourly_demand_and_forecast_summer_and_winter_peak_demand_and_annual_net_energy_for_load_03_2_duration` (only contains abbreviations like "EDT" which are needed to adjust all the timestamps)
  * `planning_area_hourly_demand_and_forecast_summer_and_winter_peak_demand_and_annual_net_energy_for_load_03_2_instant` (contains the hourly data)
* The biggest hassle with this table will probably be localizing the timestamps. We tried many complicated approaches initially in `pudl.transform.ferc714` and none of them were awesome, so we fell back on something kind of dumb but simple.  The only indications of what timezone and daylight savings vs. standard time the timestamps correspond to come from the abbreviations in the duration table. They may or may not have 23/25 hours on days with the time change -- those timestamps can have a gap or a duplicate value.
* There are also some sweeping manual fixes being applied to the older hourly load data, e.g. in some cases they flip the sign of the number they're reporting for a year here or there, and we flip it back to align with all the rest of the data.

### LINKED ISSUE #3815: Extract the FERC 714 XBRL repondent ID table
* In PUDL we call this table `core_ferc714__respondent_id`
* In the CSVs exported from DBF this table is "Respondent IDs.csv" Note that there is no annual reporting of this data -- the old FERC-714 data comes as one big dump containing all years, and there's nothing like a `report_date` in the `respondent_id` table.
* In the XBRL data this table is named: `identification_and_certification_01_1_duration` and it has a lot more columns than in the old FERC-714 data. Some of these values look like they're probably constant.
* In the old FERC-714 data, it was unclear whether the `eia_code` referred to EIA Balancing Authorities or Utilities, which actually have separate sets of IDs (even though the same integers are often used to refer to a BA and a Utility, when the Utility also acts as a BA, which was very confusing).
* The old and new respondent tables are pretty different from each other. The old table has no sense of time passing -- it claims to apply to all years of data. The new table has significantly more columns, and the potential to link to both balancing authorities and "planning areas" in other datasets.
* In the new XBRL data, most respondents report an EIA ID, but because the new form doesn't specifically request an EIA ID, there are also some other totally random kinds of IDs.
* We didn't try to decide whether the EIA IDs that were reported referred to a utility or a balancing authority (since both seemed to be showing up sometimes) and instead identified both what Utility and what Balancing Authority the respondent **would** correspond to, based on the ID.
* The fact that the old Respondent ID table has no sense of time, and the new one will (since it's reported every year) will probably complicate the integration of this table.

aesharpe:
# Overview

closes #3813
closes #3814
closes #3815

We don't currently extract XBRL tables for the FERC 714 data. This code adds the capability to do so.

## What did you change?

**`pudl/extract/ferc714.py`**
* Add `_csv` suffix to csv-specific extraction functions/variables.
* Rename csv-specific encoding dictionary to `FERC714_CSV_ENCODING`
* Add `TABLE_NAME_MAP_FERC714` dictionary
* Add a function called `create_raw_ferc714_xbrl_assets` to reference the raw xbrl tables in the FERC 714 SQLite db and call it.
* Add a function called `raw_ferc714_xbrl__metadata_json` to make an asset for the metadata table

**`pudl/io_managers.py`**
* Upgrade the `FercXBRLSQLiteIOManager` to accommodate more than just `ferc1`
* Add a `ferc714_xbrl_sqlite_io_manager` function

**`pudl/settings.py`**
* Add `xbrl_years` functino to the Ferc714Settings class to be able to access XBRL years. 
* Remove comments about not processing the XBRL data.

**`pudl/metadata/sources.py`**
* Update the FERC714 working partitions.
* Remove comments about not processing XBRL data.

# Testing

Load each of the raw assets via and make sure they contain the expected years: 
```
defs.load_asset_value("raw_ferc714_xbrl__table_name")
```

```[tasklist]
# To-do list
```


src/pudl/extract/ferc714.py
```diff
@@ -88,19 +136,80 @@ def _extract_raw_ferc714(context):
...
+    """Create SourceAssets for raw FERC 714 XBRL tables.
+
+    SourceAssets allow you to access assets that are generated elsewhere.
+    In our case, the XBRL database contains the raw FERC 714 assets from
+    2021 onward. Prior to that, the assets are distributed as CSVs and
+    are extracted with the ``raw_ferc714_csv_asset_factory`` function.
+
+    Returns:
+        A list of FERC 714 SourceAssets.
+    """
```
aesharpe, line 198: I'm not sure how to test whether this works because it doesn't create assets, rather references to non-Dagster tables.

src/pudl/io_managers.py
```diff
@@ -855,12 +855,13 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
-        # TODO (daz): this is hard-coded to FERC1, though this is nominally for all FERC datasets.
-        ferc1_settings = context.resources.dataset_settings.ferc1
+        ferc_settings = context.resources.dataset_settings[
+            self.db_name.replace("_xbrl", "")
+        ]
 
         table_name = get_table_name_from_context(context)
         # Remove preceeding asset name metadata
-        table_name = table_name.replace("raw_ferc1_xbrl__", "")
+        table_name = table_name.replace(f"raw_{self.db_name}__", "")
```
aesharpe, line 864: This was an attempt to not hard-code FERC1 to this function, but it causes issues with the tests in `test/unit/io_manager`.

The `test_ferc_xbrl_sqlite_io_manager_dedupes` function sets the `db_name` to `test_db` which causes problems here. Need to figure out a good work-around. Not sure whether to fix the tests or fix this reference.

cmgosnell: this seems like a good change! imo we should edit the test to enable it to deal with this change

src/pudl/extract/ferc714.py
cmgosnell:
```python
from pudl.extract.ferc714 import FERC714_XBRL_FILES
raw_table_names = [
        v for value in FERC714_XBRL_FILES.values() for v in value.values()
    ]
defs.load_asset_value(f"raw_ferc714_xbrl__{raw_table_names[0]}")
```

src/pudl/io_managers.py
```diff
@@ -855,12 +855,13 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
             context: dagster keyword that provides access output information like asset
                 name.
         """
-        # TODO (daz): this is hard-coded to FERC1, though this is nominally for all FERC datasets.
-        ferc1_settings = context.resources.dataset_settings.ferc1
+        ferc_settings = context.resources.dataset_settings[
+            self.db_name.replace("_xbrl", "")
+        ]
```
cmgosnell, line 860: the `dataset_settings` here is a pydantic object so you can't access this directly like a dictionary. but you can do this:
```suggestion
        ferc_settings = context.resources.dataset_settings.get_datasets()[
            self.db_name.replace("_xbrl", "")
        ]
```

src/pudl/extract/ferc714.py
```diff
@@ -63,17 +68,60 @@
...
+    {
+        "core_identification_and_certification_01_1": {
+            "duration": "identification_and_certification_01_1_duration"
+        },
+        "core_generating_plants_included_in_reporting_balancing_authority_area_02_1": {
+            "duration": "generating_plants_included_in_reporting_balancing_authority_area_02_1_duration"
+        },
+        "core_generating_plants_included_in_reporting_balancing_authority_area_totals_02_1": {
+            "duration": "generating_plants_included_in_reporting_balancing_authority_area_totals_02_1_duration"
+        },
```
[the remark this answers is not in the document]
cmgosnell: i'd suggest doing something 
```python
{" core_table_name:" { 
    "dbf": {raw_table_name: 
        {"name:" string, "encoding": jabfka}},
    "xbrl": raw_table_name main slug }
}
```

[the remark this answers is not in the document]
aesharpe: Should I do this even though it's csv not dbf for 714? Or should I replace dbf with csv in this case?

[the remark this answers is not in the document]
aesharpe: > another thing that the ferc1 extract does is to only write out the main bit of the raw xbrl table name and let raw_ferc1_xbrl__metadata_json handle whether there is an instant of a duration table using squash_period. rn there is a lot of repetition here with the main chunk of the table name. i haven't looked at the 714 metadata but if it is even semi similarly strucutred.

My initial thinking here was to create a dictionary that wouldn't produce NA values for non-existent instant or duration tables in the metadata (which the squash period does). But given our conversation about dictionary/metadata consistency, I think it's fine to revert to using this function.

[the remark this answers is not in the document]
cmgosnell: oh lol yes I meant to use `"csv"` instead of `"dbf"` in my example dicts.

src/pudl/io_managers.py
aesharpe: done!

src/pudl/settings.py
```diff
@@ -139,12 +139,16 @@ class Ferc714Settings(GenericDatasetSettings):
...
-    # monolithic CSV files, so asking for any year processes all of them.
     years: list[int] = data_source.working_partitions["years"]
 
+    # The older 714 data is distributed as CSV files and has a different extraction
+    # process than the FERC DBF extraction process.
+
+    @property
+    def xbrl_years(self):
+        """Return validated years for which XBRL data is available."""
+        return [year for year in self.years if year <= 2021]
```
zaneselvans, line 150: I think you mean `year >= 2021` rather than `year <= 2021` here, and it might be why the unit tests are failing.

```suggestion
        return [year for year in self.years if year >= 2021]
```

aesharpe: good catch!

aesharpe: Sadly not why the unit tests are failing though.

aesharpe: Unit test failures have something to do with this snipit I changed in `io_managers.py`: 

```
        ferc_settings = context.resources.dataset_settings.get_datasets()[
            self.db_name.replace("_xbrl", "")
        ]
```

The `pytest test/unit/io_managers_test.py::test_ferc_xbrl_sqlite_io_manager_dedupes` function does not like this.

aesharpe: Ok I fixed it! changed that snipit to: 

```
        ferc_settings = getattr(
            context.resources.dataset_settings, self.db_name.replace("_xbrl", "")
        )
```

aesharpe:
Unit test failures have something to do with this snipit I changed in io_managers.py:
```
        ferc_settings = context.resources.dataset_settings.get_datasets()[
            self.db_name.replace("_xbrl", "")
        ]
```
The `pytest test/unit/io_managers_test.py::test_ferc_xbrl_sqlite_io_manager_dedupes` function does not like this.

**UPDATE:**

I fixed this by changing this snipit to: 

```
        ferc_settings = getattr(
            context.resources.dataset_settings, self.db_name.replace("_xbrl", "")
        )
```

src/pudl/extract/ferc714.py
```diff
@@ -88,19 +111,97 @@ def _extract_raw_ferc714(context):
...
 
-    return _extract_raw_ferc714
+    return _extract_raw_ferc714_csv
+
+
+@asset
+def raw_ferc714_xbrl__metadata_json(
+    context,
+) -> dict[str, dict[str, list[dict[str, Any]]]]:
+    """Extract the FERC 714 XBRL Taxonomy metadata we've stored as JSON.
```
aesharpe, line 131: This is basically copied verbatim from the ferc1 version of this.

src/pudl/extract/ferc714.py
```diff
@@ -88,19 +111,97 @@ def _extract_raw_ferc714(context):
...
+    return xbrl_meta_out
+
+
+def create_raw_ferc714_xbrl_assets() -> list[SourceAsset]:
+    """Create SourceAssets for raw FERC 714 XBRL tables.
 
+    SourceAssets allow you to access assets that are generated elsewhere.
+    In our case, the XBRL database contains the raw FERC 714 assets from
+    2021 onward. Prior to that, the assets are distributed as CSVs and
+    are extracted with the ``raw_ferc714_csv_asset_factory`` function.
```
aesharpe, line 179: This is also copied over from the ferc1 module.

test/unit/io_managers_test.py
```diff
@@ -360,12 +360,12 @@ def test_error_when_reading_view_without_metadata(fake_pudl_sqlite_io_manager_fi
 
 
 def test_ferc_xbrl_sqlite_io_manager_dedupes(mocker, tmp_path):
-    db_path = tmp_path / "test_db.sqlite"
+    db_path = tmp_path / "ferc1_xbrl.sqlite"
```
cmgosnell, line 363: i think this change is unnecessary and a little bad. this test actually adds the little test table into the database (via the line `df.to_sql("test_table_instant", conn)`). So whenever this unit test is run a db table is being added to the main `ferc1_xbrl` db.

Also, I did a quick poke at it and i don't think the way you've currently changed the io_managers requires this name change.

src/pudl/settings.py
cmgosnell, line 1: do we also need `csv_years`? it looks like we probably should because inside `raw_ferc714_csv_asset_factory` we'd now be trying to loop over the xbrl years as well.

cmgosnell (changes requested):
two requests for changes in comments, plus the settings yml files need to updated! but those seem like minor tweaks overall this is looking great imo and I was able to materialize these xbrl raw assets locally

test/unit/io_managers_test.py
aesharpe: Unfortunately the test fails if the sqlite isn't named the same as `db_name`. I'll look into how we can work around this.

cmgosnell: i gave this a quick test locally and it looks like if you rename all of the instances in this test it works fine.

aesharpe: Can you push that code? I keep getting a `ValueError`

cmgosnell (changes requested):
two blocking needs here:

to change 
`years = ", ".join(map(str, ferc714_settings.years))`
to
` years = ", ".join(map(str, ferc714_settings.csv_years))`

in _extract_raw_ferc714_csv

and again to add the new years in the settings yml files.

aesharpe:
> and again to add the new years in the settings yml files.

The reason I didn't do this is because I think it will break the transforms and I didn't want to merge it into `main` that way.

cmgosnell:
> The reason I didn't do this is because I think it will break the transforms and I didn't want to merge it into main that way.

hm this is a good point... but I think in this context because the only existing transforms are based on the csv extracted tables they will be fine. but still i hadn't put it in context that this was only the extraction so i will renege on this suggestion.

cmgosnell (approved):

aesharpe:
@zaneselvans thank you for that suggestion! I just had to add one more fixture for `ferc714_xbrl_extract` and it was good to go.

aesharpe:
mmm getting a unit test failure seemingly related to timeseries. Is timeseries cleaning part of the extraction or transformation phase @zaneselvans? 

```
        for method in "tubal", "tnn":
            # Impute null values
            imputed0 = s.impute(mask=mask, method=method, rho0=1, maxiter=1)
            imputed = s.impute(mask=mask, method=method, rho0=1, maxiter=10)
            # Deviations between original and imputed values
            fit0 = s.summarize_imputed(imputed0, mask)
            fit = s.summarize_imputed(imputed, mask)
            # Mean MAPE (mean absolute percent error) is converging
>           assert fit["mape"].mean() < fit0["mape"].mean()
E           assert 0.030191300830775987 < 0.024084802339886784
E            +  where 0.030191300830775987 = mean()
E            +    where mean = 0    0.002769\n1    0.001304\n2    0.011850\n3    0.002106\n4    0.008777\n5    0.007833\n6    0.135690\n7    0.001992\n8    0.118109\n9    0.011482\nName: mape, dtype: float64.mean
E            +  and   0.024084802339886784 = mean()
E            +    where mean = 0    0.010499\n1    0.011460\n2    0.010525\n3    0.005897\n4    0.007852\n5    0.015753\n6    0.054820\n7    0.022598\n8    0.045421\n9    0.056023\nName: mape, dtype: float64.mean

test/unit/analysis/timeseries_cleaning_test.py:109: AssertionError
```

zaneselvans:
Oh, the timeseries issue is probably a rare stochastic failure -- it happens once in a blue moon, probably having something to do with the seed being used for the random number generator. If you re-run the unit tests I suspect they'll pass just fine.

aesharpe:
> Oh, the timeseries issue is probably a rare stochastic failure -- it happens once in a blue moon, probably having something to do with the seed being used for the random number generator. If you re-run the unit tests I suspect they'll pass just fine.

That is wild and also strangely comforting. I'm glad I didn't try and preemptively figure this out...haha

zaneselvans:
I spent a bunch of time at some point hunting down all the RNG initializations and tried to give them fixed seeds in the tests so we always get the same outputs, but I must have missed one somewhere, and it bites us every so often. It just didn't seem like it was worth continuing the hunt because it's so infrequent.

zaneselvans:
Especially in the unit tests, whenever there's something like this that seems totally out of left field and unrelated to what's being changed my first instinct is just to re-run the tests to see if it's real. Doesn't take much time or effort, and often it works.

PR #3823, opened 2024-09-02
Update conda lockfile for week of 2024-09-02

pudlbot:


zaneselvans (approved):

zaneselvans:
I don't actually expect this to work, since it looks like combining the `coverage.xml` files won't work, and the `.coverage` files are apparently not getting uploaded as artifacts, even though they [were last week](https://github.com/catalyst-cooperative/pudl/actions/runs/10646946361/job/29514358948#step:8:16). But I added some additional logging output that should show whether the `.coverage` files are getting created at all.

zaneselvans:
Looking at the new logging output from `pytest-unit` it looks like the `.coverage` output isn't being created at all, even though the `coverage.xml` report ultimately is, and gets uploaded as an artifact to be combined later.

We've been using
- `coverage==7.6.1`
- `pytest-cov==5.0.0`

for weeks at least, so it doesn't seem like changes to those libraries are the issue.

zaneselvans:
Picking apart the pytest coverage action a little bit, it seems like we've been unnecessarily generating these `coverage.xml` files -- they don't get used for anything, and the `.coverage` SQLite output is combined and reported as XML in the final `ci-coverage` job so it can be uploaded to CodeCov. So I have removed all of the other XML stuff to simplify the situation. Maybe it was interfering somehow with the generation of `.coverage` outputs?

zaneselvans:
Oooookay, [here's the culprit](https://github.blog/changelog/2024-08-19-notice-of-upcoming-deprecations-and-breaking-changes-in-github-actions-runners/):

> Exclude hidden files by default in Upload Artifact GitHub Actions
From September 2nd, 2024, we will no longer include hidden files and folders as part of the default upload of the v3 and v4 upload-artifact actions. This reduces the risk that credentials are accidentally uploaded into artifacts. Customers who need to continue to upload these files can use a new option, ‘include-hidden-files’, to continue to do so.

PR #3834, opened 2024-09-07
Update to Splink v4.0

### LINKED ISSUE #3735: Update to Splink v4.0
[Splink v4.0 was released](https://moj-analytical-services.github.io/splink/blog/2024/07/24/splink-400-released.html) on 2024-07-24. Mostly it updates the API to be cleaner and the codebase to be more maintainable, so this isn't an imperative -- there's no immediate change in functionality.  But we should make the minor changes required to work with the new version so that improved blocking, labeling functions, and other updates are available.

zaneselvans:
# Overview

Closes #3735

* [Splink v4](https://moj-analytical-services.github.io/splink/blog/2024/07/24/splink-400-released.html) has been out for a few months. This PR updates our imports and API calls to use the new version.
* The underlying methods haven't changed, but the code and API have been cleaned up, and there are some performance improvements.

# Testing

```[tasklist]
# To-do list
- [x] Update Splink imports to use the new API
- [x] Replace `settings_dict` with the new `SettingsCreator`
- [x] Use the new call signature for the `block_on()` function.
- [x] Get unit tests to passing
- [x] Get integration tests to passing
- [x] Run the `build-deploy-pudl` GitHub Action manually.
- [ ] Run the full ETL locally.
```


src/pudl/analysis/record_linkage/eia_ferc1_model_config.py
```diff
@@ -5,10 +5,9 @@
 model.
 """
 
-import splink.duckdb.comparison_level_library as cll
-import splink.duckdb.comparison_library as cl
-import splink.duckdb.comparison_template_library as ctl
-from splink.duckdb.blocking_rule_library import block_on
+import splink.comparison_level_library as cll
+import splink.comparison_library as cl
+from splink import block_on
```
zaneselvans, line 10: There was a big namespace reorganization in v4. Same modules and functions, but imported from different places.

src/pudl/analysis/record_linkage/eia_ferc1_model_config.py
```diff
@@ -19,7 +18,7 @@
 blocking_rule_7 = "l.report_year = r.report_year and l.capacity_mw = r.capacity_mw and substr(l.plant_name_mphone,1,2) = substr(r.plant_name_mphone,1,2)"
 blocking_rule_8 = "l.report_year = r.report_year and l.installation_year = r.installation_year and substr(l.plant_name_mphone,1,2) = substr(r.plant_name_mphone,1,2)"
 blocking_rule_9 = "l.report_year = r.report_year and l.construction_year = r.construction_year and substr(l.plant_name_mphone,1,2) = substr(r.plant_name_mphone,1,2)"
-blocking_rule_10 = block_on(["report_year", "net_generation_mwh"])
+blocking_rule_10 = block_on("report_year", "net_generation_mwh")
```
zaneselvans, line 21: Change in the function signature. Takes an arbitrary number of positional args now.

src/pudl/analysis/record_linkage/eia_ferc1_model_config.py
```diff
@@ -33,59 +32,57 @@
     blocking_rule_10,
 ]
 
-plant_name_comparison = ctl.name_comparison(
+plant_name_comparison = cl.NameComparison(
```
zaneselvans, line 35: A lot of things that were previously functions have been replaced with classes.

src/pudl/analysis/record_linkage/eia_ferc1_model_config.py
```diff
@@ -33,59 +32,57 @@
...
-    damerau_levenshtein_thresholds=[],
     jaro_winkler_thresholds=[0.9, 0.8, 0.7],
-    term_frequency_adjustments=True,
-)
-fuel_type_code_pudl_comparison = cl.exact_match(
-    "fuel_type_code_pudl", term_frequency_adjustments=True
 )
+utility_name_comparison.configure(term_frequency_adjustments=True)
+fuel_type_code_pudl_comparison = cl.ExactMatch("fuel_type_code_pudl")
+fuel_type_code_pudl_comparison.configure(term_frequency_adjustments=True)
```
zaneselvans, line 45: This seems like a funny change to me, but the `.configure()` method is used to set attributes which are common to all the different kinds of match classes now.

src/pudl/analysis/record_linkage/eia_ferc1_model_config.py
```diff
@@ -33,59 +32,57 @@
...
     """Get date comparison template for column."""
-    return ctl.date_comparison(
+    return cl.DateOfBirthComparison(
         column_name,
-        damerau_levenshtein_thresholds=[],
-        datediff_thresholds=[1, 2],
-        datediff_metrics=["year", "year"],
+        input_is_string=False,
+        datetime_thresholds=[1, 2],
+        datetime_metrics=["year", "year"],
```
zaneselvans, line 85: Weirdly they went to calling this `DateOfBirthComparison` as if that's the only kind of date you might want to compare.

src/pudl/analysis/record_linkage/eia_ferc1_record_linkage.py
```diff
@@ -211,29 +212,30 @@ def get_training_data_df(inputs):
...
+        retain_matching_columns=True,
+        retain_intermediate_calculation_columns=True,
+        probability_two_random_records_match=(1.0 / len(eia_df)),
+    )
+    linker = Linker(
         [eia_df, ferc_df],
+        settings=settings,
         input_table_aliases=["eia_df", "ferc_df"],
-        settings_dict=settings_dict,
+        db_api=DuckDBAPI(),
```
zaneselvans, line 229: The `Linker` class is agnostic as to what database backend it uses -- you pass in a class that is used to interface with it.

src/pudl/analysis/record_linkage/eia_ferc1_model_config.py
```diff
@@ -33,59 +32,57 @@
...
 
 def get_date_comparison(column_name):
     """Get date comparison template for column."""
-    return ctl.date_comparison(
+    return cl.DateOfBirthComparison(
         column_name,
-        damerau_levenshtein_thresholds=[],
-        datediff_thresholds=[1, 2],
-        datediff_metrics=["year", "year"],
+        input_is_string=False,
```
zaneselvans, line 83: This `input_is_string=False` parameter is the only thing I was worried about being more than a rename -- it means that the date field that's being matched is a `Date` or `DateTime` type, which I wasn't sure about. But the ETL worked, so I'm assuming it was correct.

pudlbot (approved):

PR #3841, opened 2024-09-11
Add pudl usage metrics gcp infrastructure

bendnorman:
# Overview

- Mount the Cloud SQL `pudl-usage-metrics-db` to Superset cloud run instance
- Create a new bucket for raw usage metrics archives
- Give github action service account permission to write to the bucket

# Testing

The usage metrics data can be viewed in superset and the github metrics are being saved to the new bucket.

```[tasklist]
# To-do list
```

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -426,8 +426,9 @@ resource "google_storage_bucket_iam_member" "usage_metrics_archiver_gcs_iam" {
 }
 
 resource "google_storage_bucket_iam_member" "usage_metrics_etl_gcs_iam" {
+  for_each = toset(["roles/storage.legacyBucketReader", "roles/storage.objectViewer"])
 
   bucket = google_storage_bucket.pudl_usage_metrics_archive_bucket.name
-  role = "roles/storage.legacyBucketReader"
+  role = each.key
   member = "serviceAccount:pudl-usage-metrics-etl@catalyst-cooperative-pudl.iam.gserviceaccount.com"
 }
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -432,3 +432,11 @@ resource "google_storage_bucket_iam_member" "usage_metrics_etl_gcs_iam" {
   role = each.key
   member = "serviceAccount:pudl-usage-metrics-etl@catalyst-cooperative-pudl.iam.gserviceaccount.com"
 }
+
+resource "google_storage_bucket_iam_member" "usage_metrics_etl_s3_logs_gcs_iam" {
+  for_each = toset(["roles/storage.legacyBucketReader", "roles/storage.objectViewer"])
+
+  bucket = "pudl-s3-logs.catalyst.coop"
+  role = each.key
+  member = "serviceAccount:pudl-usage-metrics-etl@catalyst-cooperative-pudl.iam.gserviceaccount.com"
+}
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -424,3 +424,10 @@ resource "google_storage_bucket_iam_member" "usage_metrics_archiver_gcs_iam" {
   role = each.key
   member = "serviceAccount:${google_service_account.usage_metrics_archiver.email}"
 }
+
+resource "google_storage_bucket_iam_member" "usage_metrics_etl_gcs_iam" {
+
+  bucket = google_storage_bucket.pudl_usage_metrics_archive_bucket.name
+  role = "roles/storage.legacyBucketReader"
+  member = "serviceAccount:pudl-usage-metrics-etl@catalyst-cooperative-pudl.iam.gserviceaccount.com"
+}
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -228,7 +228,7 @@ resource "google_cloud_run_v2_service" "pudl-superset" {
     volumes {
       name = "cloudsql"
       cloud_sql_instance {
-        instances = ["catalyst-cooperative-pudl:us-central1:superset-database"]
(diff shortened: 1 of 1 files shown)


terraform/main.tf
```diff
@@ -396,3 +396,31 @@ resource "google_service_account_iam_member" "gce-default-account-iam" {
   role               = "roles/iam.serviceAccountUser"
   member             = "serviceAccount:345950277072@cloudbuild.gserviceaccount.com"
 }
+
+resource "google_secret_manager_secret" "pudl_usage_metrics_db_connection_string" {
+  secret_id = "pudl-usage-metrics-db-connection-string"
+  replication {
+    auto {}
+  }
+}
```
bendnorman, line 405: I figured I'd save the connection string in case we need to reconnect the db to superset.

terraform/main.tf
```diff
@@ -396,3 +396,31 @@ resource "google_service_account_iam_member" "gce-default-account-iam" {
...
+  location      = "US"
+  storage_class = "STANDARD"
+
+  uniform_bucket_level_access = true
+}
+
+resource "google_service_account" "usage_metrics_archiver" {
+  account_id   = "usage-metrics-archiver"
+  display_name = "PUDL usage metrics archiver github action service account"
+}
```
bendnorman, line 418: I did create a service account key for the GitHub action in the business repo. @jdangerx would love a WIF tutorial soon!

terraform/main.tf
```diff
@@ -396,3 +396,39 @@ resource "google_service_account_iam_member" "gce-default-account-iam" {
...
+resource "google_storage_bucket_iam_member" "usage_metrics_archiver_gcs_iam" {
+  for_each = toset(["roles/storage.objectCreator", "roles/storage.objectViewer"])
+
+  bucket = google_storage_bucket.pudl_usage_metrics_archive_bucket.name
+  role = each.key
+  member = "serviceAccount:${google_service_account.usage_metrics_archiver.email}"
+}
+
+resource "google_storage_bucket_iam_member" "usage_metrics_etl_gcs_iam" {
+  for_each = toset(["roles/storage.legacyBucketReader", "roles/storage.objectViewer"])
```
bendnorman, line 429: I couldn't find a non legacy role that gives a principle the `storage.buckets.get` and `storage.objects.get` permissions. Seems like the [GCS python client](https://github.com/catalyst-cooperative/pudl-usage-metrics/actions/runs/10817188712/job/30010032641) wants both to access objects in a bucket.

terraform/main.tf
jdangerx: I've forgotten everything I know about WIF but could re-learn it!

terraform/main.tf
jdangerx: This is probably fine. If we want to switch to non-legacy roles, it looks like we could give `roles/storage.objectUser` for `objects.get` and the confusingly named `roles/storage.insightsCollectorService` gets you `buckets.get`.

terraform/main.tf
```diff
@@ -396,3 +396,47 @@ resource "google_service_account_iam_member" "gce-default-account-iam" {
...
+
+  bucket = google_storage_bucket.pudl_usage_metrics_archive_bucket.name
+  role = each.key
+  member = "serviceAccount:pudl-usage-metrics-etl@catalyst-cooperative-pudl.iam.gserviceaccount.com"
+}
+
+resource "google_storage_bucket_iam_member" "usage_metrics_etl_s3_logs_gcs_iam" {
+  for_each = toset(["roles/storage.legacyBucketReader", "roles/storage.objectViewer"])
+
+  bucket = "pudl-s3-logs.catalyst.coop"
```
jdangerx, line 439: Should we manage this bucket via TF too?

jdangerx (changes requested):
This all looks fine, just wondering if we should manage the s3-logs bucket via terraform as well. Other than this you're good to go!

terraform/main.tf
bendnorman: Probably! How can we manage a resource in terraform that has already been creating in the UI? Also, we should probably move the contents of `pudl-s3-logs.catalyst.coop` to `pudl-usage-metrics-archives.catalyst.coop/s3/` for consistency.

jdangerx: Here's the [documentation](https://developer.hashicorp.com/terraform/tutorials/state/resource-drift#introduce-drift) about handling "resource drift"!

bendnorman: I think we probably don't need the `pudl` part because the bucket lives in the `pudl-catalyst-cooperative` GCP project. It would be nice to keep `archive` because I could imagine having a bucket to store the outputs of our usage metric ETL to parquet files if we ever move to Big Query. How about `usage-metrics-archive.catalyst.coop`?

e-belfer: @bendnorman Would love to coordinate a name change with the flight of PRs in `pudl-usage-metrics` as much as possible, so just let me know what you're thinking.

e-belfer:
@bendnorman A question from @jdangerx on this `usage-metrics` PR I just merged: https://github.com/catalyst-cooperative/pudl-usage-metrics/pull/167#discussion_r1763293382

(Tldr - do we need to terraform the account credentials set up on the daily raw logs archiver?)

jdangerx:
I think, if the name of the bucket is still under discussion & we're considering moving buckets, then we should just do that as a separate PR and get this merged in.

terraform/main.tf
bendnorman: I think we'll keep it as is for now but I created an issue to rename [it down the line](https://github.com/catalyst-cooperative/pudl-usage-metrics/issues/170).

PR #3842, opened 2024-09-12
FERC 714: transform of hourly demand table (dbf +xbrl)

### LINKED ISSUE #3838: Write transform function to clean and normalize FERC 714 XBRL hourly historic load table
```[tasklist]
### Tasks
- [x] decide if/how to use pieces of the form 1 transforms (https://github.com/catalyst-cooperative/pudl/issues/3809#issuecomment-2344445336)
- [x] pull relevant methods out of transform.ferc1/classes  into more accessible functions
- [x] pull both csv and xbrl files into one asset
- [x] pull csv-based transform code out of existing transform function into their own modular functions
- [x] prep duration_xbrl: (convert the yearly reported records)
- [x] prep instant_xbrl: nada
- [x] prep xbrl: SO MUCH TIME SERIES CLEANING. objective: complete, continuous, non-duplicative datetime64[ns]
- [x] document small number of gaps in xbrl (mostly from daylight savings time hour)
- [x] prep csv: migrate csv-only relevant code into csv step
- [x] map_respondent_id_ferc714: csv/xbrl -> PUDL id (first phase done in #3846)
- [x] concatenate CSV + XBRL
- [x] deduplicate time series
- [x] map new bad utc offset codes
- [x] plot all respondents and do major spot fixing/timezone fixing
- [ ] [maybe] rename this table from `out_` to `core_`?
- [x] add the respondent_id_ferc714_csv and respondent_id_ferc714_xbrl into the table schema
- [x] add the #3846 little glue respondent id tables into db
- [x] back/forward fill the source respondent ids
- [x] add any additional info about the table into the metadata
- [x] sort the table before loading to make it most efficient for parquet
- [x] release notes
- [x] run all of the downstream analysis outputs
```

### out of scope here -> issue or pr or otherwise
- make the extract loading wayyyy faster or skip it for this hourly table -> dealing w/ this in #3861

cmgosnell:
# Overview

Closes #3838. There is a tasklist in the issue!

most of the work here has been in cleaning the date formats 🙄 

- datetimes omigosh....
   - 🟢 `convert_dates_to_zero_offset_hours_xbrl` hours being 01-24 or 01-00 (of next day.. probably  I put in some emails to ferc about this)
   - 🟢 `convert_dates_to_zero_seconds`: some last record of the days being last second of the day (T23:59)
   - GAPS. There are 41 gaps in the timeseries 
   - there are lots of overlapping timestamps rn
   - 
- ...

## TO-DO once all tables are merged in
- [x] get rid of `_post_process` function

# Testing

```[tasklist]
# To-do list
- [x] If updating analyses or data processing functions: make sure to update or write data validation tests (e.g. `test_minmax_rows()`)
- [x] For bigger ETL or data changes run the full ETL locally and then [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `make pytest-validate`.
```

### DIFF
--- a/src/pudl/analysis/state_demand.py
+++ b/src/pudl/analysis/state_demand.py
@@ -263,10 +263,8 @@ def load_ventyx_hourly_state_demand(path: str) -> pd.DataFrame:
         "_out_ferc714__hourly_pivoted_demand_matrix": AssetOut(),
         "_out_ferc714__utc_offset": AssetOut(),
     },
-    required_resource_keys={"dataset_settings"},
 )
 def load_hourly_demand_matrix_ferc714(
-    context,
     out_ferc714__hourly_planning_area_demand: pd.DataFrame,
 ) -> tuple[pd.DataFrame, pd.DataFrame]:
     """Read and format FERC 714 hourly demand into matrix form.


src/pudl/transform/ferc714.py
```diff
@@ -319,6 +340,270 @@ def _pre_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
     return out_df
 
 
+def melt_hourx_columns_csv(df):
```
cmgosnell, line 343: a lot of these smaller functions (especially the `_csv` suffixed ones) are just code that I pulled out of the hourly function. I pulled them out for several reasons:
- i wanted to be able to intersperse transform.classes functions with existing ferc714 specific functions. i think it looks much nicer and more readable with the main table function mostly being a chain of various transform functions.
- i needed to add many steps into the transform of this table. if i kept everything in there it would have exploded.
- these smaller steps are all now much much more unit test-able bc they have clear inputs and contained scopes
- now you can tell which pieces should pertain to which stage of the processing with the function suffix (`_csv`, `_xbrl`, `_instant_xbrl`, `_duration_xbrl`, or nothing (the concated table)).

The main downside i see here is that you can't easily see which of these functions are general or specific to a particular table. I think we should transform the two other tables and see if there is some way to clump them reasonably even just with module headers.

src/pudl/transform/ferc714.py
```diff
@@ -319,6 +340,270 @@ def _pre_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+        value_vars=range(24),
+        var_name="hour",
+        value_name="demand_mwh",
+    )
+    return df
+
+
+def map_respondent_id_ferc714(df, source: Literal["csv", "xbrl"]):
+    """TODO: Make this map!"""
+    return df
```
cmgosnell, line 370: right now the table gets fully through the transforms except for the enforce_schema step because the respondent_id_ferc714 now contains both the old int id and the new string id. we need to build the map and replace them here.

src/pudl/transform/ferc714.py
```diff
@@ -319,6 +340,270 @@ def _pre_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+
+
+def assign_report_day(df: pd.DataFrame, date_col: str) -> pd.DataFrame:
+    """Add a report_day column."""
+    return df.assign(
+        report_day=pd.to_datetime(df[date_col], format="%Y-%m-%d", exact=False)
+    )
+
+
+def merge_instant_and_duration_tables_xbrl(
```
cmgosnell, line 416: i almost used the ferc1 version of this but backed out and decided not to. that method calls other methods within the broader ferc table transformer class and it was hard to disentangle. also it was way more complex than this one needed to be. and also also this one needed to be done using different PKs and we needed to assign the day to merge on that. so i used the name and the general shape but not the code itself

aesharpe:
Just a lil lonely comment because it's not part of the code you edited therefore I can't comment directly on the line...

The doc strings for the `OFFSET_CODES` dictionary seems like it has a copy and paste error in it?

src/pudl/transform/ferc714.py
```diff
@@ -409,139 +402,385 @@ def core_ferc714__respondent_id(
...
+                "We expect there to be nearly no gaps in the time series."
+                f"but we found these gaps:\n{gappy_dates}"
+            )
+        return df.drop(columns=["gap"])
+
+    @staticmethod
+    def parse_date_strings_xbrl(xbrl: pd.DataFrame) -> pd.DataFrame:
+        """Convert report_date into pandas Datetime types."""
+        xbrl = xbrl.astype({"report_date": "datetime64[ns]"})
+        return xbrl
```
aesharpe, line 642: Why did you give this it's own function? Seems pretty small?

aesharpe (commented):
This looks really good! I found it really easy to go through all the functions and understand what was happening to the table. Most of my comments are non-blocking. Love that all of the functions are bite-sized! :)

src/pudl/transform/ferc714.py
```diff
@@ -319,6 +340,270 @@ def _pre_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
     return out_df
 
 
+def melt_hourx_columns_csv(df):
+    """Melt hourX columns into hours."""
```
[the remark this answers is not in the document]
aesharpe: nevermind!

src/pudl/transform/ferc714.py
cmgosnell: yea i agree. previously this function was more like the `parse_date_strings_csv` in that it both converted dates and did the `ensure_dates_are_complete_and_unique_xbrl`. but i needed to do yet another time cleaning step before the ensure step so i broke them apart. I will just do this inline in the `run` method

src/pudl/transform/ferc714.py
```diff
@@ -234,73 +272,54 @@
 
 RENAME_COLS = {
     "core_ferc714__respondent_id": {
```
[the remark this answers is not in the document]
cmgosnell: I used the rename function from the `transform.classes` for this and the `RenameColumns` param. Its not a bad idea to imagine doing these renamings in csv's like the EIA data but instead of years it could be source.

But that'd be a bigger structural lift to use the extractor tooling over here! Right now in the `transform.classes.rename_columns` we are checking if there are columns we are trying to rename that doesn't show up in the data but ya the cross source validation isn't here.

src/pudl/transform/ferc714.py
```diff
@@ -234,73 +272,54 @@
...
+    csv: RenameColumns = RenameColumns()
+    xbrl: RenameColumns = RenameColumns()
+
+
 ##############################################################################
 # Internal helper functions.
 ##############################################################################
-def _pre_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
+def pre_process_csv(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
     """A simple transform function for until the real ones are written.
```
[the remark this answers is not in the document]
cmgosnell: lol yea

src/pudl/transform/ferc714.py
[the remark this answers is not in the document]
aesharpe: This could be a good thing to circle back to if we have time later.

aesharpe:
oh also, @cmgosnell reminder to add some color to the `metadata/resources/ferc714.py` module about the `out_ferc714__hourly_planning_area_demand` table.

src/pudl/metadata/resources/ferc714.py
```diff
@@ -21,11 +22,27 @@
...
+            "24 instances of hourly demand for each day. The form is labeled with hour "
+            "1-24. There is no indication if hour 1 begins at midnight.\nThe XBRL data "
+            "contained several formats of timestamps. Most records corresponding to hour "
+            "1 of the Form have a timestamp with hour 1 as T1. About two thirds of the records "
+            "in the hour 24 location of the form have a timestamp with an hour reported as "
+            "T24 while the remaining third report this as T00 of the next day. T24 is not a"
+            "valid format for the hour of a datetime, so we convert these T24 hours into "
+            "T00 of the next day. A smaller subset of the respondents reports the 24th hour "
+            "as the last second of the day - we also convert these records to the T00 of the "
+            "next day."
```
aesharpe, line 39: This looks good! Maybe explain that we add a `respondent_id_ferc714` to combine the xbrl and csv sources because they differ.

src/pudl/transform/ferc714.py
```diff
@@ -805,7 +805,7 @@ def run(
         process the combined datasets.
 
         The main transforms include spot-fixing forecast years with
-        :meth:`spot_fix_forecast_years_xbrl`and averaging out duplicate forecast values
+        :func:`spot_fix_forecast_years_xbrl` and averaging out duplicate forecast values
```
cmgosnell: this guy is a method not a function! so imma swap this back

src/pudl/metadata/resources/ferc714.py
cmgosnell: done! plus i added better color into the field descriptions

src/pudl/package_data/settings/etl_fast.yml
```diff
@@ -21,7 +21,7 @@ ferc_to_sqlite_settings:
   ferc60_xbrl_to_sqlite_settings:
     years: [2021, 2022]
   ferc714_xbrl_to_sqlite_settings:
-    years: [2021, 2022]
+    years: [2021, 2023]
```
cmgosnell, line 24: the years here and in the pudl-etl settings were disjoint (here had 2022 and the pudl etl settings had 2023. this was causing several errors and was hard to track down bc it only happened when i re-ran the xbrl_to_sqlite stuff. because of this i added a validation (and corresponding unit test) in to the EtlSettings

src/pudl/transform/ferc714.py
```diff
@@ -1147,7 +1148,10 @@ def average_duplicate_pks_csv(df):
             & (df["forecast_year"] == 2014)
             & (df["net_demand_forecast_mwh"] == 0)
         )
-        assert len(df[error_mask] == 1)
+        if (len_dupes := len(df[error_mask])) >= 1:
+            raise AssertionError(
+                f"We found {len_dupes} duplicate errors, but expected 1 or less:\n{df[error_mask]}"
```
zaneselvans, line 1162: Does this message line up with the if statement?  The if says that finding 1 is not acceptable, but the message seems to say that 1 is acceptable.

src/pudl/transform/ferc714.py
```diff
@@ -957,12 +957,12 @@ def drop_missing_utc_offset(df):
         # Assert that all records missing UTC offset have zero demand
         missing_offset = df["utc_offset"].isna()
         bad_offset_and_demand = df.loc[missing_offset & (df.demand_mwh != 0)]
-        if not bad_offset_and_demand.empty:
+        if len(bad_offset_and_demand) > 12:
```
zaneselvans: Do we understand why 12 is the right number now? The error message still indicates that there should be 0. An explanation of the magic number would be helpful.

src/pudl/analysis/state_demand.py
```diff
@@ -390,17 +406,23 @@ def impute_ferc714_hourly_demand_matrix(df: pd.DataFrame) -> pd.DataFrame:
...
+    # the newer years of data first. This is so we can see early if
+    # new data causes any failures.
+    df = df.sort_index(ascending=False)
+    for year, gdf in df.groupby(df.index.year, sort=False):
+        if year in years:
+            logger.info(f"Imputing year {year}")
+            keep = df.columns[~gdf.isnull().all()]
+            tsi = pudl.analysis.timeseries_cleaning.Timeseries(gdf[keep])
+            result = tsi.to_dataframe(tsi.impute(method="tnn"), copy=False)
+            results.append(result)
```
cmgosnell, line 425: i did two things here:
- make the newer years run first - bc this takes 15 minutes to run omigosh and presumably the new failures will come in the new years.
- restrict the years that this impute is run on - bc the xbrl data introduced "Hour 24" records that are the first second of Jan 1 of the following year & this impute needs... well at least more than one record for the whole year

src/pudl/transform/ferc714.py
cmgosnell: it says one or less

cmgosnell: one is acceptable

zaneselvans: But I think the assertion will be raised if it's 1, because the if says `>= 1`. I think it should probably be `> 1`?

cmgosnell: yes you are right ty!

src/pudl/transform/ferc714.py
cmgosnell: this is a good flag. i originally found these 12 baddies just on the fast etl run just through the pytest-coverage test but I now know those runs were a little junked up for settings reasons. i will try to revert this change and see what happens or document the weirdness if it is really there

cmgosnell: these are real 12 bad guys but they should be dropped and they only show up in the fast test. so imma keep this here

src/pudl/transform/ferc714.py
cmgosnell: fixed

src/pudl/analysis/state_demand.py
aesharpe: hmm ok! I thought we fixed the weird hour data for XBRL. I remember the issue was with Hour 0 vs. Hour 24 etc. but maybe we just made then all Hour 24. Should we correct it so it's not the first second of the new year?

src/pudl/analysis/state_demand.py
```diff
@@ -390,17 +392,28 @@ def impute_ferc714_hourly_demand_matrix(df: pd.DataFrame) -> pd.DataFrame:
...
+    # sort here and then don't sort in the groupby so we can process
+    # the newer years of data first. This is so we can see early if
+    # new data causes any failures.
+    df = df.sort_index(ascending=False)
+    for year, gdf in df.groupby(df.index.year, sort=False):
+        # remove the records o/s of the working years because some
+        # respondents report one record of midnight of January first
+        # of the next year (report_date.dt.year + 1). and
+        # impute_ferc714_hourly_demand_matrix chunks over years at a time
+        # and having only one record
```
aesharpe, line 410: This looks unfinished

src/pudl/analysis/state_demand.py
```diff
@@ -390,17 +392,28 @@ def impute_ferc714_hourly_demand_matrix(df: pd.DataFrame) -> pd.DataFrame:
...
+        if year in years:
+            logger.info(f"Imputing year {year}")
+            keep = df.columns[~gdf.isnull().all()]
+            tsi = pudl.analysis.timeseries_cleaning.Timeseries(gdf[keep])
+            result = tsi.to_dataframe(tsi.impute(method="tnn"), copy=False)
+            results.append(result)
```
aesharpe, line 416: Maybe we should add a check here to make sure there aren't too many `year` values outside of `years`

aesharpe (approved):
We don't have any tests for the yearly_forecast table, but I think it's ok!

PR #3843, opened 2024-09-13
Adjust superset row limits

bendnorman:
# Overview

This PR:
- Adds superset config variables for changing the row of charts and queries
- Enables the Public superset role which allows us to share specific superset resources with folks who don't have an account
- Adds the Cloud Auth Proxy to the docker compose setup

# Testing

- You can try to connect to the production database!
- I was able to view a dashboard without registering

```[tasklist]
# To-do list
```


superset/docker-compose.yml
```diff
@@ -29,3 +29,14 @@ services:
       POSTGRES_PASSWORD: superset
     ports:
       - 8081:5432
+  cloud_auth_proxy:
+    image: gcr.io/cloud-sql-connectors/cloud-sql-proxy:2.13.0
+    profiles: [prod]
```
bendnorman, line 34: I thought creating a separate profile for prod would be another way to avoid accidentally making changes to the production database. Think it makes sense to have to have a separate `local` profile?

jdangerx: Having two profiles makes sense, I think:
- `local`: spin up the web UI + postgres; require all the auth0 stuff to be set, superset secret key, etc. 
- `prod`: just spin up cloud auth proxy and a bash shell that has all the superset dependencies - this lets you run admin commands like
  ```bash
  $ superset db upgrade
  $ superset init
  ```
  But doesn't let you do weird stuff to the prod database, like accidentally add a user through auth0.

jdangerx (changes requested):
I think your suggestion is good & smart! And that it should be done before merging :)

superset/README.md
```diff
@@ -37,7 +37,50 @@ docker compose up
...
+```
+
+and then set the environment variables:
+
+```
+export SUPERSET_DB_HOST=cloud_auth_proxy
+export SUPERSET_DB_PORT=5432
+export SUPERSET_DB_USER=postgres
+export SUPERSET_DB_PASS={Grab the password from GCP Secret Manager}
+export SUPERSET_DB_NAME=postgres
```
jdangerx, line 75: Oh and we also need to export the `CLOUD_SQL_CONNECTION_NAME` env var.

superset/docker-compose.yml
jdangerx: Sweet!

I think the superset service can just know which env it's in via the environment variables, for now.

bendnorman: Is there a better way you'd like to move to in the future?

jdangerx: I think we could use [env files](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute) so we just do

```
docker compose up --profile prod --env-file prod.env # prod.env is .gitignored
docker compose up --profile local --env-file local.env # local.env is tracked in git
```

I *don't* think there's a way to set `env-file` based on `profile`, unfortunately - that seems like it'd be the nicest.

jdangerx: That *seems* like it'd be a small change, actually, so if you think that sounds promising you should go for it! But I'm OK with keeping things as is for now.

bendnorman: I couldn't figure out a clean way to do this but I did add some interpolation logic that requires some env vars!

PR #3844, opened 2024-09-13
Remove readme dynamic information

Nancy9ice:
Overview:
Trim down the readme by removing duplicate dynamic information. These include the following headings: ‘What data is available?’, ‘Other data of interest’, ‘Organizations Using PUDL’. These headings are removed because they are already found on the PUDL website.

The headings weren't exactly removed as stated in the task. Instead, they were edited to reflect information that can be found in the more flexible sources (readthedocs or the website). As @aesharpe suggested, I changed the data source URLs to link to only the data source information on the readme. However, some of these sources weren't on the readthedocs so I left the US government website source instead. I would appreciate feedback on how best you would like me to go about this.

Testing
Since changes were made to the readme, I only checked if changes reflected properly on the readme before making a pull request.

@aesharpe @zaneselvans @e-belfer


aesharpe (commented):
I wonder if we can do a better job of deduplicating the data here, on the [website](https://catalyst.coop/data/), and in the [readthedocs](https://catalystcoop-pudl.readthedocs.io/en/nightly/#available-data) pages. @Nancy9ice where do you think this information should live? Ideally it's only in one place and we link to it. My gut says the readthedocs.

Same goes for other information in the ReadMe.

Nancy9ice:
> I wonder if we can do a better job of deduplicating the data here, on the [website](https://catalyst.coop/data/), and in the [readthedocs](https://catalystcoop-pudl.readthedocs.io/en/nightly/#available-data) pages. @Nancy9ice where do you think this information should live? Ideally it's only in one place and we link to it. My gut says the readthedocs.
> 
> Same goes for other information in the ReadMe.

I think it's best to have the organizations that have worked with PUDL on the website because the website is more like a branding material for PUDL. Then we can link the organizations heading in readthedocs to the one on the website. For the PUDL source links, it's great to that we linked them from the readme to the PUDL links. Do you have better ideas? Would love to hear it

README.rst
```diff
@@ -180,18 +170,12 @@ Organizations using PUDL
 
 .. _pudl_orgs_start:
 
-This is a partial list of organizations that have used PUDL in their work.
```
e-belfer, line 183: Even though it is duplicative, I think we want to leave this here - most people won't follow the link, and it's good to highlight our partners in multiple places.

e-belfer (changes requested):
I think we want to leave the organizations in our README, even though it's duplicative - otherwise, this looks good.

PR #3848, opened 2024-09-17
added introduction

Nancy9ice:
Overview:
Add the readthedocs landing page as the Introduction in the Table of Contents

This adds the landing page to the table of contents to ease navigation

Testing
I tested it by running make `make docs-build`

@aesharpe @zaneselvans @e-belfer


docs/index.rst
```diff
@@ -1,231 +1,8 @@
...
-   :start-after: .. _pudl_orgs_start:
-   :end-before: .. _pudl_orgs_end:
-
-
-
 .. toctree::
   :hidden:
   :maxdepth: 2
 
+  Introduction <Introduction>
```
e-belfer, line 5:
```suggestion
  About PUDL <self>
```

This will create a reference to the page in the table of contents without needing to create a new page or having an empty index.

e-belfer (changes requested):
Can we call this "About PUDL" instead of "Introduction"?

Rather than creating a blank index page and a new Introduction page, I think we need to add the index page to the TOC.

See: https://stackoverflow.com/questions/16123951/how-do-i-include-the-homepage-in-the-sphinx-toc

e-belfer (approved):
Apologies for the delay in re-reviewing, this looks great!

PR #3853, opened 2024-09-18
Increase nightly build VM disk from 80GB to 100GB

### LINKED ISSUE #3852: Fix PUDL Kaggle dataset updates
The PUDL Kaggle dataset last updated on August 24th, but it's supposed to pull new data every Monday from the S3 bucket, so something is broken.

I attempted a manual update and it failed with an error, but I can't actually *see* the error.  I made [a forum post](https://www.kaggle.com/discussions/product-feedback/534701) reporting the issue.

I'm worried that we might have run up against some kind of quota, and searching around the support forums it sounds like it may be impossible to delete old versions of a dataset, meaning the only way to continue with updates would be to create an entirely new dataset, which would be lousy for continuity.

zaneselvans:
# Overview

Increase the disk allocated to the nightly build machine from 80GB to 100GB since we're getting an error about running out of disk when we start zipping the output SQLite DBs.

Closes #3852 (hopefully)

# Testing

Can't test this short of running the nightly builds.

```[tasklist]
# To-do list
- [ ] Cause `clean_up_outputs_for_distribution()` to fail if the zip command fails?
```

### DIFF
--- a/devtools/generate_batch_config.py
+++ b/devtools/generate_batch_config.py
@@ -58,7 +58,7 @@ def to_config(
                     "computeResource": {
                         "cpuMilli": 8000,
                         "memoryMib": int(63 * MIB_PER_GB),
-                        "bootDiskMib": 80 * 1024,
+                        "bootDiskMib": 100 * 1024,
                     },
                     "maxRunDuration": f"{60 * 60 * 12}s",
                 }


bendnorman (approved):
Seems worth a try!

Issue #3809, opened 2024-08-23
Extend FERC 714 forecasted and hourly load tables to 2023 by integrating XBRL tables and combining with existing CSV tables


aesharpe:
This epic will track the progress of integrating the FERC 714 annual load forecast, hourly historical load, and respondent ID tables through 2023. Currently we have both tables available through 2020 (the last year FERC data was published in the DBF format).

These issues will track the integration of the XBRL data, knitting together the IDs from the old DBF and new XBRL files, and subsequently mapping any new FERC IDs to EIA IDs.

## Additional context
* Note that we haven't done any XBRL data ETL outside of the FERC Form 1, so there will be substantial chunks of new infrastructure to design and build (or adapt / use from FERC 1) in the `pudl.extract.ferc714` and `pudl.transform.ferc714` modules. E.g. we'll need to be able to combine instant and duration tables and rename the columns so that they match the PUDL database column names like we did in FERC 1. Do we want to re-use that code?
* Most of the complex FERC 1 transforms will probably not be required here. Both of the data tables that we're integrating here are relatively simple on their own. There are only a few columns in each of them. Localizing the hourly data to the right timezone is a hassle, but we've got structures for that which can be reused.
* Most of the complexity in what we need to do here is record linkage between the old and new FERC IDs, and then record linkage between the unified FERC IDs we create (which span both old and new data) and the EIA entities that they correspond to.
* It's not always clear what the distinction is between a Planning Area and a Balancing Authority.
* There are a number of output tables downstream of the FERC-714 historical demand and respondent ID tables, which refer to BA IDs and Utility IDs, and compile service territories for both BAs and Utility IDs.
* There's also the state-level hourly demand estimates, which depend on the historical hourly demand and those service territories.
* Maybe all of those downstream tables will Just Work, but it might also require fiddling, in which case we should probably put it off.

```[tasklist]
# Extract XBRL Tables
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3813
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3814
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3815
```

```[tasklist]
# Clean & Normalize XBRL Tables
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3837
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3838
- [ ] https://github.com/catalyst-cooperative/pudl/issues/3839
```

```[tasklist]
# Link IDs
[...]

PR #3856, opened 2024-09-19
Transform XBRL core_ferc714__yearly_planning_area_demand_forecast table

### LINKED ISSUE #3837: Write transform function to clean and normalize FERC 714 XBRL annual 10-year load forecast table

aesharpe:
# Overview

Closes #3837 .

Adds XBRL data (2021 +) for the annual demand forecast table.

## What did you change?

- Add a `YearlyPlanningAreaDemandForecast` class and class functions.
- Migrate CSV transforms into these class functions.
- Add `forecast` to the column names with forecasted values.

## Ideas
We could add some tests for the 714 data that made sure the mwh and mw values fell within a reasonable range. Might be something we put on the backburner for now.

# Testing

Materialize the `core_ferc714__yearly_planning_area_demand_forecast` table in dagster.

```[tasklist]
# To-do list
```


src/pudl/transform/ferc714.py
```diff
@@ -771,70 +781,152 @@ def out_ferc714__hourly_planning_area_demand(
...
-@asset(
-    io_manager_key="pudl_io_manager",
-    compute_kind="pandas",
-)
-def core_ferc714__yearly_planning_area_demand_forecast(
-    raw_ferc714_csv__yearly_planning_area_demand_forecast: pd.DataFrame,
-) -> pd.DataFrame:
-    """Transform the yearly planning area forecast data per Planning Area.
+class YearlyPlanningAreaDemandForecast:
+    """Class for building the :ref:`core_ferc714__yearly_planning_area_demand_forecast` asset.
```
aesharpe, line 785: The hourly table has an `out` prefix. Which one should we have here?

aesharpe (commented):
Just a high-level comment, but I spent some time tackling the duplicate PK values in the XBRL data before I realized that they were getting handled in the `FercXBRLSQLiteIOManager` class `load_input` function. Specifically, I didn't realize that the older `publication_date` values were getting dropped and the `report_year` column was getting made there. It's a little confusing to have that happen in the IOManager vs. the transforms. I understand that it might be more efficient to do it that way though. 

Out of scope to change for this PR, just a thought I had.

src/pudl/transform/ferc714.py
cmgosnell: the hourly table is weird bc its so big we will probably never make a de-normalized version of it. I personally think it makes sense to change it to `core_`. But there is a sounds argument for calling it `out_` bc its a table we want users to interact with. ([see comment](https://github.com/catalyst-cooperative/pudl/issues/3809#issuecomment-2342019024))

which is all to say i think this one should stay as `core_`

src/pudl/metadata/fields.py
```diff
@@ -2680,9 +2680,9 @@
         ),
         "unit": "MW",
     },
-    "net_demand_mwh": {
+    "net_demand_forecast_mwh": {
         "type": "number",
-        "description": "Net electricity demand for the specified period in megawatt-hours (MWh).",
+        "description": "Net forecasted electricity demand for the specific period in megawatt-hours (MWh).",
```
cmgosnell, line 2685: I love the addition of forecast in here. it makes it much more clear.

not blocking... but is there a difference between this and `demand_forecast_mwh`?

src/pudl/transform/ferc714.py
```diff
@@ -771,70 +781,161 @@ def out_ferc714__hourly_planning_area_demand(
...
+        This function fixes the following errors:
+
+        - There's one record with an NA forecast_year value. This row
+          also has no demand forcast values. Because forcast_year is a primary key
+          we can't have any NA values. Because there are no substantive forcasts
+          in this row, we can safely remove this row.
+        - respondent_id_ferc714 number 107 reported their forecast_year
+          as YY instead of YYYY values.
+        - There's also at least one forecast year value reported as 3022 that should
+          be 2033.
```
cmgosnell, line 846: lol

src/pudl/transform/ferc714.py
```diff
@@ -771,70 +781,161 @@ def out_ferc714__hourly_planning_area_demand(
...
+        mask = (
+            (df["respondent_id_ferc714"] == 17)
+            & (df["report_year"] == 2023)
+            & (df["forecast_year"] == 3033)
+        )
+        df.loc[mask, "forecast_year"] = 2033
+        # Make sure forecast_year values are expected
+        assert (
+            df["forecast_year"].isin(range(2021, 3001)).all()
+        ), "Forecast year values not in expected range"
```
cmgosnell, line 872: the forecasts go through 3000 eeep really?!

src/pudl/transform/ferc714.py
```diff
@@ -771,70 +781,161 @@ def out_ferc714__hourly_planning_area_demand(
...
+                    "net_demand_forecast_mwh",
+                ]
+            ]
+            .mean()
+            .reset_index()
+        )
+        # Make sure no more rows were dropped than expected
+        assert (
+            original_len - len(df) == 21
+        ), f"dropped {original_len - len(df)} rows, expected 26"
```
cmgosnell, line 914: nit but these expected numbers are off

src/pudl/transform/ferc714.py
```diff
@@ -771,70 +781,161 @@ def out_ferc714__hourly_planning_area_demand(
...
+            df.groupby(["respondent_id_ferc714", "report_year", "forecast_year"])[
+                [
+                    "summer_peak_demand_forecast_mw",
+                    "winter_peak_demand_forecast_mw",
+                    "net_demand_forecast_mwh",
+                ]
+            ]
+            .mean()
+            .reset_index()
+        )
```
cmgosnell, line 910: completely non-blocking nit/i know you didn't introduce this. I personally like doing a split/apply/combine process when altering only a few records. in theory this `mean` will do nothing to a group of 1. but i like to stay cautious about that.

```python
dupe_mask = {soemthing}
deduped = df[dupe_mask].groupby()...stuff
pd.concat([df[~dupe_mask], deduped]) 
```

cmgosnell (changes requested):
this looks grand!! i have two real but minor questions and one nittiest of nit.

src/pudl/metadata/fields.py
aesharpe: Yes, `demand_forecast_mwh` is from 930 and is a day-ahead forecast. This is a year-ahead forecast. The 930 value is also demand vs. the 714 value which is net demand.

aesharpe: We could probably add a day-ahead suffix or something to the 930 value! What do you think?

aesharpe: _Technically_ for the FERC714 table it's net_load vs net_demand, but they are sooo similar idk if it matters. But it's called net load in the FERC taxonomy.

src/pudl/transform/ferc714.py
aesharpe: No haha I was just trying to think of a reasonable range. We could certainly lower this! I was just not sure if it was something we wanted to update every year or use as a check for obvious outliers.

src/pudl/transform/ferc714.py
aesharpe: Makes sense! I sometimes fear that I'll introduce more problems with split/apply/combine than just applying something to an intact database. But I should trust my code more!

src/pudl/transform/ferc714.py
aesharpe: I think I can probably safely lower it to 2100

src/pudl/transform/ferc714.py
aesharpe: idk what that was about haha but fixed!

src/pudl/metadata/fields.py
cmgosnell: yea it makes sense to me to stay true to the source taxonomy unless we learn its wrong somehow, but that's not the case here

cmgosnell (approved):
🎉

cmgosnell:
oh! @aesharpe ! can you add the new years of data into the `Ferc714CheckSpec` asset check?

PR #3857, opened 2024-09-19
FERC 714: Integrate the XBRL data Respondent ID table 

### LINKED ISSUE #3839: Write transform function to clean and normalize FERC 714 XBRL respondent ID table
```[tasklist]
### Tasks
- [ ] ensure no sructural changes in new and old data
- [ ] do id's change year-to-year?
- [ ] are there new id's that are different than the old ids?
- [ ] fix any baddies
```

cmgosnell:
# Overview

Closes #3839 and #3858

What did you change?
There were two main threads that needed pulling to get this table updates:
* **Table compatibility**: The csv table is static while the xbrl table is reported annually. A lot of the downstream analysis expects this table to be static. So the first step was to check whether or not the columns that we have in the CSV years had consistent data over the few XBRL years that we have. There were a small number of `eia_code`'s we needed to clean up, but besides that it was static. I then converted the XBRL data into a static table, then I concat-ed the tables and checked the static-ness again.
* **eia_code cleaning**: Not a ton but some cleaning necessary. Done all in `spot_fix_eia_codes` & `EIA_CODE_FIXES`

still todo that I'd rather finish in `transform-714-xbrl` [3842](https://github.com/catalyst-cooperative/pudl/pull/3842):
* metadata/schema updates: add csv and xbrl respondent id's into table.

# Testing

```[tasklist]
# To-do list
- [x] Materialize the assets locally
```

### DIFF
--- a/src/pudl/transform/ferc714.py
+++ b/src/pudl/transform/ferc714.py
@@ -196,6 +196,9 @@ EIA_CODE_FIXES = {
     323: 58790,  # Gridforce Energy Management (missing)
     324: 58791,  # NaturEner Wind Watch LLC (Fixes bad ID 57995)
     329: 39347,  # East Texas Electricity Cooperative (missing)
+    # TODO: this is a placeholder! this is the new respondent_id_ferc714
+    # and everything above is a respondent_id_ferc714_csv
+    218: 56365,  # NaturEner Power Watch LLC (Fixes bad ID 57049, 57050)
 }
 """Overrides of FERC 714 respondent IDs with wrong or missing EIA Codes."""
 
@@ -204,7 +207,13 @@ RENAME_COLS = {
         "csv": {
             "respondent_id": "respondent_id_ferc714_csv",
             "respondent_name": "respondent_name_ferc714",
-        }
+            "eia_code": "eia_code",
+        },
+        "xbrl": {
+            "entity_id": "respondent_id_ferc714_xbrl",
+            "respondent_legal_name": "respondent_name_ferc714",
+            "respondent_identification_code": "eia_code",
+        },
     },
     "out_ferc714__hourly_planning_area_demand": {
         "csv": {
@@ -315,12 +324,96 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
     return PUDL_PACKAGE.get_resource(table_name).enforce_schema(df)
 
 
+class RespondentId:
+    """Class for building the :ref:`out_ferc714__hourly_planning_area_demand` asset.
+
+    Most of the methods in this class as staticmethods. The purpose of using a class
+    in this instance is mostly for organizing the table specific transforms under the
+    same name-space.
+    """
+
+    @classmethod
+    def run(
+        cls, raw_csv: pd.DataFrame, raw_xbrl_duration: pd.DataFrame
+    ) -> pd.DataFrame:
+        """Build the table for the :ref:`core_ferc714__respondent_id` asset.
+
+        TODO: Determine where the existing
+        """
+        table_name = "core_ferc714__respondent_id"
+        csv = _pre_process_csv(raw_csv, table_name)
+
+        xbrl = rename_columns(
+            raw_xbrl_duration,
+            params=RenameColumns(columns=RENAME_COLS[table_name]["xbrl"]),
+        ).pipe(cls.clean_eia_codes_xbrl)
+
+        df = pd.concat([csv, xbrl]).reset_index(drop=True)
+
+        return df
+
+    @staticmethod
+    def clean_eia_codes_xbrl(xbrl):
+        """Make eia_code's cleaner.
+
+        Desired outcomes here include all respondents have only one non-null
+        eia_code and all eia_codes that are actually the respondent_id_ferc714_xbrl
+        are nulled.
+
+        TODO: move the non-eia_code asserts
+        TODO: Convert bare asserts into Assertions
+        """
+        # we expect all of these submissions to be from the last Q
+        assert all(xbrl.report_period == "Q4")
+        # the CSV data does not vary by year, so we need to check if that is also going to be the case for the XBRL data
+        assert all(
+            xbrl.groupby(["respondent_id_ferc714_xbrl"])[  # noqa: PD101
+                ["respondent_name_ferc714"]
+            ].nunique()
+            == 1
+        )
+        # first we are gonna null out all of the "EIA" codes that are really just the respondent id
+        code_is_respondent_id_mask = xbrl.eia_code.str.startswith("C") & (
+            xbrl.respondent_id_ferc714_xbrl == xbrl.eia_code
+        )
+        xbrl.loc[code_is_respondent_id_mask, "eia_code"] = pd.NA
+
+        # lets null out some of the eia_code's from XBRL that we've manually culled
+        respondent_id_xbrl_to_bad_eia_code = {
+            "C002422": ["5776"],
+            "C011374": ["8376"],
+            "C002869": ["F720204"],
+            "C002732": ["F720204", "57049, 57050"],
+            "C011420": ["16606"],
+        }
+        for rid_xbrl, bad_eia_codes in respondent_id_xbrl_to_bad_eia_code.items():
+            xbrl.loc[
+                (xbrl.respondent_id_ferc714_xbrl == rid_xbrl)
+                & (xbrl.eia_code.isin(bad_eia_codes)),
+                "eia_code",
+            ] = pd.NA
+        assert all(
+            xbrl.dropna(subset=["eia_code"])  # noqa: PD101
+            .groupby(["respondent_id_ferc714_xbrl"])[["eia_code"]]
+            .nunique()
+            == 1
+        )
+        xbrl = xbrl.astype({"eia_code": pd.Int64Dtype()})
+        return xbrl
+
+
 @asset(
     io_manager_key="pudl_io_manager",
+    ins={
+        "raw_csv": AssetIn(key="raw_ferc714_csv__respondent_id"),
+        "raw_xbrl_duration": AssetIn(
+            key="raw_ferc714_xbrl__identification_and_certification_01_1_duration"
+        ),
+    },
     compute_kind="pandas",
 )
 def core_ferc714__respondent_id(
-    raw_ferc714_csv__respondent_id: pd.DataFrame,
+    raw_csv: pd.DataFrame, raw_xbrl_duration: pd.DataFrame
 ) -> pd.DataFrame:
     """Transform the FERC 714 respondent IDs, names, and EIA utility IDs.
 
@@ -329,15 +422,17 @@ def core_ferc714__respondent_id(
     corresponding EIA utility IDs provided by FERC for some reason (including
     PacifiCorp).
 
+    TODO: migrate transforms up into RespondentId.run() and just call that here.
+
     Args:
-        raw_ferc714_csv__respondent_id: Raw table describing the FERC 714 Respondents.
+        raw_csv: Raw table describing the FERC 714 Respondents from the CSV years.
+        raw_xbrl_duration: Raw table describing the FERC 714 Respondents from the
+            XBRL years.
 
     Returns:
         A clean(er) version of the FERC-714 respondents table.
     """
-    df = _pre_process_csv(
-        raw_ferc714_csv__respondent_id, table_name="core_ferc714__respondent_id"
-    )
+    df = _pre_process_csv(raw_csv, table_name="core_ferc714__respondent_id")
     df["respondent_name_ferc714"] = df.respondent_name_ferc714.str.strip()
     df.loc[df.eia_code == 0, "eia_code"] = pd.NA
     # There are a few utilities that seem mappable, but missing:
--- a/src/pudl/transform/ferc714.py
+++ b/src/pudl/transform/ferc714.py
@@ -174,7 +174,7 @@ TIMEZONE_CODES = {
 }
 """Mapping between standardized time offset codes and canonical timezones."""
 
-EIA_CODE_FIXES = {
+EIA_CODE_FIXES: dict[Literal["combined", "csv", "xbrl"], dict[int | str], int] = {
     "combined": {
         # FERC 714 Respondent ID: EIA BA or Utility ID
         125: 2775,  # EIA BA CAISO (fixing bad EIA Code of 229)
@@ -215,7 +215,13 @@ EIA_CODE_FIXES = {
         329: 39347,  # East Texas Electricity Cooperative (missing)
     },
 }
-"""Overrides of FERC 714 respondent IDs with wrong or missing EIA Codes."""
+"""Overrides of FERC 714 respondent IDs with wrong or missing EIA Codes.
+
+This is used in :meth:`RespondentId.spot_fix_eia_codes`. The dictionary
+is organized by "source" keys ("combined", "csv", or "xbrl"). Each source's
+value is a secondary dictionary which contains source respondent ID's as keys
+and fixes for EIA codes as values.
+"""
 
 RENAME_COLS = {
     "core_ferc714__respondent_id": {
@@ -320,7 +326,17 @@ def _assign_respondent_id_ferc714(
 def _fillna_respondent_id_ferc714_source(
     df: pd.DataFrame, source: Literal["csv", "xbrl"]
 ) -> pd.DataFrame:
-    """Fill missing CSV or XBRL respondent id."""
+    """Fill missing CSV or XBRL respondent id.
+
+    The source (CSV or XBRL) tables get assigned a PUDL-derived
+    ``respondent_id_ferc714`` ID column (via :func:`_assign_respondent_id_ferc714`).
+    After we concatenate the source tables, we sometimes to backfill and
+    forward-fill the source IDs (``respondent_id_ferc714_csv`` and
+    ``respondent_id_ferc714_xbrl``). This way the older records from the CSV years
+    will also have the XBRL ID's and vice versa. This will enable users to find
+    the full timeseries of a respondent that given either source ID (instead of
+    using the source ID to find the PUDL-derived ID and then finding the records).
+    """
     respondent_map_ferc714 = pd.read_csv(
         importlib.resources.files("pudl.package_data.glue")
         / "respondent_id_ferc714.csv"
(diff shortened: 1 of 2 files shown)


src/pudl/transform/ferc714.py
```diff
@@ -175,27 +175,45 @@
 """Mapping between standardized time offset codes and canonical timezones."""
 
 EIA_CODE_FIXES = {
```
cmgosnell, line 177: instead of converting all of these code fixes into the pudl-derived respondent id-based fixes, I kept the sourcey-ness. This is mostly to enable checking things at each stage - primarily `ensure_eia_code_uniqueness`

src/pudl/transform/ferc714.py
```diff
@@ -293,6 +317,26 @@ def _assign_respondent_id_ferc714(
     return df
 
 
+def _fillna_respondent_id_ferc714_source(
```
cmgosnell, line 320: this is actually also necessary in the hourly table as well!

src/pudl/transform/ferc714.py
```diff
@@ -293,6 +317,26 @@ def _assign_respondent_id_ferc714(
...
+    # use the source utility ID column to get a unique map and for merging
+    resp_id_col = f"respondent_id_ferc714_{source}"
+    resp_map_series = respondent_map_ferc714.dropna(subset=[resp_id_col]).set_index(
+        "respondent_id_ferc714"
+    )[resp_id_col]
+
+    df[resp_id_col] = df[resp_id_col].fillna(
+        df["respondent_id_ferc714"].map(resp_map_series)
+    )
+    return df
```
aesharpe, line 337: I'm not sure I understand how this is working. The `_assign_respondent_id_ferc714` function maps the `respondent_id_ferc714` column and then this column appears to work backwards from that to map on missing `respondent_id_ferc714_source` values. How is that possible? If there is no `respondent_id_ferc714_source` to begin with how can we map a `respondent_id_ferc714` value onto it?

src/pudl/transform/ferc714.py
aesharpe: Worth checking whether it's necessary for the hourly table

src/pudl/transform/ferc714.py
```diff
@@ -306,6 +350,8 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
     Applies standard data types and ensures that the tables generally conform to the
     schemas we have defined for them.
 
+    TODO: rip this out. enforce_schema happens via the io_managers now.
+
```
aesharpe, line 354: What's stopping us from removing it now?

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +361,223 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
     return PUDL_PACKAGE.get_resource(table_name).enforce_schema(df)
 
 
+class RespondentId:
+    """Class for building the :ref:`out_ferc714__hourly_planning_area_demand` asset.
```
aesharpe, line 365: Update to say `core_ferc714__respondent_id` instead of hourly table

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +361,223 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+            .pipe(cls.ensure_eia_code_uniqueness, "xbrl")
+            .pipe(cls.convert_into_static_table_xbrl)
+            .assign(source="xbrl")
+        )
+        # CONCATED STUFF
+        df = (
+            pd.concat([csv, xbrl])
+            .reset_index(drop=True)
+            .convert_dtypes()
+            .pipe(cls.spot_fix_eia_codes, "combined")
```
aesharpe, line 426: What's the rationale for having a "combined" spot fix if you are also going to have breakout xbrl and csv spot fixes? Would be nice to add a little sentence about this in the `EIA_CODE_FIXES` dict above.

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +361,223 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+            .pipe(cls.ensure_eia_code_uniqueness, "combined")
+            .pipe(cls.condense_into_one_source_table)
+            .pipe(_fillna_respondent_id_ferc714_source, "csv")
+            # the xbrl version of this is fillna is not *strictly necessary*
+            # bc we are sorting the records to grab the xbrl record
+            .pipe(_fillna_respondent_id_ferc714_source, "xbrl")
```
aesharpe, line 432: How come this is down in the combined section if it is specific to csv and xbrl?

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +361,223 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+        )
+        return df
+
+    @staticmethod
+    def spot_fix_eia_codes(
+        df: pd.DataFrame, source: Literal["csv", "xbrl", "combined"]
+    ) -> pd.DataFrame:
+        """Spot fix the eia_codes."""
+        df.loc[df.eia_code == 0, "eia_code"] = pd.NA
```
aesharpe, line 441: Maybe add a brief description of why this is done by source instead of as a conjoined table

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +361,223 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+            )
+        return df.drop(columns=["eia_code_count"])
+
+    @staticmethod
+    def clean_eia_codes_xbrl(xbrl: pd.DataFrame) -> pd.DataFrame:
+        """Make eia_code's cleaner coming from the XBRL data.
+
+        Desired outcomes here include all respondents have only one non-null
+        eia_code and all eia_codes that are actually the respondent_id_ferc714_xbrl
+        are nulled.
```
aesharpe, line 474: Is it possible that the `respondent_id_ferc714_xbrl` and `eia_code` are the same?

aesharpe: mmm looks like no because they start with C!

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +361,223 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+        """
+        # we expect all of these submissions to be from the last Q
+        assert all(xbrl.report_period == "Q4")
+        # first we are gonna null out all of the "EIA" codes that are really just the respondent id
+        code_is_respondent_id_mask = xbrl.eia_code.str.startswith("C") & (
+            xbrl.respondent_id_ferc714_xbrl == xbrl.eia_code
+        )
+        xbrl.loc[code_is_respondent_id_mask, "eia_code"] = pd.NA
+
+        # lets null out some of the eia_code's from XBRL that we've manually culled
```
aesharpe, line 484: By manually culled do you mean that these IDs are not the actual eia_code?

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +361,223 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+            xbrl.loc[
+                (xbrl.respondent_id_ferc714_xbrl == rid_xbrl)
+                & (xbrl.eia_code.isin(bad_eia_codes)),
+                "eia_code",
+            ] = pd.NA
+        return xbrl
+
+    @staticmethod
+    def convert_into_static_table_xbrl(xbrl: pd.DataFrame) -> pd.DataFrame:
+        """Convert this annually reported table into a skinner, static table.
```
aesharpe, line 502: Do you mean skinnier?

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +361,223 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+        whether or not the shared variables change over time and then
+        converting this table into a non-time-varying table.
+        """
+        # the CSV data does not vary by year, so we need to check if that is
+        # also going to be the case for the XBRL data. we check the eia_codes
+        # during ensure_eia_code_uniqueness. The name is less crucial but we
+        # should still check.
+        assert all(
+            xbrl.groupby(["respondent_id_ferc714_xbrl"])[  # noqa: PD101
+                ["respondent_name_ferc714"]
```
aesharpe, line 516: Is it possible there are some different spellings of the respondent name that would cause this to look like a 1:1 ratio when it's not?

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +361,223 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+        # name is the name we get
+        return (
+            xbrl.sort_values(["report_year"], ascending=False)[cols_to_keep]
+            .sort_values(["respondent_id_ferc714", "eia_code"])
+            .drop_duplicates(subset=["respondent_id_ferc714"], keep="first")
+        )
+
+    @staticmethod
+    def condense_into_one_source_table(df):
+        """Condense the CSV and XBRL records together into one record.
```
aesharpe, line 538: Not blocking at all, but in the other two tables we just put this code directly into the `run` function instead of creating it's own function wrapper. NBD, but it would be good to be consistent.

aesharpe: OR we could very easily make this a function vs. a class method because I think it's the same for all tables. We could just have a column parameter or something.

aesharpe (commented):
Mostly non-blocking comments or questions!

src/pudl/transform/ferc714.py
```diff
@@ -204,7 +222,13 @@
         "csv": {
             "respondent_id": "respondent_id_ferc714_csv",
             "respondent_name": "respondent_name_ferc714",
-        }
+            "eia_code": "eia_code",
+        },
+        "xbrl": {
+            "entity_id": "respondent_id_ferc714_xbrl",
+            "respondent_legal_name": "respondent_name_ferc714",
+            "respondent_identification_code": "eia_code",
```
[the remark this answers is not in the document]
aesharpe: I supposed the csv version had eia_code so it makes sense that this is too.

src/pudl/transform/ferc714.py
aesharpe: Also not sure I understand why this is important when we drop the respondent id source columns anyways?

src/pudl/transform/ferc714.py
cmgosnell: on this branch its being called in the annual table! so i wanted to leave this note to remind myself to rip it out once we get all these tables merged into one branch

aesharpe: We should create a list somewhere. I'll add it to your original hourly PR

src/pudl/transform/ferc714.py
cmgosnell: Zane gave me verbal feedback about supporting my leaning towards keeping the source IDs at least in the long hourly table.

This step really only makes sense to run *after* we have added in the `respondent_id_ferc714`. so everything has a pudl-derived id. but also in the context of the hourly table, doing this would enable users to come to the data with the source ID that they are the most familiar with and get the full time series (so if they knew the CSV-based respondent ID and selected records based on that they'd get the all of the years for that respondent including XBRL years if the respondent is mapped the an xbrl respondent.

src/pudl/transform/ferc714.py
cmgosnell: i thought about combining them all of keeping them all separate. I liked this option the best because i could only add things to the combined section if they actually needed to get fixed in both sources. I will add docs about it!

src/pudl/transform/ferc714.py
cmgosnell: this is fair to flag bc i guess it does look weird. but i think this answer will be more clear if you read the response to your question above. but if its still unclear lmk! also i'll try to update the docs!

src/pudl/transform/ferc714.py
cmgosnell: this isn't a concat step like the other ones. actually this table does get concat-ed first. the output of this table is one record per respondent, so before this we have one from csv and one from xbrl. we've made sure the eia_codes are unique and this step just drop dupes.

src/pudl/transform/ferc714.py
aesharpe: That's kind of what I though--I think your docs updates will help!

src/pudl/transform/ferc714.py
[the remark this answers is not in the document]
cmgosnell: yea its not definitely an eia code in its raw form.... but we clean it up to kinda force it to be that

src/pudl/transform/ferc714.py
cmgosnell: I think it will be nice to add them in to that table as well!

src/pudl/transform/ferc714.py
cmgosnell: okay i added much more into that doc string. i'm going to resolve this comment but feel free to un-resolve it if it still needs attention.

src/pudl/transform/ferc714.py
cmgosnell: yes i added more details here

src/pudl/transform/ferc714.py
cmgosnell: i was worried that there would be different names here, which is why i added this assert in here.

src/pudl/transform/ferc714.py
```diff
@@ -174,37 +174,67 @@
...
+        329: 39347,  # East Texas Electricity Cooperative (missing)
+    },
 }
-"""Overrides of FERC 714 respondent IDs with wrong or missing EIA Codes."""
+"""Overrides of FERC 714 respondent IDs with wrong or missing EIA Codes.
+
+This is used in :meth:`RespondentId.spot_fix_eia_codes`. The dictionary
+is organized by "source" keys ("combined", "csv", or "xbrl"). Each source's
+value is a secondary dictionary which contains source respondent ID's as keys
+and fixes for EIA codes as values.
```
aesharpe, line 223: Can you add an explanation here as to why there is a "combined" section?

src/pudl/transform/ferc714.py
```diff
@@ -293,6 +323,36 @@ def _assign_respondent_id_ferc714(
...
+    """Fill missing CSV or XBRL respondent id.
+
+    The source (CSV or XBRL) tables get assigned a PUDL-derived
+    ``respondent_id_ferc714`` ID column (via :func:`_assign_respondent_id_ferc714`).
+    After we concatenate the source tables, we sometimes to backfill and
+    forward-fill the source IDs (``respondent_id_ferc714_csv`` and
+    ``respondent_id_ferc714_xbrl``). This way the older records from the CSV years
+    will also have the XBRL ID's and vice versa. This will enable users to find
+    the full timeseries of a respondent that given either source ID (instead of
+    using the source ID to find the PUDL-derived ID and then finding the records).
```
aesharpe, line 338: ohh ok I understand now! It's backfilling old, non XBRL years and forward filling non CSV years.

There's one little type here: `we sometimes to backfill`

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +377,235 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+        df = (
+            pd.concat([csv, xbrl])
+            .reset_index(drop=True)
+            .convert_dtypes()
+            .pipe(cls.spot_fix_eia_codes, "combined")
+            .pipe(cls.ensure_eia_code_uniqueness, "combined")
+            .pipe(cls.condense_into_one_source_table)
+            .pipe(_fillna_respondent_id_ferc714_source, "csv")
+            # the xbrl version of this is fillna is not *strictly necessary*
+            # bc we are sorting the records to grab the xbrl record
```
aesharpe, line 447: I don't understand this comment...but I do understand the `_fillna` function now!

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +377,235 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+        """Spot fix the eia_codes.
+
+        We have manually compiled fixes to the EIA Codes that are reported
+        in :py:const:`EIA_CODE_FIXES`. We separated these fixes by either coming
+        directly from the CSV data, the XBRL data, or the combined data. We use the
+        corresponding source or PUDL-derived respondent ID to identify the EIA code to
+        overwrite. We could have combined these fixes all into one set of combined fixes
+        identified by the PUDL-derived ``respondent_id_ferc714``, but this way we can do
+        more targeted source-based cleaning and test each source's EIA codes before the
+        sources are concatenated together.
```
aesharpe, line 465: I feel like some of this info can live in the dictionary description instead.

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +377,235 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+            == 1
+        )
+        cols_to_keep = [
+            "respondent_id_ferc714",
+            "respondent_id_ferc714_xbrl",
+            "respondent_name_ferc714",
+            "eia_code",
+        ]
+        # we are not checking whether the respondent_name_ferc714 is exactly
+        # consistent across the submissions so before we grab the one true eia_code
```
aesharpe, line 555: Is this not what's happening in the assertion statement above?

src/pudl/transform/ferc714.py
```diff
@@ -315,35 +377,235 @@ def _post_process(df: pd.DataFrame, table_name: str) -> pd.DataFrame:
...
+        """Condense the CSV and XBRL records together into one record.
+
+        We have two records coming from each of the two sources in this table.
+        This method simply drops duplicates based on the PKs of the table.
+        We know that the names are different in the CSV vs the XBRL source.
+        We are going to grab the XBRL names because they are more recent.
+
+        NOTE: We could have merged the data in run instead of concatenating
+        along the index. We would have had to develop different methods for
+        :meth:`ensure_eia_code_uniqueness`.
```
aesharpe, line 575: What does "merge the data in run" mean?

aesharpe (approved):
Just some questions and non-blocking stuff!

src/pudl/transform/ferc714.py
aesharpe: ohhh the function run. Maybe make that a tad clearer?

src/pudl/transform/ferc714.py
cmgosnell: oh yea you are right

src/pudl/transform/ferc714.py
cmgosnell: ah yes! I will add a :meth: reference here so that is more clear!

src/pudl/transform/ferc714.py
cmgosnell: i kinda like the explanation better down in the place where the transform actually happens... but i moved it up here.

src/pudl/transform/ferc714.py
cmgosnell: i change the language here a bit to reference this sorting i'm talking about.

PR #3860, opened 2024-09-20
Add FERC1 validation tests

jdangerx:
We've add

This lets us test things that aren't in `PudlTabl`.

# Overview

@cmgosnell and I were working through some changes to the FERC1 machinery, related to the 714 work, and wanted to make sure things didn't break unexpectedly. This reminded us that only a small subset of our FERC tables are covered by row-count tests!

So I added those row-count tests. That required the following changes:

* using Dagster's `AssetValueLoader` to pull the dataframes out of Parquet. This is more performant than `defs.load_asset_value` for loading multiple assets.
* removing several database-access fixture references. Since these validation tests all check for `--live-dbs` I figured we didn't even need to keep those fixtures around to keep track of logical dependencies.
* changing the short `df_name`s that refer to PudlTabl methods to the asset keys we use everywhere else

A fun bonus is that using the `AssetValueLoader` decreased the runtime of this test suite from ~40s to ~8s on my machine. I then re-increased that to ~14s by adding a bunch more tests, but that is still much faster than before 🎉 

# Testing

* downloaded the latest nightly build outputs
* `pytest test/validate/ferc1.py --live-dbs`

```[tasklist]
# To-do list
```


zaneselvans:
Wasn’t there some issue with accessing Dagster assets via dagster in the tests because there’s no real `$DAGSTER_HOME` associated with the in-process executor? Does that not apply here or get around it somehow because the validation tests are only run with `--live-dbs?`

We have another pattern associated with some of our newer tables for verifying expected row counts of tables that aren’t part of PudlTabl using asset checks, with per-year row expectations so it can work on both the fast and full ETL and get run during the ETL instead of after the fact and only in the nightly builds, which would surface errors sooner in the development process. I had imagined we’d probably move in that direction as we rip out PudTabl.  E.g. in `pudl.transform.ferc714` the `spec_checks` at the end of the module:

```py

@dataclass
class Ferc714CheckSpec:
    """Define some simple checks that can run on FERC 714 assets."""

    name: str
    asset: str
    num_rows_by_report_year: dict[int, int]


check_specs = [
    Ferc714CheckSpec(
        name="yearly_planning_area_demand_forecast_check_spec",
        asset="core_ferc714__yearly_planning_area_demand_forecast",
        num_rows_by_report_year={
            2006: 1819,
            2007: 1570,
            2008: 1540,
            2009: 1269,
            2010: 1259,
            2011: 1210,
            2012: 1210,
            2013: 1192,
            2014: 1000,
            2015: 990,
            2016: 990,
            2017: 980,
            2018: 961,
            2019: 950,
            2020: 950,
        },
    )
]


def make_check(spec: Ferc714CheckSpec) -> AssetChecksDefinition:
    """Turn the Ferc714CheckSpec into an actual Dagster asset check."""

    @asset_check(asset=spec.asset, blocking=True)
    def _check(df):
        errors = []
        for year, expected_rows in spec.num_rows_by_report_year.items():
            if (num_rows := len(df.loc[df.report_year == year])) != expected_rows:
                errors.append(
                    f"Expected {expected_rows} for report year {year}, found {num_rows}"
                )

        if errors:
            return AssetCheckResult(passed=False, metadata={"errors": errors})

        return AssetCheckResult(passed=True)

    return _check


_checks = [make_check(spec) for spec in check_specs]
```

jdangerx:
Ooh, I like the per-year checks, that's definitely more powerful! And being an asset check seems nice as well.

> Wasn’t there some issue with accessing Dagster assets via dagster in the tests because there’s no real $DAGSTER_HOME associated with the in-process executor? Does that not apply here or get around it somehow because the validation tests are only run with --live-dbs?

Yeah, this gets around it with `--live-dbs`.

I think the next steps then are:

1. port the existing FERC1 minmax row asset checks to asset-check world
2. add the new FERC1 checks in asset-check world.

With maybe step 0 of turning `Ferc714CheckSpec` into `MinMaxCheck` or something. Is `name` just used for human identification when looking at the code? Seems a little redundant with the `asset` key.

Though if we want to merge these tests first to unblock the 714 work, that seems like it would be fine too.

zaneselvans:
Yeah, nothing wrong with the set up in this PR, more thinking about the future we want to migrate toward, and whether we want to start in that direction now.

My reading on the structure & naming was that there would be a variety of different asset checks parameterized within the `Ferc714ChecSpec` class -- e.g. no completely null columns, or the statistical distribution of coal heat contents or whatever flavors of checks we've got floating around in the `pudl.validate` module now -- in which case `MinMaxCheck` wouldn't make sense.

jdangerx:
> Yeah, nothing wrong with the set up in this PR, more thinking about the future we want to migrate toward, and whether we want to start in that direction now.

Agree. I think this is the shortest path to getting long overdue tests into the codebase and doesn't make the migration harder, so we should merge this roughly as-is.

> My reading on the structure & naming was that there would be a variety of different asset checks parameterized within the `Ferc714ChecSpec` class -- e.g. no completely null columns, or the statistical distribution of coal heat contents or whatever flavors of checks we've got floating around in the `pudl.validate` module now -- in which case `MinMaxCheck` wouldn't make sense.

I think there are common building blocks like `MinMaxRowsCheck` or `QuantilesCheck` that we'd then want to bundle together in various `Ferc714` or `EiaAeo` checks... though I think we should look closer into just using Great Expectations before rolling our own format for these.

> And if I understand the way asset checks work, each one of the reads in the asset to do the check, so ideally there'd be some way to just read the parquet file once and do all the checks while it's in memory, rather than reading it repeatedly.

Yeah, that makes sense, doesn't seem like that would be hard to set up.

test/conftest.py
```diff
@@ -105,6 +107,11 @@ def live_databases(request) -> bool:
     return request.config.getoption("--live-dbs")
 
 
+@pytest.fixture(scope="session")
+def asset_value_loader() -> AssetValueLoader:
+    return defs.get_asset_value_loader()
+
+
```
cmgosnell, line 114: what is the benefit of having this as a session fixture instead of just making every test call `defs.get_asset_value_loader`? Is it just to avoid all the extra `defs` and imports and such? Either way it'd be good to add docs in here to communicate how to use this.

test/validate/ferc1_test.py
```diff
@@ -59,76 +58,109 @@ def test_record_id_dupes(pudl_engine, table_name):
...
         pytest.skip("Data validation only works with a live PUDL DB.")
     pv.no_null_cols(
-        pudl_out_ferc1.__getattribute__(df_name)(), cols=cols, df_name=df_name
+        asset_value_loader.load_asset_value(asset_key), cols=cols, df_name=asset_key
     )
 
 
+# TODO 2024-09-20: many of these row counts haven't actually been confirmed to
+# be "correct", they're just "what the values were when we added them to the
+# test." Those tables are marked as "unverified" below
```
cmgosnell, line 86: in a very real way i think the main way we have validated these in the past is just that they stay constant or change in direction and scale that we expect (i.e. if we had 10 records for 10 years of data and we add a new year we should get 11 records). So while i like the idea of doing some more nuanced validation here, i'm not sure this unverified-ness makes a whole lot of a difference.

test/validate/ferc1_test.py
```diff
@@ -59,76 +58,109 @@ def test_record_id_dupes(pudl_engine, table_name):
...
     Args:
-        pudl_out_ferc1: A PudlTabl output object.
         live_dbs: Boolean (wether we're using a live or testing DB).
         expected_rows (int): Expected number of rows that the dataframe should
             contain when all data is loaded and is output without aggregation.
-        df_name (str): Shorthand name identifying the dataframe, corresponding
-            to the name of the function used to pull it from the PudlTabl
-            output object.
+        asset_key (str): Shorthand name identifying the dataframe, corresponding
+            to the name of the asset.
```
cmgosnell, line 130: oh this is a nit but this no longer a shorthand name

docs/release_notes.rst
```diff
@@ -16,6 +16,26 @@ Bug Fixes
 ^^^^^^^^^
 * Include more retiring generators in the net generation and fuel consumption
   allocation. Thanks to :user:`grgmiller` for this contirbution :pr:`3690`.
+* Added preliminary data validation checks for several FERC 1 tables that were
+  missing it:
+
+  * out_ferc1__yearly_balance_sheet_assets_sched110
```
cmgosnell, line 22: this is a complete personal preference but I am not sure what a user reading our release notes will get out of this list of tables. which is to say i would keep the note but not the list in here. but 1000% not a hill i will die on its so fine.

cmgosnell (changes requested):
sorry for letting this languish!! two lil guys (a request for docs and a non-blocking nudge to remove the unverified notes)

aesharpe (commented):
I don't have any additional comments - just to merge with main and make sure there are no conflicts in the release notes (will probably have to move these now that we've done another release).

test/conftest.py
jdangerx: It makes things less verbose, yes, but also asset loaders are kind of slow to initialize so this gets us a performance win too.

cmgosnell: okay so even though w/o this guy you'll still need to load_asset_value every time w/ this fixture you've initialized the asset loader

jdangerx: `defs.load_asset_value()` makes a new asset loader each time; `asset_value_loader.load_asset_value()` (🙄) uses the pre-initialized one

cmgosnell (approved):

PR #3861, opened 2024-09-20
Migrate where/when we filter for the freshest XBRL data

cmgosnell:
# Overview
This was done in conjunction with @jdangerx !

It came to our attention in #3857 that `filter_for_freshest_data` being inside of the XBRL io_manager was... a lot! It included a pretty slow step that was mostly a validation step.

What did you change?
- move the `filter_for_freshest_data` processing out of the `io_manager` into `transform.ferc1`. this took the longest to figure out where/when in the process this should happen.... we ended up putting it in `ferc1_transform_asset_factory` because that is where the raw tables get concatenated when there is more than one input raw table. So that is the last place where we have individual raw tables.
- move the `__compare_dedupe_methodologies` into a validation test
- move the unit tests
- also added a little defensive check in `ferc1_transform_asset_factory` to make sure all of the raw tables either have an instant or duration table... because we had a type in one of the input tables 😬 

Implications??
- after this any non-ferc1 raw xbrl table that gets extracted will no longer be filtered in this way. That actually seems like it is going to be a problem for the annual 714 table that @aesharpe  is processing in #3856.

# Testing

```[tasklist]
# To-do list
- [x] Run all of the core_ferc1 assets!
```


src/pudl/extract/ferc1.py
```diff
@@ -197,7 +197,7 @@
             "sales_of_electricity_by_rate_schedules_account_445_other_sales_to_public_authorities_304",
             "sales_of_electricity_by_rate_schedules_account_446_sales_to_railroads_and_railways_304",
             "sales_of_electricity_by_rate_schedules_account_448_interdepartmental_sales_304",
-            "sales_of_electricity_by_rate_schedules_account_4491_provision_for_rate_refunds_304",
+            "sales_of_electricity_by_rate_schedules_account_449_1_provision_for_rate_refunds_304",
```
cmgosnell, line 200: ugh this was just a bug/typo. this 4991 table does not/did not exist. I added an assertion into the form1 transforms to make sure we either get an instant or a duration version of all of the source assets.

src/pudl/transform/ferc1.py
```diff
@@ -6121,13 +6179,13 @@ def ferc1_transform_asset_factory(
...
-    ins = {f"raw_dbf__{tn}": AssetIn(f"raw_ferc1_dbf__{tn}") for tn in dbf_tables}
+    ins = {f"raw_ferc1_dbf__{tn}": AssetIn(f"raw_ferc1_dbf__{tn}") for tn in dbf_tables}
     ins |= {
-        f"raw_xbrl_instant__{tn}": AssetIn(f"raw_ferc1_xbrl__{tn}_instant")
+        f"raw_ferc1_xbrl__{tn}_instant": AssetIn(f"raw_ferc1_xbrl__{tn}_instant")
         for tn in xbrl_tables
     }
     ins |= {
-        f"raw_xbrl_duration__{tn}": AssetIn(f"raw_ferc1_xbrl__{tn}_duration")
+        f"raw_ferc1_xbrl__{tn}_duration": AssetIn(f"raw_ferc1_xbrl__{tn}_duration")
```
cmgosnell, line 6188: i changed these keys to be the same as the asset name just bc that was confusing to keep straight

src/pudl/transform/params/ferc1.py
```diff
@@ -4854,6 +4854,7 @@
                 "other_sales_to_public_authorities_axis",
                 "sales_to_railroads_and_railways_axis",
                 "interdepartmental_sales_axis",
+                "provision_for_rate_refunds_axis",
```
cmgosnell, line 4857: this got added in bc this is effectively a new table that we are now processing as a part of the rate schedule table bc of fixing that typo

test/unit/io_managers_test.py
```diff
@@ -359,161 +355,6 @@ def test_error_when_reading_view_without_metadata(fake_pudl_sqlite_io_manager_fi
         fake_pudl_sqlite_io_manager_fixture.load_input(input_context)
 
 
-def test_ferc_xbrl_sqlite_io_manager_dedupes(mocker, tmp_path):
```
cmgosnell, line 362: this got moved over into the transform unit test. but it also got significantly gutted bc we needed basically none of this complex setup to test it now

test/unit/io_managers_test.py
```diff
@@ -359,161 +355,6 @@ def test_error_when_reading_view_without_metadata(fake_pudl_sqlite_io_manager_fi
...
-        },
-    )
-    io_manager = FercXBRLSQLiteIOManager(base_dir=tmp_path, db_name="ferc1_test_db")
-    observed_table = io_manager.load_input(input_context)
-
-    assert len(observed_table) == 1
-    assert observed_table.str_factoid.to_numpy().item() == "updated 2021 EOY value"
-
-
-example_schema = pandera.DataFrameSchema(
```
cmgosnell, line 462: this example schema one also got moved and that didn't involve nearly as much gutting

test/validate/ferc1_test.py
```diff
@@ -160,3 +164,116 @@ def test_unique_rows_ferc1(pudl_out_ferc1, live_dbs, df_name, unique_subset):
...
+
+@pytest.mark.parametrize(
+    "table_name",
+    [
+        "core_ferc1__yearly_sales_by_rate_schedules_sched304",
+        "core_ferc1__yearly_pumped_storage_plants_sched408",
+        "core_ferc1__yearly_steam_plants_sched402",
+        "core_ferc1__yearly_hydroelectric_plants_sched406",
+        "core_ferc1__yearly_transmission_lines_sched422",
+    ],
```
cmgosnell, line 177: a sample of the widest tables dazhong could find

test/validate/ferc1_test.py
```diff
@@ -160,3 +164,116 @@ def test_unique_rows_ferc1(pudl_out_ferc1, live_dbs, df_name, unique_subset):
...
+)
+def test_filter_for_freshest_data(
+    ferc1_engine_xbrl, table_name: pd.DataFrame
+) -> pd.DataFrame:
+    """Test if we are unexpectedly replacing records during filter_for_freshest_data.
+
+    NOTE: None of this actually calls any of the code over in pudl.transform.ferc1 rn.
+    That's maybe okay because we are mostly testing the data updates mostly here.
```
cmgosnell, line 185: does this feel okay? we could pull some of the bits from the transform filter out into smaller functions to actually run here... but i don't think it is worth it looking at the bits and the setup.

test/unit/transform/ferc1_test.py
```diff
@@ -1231,3 +1235,92 @@ def test_unexpected_total_components():
...
+@pytest.mark.slow
+@hypothesis.settings(print_blob=True, deadline=2_000)
+@hypothesis.given(example_schema.strategy(size=3))
+def test_filter_for_freshest_data_xbrl(df):
+    # XBRL context is the identifying metadata for reported values
+    xbrl_context_cols = ["entity_id", "date", "utility_type"]
+    filing_metadata_cols = ["publication_time", "filing_name"]
+    primary_key = xbrl_context_cols + filing_metadata_cols
+    deduped = filter_for_freshest_data_xbrl(df, primary_keys=primary_key)
+    example_schema.validate(deduped)
```
cmgosnell, line 1306: hm @jdangerx  this is failing rn bc the `validate` expected the `deduped` to have `publication_time` in there.. which i am now dropping as a part of `filter_for_freshest_data_xbrl`. I could just not drop it in there are always remember to drop it after `filter_for_freshest_data_xbrl` is called. but idk it does seem like it should be dropped always so it'd be nice to just make the test work. any thoughts/ideas?

```python
test/unit/transform/ferc1_test.py:1306: in test_filter_for_freshest_data_xbrl
    example_schema.validate(deduped)
../../miniforge3/envs/pudl-dev/lib/python3.12/site-packages/pandera/api/pandas/container.py:126: in validate
    return self._validate(
../../miniforge3/envs/pudl-dev/lib/python3.12/site-packages/pandera/api/pandas/container.py:155: in _validate
    return self.get_backend(check_obj).validate(
../../miniforge3/envs/pudl-dev/lib/python3.12/site-packages/pandera/backends/pandas/container.py:105: in validate
    error_handler = self.run_checks_and_handle_errors(
../../miniforge3/envs/pudl-dev/lib/python3.12/site-packages/pandera/backends/pandas/container.py:180: in run_checks_and_handle_errors
    error_handler.collect_error(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <pandera.api.base.error_handler.ErrorHandler object at 0x28f24cc50>, error_type = <ValidationScope.SCHEMA: 'schema'>
reason_code = <SchemaErrorReason.COLUMN_NOT_IN_DATAFRAME: 'column_not_in_dataframe'>
schema_error = SchemaError("column 'publication_time' not in dataframe. Columns in dataframe: ['entity_id', 'date', 'utility_type', 'int_factoid', 'float_factoid', 'str_factoid']")
original_exc = None

    def collect_error(
        self,
        error_type: ErrorCategory,
        reason_code: SchemaErrorReason,
        schema_error: SchemaError,
        original_exc: Union[BaseException, None] = None,
    ):
        """Collect schema error, raising exception if lazy is False.

        :param error_type: type of error
        :param reason_code: string representing reason for error
        :param schema_error: ``SchemaError`` object.
        """
        if not self._lazy:
>           raise schema_error from original_exc
E           pandera.errors.SchemaError: column 'publication_time' not in dataframe. Columns in dataframe: ['entity_id', 'date', 'utility_type', 'int_factoid', 'float_factoid', 'str_factoid']
E           Falsifying example: test_filter_for_freshest_data_xbrl(
E               df=
E                     entity_id       date utility_type publication_time  int_factoid  float_factoid str_factoid
E                   0         C 1970-01-01     electric       1970-01-01            0            0.0
E                   1         C 1970-01-01     electric       1970-01-01            0            0.0
E                   2         C 1970-01-01     electric       1970-01-01            0            0.0
E               ,  # or any other generated value
E           )
E
E           You can reproduce this example by temporarily adding @reproduce_failure('6.112.1', b'AXicY2AYTAAAAI4AAQ==') as a decorator on your test case

../../miniforge3/envs/pudl-dev/lib/python3.12/site-packages/pandera/api/base/error_handler.py:54: SchemaError
```

src/pudl/transform/ferc1.py
```diff
@@ -6092,6 +6093,63 @@ class OtherRegulatoryLiabilitiesTableTransformer(Ferc1AbstractTableTransformer):
 }
 
 
+def filter_for_freshest_data_xbrl(df: pd.DataFrame, primary_keys) -> pd.DataFrame:
```
jdangerx, line 6096: If we want to use these for 714 also, we'll want to pull it into its own `pudl.transform.ferc_utils` module or something... but for now this is fine!

src/pudl/transform/ferc1.py
```diff
@@ -6092,6 +6093,63 @@ class OtherRegulatoryLiabilitiesTableTransformer(Ferc1AbstractTableTransformer):
...
+    either reports a null value for it or simply omits it from the report,
+    we keep the old non-null value, which may be erroneous. This appears to
+    be fairly rare, affecting < 0.005% of reported values.
+    """
+
+    def __apply_diffs(
+        duped_groups: pd.core.groupby.DataFrameGroupBy,
+    ) -> pd.DataFrame:
+        """Take the latest reported non-null value for each group."""
+        return duped_groups.last()
```
jdangerx, line 6120: non-blocking: we had talked about in-lining this function, do you still want to do that?

src/pudl/transform/ferc1.py
```diff
@@ -6160,13 +6218,31 @@ def ferc1_transform_asset(**kwargs: dict[str, pd.DataFrame]) -> pd.DataFrame:
...
+            for tn, df in kwargs.items()
+            if tn.startswith("raw_ferc1_xbrl__")
+        }
+        for raw_xbrl_table_name in listify(TABLE_NAME_MAP_FERC1[table_name]["xbrl"]):
+            if (
+                raw_xbrls[f"raw_ferc1_xbrl__{raw_xbrl_table_name}_instant"].empty
+                and raw_xbrls[f"raw_ferc1_xbrl__{raw_xbrl_table_name}_duration"].empty
+            ):
+                raise AssertionError(
+                    "We expect there to be no raw xbrl tables that have neither instant or duration "
```
jdangerx, line 6237: nit: `f"{raw_xbrl_table_name} has neither instant nor duration tables. Is it spelled correctly in pudl.extract.ferc1.TABLE_NAME_MAP_FERC1?"` is a bit more information dense

jdangerx (changes requested):
Whee! A few changes requested in testing (mostly around actually testing the code we're running in production) but otherwise looking good!

src/pudl/transform/ferc1.py
cmgosnell: yea... now that the 714 stuff got pulled into main i do think a new module is warrented.

src/pudl/transform/ferc1.py
cmgosnell: yea i think it's necessary to actuall run it over in the tests.

test/validate/ferc1_test.py
```diff
@@ -160,3 +164,116 @@ def test_unique_rows_ferc1(pudl_out_ferc1, live_dbs, df_name, unique_subset):
...
+                f"currently  implemented apply_diffs methodology (#{n_diffs}) as "
+                f"compared to the best_snapshot methodology (#{n_best}). We expected"
+                " the apply_diffs methodology to result in no more than "
+                f"{threshold_ratio:.2%} non-null records but found {found_ratio:.2%}.\n\n"
+                "We are concerned about excess non-null values because apply-diffs "
+                "grabs the most recent non-null values. If this error is raised, "
+                "investigate filter_for_freshest_data."
+            )
+
+    table_name = "core_ferc1__yearly_sales_by_rate_schedules_sched304"
```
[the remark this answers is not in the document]
cmgosnell: oh doh yes ty!

test/unit/transform/ferc1_test.py
cmgosnell: okay this is simple. does this mean that example_schama is a dataframe when it is in here (i've never worked with hypothesis)

test/validate/ferc1_test.py
cmgosnell: hm after trying to do this, the main snag here is that `filter_for_freshest` does the prep to get the `never_duped` and `duped_groups`, does the filtering, then concats them back together.

So instead of passing in an optional `deduper` method I think it'll be simpler to pass in a `compare_methods` bool, which defaults to `False`

test/validate/ferc1_test.py
```diff
@@ -160,3 +164,116 @@ def test_unique_rows_ferc1(pudl_out_ferc1, live_dbs, df_name, unique_subset):
...
+
+        The main thing we want to keep tabs on is apply-diff adding new
+        non-null values compared to best-snapshot, because some of those
+        are instances of a value correctly being reported as `null`.
+
+        Instead of stacking the two datasets, merging by context, and then
+        looking for left_only or right_only values, we just count non-null
+        values. This is because we would want to use the report_year as a
+        merge key, but that isn't available until after we pipe the
+        dataframe through `refine_report_year`.
```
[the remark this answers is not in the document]
cmgosnell: or do you mean merge the two filtered dfs and check column by column for nulls in one filter method but not in the other?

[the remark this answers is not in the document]
jdangerx: Sorry, I wasn't clear! I had meant that we should check each individual XBRL fact to see if it had changed:
* from null in apply-diffs to not-null in snapshot
* vice versa
* from one value in apply-diffs to a different value in snapshot

Originally / in this comment, I had envisioned this Pandas workflow:

* stack datasets so there's only one data column
* merge by XBRL context columns + variable name column
* look at the prevalence of `left_only`/`right_only` values & any rows where the `value_left` and `value_right` were different

But! I think your plan of merging the dfs and checking column-by-column also works! In my head, the merge indicator operates on a row-by-row basis, so I figured stacking to one-xbrl-fact-per-row would be easiest. But you are the one who is writing this and also has more `pandas` experience so I'll defer to your judgment.

src/pudl/transform/ferc1.py
```diff
@@ -305,7 +306,7 @@ def wide_to_tidy(df: pd.DataFrame, params: WideToTidy) -> pd.DataFrame:
     )
     df_out.columns = new_cols
     df_out = (
-        df_out.stack(params.stacked_column_name, dropna=False)
+        df_out.stack(params.stacked_column_name, future_stack=True)
```
cmgosnell, line 309: sorry this is fully unrelated... but this was to remove an annoying future warning. i tested the update by re-running all the core and out ferc assets then running the ferc validation tests.

src/pudl/workspace/setup.py
```diff
@@ -30,6 +30,7 @@ def create_directories(self: Self):
         """Create PUDL input and output directories if they don't already exist."""
         self.input_dir.mkdir(parents=True, exist_ok=True)
         self.output_dir.mkdir(parents=True, exist_ok=True)
+        return self
```
cmgosnell, line 33: this was to remove another more different warning about `model_validator`'s needing `self` to be returned. i honestly don't know how this didn't junk everything up before

test/validate/ferc1_test.py
```diff
@@ -160,3 +164,44 @@ def test_unique_rows_ferc1(pudl_out_ferc1, live_dbs, df_name, unique_subset):
...
+        "core_ferc1__yearly_pumped_storage_plants_sched408",
+        "core_ferc1__yearly_steam_plants_sched402",
+        "core_ferc1__yearly_hydroelectric_plants_sched406",
+        "core_ferc1__yearly_transmission_lines_sched422",
+        # some sample guys found to have higher filtering diffs
+        "core_ferc1__yearly_utility_plant_summary_sched200",
+        "core_ferc1__yearly_plant_in_service_sched204",
+        "core_ferc1__yearly_operating_expenses_sched320",
+        "core_ferc1__yearly_income_statements_sched114",
+    ],
```
cmgosnell, line 182: i added some more guys

test/validate/ferc1_test.py
```diff
@@ -160,3 +164,44 @@ def test_unique_rows_ferc1(pudl_out_ferc1, live_dbs, df_name, unique_subset):
...
+        for tn in raw_table_names
+    )
+    for raw_table_name in xbrls_with_periods:
+        logger.info(f"Checking if our filtering methodology works for {raw_table_name}")
+        xbrl_table = defs.load_asset_value(raw_table_name)
+        if not xbrl_table.empty:
+            primary_keys = get_primary_key_raw_xbrl(
+                raw_table_name.removeprefix("raw_ferc1_xbrl__"), "ferc1"
+            )
+            filter_for_freshest_data_xbrl(
```
jdangerx, line 205: It's a little confusing that the assertions don't live in the test function, but I guess this works! I think someone confused about what this is actually testing would just go into `filter_for_freshest` and immediately see the assertion errors.

jdangerx (approved):
🚢

test/validate/ferc1_test.py
cmgosnell: yea i absolutely agree with this. it just would have been a lot of contrived passing things around and making little functions to get the right inputs in the right place. this feels simpler for now but also i can def imagine wanting to come back and break this up so the __compare_dedupe_methodologies step happens entirely in the test.

cmgosnell:
ah i re-added this to the merge queue w/o pushing my changes.... oopsies.

PR #3870, opened 2024-09-26
Lint GHA workflows in pre-commit.

jdangerx:
# Overview

This takes basically no additional time in pre-commit and will stop you from accidentally committing a bad GHA workflow & tearing your hair out over it.

## What did you change?

Add `actionlint`, which appears to be the community standard GHA linter.

# Testing

I added a typo and it yelled at me.


zaneselvans:
Oh this is a good idea. You want to add it to our other repos as well? The archiver, cheshire, user metrics, and ferc xbrl extractor?

PR #3883, opened 2024-09-30
Create an asset factory for FERC 1 output tables

### LINKED ISSUE #3147: Consolidate ferc1 outputs using Dagster asset factories

e-belfer:
# Overview

Closes #3147 and supercedes #3557.

Taking over the draft PR in #3557 started by @hfireborn and finishing it. Creates an asset factory for all identical `out_ferc1__` tables which just join in the utility IDs. Adding more complex FERC1 output tables into this pattern is out of scope for now.

## What did you change?
Adds `ferc1_output_asset_factory` to `pudl.output.ferc1.py` and removes the corresponding table asset definitions.

# Documentation

```[tasklist]
```

# Testing

Regenerated `out_ferc1` asset group and ran `pytest ${pytest_args} -n 4 --no-cov --live-dbs test/validate/ferc1_test.py`.

```[tasklist]
# To-do list
```


zaneselvans (approved):
Thanks for finishing this off and pulling it in.

PR #3887, opened 2024-09-30
Add source metadata for `vceregen`

### LINKED ISSUE #3886: Add source metadata for vceregen data
This is an easy task to do before we get any access to the raw data. Just fill out the `pudl/metadata/sources.py` for this table. This metadata is used by the `pudl-archiver` to understand the data is it trying to archive.

It is used in the programmatic construction of the `datapackage.json` which accompanies our archived data on Zenodo, and which is necessary for the PUDL datastore script to retrieve and verify a particular partition of the dataset.

aesharpe:
# Overview

Closes #3886

## What did you change?

Adds source metadata for `vceregen` so it's ready to be ingested by the archiver.

### DIFF
--- a/src/pudl/metadata/constants.py
+++ b/src/pudl/metadata/constants.py
@@ -186,8 +186,17 @@ CONTRIBUTORS: dict[str, dict[str, str]] = {
         "organization": "Catalyst Cooperative",
         "orcid": "0000-0001-9784-8531",
     },
+    "vibrant-clean-energy": {
+        "title": "Vibrant Clean Energy",
+        "email": "info@vibrantcleanenergy.com",
+        "path": "https://vibrantcleanenergy.com/",
+        "role": "author",
+        "zenodo_role": "producer",
+        "organization": "Vibrant Clean Energy",
+    },
 }
-"""PUDL Contributors for attribution."""
+"""PUDL Contributors for attribution. See the Data Package metadata https://specs.frictionlessdata.io/data-package/#metadata
+for examples. For zenodo_role, see the contributor types in the Zenodo documentation https://developers.zenodo.org/#representation."""
 
 KEYWORDS: dict[str, list[str]] = {
     "electricity": [
--- a/src/pudl/metadata/sources.py
+++ b/src/pudl/metadata/sources.py
@@ -835,5 +835,42 @@ SOURCES: dict[str, Any] = {
         "license_raw": LICENSES["us-govt"],
         "license_pudl": LICENSES["cc-by-4.0"],
     },
+    "vceregen": {
+        "title": "Vibrant Clean Energy Renewable Generation",
+        "path": "https://vibrantcleanenergy.com/products/datasets/",
+        "description": (
+            "Hourly, county-level renewable generation profiles in the continental "
+            "United States. Compiled by Vibrant Clean Energy based on outputs from the "
+            "NOAA HRRR weather model. Profiles are stated as a capacity factor (a "
+            "fraction of nameplate capacity) and exist for onshore wind, offshore "
+            "wind, and fixed-tilt solar generation types."
+            ""
+        ),
+        "source_file_dict": {
+            "sorce_format": "Comma Separated Value (.csv)",
+        },
+        "keywords": sorted(
+            {
+                "solar",
+                "wind",
+                "time series",
+                "energy",
+                "electricity",
+                "generation",
+                "weather",
+                "capacity factor",
+                "hourly",
+                "united states",
+                "usa",
+                "resource adequacy",
+                "gridpath",
+                "vibrant clean energy",
+                "county",
+            }
+        ),
+        "license_raw": LICENSES["cc-by-4.0"],
+        "license_pudl": LICENSES["cc-by-4.0"],
+        "working_partitions": {"years": sorted(set(range(2019, 2024)))},
+    },
 }
 """Data source attributes by PUDL identifier."""
--- a/src/pudl/metadata/sources.py
+++ b/src/pudl/metadata/sources.py
@@ -844,7 +844,6 @@ SOURCES: dict[str, Any] = {
             "NOAA HRRR weather model. Profiles are stated as a capacity factor (a "
             "fraction of nameplate capacity) and exist for onshore wind, offshore "
             "wind, and fixed-tilt solar generation types."
-            ""
         ),
         "source_file_dict": {
             "sorce_format": "Comma Separated Value (.csv)",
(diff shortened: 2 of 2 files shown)


src/pudl/metadata/sources.py
```diff
@@ -835,5 +835,41 @@
...
     },
+    "vceregen": {
+        "title": "Vibrant Clean Energy Renewable Generation Profiles",
+        "path": "https://vibrantcleanenergy.com/products/datasets/",
+        "description": (
+            "Hourly, county-level renewable generation profiles in the continental "
+            "United States. Compiled by Vibrant Clean Energy based on outputs from the "
+            "NOAA HRRR weather model. Profiles are stated as a capacity factor (a "
+            "fraction of nameplate capacity) and exist for onshore wind, offshore "
+            "wind, and fixed-tilt solar generation types."
```
aesharpe, line 846: Based on the description of `gridpathtoolkit`

src/pudl/metadata/sources.py
```diff
@@ -835,5 +835,41 @@
         "license_raw": LICENSES["us-govt"],
         "license_pudl": LICENSES["cc-by-4.0"],
     },
+    "vceregen": {
+        "title": "Vibrant Clean Energy Renewable Generation Profiles",
+        "path": "https://vibrantcleanenergy.com/products/datasets/",
```
aesharpe, line 840: Was not sure which link to put here because the data are proprietary. Could have linked to either of these PDFs as well: 

https://vibrantcleanenergy.com/wp-content/uploads/2021/04/2018%20VCE%20Study_Dataset%20Methods%20and%20Analysis536960.pdf 

https://vibrantcleanenergy.com/wp-content/uploads/2020/08/VCE-Weather-Dataset-Overview_August2020.pdf

src/pudl/metadata/sources.py
```diff
@@ -835,5 +835,41 @@
...
+                "generation",
+                "weather",
+                "capacity factor",
+                "hourly",
+                "united states",
+                "usa",
+                "resource adequacy",
+                "gridpath",
+                "vibrant clean energy",
+                "county",
```
aesharpe, line 867: same as `gridpathtoolkit` but with `vibrant clean energy` and `county` added

src/pudl/metadata/sources.py
```diff
@@ -835,5 +835,41 @@
...
+            }
+        ),
+        "license_raw": LICENSES["cc-by-4.0"],
```
aesharpe, line 870: I assumed it would be the same as `gridpathtoolkit`

e-belfer (approved):
I have a question for @zaneselvans about the licensing.

We might want to add the documentation PDFs somewhere - left some thoughts in the comments.

src/pudl/metadata/sources.py
zaneselvans: I asked Chris in email which was the appropriate document and he said the [Dataset Overview](https://vibrantcleanenergy.com/wp-content/uploads/2020/08/VCE-Weather-Dataset-Overview_August2020.pdf) (Aug, 2020) was the better one to reference.  I also asked permission to archive it alongside the inputs (or the way we do in our GitHub repo) so that when the VCE website goes down it's still available.

src/pudl/metadata/sources.py
zaneselvans: Yes, we got them to agree to CC-BY-4.0 after a bit of discussion. We need to make sure that the dataset is clearly attributed to VCE in the Zenodo archive. And also in the metadata that accompanies the PUDL outputs. This isn't something we've had to deal with in the past because all the input data has come from the US Govt which doesn't have copyrights.

src/pudl/metadata/sources.py
e-belfer: Will we be able to commit a large PDF to git, space-wise?

src/pudl/metadata/sources.py
e-belfer: Ah, that's not in the licensing, that's in the contributors - we need to define a new organization in the CONTRIBUTOR field for VCE and define it here.

zaneselvans: Yeah, that sounds right!  I think so long as VCE has the "creator" role or whatever is equivalent, there's a CC-BY-4.0 license, they're mentioned explicitly in the description (which ends up in the description on Zenodo) and linked to as well, and in the Data Sources page in our docs, and we get analogous appropriate metadata into our Data Release archives on Zenodo, we'll be okay.

e-belfer (changes requested):
We should add VCE in the CONTRIBUTORS dictionary and make them the contributor of this dataset, as discussed in the comments. Also worth seeing if we can check in the key PDF into our data sources folder in Git.

src/pudl/metadata/sources.py
aesharpe: Has he given us the go-ahead to archive the PDF or are we still waiting for a response?

e-belfer: It's public on the website so I don't see why it would be a problem.

src/pudl/metadata/sources.py
```diff
@@ -835,5 +835,44 @@
         "license_raw": LICENSES["us-govt"],
         "license_pudl": LICENSES["cc-by-4.0"],
     },
+    "vceregen": {
+        "title": "Vibrant Clean Energy Renewable Generation Profiles",
+        "path": "https://vibrantcleanenergy.com/wp-content/uploads/2020/08/VCE-Weather-Dataset-Overview_August2020.pdf",
+        "description": (
```
aesharpe, line 841: Could make this a path to the internal file, but wasn't 100% sure how to do that.

e-belfer (approved):
Looks good!

src/pudl/metadata/constants.py
```diff
@@ -186,8 +186,17 @@
...
         "orcid": "0000-0001-9784-8531",
     },
+    "vibrant-clean-energy": {
+        "title": "Vibrant Clean Energy",
+        "email": "info@vibrantcleanenergy.com",
+        "path": "https://vibrantcleanenergy.com/",
+        "role": "author",
+        "zenodo_role": "producer",
+        "organization": "Vibrant Clean Energy",
+    },
```
zaneselvans, line 196: Once we've got an archiver that works for a private file handoff, it'd be good to re-archive the GridPath RA Toolkit data using it, and do something similar for attribution / licensing there.

aesharpe: @e-belfer

e-belfer: Definitely, though slightly tricky to refresh the metadata without new files but it can be done. Want to write up an issue so it doesn't get lost?

zaneselvans: See #3895

Issue #3889, opened 2024-10-01
Investigate source of fill bug + propose some potential fixes


cmgosnell:
On June 18th 2024, we received a question about some notably high values in the `out_eia923__fuel_receipts_costs` table. This was resulting in anomalously high fuel prices in some cases, identified first in deliveries of BIT to plants with plant_id_eia `8102` and `2832`.

These values were not present in `core_eia923__monthly_fuel_receipts_costs`, and it turns out that the error was introduced on our end. In #3716, we turned off the rolling average behavior, which affected about 2% of records in `out_eia923__fuel_receipts_costs`. This is distinct from the values filled from the bulk electricity data.

# Part 1: Diagnose the problem

Success criteria: Identify what section of the code created the incorrectly calculated values, and whether `out_ferc1__yearly_steam_plants_sched402` is also affected.

It's not clear whether the issue is resulting from the filling step, or the calculation of the rolling averages. If it's in the calculation of the rolling averages, then the `out_ferc1__yearly_steam_plants_sched402` table could also be affected, as the same routine is used there to calculate a smoothed estimate of ongoing capital additions.

```[tasklist]
- [x] Download nightly build data from June 18.
- [x] Regenerate the problematic `out_eia923__fuel_receipts_costs` with the rolling behavior turned off, and compare outputs using the code snippet in #3716
- [x] As part of your diagnostic process, write a series of unit tests for `pudl.helpers.fillna_w_rolling_avg()` and `pudl.helpers.generate_rolling_avg()` that make sure the function is behaving as expected and diagnose any potential behavioral failures
- [x] Identify whether the divergence occurs in `pudl.helpers.fillna_w_rolling_avg()` or `pudl.helpers.generate_rolling_avg()`, and what line of code is the source of the problem
- [x] Assess whether the `out_ferc1__yearly_steam_plants_sched402` is also affected by the same problem. (it totally was affecting the capital additions as well... it had a window of 3 so there were less records effected overall but still :grimacing:
```

# Part 2: Design and implement a solution
Success criteria: Decide whether to re-implement the rolling average behavior. If so, identify any necessary code changes and implement them.

If we want to turn this back on, we'll need to not only understand what the problem was, but also to understand how to fix it.
```[tasklist]
- [x] Once the problem is diagnosed, propose a new design for a rolling average or assess the scale of data that would be left unfilled if we leave this off
[...]

PR #3892, opened 2024-10-01
fix rolling fuel cost average bug

cmgosnell:
# Overview

Closes #[3889](https://github.com/catalyst-cooperative/pudl/issues/3889).

## What did you change?

# Documentation

```[tasklist]
```

# Testing

```[tasklist]
# To-do list
- [x] Alternatively, run the `build-deploy-pudl` GitHub Action manually.
```


src/pudl/helpers.py
```diff
@@ -1242,11 +1242,10 @@ def generate_rolling_avg(
...
     )
     # with the aggregated data, get a rolling average
-    roll = bones.rolling(window=window, center=True, **kwargs).agg({data_col: "mean"})
-    # return the merged
-    return bones.merge(
-        roll, on=group_cols + ["report_date"], suffixes=("", "_rolling")
-    ).reset_index()
+    bones[f"{data_col}_rolling"] = bones.groupby(by=group_cols)[data_col].transform(
+        lambda x: x.rolling(window=window, center=True, **kwargs).mean()
+    )
```
cmgosnell, line 1247: the source of the problem.... the `rolling` was being done after the groupby above but it was not actually using the `group_cols`. (the two example bad plants had lower cost BIT and higher cost DTO and the BIT records were being filled with the higher cost DTO because those records came right after the BIT records. bad bad).

I swapped the `DataFrame.rolling` for a `groupby.DataFrameGroupBy.rolling`.

cmgosnell:
I'm still running tests - I'd especially like to run the validation tests on both the EIA output table and the FERC1 plant tables. but I think the upstream stuff is currently ready for review.

src/pudl/metadata/fields.py
```diff
@@ -1667,9 +1667,17 @@
...
+        "type": "string",
+        "description": (
+            "Indicates what the source of the values in the fuel_cost_per_mmbtu "
+            "column. The fuel cost either comes directly from the EIA forms "
+            "(original), was filled in from the EIA's API using state-level averages "
+            "(eiaapi), was filled in using a rolling average (rolling_avg) or "
+            "When the records get aggregated together and contain multiple "
+            "sources (mixed)."
+        ),
+        "constraints": {"enum": ["original", "eiaapi", "rolling_avg", "mixed"]},
```
cmgosnell, line 1680: we previously had this bool is it from the api column. I swapped that out for this enum source column. Feels more informative overall

src/pudl/helpers.py
```diff
@@ -1285,6 +1284,22 @@ def fillna_w_rolling_avg(
...
+    aggregating records with fuel_cost_per_mmbtu_source.
+    """
+    sources = [source for source in x.tolist() if isinstance(source, str)]
+    if len(sources) > 1:
+        source = "mixed"
+    elif len(sources) == 1:
+        source = sources[0]
+    else:
+        source = pd.NA
+    return source
```
cmgosnell, line 1300: any time we were previously aggregating the `fuel_cost_from_eiaapi`  - which we did many times to make the frc table monthly or annual - we were just using `any`. This custom gb agg function replaces that functionality but instead of saying True for fuel_cost_from_eiaapi if any record came from the api, now it says "mixed" if there are more than one non-null sources.

Unsurprisingly the annual tables are mostly mixed while the monthly aggregation has a pretty even mix of outputs from this.

src/pudl/helpers.py
zaneselvans: Uuuugh. 🤦🏼

src/pudl/helpers.py
zaneselvans: Can we say anything explicit about the proportion of values in the final output? (original, eiaapi, rolling_avg, and still NULL) or anything about how the source varies by time, location, or energy source code?

zaneselvans (approved):
The explicit enumeration of the source for the fuel price is a big improvement!

Nothing blocking here, but it would be nice to see some explicit numbers on the proportion of different `fuel_cost_per_mmbtu_source` values (including NA) broken down by:
- `state`
- `energy_source_code`
- `report_date.dt.year`

to better understand how this filling process is modifying the original data.

This could just be 3 bar charts showing the proportion of each of the 4 values in groupbys of those 3 partitioning variables.

src/pudl/helpers.py
cmgosnell: I just did a quick look at the value_counts of the YS and MS aggregated table.

Monthly:
```python
fuel_cost_per_mmbtu_source
original       97146
mixed          80118
eiaapi         64646
<NA>           28322
rolling_avg     5876
```

Yearly:
```python
fuel_cost_per_mmbtu_source
mixed          23844
<NA>            1865
original         519
eiaapi           419
rolling_avg       62
Name: count, dtype: int64
```

Are you suggesting I add a validation tests about this?

zaneselvans: I was suggesting a way to visually sanity check or better understand the consequences of all of this filling, since it's affecting a large number of records, and it's a modification that we're responsible for.

Maybe once it's merged in I can make a `pudl-examples` notebook that looks at this.

cmgosnell: I was going spot visual spot checking of the two known to be bad plants as well as a handful of others over the in issue. we also have a rn very not working wip notebook which i believe does what you are thinking about... I can probably update it quickly. But if it takes a long time that would feel oos for this fix the bug issue.

zaneselvans: Like I said this is not blocking. If it's any hassle don't worry about it and I can play with it on my own time for my own sanity or as an example notebook under our example notebook documentation budget.

cmgosnell: okay i updated the notebook and will check in shortly. here are some example peeps at some outputs

image

some example plants 
image
image
image
image

aaaand by fuel type
image

PR #3893, opened 2024-10-02
Extract VCE renewable generation profiles and remove deprecated `gsutil` from workflows

### LINKED ISSUE #3880: Write an extractor for the `vpfregen` data

e-belfer:
# Overview

Closes catalyst-cooperative/pudl#3880.

Extracts VCE renewable generation profile data into raw dagster assets. Also fixes a nightly build failure due to the deprecation of `gsutil` (IAM permissions were updated manually and need to be added in terraform, but this is out of scope for this PR).

## What did you change?
- Add a module for `vcerare` extraction and specify valid run settings
- Add `vcerare` to the fast and full ETL runs
- Swap `gsutil` with `gcloud storage`

```[tasklist]
# To-do list
- [x] Swap fake DOI once catalyst-cooperative/pudl-archiver#434 merged
- [x] Extract the tiny lat/lon/fips CSV
- [x] Check in `column_maps` folder or add option to return no error when folder does not exist
- [x] Swap out ancient `gsutil` for `gcloud storage` in `zenodo-cache-sync`
```

# Testing

```[tasklist]
# To-do list
```


src/pudl/extract/vceregen.py
```diff
@@ -0,0 +1,183 @@
...
+            # # Get path to zipfile
+            # zippath = zf.filename
+            # Get list of file names in the zipfile
+            files = zf.namelist()
+            # Get the particular file of interest
+            file = next(
+                (x for x in files if self.source_filename(page, **partition) in x), None
+            )
+            # # Read it in using dask
+            df = pd.read_csv(BytesIO(zf.read(file)), **self.READ_CSV_KWARGS)
```
zaneselvans, line 116: I think there are some tricks you can play with the keyword args to `read_csv()` if you need to reduce memory usage. Like if you can give it a specific and smaller dtype (float32) to cast all the data to on read, rather than doing it after the fact, using `low_memory=True` and giving it a chunksize so that the textual data is read and re-cast incrementally, you might try `engine="pyarrow"` and `dtype_backend="pyarrow"`

e-belfer: @zaneselvans Definitely, I was just trying to get a functional version out asap and I still have a bit of clean up to do. It seems like you're hesitant about using dask instead of pandas - is there a reason why this (and eventually cems for similar reasons) wouldn't be better use cases for dask than pandas?

zaneselvans: Well, it's another layer of complexity and it has a lot of overlap with the problems that Dagster is trying to solve -- building a DAG that defines dependencies, and respecting them when doing the computation in parallel.  AFAIK there isn't a good way within Dagster to manage this kind of nested parallelism, and it's caused us trouble in some assets where we've got multi-threaded linear algebra happening through Numpy.

We could use Dagster's Dask executor rather than doing "local" execution both locally and in Google Batch, and maybe that would gracefully handle both kinds of parallelism, since the Dask cluster would be handling both? But I feel like we tested non-local execution and found that with our large number of smaller assets the overhead of distributed processing actually meant the whole ETL ran significantly slower.

Or are you thinking about using Dask to manage serialization of the CSV processing? That might work well for a fast but memory-intensive process where we don't really need to parallelize it to run faster, but we do need to reduce peak resource utilization.

e-belfer: I was thinking about this mostly through the extraction and transform steps, since we're effectively doing a small number of quick transforms (pivot, convert datetime) on a giant dataset where the parts don't depend much on one another (e.g., hours 10-20 and hours 21-30 of the year will want the same kind of transforms, and dtypes are consistent throughout the dataset). So yes, more to reduce peak memory than to improve speed.

It looks as simple as setting up a client config and passing a [dask_resource](https://docs.dagster.io/deployment/guides/dask) to dagster, but always simpler said than done ofc. I'll take a closer look at peak memory when I finish the archiver and see what things look like at peak.

e-belfer: Decided this is out of scope for this PR.

docker/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apt-get update && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-# Configure gsutil authentication
+# Configure gcloud authentication
```
jdangerx, line 18: Sort of surprised this is needed, it's writing to `boto.cfg` which sounds like it's for AWS. But no need to mess with that now I guess.

src/pudl/extract/extractor.py
```diff
@@ -66,6 +61,16 @@ def _load_csv(self, package: str, filename: str) -> pd.DataFrame:
...
+    def _load_column_maps(self, column_map_pkg: str) -> dict:
+        """Create a dictionary of all column mapping CSVs to use in get_column_map()."""
+        column_dict = {}
+        for res_path in importlib.resources.files(column_map_pkg).iterdir():
+            # res_path is expected to end with ${page}.csv
+            if res_path.suffix == ".csv":
+                column_map = self._load_csv(column_map_pkg, res_path.name)
+                column_dict[res_path.stem] = column_map
+        return column_dict
+
```
aesharpe, line 73: is this just a formality seeing as we don't have any column maps for this dataset?

docker/gcp_pudl_etl.sh
```diff
@@ -298,13 +298,13 @@ if [[ $ETL_SUCCESS == 0 ]]; then
     # If running a tagged release, ensure that outputs can't be accidentally deleted
     # It's not clear that an object lock can be applied in S3 with the AWS CLI
     if [[ "$GITHUB_ACTION_TRIGGER" == "push" && "$BUILD_REF" == v20* ]]; then
-        gsutil -m -u catalyst-cooperative-pudl retention temp set "gs://pudl.catalyst.coop/$BUILD_REF/*" 2>&1 | tee -a "$LOGFILE"
+        gcloud storage --billing-project="catalyst-cooperative-pudl" objects update "gs://pudl.catalyst.coop/$BUILD_REF/*" --temporary-hold 2>&1 | tee -a "$LOGFILE"
```
jdangerx, line 301: nit: seems like we could use `"$GCP_BILLING_PROJECT"` instead of `"catalyst-cooperative-pudl"` here.

jdangerx (approved):
The `gsutil`/`gcloud storage` changes look good to me!

src/pudl/extract/vcerare.py
```diff
@@ -0,0 +1,191 @@
+"""Extract VCE Resource Adequacy Renewable Energy (RARE) generation profile data.
+
+This dataset has 1,000s of columns, so we don't want to manually specify a rename on
+import because we'll pivot these to a column. We adapt the standard extraction
+infrastructure to simply read in the data.
```
aesharpe, line 5: Maybe add "in the PUDL pipeline" after "pivot these to a column"

src/pudl/extract/vcerare.py
```diff
@@ -0,0 +1,191 @@
...
+
+Each annual zip folder contains a folder with three files:
+Wind_Power_140m_Offshore_county.csv
+Wind_Power_100m_Onshore_county.csv
+Fixed_SolarPV_Lat_UPV_county.csv
+
+The drive also contains one more file: vce_county_lat_long_fips_table.csv. This file is
+not partitioned, so we always read it in regardless of the partitions configured for the
+run.
```
aesharpe, line 14: one more CSV file (because there is also the readme)

src/pudl/extract/vcerare.py
```diff
@@ -0,0 +1,191 @@
...
+        df = pd.concat(dfs, sort=True, ignore_index=True)
+
+        return self.process_final_page(df, page)
+
+
+raw_vcerare__all_dfs = raw_df_factory(Extractor, name="vcerare")
+
+
+def raw_vcerare_asset_factory(part: str) -> AssetsDefinition:
+    """An asset factory for VCE hourly renewable generation profiles."""
```
aesharpe, line 157: Same name comment

src/pudl/extract/vcerare.py
```diff
@@ -0,0 +1,191 @@
...
+    asset_kwargs = {
+        "name": f"raw_vcerare__{part}",
+        "required_resource_keys": {"datastore", "dataset_settings"},
+    }
+
+    @asset(**asset_kwargs)
+    def _extract(context, raw_vcerare__all_dfs):
+        """Extract raw GridPath RA Toolkit renewable energy generation profiles.
```
aesharpe, line 165: This seems like an accidental copy-over from the gridpath extractor.

src/pudl/extract/vcerare.py
```diff
@@ -0,0 +1,191 @@
...
+
+VCERARE_PAGES = [
+    "offshore_wind_power_140m",
+    "onshore_wind_power_100m",
+    "fixed_solar_pv_lat_upv",
+]
+
+
+class VCEMetadata(GenericMetadata):
+    """Special metadata class for VCE RARE renewable generation profiles."""
```
aesharpe, line 38: the RE in rare stands for renewable energy so I think we can take out the word renewable here

src/pudl/extract/vcerare.py
```diff
@@ -0,0 +1,191 @@
...
+        """Hard code the page names, which usually are pulled from column rename spreadsheets."""
+        return VCERARE_PAGES
+
+    def get_file_name(self, page, **partition):
+        """Returns file name of given partition and page."""
+        return self._file_name.loc[page, str(self._get_partition_selection(partition))]
+
+
+class Extractor(CsvExtractor):
+    """Extractor for VCE renewable generation profiles."""
```
aesharpe, line 63: Should call it by RARE name here

src/pudl/settings.py
```diff
@@ -401,6 +401,19 @@ class EiaAeoSettings(GenericDatasetSettings):
     years: list[int] = data_source.working_partitions["years"]
 
 
+class VCERARESettings(GenericDatasetSettings):
```
aesharpe, line 404: I think the way we've named the other sources is with Pascal case (VceRareSettings)

aesharpe (commented):
This looks good! a few comments about format etc. 

I have one overarching question about the dataset name. Technically it is called the "VCE RARE Power Dataset" but we refer to it here as the "VCE RARE generation profiles". I'm not sure how important the last few words after VCE RARE are...I like the way we currently have it because "generation profiles" feels more descriptive than "power dataset" but maybe we should just use the official name. I don't feel strongly about this, just wanted to flag.

docker/gcp_pudl_etl.sh
e-belfer: I agree!

src/pudl/extract/extractor.py
e-belfer: This is moving the existing load step into a separate method in the base class, so that i'm able to subclass and change it in the VCE extractor.

src/pudl/extract/vcerare.py
e-belfer: Sure is, good catch!

src/pudl/package_data/vcerare/__init__.py
```diff
@@ -0,0 +1 @@
+"""CSV file extraction maps for VCE RARE generation profiles."""
```
aesharpe, line 1: Just a note for consistency if you want to change this to say "Power Dataset" to match the others.

src/pudl/settings.py
```diff
@@ -571,6 +584,7 @@ class DatasetsSettings(FrozenBaseModel):
     phmsagas: PhmsaGasSettings | None = None
     nrelatb: NrelAtbSettings | None = None
     gridpathratoolkit: GridPathRAToolkitSettings | None = None
+    vcerare: VCERareSettings | None = None
```
aesharpe, line 587: another nit, but I think even the VCE can be lowercased to VceRare... to match the others.

src/pudl/extract/vcerare.py
```diff
@@ -0,0 +1,190 @@
...
+        return Output(value=raw_vcerare__all_dfs[part])
+
+    return _extract
+
+
+raw_vcerare_assets = [raw_vcerare_asset_factory(part) for part in VCERARE_PAGES]
+
+
+@asset(required_resource_keys={"datastore", "dataset_settings"})
+def raw_vcegen__lat_lon_fips(context) -> pd.DataFrame:
```
aesharpe, line 178: I think this should be `raw_vcerare__`

e-belfer: Sure should be, I'll update now

PR #3898, opened 2024-10-04
Transform vceregen renewable generation profiles

### LINKED ISSUE #3874: Clean, transform, and sanity check the `vceregen` data
### Data Structure Description
Under our current system, this data will be published as a single Apache Parquet file (as with all our other hourly tables). Internally it would be partitioned by year or possibly [year, state] into a number of distinct row groups for more efficient querying.

**Primary Key / Index Columns:**
- Hourly UTC Timestamp
- 5-digit county FIPS ID (a string of digits, including leading zeros, for easy integration with geospatial & other data)
- 3 data columns, each containing floating point values between 0.0 & 1.0 (could be float16?)
   - Onshore Wind
   - Offshore Wind
   - Fixed-tilt Solar PV

**Possible additional columns:**
- 2-letter state abbreviation for ergonomic filtering & legibility (categorical string, e.g. “CA”)
- Canonical county name for legibility (categorical string, e.g. “San Bernardino”)
- Approximate county time zone for use in time localization (categorical string, e.g. “America/Los_Angeles”)
- Longitude & Latitude for the center of the county, for spatial queries (low precision floats or GeoPoint)

This structure & format is intended to maximize programmatic usability. It minimizes the need for additional external information or context to make the data usable and allows it to be queried efficiently remotely, and will let us and others build on top of it with minimal additional glue required – e.g. through the interactive dashboard mentioned below. With this structure, 5 years of data would have ~140M rows and the one Parquet file will probably be a few GB on disk.

### Data Transformations
- [x] Add missing leading zeros to provided county FIPS IDs
- [x] Add UTC timestamp column based on row number to be used as part of the index
- [x] Pivot county column headers into a single county column.
- [x] Map county_state name strings to the provided county FIPS IDs
- [x] Divide values by 100 to convert percentages to capacity factors if necessary

### Data Sanity Checks
- [ ] Do the provided FIPS codes and their associated county & state names match current US Census data? In [the sample data](https://docs.google.com/spreadsheets/d/1EjrF88iBrOCN85K7zQbZASm8jAbpBGkv_divjrMbt6I/edit?gid=1719280373#gid=1719280373) the FIPS codes have lost their leading zeros.
- [ ] Do the provided county lon/lat values fall within the expected county boundaries in US Census/TIGER or DP1 data?
- [ ] Plot histograms of capacity factors for the 3 different generator types. Do they look right?
[Sample data](https://docs.google.com/spreadsheets/d/1Q4KIrbMddN0dn3iTheMlmGnahGRFKce9uPfXS95sbW8/edit?gid=388924446#gid=388924446) seems to be in percentages (0.0-100.0) rather than fractions (0.0-1.0).
- [ ] Do the capacity factors vary diurnally and seasonally as expected?

aesharpe:
# Overview

Closes #3874

Transform the `vceregen` tables!

## What did you change?

Add and populate a transform module for the `vceregen` data source. The output is a single table that combines the three capacity factor tables with the FIPS table. I will highlight any other lingering questions in the comments.

```[tasklist]
# TO-DO: Other transforms
- [x] divide capacity factor values by 100 based on [this](https://github.com/catalyst-cooperative/pudl/issues/3874#issuecomment-2394499633) thread.
- [x] figure out how to handle capacity factor values > 100.
- [x] Decide whether to have year_hour values start at 0 or 1
- [x] I noticed that not all of the counties in the capacity factor tables show up in the FIPS table. Look into this more.
- [x] Add more sanity checks.
- [x] Add asset checks.
- [x] finalize dataset name
- [x] Decide whether to revert back to 0-100 scale for capacity factor
```

```[tasklist]
# TO-DO: Once we agree on the transforms
- [x] add `RESOURCE_METADATA`
- [x] add any new fields to `fields.py`
```

# Documentation

```[tasklist]
# TO-DO: Documentation
- [x] Update relevant Data Source jinja templates (see `docs/data_sources/templates`) -- there's a separate branch for this.
```

# Testing

```[tasklist]
# TO-DO: Testing
```


src/pudl/transform/vceregen.py
```diff
@@ -0,0 +1,206 @@
...
+    to_leap_year_dates = pd.date_range(start="2019-01-01", end="2020-12-31", freq="h")[
+        :-1
+    ]
+    post_leap_year_dates = pd.date_range(
+        start="2021-01-01", end="2024-01-01", freq="h"
+    )[:-1]
+    # Add date sequence to pre and post leap year dates
+    df.loc[:, "report_year"] = df.report_year.astype("Int32")
+    df.loc[df["report_year"] < 2021, "hour"] = to_leap_year_dates
+    df.loc[df["report_year"] > 2020, "hour"] = post_leap_year_dates
```
aesharpe, line 93: Highlighting this chunk of code because it's important to note! The leap year value that gets dropped is December 31st, 2020, not 2023. I hope this won't be an issue with datetime continuity, but it might @zaneselvans.

src/pudl/transform/vceregen.py
```diff
@@ -0,0 +1,206 @@
...
+    This function adds a datetime column and a year_hour column.
+    The datetime column is important for merging the data with other
+    data, and the year_hour 0-8769 column is important for modeling
+    purposes. The report_year column is also helpful for filtering,
+    so we keep all three!
+
+    For leap years (2020), December 31st is excluded.
+
+    TO-DO: decide whether to have the year_hour start at 0 or 1.
+    and update the date_range accordingly.
```
aesharpe, line 74: Flagging this decision

aesharpe (commented):
Hello! Here is the basis for `vceregen` transforms. And here's a little bit of my reasoning. 

There are four tables, but I decided to combine them all into one sudo-output table because it's so big. I could have saved the FIPS table as it's own asset, but that didn't feel particularly useful because we already have a table like that and this one is so hyper-specific to this table.

To the best of my ability, these transforms are designed to preserve memory. If there are any further memory-saving opportunities that you see, let me know! 

There are plenty of other little defensive checks I could add it--let me know if you see an opportunity to do so! 

I tried to minimize function nesting wherever possible so you can follow what's happening to the tables clearly in `out_vceregen__hourly_available_capacity_factor()`

src/pudl/transform/vceregen.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    logger.info(
+        "Preping Lat-Long-FIPS table for merging with the capacity factor tables"
+    )
+    # Get state name patterns from the POLITICAL_SUBDIVISIONS table
+    ps_usa_df = POLITICAL_SUBDIVISIONS[POLITICAL_SUBDIVISIONS["country_code"] == "USA"]
+    state_names = cleanstrings_snake(
+        ps_usa_df, ["subdivision_name"]
+    ).subdivision_name.tolist()
+    state_pattern = "|".join(state_names)
```
aesharpe, line 32: We could also just literally use the state and county _values_ in the `POLITICAL_SUBDIVISIONS` df....but I guess I felt like I should use the ones in this table? If that feels silly then lmk.

src/pudl/transform/vceregen.py
```diff
@@ -0,0 +1,206 @@
...
+        start="2021-01-01", end="2024-01-01", freq="h"
+    )[:-1]
+    # Add date sequence to pre and post leap year dates
+    df.loc[:, "report_year"] = df.report_year.astype("Int32")
+    df.loc[df["report_year"] < 2021, "hour"] = to_leap_year_dates
+    df.loc[df["report_year"] > 2020, "hour"] = post_leap_year_dates
+    # Add a year_date column for modeling purposes.
+    # Right now each year should go from 0-8759. We
+    # can decide to make that 1-8760 if we want to!
+    df.loc[:, "year_hour"] = df.hour.dt.hour + (df.hour.dt.dayofyear - 1) * 24
```
aesharpe, line 97: do we want this to reset every year (i.e.: 0-8759) or do we want it to go strait through the years (i.e.: 0-43800)

zaneselvans: Given that the files already come with an hour index that starts with 1, I think we should start with 1.

The hour-of-year should reset every year.

src/pudl/transform/vceregen.py
```diff
@@ -0,0 +1,206 @@
...
+    return df_stacked
+
+
+def _combine_all_cap_fac_dfs(cap_fac_dict: dict[str, pd.DataFrame]) -> pd.DataFrame:
+    """Combine capacity factor tables."""
+    logger.info("Merging all the capacity factor tables into one")
+    merge_keys = ["report_year", "hour", "year_hour", "county_state_names"]
+    # Can't merge all at once using reduce because that's too much memory
+    # So we'll merge one at a time.
+    mega_df = pd.merge(
```
zaneselvans, line 129: If you set the merge columns as the index you can use [`pd.concat()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.concat.html) to combine them, which I think should be less resource intensive than `pd.merge()` especially if they're all sorted by index first.

src/pudl/transform/vceregen.py
zaneselvans: The data comes with an hour-of-year starting at 1, so I think we should probably stick with that convention.

src/pudl/transform/vceregen.py
zaneselvans: Rather than treating the normal and leap years separately, you could define the date range using `periods=8760` rather than setting the `end` which would simplify the logic here, and work for any past or future year of 8760 data as well.

zaneselvans: It might look like:
```python
all_years = [2020, 2021]
datetime8760_index = pd.DatetimeIndex(
    pd.concat(
        [
            pd.Series(
                pd.date_range(start=f"{year}-01-01", periods=8760, freq="h")
            ) for year in all_years
        ]
    )
)
```

src/pudl/transform/vceregen.py
zaneselvans: Given that the data comes with the hour-of-year we should make sure that we aren't altering it and that the datetime we've created is consistent with it. Either by constructing the datetime from it, or by checking that the provided hour-of-year matches what we would derive from our datetime.

When I suggested creating a datetime column and deriving the hour-of-year from it I thought that we had no indication of the time other than the filename (for the year) and the row number (for the hour).

zaneselvans: `hour_of_year` feels like it's more consistent with the `day_of_year` and `day_of_week` naming that pandas uses for counting subdivisions.

src/pudl/transform/vceregen.py
zaneselvans: The political subdivisions only has state level information, so we can't get the county names there.

But given the FIPS county code, we can look up the county name and state in the Census DP1 data, or (probably?) using `addfips` though there's a chance that a handful of the values would be different or some of the FIPS codes wouldn't show up, since they do evolve slightly over time. You might try merging in the county names and states from DP1 and see if they all match up as expected.

The only thing we really need the `County_State` strings for is associating the FIPS code with each column of the original data. If we have the FIPS code, that's the authoritative ID, and the county name is really just for human readability. The state will be useful for geographic filtering. We use the 2-letter state abbreviation for state columns elsewhere and they're easier to work with than the long state names. And those you should definitely be able to look up in the political subdivision dataframe given the first 2-digits of the FIPS code.

src/pudl/transform/vceregen.py
aesharpe: oh cool, this is interesting to know that `periods` works like that.

e-belfer:
Some asset checks to consider:

- The difference between the datetime index in our df and one that runs from the start to the end date of the selected years in the run is only missing the known leap year timestamps (see, e.g. https://stackoverflow.com/questions/52044348/check-for-any-missing-dates-in-the-index)
- No capacity factors <0

src/pudl/transform/vceregen.py
aesharpe: I can't decide what to do about the `county` name column. Should I keep the county name as extracted from this VCE FIPS table or should I use the county name from the census dp1 table? I think @e-belfer mentioned that the census table was older and possibly less accurate than their provided FIPS/county combos.

src/pudl/metadata/fields.py
```diff
@@ -1970,6 +1982,15 @@
...
     },
+    "hour_of_year": {
+        "type": "number",
+        "description": "Number between 1 and 8670 representing the hour in a given year.",
+    },
+    "hour_utc": {
+        "type": "datetime",
+        "description": "Hourly datetime in UTC.",
+        "unit": "UTC",
+    },
```
aesharpe, line 1993: There's a similar field for CEMS, but it's called `operating_hour_utc` which felt a little different.

src/pudl/metadata/resources/vceregen.py
```diff
@@ -0,0 +1,37 @@
...
+                "county_id_fips",
+                "county",
+                "state",
+                "latitude",
+                "longitude",
+                "capacity_factor_solar_pv",
+                "capacity_factor_onshore_wind",
+                "capacity_factor_offshore_wind",
+            ],
+            "primary_key": ["hour_utc", "county_id_fips", "county"],
```
aesharpe, line 30: This is kind of weird and annoying, but the FIPS table provided by VCE has multiple county names per FIPS code. I'm not exactly sure why? Maybe from multiple years? This makes it so that hour and county_id alone aren't unique identifiers for the data.

src/pudl/transform/vceregen.py
aesharpe: I'm inclined to just use the county name from the VCE fips table (instead of the dp1 table) because of how the VCE table actually has multiple county names per FIPS code. As I mention above in relation to the primary keys, I'm not sure why, but that's all to say that mapping to the dp1 table by FIPS code would be inaccurate.

src/pudl/transform/vceregen.py
```diff
@@ -0,0 +1,257 @@
...
+            .pipe(_make_cap_fac_frac, df_name)
+            .pipe(_add_time_cols)
+            .pipe(_stack_cap_fac_df, df_name)
+        )
+    # Make sure there's no funny business
+    if len(df_dict) != 3:
+        raise AssertionError("Incorrect number of dataframes in the df_dict")
+    # Combine the data!
+    out_df = _combine_all_cap_fac_dfs(df_dict).pipe(
+        _combine_cap_fac_with_fips_df, fips_df
```
aesharpe, line 255: Tempted to try combining each of the cap fac tables with fips in the for loop above instead of here to see if that would save any computing power.

aesharpe (commented):
Ready for another review @e-belfer

src/pudl/transform/vceregen.py
```diff
@@ -0,0 +1,257 @@
...
+
+def _make_cap_fac_frac(df: pd.DataFrame, df_name: str) -> pd.DataFrame:
+    """Make the capacity factor column a fraction instead of a percentage.
+
+    This step happens before the table gets stacked to save memory.
+    """
+    logger.info("Converting capacity factor into a fraction.")
+    county_cols = [x for x in df.columns if x not in ["report_year", "unnamed_0"]]
+    df[county_cols] = df[county_cols] / 100
+    if (max_val := df[county_cols].max().max()) > 1.02:
```
aesharpe, line 142: I'm not exactly sure why, but there are some capacity factor values in the solar table that are ~102. This isn't so far over 100 that I'm concerned, but I'm curious. Is there any explanation for this?

src/pudl/metadata/fields.py
```diff
@@ -466,6 +466,18 @@
...
+        "type": "number",
+        "description": "Fraction of potential generation calculated for offshore wind.",
+    },
+    "capacity_factor_onshore_wind": {
+        "type": "number",
+        "description": "Fraction of potential generation calculated for onshore wind.",
+    },
+    "capacity_factor_solar_pv": {
+        "type": "number",
+        "description": "Fraction of potential generation calculated for solar pv.",
```
e-belfer, line 479:
```suggestion
        "description": "Fraction of potential generation calculated for fixed-tilt solar PV installations.",
```

src/pudl/metadata/fields.py
```diff
@@ -1970,6 +1982,15 @@
         "description": "The energy contained in fuel burned, measured in million BTU.",
         "unit": "MMBtu",
     },
+    "hour_of_year": {
+        "type": "number",
+        "description": "Number between 1 and 8670 representing the hour in a given year.",
+    },
+    "hour_utc": {
+        "type": "datetime",
+        "description": "Hourly datetime in UTC.",
```
e-belfer, line 1991: What do you mean by "hourly datetime" and "hour_utc"? If this is just a timestamp (2021-01-01 01:00:00), as compared to a date (2021-01-01), then "datetime_utc" column that already exists should work just fine instead of creating a new one.

src/pudl/metadata/fields.py
```diff
@@ -1970,6 +1982,15 @@
...
+        "unit": "UTC",
```
e-belfer, line 1992: I don't think "UTC" makes sense as a unit, as the unit here is a measure of time not a timezone.

src/pudl/metadata/fields.py
```diff
@@ -466,6 +466,18 @@
         "type": "number",
         "description": "Fraction of potential generation that was actually reported for a plant part.",
     },
+    "capacity_factor_offshore_wind": {
+        "type": "number",
+        "description": "Fraction of potential generation calculated for offshore wind.",
+    },
+    "capacity_factor_onshore_wind": {
+        "type": "number",
+        "description": "Fraction of potential generation calculated for onshore wind.",
```
e-belfer, line 475: Do you want to specify about the 100m here?

e-belfer (commented):
Will review thoroughly and run the code when asset checks are complete, but here's a first pass of feedback!

src/pudl/transform/vceregen.py
```diff
@@ -0,0 +1,206 @@
...
+) -> pd.DataFrame:
+    """Transform raw Vibrant Clean Energy renewable generation profiles.
+
+    Concatenates the solar and wind capacity factors into a single table and turns
+    the columns for each county into a single county column.
+
+    Note that this transform is a bit unusual, in that it is producing a highly
+    processed output table. That's because we're working backwards from an archived
+    finished product to be able to provide a minimum viable product. Our intent is to
+    integrate or reimplement the steps required to produce this output table from
```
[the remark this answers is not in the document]
zaneselvans: This was true for the GridPath RA Toolkit description and doesn't apply here.

src/pudl/metadata/fields.py
zaneselvans: Might also specify that the tilt is equal to the latitude.

src/pudl/metadata/fields.py
```diff
@@ -466,6 +466,18 @@
         "type": "number",
         "description": "Fraction of potential generation that was actually reported for a plant part.",
     },
+    "capacity_factor_offshore_wind": {
+        "type": "number",
+        "description": "Fraction of potential generation calculated for offshore wind.",
```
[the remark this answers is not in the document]
zaneselvans: I agree having the hub heights in these descriptions (and possibly even the column names?) would be helpful for legibility of the data, and for potentially differentiating between future generation profiles with different technology specs.

src/pudl/metadata/fields.py
zaneselvans: IIRC the EPA CEMS "operating hour" was an integer hour-of-day, and it's only present in the raw input and not the outputs. We do have an `operating_datetime_utc` in CEMS, but that's a relic we should really rename to be consistent with all of the other hourly data where we're using `datetime_utc`

src/pudl/metadata/resources/vceregen.py
zaneselvans: Hmm. This seems like something we might want to ask Chris about, since it suggests that the "counties" might not actually be counties as described in the README. If it really is the case that there are multiple different locations with data being reported that are part of the same FIPS code, then maybe we have no choice but to use the county as the primary key, and NOT the FIPS code, since it then becomes dependent on the county/place name. But this will mess with any attempt to connect the data to geospatial information / the county geometries.

src/pudl/transform/vceregen.py
```diff
@@ -0,0 +1,286 @@
...
+    merge_keys = ["report_year", "hour_utc", "hour_of_year", "county_state_names"]
+    for name, df in cap_fac_dict.items():
+        cap_fac_dict[name] = df.set_index(merge_keys)
+    mega_df = pd.concat(
+        [
+            cap_fac_dict["solar_pv"],
+            cap_fac_dict["offshore_wind"],
+            cap_fac_dict["onshore_wind"],
+        ],
+        axis=1,
```
zaneselvans, line 169: We've been trying to use the index/columns rather than 0/1 syntax for functions like concat since they're a bit less mysterious.

src/pudl/transform/vceregen.py
aesharpe: That makes sense!

src/pudl/metadata/resources/vceregen.py
aesharpe: Yeah, it's tricky. I think for now I'm going to make this `county` column a `region` or something similar column. We need either `county_id_fips` or `state` as part of the primary key to differentiate between counties of the same name.

src/pudl/transform/vceregen.py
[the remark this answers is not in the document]
aesharpe: Thanks for flagging, will remove!

src/pudl/metadata/fields.py
aesharpe: good point!

src/pudl/metadata/resources/vceregen.py
zaneselvans: If these non-county things are lakes or city centers like Ella suggested then I'm curious if they're really supposed to be in there, or what geometry they're supposed to be linked to -- all of these capacity values are averages of the gridded model outputs from some area.

src/pudl/transform/vceregen.py
aesharpe: Can you give me an example? I'm not sure what you mean by this @zaneselvans

src/pudl/metadata/resources/vceregen.py
aesharpe: We should ask Chris, I'll get a list of them.

aesharpe:
Here are the subregions in question: 

**Lakes**

* Lake Michigan (four slightly different coordinates for the places it touches different states)
* Lake Superior (three different coordinates for different states)
* Lake Hurron
* Lake St. Clair
* Lake Ontario
* Lake Erie

**Cities**

* Clifton Forge City, VA
* Bedford City, VA

|      |   county_id_fips | county_or_subregion   | state   |   latitude |   longitude |
|-----:|-----------------:|:----------------------|:--------|-----------:|------------:|
|  624 |            17031 | cook                  | IL      |    41.8707 |    -87.7498 |
| 1545 |            17031 | lake_michigan         | IL      |    42.1328 |    -87.4945 |
| 1543 |            18089 | lake                  | IN      |    41.4351 |    -87.3776 |
| 1546 |            18089 | lake_michigan         | IN      |    41.865  |    -87.3295 |
| 1479 |            26083 | keweenaw              | MI      |    47.3549 |    -88.2192 |
| 1557 |            26083 | lake_superior         | MI      |    47.2335 |    -88.2031 |
| 1547 |            26089 | lake_michigan         | MI      |    45.3633 |    -86.0358 |
| 1623 |            26089 | leelanau              | MI      |    44.9188 |    -85.7801 |
| 1541 |            26141 | lake_hurron           | MI      |    45.7221 |    -83.7262 |
| 2320 |            26141 | presque_isle          | MI      |    45.3412 |    -83.9243 |
| 1556 |            26147 | lake_st_clair         | MI      |    42.5959 |    -82.6699 |
| 2475 |            26147 | saint_clair           | MI      |    42.955  |    -82.6584 |
|  625 |            27031 | cook                  | MN      |    47.8447 |    -90.4528 |
| 1558 |            27031 | lake_superior         | MN      |    47.3559 |    -90.8572 |
| 1403 |            36045 | jefferson             | NY      |    44.0287 |    -75.9595 |
| 1553 |            36045 | lake_ontario          | NY      |    44.0983 |    -76.347  |
|  874 |            39043 | erie                  | OH      |    41.4035 |    -82.5877 |
| 1539 |            39043 | lake_erie             | OH      |    41.6007 |    -82.6119 |
|   42 |            51005 | alleghany             | VA      |    37.7895 |    -80.0076 |
|  563 |            51005 | clifton_forge_city    | VA      |    37.8157 |    -79.8172 |
|  161 |            51019 | bedford_city          | VA      |    37.3197 |    -79.5385 |
|  164 |            51019 | bedford               | VA      |    37.3153 |    -79.5241 |
|  148 |            55007 | bayfield              | WI      |    46.5312 |    -91.1895 |
| 1559 |            55007 | lake_superior         | WI      |    46.9673 |    -90.8044 |
|  796 |            55029 | door                  | WI      |    44.7692 |    -87.5016 |
| 1548 |            55029 | lake_michigan         | WI      |    45.0588 |    -87.1159 |

src/pudl/transform/vceregen.py
aesharpe: ah, yes this is way clearer! Thank you.

zaneselvans:
A couple of weird things that appear in the non-county places:
* It looks like Lake Huron is misspelled (in the original data!) with an extra R ("Hurron").
* For "Saint Claire" and "Lake St. Claire" it seems odd that one would be abbreviated but the other not (also, coming from the original CSV).

My main motivation for wanting to ensure that there are FIPS codes included with the records is that way they can be tied back to a well defined geometry over which the modeled generation was averaged (and a bunch of other data that's collected in associated with that physical area). In the case of the lakes here -- and probably also the offshore regions on the Pacific, Atlantic, and Gulf coasts too -- it's not obvious to me what that area is and it's not documented in the README or the dataset overview. For the Great Lakes is there some well-defined state territorial water? Does "Lake Ontario, NY" refer to the entire portion of Lake Ontario that falls under NY State jurisdiction? If so, why is being associated with a particular county's FIPS code? There are many counties that border Lake Ontario in NY.

Similarly for offshore wind production on the coasts, what offshore area is responsible for the offshore generation associated with say, Santa Cruz, California? How far offshore does it go? What divides that area from the counties to the north and south? Is it a small area that might be biased by only having a couple of model grid points in it? Or is it a huge area that will be representative on average, but not very specific? I don't know for sure but I would guess that offshore wind capacity factors vary systematically somehow with distance from shore, so knowing how far offshore the modeled area extends might inform how someone interprets these capacity factors.

Are the lat/lon points the centroid of the aggregated area? Or a simple mean of all the lat/lon values for the model grid points within the area? Or something else?

aesharpe:
Updates from Chris: 

> Good catch here. So keep as the 0-100 (covert if you wish, but better to leave). The over 100 is a feature (should be only solar) due to cooler temperature on panels.

@zaneselvans do you have specific feelings about whether we should keep convert 0-1 or revert back to 0-100? 

> This leads to another point. We should note in the documentation that the PV is DC values and users should convert to AC using a DC-AC ratio (current average in US is 1.34 according to NREL). To convert you multiply by 1.34 (DC-AC ratio) and any value over 100 you then clip.

@zaneselvans do you think we should convert to AC in the PUDL table? What is our standard for other solar data? It seems like we generally convert DC to AC, but we could have both or just provide the multiplication factor as suggested.

zaneselvans:
I figured he was referring to the raw data.  I think we should leave the raw data 0-100 (and edit the README to reflect that, including a note about the >100% capacities due to colder than standard capacity measurement conditions) and in our processed data we make it 0.0-1.0, both for consistency with the analogous GridPath data, and because all of these values are going to be used as multipliers to scale nameplate capacities -- they aren't directly useful as percentages.

On AC vs. DC capacity, I think we just want to document that it refers to the DC capacity and note in the README that the current average is 1.34 and/or provide a pointer to the NREL data. The current average might be 1.34, but the inverter is really project-by-project decision so if someone wanted to model a project with a different DC/AC capacity ratio, or a change in the ratio over time I think that should be left up to the user.

Also world-famous Solar Opinion Haver Jenny Chase [recommends DC capacity](https://mastodon.green/@solar_chase/113215017127217964) as the best single number to report.

On the FIPS / lakes / non-county locations, it seems like we might want to null out the FIPS codes for those entries that are not actually counties, and use the datetime, state and placename as the primary key, since not all records actually have a FIPS county geometry or other relevant Census data that they correspond to.

src/pudl/metadata/resources/vceregen.py
aesharpe: Unclear, but it seems like Chis did know they were there. He said we could average the counties and cities so I did that. Left the lakes in.

src/pudl/transform/vceregen.py
aesharpe: Chis said this is due to cooler panel temperatures.

aesharpe:
> it seems like we might want to null out the FIPS codes for those entries that are not actually counties,

I agree with everything you said--FIPS should not be a primary key--but I think we should keep the county FIPS codes (not null them).

zaneselvans:
Can you say more about why you don't think the FIPS codes should be null for the records that don't correspond to geometries associated with any FIPS code?

To me it seems like this will cause confusion and errors if anyone tries to do analysis based on those associations. E.g. if you want to estimate how many mode grid points are being averaged to get the generation profile, it'll be valid for counties, but wrong for non-counties. Or if you wanted to make an animated map that colors counties by capacity factor to watch them evolve, there would be two profiles mapped to the same territory, but only one of them would be associated with that territory.

From Chris' email it sounds like the geometries associated with the non-counties can't be identified by a FIPS code, and aren't going to be made available in any form, so NA seems like the most accurate value available.

Also I requested the FIPS codes from Chris to save us the trouble of matching the county names with addfips, so they're really kind of an add-on / not something they were originally planning to share, so I'm not totally surprised they're a little flaky.

@e-belfer what do you think is the right answer here?

aesharpe:
> Can you say more about why you don't think the FIPS codes should be null for the records that don't correspond to geometries associated with any FIPS code?

I was thinking more for filtering purposes. If someone wanted to filter for regions pertinent to a specific county they would miss the lake regions (and therefore offshore wind) if they didn't have FIPs codes associated with them. 

I don't have super strong feelings about this though. Your points are also valid.

zaneselvans:
I think the latitude and longitude coordinates can still give folks the ability to filter spatially pretty easily without the same kind of confusion about the county geometries.

docs/templates/vcerare_child.rst.jinja
```diff
@@ -0,0 +1,60 @@
...
+The data include some sub-county regions such as lakes and cities meaning that the
+county_id_fips and datetime_utc fields are many:1 rather than 1:1. We've averaged out
+the cities but kept the lakes in the data. Note that lakes bordering multiple states
+will appear more than once in the data. VCE used a nearest neighbor technique to assign
+the state waters to the counties (this pertains to coastal areas as well).
+
+Capacity factors > 1
+--------------------
+There are a couple of capacity factor values for the solar pv data that exceed
+the maximum value of 1 for capacity factor. This is due to power production performance
```
e-belfer, line 51: Think it's fine to leave this as 1 because we convert from 100->1, but may be worth  mentioning somewhere that we do this conversion.

src/pudl/metadata/fields.py
```diff
@@ -466,6 +466,56 @@
...
     },
+    "capacity_factor_offshore_wind": {
+        "type": "number",
+        "description": (
+            "Estimated capacity factor (0-1) calculated for offshore wind "
+            "assuming a 140m hub height and 120m rotor diameter."
+            "Based on outputs from the NOAA HRRR operational numerical "
+            "weather prediction model. Capacity factors are normalized "
+            "to unity for maximal power output. "
+            "vertical slices of the atmosphere are considered across the "
```
e-belfer, line 477:
```suggestion
            "Vertical slices of the atmosphere are considered across the "
```

src/pudl/metadata/fields.py
```diff
@@ -466,6 +466,56 @@
...
+            "defined rotor swept area. Bringing together wind speed, density, "
+            "temperature and icing information, a power capacity is estimated "
+            "using a representative power coeffienct (Cp) curve to determine "
```
e-belfer, line 480:
```suggestion
            "using a representative power coefficient (Cp) curve to determine "
```

src/pudl/metadata/fields.py
```diff
@@ -466,6 +466,56 @@
...
+    },
+    "capacity_factor_onshore_wind": {
+        "type": "number",
+        "description": (
+            "Estimated capacity factor (0-1) calculated for onshore wind "
+            "assuming a 100m hub height and 120m rotor diameter."
+            "Based on outputs from the NOAA HRRR operational numerical "
+            "weather prediction model. Capacity factors are normalized "
+            "to unity for maximal power output. "
+            "vertical slices of the atmosphere are considered across the "
```
e-belfer, line 493:
```suggestion
            "Vertical slices of the atmosphere are considered across the "
```

src/pudl/metadata/fields.py
```diff
@@ -466,6 +466,56 @@
...
+            "defined rotor swept area. Bringing together wind speed, density, "
+            "temperature and icing information, a power capacity is estimated "
+            "using a representative power coeffienct (Cp) curve to determine "
```
e-belfer, line 496:
```suggestion
            "using a representative power coefficient (Cp) curve to determine "
```

src/pudl/metadata/fields.py
```diff
@@ -466,6 +466,56 @@
...
+            "the power from a given wind speed, atmospheric density and "
+            "temperature. There is no wake modeling included in the dataset."
+        ),
+    },
+    "capacity_factor_solar_pv": {
+        "type": "number",
+        "description": (
+            "Estimated capacity factor (0-1) calculated for solar pv "
```
e-belfer, line 504:
```suggestion
            "Estimated capacity factor (0-1) calculated for solar PV "
```

src/pudl/metadata/resources/vcerare.py
```diff
@@ -0,0 +1,62 @@
...
+
+from typing import Any
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "out_vcerare__hourly_available_capacity_factor": {
+        "description": (
+            "The data in this table were produced by Vibrant Clean Energy, and are "
+            "licensed to the public under the Creative Commons Attribution 4.0 International "
+            "license (CC-BY-4.0). The table consists of estimated county-averaged hourly "
+            "capacity factor for wind and solar generating facilities across the contiguous "
```
e-belfer, line 11:
```suggestion
            "capacity factors for wind and solar generating facilities across the contiguous "
```

src/pudl/metadata/resources/vcerare.py
```diff
@@ -0,0 +1,62 @@
...
+            "United States [US] to be used as a tool and input for resource adequacy modeling "
```
e-belfer, line 12:
```suggestion
            "United States (US) to be used as a tool and input for resource adequacy modeling "
```

src/pudl/metadata/resources/vcerare.py
```diff
@@ -0,0 +1,62 @@
...
+            "algorithms. This forecast hour is chosen to trade-off the impact of the measurement "
+            "and data assimilation procedure of the HRRR with the physics of the model to derive "
+            "the most complete picture of the atmosphere at the forecast time horizon. "
+            "Hourly capacity factors are spatially averaged across each county over the contiguous "
+            "USA. There are a handful of counties that are too small to pick up representation on "
+            "the HRRR operational forecast grid. As such, these counties will have no wind or solar "
+            "power production curves.\n\n"
+            "For wind capacity factors: vertical slices of the atmosphere are considered across "
+            "the defined rotor swept area. Bringing together wind speed, density, temperature and "
+            "icing information, a power capacity is estimated using a representative power coeffienct "
```
e-belfer, line 33:
```suggestion
            "icing information, a power capacity is estimated using a representative power coefficient "
```

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    logger.info(
+        "Preping Lat-Long-FIPS table for merging with the capacity factor tables"
+    )
+    ps_usa_df = POLITICAL_SUBDIVISIONS[POLITICAL_SUBDIVISIONS["country_code"] == "USA"]
+    state_names = cleanstrings_snake(
+        ps_usa_df, ["subdivision_name"]
+    ).subdivision_name.tolist()
+    state_pattern = "|".join(state_names)
+    # Transform lat long fips table by making the county_state_names lowercase
```
e-belfer, line 38: Non-blocking: It'd be more helpful to split these docstrings and add each on the line where the change is being made, since you're chaining many steps here.

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+        .assign(
+            county_state_names=lambda x: x.county_state_names.str.lower().replace(
+                {r"\.": "", "-": "_"}, regex=True
+            )
+        )
+        .assign(county_id_fips=lambda x: zero_pad_numeric_string(x.fips, 5))
+        .assign(state_id_fips=lambda x: x.county_id_fips.str.extract(r"(\d{2})"))
+        .assign(
+            county_or_lake_name=lambda x: x.county_state_names.str.extract(
+                rf"(.+)_({state_pattern})$"
```
e-belfer, line 52: Is there any reason not to more narrowly capture [a-z]+ here, instead of .+?

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+
+    Prep entails making sure the formatting and column names match those in the
+    capacity factor tables, adding 0s to the beginning of FIPS codes with 4 values,
+    and making separate county/subregion and state columns.
+
+    The county portion of the county_state column does not map directly to FIPS ID.
+    Some of the county names are actually subregions like cities or lakes. For this
+    reason we've named the column county_or_lake_name and it should be considered
+    part of the primary key. There are several instances of multiple subregions that
+    map to a single county_id_fips value.
```
e-belfer, line 28: Would be helpful to note that we're also adding state shortcodes from the Census data (via POLITICAL_SUBDIVISIONS) in this docstring.

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+        }
+    )
+    df = pd.concat(
+        [df.reset_index(drop=True), new_time_col.reset_index(drop=True)], axis="columns"
+    ).rename(columns={"unnamed_0": "hour_of_year"})
+    return df
+
+
+def _stack_cap_fac_df(df: pd.DataFrame, df_name: str) -> pd.DataFrame:
+    """Funciton to transform each capacity factor table individually to save memory.
```
e-belfer, line 109:
```suggestion
    """Function to transform each capacity factor table individually to save memory.
```

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+
+    The main transforms are turning count/subregion columns into county/subregion rows
```
e-belfer, line 111:
```suggestion
    The main transforms are turning county/subregion columns into county/subregion rows
```

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+    return df_stacked
+
+
+def _make_cap_fac_frac(df: pd.DataFrame, df_name: str) -> pd.DataFrame:
+    """Make the capacity factor column a fraction instead of a percentage.
+
+    This step happens before the table gets stacked to save memory.
+    """
+    logger.info(f"Converting capacity factor into a fraction for {df_name} table.")
+    county_cols = [x for x in df.columns if x not in ["report_year", "unnamed_0"]]
```
e-belfer, line 137: Non-blocking: It might be clearer to rename `unnamed_0` before you do any of these steps, and then reference it directly so it's more obvious why it's getting excluded here.

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+    return combined_df
+
+
+def _combine_city_county_records(df: pd.DataFrame) -> pd.DataFrame:
+    """Average cap fac values for city and county values with the same FIPS ID.
+
+    There are several duplicate FIPS code values in the data. Most of them
+    are lakes within a county, but some of them are cities. Only one of the
+    cities, clifton forge city, has values, so we average those values
+    with that of the county. The other city, XXX, has no values so we
```
e-belfer, line 199: Is the city called XXX or is this a leftover placeholder?

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+    combined_df = pd.merge(cap_fac_df, fips_df, on="county_state_names", how="left")
+    return combined_df
+
+
+def _combine_city_county_records(df: pd.DataFrame) -> pd.DataFrame:
+    """Average cap fac values for city and county values with the same FIPS ID.
+
+    There are several duplicate FIPS code values in the data. Most of them
+    are lakes within a county, but some of them are cities. Only one of the
+    cities, clifton forge city, has values, so we average those values
```
e-belfer, line 198:
```suggestion
    cities, Clifton Forge City, has values, so we average those values
```

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+    bad_county_mask = df["county_state_names"].isin(
+        ["alleghany_virginia", "clifton_forge_city_virginia"]
+    )
+    # Take the subset of the data that's relevant to the changes and sort the values
+    # by county_state so that alleghany is first and those are the values that are
+    # preserved by the groupby "first" below.
+    subset_df = df[bad_county_mask].sort_values(by=["county_state_names"])
+    cap_fac_cols = [x for x in subset_df.columns if "capacity_factor" in x]
+    other_cols = [
+        x for x in subset_df.columns if x not in cap_fac_cols + ["datetime_utc"]
```
e-belfer, line 212: Is `datetime_utc` not in `subset_df.columns`? Just curious why this is necessary.

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+    """Combine the combined capacity factor df with the fips df."""
+    logger.info(
+        "Merging the combined capacity factor table with the Lat-Long-FIPS table"
+    )
+    combined_df = pd.merge(cap_fac_df, fips_df, on="county_state_names", how="left")
+    return combined_df
+
+
+def _combine_city_county_records(df: pd.DataFrame) -> pd.DataFrame:
+    """Average cap fac values for city and county values with the same FIPS ID.
```
e-belfer, line 194: Personally think it's almost neater to just drop the city value, but you've gone through the work of averaging it so I think we can just keep it. It would be helpful to add an assertion that ensures that this is the only non-lake case with a duplicated county and values.

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+    return out_df[~bad_county_mask_2]
+
+
+@asset(
+    io_manager_key="parquet_io_manager",
+    compute_kind="pandas",
+    op_tags={"memory-use": "high"},
+)
+def out_vcerare__hourly_available_capacity_factor(
+    raw_vcegen__lat_lon_fips: pd.DataFrame,
```
e-belfer, line 239:
```suggestion
    raw_vcerare__lat_lon_fips: pd.DataFrame,
```
This should get updated, right?

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+    """Transform raw Vibrant Clean Energy renewable generation profiles.
+
+    Concatenates the solar and wind capacity factors into a single table and turns
+    the columns for each county or subregion into a single county_or_lake_name column.
+    """
+    logger.info("Transforming the hourly available capacity factor tables")
+    # Clean up the FIPS table
+    fips_df = _prep_lat_long_fips_df(raw_vcegen__lat_lon_fips)
+    # Apply the same transforms to all the capacity factor tables. This is slower
+    # than doing it to a concatinated table but less memory intensive because
```
e-belfer, line 253:
```suggestion
    # than doing it to a concatenated table but less memory intensive because
```

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+    if (length := len(asset_df)) != 136437000:
+        return AssetCheckResult(
+            passed=False,
+            description="Table unexpected length",
+            metadata={"table_length": length, "expected_length": 136437000},
+        )
+    # Make sure there are no null values
+    if asset_df.isna().any().any():
+        return AssetCheckResult(
+            passed=False, description="Found NA values when there should be none"
```
e-belfer, line 292: Non-blocking: could be helpful to return the offending row PKs as part of the metadata

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+        )
+    # Make sure the capacity_factor values are below the expected value
+    # There are some solar values that are slightly over 1 due to colder
+    # than average panel temperatures.
+    if (asset_df.iloc[:, asset_df.columns.str.contains("cap")] > 1.02).all().all():
```
e-belfer, line 297: I always get these confused - wouldn't you want .any().any() here and below?

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+        return AssetCheckResult(
+            passed=False,
+            description="Found capacity factor fraction values less than 0",
+        )
+    # Make sure the highest hour_of_year values is 8760
+    if asset_df["hour_of_year"].max() != 8760:
+        return AssetCheckResult(
+            passed=False, description="Found hour_of_year values larger than 8760"
+        )
+    # Make sure Dec 31/2029 is missing
```
e-belfer, line 313:
```suggestion
    # Make sure Dec 31, 2020 is missing (leap year handling is correct)
```

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+        .assign(state_id_fips=lambda x: x.county_id_fips.str.extract(r"(\d{2})"))
+        .assign(
+            county_or_lake_name=lambda x: x.county_state_names.str.extract(
+                rf"(.+)_({state_pattern})$"
+            )[0]
+        )
+        .merge(
+            ps_usa_df[["state_id_fips", "subdivision_code"]],
+            on=["state_id_fips"],
+            how="left",
```
e-belfer, line 58: Do you want to validate "m:1" here?

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+def _combine_cap_fac_with_fips_df(
+    cap_fac_df: pd.DataFrame, fips_df: pd.DataFrame
+) -> pd.DataFrame:
+    """Combine the combined capacity factor df with the fips df."""
+    logger.info(
+        "Merging the combined capacity factor table with the Lat-Long-FIPS table"
+    )
+    combined_df = pd.merge(cap_fac_df, fips_df, on="county_state_names", how="left")
```
e-belfer, line 189: Do you want to validate "m:1" here?

e-belfer (changes requested):
I have a bunch of fairly minor documentation changes, and a handful of questions.

Re the FIPS - I think we'd either want to leave this as N/A or make a separate output table for lakes vs. county data. At present, I think nulling for lakes seems fine, since a lake could well correspond to multiple FIPS codes.

I'm still running the code, but just wanted to flag these changes first. Might pop back with other comments once I see the finished df!

src/pudl/transform/vceregen.py
e-belfer: @aesharpe Can you explain why hour 1 is equal to YYYY-MM-DD 00:00 instead of 01:00? This isn't intuitive to me.

src/pudl/transform/vcerare.py
aesharpe: no reason! think I didn't want to run the risk of there being a hyphen or something in there. These county names are relatively informal, so whatever comes before the state pattern is what we are taking. If we wanted something more official we could just merge the County Name from our dp1 census table.

src/pudl/transform/vcerare.py
aesharpe: `datetime_utc` is in `subset_df.columns` but datetime_utc becomes the index value when we run the groupby so we don't want to included in the list of `other_cols` which is used to take the "first" value (vs. the cap fac values which use "mean") when the data are aggregated. I can try and make this more clear with some inline comments.

src/pudl/transform/vcerare.py
aesharpe: Oye yeah these are confusing. I think you're right!

src/pudl/transform/vcerare.py
aesharpe: I think it's easy enough to search the data for NA values I'm not too concerned about adding metadata for it.

src/pudl/transform/vcerare.py
aesharpe: I think this should actually be 1:1

src/pudl/metadata/resources/vcerare.py
```diff
@@ -0,0 +1,62 @@
...
+            "I-V curve translation to power capacity factors."
+        ),
+        "schema": {
+            "fields": [
+                "datetime_utc",
+                "hour_of_year",
+                "report_year",
+                "county_id_fips",
+                "county_or_lake_name",
+                "state",
```
zaneselvans, line 48: Note that for the string columns that contain a small number of distinct values (like state, FIPS codes, and the names) it may be much more space/memory efficient to store them as categoricals. We do this with the state column in the EPA CEMS hourly emissions table.

zaneselvans (changes requested):

docs/templates/vcerare_child.rst.jinja
```diff
@@ -0,0 +1,60 @@
...
+file contains an index column for time (simply 1, 2, 3...8760 to
+represent the hours in a year) and columns for each county populated with capacity
+factor values as a percentage from 0-100.
+{% endblock %}
+
+{% block notable_irregularities %}
+Sub-county regions
+------------------
+
+The data include some sub-county regions such as lakes and cities meaning that the
```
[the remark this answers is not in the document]
aesharpe: good point!

src/pudl/transform/vcerare.py
zaneselvans: Looking at [Google Maps](https://www.google.com/maps/place/Alleghany+County,+VA/@37.8614873,-80.0798489,87885m/data=!3m1!1e3!4m6!3m5!1s0x884cc397910e13fd:0xc43980e08ca7e1d0!8m2!3d37.8531975!4d-80.0534923!16zL20vMG1wZGg?hl=en&entry=ttu&g_ep=EgoyMDI0MTAxNS4wIKXMDSoASAFQAw%3D%3D) Clifton Forge appears to be quite a small town in Alleghany County, VA so dropping it rather than having it influence the overall generation profile for the county seems like it would be the bbest option.

aesharpe: Ok easy enough -- I changed the function so that it drops these two cities.

src/pudl/transform/vceregen.py
aesharpe: This is tricky -- in `datetime` the year starts at `2022-01-01 00:00:00`. If we bumped it up by an hour and coorelated hour 1 with `2022-01-01 01:00:00` then the 8760th hour would be the next year: `2023-01-01 00:00:00 ` and not line up with the report year column which would still say 2022.

aesharpe: But yeah, it's weird. I could see us going either way.

docs/data_sources/vcerare/2018 VCE Study_Dataset Methods and Analysis.pdf
zaneselvans, line 1: We asked Chris which of the two reports ought to be used as a reference and he indicated the 2020 one was preferable. Is there a reason to include this one too?

src/pudl/transform/vcerare.py
```diff
@@ -0,0 +1,399 @@
...
+        .pipe(_null_non_county_fips_rows)
+    )
+
+
+@asset_check(
+    asset=out_vcerare__hourly_available_capacity_factor,
+    blocking=True,
+    description="Check that output table is as expected.",
+)
+def check_hourly_available_cap_fac_table(asset_df: pd.DataFrame):  # noqa: C901
```
zaneselvans, line 310: Not a blocker, but with a lot of different checks inside a single function like this, we probably want to accumulate some kind of error report and then at the end of the function, if there are any errors in that report, we return the `AssetCheckResult` and include all of that information, rather than bailing as soon as we find any failure. Otherwise if there are multiple failures you'll have to run the whole asset check multiple times to find them all.

zaneselvans (changes requested):
- There are a couple of small changes in the order of operations that may speed things up significantly.
- The row count check looks like it will always fail in the fast ETL.
- There are two lake name irregularities (misspelled Hurron, St. vs. Saint Claire)

Do we want to try and make the low-cardinality string columns into categoricals to save memory / space? If you read the Parquet into Pandas from Parquet naively now, it takes like 25-28GB of memory.

docs/data_sources/vcerare/2018 VCE Study_Dataset Methods and Analysis.pdf
aesharpe: My reasoning was that it has more detail about the methodology. When Chris said the other one was preferable I didn't necessarily take that to mean that this one wasn't accurate, rather that the other was more succinct. But we can remove it if you think it might not be what Chris intended!

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   "source": [
+    "# Using the VCE Resource Adequacy Renewable Energy (RARE) Power Dataset Parquet Files"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d6d8c7fe-8e6d-44ae-b68f-7d659c45f23d",
+   "metadata": {},
+   "source": [
+    "The RARE dataset, `vcerare` in PUDL, was produced by Vibrant Clean Energy and is licensed to the public under the Creative Commons Attribution 4.0 International license (CC-BY-4.0). The data consists of hourly, county-level renewable generation profiles for solar pv, onshore, and offshore wind in the continental United States. It was compiled based on outputs from the NOAA HRRR weather model. Visit our [VCE Data Source](https://catalystcoop-pudl.readthedocs.io/en/nightly/data_sources/vcerare.html) page to learn more."
```
e-belfer, line 16:
```suggestion
    "The RARE dataset, `vcerare` in PUDL, was produced by Vibrant Clean Energy and is licensed to the public under the Creative Commons Attribution 4.0 International license (CC-BY-4.0). The data consists of hourly, county-level renewable generation profiles for solar PV, onshore, and offshore wind in the continental United States. It was compiled based on outputs from the NOAA HRRR weather model. Visit our [VCE Data Source](https://catalystcoop-pudl.readthedocs.io/en/nightly/data_sources/vcerare.html) page to learn more."
```

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    "See the [Zenodo archive README](https://zenodo.org/records/13937523) for more detail on the raw data schema and contents. \n",
+    "\n",
+    "### Processed PUDL Parquet files (S3 bucket)\n",
+    "\n",
+    "The processed data are published as Apache Parquet files--a file format designed for efficient data storage and retreival. These processed files combine all years and generation sources into one table. They also restructure the data from a wide format to a tall format by pulling county into a column value rather than a column header. Additional informational columns are added to the processed data including: \n",
+    "\n",
+    "* Latitude\n",
+    "* Longitude\n",
+    "* County FIPS code\n",
+    "* Report rear\n",
```
e-belfer, line 51:
```suggestion
    "* Report year\n",
```

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "49394cc9-116b-4fee-a445-02afcc3215b1",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "rare_df = pd.read_parquet(\n",
+    "    pudl_paths.pudl_output / \"parquet/out_vceregen__hourly_available_capacity_factor.parquet\"\n",
```
e-belfer, line 107:
```suggestion
    "    pudl_paths.pudl_output / \"parquet/out_vcerare__hourly_available_capacity_factor.parquet\"\n",
```

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    ")\n",
```
e-belfer, line 108: For me this crashed the kernel several times. Adding `engine=pyarrow` and dtype_backend='pyarrow' here might be helpful, with a short explanation of why.

E.g., 

```
rare_df = pd.read_parquet(
    pudl_paths.pudl_output / "parquet/out_vcerare__hourly_available_capacity_factor.parquet", engine='pyarrow', dtype_backend='pyarrow'
)
```

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    "rare_df.sample(10)"
```
e-belfer, line 109:
```suggestion
    "rare_df.sample(10)" # Display ten random hourly records, to give an overview of what the data looks like
```

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   "execution_count": null,
+   "id": "82d0cc7c-3b84-4dc8-b07d-4f2f6de8df79",
+   "metadata": {
+    "jupyter": {
+     "outputs_hidden": true
+    }
+   },
+   "outputs": [],
+   "source": [
+    "filtered_df = rare_df[rare_df[\"state\"]==\"RI\"]\n",
```
e-belfer, line 150:
```suggestion
    "filtered_df = rare_df[rare_df[\"state\"]==\"RI\"] # Filter to data from Rhode Island\n", 
```

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    "filtered_df.head(10)"
```
e-belfer, line 151:
```suggestion
    "filtered_df.head(10) #Inspect first ten rows"
```

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   "id": "3915c2de-215f-4b25-b718-15f14ed3dbd6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Non-editable variables\n",
+    "plot_df = rare_df[\n",
+    "    (rare_df[\"report_year\"]==year)\n",
+    "    & (rare_df[\"county_id_fips\"]==county_id)\n",
+    "]\n",
+    "county_name = plot_df[plot_df[\"county_id_fips\"]==county_id].county_or_lake_name.unique().item()\n",
```
e-belfer, line 198: If using pyarrow, item() doesn't work so you'll need to use [0] here

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    "filtered_df = rare_df[rare_df[\"state\"]==\"RI\"]\n",
+    "filtered_df.head(10)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1e5c6c8d-8a8e-48be-918d-6a9d1e14741f",
+   "metadata": {},
+   "source": [
+    "### Visualize the data"
```
e-belfer, line 159: We should add at the very least minimal text to explain what the plot is we're making.

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "cc41d9ee-3f6b-4da6-8450-6f649bef2744",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# random county_id generator:\n",
+    "random_county = random.choice(rare_df[\"county_id_fips\"])"
```
e-belfer, line 170: Ideally here, "Select a random county by default, or substitute in your own 5-digit FIPS ID". I think called the variable `county_fips_id` instead of `random_county` will be better here.

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    "]\n",
+    "county_name = plot_df[plot_df[\"county_id_fips\"]==county_id].county_or_lake_name.unique().item()\n",
+    "state_name = plot_df[plot_df[\"county_id_fips\"]==county_id].state.unique().item()\n",
+    "\n",
+    "# Make the chart\n",
+    "plt.hist(plot_df[f\"capacity_factor_{gen_type}\"], bins=100, range=(0, 1))\n",
+    "plt.title(\n",
+    "    f\"county: {county_name}; state: {state_name}; gen: {gen_type}; max:{plot_df[f\"capacity_factor_{gen_type}\"].max():.2f}\")\n",
+    "plt.grid()\n",
+    "plt.show();"
```
e-belfer, line 206:
```suggestion
    "plt.show()"
```

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    "plot_df = rare_df[\n",
+    "    (rare_df[\"report_year\"]==year)\n",
+    "    & (rare_df[\"county_id_fips\"]==county_id)\n",
+    "]\n",
+    "county_name = plot_df[plot_df[\"county_id_fips\"]==county_id].county_or_lake_name.unique().item()\n",
+    "state_name = plot_df[plot_df[\"county_id_fips\"]==county_id].state.unique().item()\n",
+    "\n",
+    "# Make the chart\n",
+    "plt.hist(plot_df[f\"capacity_factor_{gen_type}\"], bins=100, range=(0, 1))\n",
+    "plt.title(\n",
```
e-belfer, line 203: We should add x and y axis labels here.

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "13289ced-0028-4d79-9e5a-cf79bc464d80",
+   "metadata": {},
+   "source": [
+    "#### Memory check - will it Excel? \n",
+    "\n",
+    "If you want to know whether your table is capable of being processed as a csv you can use this memory estimator. If the estimated memory exceeds 500 MB it's too big! Cutting columns or making the filter scope smaller will help reduce the file size."
```
e-belfer, line 224: If we're saying "cutting columns", we should show people how to drop columns probably.

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    "print(f\"Estimated CSV size: {csv_size_estimate_mb:.2f} MB\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "17bff202-4766-44e1-8442-86108335162b",
+   "metadata": {},
+   "source": [
+    "#### Download!\n",
+    "Specify your desired download location by filling in the `download_path` and running this cell will output the data to that location under the name `rare_power_data.csv`."
```
e-belfer, line 251: We'll need to give people more guidance here, since path names often trip people up. Also, having the slash on the file name will cause problems if people just leave this blank.

I'd suggest:

```
# Add the file path you want to download the data to
# Leave this blank to save the data in the same folder as this notebook.
download_path = "" 
# e.g. download_path = "/home/user/Desktop/folder/data_folder/"
plot_df.to_csv(download_path+"rare_power_data.csv")
```

notebooks/work-in-progress/VCERARE.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7c880907-818c-47e0-bfc9-2ee270d8893d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Add the file path you want to download the data to\n",
```
e-belfer, line 261: We'll need to give people more guidance here, since path names often trip people up. Also, having the slash on the file name will cause problems if people just leave this blank.

I'd suggest:

```
# Add the file path you want to download the data to
# Leave this blank to save the data in the same folder as this notebook.
download_path = "" 
# e.g. download_path = "/home/user/Desktop/folder/data_folder/"
plot_df.to_csv(download_path+"rare_power_data.csv")
```

e-belfer (dismissed):
Some comments on the notebook, which is very close but doesn't currently work out of the box.

notebooks/work-in-progress/VCERARE.ipynb
e-belfer: Also should note I'm not sure that you can just accept this commit because of the weirdness of notebook formatting in git.

zaneselvans:
* The two small order-of-operations changes didn't have big impacts.
* Converting the input dataframes to use `dtype_backend="pyarrow"` also did not help.
* Peak memory usage for the output asset appears to be about 13GB (based on watching btop locally)
* The `df.stack()` operations each take about 20 sec to run, and spike memory usage.
* The `df.concat()` operation takes about 2 min to run and also spikes memory usage.

I haven't found a simple way around it, so I have created a new superchonk asset tag: `"memory-use": "very-high"` that limits concurrency to 1 for the VCE asset.

I've also added `"memory-use": "high"` to the asset check since it seems to be peaking at more than 4GB of memory usage.

notebooks/work-in-progress/VCERARE.ipynb
zaneselvans: Yeah I think it'll b0rk here, but I will integrate all this feedback into the version of the notebook that goes on Kaggle.

zaneselvans:
* Unfortunately concurrency limits only operate within a single tag group, so the `very-high` tag didn't help.
* Additional configuration is required to make retries work, so it won't help us immediately.
* Going with the brute force approach of making the VM bigger temporarily.

zaneselvans:
I've compiled some loose ends to clean up after the data release in #3914

zaneselvans (approved):

kmk142789 (approved):

PR #3904, opened 2024-10-10
Update to micromamba 2.0.2 Docker image.

zaneselvans:
# Overview

- Update to the micromamba v2 Docker image.
- Update to the newest version of the `hadolint-py` pre-commit hook since the old one started failing.

Note: [hadolint](https://github.com/hadolint/hadolint) seems to have been abandoned and [the prettier commit hook](https://github.com/pre-commit/mirrors-prettier) has been archived due to incompatible changes in the underlying prettier library, so we should probably try and replace those hooks (Dockerfile linting, and YAML linting/formatting)

# Testing

- Ran a branch deployment using the new Dockerfile

```[tasklist]
# To-do list
- [x] Run the `build-deploy-pudl` GitHub Action manually to make sure the new Docker image works.
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,7 +47,7 @@ repos:
 
   # Lint Dockerfiles for errors and to ensure best practices
   - repo: https://github.com/AleksaC/hadolint-py
-    rev: v2.12.0.3
+    rev: v2.12.1b3
     hooks:
       - id: hadolint
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,6 @@
 FROM mambaorg/micromamba:2.0.2
 
+ENV CONTAINER_HOME=/home/$MAMBA_USER
 ENV PGDATA=${CONTAINER_HOME}/pgdata
 
 USER root
@@ -29,7 +30,6 @@ RUN pg_createcluster 15 dagster -u "$MAMBA_USER" -- -A trust
 
 # Switch back to being non-root user and get into the home directory
 USER $MAMBA_USER
-ENV CONTAINER_HOME=/home/$MAMBA_USER
 WORKDIR ${CONTAINER_HOME}
 
 ENV CONDA_PREFIX=${CONTAINER_HOME}/env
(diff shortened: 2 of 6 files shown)


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mambaorg/micromamba:1.5.10
+FROM mambaorg/micromamba:2.0.2
 
 ENV PGDATA=${CONTAINER_HOME}/pgdata
```
jdangerx, line 3: I guess we've just been putting the Postgres data in `/pgdata` this whole time? That seems fine, I guess, though we should probably move this down to where `CONTAINER_HOME` *is* defined.

.pre-commit-config.yaml
```diff
@@ -47,7 +47,7 @@ repos:
 
   # Lint Dockerfiles for errors and to ensure best practices
   - repo: https://github.com/AleksaC/hadolint-py
-    rev: v2.12.0.3
+    rev: v2.12.1b3
     hooks:
       - id: hadolint
```
jdangerx, line 52: Tried out the official `hadolint` [pre-commit hook](https://github.com/hadolint/hadolint/blob/master/docs/INTEGRATION.md#pre-commit) using this configuration:

```yaml
  - repo: https://github.com/hadolint/hadolint
    rev: v2.12.0
    hooks:
      - id: hadolint-docker
```

And it worked locally with `pre-commit run -a hadolint-docker`. Should we switch over if `hadolint-py` is unmaintained?

jdangerx (approved):
Fine to go in as is but I have a few small suggestions to clean things up a bit.

.pre-commit-config.yaml
zaneselvans: I don't think `hadolint-docker` will work without Docker installed and running locally will it?

And in any case, it's  the underlying `hadolint` that's been abandoned (I misspoke in the initial comment) so I'm wondering if there are alternatives for Dockerfile linting out there.

docker/Dockerfile
zaneselvans: I suspect that we need `$PGDATA` to be defined prior to installing postgres and creating a postgres cluster, so maybe it would be better to move the definition of `$CONTAINER_HOME` up to the top.

jdangerx: Yes, that's probably better!

Issue #3770, opened 2024-08-07
Integrate PHMSA distribution data into PUDL


e-belfer:
## PHMSA distribution data (1990-present)

**SEE https://github.com/catalyst-cooperative/pudl/pull/4005 and closing this one.**

PHMSA distribution data explains the extent, safety record, and characteristics of each operator's distribution system by state and commodity group. The first goal for PHMSA data integration should be to create cleaned and normalized tables from the PHMSA distribution data (likely 3-4 separate tables). Unlike transmission, distribution is all published in one form. The raw data is archived [here](https://zenodo.org/records/13149085), and each zip file contains the PDF form that is filled out, which will be helpful for visualizing data structure.

### Prep
Before getting into the table transforms, let's update the data itself.

```[tasklist]
### Tasks
- [x] Update DOI in `pudl.workspace.datastore` to latest archive
- [x] In `pudl.metadata.sources.SOURCES`, update the working_partitions to reflect the years, months, or quarters of data that are available
- [x] Update the partitions of data to be processed in the `etl_full.yml` and `etl_fast.yml` settings files stored under `src/pudl/package_data/settings`
- [x] Use the `pudl_datastore` script to download the new raw data
- [x] Run the extraction on the new data by generating `raw_phmsa` assets and debug any issues
- [x] Create new module: `transform.phmsa` to handle transforms
```

### Table design
- All form parts below are based on the 2021 form. Form part letters change over time, so this will require pairing older fields from .xlsx files to their correct tables. The original files (CSV and Excel) are not split by form part.
- Each report corresponds to 1 state and one commodity group. Further parts of the table have dozens of columns that we should tidy into one categorical column, creating different PKs within the parts of the form. We should try to structure our tables to match these PKs as much as possible.

### `core_phmsagas__yearly_distribution_operators`
Part A: Operator Information
- Essentially just the operator ID, name, address, and info on the state and commodity pertinent to the report.

Part D-I
- Each field occurs once per report, these should probably be kept in one table with Part A.
- Part D, Excavation Damage: 6 fields of summary stats
- Part E, EFV and Service Valve Data: 4 fields of summary stats
- Part F: Total number of leaks on federal lands scheduled for repair or repaired.
- Part G: %age of unaccounted gas.
- Part H: Additional Info: a Notes field, discusses corrections, changes in calculations, and other ambiguities.
- Part I: Preparer's info: Email, initial or supplemental report, contact info for preparer.

```[tasklist]
[...]

PR #3929, opened 2024-10-24
Issue 3770 transform phmsagas data

seeess1:
# Overview

This is still WIP!

Relates to #3770.

Goal of the PR is to complete the first transformation of raw PHMSA data into a core asset.

## What did you change?

- Added a new transform script for PHMSA data.
- Made one helper method (`fix_eia_na`) more generic since it can be applied across data sets (and updated references accordingly).
- Added a few other helper methods.
- Updated column map files with 2023 values (pulled from whatever was in 2022).
- Specifically changed the ordering of columns in `/Users/sam/Documents/pudl/src/pudl/package_data/phmsagas/column_maps/yearly_distribution.csv` since we had fax columns mapped to emails and vice versa.

Will come back to everything below once this draft PR has gone through an initial round of review.

# Documentation

```[tasklist]
```

# Testing

```[tasklist]
# To-do list
```


src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,164 @@
...
+    df.loc[mask, "report_year"] = df.loc[mask, "report_year"].apply(
+        lambda x: 2000 + x if x < 50 else 1900 + x
+    )
+
+    # Fill NA values with zeroes because these columns are simply counts.
+    # Note that "excavation_damage..." columns should sum up to the value in "excavation_damage_total". However, many rows
+    # (on the scale of thousands) do not actually sum up to "excavation_damage_total".
+    df[YEARLY_DISTRIBUTION_OPERATORS_COLUMNS["columns_with_nas_to_fill"]] = df[
+        YEARLY_DISTRIBUTION_OPERATORS_COLUMNS["columns_with_nas_to_fill"]
+    ].fillna(0)
```
e-belfer, line 126: We distinguish between NA as "not reported" and 0 as "reported as 0", so filling this would actually change the meaning here (from "I didn't fill this in" to "I had 0 incidents"). Please drop this step.

src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,164 @@
...
+    mask = df["report_year"] < 100
+
+    # Convert 2-digit years to appropriate 4-digit format (assume cutoff at year 50)
+    # We could also use the first 4 digits of the "report_number" but there was at least one anomaly here with an invalid year
+    df.loc[mask, "report_year"] = df.loc[mask, "report_year"].apply(
+        lambda x: 2000 + x if x < 50 else 1900 + x
+    )
+
+    # Fill NA values with zeroes because these columns are simply counts.
+    # Note that "excavation_damage..." columns should sum up to the value in "excavation_damage_total". However, many rows
```
e-belfer, line 122: We should definitely add this validation as an asset check, and use fillna(0) there to do the comparison calculation.

e-belfer (commented):
This is looking really good! A few comments on what's already been done, you're making great progress.

Some next steps:
- the values in `percent_unaccounted_for_gas` look like they could use some normalization (I see some negative values)
- once you're happy with the table, we can go ahead and write it to the database. See [here](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-pudl-db-schema) for guidance on how to do this.

src/pudl/helpers.py
```diff
@@ -1009,19 +1009,19 @@ def remove_leading_zeros_from_numeric_strings(
     return df
 
 
-def fix_eia_na(df: pd.DataFrame) -> pd.DataFrame:
+def fix_na(df: pd.DataFrame) -> pd.DataFrame:
     """Replace common ill-posed EIA NA spreadsheet values with np.nan.
```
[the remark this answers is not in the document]
zaneselvans: If we're changing this function name, maybe something more descriptive like `standardize_na_values()` would be better than "fix" since I could see fixing NA values meaning filling them in with non-null values.

src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,164 @@
...
+            YEARLY_DISTRIBUTION_OPERATORS_COLUMNS["capitalization_exclusion"]
+        )
+    ] = df[
+        df.select_dtypes(include=["object"]).columns.difference(
+            YEARLY_DISTRIBUTION_OPERATORS_COLUMNS["capitalization_exclusion"]
+        )
+    ].apply(lambda col: col.str.title())
+
+    # List of state columns to standardize
+    df = standardize_state_columns(
```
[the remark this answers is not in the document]
zaneselvans: I think we've used an `enum` constraint for the state column in EPA CEMS because it's a huge table and we want it to be interpreted as a categorical value to save space, and we don't use foreign key constraints in the hourly data that's only written to parquet. But for tables going into the database I think we would constrain the state values with a foreign key constraint that points at the `core_pudl__codes_subdivisions` table, and use the encoder setup to do the cleaning.

src/pudl/helpers.py
```diff
@@ -2185,3 +2185,305 @@ def retry(
...
+        "Vermont": "VT",
+        "Virginia": "VA",
+        "Washington": "WA",
+        "West Virginia": "WV",
+        "Wisconsin": "WI",
+        "Wyoming": "WY",
+    }
+
+    for col in state_columns:
+        df[col] = df[col].replace(state_abbreviations).str.upper()
```
zaneselvans, line 2487: When you have a complete mapping between old and new values, the [col.map()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.map.html) method should be much faster. However any value that appears in the series but not in the dictionary defining the mapping gets converted to NA.

src/pudl/helpers.py
```diff
@@ -2185,3 +2185,305 @@ def retry(
...
+    Numbers with extensions will be appended with "x#".
+    Non-numeric entries will be returned as np.nan. Entries with fewer than
+    10 digits will be returned with no hyphens.
+
+    Args:
+    df: The DataFrame to modify.
+    columns: A list of the names of the columns that need to be standardized
+    Returns:
+    The modified DataFrame with standardized phone numbers in the same column.
+    """
```
zaneselvans, line 2203: This docstring won't render correctly in the docs. Check how it looks in the built API docutmentation.

```suggestion
    """Standardize phone numbers in the specified columns of the DataFrame.
    
    US numbers: ###-###-####
    International numbers with the international code at the beginning.
    Numbers with extensions will be appended with "x#".
    Non-numeric entries will be returned as np.nan. Entries with fewer than
    10 digits will be returned with no hyphens.

    Args:
        df: The DataFrame to modify.
        columns: A list of the names of the columns that need to be standardized

    Returns:
        The modified DataFrame with standardized phone numbers in the same column.
    """
```

src/pudl/helpers.py
```diff
@@ -2185,3 +2185,305 @@ def retry(
...
+        df[column] = df[column].apply(standardize_phone_number)
+
+    return df
+
+
+def analyze_missing_values(df: pd.DataFrame) -> list[str]:
+    """Analyze columns of a DataFrame for missing or invalid values. Note that this is purely for analysis
+    and does not perform any data transformation or cleaning.
+    This function checks each column for missing or custom missing values and prints
+    a summary of the findings for string (object), numeric, and datetime columns.
```
zaneselvans, line 2267:
```suggestion
    """Analyze columns of a DataFrame for missing or invalid values.
    
    Note that this is purely for analysis and does not perform any data
    transformation or cleaning. This function checks each column for
    missing or custom missing values and prints a summary of the findings
    for string (object), numeric, and datetime columns.
```

seeess1:
> This is looking really good! A few comments on what's already been done, you're making great progress.
> 
> Some next steps:
> 
> * the values in `percent_unaccounted_for_gas` look like they could use some normalization (I see some negative values)
> * once you're happy with the table, we can go ahead and write it to the database. See [here](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#update-the-pudl-db-schema) for guidance on how to do this.

Ah yep I meant to point that part out about `percent_unaccounted_for_gas`. I noticed that too. I honestly wasn't sure how we wanted to handle this one. Any suggestions?

src/pudl/helpers.py
```diff
@@ -2185,3 +2185,305 @@ def retry(
...
+        # Split phone number and extension, if present
+        phone_parts = re.split(r"[xX]", str(phone))
+        phone_main = phone_parts[0]  # Main phone number
+        extension = (
+            phone_parts[1].strip() if len(phone_parts) > 1 else None
+        )  # Extension, if it exists
+
+        # Remove unwanted characters (parentheses, spaces, periods, and dashes) from the main phone number
+        phone_main = re.sub(
+            r"[^\d]", "", phone_main.replace(".0", "")
```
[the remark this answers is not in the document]
seeess1: There were lots of phone numbers in the raw data with ".0" at the end of the value. So I'm using `replace` to remove those first.

src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,164 @@
...
+
+##############################################################################
+# PHMSAGAS transformation logic
+##############################################################################
+
+
+@asset(
+    ins={"raw_data": AssetIn(key="raw_phmsagas__yearly_distribution")},
+    io_manager_key=None,  # TODO: check on this one ("pudl_io_manager"? something else?)
+    compute_kind=None,  # TODO: check on this one
```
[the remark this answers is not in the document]
seeess1: Good to know. I'll play around with this a little bit more.

src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,164 @@
...
+    """Build the :ref:`core_phmsagas__yearly_distribution_operators`."""
+    df = raw_data[YEARLY_DISTRIBUTION_OPERATORS_COLUMNS["columns_to_keep"]]
+    df[YEARLY_DISTRIBUTION_OPERATORS_COLUMNS["columns_to_convert_to_ints"]] = df[
+        YEARLY_DISTRIBUTION_OPERATORS_COLUMNS["columns_to_convert_to_ints"]
+    ].astype("Int64")
+
+    # Ensure all "report_year" values have four digits
+    mask = df["report_year"] < 100
+
+    # Convert 2-digit years to appropriate 4-digit format (assume cutoff at year 50)
```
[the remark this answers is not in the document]
seeess1: No values below 50 in the data right now. I mostly put this in to keep it flexible enough to also be able to handle post-2000 data if we started seeing values like "00", "01", etc. The column still feels pretty reliable to me overall. It was just a 2 vs 4 digit year formatting issue.

src/pudl/helpers.py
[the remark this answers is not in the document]
zaneselvans: You might want to use a regex that only matches this pattern at the end of the string to avoid accidentally removing characters from the middle of a number that could be reported like 303.443.0123

zaneselvans:
Would `percent_unaccounted_for_gas < 0` mean that extra gas appeared from somewhere?

Is the distribution of the absolute values of positive and negative values similar?

Does it seem likely that different respondents just assumed different sign conventions here? Can we see what question they're responding to and if it is phrased ambiguously? Like is "5% of the gas was unaccounted for" supposed to be -5% or 5%?

src/pudl/transform/phmsagas.py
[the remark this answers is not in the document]
seeess1: Getting rid of my duplicate function and utilizing `POLITICAL_SUBDIVISIONS`. Plus adding an `enum` constraint to "state" now.

src/pudl/helpers.py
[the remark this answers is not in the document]
seeess1: Updated

src/pudl/helpers.py
```diff
@@ -2185,3 +2185,305 @@ def retry(
...
+            print(f"Caught exception for column {col}: {e}\n")
+            continue
+
+    print(f"Columns with NaNs or custom missing values: {nan_cols}")
+    print(f"Columns with exceptions during processing: {exception_cols}")
+
+    return exception_cols
+
+
+def standardize_state_columns(df: pd.DataFrame, state_columns: list) -> pd.DataFrame:
```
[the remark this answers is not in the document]
seeess1: Updated

src/pudl/transform/phmsagas.py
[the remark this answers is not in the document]
seeess1: Updated

src/pudl/transform/phmsagas.py
seeess1: Updated

src/pudl/helpers.py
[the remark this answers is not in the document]
seeess1: Good point. I'm removing that logic. The regex should be all that's needed here.

src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # Function for looking up state names and abbreviations
+    def standardize_state(state):
+        if pd.isna(state):
+            return state
+        state = state.strip()
+        standardized_state = state_to_abbr.get(state, state)
+        if standardized_state not in state_to_abbr.values():
+            return np.nan
+        return standardized_state
```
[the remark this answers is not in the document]
zaneselvans: Also note that you need to be careful to avoid mixing incompatible NA types.  `np.nan` is only appropriate for numpy floating point values. If it gets mixed in with other types in a column, the whole column has to be treated as an `object`.

For all the nullable Pandas dtypes (e.g. `Int64` and `string`) the appropriate NA value will be `pd.NA`.

src/pudl/helpers.py
```diff
@@ -2185,3 +2194,219 @@ def retry(
...
+
+    Args:
+    df: The DataFrame to modify.
+    columns: A list of the names of the columns that need to be standardized
+    Returns:
+    The modified DataFrame with standardized phone numbers in the same column.
+    """
+
+    # Function to clean and standardize phone numbers
+    def standardize_phone_number(phone: str) -> str:
```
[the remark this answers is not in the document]
zaneselvans: See my related comment in the PHMSA transform module.

src/pudl/helpers.py
```diff
@@ -2185,3 +2194,219 @@ def retry(
...
+        "NA",
+        "N/A",
+        "NULL",
+        "-",
+        "None",
+        "NaN",
+        "?",
+        "*",
+        "#",
+    ]
```
zaneselvans, line 2285: If this function is going to stick around post-PR then we probably want the user to be able to define their own set of custom missing values -- though these could be retained as a default set.

zaneselvans: Probably also want to switch all of the `print()` statements to `logger.info()` outputs so it can be managed along with all of the other outputs.

zaneselvans (changes requested):

src/pudl/transform/phmsagas.py
```diff
@@ -165,10 +172,10 @@ def standardize_state(state):
...
     df["office_address_state"] = df["office_address_state"].apply(standardize_state)
 
-    # Trim all the object-type columns
-    df[df.select_dtypes(include=["object"]).columns] = df.select_dtypes(
-        include=["object"]
-    ).applymap(lambda x: x.strip() if isinstance(x, str) else x)
+    # Strip whitespace from all object (string) columns in the DataFrame
+    df[df.select_dtypes(include=["object"]).columns] = df.select_dtypes(include=["object"]).apply(
+        lambda col: col.map(lambda x: x.strip() if isinstance(x, str) else x)
+    )
```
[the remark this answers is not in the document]
seeess1: It's been running very quickly on my dinky computer (less than half a second).

src/pudl/transform/phmsagas.py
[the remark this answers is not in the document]
seeess1: Ah whoops thanks for catching that. Fixing that now.

src/pudl/helpers.py
seeess1: Updated

src/pudl/transform/phmsagas.py
seeess1: Got it. I'm seeing that even if I fill in NAs with zero, there are almost 4k rows where the `excavation_damage...` values don't sum to `excavation_damage_total`. How would you recommend handling these?

src/pudl/helpers.py
seeess1: Updated

src/pudl/helpers.py
seeess1: Updated

src/pudl/helpers.py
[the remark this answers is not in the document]
seeess1: Actually I'm gonna leave in one line toward the top to get rid of these substrings just to make sure the rest of the number standardization is working with cleaner data.

Updated.

src/pudl/helpers.py
seeess1: Updated logic for standardizing state abbreviations here:

https://github.com/catalyst-cooperative/pudl/pull/3929/files#diff-941a3989c384fa527b5188c725a5b3beb1f1e0dbf9d427dae9dd82f4f9fe5c09R126

src/pudl/helpers.py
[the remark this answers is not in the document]
seeess1: Thanks for pointing this out. Logic has been updated.

src/pudl/transform/phmsagas.py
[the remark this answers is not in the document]
seeess1: Updated

seeess1:
> Would `percent_unaccounted_for_gas < 0` mean that extra gas appeared from somewhere?
> 
> Is the distribution of the absolute values of positive and negative values similar?
> 
> Does it seem likely that different respondents just assumed different sign conventions here? Can we see what question they're responding to and if it is phrased ambiguously? Like is "5% of the gas was unaccounted for" supposed to be -5% or 5%?

Here's the part of the form where this is coming from (from the [2004-2009 zip file](https://zenodo.org/records/13149085)):

image

Maybe negative values are coming from "appropriate adjustments"? That part's pretty vague. 

In either case, only 3.37% of values in this column are negative. I'm seeing negative values from across many different years, operators, states, etc. I was wondering if there might be a clear pattern here (one operator continually filing negative values for some reason) but so far I'm not seeing that. Can keep digging if needed.

seeess1:
@e-belfer @zaneselvans I've responded to the initial round of comments. I still haven't solved the riddle of the negative unaccounted gas values. And I need to take a closer look at `io_manager_key` and `compute_kind`. But hopefully the PR is starting to shape up here.

src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,159 @@
...
+
+@asset(
+    ins={"raw_data": AssetIn(key="raw_phmsagas__yearly_distribution")},
+    io_manager_key=None,  # TODO: check on this one ("pudl_io_manager"? something else?)
+    compute_kind=None,  # TODO: check on this one
+)
+def core_phmsagas__yearly_distribution_operators(
+    raw_data: pd.DataFrame,
+) -> pd.DataFrame:
+    """Build the :ref:`core_phmsagas__yearly_distribution_operators`."""
```
e-belfer, line 89: Here, include a short bulleted list of the transforms you make in this function so that this shows up in our docs. See pudl.transform.eia860 for some examples.

src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,159 @@
...
+    for col in df.select_dtypes(include=["object"]).columns:
+        df[col] = df[col].str.strip()
+
+    # Specify the columns to convert to integer type
+    cols_to_convert = YEARLY_DISTRIBUTION_OPERATORS_COLUMNS[
+        "columns_to_convert_to_ints"
+    ]
+
+    # Fill NaN values with pd.NA, then cast to "Int64" nullable integer type
+    df[cols_to_convert] = df[cols_to_convert].fillna(pd.NA).astype("Int64")
```
e-belfer, line 107: Non-blocking: `convert_dtypes()` might more simply achieve what you're trying to do here, since it handles the conversion of nulls to Pandas nullable dtypes.

src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,159 @@
...
+        "federal_land_leaks_repaired_or_scheduled",
+        "percent_unaccounted_for_gas",
+        "additional_information",
+        "preparer_email",
+        "preparer_fax",
+        "preparer_name",
+        "preparer_phone",
+        "preparer_title",
+    ],
+    "columns_to_convert_to_ints": [
```
e-belfer, line 58: `federal_land_leaks_repaired_or_scheduled` should also be in this list, I believe!

src/pudl/transform/phmsagas.py
```diff
@@ -0,0 +1,159 @@
...
+        "excavation_damage_one_call_notification",
+        "excavation_damage_other",
+        "excavation_damage_total",
+        "excavation_tickets",
+        "services_efv_in_system",
+        "services_efv_installed",
+        "services_shutoff_valve_in_system",
+        "services_shutoff_valve_installed",
+    ],
+    "capitalization_exclusion": ["headquarters_address_state", "office_address_state"],
```
e-belfer, line 73: `preparer_email` should also get excluded here, emails are case insensitive

e-belfer (changes requested):
A few minor comments, but I think you're basically ready to move on to writing this to the PUDL DB. The next steps should be outlined in the issue, but obviously let me know if you have questions about anything!

src/pudl/workspace/datastore.py
```diff
@@ -206,7 +206,7 @@ class ZenodoDoiSettings(BaseSettings):
     ferc60: ZenodoDoi = "10.5281/zenodo.13149090"
     ferc714: ZenodoDoi = "10.5281/zenodo.13149091"
     gridpathratoolkit: ZenodoDoi = "10.5281/zenodo.10892394"
-    phmsagas: ZenodoDoi = "10.5281/zenodo.10493790"
+    phmsagas: ZenodoDoi = "10.5281/zenodo.13624526"
```
e-belfer, line 209: @seeess1 One more thing to flag - we published new PHMSA data this month, so the DOI should now be:

```suggestion
    phmsagas: ZenodoDoi = "10.5281/zenodo.14026420"
```

src/pudl/helpers.py
```diff
@@ -2189,111 +2203,105 @@ def retry(
...
 
     return df
 
 
-def analyze_missing_values(df: pd.DataFrame) -> list[str]:
-    """Analyze columns of a DataFrame for missing or invalid values. Note that this is purely for analysis
-    and does not perform any data transformation or cleaning.
-    This function checks each column for missing or custom missing values and prints
-    a summary of the findings for string (object), numeric, and datetime columns.
+def analyze_missing_values(
```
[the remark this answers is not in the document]
seeess1: Yep all of that is correct. This was just a tool for analysis that could be used in the middle of dev work. 

But I honestly don't feel too precious about it. I'll add a note in the docstring now to make sure that it's clear this is just an analytical tool and not one used in any data transformation. But I'm also totally okay if you want to drop it entirely.

src/pudl/transform/phmsagas.py
seeess1: Good catch

src/pudl/metadata/enums.py
```diff
@@ -640,3 +640,11 @@
...
 ]
 """Income types for FERC Form 1 data."""
+
+US_STATE_CODES: set[str] = set(
+    POLITICAL_SUBDIVISIONS.loc[
+        (POLITICAL_SUBDIVISIONS.country_code == "USA")
+        & (POLITICAL_SUBDIVISIONS.subdivision_type == "state"),
+        "subdivision_name",
+    ]
+)
```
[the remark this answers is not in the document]
seeess1: Updated

src/pudl/metadata/fields.py
```diff
@@ -4249,6 +4250,7 @@
         # different tables. E.g. state of the utility's HQ address vs. state that a
         # plant is located in vs. state in which a utility provides service.
         "description": "Two letter US state abbreviation.",
+        "constraints": {"enum": US_STATE_CODES},
```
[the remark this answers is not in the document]
seeess1: I'm going to work on this part next.

src/pudl/helpers.py
```diff
@@ -1009,19 +1009,33 @@ def remove_leading_zeros_from_numeric_strings(
...
+    na_patterns = r"(^\.$|^\s*$|^-+$)"
+
+    # Replace empty strings with np.nan
+    df = df.replace("", np.nan)
+
+    # Replace matching patterns in all object columns with np.nan
+    df = df.replace(na_patterns, np.nan, regex=True)
+
+    # Attempt to infer original types where possible to avoid unwanted type changes
+    df = df.infer_objects()
```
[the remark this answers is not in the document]
seeess1: Updated

src/pudl/helpers.py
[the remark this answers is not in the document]
seeess1: Updated

src/pudl/transform/phmsagas.py
seeess1: Updated

src/pudl/transform/phmsagas.py
seeess1: Updated

src/pudl/transform/phmsagas.py
seeess1: Updated

src/pudl/workspace/datastore.py
seeess1: Updated

src/pudl/transform/phmsagas.py
seeess1: Updated

src/pudl/helpers.py
```diff
@@ -1009,19 +1009,33 @@ def remove_leading_zeros_from_numeric_strings(
...
 
 
-def fix_na(df: pd.DataFrame) -> pd.DataFrame:
+def standardize_na_values(df: pd.DataFrame) -> pd.DataFrame:
     """Replace common ill-posed EIA NA spreadsheet values with np.nan.
 
     Currently replaces empty string, single decimal points with no numbers,
-    any single whitespace character, and hyphens with np.nan.
+    any single whitespace character, and hyphens with np.nan. Also converts
+    columns to their original types where possible (using 'infer_objects')
```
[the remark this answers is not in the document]
e-belfer: Ah this was vaguely in reference to my comment about `convert_dtypes()` vs `infer_objects()`, sorry that wasn't clearer!

zaneselvans:
I could see the change of adding an `ENUM` constraint on the `state` field being complicated since it changes the type to `categorical` rather than `string` which may make it hard to work with different `state` columns in different tables together. If that's a frustration or a hold up then I think just cleaning the column and then leaving it unconstrained (until we can come back and really deal with `state` across the board) is also totally fine.

src/pudl/helpers.py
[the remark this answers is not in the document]
seeess1: Whoops! Looks like my documentation was misleading. No call is made to `infer_objects` in this method. I've removed that part of the function description.

seeess1:
Hey @e-belfer! Made some more updates (commit [here](https://github.com/catalyst-cooperative/pudl/pull/3929/commits/6f504e5a2247bf2f6a12043a4d6ede09f7e891f6)):

- Updated `.init` files.
- Added a new PHMSA resource in `pudl.metadata.resources`.
- Added new fields.

I think the last couple things are for me to:

- Define the PK (more on that below).
- Create a migration.
- Run all tests to make sure nothing is off.

Does that sound right?

**Primary keys**

Regarding primary keys, I figured that the logical PK identifier for this asset would be the combined `operator_id_phmsa` and `report_number`. But apparently there are a few rows with the same values here but slightly different metadata (for example, different office addresses for each row). I've attached the output from filtering down to rows that have the same combined values for these columns. Let me know how you'd like to handle these.
[duplicates.csv](https://github.com/user-attachments/files/18049110/duplicates.csv)

e-belfer:
@seeess1 Always some annoying duplicates! Low stakes here since all the actual data is luckily identical, but thanks for catching this. Some thoughts:

- The last four look completely identical to me and could get handled by `df.drop_duplicates()` unless I'm missing some tiny difference?
- For Williamsville, the one that doesn't have any "testing" in the names will be the one we want.
- For the other 2 cases, would prefer to just have a consistent rule and in this case the first row for each looks intuitively more correct (both city addresses actually match the system they're located in). Would be good to add a tiny assertion in here that checks that you aren't dropping more than the expected number of rows of data (2) in case this becomes a huge problem in the next year of data, e.g.

src/pudl/helpers.py
```diff
@@ -2264,14 +2259,17 @@ def analyze_missing_values(
...
+    any data transformation or cleaning.
+
+    This function checks each column for missing or custom missing values
+    and logs a summary of the findings for string (object), numeric, and
+    datetime columns.
 
     Args:
         df: The DataFrame to analyze.
-        custom_missing_values: Optional list of custom values to consider
+    custom_missing_values: Optional list of custom values to consider
```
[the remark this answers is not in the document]
seeess1: Updated

seeess1:
> Looking on track! I can do a final review once the last little bits are in place, but your remaining to-do list looks right to me.

Pushed up my latest updates. Let me know what you think.

src/pudl/transform/phmsagas.py
```diff
@@ -219,10 +318,31 @@ def _check(df):
         ):
             error = "Percentage of rows with negative percent_unaccounted_for_gas values: {negative_percentage:.2f}"
             logger.info(error)
+            return AssetCheckResult(passed=False, metadata={"errors": error})
+
+        return AssetCheckResult(passed=True)
 
-        if error:
+    @asset_check(asset=spec.asset, blocking=True)
```
e-belfer, line 325: Not entirely following the logic here - the asset check will run on the `df` after you've dropped these rows in the transform step to create a unique PK, so the `non_unique_groups` returned in this asset check should always be 0. Rather than writing an asset check here, we'll want to actually embed this assertion in the code itself. 

Could be as simple as something like this in the actual filtering functions you've written:
```
assert unfiltered_df.shape[0] - filtered_df.shape[0]  <=3, "Dropped more than expected number of rows due to non-unique groups. Rows dropped: {dropped_rows}
```

src/pudl/transform/phmsagas.py
```diff
@@ -179,16 +179,113 @@ def core_phmsagas__yearly_distribution_operators(
...
+    """Filters out rows with "test" in address columns. The logic is as follows:
+
+    1. For any group of rows with the same combination of "operator_id_phmsa"
+       and "report_number":
+        - If at least one row in the group does not contain the string "test"
+          (case-insensitive) in either "office_address_street" or
+          "headquarters_address_street", keep only the rows in the group
+          that do not contain "test" in these columns.
+        - If all rows in the group contain "test" in either of the columns,
+          leave the group unchanged.
```
e-belfer, line 215: The indentation here is causing `make docs-build` to fail. Should look like:

```suggestion
    1. For any group of rows with the same combination of "operator_id_phmsa"
       and "report_number":
       - If at least one row in the group does not contain the string "test"
       (case-insensitive) in either "office_address_street" or
       "headquarters_address_street", keep only the rows in the group
       that do not contain "test" in these columns.
       - If all rows in the group contain "test" in either of the columns,
       leave the group unchanged.
```

src/pudl/transform/phmsagas.py
```diff
@@ -219,10 +318,31 @@ def _check(df):
...
+                f"Number of filtered non-unique rows ({len(filtered_non_unique_rows)})\n"
+                f"Exceeds the threshold of {spec.pk_deduplication_theshold}"
+            )
+            logger.info(error)
             return AssetCheckResult(passed=False, metadata={"errors": error})
 
         return AssetCheckResult(passed=True)
 
-    return _check
+    return [_check_percent_unaccounted_for_gas, _check_pk_deduplication]
```
e-belfer, line 348: Right now these asset checks aren't loaded into dagster, at least partially because the function to produce them isn't called anywhere but possibly also because you're returning a list a checks here rather than just one check. If we aren't parameterizing the checks (e.g., specifying expected values by year) then looking at the way in which multiple checks are defined in `eia860.py` would be a better analogy than the FERC 714 model.

e-belfer (changes requested):
Right now the pandera schema check is failing when I run this locally, and I'm not totally sure why based on the metadata returned from the check failure.

Are you seeing the same issue on your end? I'd suggest getting this into alembic and written to the DB and then we can try to debug the problem, as we have some dtype enforcement built into the IO managers.

Otherwise, one recommendation about the other asset checks - which is to simplify them, as I think the way you've set this up won't work conceptually, seeing as the check gets called after the problematic data has already been dropped. Let me know if that all makes sense?

(As always, just let me know if you need any additional support on my end to get this over the line!)

e-belfer:
@seeess1 Thanks for all your work on this one! I'm closing this and picking up your work in #4005 to get this over the last few hurdles!

PR #3931, opened 2024-10-26
1357 convert unittest to pytest

### LINKED ISSUE #1357: Convert tests using unittest to pytest
Our codebase currently uses a mixture of `pytest` and `unittest` to run tests. In order to standardize, let's convert the remaining instances of `unittest` into `pytest`. To do this:

- locate all instances of `unittest`
- write these tests in equivalent `pytest` code. For example, switch from importing MagicMock to calling `mocker.MagicMock`, as we currently do in `test_retry()` in `pudl.test.unit.helpers_test.py`.
- remove imports of `unittest` package

`unittest` shows up in the following places in our codebase:

`$ git grep unittest`

```
test/unit/extract/csv_test.py:from unittest.mock import MagicMock, patch
test/unit/extract/excel_test.py:import unittest
test/unit/extract/excel_test.py:from unittest import mock as mock
test/unit/extract/excel_test.py:class TestMetadata(unittest.TestCase):
test/unit/extract/excel_test.py:class TestExtractor(unittest.TestCase):
test/unit/extract/phmsagas_test.py:from unittest.mock import MagicMock, patch
test/unit/output/ferc1_test.py:import unittest
test/unit/output/ferc1_test.py:class TestForestSetup(unittest.TestCase):
test/unit/transform/classes_test.py:from unittest.mock import MagicMock
test/unit/workspace/datastore_test.py:import unittest
test/unit/workspace/datastore_test.py:class TestDatapackageDescriptor(unittest.TestCase):
test/unit/workspace/datastore_test.py:class TestZenodoFetcher(unittest.TestCase):
test/unit/workspace/resource_cache_test.py:import unittest
test/unit/workspace/resource_cache_test.py:class TestGoogleCloudStorageCache(unittest.TestCase):
test/unit/workspace/resource_cache_test.py:class TestLocalFileCache(unittest.TestCase):
test/unit/workspace/resource_cache_test.py:class TestLayeredCache(unittest.TestCase):
```

ggurjar333:
# Overview

Closes #1357.

Replacing Unittest with Pytest to standardizing the testing code.
## What did you change?
I have changed Unittest based code with Pytest
# Documentation

```[tasklist]
```

# Testing

```[tasklist]
# To-do list
```


zaneselvans:
@ggurjar333 I think you'll want to install the git pre-commit hooks locally so that all the static code analysis and formatting checks can be run before anything gets pushed. Some pointers [in our docs here](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/dev_setup.html#set-up-code-linting)

ggurjar333:
> @ggurjar333 I think you'll want to install the git pre-commit hooks locally so that all the static code analysis and formatting checks can be run before anything gets pushed. Some pointers [in our docs here](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/dev_setup.html#set-up-code-linting)

I'm facing these errors.
{150E6F8C-AF51-4A8C-B4B2-679322C02081}

I have installed pre-commit 

{9939A25D-EE21-4667-9D7E-F49DB1D84C11}

I think it is messing around here somewhere.
{AA83BF4C-1313-41DC-9025-26B1D6199195}

How to fix this one?
{CF70B44F-383D-47FF-9E6C-310B3060A38F}

zaneselvans:
It looks like there are a bunch of genuine test failures on your branch, so I think we should try and separate the errors that are due to changes from the errors that are due to environment / setup issues.

* Were you able to successfully run the unit tests before making any changes?
* Check out the `main` branch and run `make pytest-unit` from the top-level of the repository.
* What kind of errors come up then?
* How did you create the `pudl-dev` conda environment?

ggurjar333:
> It looks like there are a bunch of genuine test failures on your branch, so I think we should try and separate the errors that are due to changes from the errors that are due to environment / setup issues.
> 
> * Were you able to successfully run the unit tests before making any changes?
> * Check out the `main` branch and run `make pytest-unit` from the top-level of the repository.
> * What kind of errors come up then?
> * How did you create the `pudl-dev` conda environment?

**Explanation of the issue** for the record here:
I am using WSL on Windows. The issue was I had mounted a volume from the Linux side, the disconnect may be what those two Windows vs Linux file paths difference, even if they were trying to actually pointing at the same files on disk. I was trying to commit from within PyCharm folder from linux (as the folder was on Windows) or some shell thing that PyCharm does it might see the Windows path, while ```pre-commit``` sees the Linux path.

Solution:
Zane helped me to understand and I moved the folder from Windows path to Linux path at the root directory. It Works!

test/unit/extract/excel_test.py
```diff
@@ -1,35 +1,42 @@
...
+import pytest
 
 from pudl.extract import excel
 
 
-class TestMetadata(unittest.TestCase):
+class TestMetadata:
     """Tests basic operation of the excel.Metadata object."""
 
+    @pytest.fixture(autouse=True)
```
jdangerx, line 12: This works, but the recommended way of implementing these xunit-style setup methods is to use `def setup_class`: https://docs.pytest.org/en/7.1.x/how-to/xunit_setup.html#class-level-setup-teardown

But, the even more recommended thing is to just turn this into a normal fixture called `metadata` and refer to that everywhere we need it.

test/unit/extract/phmsagas_test.py
```diff
@@ -8,20 +6,20 @@
 
 
 class FakeExtractor(Extractor):
-    def __init__(self):
+    def __init__(self, mocker):
```
jdangerx, line 9: See above non-blocking comment about passing in the actual dependencies vs. passing in `mocker`.

test/unit/extract/phmsagas_test.py
```diff
@@ -8,20 +6,20 @@
...
+def extractor(mocker):
     # Create an instance of the CsvExtractor class
-    return FakeExtractor()
+    return FakeExtractor(mocker)
 
 
-@patch("pudl.extract.phmsagas.logger")
-def test_process_renamed_drop_columns(mock_logger, extractor):
+def test_process_renamed_drop_columns(mocker, extractor):
+    mock_logger = mocker.patch("pudl.extract.phmsagas.logger")
```
jdangerx, line 22: Should we be using `with mocker.patch(...)` here and everywhere else with a naked `mocker.patch()`? This usage is not documented behavior of `patch`.

test/unit/output/ferc1_test.py
```diff
@@ -90,7 +87,9 @@ def build_forest_and_annotated_tags(
 
 
 class TestPrunnedNode(TestForestSetup):
-    def setUp(self):
+    @pytest.fixture(autouse=True)
```
jdangerx, line 90: See above comment about using a normal fixture instead of using autouse here.

test/unit/output/ferc1_test.py
```diff
@@ -133,7 +133,11 @@ def test_pruned_nodes(self):
 
 
 class TestTagPropagation(TestForestSetup):
-    def setUp(self):
+    """Test tag propagation functionality."""
+
+    @pytest.fixture(autouse=True)
```
jdangerx, line 138: See above re: autouse.

test/unit/output/ferc1_test.py
```diff
@@ -185,6 +189,7 @@ def setUp(self):
         )
 
     def test_leafward_prop_undecided_children(self):
+        """Test leadward propagation with undecided children."""
```
jdangerx, line 192: Were pre-commit checks yelling at you for not adding docstrings to these test methods? If so we should tell the check to ignore test docstrings.

jdangerx, line 192:
```suggestion
        """Test leafward propagation with undecided children."""
```

test/unit/workspace/datastore_test.py
```diff
@@ -243,7 +209,8 @@ class TestZenodoFetcher(unittest.TestCase):
         r"^10\.(5072|5281)/zenodo\.(\d+)$", PROD_EPACEMS_DOI
     ).group(2)
 
-    def setUp(self):
+    @pytest.fixture(autouse=True)
+    def setup(self):
```
jdangerx, line 213: Another flag for turning this into normal pytest fixtures.

test/unit/workspace/resource_cache_test.py
```diff
@@ -13,85 +12,85 @@
...
+        assert bad_request_predicate(requests_exceptions.Timeout())
+        assert bad_request_predicate(BadRequest(message="Bad request!"))
 
 
-class TestLocalFileCache(unittest.TestCase):
+class TestLocalFileCache:
     """Unit tests for the LocalFileCache class."""
 
-    def setUp(self):
+    def setup_method(self):
```
jdangerx, line 34: I think this setup/teardown should be turned into a fixture as well.

test/unit/workspace/resource_cache_test.py
```diff
@@ -13,85 +12,85 @@
...
-        self.assertTrue(ro_cache.contains(res))
+        assert ro_cache.contains(res)
 
 
-class TestLayeredCache(unittest.TestCase):
+class TestLayeredCache:
     """Unit tests for LayeredCache class."""
 
-    def setUp(self):
+    def setup_method(self):
```
jdangerx, line 93: Flagging another spot to fixture-ify!

jdangerx (changes requested):
Overall this looks much nicer when we're consistently using `pytest` - thank you! I do think that there are some spots where we're using the legacy setup / teardown behavior - while we're here converting everything to the new style, I think we should finish the job and convert those into pytest fixtures.

test/unit/output/ferc1_test.py
```diff
@@ -37,7 +36,104 @@
 logger = logging.getLogger(__name__)
 
 
-class TestForestSetup(unittest.TestCase):
+@pytest.fixture
+def forest_setup():
```
jdangerx, line 40: Pulling out just the *nodes* as a fixture doesn't help much, as you see you still have to unpack and add edges etc. The actual setup that has to get done for each test is "make a forest, with nodes, edges, and tags" - and each test has a slight variation on what those are. To me that means we actually want a helper function in this module that makes creating these forests a little less verbose, rather than a few fixtures.

But that involves a fair amount of restructuring of the tests, etc., and this PR has already gone on for a long time. I would just return to the xunit-style setup using the `setup_method` flow you use below.

jdangerx (changes requested):
Mostly looks good, I think we should revert back to the xunit-style setup for the FERC Form 1 forest test. Everything else is non-blocking. Thanks for working on this!

jdangerx:
I went ahead and made the small change I requested in #4014 . Thanks for doing all this work @ggurjar333 !

PR #3953, opened 2024-11-08
Refactor entity matching name cleaner to be more efficient

katie-lamb:
# Overview

As part of the SEC to EIA record linkage development, I had to make some changes to the PUDL company name cleaning module to make it more efficient and useful. The code for this module was originally pulled OS Climate's repo, but it was no longer maintained there. I didn't make significant changes when I pulled out that module, and thus it had some quirks and inefficiencies.  

* The name cleaner was very slow. It's still pretty slow on big datasets, but is about 3x faster than previously

## What did you change?
* Instead of using `apply` to apply the regex replacement rules, I used `pd.Series.replace` so that this replacement is vectorized.
* Removed some coupling in the cleaning rules and restructured the `CompanyNameCleaner` class
* Made some updates to the regex rules to be more effective

# Documentation

```[tasklist]
```

# Testing

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/analysis/record_linkage/name_cleaner.py
+++ b/src/pudl/analysis/record_linkage/name_cleaner.py
@@ -5,37 +5,33 @@ import json
 import logging
 import re
 from importlib.resources import files
-from typing import Literal
+from typing import Self
 
 import pandas as pd
-from pydantic import BaseModel
+from pydantic import BaseModel, Field, model_validator
 
 logger = logging.getLogger(__name__)
 
 CLEANING_RULES_DICT = {
     "remove_email": [" ", r"\S*@\S*\s?"],
     "remove_url": [" ", r"https*\S+"],
-    "remove_word_the_from_the_end": [" ", r"the$"],
-    "remove_word_the_from_the_beginning": [" ", r"^the"],
-    "place_word_the_at_the_beginning": [" ", r"the$"],
+    "remove_word_the_from_the_end": ["", r"\s+the$"],
+    "remove_word_the_from_the_beginning": ["", r"^the\s+"],
     "remove_www_address": [" ", r"https?://[.\w]{3,}|www.[.\w]{3,}"],
     "enforce_single_space_between_words": [" ", r"\s+"],
-    "replace_amperstand_by_AND": [" and ", r"&"],
-    "add_space_between_amperstand": [" & ", r"&"],
+    "replace_ampersand_by_AND": [" and ", r"\s*&\s*"],
+    "add_space_between_ampersand": [" & ", r"&"],
     "add_space_before_opening_parentheses": [" (", r"\("],
     "add_space_after_closing_parentheses": [") ", r"\)"],
-    "replace_amperstand_between_space_by_AND": [" and ", r"\s+&\s+"],
-    "replace_hyphen_by_space": [" ", r"-"],
-    "replace_hyphen_between_spaces_by_single_space": [" ", r"\s+-\s+"],
-    "replace_underscore_by_space": [" ", r"_"],
-    "replace_underscore_between_spaces_by_single_space": [" ", r"\s+_\s+"],
+    "replace_hyphen_by_space": [" ", r"\s*-\s*"],
+    "replace_underscore_by_space": [" ", r"\s*_\s*"],
     "remove_all_punctuation": [" ", r"([^\w\s])"],
     "remove_punctuation_except_dot": [" ", r"([^\w\s.])"],
     "remove_mentions": [" ", r"@\S+"],
     "remove_hashtags": [" ", r"#\S+"],
     "remove_numbers": [" ", r"\w*\d+\w*"],
-    "remove_text_puctuation": [" ", r'\;|\:|\,|\.|\?|\!|"'],
-    "remove_text_puctuation_except_dot": [" ", r'\;|\:|\,|\?|\!|"'],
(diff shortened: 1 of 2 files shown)


katie-lamb:
@zschira I'm not sure how to fix the docs error... I posted about this in Slack, but here's the problem:

The warning (treated as error) occurs [here](https://github.com/catalyst-cooperative/pudl/actions/runs/12129239890/job/33817270455?pr=3953#step:6:3641) and is about duplicate object descriptions for the class variables that I describe [here](https://github.com/catalyst-cooperative/pudl/blob/rl-cleaning-upates/src/pudl/analysis/record_linkage/name_cleaner.py#L101) in my CompanyNameCleaner class. When I look at the doc page that's generated by sphinx, it looks fine, although it lists all of these attributes twice, first with the full description and then with no description. Am I not doing the docstring for this class correctly? Is there something I should include so sphinx ignores this?

src/pudl/analysis/record_linkage/name_cleaner.py
```diff
@@ -56,69 +70,109 @@ class LegalTermLocation(enum.Enum):
...
+    )
+    with json_source.open() as json_file:
+        legal_terms_dict = json.load(json_file)[NAME_JSON_ENTRY_LEGAL_TERMS]["en"]
+    return legal_terms_dict
+
+
+class CompanyNameCleaner(BaseModel):
+    """Class to normalize/clean up text based company names.
+
+    Attributes:
```
bendnorman, line 101: I resolve the warning/error you need to move the attribute doc strings directly below the attribute definitions:

```python
cleaning_rules_list: list[str] = DEFAULT_CLEANING_RULES_LIST
    """A list of cleaning rules that the
    
    CompanyNameCleaner should apply. Will be validated to ensure
    rules comply to allowed cleaning functions."""
```

You can use this [pudl class](https://github.com/catalyst-cooperative/pudl/blob/c33a675d9931ef96ab8c58461ab920bba0dec5f9/src/pudl/metadata/classes.py#L328) as a reference. 

I think you're getting a duplicate warning because [sphinx is creating empty doc strings](https://github.com/sphinx-doc/sphinx/issues/8664#issuecomment-757087741) for the attributes and then it looks at the doc string of the class and creates doc strings for the attributes again.

katie-lamb:
Thanks for the help @bendnorman ! I think now it's just the same error that you mentioned was generally causing problems in PUDL?

src/pudl/analysis/record_linkage/name_cleaner.py
```diff
@@ -5,37 +5,33 @@
...
+    "replace_underscore_by_space": [" ", r"\s*_\s*"],
     "remove_all_punctuation": [" ", r"([^\w\s])"],
     "remove_punctuation_except_dot": [" ", r"([^\w\s.])"],
     "remove_mentions": [" ", r"@\S+"],
     "remove_hashtags": [" ", r"#\S+"],
     "remove_numbers": [" ", r"\w*\d+\w*"],
-    "remove_text_puctuation": [" ", r'\;|\:|\,|\.|\?|\!|"'],
-    "remove_text_puctuation_except_dot": [" ", r'\;|\:|\,|\?|\!|"'],
+    "remove_text_punctuation": ["", r'\;|\:|\,|\.|\?|\!|"|\''],
+    "remove_text_punctuation_except_dot": [" ", r'\;|\:|\,|\?|\!|"\''],
```
zschira, line 34: Should this replace with an empty string like the one above?

src/pudl/analysis/record_linkage/name_cleaner.py
```diff
@@ -133,61 +192,69 @@ def _apply_regex_rules(
...
+            remove_the_from_end_replacement,
+            regex=True,
+        )
+        clean_col = clean_col.where(~the_at_end_matches, "the " + clean_col)
+        return clean_col
 
-    def _apply_cleaning_rules(self, company_name: str) -> str:
+    def _apply_cleaning_rules(self, col: pd.Series) -> pd.Series:
         """Apply the cleaning rules from the dictionary of regex rules."""
+        if self.place_word_the_at_beginning:
```
zschira, line 234: Why is this rule handled differently? Is it a combination of two rules?

zschira (approved):
A couple of small questions, but overall looks good!

One random thing I started thinking about while looking at this is that maybe we were going about this the wrong way when trying to think up a way to avoid illegal states with string cleaning, and instead a better approach would be to just provide as much insight into what's happening as possible. It could be cool to try to turn this into a series of dagster `op`'s so you could see the order in the dagster UI and inspect intermediate outputs.

This is all just some random thoughts though, not something I think we should try to actually do right now.

src/pudl/analysis/record_linkage/name_cleaner.py
katie-lamb: yep, fixed

src/pudl/analysis/record_linkage/name_cleaner.py
katie-lamb: This has to do with the relationship between the `"remove_word_the_from_the_end"`, `"remove_word_the_from_the_beginning"`, and `place_word_the_at_beginning` rules. I decided that if you want to place the word the at the beginning, then you should do this first in case you want to then remove "the" from the end or "the" from the beginning. These rules are all kind of in conflict and feed into the idea that there are "states" that we go through - an `op` to handle "the" would make sense in a later refactor.

Issue #3926, opened 2024-10-23
Refactor VCE RARE asset checks


jdangerx:
```[tasklist]
### Tasks
- [x] Review asset checks and see if their memory footprint can be reduced
- [x] Remove any asset checks that seem low value and resource intensive
- [x] Accumulate an error report and return it if any of the piecemeal checks fail
- [x] Fix the expected row-count asset check so that it can pass in the fast ETL (#3993)
```

zaneselvans:
Most of these issues were addressed in #3959 but the row-count check still only passes if all data is processed, meaning the `fast_etl` will always fail if asset checks are enabled, which they should be.

Issue #3928, opened 2024-10-23
Make asset checks run during integration tests


zaneselvans:
## Overview

Asset check failures do not appear to get run or be treated as blocking when run as part of the integration tests.

I swear @jdangerx already tackled this in #3687 but it seems to be back from the dead?

I verified this by running the VCE RARE subset of the DAG using the `fast_etl` job in the Dagster UI (getting a failure) while the integration tests in the merge-queue requirements / local integration tests have been passing.

**Note:** The asset checks not running during tests is driving down our test coverage.  In #3959 there's a significant amount of asset checking code, and as a result it dropped the test coverage below the required threshold. The required threshold has been temporarily reduced, but this will only get worse as we refactor the data validation and bring them into the main ETL.

```[tasklist]
- [ ] Figure out why asset checks are not running in tests (again)
- [ ] Verify that the VCE row count check fails in the integration tests.
- [ ] Fix VCE row-counts check to close out #3926
- [ ] After it's fixed, bump required test coverage back up to 92.7%
```

jdangerx:
Looks like we have two failing asset checks when running `etl_fast` that block #3990 from getting merged:

* `core_ferc714__yearly_planning_area_demand_forecast` seems to fail the `_check` pandera schema check - @cmgosnell do you have bandwidth to look into this?
* `out_vcerare__hourly_available_capacity_factor`'s `check_rows` fails - that's already tracked in #3926 . @aesharpe or maybe @zschira do you have bandwidth for this one?

cmgosnell:
I can take a look at the 714 fail. I will try to get to it today but it might be tomorrow.

PR #3959, opened 2024-11-14
Reduce peak memory use of VCE RARE assets

### LINKED ISSUE #3925: Refactor VCE RARE transform to reduce peak memory usage

zschira:
# Overview

This PR refactors VCE RARE transforms and the `asset_check` to reduce peak memory usage.

Closes #3925

Makes progress on #3926 but row count asset check still fails on fast ETL.

## Approach
To reduce memory usage, I refactored the transforms to work on a single year of data at a time and write outputs to parquet files. This approach writes directly to parquet without an IO manager, which doesn't feel ideal, but it does work well to reduce memory usage.

To deal with asset checks, I refactored from loading the entire table into pandas to using duckdb to query the parquet outputs. This approach is also a bit messy, but is fast/efficient and we could probably build tooling around this approach and standardize it as a part of the validation framework design.

### Partitioned assets
I think using partitioned assets would be a good approach for this asset, as well as other resource intensive assets, but I believe that there are a couple reasons why that's not easy/feasible at this point. Mainly, it seems like it's best practice to maintain 1 consistent partitioning scheme per job. I think splitting the ETL into multiple jobs with partitioned assets could be a good pattern to adopt, but this felt like too big of a can of worms to tackle right now.

### Dynamic op graph
The other approach I investigated  was using dynamic op graphs to process years in parallel, but I found this could still lead to significant memory spikes depending on what all ends up running together and the unparallelized version doesn't take too long to run.

```[tasklist]
- [x] Reduce peak transform memory usage
- [x] Reduce peak `asset_check` memory usage
```


src/pudl/transform/vcerare.py
```diff
@@ -283,107 +289,232 @@ def out_vcerare__hourly_available_capacity_factor(
...
-)
-def check_hourly_available_cap_fac_table(asset_df: pd.DataFrame):  # noqa: C901
-    """Check that the final output table is as expected."""
+@asset
+def out_vcerare__hourly_available_capacity_factor(
+    raw_vcerare__lat_lon_fips: pd.DataFrame,
+    raw_vcerare__fixed_solar_pv_lat_upv: pd.DataFrame,
+    raw_vcerare__offshore_wind_power_140m: pd.DataFrame,
+    raw_vcerare__onshore_wind_power_100m: pd.DataFrame,
+):
```
zaneselvans, line 312: This only creates a set of single-year parquet outputs, rather than creating a single concatenated parquet output that conforms to the PyArrow schema defined for the table, so it seems like there's a missing concatenation step that's required to get the outputs into the same shape as all of our other outputs -- a single `table_name.parquet` file in the `PUDL_OUTPUT/parquet` directory, similar to the use of `pq.ParquetWriter` in `pudl.etl.epacems_assets`.

But instead of having the transform embedded within that function as an op graph, it seems like we can just iterate over the annual files and create pyarrow Tables out of them, and write them serially to the single output file with `pq.ParquetWriter` using the PyArrow schema defined for the table in the PUDL resource metadata.  Using the resource metadata defined PyArrow schema also stamps the table and column descriptions into the Parquet metadata directly.

src/pudl/transform/vcerare.py
```diff
@@ -283,107 +289,232 @@ def out_vcerare__hourly_available_capacity_factor(
...
-            },
-        )
+    columns = [c for c in vce.columns if c != "county_id_fips"]
+    for c in columns:
+        nulls = duckdb.query(f"SELECT {c} FROM vce WHERE {c} IS NULL").fetchall()  # noqa: S608
+        if len(nulls) > 0:
+            return AssetCheckResult(
+                passed=False,
+                description=f"Found NA values in column {c}",
+            )
```
zaneselvans, line 360: Rather than bailing as soon as we get any unexpected nulls, can we check all the columns, and tell the user what columns had unexpected nulls? Will make debugging less frustrating.

src/pudl/transform/vcerare.py
```diff
@@ -283,107 +289,232 @@ def out_vcerare__hourly_available_capacity_factor(
...
+
+def _check_wind_capacity_factor_upper_bound(vce) -> AssetCheckResult | None:
+    columns = [c for c in vce.columns if c.endswith("wind")]
+    for c in columns:
+        cap_oob = duckdb.query(f"SELECT {c} FROM vce WHERE {c} > 1.0").fetchall()  # noqa: S608
+        if len(cap_oob) > 0:  # noqa: S608
+            return AssetCheckResult(
+                passed=False,
+                description=f"Found wind capacity factor values greater than 1.0 in column {c}",
+            )
```
zaneselvans, line 388: As above, rather than fast failing, it would be nice to know all the columns that have problems after a single run.

src/pudl/transform/vcerare.py
```diff
@@ -283,107 +289,232 @@ def out_vcerare__hourly_available_capacity_factor(
...
-            description="Found capacity factor values less than 0",
-        )
+    columns = [c for c in vce.columns if c.startswith("capacity_factor")]
+    for c in columns:
+        cap_oob = duckdb.query(f"SELECT {c} FROM vce WHERE {c} < 0.0").fetchall()  # noqa: S608
+        if len(cap_oob) > 0:
+            return AssetCheckResult(
+                passed=False,
+                description=f"Found capacity factor values less than 0 from column {c}",
+            )
```
zaneselvans, line 401: As above, rather than fast failing, it would be nice to know all the columns that have problems after a single run.

src/pudl/transform/vcerare.py
```diff
@@ -283,107 +289,232 @@ def out_vcerare__hourly_available_capacity_factor(
...
+    return None
+
+
+def _check_max_hour_of_year(vce) -> AssetCheckResult | None:
     logger.info("Check max hour of year in VCE RARE table is 8760.")
-    if asset_df["hour_of_year"].max() != 8760:
+    max_hours = duckdb.query(
+        "SELECT hour_of_year FROM vce WHERE hour_of_year > 8760"
+    ).fetchall()
+    if len(max_hours) > 0:
```
zaneselvans, line 410: This is a slightly less specific test than the one it replaces. The max hour should always be exactly 8760 -- neither more nor less.

src/pudl/transform/vcerare.py
zaneselvans: Does the lower memory footprint of running the asset checks with DuckDB depend on the inputs being split into separate parquet files? Or will it still be memory efficient if it operates on a single concatenated parquet file that has annual row-groups?  Ideally we would want to validate the thing that we're actually going to distribute.

zaneselvans (changes requested):
One major issue -- this isn't producing a concatenated parquet file that corresponds to the output table we want to distribute, and it isn't using the PyArrow schema that is defined by the resource metadata for this output table.

Non blocking but if there's an easy way to not fast-fail the asset checks and provide more comprehensive feedback to users that would be nicer for debugging. See specific notes.

src/pudl/transform/vcerare.py
```diff
@@ -4,12 +4,18 @@
 in this module, as they have exactly the same structure.
 """
 
+import duckdb
```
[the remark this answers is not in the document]
zaneselvans: (I went ahead and made this change and merged in the changes from `main`)

src/pudl/etl/__init__.py
```diff
@@ -216,7 +216,7 @@ def _get_keys_from_assets(
     {
         "key": "memory-use",
         "value": "high",
-        "limit": 4,
+        "limit": 2,
```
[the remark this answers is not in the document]
zschira: Oh no this was from testing with op graphs and I meant to revert, fixed now

src/pudl/transform/vcerare.py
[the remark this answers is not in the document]
zschira: Good call, thanks!

src/pudl/transform/vcerare.py
```diff
@@ -283,107 +289,232 @@ def out_vcerare__hourly_available_capacity_factor(
...
+
+    if (error := _check_hour_from_date(vce)) is not None:
+        return error
+
+    if (error := _check_unexpected_counties(vce)) is not None:
+        return error
+
+    if (error := _check_duplicate_county_id_fips(vce)) is not None:
+        return error
+
```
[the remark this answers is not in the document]
zschira: I've split them into a bunch of individual checks, and tested running in parallel and memory usage looks fine

src/pudl/transform/vcerare.py
```diff
@@ -283,107 +289,232 @@ def out_vcerare__hourly_available_capacity_factor(
...
         .pipe(_combine_cap_fac_with_fips_df, fips_df)
         .sort_values(by=["state", "county_or_lake_name", "datetime_utc"])
         .reset_index(drop=True)
     )
+    parquet_path = _get_parquet_path()
+    parquet_path.mkdir(exist_ok=True)
+    df.to_parquet(
+        parquet_path / f"vcerare_hourly_available_capacity_factor-{year}",
+        index=False,
+    )
```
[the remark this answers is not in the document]
zschira: I've switched to creating a monolithic parquet file, and passing the schema during writes.

src/pudl/transform/vcerare.py
zschira: Fixed to find all null columns

src/pudl/transform/vcerare.py
zschira: Same

src/pudl/transform/vcerare.py
zschira: Good catch, fixed

zschira:
@zaneselvans I've updated to produce a single monolithic parquet file, which definitely makes more sense. I've also split the one big asset check into a bunch of little ones. Memory usage can get a little high when running them all in parallel, but it's not too bad. If we encounter issues running the full ETL we might want to mark them all as high-memory.

It looks like one remaining issue is a docs failure from adding the duckdb python dependency. Do you have any idea what's causing that?

src/pudl/transform/vcerare.py
zaneselvans: I ran the VCE RARE ETL locally and the asset checks were a few 100 MB each of memory use, so DuckDB seems to be doing great!

zaneselvans (approved):

zaneselvans:
- RTD docs build issue was because I'd used the conda rather than PyPI name for the package in `pyproject.toml`
- Removed a couple of leftover `| None` return types for the asset checks.
- Changed the nightly build machine specs back to 8CPU/64GB
- Running a branch deployment now to see if it'll actually run on the smaller machine, how long it takes, and whether the Zenodo sandbox is still b0rked.

zaneselvans:
Yay! It works on the smaller VM just fine.

zaneselvans:
Boo, the fact that asset checks aren't run in the integration tests means that we failed the coverage requirement.  I reduced it slightly since this really needs to be fixed in #3928.

PR #3976, opened 2024-11-26
Updates to splink FERC to EIA record linkage notebook

katie-lamb:
# Overview

This PR updates the `devtools` notebook for FERC to EIA record linkage with `splink`. This notebook isn't actually run in production, but it's helpful for visualizing the model weights and the decisions that go into making the match. 

The notebook now runs with the latest version of `splink` and reads in parquet inputs from S3 instead of from dagster storage.

## What did you change?

Updated the notebook to have more documentation cells and run with the latest version of splink. Also, fixed an error that was cropping up in the plant parts list creation with datetime types. I think this came up because I'm now reading in parquet inputs from S3 instead of dagster storage, and the `report_date` column was an object instead of a datetime.

# Documentation

```[tasklist]
```

# Testing

Run the FERC to EIA match notebook from top to bottom.

```[tasklist]
# To-do list
- [x] If updating analyses or data processing functions: make sure to update or write data validation tests (e.g. `test_minmax_rows()`).
```

### DIFF
--- a/src/pudl/analysis/plant_parts_eia.py
+++ b/src/pudl/analysis/plant_parts_eia.py
@@ -1483,10 +1483,14 @@ def match_to_single_plant_part(
     """
     # select only the plant-part records that we are trying to scale to
     ppe_part_df = ppe[ppe.plant_part == part_name]
+    if not pd.api.types.is_datetime64_any_dtype(ppe_part_df["report_date"]):
+        ppe_part_df["report_date"] = pd.to_datetime(ppe_part_df["report_date"])
+    if not pd.api.types.is_datetime64_any_dtype(multi_gran_df["report_date"]):
+        multi_gran_df["report_date"] = pd.to_datetime(multi_gran_df["report_date"])
     # convert the date to year start - this is necessary because the
     # depreciation data is often reported as EOY and the ppe is always SOY
     multi_gran_df.loc[:, "report_date"] = pd.to_datetime(
-        multi_gran_df.report_date.dt.year, format="%Y"
+        multi_gran_df["report_date"].dt.year, format="%Y"
     )
     out_dfs = []
     for merge_part in PLANT_PARTS:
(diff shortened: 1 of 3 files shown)


src/pudl/analysis/plant_parts_eia.py
```diff
@@ -1483,10 +1483,14 @@ def match_to_single_plant_part(
     """
     # select only the plant-part records that we are trying to scale to
     ppe_part_df = ppe[ppe.plant_part == part_name]
+    if not pd.api.types.is_datetime64_any_dtype(ppe_part_df["report_date"]):
+        ppe_part_df["report_date"] = pd.to_datetime(ppe_part_df["report_date"])
+    if not pd.api.types.is_datetime64_any_dtype(multi_gran_df["report_date"]):
+        multi_gran_df["report_date"] = pd.to_datetime(multi_gran_df["report_date"])
```
katie-lamb, line 1489: This is kind of janky but I think when inputs are read in from S3 (instead of from Dagster storage) the report date is an object not a datetime. I was getting errors, so I threw this in there, but it's a little sloppy.

zaneselvans: Should this be in the python module? Or a pre-processing step in the notebook? If the dataframe doesn't match the expected schema, then maybe it should fail?

When we read from Dagster storage if it's not a persisted asset it'll be a pickled dataframe, with the datetime type intact. If we're reading from a persisted asset, it'll be coming from Parquet, and IIRC our IOManager will enforce the correct PUDL dtypes when we read the data in. If you're reading a Parquet file from S3 and passing it in outside of the Dagster context, you might try using `dtype_backend="pyarrow"` to ensure that the datetime column gets parsed not as an object.

katie-lamb: Ah okay let me try using the pyarrow backend.

katie-lamb: The pyarrow backend introduced other type issues, so I updated to cast to a datetime within the notebook, after reading in the Parquet inputs.

zschira (approved):
Looks good!

PR #3988, opened 2024-12-08
Integrate EIA 860 multifuel table

### LINKED ISSUE #3438: Integrate EIA-860 Multifuel table
### Overview

3_5_Multifuel_Y#### is a table in EIA-860 that includes detailed information about multifuel generators. 

Why do you want it in PUDL?

Some of the columns in the multifuel table are in pudl from other EIA-860 tables, but there are columns in the multifuel table that are unique to that table (for example, the time to switch from oil to gas), and the data in this table seems more complete than the duplicate columns in other tables (for example, this table indicates that plant 10 can switch fuels while operating (`operating_switch = "Y"`, but the data for this plant currently integrated into pudl has this as a missing value. 

### Logistics

This is public and published alongside all of the other EIA-860 tables. Integrating may require reconciling overlapping columns from other EIA-860 tables if they disagree with each other. 

### What do you know about it so far?

It would be great to have this to ensure that `multiple_fuels`, 'cofire_fuels`, `switch_oil_gas`, and `operating_switch` are accurate and complete. 

It would also be helpful to have some new data such as the specific breakout of different cofire energy sources (as distinct from `energy_source_code` and `startup_source_code`); the time to switch columns, and the factors that limit switching columns.

jmelot:
# Overview

Closes #3438

## What did you change?

# Documentation

```[tasklist]
```

# Testing

```[tasklist]
# To-do list
```


e-belfer:
@jmelot Hello! Thanks for opening this PR, excited to get this in! I'm happy to answer any questions you have as you work on this, and review it when you reach a good pausing point. Just ping me!

jmelot:
Thanks @e-belfer, I appreciate it!

* I have been able to run the pipeline end to end and examine the multifuel outputs. I followed some of the other transformations I saw applied in `transform/eia860.py` but I doubt I got everything that's needed, could you please let me know what else would be useful? I'll also do another pass to make sure I got all the boolean columns and spot check the proposed column mapping.
* One thing I noticed was that the `operable` and `retired` multifuel tables have `Operating month` and `Operating year` columns while the `proposed` table has `Current month` and `Current year` tables which presumably have a different meaning. I wasn't sure whether it would be better to store the proposed values in the same `generator_operating_{month,year}` columns or keep them separate. For now, I've stored them in the same column.

aesharpe (changes requested):
# Summary of feedback
Thanks for your additions @jemelot! You did a nice job mimicking existing functions for other tables. I've added a few comments and suggested additions below!

## Regarding your comment about operating date:
> One thing I noticed was that the operable and retired multifuel tables have Operating month and Operating year columns while the proposed table has Current month and Current year tables which presumably have a different meaning. I wasn't sure whether it would be better to store the proposed values in the same generator_operating_{month,year} columns or keep them separate. For now, I've stored them in the same column.

In the proposed generators table, this "current" column refers to the current planned operating date and is referred to as `current_planned_generator_operating_month/year`. The generator table also has an `original_planned_generator_operating_month/year` column, so it makes sense to distinguish the two with "current" and "original" prefixes. Because this table doesn't have an "original" column, I could see us going either way: 

A) Keep the column name `generator_operating_month/year` and relying on the `operational_status` column to know whether it's planned or in effect.

B) Re-use the `current_planned_generator_operating_month/year` column name for consistency and extra specificity. 

@e-belfer or @cmgosnell what do you think? 

## Add table to harvested asset factory
Add this table to the `harvested_entity_asset_factory` in `src/pudl/transform/eia.py`. See the description of the `core` layer in our [docs](https://catalystcoop-pudl.readthedocs.io/en/latest/index.html#core-layer) for more detail about the harvesting process.

## Add table to `RESOURCE_METADATA`
Add the table schema and information to the `src/pudl/metadata/resources/eia860.py/RESOURCE_METADATA` dictionary.

## Update release notes
You'll want to add a description of your changes to our `docs/release_notes.rst` file

## Testing
This is more of a question for @e-belfer and @cmgosnell -- should we add this table to any of the testing infrastructure? E.g.: `test_minmax_rows` or `test_unique_rows_eia` in `eia_test.py`? It's not clear to me given the exclusion of the other generator subset tables (`scd_generators_solar` etc.) from both of these functions.

cmgosnell:
###  Regarding your comment about operating date:
* definitely think this looks and sounds like `current_planned_generator_operating_month`/year. great catch that these columns look different! So i think this clearly fits under your suggestion B) austne

cmgosnell:
### Re testing
All of the `test/validate/eia_test.py` stuff is all dependent on the `PudlTabl`, which we are deprecating. But rn most or all of the scd tables are included in these tests via `_register_output_methods`. and then the name of the values in that `_register_output_methods` dictionary are used in these validation tests.

We are planning on re-vamping our data validation process and depreciating `PudlTabl` ( an early example of that is using the similar tests for FERC1 in this pr #3860)


if you wanted to add new tables into `test_minmax_rows` I believe the steps would be:
1. add any new tables to `_register_output_methods`. keys are the asset names (which would be the `core_` table names) - value can be the same tbh the shortened names are only the maintain backwards compatibility with our old method names.
2. add these new tables with the parameters for `test_minmax_rows` with the counts you see now.
3. w/ or w/o adding new tables into the validation tests you'll definitely want to [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `make pytest-validate`

Also general comment, with all of these steps to fully fully integrate these new tables, please feel free to say "I'd like to pass this off" or "can this be out of scope" or "i'll need more support in order to make ___ happen" or whatever! The extract mappings and the transform function is already really helpful!

jmelot:
Thanks @aesharpe and @cmgosnell (and sorry to be slow getting back to you). On a quick skim, this makes sense. I'll have much more time next week, and I'll take a closer look and try to make the necessary edits before the new year.

src/pudl/metadata/resources/eia860.py
```diff
@@ -906,6 +906,73 @@
...
+                "fuel_switch_energy_source_1",
+                "fuel_switch_energy_source_2",
+                "fuel_switch_energy_source_3",
+                "fuel_switch_energy_source_4",
+                "fuel_switch_energy_source_5",
+                "fuel_switch_energy_source_6",
+            ],
+            "primary_key": [
+                "utility_id_eia",
+                "generator_id",
```
aesharpe, line 969: Did you double check that this is the primary key? My guess is that `plant_id_eia` might also be part of that / we usually add it in in case there are future instances where a utility owns multiple plants with a `generator_id` 1 for example.

aesharpe (commented):
@jmelot I left a little comment about boolean column naming, otherwise just need the release notes! 

I'll try and materialize the assets in a bit to see if it works. I'll also get back to you with more specific testing specifications in a moment.

jmelot:
Thanks, on it (and sorry this is dragging out a bit, the holidays were busier than I expected)! I think I was running into some kind of schema mismatch error when I last ran `etl_full` so after I've finished updating the boolean column names I'll give it another try and see if I can track that down.

src/pudl/metadata/resources/eia860.py
jmelot: Done!

jmelot:
I updated the boolean column names, added a short release note, and updated the primary key!

Two questions:

* When I was updating `src/pudl/metadata/fields.py` I felt like I was kind of guessing at some of the columns' meanings and on a few of them I was unsure enough that I just wrote "TK" for now. I looked around [here](https://www.eia.gov/electricity/data/eia860/) but didn't see any column-level documentation. Do you know of a good resource where I could look to see the meaning of these columns so I can improve the column descriptions?
* I tried rerunning the `etl_full` pipeline but it died with the error below. I made sure to update my database schema so I'm not sure what I'm doing wrong here - none of my other column renamings resulted in a similar error message. Does anything stand out to you?

Screenshot 2025-01-04 at 4 38 34 PM

aesharpe:
Hi @jmelot -- sorry for the delay, I was on vacation. 

> Do you know of a good resource where I could look to see the meaning of these columns so I can improve the column descriptions?

Yes, in the raw Zenodo archives for EIA 860 there is a Layout file containing column-level definitions!

> I tried rerunning the etl_full pipeline but it died with the error below. I made sure to update my database schema so I'm not sure what I'm doing wrong here - none of my other column renamings resulted in a similar error message. Does anything stand out to you?

As for this missing column error, I think the problem is in the `package_data`. You'll need to update the columns you manually mapped in the `column_maps` directory to match the new boolean column names. The schema can't find those columns because they are listed differently in the `RESOURCE_METADATA` (used for confirming the table schema) and manually mapped column data. Does that make sense? 

Also, when testing whether the table successfully materializes it'll be _must faster and easier_ to materialize the `core_eia860__scd_generators_multifuel` from the `core_eia` asset group and the `_core_eia860_generators_multifuel` from the `core_eia860` asset group as opposed to running the entire `etl_full`. If you search those assets in dagster you should be able to click on them and run them independently. If you update the column names, you'll also want to materialize the `raw_eia860_multifuel_` assets in the `raw_eia860` asset class before materializing the transformed tables. Let me know if you want some help with this!

src/pudl/metadata/resources/eia860.py
```diff
@@ -906,6 +906,75 @@
         "sources": ["eia860"],
         "etl_group": "eia860",
     },
+    "core_eia860__scd_generators_multifuel": {
+        "description": (
+            "Annually reported information about energy storage from EIA-860 Schedule 3. "
+            "Contains data on fuel-switching and the use of multiple fuels by surveyed generators."
```
aesharpe, line 912: I don't think the first sentence here is relevant to this table. As far as I understand it's about generators not energy storage entities. Maybe rework the sentence to say the following: 

```suggestion
        "description": (
            "Annually reported information from EIA-860 Schedule 3 about generators that use multiple fuel "
            " sources or engage in fuel switching. " 
```

src/pudl/metadata/fields.py
```diff
@@ -731,7 +747,31 @@
...
+        "type": "string",
+        "description": "Type of fuel.",
+    },
+    "cofire_energy_source_5": {
+        "type": "string",
+        "description": "Type of fuel.",
+    },
+    "cofire_energy_source_6": {
+        "type": "string",
+        "description": "Type of fuel.",
```
aesharpe, line 774: Should probably specify this description a little further to something like: "The codes representing the type of fuel that will be co-fired"

src/pudl/metadata/fields.py
```diff
@@ -449,6 +449,22 @@
...
+        "description": "Can the generator co-fire oil and gas?",
+    },
+    "can_fuel_switch": {
+        "type": "boolean",
+        "description": "TK",
+    },
+    "can_switch_when_operating": {
+        "type": "boolean",
+        "description": "Indicates whether the generator switches fuel while operating.",
+    },
```
aesharpe, line 467: I like the way you formatted the second description better than the first one. I.e., use "Indicates whether...." rather than framing the definition as a question.

aesharpe:
This PR is almost ready! I caught one dirty little secret while digging into the column descriptions, however. It looks like the `limit` columns have slightly different definitions over time. 

- `has_storage_limits`
- `has_air_permit_limits`
- `other_limits`

In 2012 and prior the definitions for these fields were measuring factors that limits the **operation of the generator when running on 100 percent oil** 

In 2013 and after the definitions for these fields measure factors that limits the generator's **ability to switch between oil and natural gas**

I emailed EIA to make sure that these were indeed different data points getting reported under the same column name.

**OPTIONS:**
- Update the definition for these columns to explain the difference pre and post 2012.
- Split each column into two columns, one for pre 2012 and one for post.

I think my decisions will be informed by whatever EIA says back, but I think either is fine. I'm inclined to keep them as the same columns and include the caveat in the column description (mostly because I hate columns with lots of NA values), but maybe that is bad data practice.

aesharpe:
EIA's response: 

> That change was an attempt to improve the meaningfulness of the question.  We originally sought to know why full capacity could not be reached while on the other fuel, usually a gas boiler switching over to oil, but I think we realized that we could ask the question a little better.  I don’ t recall that it had a large effect on the reported data.  By 2013, fewer gas boilers were switching to oil.  

As a result of this response, I am going to use the definitions from 2013+ and call it a day.

aesharpe (approved):

jmelot:
Thanks for your help getting this over the finish line (and throughout). And sorry for being so slow to reply... during the time this PR's been open I changed jobs and started planning my wedding, it has been an unexpectedly busy time. I should have communicated better! If it's ok, I will check back in in a few months when things settle down and see if I can pitch in on another issue (hopefully in a more timely manner...).

aesharpe:
@jmelot no worries at all; I was slow with my comments as well, so it's not your fault :) Happy to have you as a contributor! And wow, that's a lot going on--congrats on the new job and upcoming wedding 🍾 🍾

Issue #3985, opened 2024-12-04
readthedocs build failure


bendnorman:
Our readthedocs builds are failing because of a confusing dependency conflict. Full error:

```
warning  libmamba Could not parse mod/etag header
warning  libmamba Could not parse mod/etag header
Could not solve for environment specs
The following packages are incompatible
├─ cryptography ==44.0.0 py312hda17c39_0 is requested and can be installed;
└─ pyopenssl ==21.0.0 pyhd8ed1ab_0 is not installable because it requires
   └─ cryptography >=3.3,<39.0a0 , which conflicts with any installable versions previously reported.
```

- [Logs](https://app.readthedocs.org/projects/catalystcoop-pudl/builds/26465978/) from the most recent successful build
- Logs from [the build failure](https://app.readthedocs.org/projects/catalystcoop-pudl/builds/26487098/)

I'm confused because the `conda-linux-64.lock.yml` in the successful and failed builds both use `cryptography ==44.0.0` and `pyopenssl ==21.0.0`. The mamba versions are also the same.

I tried adding `cryptography<39.0a0` to `pyproject.toml` but got a slew of other dependency issues.

bendnorman:
Resolved by #3989

PR #3989, opened 2024-12-09
Update conda lockfile for week of 2024-12-09

pudlbot:

### DIFF
--- a/environments/conda-lock.yml
+++ b/environments/conda-lock.yml
@@ -128,39 +128,39 @@ package:
     category: main
     optional: false
   - name: aiohappyeyeballs
-    version: 2.4.3
+    version: 2.4.4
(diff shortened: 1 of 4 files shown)


environments/conda-lock.yml
```diff
@@ -3934,52 +3894,52 @@ package:
     category: main
     optional: false
   - name: cryptography
-    version: 44.0.0
+    version: 43.0.3
```
e-belfer, line 3897: @bendnorman Looks like a rollback of the package that was causing problems - maybe this will fix #3985?

bendnorman: Ah hopefully! I'll merge it in.

bendnorman (approved):

PR #3990, opened 2024-12-13
Make asset checks run in integration test

### LINKED ISSUE #3928: Make asset checks run during integration tests

jdangerx:
# Overview

Closes #3928 .

#3711 inadvertently disabled the asset checks again, in service of fixing `pudl_etl` 😞 

Getting rid of `pudl_etl` (#3161) seemed out of scope here, but I made the tests use the same `Definitions` as the nightly builds again.

# Testing

Made the pandera checks all fail, ran `make integration-test`, and checked to make sure that the integration tests bailed out once we ran into a failure.

Then got other people to fix the failing asset checks and re-ran integration tests. They passed.

```[tasklist]
# To-do list
- [x] Un-break the pandera checks
- [x] Fix the various asset checks that have failed over time through the CI not running them
```

### DIFF
--- a/src/pudl/etl/cli.py
+++ b/src/pudl/etl/cli.py
@@ -36,10 +36,11 @@ def pudl_etl_job_factory(
         The job definition to be executed.
     """
 
-    def get_pudl_etl_job():
+    def get_pudl_etl_job(job_name: str | None = None):
         """Create an pudl_etl_job wrapped by to be wrapped by reconstructable."""
         pudl.logging_helpers.configure_root_logger(logfile=logfile, loglevel=loglevel)
-        job_name = "etl_full_no_cems" if not process_epacems else "etl_full"
+        if job_name is None:
+            job_name = "etl_full_no_cems" if not process_epacems else "etl_full"
         return defs.get_job_def(job_name)
 
     return get_pudl_etl_job
--- a/src/pudl/etl/cli.py
+++ b/src/pudl/etl/cli.py
@@ -8,14 +8,13 @@ import click
 import fsspec
 from dagster import (
     DagsterInstance,
-    Definitions,
     JobDefinition,
     build_reconstructable_job,
-    define_asset_job,
     execute_job,
 )
 
 import pudl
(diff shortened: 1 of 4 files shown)


src/pudl/etl/cli.py
```diff
@@ -37,24 +36,12 @@ def pudl_etl_job_factory(
...
-                    selection=pudl.etl.create_non_cems_selection(
-                        pudl.etl.default_assets
-                    ),
-                )
-            ]
-        return Definitions(
-            assets=pudl.etl.default_assets,
-            resources=pudl.etl.default_resources,
-            jobs=jobs,
-        ).get_job_def("etl_job")
```
bendnorman, line 57: Oh I see... we redefined the definitions here but didn't add the asset checks so they weren't getting run. Importing [`defs` from `pudl.etl`](https://github.com/catalyst-cooperative/pudl/blob/23612af209a58c382fca7b424596c90563b36b1a/src/pudl/etl/__init__.py#L266-L269) includes the assets checks.

bendnorman (approved):
Yay! Thanks for fixing this.

PR #3993, opened 2024-12-19
Fix row count asset check for vcerare

zschira:
# Overview

This PR fixes the VCERARE row count `asset_check`. This check was using the expected row count for the full ETL, and would fail for the fast ETL. Now it handles both correctly.

### DIFF
--- a/src/pudl/transform/vcerare.py
+++ b/src/pudl/transform/vcerare.py
@@ -373,7 +373,7 @@ def check_rows(context: AssetCheckExecutionContext) -> AssetCheckResult:
     vce = _load_duckdb_table()  # noqa: F841
     (length,) = duckdb.query("SELECT COUNT(*) FROM vce").fetchone()
     if (
-        expecteded_length := row_counts[context.op_execution_context.job_name]
+        expected_length := row_counts[context.op_execution_context.job_name]
     ) != length:
         return AssetCheckResult(
             passed=False,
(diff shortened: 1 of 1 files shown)


src/pudl/transform/vcerare.py
```diff
@@ -359,16 +360,25 @@ def _load_duckdb_table():
...
+    row_counts = {
+        "etl_full": 136437000,
+        "etl_fast": 27287400,
+    }
+
     vce = _load_duckdb_table()  # noqa: F841
     (length,) = duckdb.query("SELECT COUNT(*) FROM vce").fetchone()
-    if length != 136437000:
+    if (
+        expecteded_length := row_counts[context.op_execution_context.job_name]
```
jdangerx, line 376:
```suggestion
        expected_length := row_counts[context.op_execution_context.job_name]
```

src/pudl/transform/vcerare.py
```diff
@@ -359,16 +360,25 @@ def _load_duckdb_table():
...
+    ) != length:
         return AssetCheckResult(
             passed=False,
             description="Table unexpected length",
-            metadata={"table_length": length, "expected_length": 136437000},
+            metadata={"table_length": length, "expected_length": expecteded_length},
```
jdangerx, line 381:
```suggestion
            metadata={"table_length": length, "expected_length": expected_length},
```

jdangerx (approved):
I believe you!

Also tiny typo fix. But otherwise good to go!

zschira:
lol expectedededed

PR #3996, opened 2024-12-27
Make the reconstructable job not take any arguments at all.

jdangerx:
# Overview

... the nightly builds are still broken because the reconstructable job doesn't want any parameters at all.

## What did you change?

Moved the parameters to the wrapper function.

# Documentation

# Testing

* [x] Run `make pytest-integration`
* [x] Run `pudl_etl --gcs-cache-path gs://internal-zenodo-cache.catalyst.coop ~/work/pudl/src/pudl/package_data/settings/etl_full.yml` - see if it even starts

### DIFF
--- a/src/pudl/transform/vcerare.py
+++ b/src/pudl/transform/vcerare.py
@@ -365,6 +365,8 @@ def check_rows(context: AssetCheckExecutionContext) -> AssetCheckResult:
     logger.info("Check VCE RARE hourly table is the expected length")
 
     # Define row counts for fast/full etl
+    # TODO 2024-12-27: make this check row counts per year instead of having
+    # two different counts based on job name - less brittle.
     row_counts = {
         "etl_full": 136437000,
         "etl_fast": 27287400,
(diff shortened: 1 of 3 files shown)


src/pudl/transform/vcerare.py
```diff
@@ -365,6 +365,8 @@ def check_rows(context: AssetCheckExecutionContext) -> AssetCheckResult:
     logger.info("Check VCE RARE hourly table is the expected length")
 
     # Define row counts for fast/full etl
+    # TODO 2024-12-27: make this check row counts per year instead of having
```
e-belfer, line 368: Agree, but we'll probably eventually wind up with distribution tests that do want fast vs. full specified and can't be parameterized by year, so the solution will probably come in handy anyways?

e-belfer (approved):
Assuming this runs as expected, this seems to be a reasonable reconfiguration to me. Left a note about the parameterization by year (tldr - I agree but we might need this kind of implementation for other tests).

PR #4003, opened 2025-01-07
Make it possible to pass another `sources` dict to DataSource

e-belfer:
# Overview

Addresses a lightweight version of https://github.com/catalyst-cooperative/pudl-archiver/issues/499 in concert with https://github.com/catalyst-cooperative/pudl-archiver/pull/506.

In the `pudl-archiver` repository, we want to make it possible to archive datasets that will never end up in the PUDL repository. Eventually, we'll want to fully separate these two repositories, which raises bigger questions about where the metadata should be defined, etc. But for now, we want to make it possible to add a dictionary of `NON_PUDL_SOURCES` to the `pudl-archiver` repository, to be called in a method similar to `from_pudl_metadata` in `frictionless.py`.

## What did you change?
This method imports the `DataSource.from_id()` method from PUDL, which currently hardcodes `SOURCES` into the method. In order to make it possible to provide an alternative dictionary of source metadata into this method, I have added a `sources` parameter to this method which defaults to `SOURCES`, but can take a different SOURCE dictionary.

# Documentation

```[tasklist]
```

# Testing

```[tasklist]
# To-do list
```

### DIFF
--- a/src/pudl/metadata/classes.py
+++ b/src/pudl/metadata/classes.py
@@ -1035,20 +1035,22 @@ class DataSource(PudlMeta):
             sys.stdout.write(rendered)
 
     @classmethod
-    def from_field_namespace(cls, x: str) -> list["DataSource"]:
+    def from_field_namespace(
+        cls, x: str, sources: dict[str, Any] = SOURCES
+    ) -> list["DataSource"]:
         """Return list of DataSource objects by field namespace."""
         return [
             cls(**cls.dict_from_id(name))
-            for name, val in SOURCES.items()
+            for name, val in sources.items()
             if val.get("field_namespace") == x
         ]
 
     @staticmethod
-    def dict_from_id(x: str) -> dict:
+    def dict_from_id(x: str, sources: dict[str, Any] = SOURCES) -> dict:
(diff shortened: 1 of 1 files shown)


src/pudl/metadata/classes.py
```diff
@@ -1035,20 +1035,22 @@ def to_rst(
             sys.stdout.write(rendered)
 
     @classmethod
-    def from_field_namespace(cls, x: str) -> list["DataSource"]:
+    def from_field_namespace(
```
e-belfer, line 1038: This doesn't get called in `pudl-archiver`, so I could revert this change. Just figured for consistency it'd make sense to have the same behavior as below.

src/pudl/metadata/classes.py
```diff
@@ -1285,6 +1287,7 @@ class Resource(PudlMeta):
             "pudl",
             "nrelatb",
             "vcerare",
+            "phmsagas",
```
e-belfer, line 1290: Oops! This crossed over from another branch, I'll remove this.

zschira (approved):
This seems like a reasonable/simple solution to avoid cluttering our main metadata collection

Issue #3971, opened 2024-11-19
Data validation project design conversation


jdangerx:
### Overview

We need to have a plan before we jump into this data validation project. Plus the plan has to fit roughly into 75 hours, less the design time (and needs to take review / adjustments into account!)

### Success criteria:
* [x] short document that lays out the **product design** of the system: what specific use cases are we solving for? what hard things are we not going to do? how does the user (us, as developers) interact with the new data validation library? 
* [x] short document laying out **technical design** - what components come together to present the high level interface from the product design? how do they interact with each other / what operations do they expose? what data structures do they exchange?

### Next steps:
* [x] get people to complain about the data validation system
* [ ] decide who needs to "approve" the product design
* [ ] synthesize the complaints into the product design
* [ ] get approval
* [ ] figure out approvers, work out technical design, get approval there too

At the very minimum we should pick @zaneselvans 's brain about his complaints before he leaves, and ideally we should get some input from him in the product design synthesis part.

zschira:
## Background: 
I was moving forward with the design/prototyping of our validation framework based on great expectations upon discovering that there are dependency conflicts with PUDL stemming from numpy/pandas.

## Possible solutions:
### Downgrade dependencies-
We could downgrade several dependencies to make great expectations installable. The following dependencies would need to be downgraded:
- pandas < 2.2
- numpy < 2.0
- splink < 4

### Use an alternate tool-
#### [Soda](https://github.com/sodadata/soda-core/tree/main)-
I was initially hesitant about Soda because it seemed really coupled to it’s paid ecosystem. However, after spending more time with other tools and diving deeper into their docs, I think that there’s a decent amount of functionality that we could get out of the core library

Pros:
- duckdb support
- Checks are highly configurable
- Good support for custom sql checks

Cons:
- Not available on `conda-forge`. We could probably help them contribute a feedstock, but this would be a blocker to adding `soda` as a dependency

#### Pandera-
I'm including `pandera` since it's a tool we already use and have looked at before, but I don't think it's particularly well suited to our data validation needs.

Pros:
- Really nice expressive API
- Useful integration with
- Supports custom checks

Cons:
- Limited support for checks involving multiple columns
- Little control over when/how checks are executed
- No support for multi-table checks

### Create a separate environment for data validation
[...]

PR #4011, opened 2025-01-13
Dbt setup

### LINKED ISSUE #3997: Port 2-3 logically complicated validations to dbt
## Success criteria
* [ ] Convert basic validation like bounds checks
* [ ] Convert high memory validations (vcerare)
* [ ] Convert eia923 `vs_historical` test
* [ ] Fully setup dbt project in basic `pudl-dev` environment install

## Not in scope
* connecting dbt source test to dagster

## Next steps
* [x] setup multiple profiles for testing nightly builds/etl full/etl fast
* [x] get dbt set up with dbt-utils, dbt-expectations
* [ ] port those dang tests

zschira:
### Overview
This PR sets up a `dbt` project within the PUDL repo that will be used for data testing. Details on setup and usage can all be found in the `Readme.md`. This PR also includes several data validations that have been converted to `dbt` tests. The tests currently converted are all `vcerare` `asset_check`s and FERC fuel by plant cost per mmbtu range checks.

### In scope
- [x] Standardize organization and naming conventions
- [x] Develop some basic macros that will enable migrating existing tests
    - [x] Weighted quantile
    - [x] Review `validate.py` for any other obvious candidates
- [x] Auto-generate row count checks
- [x] Update `readme` with directions for migration (how to add new asset, types of test in scope for immediate migration)

### Out of scope
- [ ] Make tests run in asset checks
- [ ] Plan alternative to complex tests like the existing `vs_historical`
- [ ] Migrate tests

closes #3997 #3971


src/pudl/dbt/models/schema.yml
```diff
@@ -0,0 +1,90 @@
...
+    meta:
+      external_location: |
+        {%- if target.name == "nightly" -%} 'https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/{name}.parquet'
+        {%- else -%} '{{ env_var('PUDL_OUTPUT') }}/parquet/{name}.parquet'
+        {%- endif -%}
+    tables:
+      - name: out_eia923__boiler_fuel
+      - name: out_eia923__monthly_boiler_fuel
+      - name: out_ferc1__yearly_steam_plants_fuel_by_plant_sched402
+      - name: out_vcerare__hourly_available_capacity_factor
```
zaneselvans, line 14: In a data warehouse with hundreds of tables, would this file be created and managed by hand? Or would there be some rule-based way to generate it, or parts of it, along the lines of what we're doing with the Pandera schema checks right now? For example, the `not_null` tests here are a 2nd place that that restriction is being specified -- it's already present in our table metadata, which seems like recipe for them getting out of sync.

Or in the case of row counts, is there a clean, non-manual way to update the row counts to reflect whatever the currently observed counts are?  Especially if we're trying to regenerate expected row counts for each individual year, filling it all in manually could be pretty tedious and error prone.  We've moved toward specifying per-year row counts on the newer assets so that they work transparently in either the fast or full ETL cases, and the asset checks don't need to be aware of which kind of job they're being run in, which seems both more specific and more robust.

src/pudl/dbt/models/schema.yml
```diff
@@ -0,0 +1,90 @@
+version: 2
+
+sources:
+  - name: pudl
+    meta:
+      external_location: |
+        {%- if target.name == "nightly" -%} 'https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/{name}.parquet'
```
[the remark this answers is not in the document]
zaneselvans: Interestingly, with the `s3://` URL it didn't give me any error, but it also didn't seem to be making much progress. There was just a ton of data being downloaded. Not sure why.

src/pudl/dbt/models/validation/ferc1_fbp_cost_per_mmbtu.sql
zaneselvans, line 1: If we do end up needing to define these intermediate tables it seems like we would want to have some kind of clear naming convention for them?

src/pudl/dbt/models/schema.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: out_ferc1__yearly_steam_plants_fuel_by_plant_sched402
+      - name: out_vcerare__hourly_available_capacity_factor
+        data_tests:
+          - dbt_expectations.expect_table_row_count_to_equal:
+              value: |
+                {%- if target.name == "etl-fast" -%} 27287400
+                {%- else -%} 136437000
+                {%- endif -%}
+          - dbt_expectations.expect_compound_columns_to_be_unique:
+              column_list: ["county_id_fips", "datetime_utc"]
```
zaneselvans, line 22: Could be generated based on the PK that's defined for every table?

src/pudl/dbt/models/schema.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: out_eia923__boiler_fuel
+      - name: out_eia923__monthly_boiler_fuel
+      - name: out_ferc1__yearly_steam_plants_fuel_by_plant_sched402
+      - name: out_vcerare__hourly_available_capacity_factor
+        data_tests:
+          - dbt_expectations.expect_table_row_count_to_equal:
+              value: |
+                {%- if target.name == "etl-fast" -%} 27287400
+                {%- else -%} 136437000
+                {%- endif -%}
```
zaneselvans, line 20: Is there a clean way to specify the expected row counts for each year of data (or some other meaningful subset) within a table, as we've started doing for the newer assets in Dagster asset checks, so we don't have to differentiate between fast and full validations, and can identify where the changes are?

src/pudl/dbt/models/schema.yml
```diff
@@ -0,0 +1,90 @@
...
+              - dbt_expectations.expect_column_values_to_not_be_in_set:
+                  value_set: ["bedford_city", "clifton_forge_city"]
+models:
+  - name: ferc1_fbp_cost_per_mmbtu
+    columns:
+      - name: gas_cost_per_mmbtu
+        data_tests:
+          - dbt_expectations.expect_column_quantile_values_to_be_between:
+              quantile: 0.05
+              min_value: 1.5
```
zaneselvans, line 62: I'm guessing these are not using the weighted quantiles?

src/pudl/dbt/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### Adding new tables
+To add a new table to the project, you must add it as a
+[dbt source](https://docs.getdbt.com/docs/build/sources). You can do this by editing
+the file `src/pudl/dbt/models/schema.yml`. I've already added the table
+`out_vcerare__hourly_available_capacity_factor`, which can be used as a reference.
+
+#### Adding tests
+Once a table is included as a `source`, you can add tests for the table. You can
+either add a generic test directly in `src/pudl/dbt/models/schema.yml`, or create
```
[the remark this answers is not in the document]
jdangerx: Pretty sure you can just shove more YAMLs into `/models` and DBT will just pick them up... here's a [guide for how to structure DBT projects](https://docs.getdbt.com/best-practices/how-we-structure/1-guide-overview).

src/pudl/dbt/models/schema.yml
jdangerx: Looks like the "X column is not null" checks are currently defined in `fields.py` under the field constraints, is that what you're thinking about?

I think it would be nice to have auto-generated tests like the non-null tests & row counts defined alongside manually added tests. Then all the tests will be defined in one place, except for the tests that we need to write custom Python code for.

That seems pretty doable - YAML is easy to work with, and dbt lets us tag tests, so we could easily tag all the auto-generated tests so our generation scripts know to replace them but leave the manually-added tests alone.

zaneselvans: In addition to the field specific constraints I think we automatically add `NOT NULL` check constraints to the PK fields when we construct the SQLite database -- but more generally I'm just saying that we need to get all of these generated tests integrated non-duplicatively into the `dbt` tests somehow.

src/pudl/dbt/.user.yml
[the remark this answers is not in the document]
zschira: Probably not, removed.

src/pudl/dbt/.gitignore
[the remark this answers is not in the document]
zschira: That makes sense, I've moved the `dbt` directory to the top-level now

src/pudl/dbt/models/schema.yml
zschira: It seems totally possible to auto-generate tests, but I think there's also probably many ways to do accomplish this, so we should figure out what we want from it. For example, when we talk about auto-generating row count/not null tests, will these be generated once and committed into the repo, or will some/all of them be dynamically generated at runtime?

It definitely seems tricky to minimize duplication between `dbt`/our existing python schema info. I also wonder how this plays into any refactoring of our metadata system?

src/pudl/dbt/models/validation/ferc1_fbp_cost_per_mmbtu.sql
zschira: Yeah I think that seems like a good idea. Maybe just use a `validation_` prefix and otherwise follow existing naming conventions?

src/pudl/dbt/models/schema.yml
zschira: We'd probably need to create a custom macro for this, but that seems totally doable. Big question is how we want to generate/store all of those tests.

src/pudl/dbt/models/schema.yml
zschira: Yeah this are just basic quantiles. It's not too hard to get a `sql` query that can do a version of weighted quantiles, but the existing `vs_historical` tests are hard because they're computing a bunch of quantiles, then comparing them all

src/pudl/dbt/models/schema.yml
zaneselvans: It feels like we may need to clearly define the data tests that are ready to be migrated in a straightforward way, and the things that still need design work, so we can point Margay folks at the stuff that's ready to go and keep thinking about the things that still need some scaffolding?

src/pudl/dbt/models/schema.yml
zaneselvans: The row count tests have functionally become regression tests -- we want to know when they change, and verify that the magnitude and nature of the change is expected based on the code or data that we've changed. Given that there are hundreds of tables (and thousands of table-years) it doesn't seem practical to hand-code all of the expected row counts. 

It would be nice to have the per table-year row counts stored in (say) YAML somewhere, and be able to generate a new version of that file based on current ETL outputs. Then we could look at the diffs between the old and the new versions of the file when trying to assess changes in the lengths of the outputs.

dbt/README.md
```diff
@@ -0,0 +1,97 @@
...
+then run the following command from this directory.
+
+```
+dbt deps
+```
+
+### Adding new tables
+To add a new table to the project, you must add it as a
+[dbt source](https://docs.getdbt.com/docs/build/sources). The standard way to do
+this is to create a new file `models/{data_source}/{table_name}.yml`. If the the
```
marianneke, line 22: @zschira I am confused by the difference between this instruction that tells users to add the table schema as an individual `yml` file with the table name and the instructions below for adding tests, where you mention a `schema.yml` file that would presumably contain all models.

devtools/dbt_helper.py
```diff
@@ -0,0 +1,267 @@
+"""A basic CLI to autogenerate dbt yml."""
```
zschira, line 1: Here's the script to add a new table to the project. It will generate a file called `dbt/models/{data_source}/{table_name}/schema.yml` that defines the basic schema for a table for dbt and adds a test for row counts per partition. I've added some documentation for usage in `dbt/README.md`.

We should decide if there's any other tests we want to automatically add based on our metadata at this point.

dbt/macros/weighted_quantile.sql
```diff
@@ -0,0 +1,26 @@
+{% macro weighted_quantile(model, column_name, weight_col, quantile) %}
```
zschira, line 1: @marianneke this is the macro that's currently broken. It gets called in `expect_column_weighted_quantile_values_to_be_between.sql`. The idea is that this will compute a weighted quantile on a table, then that test will check the quantile vs bounds. I think the issue is with the syntax using `WITH` here. The macro `dbt/macros/check_row_counts_per_partition.sql` has an example where I fixed this to use the proper duckdb syntax.

dbt/macros/check_row_counts_per_partition.sql
```diff
@@ -0,0 +1,18 @@
+{% test check_row_counts_per_partition(model, table_name, partition_column) %}
```
zschira, line 1: Compares row counts in the table to expected counts which are stored in `seeds/row_counts.csv`

zschira (commented):
I think we're close to being able to convert all the existing tests using the `validate.py` framework. It seems like there are currently 3 categories of tests in this framework:

- `vs_historical`/`vs_self` tests, which we've decided are out-of-scope and could be replaced with simpler tests. @zaneselvans we should decide what to do with these tests
- Bounds checks on un-weighted quantiles. These are immediately convertable using `dbt_expectations`
- Bounds checks on weighted quantiles. These can be converted if we get the `weighted_quantile.sql` macro working

Beyond this, the script I've added to generate scaffolding for new tables will add row count checks per partition. If there's any other checks we want to autogenerate from our metadata we should decide on that now.

dbt/models/eia923/out_eia923__boiler_fuel/schema.yml
```diff
@@ -0,0 +1,28 @@
+version: 2
```
zschira, line 1: @zaneselvans do you know why `out_eia923__boiler_fuel` isn't using the `monthly` naming convention?

dbt/macros/weighted_quantile.sql
marianneke: Can you clarify what the syntax issue was and how you solved it in the row count macro?

zschira: Hmm I'm actually not sure that it was the issue after all. I was getting a similar error message error message with the row count macro and I did have a syntax error there, but this looks ok to me. It might also be something to do with the `expect_column_weighted_quantile_values_to_be_between` macro which is calling this under the hood.

dbt/models/eia923/out_eia923__boiler_fuel/schema.yml
zaneselvans: There are 3 `out_eia923__*_boiler_fuel` tables -- this one hasn't been aggregated by month, so there are probably some cases in which there are multiple records for a single month that get combined in the `__monthly` version. But I'd need to compare the monthly and unaggregated tables to be sure. I know there's at least one case in the generation tables where the monthly and unaggragated tables are literally identical.

devtools/dbt_helper.py
```diff
@@ -0,0 +1,296 @@
...
+
+    if table_name in row_counts_df["table_name"].to_numpy() and not clobber:
+        return AddTableResult(
+            success=False,
+            message=f"There are already row counts for table {table_name} in row counts table and clobber is not set.",
+        )
+
+    # Load table of interest
+    if not use_local_tables:
+        df = pd.read_parquet(_get_nightly_url(table_name))
```
zaneselvans, line 126: Rather than downloading the entire parquet file to get the row counts per partition (which will mean several GB every time this is run on all tables) can we potentially send a query that only has to access the Parquet metadata? That would require almost no data transfer, but might mean we need to change how the parquet files are indexed and possibly use `s3://` rather than `https://` URLs to access the nightly outputs.

dbt/seeds/row_counts.csv
zaneselvans, line 1: I mentioned this in our call but I'm nervous that calculating row counts based on the contents of the tables may sometimes result in different values that we get based on restricting input data to certain years, which we need to work -- since one of the main reasons we need per-partition counts is so that the integration tests / fast ETL can succeed even when only a subset of the years of data are available in the DB.

We should check that the partitions and the restricted datasets always give the same number of rows at least for the case of the fast ETL before we merge this in.

dbt/seeds/row_counts.csv
```diff
@@ -0,0 +1,270 @@
...
+out_ferc1__yearly_steam_plants_fuel_by_plant_sched402,2020,716
+out_ferc1__yearly_steam_plants_fuel_by_plant_sched402,2021,751
+out_ferc1__yearly_steam_plants_fuel_by_plant_sched402,2022,765
+out_ferc1__yearly_steam_plants_fuel_by_plant_sched402,2023,759
+out_eia923__boiler_fuel,2008-01-01,6389
+out_eia923__boiler_fuel,2008-02-01,6358
+out_eia923__boiler_fuel,2008-03-01,6435
+out_eia923__boiler_fuel,2008-04-01,6443
+out_eia923__boiler_fuel,2008-05-01,6455
+out_eia923__boiler_fuel,2008-06-01,6457
```
zaneselvans, line 42: Monthly partitions will provide a lot of granular information about where changes have happened, but using only unique values of a single column seems like it may not be very adaptable, e.g. in hourly tables, or tables where we need to select on more than one column to get well meaningful chunks, and the number of groups we're tracking in this system will be totally unmanageable by hand -- it'll have to be entirely automatable (with manual review / approval).

If we want to get much more granular data regression / diff testing than what we've been doing with the row counts, then we may want to revisit the more generalized database diff work that @rousik was doing before he vanished, especially now that he's back.

environments/conda-linux-64.lock.yml
```diff
@@ -208,13 +221,13 @@ dependencies:
   - identify=2.6.7=pyhd8ed1ab_0
   - idna=3.10=pyhd8ed1ab_1
   - imagesize=1.4.1=pyhd8ed1ab_0
-  - importlib-metadata=8.6.1=pyha770c72_0
+  - importlib-metadata=6.10.0=pyha770c72_0
```
zaneselvans, line 224: Whew that's a big backrev.

dbt/models/ferc1/out_ferc1__yearly_steam_plants_fuel_by_plant_sched402/schema.yml
```diff
@@ -0,0 +1,67 @@
...
+          - name: primary_fuel_by_mmbtu
+          - name: waste_fraction_cost
+          - name: waste_fraction_mmbtu
+
+models:
+  - name: validate_ferc1__yearly_steam_plants_fuel_by_plant_sched402
+    columns:
+      - name: gas_cost_per_mmbtu
+        data_tests:
+          - dbt_expectations.expect_column_quantile_values_to_be_between:
```
zaneselvans, line 37: My intuition is that these should probably really be weighted quantiles, or that they are relatively low-value checks because the FERC fuel reporting is such a mess.

zaneselvans (commented):
There's more SQL than I was hoping we'd need to have. It makes sense for the data tests / macros that we're defining for ourselves, but it seems like there's a fair amount that's also required to do the calculations / create ephemeral tables for the tests to run on which seems like it could be a source of friction given our generally moderate level of SQL knowledge.

zschira:
@zaneselvans I did some testing to look at the row counts per partition between the fast ETL and the full ETL. I also tested this partitioning by most granular time column in a table and partitioning by year. 
[Here](https://github.com/user-attachments/files/18872880/mismatched_counts_date_partition.csv) is a CSV of row count mismatches when partitioning by the full date, which found 131 cases across all tables we produce parquet outputs for (this also excluding thousands of mismatches found for the table  `out_gridpathratoolkit__hourly_available_capacity_factor`). [Here](https://github.com/user-attachments/files/18872953/mismatched_counts_yearly_partition.csv) are all of the mismatches when partitioning by year, which found 91 cases including the gridpath table.

Given that partitioning by year does seem more stable, I've switched the row count tests, plus the script to generate them, to this approach. However, as we can see it doesn't solve all of the problems, so I've also modified things so there is a separate count of fast/full etl row counts, which will be selectively used based on the `dbt` `target` you specify when running tests. This isn't ideal since we now have to maintain and update two separate row count tests, but I think if we make sure both counts are checked frequently during development it hopefully won't be too much of a burden.

As for the proliferation of SQL, `dbt` does allow [python-models](https://docs.getdbt.com/docs/build/python-models), however I haven't really played with this option and I'm not sure if there would be complications with managing a separate python environment used by `dbt` (maybe we could just get it to use the `pudl-dev` environment?). Another option could be to create intermediate tables in dagster and store those tables somewhere `dbt` could access them. However, I think there are also several advantages to using SQL including:

1. The tables are stored as SQL views so we don't end up with more disk usage
2. Producing these intermediate tables is very fast/memory efficient when using `duckdb`
3. If we get better at SQL we could potentially start incorporating `dbt`/`duckdb` directly into our ETL for high memory/compute transformations.

dbt/seeds/row_counts.csv
zschira: See comment below. I've switched to using yearly partitions.

devtools/dbt_helper.py
zschira: I changed this to use `duckdb` to query the nightly parquet files and I've observed significantly reduced network usage when running the script, so I don't think it's downloading the full table anymore.

dbt/README.md
```diff
@@ -0,0 +1,121 @@
...
+### Usage
+There are a few ways to execute tests. To run all tests with a single command:
+
+```
+dbt build
+```
+
+This command will first run any models, then execute all tests.
+
+For more finegrained control, first run:
```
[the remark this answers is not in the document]
zschira: I've simplified this section a bit. I think the most straightforward option for running tests with this setup is to always use the `dbt build` command with the `--select` option.

dbt/README.md
```diff
@@ -0,0 +1,121 @@
...
+for guidance on how to reference a `source` from a SQL file.
+
+The second method is to create a [model](https://docs.getdbt.com/docs/build/models)
+which will produce the intermediate table you want to execute tests on. To use this
+approach, simply add a sql file to `dbt/models/{data_source}/{table_name}/`.
+Now, add a SQL file to this directory named `validate_{table_name}` and define your model
+for producing the intermediate table here. Finally, add the model to the `schema.yml` file
+and define tests exactly as you would for a `source` table. See
+`models/ferc1/out_ferc1__yearly_steam_plants_fuel_by_plant_sched402` for an example of this
+pattern.
```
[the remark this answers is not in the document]
zschira: It gets stored as a SQL view. I've added a note about this in the README.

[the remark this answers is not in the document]
marianneke: @zaneselvans dbt allows you to choose how models are materialized. As @zschira mentioned, the default materialization is as a view, but you can also choose others, a common one being "table".

dbt/models/ferc1/out_ferc1__yearly_steam_plants_fuel_by_plant_sched402/schema.yml
```diff
@@ -0,0 +1,67 @@
...
+          - name: nuclear_fraction_mmbtu
+          - name: oil_fraction_cost
+          - name: oil_fraction_mmbtu
+          - name: primary_fuel_by_cost
+          - name: primary_fuel_by_mmbtu
+          - name: waste_fraction_cost
+          - name: waste_fraction_mmbtu
+
+models:
+  - name: validate_ferc1__yearly_steam_plants_fuel_by_plant_sched402
```
zaneselvans, line 33: In this case where we've created an ephemeral derived table (model) to run a test against, this information is all compiled by hand, right?  And `dbt` just knows to automatically look for a `.sql` file with the same name as the model to use that to create the table? Or the `.sql` is picked up by `dbt build` and that instantiates the model I guess.

dbt/profiles.yml
```diff
@@ -0,0 +1,17 @@
+pudl_dbt:
+  outputs:
+    # Define targets for nightly builds, and local ETL full/fast
+    # See models/schema.yml for further configuration
```
zaneselvans, line 4: I don't think there's a `models/schema.yml` any more is there?

dbt/profiles.yml
```diff
@@ -0,0 +1,17 @@
...
+    nightly:
+      type: duckdb
+      path: "{{ env_var('PUDL_OUTPUT') }}/pudl.duckdb"
```
zaneselvans, line 7: Putting a file named `pudl.duckdb` in the `PUDL_OUTPUT` directory alongside `pudl.sqlite` implies that this is the PUDL database, but in DuckDB format, which isn't the case and will be confusing -- this DB will only end up holding the models, seeds, and views we define, right?

dbt/profiles.yml
```diff
@@ -0,0 +1,17 @@
+pudl_dbt:
```
zaneselvans, line 1: It seems like by default dbt only uses 1 thread when running the tests, but using more definitely sped it up considerably for me. Is there some way can have it use whatever the number of available CPUs is? Or will this be getting run in a script where we can set it with a command line argument? How will this work once we get it into our asset checks?

Also wow, it ran all the tests locally in 12 seconds with 8 threads. Even referencing the nightly outputs it only took 2 minutes.

zaneselvans (commented):
Overall I think this looks great, insofar as I understand it, and it is insanely fast compared to what we have had for ages.  All the tests ran in like 12 seconds for me on 8 cores.

I think better documentation of the multiple ways the helper script is meant to be used would be helpful, including gotchas that could result in accidentally overwriting manually compiled information in the schema files.  I still feel unclear on what the process is going to look like for updating all of this information when we change the schema of a table, or need to update the row counts periodically if we are accumulating manually compiled tests in these files.

Many of the docstrings and help messages are very terse and I don't think provide enough context / detail.

I think the README should be migrated into our development documentation. Some of it is only relevant to the migration but a lot of it will be more durable and all the docs should be in the same place.

SImilarly I think we should try and keep as much of our code as possible under `src/pudl` where it can be part of the tests -- especially if it contains significant logic, which the dbt helper script seems to. And it'll probably only grow over time.  We should probably create a `scripts` directory in there the `devtools` directory is just an invitation to create code outside of our other infrastructure.

dbt/README.md
```diff
@@ -235,3 +235,70 @@ This produces the output:
 
 This is failing because the `max_value` is set to `0.7`. If we change this value
 to 0.83, this test should now pass.
+
+#### `no_null_cols` tests
```
zaneselvans, line 239: I think this test is run on a very large number of columns currently. It might be a good candidate for auto-generation, or being turned into a table-level test that by default checks all columns, but allows certain columns to be excluded (since in the `etl_fast` outputs columns that were only reported in older historical versions of the data are expected to be entirely null in modern data)

zschira:
@zaneselvans I've added some updated documentation, let me know if there's anything else needed before merge!

I'll also continue to work through your individual comments in the PR and we can collect next steps as issues/tasks in this epic I've created: #4075.

zaneselvans (changes requested):
Maybe I'm confused but it seems like there's an error in the definition of `ALL_TABLES`?

Plus a few documentation clarifications. I tried to add suggestions where I *though* I knew what it meant? But I'm not sure.

devtools/dbt_helper.py
```diff
@@ -0,0 +1,503 @@
...
+        if quantile_key in test_config
+    ]
+
+
+@click.command
+@click.option("--table-name", type=str, help="Name of table test will be applied to.")
+@click.option(
+    "--test-config-name",
+    type=str,
+    help="Name variable containing test configuration in `pudl.validate`.",
```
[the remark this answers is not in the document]
zschira: Added

devtools/dbt_helper.py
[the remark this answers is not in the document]
zschira: Fixed now

devtools/dbt_helper.py
```diff
@@ -0,0 +1,503 @@
...
+
+from pudl import validate
+from pudl.logging_helpers import configure_root_logger, get_logger
+from pudl.metadata.classes import PUDL_PACKAGE
+from pudl.workspace.setup import PudlPaths
+
+configure_root_logger()
+logger = get_logger(__file__)
+
+ALL_TABLES = [r.name for r in PUDL_PACKAGE.resources if r.name.startswith("_")]
```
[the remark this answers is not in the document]
zschira: Oh whoops no that was a temporary thing I was testing and didn't mean to commit. Fixed!

zaneselvans (approved):

PR #4019, opened 2025-01-17
Integrate the Census-based FIPS codes to replace `addfips`

### LINKED ISSUE #3884: Fix FIPS coding in PUDL
In #3531 we've learned that the `addfips` package we rely on to map state/county data into a FIPS code is irregularly maintained. This is causing problems for downstream users, and we should consider deprecating this package and moving to a different solution.

There are a few different use cases for FIPS encoding that we could hypothetically imagine for PUDL.
Right now, `add_fips_ids` in `pudl.src.helpers.py` is the only place we import the `addfips` package. This method is expects a county and state name, and a FIPS vintage:

```
def add_fips_ids(
    df: pd.DataFrame,
    state_col: str = "state",
    county_col: str = "county",
    vintage: int = 2015,
) -> pd.DataFrame:
    """Add State and County FIPS IDs to a dataframe.

    To just add State FIPS IDs, make county_col = None.
    """
    # force the columns to be the nullable string types so we have a consistent
    # null value to filter out before feeding to addfips
    df = df.astype({state_col: pd.StringDtype()})
    if county_col:
        df = df.astype({county_col: pd.StringDtype()})
    af = addfips.AddFIPS(vintage=vintage)
    # Lookup the state and county FIPS IDs and add them to the dataframe:
    df["state_id_fips"] = df.apply(
        lambda x: (
            af.get_state_fips(state=x[state_col]) if pd.notnull(x[state_col]) else pd.NA
        ),
        axis=1,
    )

    # force the code columns to be nullable strings - the leading zeros are
    # important
    df = df.astype({"state_id_fips": pd.StringDtype()})

    logger.info(
        f"Assigned state FIPS codes for "
        f"{len(df[df.state_id_fips.notnull()])/len(df):.2%} of records."
    )
    if county_col:
        df["county_id_fips"] = df.apply(
            lambda x: (
                af.get_county_fips(state=x[state_col], county=x[county_col])
                if pd.notnull(x[county_col]) and pd.notnull(x[state_col])
                else pd.NA
            ),
            axis=1,
        )
        # force the code columns to be nullable strings - the leading zeros are
        # important
        df = df.astype({"county_id_fips": pd.StringDtype()})
        logger.info(
            f"Assigned county FIPS codes for "
            f"{len(df[df.county_id_fips.notnull()])/len(df):.2%} of records."
        )
    return df
```

#### County and state names -> FIPS Code
This is our current FIPS encoding use case. We could:

- create a fork of `addfips`, and implement the fixes we want to see that solve the problem identified in #3531. We'd relatively easily benefit from any upstream changes, and if the project returns to being more actively maintained it wouldn't be particularly complex to switch back.
- write our own method to convert state/county codes to FIPS codes drawing on the `addfips` methodology. As is noted in #338 and below, the slight string cleaning implemented in `addfips` is definitely of relevance to us, so we'd want to replicate it at minimum but it shouldn't require heavy lifting to recreate.
- explore a maximalist version, which involves encoding FIPS from addresses or lat/lon coordinates rather than county/state. However, this function is currently used for EIA 860, 861 and 923 data, and at least 861 and 923 data _don't_ have complete addresses or lat/lon coordinates, so this likely won't meet our needs. The `pygris` [package](https://walker-data.com/pygris/03-data-utilities/) seems like a great API-free option for implementing this functionality down the line, though!

#### Address -> Lat lon -> FIPS Code or lat/lon -> FIPS code
As mentioned above, this wouldn't address the use case for 2/3 EIA datasets we're currently using the `addfips` package for. See  `pygris` [package](https://walker-data.com/pygris/03-data-utilities/), which seems interesting if we want to explore this further, but this is out of scope for the actual problem we're trying to resolve. See also #3531 for some discussion on other ways to implement this.

### Proposed solutions

1. Fork `addfips` and implement our desired changes, then do our own release so we can import this as a package.
Pros: We don't have to write much code, and we can keep up to date with any changes made in `addfips`.
Cons: Extremely annoying and we're likely to forget about this. We're still relying on a largely unmaintained package.

2. Recreate the logic of `addfips` in `pudl.helpers.py` using Census data directly (e.g., [this 2023 data](https://www.census.gov/geographies/reference-files/2023/demo/popest/2023-fips.html) or by calling the [Census's geometry API](https://api.census.gov/data/2023/geoinfo?get=NAME&for=county:*), see `[here](https://www.census.gov/data/developers/data-sets/geo-info.html)` for more info).
Pros: We are our own maintainers.
Cons: We have to recreate the fuzzy-matching logic in the package we're currently using to prevent performance loss.

3. Find some other new Python package to meet our needs.

- `us` is maintained but doesn't yet support counties despite an open PR on this topic from 2017: https://github.com/unitedstates/python-us/. The `census` [package](https://github.com/datamade/census) depends on it, so there's no county support there either.
- I have yet to find another comparable alternative

```
- [x] Add census metadata into `SOURCES` in pudl #4006
- [x] Add archiver https://github.com/catalyst-cooperative/pudl-archiver/pull/507
- [ ] Add extraction & light transform step for census fips codes #4019
- [ ] Recreate the logic of `addfips` in `pudl.helpers.py`
```

cmgosnell:
# Overview

Closes #3884.

## Questions
- [x] should we publish a transformed table as a `_core` or `core` table?? if `core` we may want to break it out into multiple tables bc really its many geographies all squished into one table
- [x] rn i've called this table "geocodes" because that's what census calls it. sound okay?
- [x] should i add in a lil manually compiled non-us state codes that addfips was including?

# Documentation

```[tasklist]
```

# Testing

```[tasklist]
# To-do list
```


krivard (commented):
Re:Q1 - are any of the areas both a subdivision and a place/place and consolidated city/subdivision and consolidated city, or can we file them as an "aux code" "aux type" pair? if they're mutually exclusive then we could do either:
* one table, with a variable-length concatenated FIPS (3 chars for country, 5 for state, 8 for county, 13 for aux)
* 3 tables (country could just be hard coded): state, county, aux

Re:Q2 - 👍 "geocodes"

cmgosnell:
hm @krivard  I've been assuming we want to update the vintage of the fips codes (the `addfips` package's most recent data is from 2015). But I've already found one problem with this update. CT! all the county names have been.. idk boring-afied

image

image

cmgosnell:
okay since this revelation about different years having different info in there i decided to incorporate 2015 and 2023. I'm purposefully rn _not_ merging the code using `report_date`. because even in this example the `report_date` from census don't line up. but now there are both options in the data. this seems to be working as intended.

cmgosnell:
okay i also added in 2009 bc there were a small handful of counties that were mapped w/ addfips that weren't being mapped without an older year.

rn the county codes that don't match are only two and they are bc of [name changes](https://en.wikipedia.org/wiki/Oglala_Lakota_County,_South_Dakota):
image

there are also 13 non-us state records that are not mapped
image

src/pudl/helpers.py
```diff
@@ -190,52 +190,133 @@ def download_zip_url(
...
+        r"è": "e",
+        r"à": "a",
+        r"ì": "i",
+        r"å": "a",
+    }
+    abbrevs = {"ft. ": "fort ", "st. ": "saint ", "ste. ": "sainte "}
+
+    county_types = {
+        r" (county|city|city and borough|borough|census area|municipio|municipality|district|parish)$": ""
+    }
```
cmgosnell, line 268: i liberally copy + pasted this from [here](https://github.com/fitnr/addfips/blob/c7f96a2b4906c97219d0003080b9779b2ae464cd/src/addfips/addfips.py#L28)

src/pudl/transform/censuspep.py
cmgosnell, line 1: right now i'm using/not doing anything with this shell _core asset. i think there are two possible things to do here:
* [ ] move this here `county_id_fips=lambda x: x.state_id_fips + x.county_id_fips` ?
* [ ] add a stored version of these missing guys/any bespoke fixes needed
image

src/pudl/helpers.py
```diff
@@ -190,52 +190,133 @@ def download_zip_url(
...
+        ["state_id_fips", "county_tmp"], keep=False
+    ) & ~(counties.area_name.str.lower().str.contains("city"))
+
+    if len(city_county_dupes := counties[city_county_dupe_mask]) > 10:
+        raise AssertionError(
+            "We expect there to be 10 county records that have the same name and state "
+            "but have different county fips codes. This is due to pair counties that have "
+            "a city jurisdiction and a corresponding county jurisdiction (ex. Baltimore "
+            "County and Baltimore City which is also a county). We found "
+            f"{len(city_county_dupes)}:\n{city_county_dupes}"
```
cmgosnell, line 306: lol jk i can't choose these cities across the board. there are only a handful of them but i will need to specify which to choose.

krivard (commented):
- Threading stuff through Dagster is hard
- Oh no US Territories. Would it be useful to figure out what % of our data touches US territories not included in the 50 state and Puerto Rico?

src/pudl/helpers.py
```diff
@@ -190,52 +190,133 @@ def download_zip_url(
...
-            f"Assigned county FIPS codes for "
-            f"{len(df[df.county_id_fips.notnull()]) / len(df):.2%} of records."
+    return df
+
+
+def add_county_fips_id(
+    df: pd.DataFrame, geocodes: pd.DataFrame, county_col: str
+) -> pd.DataFrame:
+    """Add the County FIPS codes to a table with State FIPS codes."""
+    diacretics = {
```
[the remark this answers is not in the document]
cmgosnell: oh no sorries needed ty!

src/pudl/helpers.py
```diff
@@ -190,52 +190,133 @@ def download_zip_url(
...
+def add_state_id_fips(
+    df: pd.DataFrame, geocodes: pd.DataFrame, state_col: str
+) -> pd.DataFrame:
+    """Add the State FIPS codes.
+
+    First we need to merge in the state abbreviation to the fips codes
 
+    Note: There are state fips codes in this pudl-compiled POLITICAL_SUBDIVISIONS
+    going with using the census based codes instead of our compiled codes even though
+    I checked they are all the same. Still feels better to sue a more og source.
```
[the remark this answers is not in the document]
cmgosnell: lol i know! but ty for this typo catch. imma change it to be a "primary source" as well bc ya know

src/pudl/extract/censuspep.py
```diff
@@ -0,0 +1,61 @@
...
+        Args:
+            ds (:class:datastore.Datastore): Initialized datastore.
+        """
+        self.METADATA = excel.ExcelMetadata("censuspep")
+        self.cols_added = []
+        super().__init__(*args, **kwargs)
+
+    @staticmethod
+    def get_dtypes(page, **partition):
+        """Returns dtypes for plant id columns and county FIPS column."""
```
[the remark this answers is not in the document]
cmgosnell: oh oops i meant to push an update to this yesterday that says something very similar. so good catch!

src/pudl/transform/eia860.py
```diff
@@ -1209,7 +1209,7 @@ def _core_eia860__boiler_stack_flue(
 
 @asset(io_manager_key="pudl_io_manager")
 def _core_eia860__cooling_equipment(
-    raw_eia860__cooling_equipment: pd.DataFrame,
+    raw_eia860__cooling_equipment: pd.DataFrame, raw_censuspep__geocodes: pd.DataFrame
```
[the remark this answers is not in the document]
cmgosnell: yea i was wondering if i can add an asset as a dependency to a function that is not an asset but instead is run while generation other assets. @bendnorman or @jdangerx do you know if there is a way to do this?

src/pudl/transform/censuspep.py
cmgosnell: I found this page when trying to find a good census source to archive but couldn't find a good way to download it in bulk or an archive on their part of past versions.

cmgosnell: hm okay wow that was easier than expected... i found where to download the [2020 codes](https://www2.census.gov/geo/docs/reference/codes2020/). but it doesn't have past versions which is not ideal in terms of replacing the fips codes from where I'm currently pulling them. but does include the non-states+PR. hm.

krivard: How far back do we need to go? I found the bulk source files for 2010 and 2020.

(feel free to ignore if that meant "oh god please no this one's almost done")

krivard: 2010 vintage is in a slightly different format but still has the needed content: https://www2.census.gov/geo/docs/reference/codes/files/national_county.txt

cmgosnell: Hm. Perhaps if I had found these before I would have used these instead. Although the version that I am using seems to have way more updates.

I think for getting this over the hump I'd like to just generate a small CSV of these non-state/non-PR records, store it in `package_data` and append it to the existing data. There just really aren't that many records here and yes as you noted this is currently almost done (at least it feels like it lol)

src/pudl/transform/eia860.py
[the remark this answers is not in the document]
cmgosnell: oh yea i don't think the way it is rn is incorrect. and i do think its natural and good for it to be clear that this table actually depends on the geocodes. it'd be nice in a way to have a default asset input into the helper function so we wouldn't have to insert it as an argument in all of the places add fips is being called but idk it is more explicit/clear this way.

src/pudl/package_data/censuspep/territories.csv
```diff
@@ -0,0 +1,21 @@
+state_id_fips,county_id_fips,area_name,fips_level,report_date
```
krivard, line 1: Nonblocking: Could you commit the/a script for generating this file? That way it's easier to update if we need to (perhaps in 6 years; the ANSI files had updates for the last two census-es so it seems likely they'll have one for 2030)

krivard (approved):
Fabulous. Two documentation nits and an optional helper script (assuming we're all still here and writing Python for the 2030 census)

test/unit/helpers_test.py
```diff
@@ -853,3 +854,49 @@ def test_generate_rolling_avg():
...
+Baltimore city,MD
+Baltimore,MD
+Bedford County,VA
+Bedford,VA
+Bedford city,VA"""
+        ),
+        dtype=pd.StringDtype(),
+    )
+    # Note: I couldn't find real instances of area's called "Bedford City"
+    # which is good because right now the ex
```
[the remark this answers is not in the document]
cmgosnell: this is classic me... thank you for catching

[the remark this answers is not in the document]
cmgosnell: ahha this is a thing that _was_ wrong in the way i setup the test and then i fixed it. so imma delete.

src/pudl/package_data/censuspep/territories.csv
cmgosnell: good idea! I did this pretty manually bc there were so few of em. but i added details about that process into the transform function doc string

cmgosnell:
oh nice okay i ran the validation tests before actually fixing the Bedford City vs County problem.... and then the row counts changed! because previously Bedford City was getting dropped because `addfips` wasn't labeling Bedford City and we were `dropna`-ing on the fips codes. so i am going to update the row counts!

src/pudl/transform/eia861.py
```diff
@@ -1050,18 +1051,10 @@ def core_eia861__yearly_service_territory(
...
         .pipe(_post_process)
     )
-    # The Virgin Islands and Guam aren't covered by addfips but they have FIPS:
-    st_croix = (df.state == "VI") & (df.county.isin(["St. Croix", "Saint Croix"]))
-    df.loc[st_croix, "county_id_fips"] = "78010"
-    st_john = (df.state == "VI") & (df.county.isin(["St. John", "Saint John"]))
-    df.loc[st_john, "county_id_fips"] = "78020"
-    st_thomas = (df.state == "VI") & (df.county.isin(["St. Thomas", "Saint Thomas"]))
-    df.loc[st_thomas, "county_id_fips"] = "78030"
-    df.loc[df.state == "GU", "county_id_fips"] = "66010"
```
cmgosnell, line 1064: i didn't realize this was here until after i added these guys into the CSV. that feels like a better place for them anyway so i delete this guy.

src/pudl/helpers.py
```diff
@@ -292,7 +292,8 @@ def _clean_area_name_col(area_name_col: pd.Series, replace_dict: dict):
     counties = counties[~city_county_dupe_mask]
 
     df = (
-        df.assign(county_tmp=_clean_area_name_col(df[county_col], {}))
+        df.astype({county_col: pd.StringDtype()})
+        .assign(county_tmp=lambda x: _clean_area_name_col(x[county_col], {}))
```
krivard, line 296: Checking understanding: this change is for two reasons:

1. we needed `county_col` as a string, and sometimes it wasn't
2. `_clean_area_name_col` couldn't handle a Series so we have to run it one row at a time

if yes, all good, if no, help?

krivard (approved):
Assuming I read that right, good to go!

src/pudl/helpers.py
zaneselvans: I thought assign with a lambda function / any callable that takes a series still does its work vectorized. Is that not true?

PR #4026, opened 2025-01-24
Enable sec10k distribution

zschira:
# Overview

This PR enables the distribution of SEC10k and other future datasets produced by what will be the `pudl-modelling` repo. This repo is currently called [mozilla_sec_eia](https://github.com/catalyst-cooperative/mozilla-sec-eia) and has been used to develop models for extracting company ownership data from SEC10k filings and linking this data to EIA data. This work lives separately from the main PUDL repo because training and running the models is highly resource intensive and would not interact well with the our current nightly build setup. It also allows us to isolate the complex dependencies from PUDL.

Interacting with this data is fairly straightforward on the PUDL side as all that is required is to load a table and schema information from cloud storage, then write this data alongside the rest of the PUDL data. Given that outside contributors won't have credentials to access cloud resources, I've added an environment variable `USE_PUDL_MODELS` that will be checked before ever trying to load the any PUDL models data. This environment variable needs to be set for testing and nightly builds, but for most developers can be ignored.


docker/Dockerfile
```diff
@@ -42,6 +42,7 @@ ENV CONTAINER_PUDL_WORKSPACE=${CONTAINER_HOME}/pudl_work
 ENV PUDL_INPUT=${CONTAINER_PUDL_WORKSPACE}/input
 ENV PUDL_OUTPUT=${CONTAINER_PUDL_WORKSPACE}/output
 ENV DAGSTER_HOME=${CONTAINER_PUDL_WORKSPACE}/dagster_home
+ENV USE_PUDL_MODELS=True
```
zschira, line 45: Set environment variable to load PUDL models data during nightly builds.

src/pudl/etl/__init__.py
```diff
@@ -107,14 +108,17 @@
...
+        itertools.chain.from_iterable(
+            load_assets_from_modules(
+                modules,
+                group_name=group_name,
+            )
+            for group_name, modules in all_asset_modules.items()
         )
-        for group_name, modules in all_asset_modules.items()
     )
+    + get_pudl_models_assets()
```
zschira, line 121: Append assets to load pudl models tables to default set of assets. If `USE_PUDL_MODELS` is not set, then no assets will be added.

src/pudl/io_managers.py
```diff
@@ -322,13 +322,21 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
 class PudlParquetIOManager(IOManager):
     """IOManager that writes pudl tables to pyarrow parquet files."""
 
+    def _get_table_resource(self, table_name: str) -> Resource:
```
zschira, line 325: This handles metadata for PUDL models tables slightly differently from all other tables.

src/pudl/metadata/classes.py
```diff
@@ -572,6 +573,24 @@ class Field(PudlMeta):
     harvest: FieldHarvest = FieldHarvest()
     encoder: Encoder | None = None
 
+    @classmethod
```
zschira, line 576: PUDL models schemas are stored as `pyarrow` schemas, which are converted to our standard metadata structures here.

katie-lamb (approved):
This looks good!

we probably don't need to generate the assets in every nightly build if it's resource intensive but good for now!

PR #4035, opened 2025-01-28
Add sec10k metadata directly in PUDL

zschira:
Add metadata for sec10k tables directly to PUDL to solve read-the-docs build issues with pulling metadata from gcs, and make the `deltalake` dependency unnecessary, which was behaving weirdly.


katie-lamb (approved):
Approved

zaneselvans:
I think you may need to set the `USE_PUDL_MODELS` envvar on the GitHub runner to get the metadata unit tests to pass -- it's finding a bunch of fields which are not being used because the SEC tables aren't included in any of the resources we've defined.

src/pudl/analysis/pudl_models.py
```diff
@@ -3,12 +3,10 @@
 import os
 
 import pandas as pd
-import pyarrow as pa
 from dagster import AssetsDefinition, asset
-from deltalake import DeltaTable
```
zaneselvans, line 8: If `deltalake` is no longer being directly imported anywhere then we should remove it from the requirements in `pyproject.toml`.

src/pudl/metadata/fields.py
```diff
@@ -331,6 +331,14 @@
...
         ),
     },
+    "block": {
+        "type": "string",
+        "description": "Title of block of data.",
+    },
+    "block_count": {
+        "type": "integer",
+        "description": "Some blocks are repeated, `block_count` defines the index of the data block.",
+    },
```
zaneselvans, line 341: What is a data block? More context in these descriptions would be helpful. Is `block` a specific enough column name?

I think the single backticks will probably not render correctly, since that's markdown formatting, and this will be getting parse as RST in the documentation.

src/pudl/metadata/fields.py
```diff
@@ -998,6 +1018,14 @@
         "description": "Code identifying a dataset available within PUDL.",
         "constraints": {"enum": list(SOURCES)},
     },
+    "date_filed": {
+        "type": "datetime",
+        "description": "Date filing was submitted.",
+    },
```
zaneselvans, line 1024: Do we already have a field name that means this, maybe in the context of the FERC XBRL data, which could be re-used here rather than creating a new field name?

src/pudl/metadata/fields.py
```diff
@@ -1611,6 +1643,18 @@
         "type": "number",
         "description": "Total number of flue gas desulfurization unit scrubber trains.",
     },
+    "former_conformed_name": {
+        "type": "string",
+        "description": "Former name of the company.",
+    },
```
zaneselvans, line 1649: By analogy with `company_name` and `company_name_raw` might this be clearer as `company_name_former`? What does "conformed" mean in this context?

src/pudl/metadata/fields.py
```diff
@@ -1687,6 +1731,10 @@
         "type": "integer",
         "description": "Four-digit year that applies to a particular forecasted value.",
     },
+    "form_type": {
+        "type": "string",
+        "description": "Specific version of SEC 10k filed.",
+    },
```
zaneselvans, line 1737: Above we have `exhibit_21_version` which seems analogous, but there we say `version` and here we say `type` -- if both are versions, and this one is specific to the SEC 10K, it seems like it would be more consistent to use `version` in both cases.  So maybe `exhibit21_version` and `sec10k_version`

In our naming conventions we also use the `_type` suffix to indicate "a human readable category from a well defined list of values"

src/pudl/metadata/fields.py
```diff
@@ -2329,6 +2377,7 @@
         "description": "Original reported energy interchange between adjacent balancing authorities.",
         "unit": "MWh",
     },
+    "irs_number": {"type": "string", "description": "ID of the company with the IRS."},
```
zaneselvans, line 2380: In our current naming conventions we use `_id` to identify numerical IDs, and put the data source at the end of the field name, so I think this would probably be `company_id_irs`.

src/pudl/metadata/fields.py
```diff
@@ -2344,6 +2393,10 @@
         "type": "string",
         "description": "The code of the plant's ISO or RTO. NA if not reported in that year.",
     },
+    "key": {
+        "type": "string",
+        "description": "Key within block.",
+    },
```
zaneselvans, line 2399: Without more context it's not clear what this means, and `key` is a very short and generic column name.

src/pudl/metadata/fields.py
```diff
@@ -2456,6 +2509,14 @@
         ),
         "unit": "MW",
     },
+    "location": {
+        "type": "string",
+        "description": "Location of subsidiary company.",
+    },
```
zaneselvans, line 2515: Do we have any analogous fields for this meaning already? Is this a freeform string? A place name? A fully specified address? Would it be clearer as `subsidary_location`? How does this field relate to / differ from `location_of_inc` below?

src/pudl/metadata/fields.py
```diff
@@ -3289,6 +3350,10 @@
         "description": "Whether each generator record is for one owner or represents a total of all ownerships.",
         "constraints": {"enum": ["owned", "total"]},
     },
+    "ownership_percentage": {
+        "type": "string",
+        "description": "Percentage of subsidiary company owned by parent.",
+    },
```
zaneselvans, line 3356: Why is ownership percentage a string instead of a number?

We have an analogous field in [the EIA 860 ownership tables](https://catalystcoop-pudl.readthedocs.io/en/nightly/data_dictionaries/pudl_db.html#core-eia860-scd-ownership) which is `fraction_owned` -- in general we have avoided using percentages (numbers from 0 to 100) and used fractions, since they can be used directly in calculations.

The existing field description is specific to generators, but you can overload it with a resource-specific field description.

src/pudl/metadata/fields.py
```diff
@@ -3297,6 +3362,10 @@
         "type": "boolean",
         "description": "Whether a plant part record has a duplicate record with different ownership status.",
     },
+    "parent_company_cik": {
+        "type": "string",
+        "description": "CIK of the company's parent company.",
+    },
```
zaneselvans, line 3368: What is a CIK? We should spell it out in the description and abbreviate in parentheses.

We have a mix of "company", "parent" and "subsidiary" field names. In any context where there could be any ambiguity as to whether we're referring to a parent or a subsidiary company it should be explicit.

src/pudl/metadata/fields.py
```diff
@@ -3912,6 +3981,14 @@
         "description": "Estimated electricity demand scaled by the total sales within a state.",
         "unit": "MWh",
     },
+    "sec_company_id": {
+        "type": "string",
+        "description": "Assigned identifier for the company.",
+    },
```
zaneselvans, line 3987: As with the IRS ID above, according to our naming conventions I think this would be `company_id_sec` Description should mention SEC.

src/pudl/metadata/fields.py
```diff
@@ -3912,6 +3981,14 @@
...
+    "sec10k_filename": {
```
zaneselvans, line 3988: `filename_sec10k`?

src/pudl/metadata/fields.py
```diff
@@ -4202,6 +4279,10 @@
         "constraints": {"enum": RELIABILITY_STANDARDS},
         # TODO: Might want to make this column more specific to outages: ex: outage calculation standard.
     },
+    "standard_industrial_classification": {
+        "type": "string",
+        "description": "The company's type of business.",
+    },
```
zaneselvans, line 4285: Whose classification system is this coming from? NAICS?  Is it a freeform text field, or a well defined set of codes or categories?

src/pudl/metadata/fields.py
```diff
@@ -4257,6 +4338,10 @@
             "pattern": r"^\d{2}$",
         },
     },
+    "state_of_incorporation": {
+        "type": "string",
+        "description": "Two letter state code where company is incorporated.",
+    },
```
zaneselvans, line 4344: Should this / can this be constrained to match only valid state abbreviations?

src/pudl/metadata/fields.py
```diff
@@ -4299,6 +4384,10 @@
         # TODO: Disambiguate as this means different things in different tables.
         "description": "Physical street address.",
     },
+    "street_address_2": {
+        "type": "string",
+        "description": "Secondary street address.",
+    },
```
zaneselvans, line 4390: We already have an `address_2` field defined.

src/pudl/metadata/fields.py
```diff
@@ -4328,6 +4417,10 @@
         "type": "integer",
         "description": "Sub-plant ID links EPA CEMS emissions units to EIA units.",
     },
+    "subsidiary": {
```
zaneselvans, line 4420: `subsidiary_company_name` or at least `subsidiary_name`?

src/pudl/metadata/fields.py
```diff
@@ -4811,6 +4904,10 @@
         "type": "date",
         "description": "The record in the changelog is valid until this date. The record is valid from the report_date up until but not including the valid_until_date.",
     },
+    "value": {
+        "type": "string",
+        "description": "String value of data point.",
+    },
```
zaneselvans, line 4910: This seems very vague without more context. What does it mean?

src/pudl/metadata/fields.py
```diff
@@ -4944,6 +5041,10 @@
         "type": "integer",
         "description": "Year the data was reported in, used for partitioning EPA CEMS.",
     },
+    "year_quarter": {
+        "type": "string",
+        "description": "Year quarter filing applies to.",
+    },
```
zaneselvans, line 5047: I think for all other columns with this kind of information in them, we are using `report_date` -- sometimes with annual and other times with monthly frequency. In the FERC data I think it also describes quarterly data (although we only pull in Q4 information). Can we store a quarter using a date type, which conveys richer information about the contents, and can be used in conjunction with other report date data?

zaneselvans (changes requested):
I made a bunch of comments about field names and our existing conventions mostly.

Do we not need to define the resources that these fields are going to be part of in `pudl/metadata/resources/sec10k.py`

zschira:
> Do we not need to define the resources that these fields are going to be part of in pudl/metadata/resources/sec10k.py

@zaneselvans yeah I had just accidentally not added the file in my commit, then got pulled away yesterday before I could fix it. Same with removing the `deltalake` dependency. Those should both be fixed now and I'll work through the comments of the `fields` this morning.

src/pudl/metadata/fields.py
zschira: This is basically a narrow tidy table with a bunch of key value pairs that contain company information. These key values are organized in connected blocks. I've changed the naming to hopefully make this structure more clear. Let me know if it makes more sense now.

src/pudl/metadata/fields.py
zschira: I wasn't able to find any fields like that, but I can definitely change this to use an existing field if there's one that makes sense.

src/pudl/metadata/fields.py
zschira: That makes sense, fixed.

src/pudl/metadata/fields.py
zschira: Changed.

src/pudl/metadata/fields.py
zschira: Same as above, I've changed the naming for this table to hopefully make it more clear.

src/pudl/metadata/fields.py
zschira: I think `subsidiary_location` makes sense, and I've changed it. I think there's both state and country names in here, does that sound right @katie-lamb?

src/pudl/metadata/fields.py
zschira: These were a bit messy and I think we just hadn't had the time to clean them fully to get into floats, but I took a crack at it and it wasn't too hard. These are now floats and using the `fraction_owned` field name with custom descriptions.

src/pudl/metadata/fields.py
zschira: I spelled it out in the name since we're using the full `central_index_key` name elsewhere.

src/pudl/metadata/fields.py
zschira: Fixed.

src/pudl/metadata/fields.py
zschira: Changed.

src/pudl/metadata/fields.py
zschira: Fixed.

src/pudl/metadata/fields.py
zschira: Fixed.

src/pudl/metadata/fields.py
zschira: Same as above, I've changed the naming for this table to hopefully make it more clear.

src/pudl/metadata/fields.py
zschira: Looks like this is a [standalone system](https://en.wikipedia.org/wiki/Standard_Industrial_Classification) that was replaced by the NAICS, but is still in use by the SEC.

src/pudl/metadata/fields.py
zschira: I think it should and maybe can. Is there a place where we're applying such a constraint right now? The base `state` field doesn't have any constraints on it.

src/pudl/metadata/fields.py
zschira: Converted to `report_date`

src/pudl/metadata/fields.py
katie-lamb: Yep, `former_confermed_name` comes directly from a field pulled from the 10-K filing. I don't see any reason to keep `conformed` though. We don't use this field in any of the modeling bits.

src/pudl/metadata/fields.py
katie-lamb: Yeah, this is the location field that's pulled directly from the Ex. 21 so it's supposed to mean "location_of_incorporation" but it's not standardized whether that should be a state, city, country etc. So I think `subsidiary_location` is fine, or later we can update the Ex. 21 model to extract this column directly into `location_of_incorporation`.

src/pudl/metadata/fields.py
katie-lamb: I had to make them strings because it was hard to clean out all the errors to make it a float within the budget constraints. Also, I was having pydantic issues when I tried to make it a float. But hopefully Zach has fixed that?

katie-lamb:
I left some responses about naming convention stuff but all the updates seem good to me. @zaneselvans I'll let you weigh in to address requested changes.

src/pudl/metadata/fields.py
```diff
@@ -998,6 +1034,14 @@
...
         "constraints": {"enum": list(SOURCES)},
     },
+    "date_filed": {
+        "type": "datetime",
+        "description": "Date filing was submitted.",
+    },
+    "date_of_name_change": {
+        "type": "datetime",
+        "description": "Date of last name change of the company.",
+    },
```
zaneselvans, line 1044: As above, the type, name, and description should all reflect the same resolution (date vs. datetime) and use `_date` or `_datetime_utc` (if UTC) as suffixes.

src/pudl/metadata/fields.py
zaneselvans: Some more detail in the description would help make it clear that this is a real thing in the world that can be referenced.  E.g.

"Four-digit Standard Industrial Classification (SIC) code identifying the company's primary industry. SIC codes have been replaced by NAICS codes in many applications, but are still used by the SEC."

Is this a string because it can have leading zeroes like FIPS codes?

Under our naming conventions I think this would probably be an `id` e.g. `primary_purpose_id_naics` though `id_sic` is a bit inscrutable.  `industry_id_sic` ???

Can we constrain this like we do for the FIPS IDs?
```py
"constraints": {
    "pattern": r"^\d{4}$",
}
```

src/pudl/metadata/fields.py
zaneselvans: Ah, you're right this is all over the place and we have yet to try and enforce it because it will probably mean chasing down issues.

src/pudl/metadata/resources/sec10k.py
```diff
@@ -0,0 +1,100 @@
...
+        "field_namespace": "sec10k",
+    },
+    "out_sec10k__parents_and_subsidiaries": {
+        "description": "Denormalized table containing SEC 10k company information with mapping between subsidiary and parent companies, as well as a linkage to EIA companies.",
+        "schema": {
+            "fields": [
+                "company_id_sec",
+                "filename_sec10k",
+                "report_date",
+                "report_year",
```
zaneselvans, line 76: Why both `report_date` and `report_year`?

Currently we've standardized on `report_date` elsewhere in PUDL regardless of reporting frequency, but have yet to go back and fix all the FERC tables that still have `report_year`

src/pudl/metadata/fields.py
```diff
@@ -738,6 +742,38 @@
         "description": "Average monthly coincident peak (CP) demand (for requirements purchases, and any transactions involving demand charges). Monthly CP demand is the metered demand during the hour (60-minute integration) in which the supplier's system reaches its monthly peak. In megawatts.",
         "unit": "MW",
     },
+    "company_id_sec": {
+        "type": "string",
+        "description": "Assigned identifier for companies that file SEC10k forms, or show up in Exhibit 21 attachments.",
+    },
```
zaneselvans, line 748: What differentiates the `company_id_sec` from the `central_index_key`?  Are we assigning one of them and the SEC assigning the other? Is there a 1-to-1 mapping between them and they just get used in different contexts? Can we say something about how they relate to each other (or don't) in the field descriptions?

zaneselvans (changes requested):

src/pudl/analysis/pudl_models.py
```diff
@@ -1,43 +1,115 @@
...
 
-import os
-
 import pandas as pd
-from dagster import AssetsDefinition, asset
+from dagster import asset
+
+
+def _load_table_from_gcs(table_name: str) -> pd.DataFrame:
+    return pd.read_parquet(f"gs://model-outputs.catalyst.coop/sec10k/{table_name}")
```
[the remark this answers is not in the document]
zschira: Part of why I was using `deltalake` is that it's a table format that provides versioning and attaches metadata to parqeut files. However, we decided awhile back that we would just treat these tables as static inputs to PUDL for the time being, which is why I'm just reading the raw parquet files here. When we're ready to change the model or integrate new data, we can use `deltalake` to read the tables from GCS and select specific versions.

src/pudl/analysis/pudl_models.py
```diff
@@ -1,43 +1,115 @@
...
+    ) / 100.0
+
+
+def _year_quarter_to_date(year_quarter: pd.Series) -> pd.Series:
+    """Clean percent ownership, convert to float, then convert percent to ratio."""
+    return pd.PeriodIndex(year_quarter, freq="Q").to_timestamp()
+
 
+@asset(
+    io_manager_key="parquet_io_manager",
```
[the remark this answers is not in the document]
zschira: I had done this because when we were trying to read metadata from GCS it would have caused schema inconsistencies for people who don't have access to the metadata there. It should be ok to add these to SQLite now though since the metadata is all in PUDL.

src/pudl/analysis/pudl_models.py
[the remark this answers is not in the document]
zschira: I agree the organization feels weird, but I don't think that would quite make sense either. We are doing some basic transformations in PUDL right now, but ideally these would all be moved back upstream so all we actually do in PUDL is read the parquet files in so they can be distributed. In that case the extract step would be creating `core`/`out` tables which doesn't feel right. Would it make sense to add this to the `etl` directory?

src/pudl/metadata/fields.py
zschira: The `company_id_sec`  is assigned by us. @katie-lamb can probably provide more detail, but my understanding is that some subsidiary companies show up in exhibit 21's, but don't file a 10k and thus don't have a CIK (or at least not that we know of), so we assign this ID instead. I've updated the description to hopefully make this more clear.

src/pudl/metadata/resources/sec10k.py
zschira: `report_year` was there because the upstream table is partitioned by year, but no need to distribute it, so I've dropped it.

src/pudl/metadata/fields.py
```diff
@@ -1643,14 +1659,14 @@
...
-    },
     "filer_count": {
         "type": "integer",
         "description": "Index company information as some filings contain information for multiple companies.",
     },
+    "filename_sec10k": {
+        "type": "string",
+        "description": "Name of filing as provided by SEC data portal.",
+    },
     "files_10k": {
```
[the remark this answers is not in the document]
zschira: Fixed

src/pudl/metadata/resources/sec10k.py
```diff
@@ -0,0 +1,100 @@
...
+                "exhibit_21_version",
+                "report_date",
+            ],
+            "primary_key": [
+                "filename_sec10k",
+            ],
+        },
+        "sources": ["sec10k"],
+        "etl_group": "pudl_models",
+        "create_database_schema": False,
```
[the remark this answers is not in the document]
zschira: Fixed

src/pudl/metadata/fields.py
zschira: Fixed date inconsistencies

src/pudl/metadata/fields.py
```diff
@@ -4417,10 +4412,14 @@
         "type": "integer",
         "description": "Sub-plant ID links EPA CEMS emissions units to EIA units.",
     },
-    "subsidiary": {
+    "subsidiary_company_name": {
         "type": "string",
         "description": "Name of subsidiary company.",
     },
+    "subsidiary_location": {
```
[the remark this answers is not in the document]
zschira: Fixed

src/pudl/analysis/pudl_models.py
```diff
@@ -1,43 +1,115 @@
...
+    return (
+        percent_ownership.str.replace(r"(\.{2,})", r"\.", regex=True)
+        .replace("\\\\", "", regex=True)
+        .replace(".", "0.0", regex=False)
+        .astype("float")
+    ) / 100.0
+
+
+def _year_quarter_to_date(year_quarter: pd.Series) -> pd.Series:
+    """Clean percent ownership, convert to float, then convert percent to ratio."""
```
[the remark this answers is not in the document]
zschira: Fixed

zschira:
@zaneselvans for some reason I can't respond to your comments about the standard industrial classification stuff, but this column was a string in the format: "text description [4-digit-id]". I've changed this to split into two columns, one with the ID as an int and one with the text description.

I also left a comment about where to put the assets. I don't think they quite fit into any existing patterns we've established.

src/pudl/metadata/fields.py
zaneselvans: Okay, this sounds analogous to the `utility_id_ferc1` that we assign to provide a connection between the disparate `utility_id_ferc1_dbf` and `utility_id_ferc1_xbrl` IDs that come from FERC.  Unless we think this ID should apply to all instances of a given company across all SEC filings (which seems unlikely since we're assigning it in the context of this particular work) then I think it should have the more narrow `_sec10k` suffix. I think being explicit that we are the ones doing the assignment ("PUDL-assigned identifier...") as we do with the FERC1 IDs will help people know who to ask for help / blame if something is weird.

zaneselvans (changes requested):
One last naming inconsistency and a dtype / constraint issue on the SIC IDs.

zaneselvans (approved):
@zschira I went ahead and made the last minor naming, description, and dtype changes that I'd requested since it's after work time for you and we need to get this merged in.

There were a couple of higher level comments in here too though.

zaneselvans:
Something went wonky with the integration tests -- it was hung trying to create the conda environment and failing for more than 90 minutes so I yanked it.

zaneselvans:
Environment creation seems to work fine for `ci-unit` so maybe it's just a glitch?

One weird thing is that `mamba` v2.0.6 seems to be getting installed, but [it doesn't seem to have been released yet](https://github.com/mamba-org/mamba/releases) -- the last release seems to be 2.0.6rc3 at the moment.

Incidentally updated the `micromamba` Docker container version since I noticed it was out of date. But don't see how that would have any impact on this.

PR #4048, opened 2025-02-05
Update SEC 10K table level descriptions

katie-lamb:
# Overview

This PR adds more detail to the table level metadata descriptions for the SEC 10K tables.

## What did you change?

Included more detail about the information in this tables, their time granularity, and the probabilistic nature of the modeling process.

# Documentation

```[tasklist]
```

# Testing

```[tasklist]
# To-do list
```


src/pudl/metadata/resources/sec10k.py
```diff
@@ -64,7 +79,21 @@
         "field_namespace": "sec10k",
     },
     "out_sec10k__parents_and_subsidiaries": {
-        "description": "Denormalized table containing SEC 10k company information with mapping between subsidiary and parent companies, as well as a linkage to EIA companies.",
+        "description": (
```
katie-lamb, line 82: @zaneselvans @zschira Are there specific guidelines for what should and shouldn't go into table level metadata? I wanted to include that this table is modeled probabilistically and will contain errors. Does that feel appropriate here?

zaneselvans: I think this is something that we're about to work on. But I think we should err on the side of providing rich information here, since it will end up in the PUDL Data Dictionary, which is the main documentation of our data, and the only documentation that most users will encounter.

"Ownership information about parent to subsidiary company relationships" is confusing to me.  What is the nature of that ownership information? Is it really just vague, unstructured information, or is it a particular kind of information? What direction does that information go?

E.g. is this correct:

Each record in this table describes an ownership relationship between a parent company and a subsidiary company. When possible it also indicates what fraction of the subsidiary is owned by the parent company. Only the most recently reported ownership data for each parent-subsidiary pair is reported.

Given that the table only contains the last identified ownership relationship, how should this information be used, or not used?

E.g. if Southern Company had spun Georgia Power off as a separate company in 2023, I imagine that we would expect to see a record describing Georgia Power as being owned by Southern Company in 2022 in this table. Is that correct?

It seems like there are two kinds of potential time evolution that one might be looking for, but that don't seem to be available.
- First, there's the evolution of ownership over time. If we only keep the last record of a parent-subsidary relationship, then there's no information about whether the parent has been buying or selling the subsidiary gradually over time.
- Second, there's the potential evolution of the mapping between SEC companies and EIA utilities over time.

Is it possible to use this table to answer a question like "In 2020, who were the owners of Public Service Company of New Mexico?" How would one do that?

Is it possible to use this table to answer a question like "In 2013, what other companies did Xcel Energy own?" How would one do that?

zaneselvans:
Another good place to put overarching documentation about the SEC to EIA tables and the process behind them would be an SEC Form 10-K [Data Source documentation page](https://catalystcoop-pudl.readthedocs.io/en/nightly/data_sources/index.html).

zschira:
@zaneselvans and @katie-lamb I think there's a core tension here in that this data is still in a "beta" state and that's not something we have clear guidelines on how to handle. Ultimately it took us a lot of time/effort to get to the point where we could actually run our full extraction model, and if we had more time at the end we could have done a more robust linkage across time and better cleaning on both the content and structure of the tables we're producing.

I definitely think that makes it particularly important to communicate the limitations of the data, but also recognizing that this is still in a bit of a "work in progress" state relative to most of the other tables in PUDL. In general it might be nice to think about some standard practice for communicating to users that a dataset is in a state like this.

src/pudl/metadata/resources/sec10k.py
```diff
@@ -24,7 +27,14 @@
...
     },
     "core_sec10k__exhibit_21_company_ownership": {
-        "description": "Company ownership data extracted from Exhibit 21 attachments to SEC 10k filings.",
+        "description": (
+            """Company ownership data extracted from Exhibit 21 attachments to SEC 10k filings.
+This data is extracted from PDFs of the Exhibit 21 attachment using an information extraction
+machine learning model. This table is connected to the SEC 10k filer information to create
+a dataset on SEC ownership relationships in `out_sec10k__parents_and_subsidiaries`.
+We only completed a first iteration of modeling to extract this data
+and thus it contains errors in the parsing and structure of the data."""
```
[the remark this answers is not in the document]
katie-lamb: Not sure how else to phrase this - most of other PUDL datasets are transformations of raw data and thus won't have errors in string parsing and the actual raw content of the table. There may be certain analysis decisions baked into other tables, but this different in nature. It's a table that's entirely created by a model, so company name strings may be wrong, companies may be associated with the wrong parent company, things may just be wrong where the model was incorrect. But I don't think we have a good way of indicating whether we think a value is wrong. I guess it's pessimistic, but not sure what else to say? Basically "we hoped our model worked well but we know it's far from perfect and it was only a first iteration".

jdangerx:
Is this superceded by #4079 ?

katie-lamb:
@jdangerx yes, closing this.

PR #4070, opened 2025-02-18
feat: output a datapackage.json for PUDL parquet files.

### LINKED ISSUE #4069: Output a `datapackage.json` for PUDL parquet files
### Overview

The current Parquet frontend metadata is grabbed from a metadata.yaml I generated by hand.

### Success Criteria

* [ ] a `datapackage.json` is published alongside nightly builds, describing the Parquet files.

```[tasklist]
### Next steps
* [ ] make `Package` output a datapackage
* [ ] write a test that validates the datapackage
```

jdangerx:
# Overview

Closes #4069 .

I wanted to output a datapackage descriptor for the PUDL viewer to use as the basis for search. Up until now it's been using a vendored version of the Datasette `metadata.yaml` that we generate.

We *could* achieve the effect of "have an up-to-date metadata for PUDL viewer" just by *not* cleaning up the metadata.yaml out of the nightly builds, but distributing some wacky bespoke metadata format seems silly. So I made the necessary changes to output a proper `datapackage.json`.

## What did you change?

* added `to_frictionless` methods to `Package` and `Resource` so we can use their "to a valid datapackage.json" functionality
* added a function to the `gcp_pudl_etl.sh` script to write this datapackage out.

# Documentation

```[tasklist]
- [ ] ~Update relevant Data Source jinja templates (see `docs/data_sources/templates`).~
- [ ] ~Update relevant table or source description metadata (see `src/metadata`).~
- [ ] ~Review and update any other aspects of the documentation that might be affected by this PR.~
```

# Testing

Kicking off a manual run of our build.

```[tasklist]
# To-do list
- [ ] ~If updating analyses or data processing functions: make sure to update or write data validation tests (e.g. `test_minmax_rows()`).~
- [ ] ~For minor ETL changes or data additions, once `make pytest-coverage` passes, make sure you have a fresh full PUDL DB downloaded locally, materialize new/changed assets and all their downstream assets and [run relevant data validation tests](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `pytest` and `--live-dbs`.~
- [ ] ~For bigger ETL or data changes run the full ETL locally and then [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `make pytest-validate`.~
- [x] Alternatively, run the `build-deploy-pudl` GitHub Action manually.
```

### DIFF
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -7,11 +7,6 @@ function send_slack_msg() {
     curl -X POST -H "Content-type: application/json" -H "Authorization: Bearer ${SLACK_TOKEN}" https://slack.com/api/chat.postMessage --data "{\"channel\": \"C03FHB9N0PQ\", \"text\": \"$1\"}"
 }
 
-function upload_file_to_slack() {
-    echo "Uploading file to slack with comment $2"
-    curl -F "file=@$1" -F "initial_comment=$2" -F channels=C03FHB9N0PQ -H "Authorization: Bearer ${SLACK_TOKEN}" https://slack.com/api/files.upload
-}
-
 function authenticate_gcp() {
     # Set the default gcloud project id so the zenodo-cache bucket
     # knows what project to bill for egress
@@ -144,6 +139,7 @@ function notify_slack() {
 
     message+="Stage status (0 means success):\n"
     message+="ETL_SUCCESS: $ETL_SUCCESS\n"
+    message+="WRITE_DATAPACKAGE_SUCCESS: $WRITE_DATAPACKAGE_SUCCESS\n"
     message+="SAVE_OUTPUTS_SUCCESS: $SAVE_OUTPUTS_SUCCESS\n"
     message+="UPDATE_NIGHTLY_SUCCESS: $UPDATE_NIGHTLY_SUCCESS\n"
     message+="UPDATE_STABLE_SUCCESS: $UPDATE_STABLE_SUCCESS\n"
@@ -152,12 +148,12 @@ function notify_slack() {
     message+="DISTRIBUTION_BUCKET_SUCCESS: $DISTRIBUTION_BUCKET_SUCCESS\n"
     message+="GCS_TEMPORARY_HOLD_SUCCESS: $GCS_TEMPORARY_HOLD_SUCCESS \n"
     message+="ZENODO_SUCCESS: $ZENODO_SUCCESS\n\n"
-    message+="*Query* logs on <https://console.cloud.google.com/batch/jobsDetail/regions/us-west1/jobs/run-etl-$BUILD_ID/logs?project=catalyst-cooperative-pudl|Google Batch Console>.\n\n"
-    message+="*Download* logs at <https://console.cloud.google.com/storage/browser/_details/builds.catalyst.coop/$BUILD_ID/$BUILD_ID.log|gs://builds.catalyst.coop/${BUILD_ID}/${BUILD_ID}.log>\n\n"
-    message+="Get *full outputs* at <https://console.cloud.google.com/storage/browser/builds.catalyst.coop/$BUILD_ID|gs://builds.catalyst.coop/${BUILD_ID}>."
+    # we need to trim off the last dash-delimited section off the build ID to get a valid log link
+    message+="<https://console.cloud.google.com/batch/jobsDetail/regions/us-west1/jobs/run-etl-${BUILD_ID%-*}/logs?project=catalyst-cooperative-pudl|*Query logs online*>\n\n"
+    message+="<https://storage.cloud.google.com/builds.catalyst.coop/$BUILD_ID/$BUILD_ID.log|*Download logs to your computer*>\n\n"
+    message+="<https://console.cloud.google.com/storage/browser/builds.catalyst.coop/$BUILD_ID|*Browse full build outputs*>"
 
     send_slack_msg "$message"
-    upload_file_to_slack "$LOGFILE" "$BUILD_ID logs:"
 }
 
 function merge_tag_into_branch() {
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -65,6 +65,12 @@ function run_pudl_etl() {
     && touch "$PUDL_OUTPUT/success"
 }
 
+function write_pudl_datapackage() {
+    echo "Writing PUDL datapackage."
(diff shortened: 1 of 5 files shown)


docker/gcp_pudl_etl.sh
```diff
@@ -65,6 +65,12 @@ function run_pudl_etl() {
     && touch "$PUDL_OUTPUT/success"
 }
 
+function write_pudl_datapackage() {
+    echo "Writing PUDL datapackage."
+    python -c "from pudl.metadata.classes import PUDL_PACKAGE; print(PUDL_PACKAGE.to_frictionless().to_json())" > "$PUDL_OUTPUT/pudl_datapackage.json"
```
zaneselvans, line 70: Is this the right place for the `pudl_datapackage.json` to go, given that the parquet files are output under `$PUDL_OUTPUT/parquet/`? The resource paths in the descriptor refer to the current directory, suggesting that the parquet files and the datapackage.json file should be in the same directory.  Or is this location anticipating that the parquet files will get moved into `$PUDL_OUTPUT` eventually? Or that wherever the JSON is deployed to will be alongside all the parquet files? For more ongoing consistency might we want to output the datapackage.json alongside the parquet files and then also move them together when they move?

Should we make the name parquet specific, given that we could also have an SQLite or DuckDB database that's annotated by a datapackage.json file?

jdangerx:
OK, so there *was* a `pudl_datapackage` uploaded in the [sandbox Zenodo deposition](https://sandbox.zenodo.org/records/169277) from the build I ran yesterday afternoon. But I think we should pull the datapackage-writing up before we do the "save raw outputs to GCS" step so we can also see those outputs there. I'm going to do that, but I also don't think it's necessarily worth running/waiting for a whole new build. WDYT @zaneselvans ?

test/unit/metadata_test.py
zaneselvans, line 1: I think there are also methods in the `frictionless` python package that can let us test the validity of the datapackage descriptor -- i.e. without looking at the data it supposedly references, check that all required fields have been populated and have the correct formats/types. Unfortunately the [API Documentation](https://framework.frictionlessdata.io/docs/framework/package.html#reference-validate) is insanely terse.

zaneselvans (changes requested):
When I move the `pudl_datapackage.json` file into the parquet directory and run:

```
frictionless validate pudl_datapackage.json
```

I get a bunch of errors saying that all the paths are invalid, but it doesn't say why.  Looking at the [data location attributes in the Frictionless spec](https://specs.frictionlessdata.io/data-resource/#data-location) I'm not sure why that would be.

If we expect this to be a valid datapackage then it seems like we should attempt to validate it and have it pass? At least structurally -- i.e. maybe we don't want to check the data contents, since we're doing that with other tooling.  Or maybe we do, for a limited subset of the data (I think the default is just a few 1000 rows per table?) just to demonstrate that this really is nominally a valid datapackage.

docker/gcp_pudl_etl.sh
jdangerx: You're right on all counts. I put this in `$PUDL_OUTPUT` because that's where the parquet files end up in distribution, and it does make more sense to call it `pudl_parquet_datapackage.json` and put it in `parquet/`. Easy fixes.

zaneselvans:
@jdangerx I left some questions about where/when the datapackage descriptor is getting written. If the mechanics for writing it is the same in every case, then writing it outside of the if/else block seems reasonable.

What context is this file going to be used in? Does it make more sense for it to reference local files in the same directory as the datapackage.json (e.g. files downloaded to the same server where the `pudl-viewer` is running)? Or remote parquet outputs on S3 (if those are the ones being referenced by `pudl-viewer`)?

jdangerx:
* re: validation: I spent some time trying to get this to work, and got it to say "the paths are valid" by running the frictionless command *from the directory that the parquet files are in*. but then it hung, presumably trying to read through all of our parquet data and validate it. I tried to figure out how to configure it to only look at the schemas but decided to not go down that rabbit-hole.
* re: context - I sort of feel like, if it's being distributed in a variety of places (zenodo, S3) but always alongside the dang data, it makes sense to have it be the relative path instead of a URL. With the PUDL viewer, we *are* directly accessing from S3 but it doesn't really use the `path` to do that, TBH.

src/pudl/metadata/classes.py
```diff
@@ -1485,6 +1486,22 @@ def to_sql(
...
 
+    def to_frictionless(self) -> frictionless.Resource:
+        """Convert to a Frictionless Resource."""
+        schema = frictionless.Schema(
+            fields=[
+                frictionless.Field(name=f.name, description=f.description)
+                for f in self.schema.fields
+            ],
+            primary_key=self.schema.primary_key,
+        )
```
zaneselvans, line 1497: I believe the `type` values that we specify in our metadata are taken directly from the [Frictionless Table Schema spec](https://specs.frictionlessdata.io/table-schema/#types-and-formats), and so could also be output here, though only `name` is required by the spec.

zaneselvans:
If we're going to distribute the datapackage file in a way that we do *not* expect it to be valid -- not in the right location relative to the data files, or unable to actually validate against those files, then we should probably note that somewhere (maybe in the file itself) so people don't get confused and think it's (unexpectedly) broken.

Maybe @roll knows the magic sauce to just validate the structure of the package via the API in our tests and also not have it dive into millions of rows of data from the CLI? There are a lot of options on `frictionless --validate --help` but I'm not seeing one like this.

zaneselvans:
FYI the table & field descriptions are also directly embedded within the Parquet file metadata and could be accessed from there.

docker/gcp_pudl_etl.sh
```diff
@@ -187,6 +189,8 @@ function clean_up_outputs_for_distribution() {
     zip -0 "$PUDL_OUTPUT/pudl_parquet.zip" ./*.parquet && \
```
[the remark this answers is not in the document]
zaneselvans: Do we not need to add `./pudl_parquet_datapackage.json` to the list of files to put in the zipfile here?

zaneselvans (changes requested):

zaneselvans (approved):

roll:
Hi @zaneselvans!

> Maybe @roll knows the magic sauce to just validate the structure of the package via the API in our tests and also not have it dive into millions of rows of data from the CLI? There are a lot of options on frictionless --validate --help but I'm not seeing one like this.

I believe @pierrecamilleri might suggest.

There is also https://github.com/frictionlessdata/dplib-py that handles only metadata aspects

zaneselvans:
Cc: @jdangerx

PR #4084, opened 2025-02-24
New citations!

krivard:
# Overview

Found these while building a draft of the next newsletter!

Includes the PhD thesis of one of the authors of the racial disparities paper we mentioned in the last newsletter. go CMU EPP! 🖤 💛


zaneselvans (approved):
Very cool to see the references starting to pile up! We should add some of these to the "PUDL in the Wild" section of the next newsletter. @e-belfer

Issue #4076, opened 2025-02-21
Migrate existing tests


zschira:

krivard:
[Additional information and questions doc](https://docs.google.com/document/d/1AEw4TWmcfR1SIekR9SmKib9B7XDy_oIoS51TYYhY_xA/edit?tab=t.0#heading=h.jy8kyvwxw7w)

aesharpe:
# Validation Test Migration Check-list

## Start Now
- [ ] Migrate bespoke validation tests to DBT
- [ ] Migrate correlation validation tests to DBT
- [ ] Migrate group_mean_continuity_check validation tests to DBT
- [ ] Migrate relationship tests to DBT

## Save for later
- [ ] Migrate no_null_cols validation tests to DBT
- [ ] Migrate no_null_rows validation tests to DBT
- [ ] Migrate not_null validation tests to DBT
- [ ] Migrate vs_bounds tests to DBT

zaneselvans:
Rather than having descriptions of the categories here, and then no information in the related issues, I moved the descriptions to the appropriate sub-issues.

zaneselvans:
@krivard I don't seem to be able to open that spreadsheet link.

krivard:
Docs permissions fixed.

krivard:
Not for any of mine or Austen's work; #4093, #4094, and #4096 closed on merge of the relevant PRs. 

For the remainder,
* #4095 is deferred as discussed yesterday, though I'm going to poke at it this afternoon.
* #4104 and #4105 were paused waiting for a macro implementation; as of yesterday it looked likely this was going to be converted to asset checks
* #4106 was paused waiting for a macro implementation that looks like it merged a while ago, but/and I don't know what else needs to happen on it or who should do it
* #4126 there are no comments on the planning doc, so we should (who should?) assign that work out if we're going to do it.
* #4129 looks like a good-first-issue for followup after this effort, unless you think it should block completion here.
* same #4130

zaneselvans:
- I think the tests set to warn and the suspicious row count checks can come after this "migration" is all closed out.
- I commented on #4106 -- those should all be migratable if they haven't been migrated already. Probs need to look through existing `vs_bounds` validations and see which ones remain / delete the ones that have already been migrated.
- It might make sense to roll the docs into the ball of infrastructure changes rather than migrations now, since it'll probably end up changing what we tell people (ourselves...) to do.

zaneselvans:
[...]

PR #4089, opened 2025-02-26
[DRAFT] dbt `not_all_null` migrations

krivard:
# Overview

Migrates `not_all_null` tests to dbt.

Draft: mostly this was getting comfortable with the helper script and `dbt build`, but if we get bonus migrated tests out of it, great!

## What did you change?

# Documentation

# Testing

```
$ dbt build --target etl-fast
```

(no failures in any of the tables touched by this PR. though I do have one in ferc that looks suspicious, and it complains about not being able to find sec10k tables even though i don't think they're even in etl-fast?)

# To-do list

### DIFF
--- a/dbt/README.md
+++ b/dbt/README.md
@@ -328,7 +328,7 @@ validate that there are no completely null columns. We've added a custom test to
       - not_all_null
 ```
 
-See `dbt/models/_out_eia__monthly_heat_rate_by_unit/schema.yml` for specific examples.
+See `dbt/models/output/_out_eia__monthly_heat_rate_by_unit/schema.yml` for specific examples.
 
 Note: there is also a builtin test in `dbt` called `not_null` that will make sure
 there are no null values at all in a column.
--- a/dbt/models/output/_out_eia__monthly_capacity_factor_by_generator/schema.yml
+++ b/dbt/models/output/_out_eia__monthly_capacity_factor_by_generator/schema.yml
@@ -9,8 +9,20 @@ sources:
               partition_column: report_date
         columns:
           - name: report_date
+            data_tests:
+              - not_all_null
           - name: plant_id_eia
+            data_tests:
+              - not_all_null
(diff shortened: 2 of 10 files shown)


zaneselvans:
I made some notes on how I think we might want to approach the check for completely null columns in #4076 

> In general, we want to run this test on every column, so rather than specifying it thousands of times, we probably want to set things up so it is run by default. However, there are cases in the fast ETL (used in the integration tests) where we do expect completely null columns, since columns appear and disappear over time, and it's sometimes the case that a column has data in an old year, but not the 1-2 years that are included in fast ETL. So there needs to be some way to override the default "apply to all columns" when we're running on the fast ETL.

I got the sense from @zschira that there was a way to set up default or table level tests, but I'm not sure how we would handle the fact that we expect different outcomes depending on what years of data are being processed.  Maybe for the handful of columns that don't have data in all years we could specify which years we expect them to be completely null, and which years we don't? Or maybe the not-all-null test just gets disabled on those columns, and there's a different per-year not-all-null test? Or the not-all-null test by default checks across all years of data, but that behavior can be overridden on a per-column basis with some per-year expectations?

I think we expect to remove all of the tables that start with an `_underscore` once we've removed the `PudlTabl` output class, which is now almost entirely restricted to being used by in our tests. If I recall correctly they only exist to make it easy to replicate the old `PudlTabl` behavior from when the class was reading data out of the database and building dataframes to return to the user (pre-Dagster). So migrating tests for those tables is probably not a priority.

marianneke:
As discussed, we'll replace this with model-level tests, so as not to clog up the schema files with basic data hygiene checks that convey no information specific to the table at hand. The macro needed to do this is in https://github.com/catalyst-cooperative/pudl/pull/4091 and we still need to write the code to automatically include this test

krivard:
Indeed! I will close this out so it is no longer confusing

Issue #4075, opened 2025-02-21
Switch to using dbt for data validation


zschira:
# Description

#4011 sets up a `dbt` project we can plan to use for testing our data outputs. This pr establishes the base infrastructure and provides tooling for migrating tests, but there will still be several major steps to allow this framework to take over all of our validation testing. This issue will track the remaining development to finalize the framework.

Some of our high-level desires for this new framework include:

- Re-implements all existing validation tests which are slow and tested relatively infrequently
- Rapid feedback during development
- Work on fast/full ETL (so they're checked in merge queue)
- Easy to add new tests
- Provide memory/compute efficient tooling

## In-scope

- Set up initial `dbt` project
- Tooling to facilitate / semi-automate migration of existing data validations
- Move existing validation tests either into `dbt` tests or `@asset_check` definitions, as appropriate.
- Extend coverage of existing data validation checks to encompass all of our outputs.
- Integrate `dbt` into our CI and testing and nightly build processes
- Retire obsolete / low-value data validation tests after some discussion.
- Documentation of the new system
- Performance improvements necessary to make running the data validations during development / ETL ergonomic
- Improved diagnostic output from data validation tests to make debugging them ergonomic (logging, failing records from dbt, output into the Dagster UI, etc.)

## Out of scope

- Creation of entirely new data validation tests. 
- https://github.com/dagster-io/dagster/issues/18641 (unless it seems easier than asset check factory)

zaneselvans:
See the follow-up Epic: #4421

PR #4090, opened 2025-02-26
make weighted quantile continuous by interpolation

marianneke:
# Overview
- Solves one of the known issues in https://github.com/catalyst-cooperative/pudl/issues/4075

- The weighted quantile macro is a discrete implementation, where it selects the first row after the quantile has been reached.

## What did you change?
- This PR interpolates the value from the two values around the target quantile using linear interpolation.
- The small number of previously migrated tests which may have had their bounds modified to work with the discrete quantiles have been removed and will need to be re-migrated.

# Documentation

Note that there's still some content in the `dbt` README.md related to the errors that would have come up due to the discrete quantile macro. @krivard were you still interested in replacing that file with a real documentation page?

# Testing

# To-do list
- [x] Re-ran the full `dbt` tests locally using fresh outputs of the full ETL.


dbt/macros/weighted_quantile.sql
```diff
@@ -17,11 +17,30 @@ QuantileData AS (
...
+    WHERE cumulative_probability <= {{ quantile }}
+),
+InterpolatedQuantile AS (
+    SELECT
+        CASE
+            WHEN {{ quantile }} = 0 THEN MIN({{ column_name }})  -- Handling quantile = 0
+            WHEN {{ quantile }} = 1 THEN MAX({{ column_name }})  -- Handling quantile = 1
+            ELSE lower_value +
+                 (upper_value - lower_value) *
+                 (({{ quantile }} - lower_prob) / (upper_prob - lower_prob))  -- Regular interpolation
```
krivard, line 37: This math looks good 👍

krivard (approved):
I am no good at window functions but the interpolation math looks right!

zaneselvans:
If this is working as expected then should we go ahead and revert the expectations in the migrated weighted quantile tests to their old values, which were based on the continuous quantiles in the python validations?

zaneselvans:
> Ah, that was a question I had: I was wondering why both this one and the old non-continuous one seem to work, but were the tests made more permissive?

@marianneke Yeah, I think that's what's going on -- at least in the one example that I looked at it was the case. (I also checked out the branch and ran all the tests expecting some of them to fail)

zaneselvans:
@marianneke Maybe the simplest thing to do is just remove all of the already migrated weighted quantile tests here, merge this PR in, and then make a separate PR that migrates all of them -- hopefully they can just be imported directly now that they're using the same calculation. Unless it's easy enough that you want to just do it all in this PR.

dbt/models/output/out_eia__monthly_generators/schema.yml
```diff
@@ -36,12 +36,7 @@ sources:
                   weight_column: capacity_mw
               - expect_column_weighted_quantile_values_to_be_between:
                   quantile: 0.65
-                  max_value: |
-                    {% if target.name == "etl-fast" %}
-                    0.9
-                    {% else %}
-                    0.7
-                    {% endif %}
+                  max_value: 0.7
```
zaneselvans, line 39: I expect that we'll see a bunch of errors if the tests are run against the fast ETL outputs -- that will probably be tthe hard part about migrating these tests.

zaneselvans (changes requested):
I re-migrated the the yearly & monthly generator capacity factor distribution tests to minimize the changes here.

**Blocking**: I tried to make the tests fail by changing the thresholds and quantiles in the YAML files and I was unable to make it fail, which seems like something is broken. We should verify that we can make these tests fail when we expect them to fail before we start migrating things.

Separately, the structure of these test parameterizations seem like they could be more intuitive. Rather than having two separate tests listing separate "min" and "max" values, why not put them in a single test with both the min/max values present? The test is named "quantile between" after all -- also then we'd only be calculating that weighted quantile once, and checking it against the two values, rather than calculating it twice.

zaneselvans:
@marianneke let me know when you're ready for me to look at this again.

marianneke:
> Separately, the structure of these test parameterizations seem like they could be more intuitive. Rather than having two separate tests listing separate "min" and "max" values, why not put them in a single test with both the min/max values present? The test is named "quantile between" after all -- also then we'd only be calculating that weighted quantile once, and checking it against the two values, rather than calculating it twice.

To be clear, this is not an issue with the implementation of the test, but with its usage, since the implementation accepts both a lower and upper bound, and leaving either one out has the expected behavior.

marianneke:
> @marianneke let me know when you're ready for me to look at this again.

@zaneselvans I found the culprit of the weighted quantile tests not failing, and that should be fixed now.

zaneselvans (approved):

PR #4091, opened 2025-02-27
add model level macro to test no null cols

marianneke:
# Overview

Part of https://github.com/catalyst-cooperative/pudl/issues/4075
Namely, the issue of templating no_null_rows/cols tests.

There is a model-level test macro for no_null_rows, but the equivalent test for no_null_cols was not implemented; the available macro just tests a single column for not being all null.
Since these tests are not informative (i.e. they are a basic hygiene test for any model), it is preferable to only use the model-level tests, in order to reduce the bloat of uninformative tests in the model schemas.

## What did you change?
Add macro to loop over all of a model's columns and do the `not_all_null` column level test, returning an error if any of the columns fail this test.

# Documentation

```[tasklist]
```

# Testing

```[tasklist]
# To-do list
```

### DIFF
--- a/dbt/README.md
+++ b/dbt/README.md
@@ -319,16 +319,47 @@ This is failing because the `max_value` is set to `0.7`. If we change this value
 The method `test_no_null_cols_mcoe` in `mcoe_test.py` checks a number of tables to
 validate that there are no completely null columns. We've added a custom test to the
 `dbt` project that can accomplish this. To use this test, simply apply the
-`not_all_null` test to a column in question:
-
+`no_null_cols` test to a model in question:
 ```
+models:
+  - name: some_model
+    data_tests:
+      - no_null_cols
     columns:
-    - name: report_date
-      data_tests:
-      - not_all_null
+      - ...
+```
+
+By default the `no_null_cols` test applies to all columns in the table. To handle deprecated columns (where it is expected that after a certain date the column will only product null values), you can add conditions to select columns for when the test should apply:
+```
+models:
+  - name: some_model
+    data_tests:
+      - no_null_cols:
+          partial_columns:
+            - column_name: "old_column"
+              expect_at_least_one_value_when: "created_at < '2024-01-01'"
+            - column_name: "deprecated_column"
+              expect_at_least_one_value_when: "is_active = TRUE"
 ```
 
 See `dbt/models/_out_eia__monthly_heat_rate_by_unit/schema.yml` for specific examples.
+
+If there are many exceptions, it may be cleaner to apply the test on a column-by-column basis, in which case you can use the `not_all_null` test on a single column instead:
+```
+    columns:
+    - name: some_column
+      data_tests:
+      - not_all_null
+```
+This test accepts an optional `filter` condition for when the test applies:
+```
+columns:
+  - name: deprecated_column
+    data_tests:
+      - not_all_null:
+          filter: "WHERE report_date < '2025-01-01'"
+
+```
 
 Note: there is also a builtin test in `dbt` called `not_null` that will make sure
 there are no null values at all in a column.
--- a/dbt/README.md
+++ b/dbt/README.md
@@ -329,7 +329,7 @@ models:
       - ...
 ```
 
-By default the `no_null_cols` test applies to all columns in the table. To handle deprecated columns (where it is expected that after a certain date the column will only product null values), you can add conditions to select columns for when the test should apply:
+By default the `no_null_cols` test applies to all columns in the table. To handle deprecated columns (where it is expected that after a certain date the column will only produce null values), you can add conditions to select the rows when the test should apply:
 ```
 models:
   - name: some_model
@@ -357,7 +357,7 @@ columns:
   - name: deprecated_column
     data_tests:
       - not_all_null:
-          filter: "WHERE report_date < '2025-01-01'"
+          filter: "report_date < '2025-01-01'"
 
 ```
 
--- a/dbt/macros/no_null_cols.sql
+++ b/dbt/macros/no_null_cols.sql
@@ -1,0 +1,23 @@
+{% macro no_null_cols(model) %}
+    {% set columns = dbt_utils.get_columns_in_relation(model) %}
+    {% set null_columns = [] %}
+
+    {% for column in columns %}
+        {% set column_name = column.name %}
+
+        -- Use the existing not_all_null test for each column
+        {% if execute_test('not_all_null', model=model, column_name=column_name) %}
+            -- If the test fails (all values are null), add the column to the null_columns list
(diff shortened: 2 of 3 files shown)


krivard:
Would this macro make it possible to skip/exclude some cols from the check? There's a list of deprecated columns in the mcoe test that became unavailable at a certain date, and we don't want to check them as part of the no-null-cols (but do want to check the other cols in the table)

zaneselvans:
> Would this macro make it possible to skip/exclude some cols from the check? There's a list of deprecated columns in the mcoe test that became unavailable at a certain date, and we don't want to check them as part of the no-null-cols (but do want to check the other cols in the table)

@marianneke made it sound like this was possible in the inframundo check-in today but maybe we should put an example in to see what it looks like?

marianneke:
@krivard I went with a "positive" filter, meaning the filter condition filters _out_ rows for which the test is not supposed to be applied. I think this is the most intuitive.

One question I have: how should I handle having filter_condition=None? I am thinking it should then just leave out tests for those columns altogether, right?

dbt/macros/no_null_cols.sql
```diff
@@ -0,0 +1,29 @@
+{% test no_null_cols(model, filter_condition=None, filtered_columns=[]) %}
+
+WITH column_checks AS (
+    {% set columns = dbt_utils.get_columns_in_relation(model) %}
```
krivard, line 4: I don't think this function exists in `dbt_utils`; did you mean `adapter`?

krivard:
You could define `filter=None` that way, or you could package the list of columns and filter together in such a way that the filter is always required (and users will set it to TRUE or FALSE if they don't need a more specific condition)

```
- no_null_cols:
    nulls_permitted_in:
      columns: [...]
      filter: ...
```

Whichever you do, you'll want to carefully document the expected behavior. Make sure we're using the word "filter" to mean similar things across all our dbt materials, or choose a name that is less ambiguous. You can be verbose -- we'd rather be legible than win a typing speed contest. Something like:

```
- no_null_cols:
    deactivated_columns:
      column_names: [...]
      active_when: ...
```

```
- no_null_cols:
    partial_columns:
      column_names: [...]
      expect_at_least_one_value_when: ...
```

```
- no_null_cols:
    disappearing_columns:
      column_names: [...]
      available_when: ...
```

```
- no_null_cols:
    special_columns:
      column_names: [...]
      only_fail_if_all_null_when: ...
```

marianneke:
@krivard I implemented your suggestions and updated the README.

Open question:
- Testing test macros is finnicky: we want the failure case behavior to be for the test to fail. Handling this as a dbt unit test is not trivial

dbt/macros/no_null_cols.sql
```diff
@@ -1,19 +1,23 @@
...
     {% for column in columns %}
         {% set column_name = column.name %}
+        {% set condition_sql = "1=1" %}  -- Default: test all rows
+
+        {# Check if this column has a specific condition #}
+        {% for partial_col in partial_columns %}
+            {% if partial_col.get('column_name') == column_name %}
+                {% set condition_sql = partial_col.get('expect_at_least_one_value_when') %}
+            {% endif %}
+        {% endfor %}
```
krivard, line 15: Just noting this is O(mn). Shouldn't be a big deal for currently expected values of m and n, but if we start seeing performance problems we can switch to a map implementation for `partial_columns`, like

```suggestion
        {% if column_name in partial_columns %}
            {% set condition_sql = partial_columns.get(column_name) %}
        {% endif %}
```
we'd lose the nice `expect_at_least_one_value_when` syntax cue though, so probably something to save for later

dbt/README.md
```diff
@@ -319,16 +319,47 @@ This is failing because the `max_value` is set to `0.7`. If we change this value
...
+          partial_columns:
+            - column_name: "old_column"
+              expect_at_least_one_value_when: "created_at < '2024-01-01'"
+            - column_name: "deprecated_column"
+              expect_at_least_one_value_when: "is_active = TRUE"
+```
+
+See `dbt/models/_out_eia__monthly_heat_rate_by_unit/schema.yml` for specific examples.
+
+If there are many exceptions, it may be cleaner to apply the test on a column-by-column basis, in which case you can use the `not_all_null` test on a single column instead:
```
krivard, line 347: 👍 nice clarification

dbt/README.md
```diff
@@ -319,16 +319,47 @@ This is failing because the `max_value` is set to `0.7`. If we change this value
...
       data_tests:
       - not_all_null
 ```
+This test accepts an optional `filter` condition for when the test applies:
+```
+columns:
+  - name: deprecated_column
+    data_tests:
+      - not_all_null:
+          filter: "WHERE report_date < '2025-01-01'"
```
krivard, line 360: this is currently being inserted inside an existing `WHERE` clause, so we shouldn't replicate the keyword

```suggestion
          filter: "report_date < '2025-01-01'"
```

dbt/macros/not_all_null.sql
```diff
@@ -1,9 +1,12 @@
-{% test not_all_null(model, column_name) %}
+{% test not_all_null(model, column_name, filter=None) %}
```
krivard, line 1: Hmm. Will this work for `etl-fast` cases where we're limited to very recent dates, such that there are no rows in the table before the deprecation date?

Given a table with

* `report_date` starting at 2023-01-01 and going forward from there
* `deprecated_column` all null, bc it stopped being published in 2020

and a schema.yml with

```
columns:
  - name: deprecated_column
    data_tests:
      - not_all_null:
          filter: "WHERE report_date < '2020-01-01'"
```

Will this pass or fail?

marianneke: Schema tests fail only when there are rows returned, so if a filter results in zero rows being returned, that would mean the test passes.

dbt/macros/no_null_cols.sql
```diff
@@ -1,19 +1,23 @@
-{% test no_null_cols(model, filter_condition=None, filtered_columns=[]) %}
+{% test no_null_cols(model, partial_columns=[]) %}
 
 WITH column_checks AS (
-    {% set columns = dbt_utils.get_columns_in_relation(model) %}
+    {% set columns = adapter.get_columns_in_relation(model) %}
 
     {% for column in columns %}
         {% set column_name = column.name %}
+        {% set condition_sql = "1=1" %}  -- Default: test all rows
```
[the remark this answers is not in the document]
marianneke: Boolean literals might behave differently depending on the type of SQL. Not sure if I encountered an error with this or whether ChatGPT told me that, but adding a condition this way should always work

zaneselvans:
I'm gonna close this as superseded by #4382

PR #4092, opened 2025-02-28
Add macro for group_mean_continuity_check

krivard:
# Overview

Three tables require a `group_mean_continuity_check`, which does the following:

1. Group by a specified column
2. Compute the mean of each group
3. Compute the percent change between successive groups
4. Check against a per-column threshold

This PR contains a draft for a possible implementation.

## What did you change?

- New macro, `group_mean_continuity_check(group_column, max_pct_change)`
- Demo of macro in use for `_core_eia923__cooling_system_information`, using columns and thresholds from [`transform/eia923.py`](https://github.com/catalyst-cooperative/pudl/blob/7ab320421f2e08438469500feec9335f9e5cee86/src/pudl/transform/eia923.py#L1340)

## Debatable design choices

- This is written as a column-level check
  - pros: macro source is simple (no loops); keeps all checks for a column together in the column entry
  - cons: the grouping column is replicated in each data-test entry
  - options: could rewrite as a table-level check with a list of columns and thresholds. We'd replicate column names doing that, but the existing python solution does it that way so maybe past-catalyst already decided that's superior?
 - Currently no support for `n_outliers_allowed`, which is definitely a blocker
   - options: maybe query with an offset?

### DIFF
--- a/dbt/macros/group_mean_continuity_check.sql
+++ b/dbt/macros/group_mean_continuity_check.sql
@@ -1,0 +1,23 @@
+{% test group_mean_continuity_check(model, column_name, group_column, max_pct_change) %}
+
+with GroupMean as (
+    select
+    {{ group_column }},
+    avg({{ column_name }}) as mean
+    from {{ model }}
+    group by {{ group_column }}
+),
+Continuity as (
+    select
+    {{ group_column }},
+    mean as newer_value,
+    LAG(mean) OVER (order by {{ group_column }}) as older_value
+    from GroupMean
+)
+select
+{{ group_column }},
+abs(newer_value - older_value)/older_value as pct_change
+from Continuity
+where newer_value is not null and older_value is not null and pct_change >= {{ max_pct_change }}
+
+{% endtest %}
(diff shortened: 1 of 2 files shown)


marianneke (commented):
I have a question: you mention pct change between successive groups. This suggests the group column is inherently ordered, right? Is this something you would generally expect to be the case?

I guess what I'm saying is: to me it is not intuitive that a "group" is always an ordered thing. Maybe this variable name should be changed to something that makes it clear that it is both a partition and something that has an inherent order to it

krivard:
@aesharpe even though 2 of the 3 arguments for the test are intended to be identical across all the column tests within a table? and if they ever change, we have to update N lines instead of just one?

I can live with that; I just want to make sure the impact is clear before we decide.

aesharpe:
> @aesharpe even though 2 of the 3 arguments for the test are intended to be identical across all the column tests within a table? and if they ever change, we have to update N lines instead of just one?
> 
> I can live with that; I just want to make sure the impact is clear before we decide.

Yes, this is good clarification. I think it's important to be able to customize things like `max_fr_change` and `n_outliers_allowed` to each column, hence, column level test :)

Issue #4088, opened 2025-02-25
Define schema for imputed EIA 930 tables


zschira:
### Background
As we apply imputation methods to EIA 930 demand data, we will be adding new columns/tables. These schema changes and additions are relatively minor/constrained, but there's a few questions worth discussing before diving too deep into this work.

### General approach
Generally, we can follow the conventions set by the `out_ferc714__hourly_estimated_state_demand` table where we currently do demand imputation. This means we will produce the following tables:

#### `out_eia930__hourly_estimated_subregion_demand`
- balancing_authority_code_eia
- balancing_authority_subregion_code_eia
- datetime_utc
- demand_mwh

#### `out_eia930__hourly_estimated_operations`
- balancing_authority_code_eia
- datetime_utc
- demand_mwh
- demand_forecast_mwh
- net_generation_mwh (if we apply imputation to net generation)

#### What to do with interchange
The upstream table `core_eia930__hourly_estimated_operations` includes interchange, but we don't plan to apply our imputation methods to interchange. Given that fact, it probably wouldn't make sense to include interchange in the `out_eia930__hourly_estimated_operations`, but maybe it would be convenient for users so they could just use the output table?

#### Should we create spatially aggregated output tables?
The imputed 930 release on zenodo which we're comparing against includes aggregations to the region, interconnect and contiguous US levels. Should we produce output tables at each of these levels of aggregation?

zaneselvans:
@zschira I don't know if `out_ferc714_hourly_estimated_state_demand` is quite the right model here. It's based on the imputed hourly data, but there's also a bunch of spatial allocation of demand that happens in there, so the results are much further removed from the originally reported data than what we're looking at doing with the EIA-930.

## New Intermediate FERC-714 Outputs

In fact... it seems that right now we aren't publishing anything from the FERC-714 data that's analogous to the imputed version of the EIA-930 data, which doesn't seem great -- users should be able to use the imputed time series for their own purposes at the originally reported spatial / entity resolution, and it would be helpful to have the imputed version of FERC-714 for direct comparison with the imputed version of EIA-930. How hard would it be to create a new intermediate FERC-714 asset that includes the imputed demand values, before the spatial reallocation has happened?  I think when we were originally developing this process, we weren't as able to process and distribute large tables, and so we probably over-economzied on the outputs that were getting saved.

[...]

PR #4113, opened 2025-03-05
Refactor Time series cleaning

### LINKED ISSUE #4112: Refactor/generalize existing timeseries cleaning
## Overview

This issue will track refactoring of the existing timeseries cleaning code, which is applied to FERC714 hourly demand data. The goal is to generalize the existing implementation so it can be applied to EIA 930 data as well. This should also include a slight restructuring of the FERC 714 tables as discussed in #4088.

- [x] Create generic imputation functions that can be used for either FERC-714 or EIA-930
- [x] Add imputed (but not aggregated to the state level) assets for FERC-714
- [x] Check old vs new imputed FERC-714 outputs for regressions -- should be nearly identical
- [x] Update docstrings to reflect new generic code
- [x] Add imputation reason codes to FERC-714 outputs

zschira:
# Overview

Closes #4112 .

This PR generalizes our timeseries cleaning/imputation code currently used on FERC 714 hourly demand data so it can also be applied to EIA 930. It will also restructure FERC 714 tables as discussed in #4088, and test that the new implementation does not break the FERC 714 imputation.

# Scope

### Generalize Timeseries imputation
- [x] Decouple timeseries cleaning from FERC 714
- [x] Update all docstrings/comments to remove references to 714 in generic timeseries cleaning code
- [x] Create new FERC 714 assets and reorganize where assets live as discussed in #4088
- [x] Update table metadata for new assets
- [x] Verify that changes have not impacted downstream 714 asset, `out_ferc714__hourly_estimated_state_demand`
- [x] Fix timeseries unit test

### Add imputation reason code table
- [x] Define core code table
- [x] Add code column to all imputed tables

## Out of scope
- Improve resource management to enable imputation on many timeseries' (930 sub region demand is relatively small and doesn't blow up resources so not necessary for initial pr)
- Apply imputation to `core_eia930__hourly_operations` table on BA level demand/net generation
- Define aggregated tables to region/interconnect levels
- Test imputation on `core_eia930__hourly_subregion_demand`


src/pudl/metadata/resources/ferc714.py
```diff
@@ -59,6 +59,47 @@
...
+            "in the hour 24 location of the form have a timestamp with an hour reported as "
+            "T24 while the remaining third report this as T00 of the next day. T24 is not a "
+            "valid format for the hour of a datetime, so we convert these T24 hours into "
+            "T00 of the next day. A smaller subset of the respondents reports the 24th hour "
+            "as the last second of the day - we also convert these records to the T00 of the "
+            "next day.\n\nThis table includes three respondent ID columns: one from the "
+            "CSV raw source, one from the XBRL raw source and another that is PUDL-derived "
+            "that links those two source ID's together. This table has filled in source IDs "
+            "for all records so you can select the full timeseries for a given respondent from "
+            "any of these three IDs."
```
zaneselvans, line 83: Rather than cut and pasting the entire description here, it might make sense to reference the core table with a minimal explanation of where it's coming from, and then describe what is new about this table compared to that one in more detail, i.e. the imputed values, which aren't mentioned here.

src/pudl/metadata/resources/ferc714.py
```diff
@@ -59,6 +59,47 @@
...
+        "schema": {
+            "fields": [
+                "respondent_id_ferc714",
+                "respondent_id_ferc714_csv",
+                "respondent_id_ferc714_xbrl",
+                "report_date",
+                "datetime_utc",
+                "timezone",
+                "demand_reported_mwh",
+                "demand_imputed_mwh",
```
zaneselvans, line 94: Need to add the imputation code right?

src/pudl/analysis/timeseries_cleaning.py
zaneselvans, line 1: Basically this is just migrating all of the bespoke demand time series massaging that we were doing over in the state demand module, right?

A fair bit of the infrastructure in here feels more specific than "time series" in general. Is there a narrower type of time series that we're really talking about? Hourly demand and generation?

src/pudl/analysis/state_demand.py
```diff
@@ -586,6 +141,25 @@ def total_state_sales_eia861(
...
+    """Impute hourly demand on core_ferc714__hourly_planning_area_demand table."""
+    df = pudl.analysis.timeseries_cleaning.impute_timeseries(
+        core_ferc714__hourly_planning_area_demand,
+        years=context.resources.dataset_settings.ferc714.years,
+        value_col="demand_mwh",
+        id_col="respondent_id_ferc714",
+    )
+    return df.drop_duplicates(subset=["respondent_id_ferc714", "datetime_utc"])
+
+
```
zaneselvans, line 162: This and the other `out_ferc714` assets feel like they would be more intuitively found in the `pudl.output.ferc714` module now, since they don't just pertain to the State Demand outputs.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1299,3 +1322,319 @@ def summarize_imputed(self, imputed: np.ndarray, mask: np.ndarray) -> pd.DataFra
...
+
+    Args:
+        df: FERC 714 hourly demand matrix,
+            as described in :func:`load_ferc714_hourly_demand_matrix`.
+        tz: FERC 714 respondent time zones,
+            as described in :func:`load_ferc714_hourly_demand_matrix`.
+
+    Returns:
+        Long-format hourly demand with columns ``respondent_id_ferc714``, report
+        ``year`` (int), ``datetime_utc``, and ``demand_mwh``.
```
zaneselvans, line 1393: Need to update all the docstrings to be generic but you know that.

src/pudl/analysis/state_demand.py
```diff
@@ -117,404 +70,6 @@ def lookup_state(state: str | int) -> dict:
...
-            else x.name
-        ).dt.tz_localize(None)
-    )
-
-
-# --- Datasets: References --- #
-
-
-def load_ventyx_hourly_state_demand(path: str) -> pd.DataFrame:
-    """Read and format Ventyx hourly state-level demand.
```
zaneselvans, line 190: This function for checking our outputs seems to have disappeared. Is that intentional?

zaneselvans (changes requested):
Some data issues that need to be explored:

- Why do we end up with more null values in the imputed outputs than the input data sometimes? Identify cases where this is happening and try to figure out what makes them different from the other areas that are getting imputed.
- What is the appropriate collection of correlated demand time series to use when imputing the BA subregions? Smaller regions are likely going to be noisier, and there are not very many subregions compared to the whole universe of BAs. Should we be using all of the whole BA time series? Instead of / In addition to the subregions? What if we just used the other subregions from within a single BA?  Would that be enough information to get a reasonable imputation? Do they tend to have correlated NA values? Of all the demand time series, which ones are the most valuable / important for modeling use cases?
- Do the BA subregions accurately aggregate up to the reported demand from the whole BAs they represent?  Needs a visualization. Could be a correlation scatterplot.

zaneselvans:
Looking at a few places where the subregion demand has been imputed, it seems like the demand curves are more variable than for the full BAs and the imputed values aren't particularly smooth. E.g. PGAE in CAISO in Feb 2019:

```py
start_date = "2019-02-01"
end_date = "2019-02-14"
df = out_sub_eia930.set_index(["balancing_authority_code_eia", "balancing_authority_subregion_code_eia", "datetime_utc"]).sort_index()
df.loc[("CISO", "PGAE", slice(start_date, end_date)), ["demand_reported_mwh", "demand_imputed_mwh"]].plot()
```

image

zaneselvans:
Looking at how the null-ness of the different BA subregions changes before/after the imputation, there are a number that end up *more* null which seems like it warrants further investigation:

```py
df = (
    out_sub_eia930
    .groupby(
        [
            "balancing_authority_code_eia",
            "balancing_authority_subregion_code_eia"
        ],
        observed=True
    )[
        [
            "demand_reported_mwh",
            "demand_imputed_mwh"
        ]
    ]
    .count()
    .assign(
        ratio=lambda x: x["demand_imputed_mwh"] / x["demand_reported_mwh"]
    )
)
df
```
|                  |   demand_reported_mwh |   demand_imputed_mwh |    ratio |
|:-----------------|----------------------:|---------------------:|---------:|
| ('CISO', 'PGAE') |                 52513 |                52607 | 1.00179  |
| ('CISO', 'SCE')  |                 52513 |                52607 | 1.00179  |
| ('CISO', 'SDGE') |                 52513 |                52607 | 1.00179  |
| ('CISO', 'VEA')  |                 52513 |                43823 | 0.834517 |
| ('ERCO', 'COAS') |                 49081 |                52608 | 1.07186  |
| ('ERCO', 'EAST') |                 49081 |                52608 | 1.07186  |
| ('ERCO', 'FWES') |                 49075 |                52608 | 1.07199  |
| ('ERCO', 'NCEN') |                 49076 |                52608 | 1.07197  |
| ('ERCO', 'NRTH') |                 49079 |                52608 | 1.0719   |
| ('ERCO', 'SCEN') |                 49081 |                52608 | 1.07186  |
| ('ERCO', 'SOUT') |                 49081 |                52608 | 1.07186  |
| ('ERCO', 'WEST') |                 49081 |                52608 | 1.07186  |
| ('ISNE', '4001') |                 52607 |                52607 | 1        |
| ('ISNE', '4002') |                 52607 |                52607 | 1        |
| ('ISNE', '4003') |                 52607 |                52607 | 1        |
| ('ISNE', '4004') |                 52607 |                52607 | 1        |
| ('ISNE', '4005') |                 52607 |                52607 | 1        |
| ('ISNE', '4006') |                 52607 |                52607 | 1        |
| ('ISNE', '4007') |                 52607 |                52607 | 1        |
| ('ISNE', '4008') |                 52607 |                52607 | 1        |
| ('MISO', '0001') |                 52573 |                52608 | 1.00067  |
| ('MISO', '0004') |                 52573 |                52608 | 1.00067  |
| ('MISO', '0006') |                 52573 |                52608 | 1.00067  |
| ('MISO', '0027') |                 52573 |                52608 | 1.00067  |
| ('MISO', '0035') |                 52573 |                52608 | 1.00067  |
| ('MISO', '8910') |                 52573 |                52608 | 1.00067  |
| ('NYIS', 'ZONA') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZONB') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZONC') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZOND') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZONE') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZONF') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZONG') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZONH') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZONI') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZONJ') |                 52604 |                52607 | 1.00006  |
| ('NYIS', 'ZONK') |                 52604 |                52607 | 1.00006  |
| ('PJM', 'AE')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'AEP')   |                 52389 |                52607 | 1.00416  |
| ('PJM', 'AP')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'ATSI')  |                 52389 |                52607 | 1.00416  |
| ('PJM', 'BC')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'CE')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'DAY')   |                 52389 |                52607 | 1.00416  |
| ('PJM', 'DEOK')  |                 52389 |                52607 | 1.00416  |
| ('PJM', 'DOM')   |                 52389 |                52607 | 1.00416  |
| ('PJM', 'DPL')   |                 52389 |                52607 | 1.00416  |
| ('PJM', 'DUQ')   |                 52389 |                52607 | 1.00416  |
| ('PJM', 'EKPC')  |                 52389 |                52607 | 1.00416  |
| ('PJM', 'JC')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'ME')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'PE')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'PEP')   |                 52389 |                52607 | 1.00416  |
| ('PJM', 'PL')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'PN')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'PS')    |                 52389 |                52607 | 1.00416  |
| ('PJM', 'RECO')  |                 52389 |                17544 | 0.334879 |
| ('PNM', 'PNM')   |                 52415 |                52607 | 1.00366  |
| ('PNM', 'TSGT')  |                 52415 |                52607 | 1.00366  |
| ('SWPP', 'CSWS') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'EDE')  |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'GRDA') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'INDN') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'KACY') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'KCPL') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'LES')  |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'MPS')  |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'NPPD') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'OKGE') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'OPPD') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'SECI') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'SPRM') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'SPS')  |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'WAUE') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'WFEC') |                 52511 |                43848 | 0.835025 |
| ('SWPP', 'WR')   |                 52511 |                43848 | 0.835025 |

zaneselvans:
@zschira I think we decided to make this PR just the generalization of the imputation code and its application to the FERC-714 data right?  With new EIA-930 work to come in a second PR.

src/pudl/metadata/resources/ferc714.py
zschira: > Need to add the imputation code right?
Updated

> Also, should we use the same naming convention as in the EIA-930 here indicating who is doing the imputation (pudl)? Hmm.

I think that's reasonable for consistency, I've updated to that pattern.

src/pudl/analysis/timeseries_cleaning.py
zschira: That's all we have plans to apply this to right now, are there any other cases where this imputation might be applicable, or should we just make it specific to those?

src/pudl/analysis/state_demand.py
zschira: Moved

src/pudl/analysis/timeseries_cleaning.py
zschira: I think I've fixed all the docstrings now

src/pudl/output/eia930.py
```diff
@@ -0,0 +1,43 @@
+"""Functions for compiling derived aspects of the EIA 930 data."""
+
+import pandas as pd
+from dagster import asset
+
+import pudl.analysis.timeseries_cleaning
+
+BA_TIMEZONE_MAP = {
+    "CISO": "America/Los_Angeles",
+    "ERCO": "America/Denver",
```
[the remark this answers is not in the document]
zschira: Holding off on all 930 specific stuff for next PR

zschira:
@zaneselvans I've updated the scope to make it specific to the refactoring and not include subregion demand. Last steps are to finish regression testing the downstream 714 assets, and fix a `timeseries_cleaning` specific unit test. I'm doing to the regression testing in a notebook that I'll push to this PR once it's ready.

src/pudl/analysis/state_demand.py
```diff
@@ -17,23 +17,12 @@
...
-
 import geopandas as gpd
-import numpy as np
 import pandas as pd
-from dagster import AssetOut, Field, asset, multi_asset
+from dagster import Field, asset
 
-import pudl.analysis.timeseries_cleaning
-import pudl.logging_helpers
-import pudl.output.pudltabl
```
zschira, line 31: For some reason when you remove this `import pudl.output.pudltabl` I end up getting a circular import error, which is breaking the tests right now. `pudltabl` was already not being used here before, so it's completely extraneous but also apparently load bearing. I'll keep trying to figure this out, but I've spent awhile already debugging. @zaneselvans let me know if you have any thoughts on this weird one.

src/pudl/analysis/imputation/timeseries_cleaning.py
```diff
@@ -0,0 +1,1654 @@
+"""Screen timeseries for anomalies and impute missing and anomalous values.
```
zschira, line 1: I moved imputation stuff into a subpackage so we could do more generic imputation within here if that makes sense.

src/pudl/analysis/imputation/timeseries_cleaning.py
```diff
@@ -0,0 +1,1654 @@
...
+    """Return DataFrame with localized ``datetime`` converted back to ``datetime_utc``."""
+    # Convert local times to UTC
+    input_df["datetime_utc"] = _local_to_utc(
+        input_df["datetime"], input_df["utc_offset"]
+    )
+    return input_df
+
+
+@dataclass
+class FlaggedTimeseries:
```
zschira, line 191: TUrned this into just a data type that gets passed around, because the state management of the old class was really confusing and felt error prone to me.

zschira: Otherwise haven't modified the methods below

src/pudl/analysis/imputation/timeseries_cleaning.py
```diff
@@ -0,0 +1,1654 @@
...
+            logger.info(f"{msg}:\n{report}")
+    # Drop respondents with no data
+    blank = df.columns[df.isnull().all()].tolist()
+    df = df.drop(columns=blank)
+    # Report dropped respondents (with no data)
+    logger.info(f"Dropped blank respondents: {blank}")
+    return df
+
+
+def impute_timeseries_asset_factory(
```
zschira, line 1555: I've turned this into an asset factory that does imputation in 3 steps:
1. pivot to timeseries matrix
2. flag values for imputation
3. impute values and melt back into original shape with new columns

I decided to turn this into assets, because it reduces peak memory usage a bit to offload intermediate products to disk.

src/pudl/analysis/timeseries_cleaning.py
zaneselvans: This particular algorithm is pretty dedicated to timeseries with some periodicity / seasonality so I don't think it'll end up anywhere else.

src/pudl/analysis/state_demand.py
zaneselvans: We're going to delete `pudltabl` as soon as the data validations / tests are refactored, so I think it's not worth debugging -- just leave it in if it makes things work.

src/pudl/analysis/imputation/timeseries_cleaning.py
zaneselvans: This would be good information to put in the docstring.

src/pudl/analysis/imputation/timeseries_cleaning.py
```diff
@@ -0,0 +1,1654 @@
...
+@pa.check_types
+def local_dataframe_to_utc(
+    input_df: DataFrame[LocalizedTimeseriesDataFrame],
+) -> DataFrame[UTCTimeseriesDataFrame]:
+    """Return DataFrame with localized ``datetime`` converted back to ``datetime_utc``."""
+    # Convert local times to UTC
+    input_df["datetime_utc"] = _local_to_utc(
+        input_df["datetime"], input_df["utc_offset"]
+    )
+    return input_df
```
zaneselvans, line 187: When I attempt to materialize the FERC-714 assets I get a type checking error here from Pandera. It seems like the `datetime` column is getting lost somewhere along the way (or maybe it's in the index?)

```
pandera.errors.SchemaError: error in check_types decorator of function 'local_dataframe_to_utc': column 'datetime' not in dataframe. Columns in dataframe: ['id_col', 'demand_imputed_pudl_mwh', 'demand_imputed_pudl_mwh_imputation_reason_code']
```

src/pudl/analysis/imputation/timeseries_cleaning.py
```diff
@@ -0,0 +1,1654 @@
...
+        matrix = localized_df.pivot(
+            index="datetime", columns="id_col", values="value_col"
+        )
+        return matrix
+
+    @multi_asset(
+        ins={"matrix": AssetIn(timeseries_matrix_asset)},
+        outs={cleaned_timeseries_matrix_asset: AssetOut(), flags_asset: AssetOut()},
+    )
+    def _flag_timeseries_matrix(
```
zaneselvans, line 1610: In the DAG I notice that there's now an asset just called `_flag_timeseries_matrix`. If we are using this factory in multiple places for EIA-930 and FERC-714 will that result in a name collision? Does the asset need a use-case specific name specified in the decorator?

zaneselvans (changes requested):
Some naming and organizational stuff, and also the ETL is failing for me because a `datetime` column is missing in the translation back to UTC.

zaneselvans:
@zschira Now I'm getting failures on different downstream assets:

* `out_ferc714__hourly_estimated_state_demand`: Column not found: `demand_imputed_mwh`
* `out_ferc714__summarized_demand`: KeyError: "['demand_imputed_mwh'] not in index"

zschira:
@zaneselvans while validating the downstream assets, I found that the the timezone conversion issues I noticed earlier are a bit more problematic than I realized before (I'll outline below). I've put together a workaround that I think solves this problem, and I'm rerunning all of the downstream assets so I can finish validating them. Given that I've changed the behavior of the imputation, I won't be able to do a straightforward regression test of these assets, but I'll try to demonstrate the changes so we can feel confident in the outputs.

The downstream assets should all be fixed right now, but when I'm actively working I might sometimes push breaking changes, and I'm happy to let you know when things are actually ready for testing.

### The problem
Earlier on I noticed that the conversion from `datetime_utc`->`datetime`->`datetime_utc` produced some duplicate values, but I thought this was only touching a very small number of records. However, I've now found that while there were only a few cases where the conversion was producing duplicates, there are many (I'm still trying to verify the exact number, but at least 10s of thousands of rows are impacted) cases where the roundtrip conversion did not produce the original `datetime_utc` values.

This issue became apparent, because I'm merging the imputed demand back on the core input table with the reported demand, so when the datetimes don't align properly, you could have an imputed value corresponding to the wrong reported value. I played around for a bit with fixing the conversion, but ultimately came up with a workaround where I don't do the roundtrip conversion at all, and instead merge back onto a version of the core table with a converted `datetime` and the original `datetime_utc` columns. This allows me to use the `datetime` column as a merge key, which should correspond exactly with the `datetime` used during imputation.

Given that when the timezone conversion fails, it results in very minor differences in the datetime (usually off by an hour), I don't expect this to have substantial impacts on the downstream assets, but they also won't be 1-1 recreations, so in the notebook I'm working on I'll try to demonstrate that these tables are pretty close to what we had before.

zaneselvans:
After uploading a parquet file of the old imputed outputs (which we were not persisting previously) this plots the correspondence between the old and new imputed FERC-714 hourly planning area demand outputs:

```python
new_imputed = defs.load_asset_value("out_ferc714__hourly_planning_area_demand")
old_imputed = pd.read_parquet("s3://pudl.catalyst.coop/nightly/_out_ferc714__hourly_imputed_demand.parquet")

both_imputed = pd.merge(
    new_imputed.set_index(["respondent_id_ferc714", "datetime_utc"]),
    old_imputed.set_index(["respondent_id_ferc714", "datetime_utc"]),
    left_index=True,
    right_index=True,
    suffixes=("_new", "_old"),
    how="outer",
)

import matplotlib.pyplot as plt
import seaborn as sns

# Reset the index to make `respondent_id_ferc714` a data column
both_imputed = both_imputed.reset_index()

# Assign a discrete color to each `respondent_id_ferc714`
unique_ids = both_imputed["respondent_id_ferc714"].unique()
palette = sns.color_palette("tab20", len(unique_ids))
color_map = {rid: palette[i] for i, rid in enumerate(unique_ids)}
colors = both_imputed["respondent_id_ferc714"].map(color_map)

# Create the scatter plot
plt.figure(figsize=(12, 12))
plt.scatter(
    both_imputed["demand_mwh"],
    both_imputed["demand_imputed_pudl_mwh"],
    c=colors,
    s=0.1,
    alpha=0.1,
)

# Set both axes to logarithmic scale
plt.xscale("log")
plt.yscale("log")

# Add gridlines
plt.grid(True, which="both", linestyle="--", linewidth=0.5)

# Optionally add labels and a title
plt.xlabel("Old Imputed FERC-714 Planning Area Demand [MWh]")
plt.ylabel("New Imputed FERC-714 Planning Area Demand [MWh]")
plt.title("Log-Log Scatter Plot of Old vs New Imputed Demand")

plt.show()
```

With some transparency in the points to show where the bulk of the data actually is:

image

With no transparency to highlight the outliers. All the big (in relative terms) outliers are associated with much smaller absolute levels of demand, while at the upper end of the demand spectrum, the correlation between the old and new implementations of the imputations become very closely aligned.

image

The main difference between the old and new implementations is the way we're translating timezone offsets. @zschira does this only affect which records line up with each other in the outputs? Or is it also changing the relative offsets of some timeseries in the imputation itself?

src/pudl/metadata/fields.py
```diff
@@ -1120,6 +1120,15 @@
...
         "unit": "MWh",
     },
+    "demand_imputed_pudl_mwh": {
+        "type": "number",
+        "description": "Electricity demand calculated by subtracting BA interchange from net generation, with outliers and missing values imputed by in PUDL.",
+        "unit": "MWh",
+    },
+    "demand_imputed_pudl_mwh_imputation_code": {
+        "type": "string",
+        "description": "Code describing why a demand value was flagged for imputation.",
```
zaneselvans, line 1130: I think this field needs to have an `enum` constraint applied to it if we want it to be treated as a categorical value and dictionary encoded in Parquet -- otherwise I think it'll treat every string as potentially unique and store them all, which will take up much more memory / disk than necessary.

src/pudl/etl/static_assets.py
```diff
@@ -43,6 +43,7 @@ def _read_static_encoding_tables(
     outs={
         table_name: AssetOut(io_manager_key="pudl_io_manager")
         for table_name in Package.get_etl_group_tables("static_pudl")
+        if table_name != "core_pudl__codes_imputation_reasons"
```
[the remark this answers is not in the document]
zschira: I've moved the asset to the `static_assets.py`, but I do think it's handy to create the table from the same Enum I use when actually flagging values so they're explicitly coupled and defined in exactly one place.

src/pudl/analysis/imputation/codes.py
```diff
@@ -0,0 +1,38 @@
...
+    GLOBAL_OUTLIER = (
+        "Indicates value is greater or less than n times the global median."
+    )
+    GLOBAL_OUTLIER_NEIGHBOR = "Indicates value is a neighbors global outliers."
+    LOCAL_OUTLIER_HIGH = "Indicates value is a local outlier on the high end."
+    LOCAL_OUTLIER_LOW = "Indicates value is a local outlier on the low end."
+    DOUBLE_DELTA = "Indicates value is very different from neighbors on either side."
+    SINGLE_DELTA = (
+        "Indicates value is significantly different from nearest unflagged value."
+    )
```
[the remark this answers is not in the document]
zschira: I've changed this so everywhere we use the names of these codes we first call `lower()` on them, so they're all snakecase.

And I used an `Enum` because it provides a nice type that I can use to annotate the `flag()` method in `timeseries_cleaning.py`. Now instead of flagging with a free-form string, then later checking that all of the flags conform to the allowed values, you can immediately tell from the type signature on the method exactly what is allowed here.

test/unit/analysis/timeseries_cleaning_test.py
```diff
@@ -67,6 +67,7 @@ def simulate_anomalies(
     return x.flat[indices] + values, indices
 
 
+@pytest.mark.xfail
```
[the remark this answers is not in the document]
zschira: I've fixed this now, it just relied on the old class which I've gotten rid of and needed to change to use the new implementation.

src/pudl/output/ferc714.py
```diff
@@ -648,16 +649,16 @@ def out_ferc714__summarized_demand(
         pd.merge(
             _out_ferc714__annualized_respondents,
             out_ferc714__hourly_planning_area_demand.loc[
-                :, ["report_date", "respondent_id_ferc714", "demand_mwh"]
+                :, ["report_date", "respondent_id_ferc714", "demand_imputed_mwh"]
```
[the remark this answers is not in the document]
zschira: Fixed

src/pudl/analysis/imputation/timeseries_cleaning.py
```diff
@@ -0,0 +1,1654 @@
...
+        matrix: pd.DataFrame,
+    ) -> tuple[pd.DataFrame, pd.DataFrame]:
+        """Flag/Null anomalous and missing values."""
+        matrix, flags = flag_ruggles(matrix)
+        matrix = filter_missing_values(matrix)
+        return matrix, flags
```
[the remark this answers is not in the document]
zschira: Marked as high-memory

src/pudl/analysis/imputation/timeseries_cleaning.py
zschira: Fixed

src/pudl/metadata/fields.py
zschira: Added

src/pudl/analysis/imputation/timeseries_cleaning.py
```diff
@@ -0,0 +1,1654 @@
...
+
+
+def impute_timeseries_asset_factory(
+    input_asset_name: str,
+    output_asset_name: str,
+    years: list[int],
+    value_col: str = "demand_mwh",
+    imputed_value_col: str = "demand_imputed_mwh",
+    reported_value_col: str = "demand_reported_mwh",
+    id_col: str = "",
```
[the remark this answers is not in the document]
zschira: I think making it required makes sense. Fixed.

src/pudl/analysis/imputation/codes.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@unique
+class ImputationReasonCodes(Enum):
+    """Defines all reasons a value might be flagged for imputation."""
+
+    ANOMALOUS_REGION = "Indicates that value is surrounded by flagged values."
+    NEGATIVE_OR_ZERO = "Indicates value is negative or zero value."
+    IDENTICAL_RUN = (
+        "Indicates value is among last values in an identical run of values."
```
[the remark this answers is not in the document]
zschira: Looking at the code, I think this actually flags all values after the first one in an identical run, in which case we should probably change this description to make that more clear. I'll try to test this to verify that that's the behavior first though.

Ultimately, it would probably be nice to have a unit test for every flagging function both to verify the behavior is as expected, and as documentation for what exactly each of these does.

src/pudl/analysis/imputation/timeseries_cleaning.py
zschira: Added

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -30,14 +30,210 @@
...
+Time zones are canonical names (e.g. 'America/Denver') from tzdata (
+https://www.iana.org/time-zones)
+mapped to their standard-time UTC offset.
+"""
+
+
+# --- Data structures --- #
+
+
+class UTCTimeseriesDataFrame(pa.DataFrameModel):
```
zschira, line 73: These pandera models are primarily for "documentation". They just allow me to add type annotations to various function signatures that give more insight to what is expected as the inputs/outputs than just `pd.DataFrame`

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -530,772 +726,932 @@ def impute_latc_tubal(  # noqa: C901
...
-        df = pd.concat(stats, names=["column", "flag"]).reset_index()
-        # Sort flags by flagged order
-        ordered = df["flag"].astype(pd.CategoricalDtype(set(self.flagged)))
-        return df.assign(flag=ordered).sort_values(["column", "flag"])
+        if overlap:
+            is_new_null &= ~is_null
+        new_nulls[:, col] = is_new_null
+    return new_nulls
 
-    def plot_flags(self, name: Any = 0) -> None:
```
zschira, line 1108: I removed this plot flags method because it seems likely to be obsolete by the analysis notebooks we're going to create, but can add a version back in if this is useful

zaneselvans (changes requested):
Let me know if you want to get on a call to talk through anything so we can get this merged in!

src/pudl/analysis/__init__.py
```diff
@@ -17,5 +17,4 @@
     service_territory,
     spatial,
     state_demand,
-    timeseries_cleaning,
```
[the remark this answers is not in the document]
zschira: Added back

src/pudl/analysis/state_demand.py
```diff
@@ -625,6 +181,9 @@ def out_ferc714__hourly_estimated_state_demand(
         Dataframe with columns ``state_id_fips``, ``datetime_utc``, ``demand_mwh``, and
         (if ``state_totals`` was provided) ``scaled_demand_mwh``.
     """
+    out_ferc714__hourly_planning_area_demand["year"] = (
+        out_ferc714__hourly_planning_area_demand.datetime_utc.dt.year
+    )
```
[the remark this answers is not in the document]
zschira: I've changed the docstring because the columns are different now, and since we're actually persisting the imputed table now you can just go look at the schema explicitly. This table doesn't have a report year column, though, and the way I'm deriving the year is consistent with the old intermediate imputed table.

src/pudl/analysis/state_demand.py
```diff
@@ -671,8 +230,8 @@ def out_ferc714__hourly_estimated_state_demand(
     )["weight"].sum()
     # Multiply respondent-state weights with demands
     df = weights.merge(
-        _out_ferc714__hourly_imputed_demand, on=["respondent_id_ferc714", "year"]
-    )
+        out_ferc714__hourly_planning_area_demand, on=["respondent_id_ferc714", "year"]
+    ).rename(columns={"demand_imputed_pudl_mwh": "demand_mwh"})
```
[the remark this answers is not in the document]
zschira: My thoughts were that while this is being derived from the imputed demand, it's not just imputed demand since this asset is doing other modifications/aggregations. Maybe there's a different name that would be more accurate?

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -30,14 +30,210 @@
...
+        Local times (tz-naive ``datetime64[ns]``).
+
+    Examples:
+        >>> s = pd.Series([pd.Timestamp(2020, 1, 1), pd.Timestamp(2020, 1, 1)])
+        >>> _utc_to_local(s, [-7, -6])
+        0   2019-12-31 17:00:00
+        1   2019-12-31 18:00:00
+        dtype: datetime64[ns]
+    """
+    return utc + pd.to_timedelta(utc_offset, unit="hours")
```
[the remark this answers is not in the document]
zschira: I've changed the naming conventions with all the timezone stuff to use the term 'aligned' datetimes, which is maybe more accurate to what we're actually doing than actually localizing them

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -30,14 +30,210 @@
...
+
+    x: np.ndarray
+    columns: pd.Index
+    index: pd.Index
+    flags: np.ndarray
+    uuid: str
+
+    def __hash__(self):
+        """Implement hash for lru_cache."""
+        return hash(self.uuid)
```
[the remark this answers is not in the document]
zschira: I think my implementation was consistent with the old class based one, since the `lru_cache` was just  looking at the other parameters to the method. That being said, I've changed the `flag` method to update the `uuid` any time it gets called, and that's the only place we modify the timeseries matrix, so now `lru_cache` will be aware of any modifications.

src/pudl/etl/static_assets.py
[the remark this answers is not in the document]
zschira: I've rearranged this a bit, so it's hopefully more consistent with other static tables. I'm still creating the enum for the type information, but I use that to generate a dataframe in `pudl.metadata.dfs`, so the actual table can be created in the same way as the other static tables.

src/pudl/analysis/state_demand.py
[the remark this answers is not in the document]
zaneselvans: Maybe `_estimated`? But also that's in the table name. I think it's fine this way.

src/pudl/analysis/timeseries_cleaning.py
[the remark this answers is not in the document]
zaneselvans: I did some additional search-and-replace on remaining `localized` variable & argument names and changed them to `aligned` and fixed a typo in the class name.  Re-running all the FERC-714 assets locally now to make sure they still work.

zaneselvans (approved):
I fixed a naming typo and renamed remaining `localized` variables and arguments to `aligned` and re-ran all the FERC-714 assets locally and they worked.

zaneselvans:
One funny thing I noticed is that I no longer seem to be able to get the imputation to use multiple threads, even if I want it to -- I bumped my local `OMP_NUM_THREADS` to 8 and it still seems to be using 1, or maybe 2.  Is there some other way that we're controlling the linear algebra parallelization?

zaneselvans:
I realized that we had not updated any of the row counts for `dbt` or added the new `core_ferc714__hourly_planning_area_demand` table schema, so I've tried to make those updates to the `dbt` tests.  Running the `pytest` workflow here to make sure it passes, and running the full ETL locally so I can check that the `dbt` tests pass before it hits the nightly builds.

zaneselvans:
Running the dbt tests on the full ETL outputs locally, I get no failures.

In CI on GitHub, even though I updated the row counts locally, I got 2 errors:

## `core_ferc714__hourly_planning_area_demand`
```
┌───────────┬────────────────┬────────────────┐
│ partition │ expected_count │ observed_count │
│  varchar  │     int32      │     int64      │
├───────────┼────────────────┼────────────────┤
│ 2022      │         784216 │          17544 │
│ 2023      │         779627 │         779571 │
└───────────┴────────────────┴────────────────┘
```

## `out_ferc714__hourly_estimated_state_demand`
```
┌───────────┬────────────────┬────────────────┐
│ partition │ expected_count │ observed_count │
│  varchar  │     int32      │     int64      │
├───────────┼────────────────┼────────────────┤
│ 2023      │     446478     │     446442     │
└───────────┴────────────────┴────────────────┘
```

Issue #4095, opened 2025-02-28
Migrate group_mean_continuity_check validation tests to DBT


aesharpe:

aesharpe:
Paused investigation on this in order to get the rest of the tests migrated

krivard:
Confirmed that:

* this is still happening with an otherwise-working Dagster env
* because they are currently implemented as Dagster asset checks, they're not blocking removal of pudl/test/validate

PR #4116, opened 2025-03-06
Add model/table-level macro for group_mean_continuity_check

krivard:
Alternate implementation of #4092. Would contribute to completion of #4095.

Where #4092 implements this check at the column level, this PR implements it at the model/table level, which significantly reduces the amount of duplicated information in the `schema.yml` files that use it... at the possible expense of having to build a pretty complicated query.

# Overview

Three tables require a [`group_mean_continuity_check`, which does the following](https://github.com/catalyst-cooperative/pudl/blob/e4bd17d8478ad44de0c7ce8175b3ae84a12b409d/src/pudl/validate.py#L163):

1. Group by a specified column
2. Compute the mean of each group
3. Compute the percent change between successive groups
4. Check against a per-column threshold

This PR contains a draft for a possible implementation.

## What did you change?

- New macro, `group_mean_continuity_check(ordered_group_column, thresholds, n_outliers_allowed)`
- Demo of macro in use for `_core_eia923__cooling_system_information`, using columns and thresholds from [`transform/eia923.py`](https://github.com/catalyst-cooperative/pudl/blob/7ab320421f2e08438469500feec9335f9e5cee86/src/pudl/transform/eia923.py#L1340)

### DIFF
--- a/dbt/macros/schema.yml
+++ b/dbt/macros/schema.yml
@@ -1,0 +1,28 @@
+version: 2
+
+macros:
+  - name: test_group_mean_continuity_check
+    description: >
+      Check that certain variables don't vary by too much. Groups the data by
+      ``ordered_group_column``, takes the mean of each group, then compares neighboring
+      means in ``ordered_group_column`` order. Useful for saying something like "the
+      average water usage of cooling systems didn't jump by 10x from 2012-2013."
+    arguments:
+      - name: ordered_group_column
+        type: string
+        description: The name of the column to group and order by (usually ``report_date`` or similar).
+      - name: thresholds
+        type: list
+        description: >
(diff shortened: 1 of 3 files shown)


dbt/macros/schema.yml
```diff
@@ -0,0 +1,28 @@
+version: 2
```
krivard, line 1: This is apparently the recommended way to document macros?

https://docs.getdbt.com/faqs/Docs/documenting-macros

aesharpe:
See my comment [here](https://github.com/catalyst-cooperative/pudl/pull/4092#issuecomment-2715856400)

krivard:
Dropping in favor of #4092!

PR #4117, opened 2025-03-06
Migrate correlation validation test

### LINKED ISSUE #4094: Migrate correlation validation tests to dbt
[Here](https://docs.google.com/spreadsheets/d/14N7C31d9SKzAf6BpwE9nYhpePiwI_JR1Bfwi34xH-Jg/edit?gid=1530740951#gid=1530740951) is a spreadsheet documenting all of our existing tests which need to be migrated to the new framework.

There's only one test in the correlation category: `test_fbp_ferc1_mmbtu_cost_correlation`

aesharpe:
# Overview

Closes #4094

Add a validation test for testing column correlation

## What did you change?

- Add a new macro generic test for column correlation
- Add that test to the `out_ferc1__yearly_steam_plants_fuel_by_plant_sched402` table `schema.yml`

# Documentation

```[tasklist]
```

# Testing

- run `dbt build --select source:pudl.out_ferc1__yearly_steam_plants_fuel_by_plant_sched402 --threads 1`


aesharpe:
@zaneselvans I want to improve our practice of justifying why we run certain tests on certain tables, and this feels like a good place to start. It might be OOS for this initial migration, but I was wondering why we are ok with 0.8 and not 0.9 correlation? (0.9 fails, I tried it). My best guess is that we are performing a sanity check on column that should be similar without delving into something more rigorous that would certainly fail and unveil hoards of little reporting issues that we can't necessarily fix. Maybe we should at least include a note that the purpose of this test is to perform a light-touch examination of data correlation to make sure nothing is wildly off. In an ideal world is there a more rigorous test we might perform here to check the data / highlight discrepancies? As we're going through and making these tests it can't hurt to document places where tests could be better for future reference.

krivard (changes requested):
This looks great! Would you add documentation for the new macro in `macros/schema.yml`?  That file doesn't exist yet, but here's how to do it:

https://docs.getdbt.com/faqs/Docs/documenting-macros

zaneselvans:
@aesharpe A lot of our data tests are pretty heuristic -- they're more like data regression tests that check whether things have changed from the way they were before.  IIRC, with this one in particular if you were to change the threshold from 0.8 to 0.9, that would mean you're making the test significantly more stringent -- rather than saying you expect at least a correlation of 0.8 between these variables, you expect a correlation of at least 0.9 (where 1.0 is the highest possible correlation) so I'm not surprised that it would fail. The 0.8 is just an observed value -- something that's slightly lower than the observed correlation that I saw when I made the test in the first place, so it'll pass, but also raise a red flag if something were to significantly decorrelate these variables (like a bad merge or bad data cleaning that we introduced accidentally)

dbt/tests/generic_tests/schema.yml
```diff
@@ -0,0 +1,52 @@
...
+    arguments:
+      - name: column1
+        type: string
+        description: The name of the first column
+      - name: column2
+        type: string
+        description: The name of the second column
+      - name: correlation_coef
+        type: float
+        description: The correlation coefficient to check for
```
krivard, line 15: Nonblocking:
Is this an exact value or a threshold? & if a threshold, which direction?

Maybe something like 
```suggestion
        description: The minimum correlation coefficient we expect between the two columns.
```

dbt/dbt_project.yml
```diff
@@ -11,9 +11,9 @@ profile: "pudl_dbt"
 # The `model-paths` config, for example, states that models in this project can be
 # found in the "models/" directory. You probably won't need to change these!
 model-paths: ["models"]
-macro-paths: ["macros"]
+macro-paths: ["macros", "tests/generic_tests"]
```
krivard, line 14: Possibly blocking: the current schematic has this as
```suggestion
macro-paths: ["macros", "tests/data_tests/generic_tests"]
```

dbt/dbt_project.yml
```diff
@@ -11,9 +11,9 @@ profile: "pudl_dbt"
...
 seed-paths: ["seeds"]
-test-paths: ["tests"]
+test-paths: ["tests/singular_tests"]
```
krivard, line 16: Possibly blocking: the current schematic places singular tests in `tests/data_tests/generic_tests`

But also I think we still want dbt to be able to find the unit tests as well, which the schematic places in `tests/unit_tests`

So probably the original 
```suggestion
test-paths: ["tests"]
```
is superior?

krivard (approved):
Self-consistent, but doesn't match the spec. Can merge as-is if we expect to readjust later once the spec is finalized.

dbt/dbt_project.yml
aesharpe: Will it be an issue if the test path is `test` while the macro path is `tests/data_tests/generic_tests`? Will that overlap cause any problems? I.e.: the generic tests getting read twice?

docs/release_notes.rst
```diff
@@ -21,6 +21,18 @@ Major Dependency Updates
...
+^^^^^^^^^
+We're in the process of migrating our tests to use the
+`dbt <https://docs.getdbt.com/docs/introduction>`__ framework.
+So far we have created the following tests:
+
+* ``check_column_correlation`` – a more generic replacement for the old
+  ``test_fbp_ferc1_mmbtu_cost_correlation`` pytest.
+  See :issue:`4094`, :pr:`4117`. You can find the implementation in the
+  `check_column_correlation.sql
+  <../../dbt/tests/data_tests/check_column_correlation.sql>`__ file.
```
krivard, line 34:
```suggestion
  <../../dbt/tests/data_tests/generic_tests/check_column_correlation.sql>`__ file.
```

docs/release_notes.rst
```diff
@@ -21,6 +21,18 @@ Major Dependency Updates
 Quality of Life Improvements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+New Tests
+^^^^^^^^^
+We're in the process of migrating our tests to use the
+`dbt <https://docs.getdbt.com/docs/introduction>`__ framework.
+So far we have created the following tests:
```
krivard, line 28:
```suggestion
So far we have converted the following tests:
```

or "moved" or "moved over"

krivard (commented):
path fix and optional wording adjustment

Issue #4119, opened 2025-03-07
Write `dbt` YAML files the way `prettier` likes them


zaneselvans:
# Overview

When updating some of the `dbt` YAMl files using the `dbt_helper.py` script I noticed that the YAML formatting which the script uses is different from what is enforced by `prettier` which we use to ensure standardized formatting in all of our YAML files via a `pre-commit` hook. This means that every update results in a large number of meaningless diffs, and makes it hard to see what actually changed.

Can we update the script to use the same indentation / formatting so there isn't guaranteed to be a diff that pre-commit has to fix, even if there haven't been any changes to the schema? It might also be good to use sorted keys to avoid diffs from just re-ordering? This might mean using another YAML library beyond the `pyyaml` one we're apparently using now.?

## Success Criteria

* [x] Using the `devtools/dbt_helper.py` script to update table schemas or tests doesn't automatically result in the `prettier` pre-commit hook needing to fix things.
* [ ] Diffs in YAML files are minimized, making it easy to identify only the meaningful changes before the pre-commit hooks run.

kalexandru:
Hi @zaneselvans, nice to meet you!
If it is possible I would like to help with this issue, you could assign it to me!

jdangerx:
Sorting the keys means we can't order keys for, like... semantic meaning. And the diffs seemed still pretty sensible and small. So I didn't do that in #4401 and I would argue that that's fine.

Issue #4065, opened 2025-02-14
Simplify sec10k integration and testing


zaneselvans:
# Overview
 
- The `sec10k` tables are only loaded when the `USE_PUDL_MODELS` environment variable is set.
- Currently it is not set in the `pytest` workflow in our CI, so the `sec10k` assets aren't tested by our integration tests when a PR gets put in the merge queue.
- If `USE_PUDL_MODELS` is set in our CI, the runner doesn't have permission to pull the `sec10k` assets, so the tests fail.
- Because there are `utility_id_eia` values from a wide range of years in the `out_sec10k__` tables, and all of them are loaded any time the asset is loaded, in the `etl_fast` integration tests, there are FK constraint violations -- since only a few years of EIA-860 data is present in the database.
- This means that if you have `USE_PUDL_MODELS` set locally, you cannot get `make pytest-integration` to pass, and `pytest-integration` only passes in CI because the `sec10k` assets aren't being tested.
- The permissions on the GCS bucket where this data is coming from also means that no open source contributor can access the data / run the ETL with it, and its provenance is murky.
- Much of this complexity can be avoided if we integrate the `sec10k` with PUDL using the same tooling.

# Success Criteria

- SEC 10-K tables are always processed in local ETL and tests, CI on GitHub, and in the nightly builds.
- SEC 10-K tables don't cause integration tests to fail either when running full or fast ETL.
- New versions of the SEC 10-K tables are pulled into PUDL as part of our monthly/quarterly updates if we re-run the upstream data pipeline that gathers and processes SEC 10-K data.

# Next Steps
 
- [x] Create a new `pudl-archiver` that pulls the SEC 10-K data from GCS and uploads it to Zenodo
- [x] Update the SEC 10-K ETL in the main PUDL repo to rely on that Zenodo archive instead.
- [x] Move the SEC 10-K processing into extract/transform/output modules & an `sec10k` asset group, using the same patterns as our other datsets.
- [x] Remove the `USE_PUDL_MODELS` dependencies and other SEC 10-K special cases wherever they exist.
- [x] Add `sec10k` to the PUDL ETL Settings classes / files.
- [x] Address failing FK constraint in Fast ETL either by restricting date ranges to match EIA-860 utilities, or temporarily removing the FK constraint

Issue #3418, opened 2024-02-22
Address various deprecation warnings


zaneselvans:
There were a bunch of deprecation warnings that didn't get addressed in #3272 that we should probably get to at some point, so here is a todo list. Many but not all of these are related to upcoming changes in Pandas 3.0.  This isn't an exhaustive catalog -- just an example of each deprecation warning I saw in the output from `make pytest-coverage`

```[tasklist]
- [ ] `/Users/zane/code/catalyst/pudl/src/pudl/transform/ferc1.py:5346: FutureWarning:` The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation. 
- [ ] `/Users/zane/code/catalyst/pudl/src/pudl/transform/ferc1.py:3897: FutureWarning:` Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value 'header' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
- [ ] `/Users/zane/code/catalyst/pudl/src/pudl/output/ferc1.py:952: DeprecationWarning:` AssetExecutionContext.op_config is deprecated and will be removed in 1.8.0. You have called the deprecated method op_config on AssetExecutionContext. Use context.op_execution_context.op_config instead.
- [ ] `/Users/zane/code/catalyst/pudl/src/pudl/analysis/fuel_by_plant.py:126: FutureWarning:` The previous implementation of stack is deprecated and will be removed in a future version of pandas. See the What's New notes for pandas 2.1.0 for details. Specify future_stack=True to adopt the new implementation and silence this warning.
- [ ] `/Users/zane/code/catalyst/pudl/src/pudl/transform/eia861.py:1464: FutureWarning:` Downcasting object dtype arrays on .fillna, .ffill, .bfill is deprecated and will change in a future version. Call result.infer_objects(copy=False) instead. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
- [ ] `/Users/zane/code/catalyst/pudl/src/pudl/transform/eia861.py:1197: FutureWarning:` Downcasting behavior in `replace` is deprecated and will be removed in a future version. To retain the old behavior, explicitly call `result.infer_objects(copy=False)`. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
[...]

PR #4122, opened 2025-03-11
Run dbt data tests in CI / integration tests

### LINKED ISSUE #4102: Integrate `dbt` testing into CI/nightly builds
One of our goals with the migration to using `dbt` for our data validations is to have those validations run and work on the fast ETL, which only processes a couple of years of data for each dataset. This is the ETL that we use in our integration tests, which must pass before any PR gets merged into `main`.

## PUDL pytest uses temporary `$PUDL_OUTPUT` directory

The `dbt` tests currently read their data out of the Apache Parquet outputs generated by the ETL, and the ETL that is run in our integration tests uses a temporary directory for its `$PUDL_OUTPUT`. This directory is defined in `test/conf.py` in a fixture named `configure_paths_for_tests()`

There's also a `pudl_io_manager` fixture defined in `test/conf.py` which should work for reading data out of any Parquet files output by the tests.

Both of these `pytest` fixtures have "session" scopes so the dbt tests would need to be run as part of the same `pytest` command that kicks off the integration tests.

Hopefully this means that `dbt` will automatically look in the right place to find the Parquet outputs generated by the CI ETL, and will drop the failures DB there as well.

## Implementation

- Since this is meant to be an interim solution until we can get the `dbt` tests running as asset checks, really we want the simplest solution possible.
- We can call the `dbt` CLI programmatically from a new test module that depends on the `pudl_io_manager` fixture (so it runs after the ETL is complete)
- It will need to be aware of whether it's running against the Fast or Full ETL outputs, since we use these tests both in CI and in the nightly builds.
- To save the failing records as an artifact we can use for debugging we'll need to stash that DB in a non-temporary directory after the `dbt` tests have completed.

## Success criteria
- [x] All of the data validation tests that are defined using `dbt` get run in CI before a PR merges into `main`
- [x] If there are failing `dbt` data validations, the PR is rejected
- [x] Any `dbt` logging output is captured/displayed in CI for debugging.
- [x] The results of `dbt` test failures are saved into `$PUDL_OUTPUT/pudl_dbt.duckdb` for investigation (locally)
- [x] The `pudl_dbt_tests.duckdb` output is saved as a GitHub Actions artifact if run in CI.
- [x] The `pudl_dbt_tests.duckdb` output is saved alongside everything else in `gs://builds.catalyst.coop` for debugging the nightly builds.

## Out of Scope
- Migrating any additional tests into the `dbt` framework.
- Running the `dbt` data validation tests in `@asset_check` associated with individual assets in the DAG.

zaneselvans:
# Overview

Ensure that all of the data validation tests we've defined for `dbt` are run during our integration tests, and thus prior to any PR getting merged into `main`.

- Updated `dbt` dependencies / environment
- Switched to using a new maintained [fork of dbt_expectations](https://github.com/metaplane/dbt-expectations) and [dbt_date](https://github.com/godatadriven/dbt-date)
- Added a new `test/inegration/dbt_test.py` module that uses the `dbt` [programmatic invocation API](https://docs.getdbt.com/reference/programmatic-invocations) to run all of our `dbt` data tests.
- The new test is currently marked `xfail` because there are a couple of outstanding test issues that need to be fixed (see below)

Closes #4102

# Outstanding Issues to create
To be addressed after this PR is merged:
- the distribution of gas capacity factors is not what we expect. There are 3-4 real weighted quantile test failures.
- The SEC 10-K tables cannot be loaded in CI on GitHub, leading to dbt test errors (missing tables)
- When the SEC 10-K tables **are** present for integration tests run against the Fast ETL outputs (e.g. locally) they result in FK constraint failures (because they reference EIA Utilities that only appear in years that are not part of the Fast ETL)
- Running the `dbt` tests in multi-threaded mode sometimes results in segmentation faults.
- Why did the expected row count for `out_ferc714__hourly_planning_area_demand` (2022) drop so much?
- Why is there a null-year row count expectation in the Plant Parts EIA table?
- Improve `dbt` failure database usability (if it's easy)

# To-do list
- [x] Tell `dbt` to save the failing records in DuckDB
- [x] Improve signal-to-noise ratio of logging output -- highlight just the failing `dbt` tests.
- [x] Verify that the `dbt` tests are actually run in CI.
- [x] Make sure that the logging output from the `dbt` tests is available.
- [x] Save the failing records database as a GitHub Actions artifact that can be downloaded for inspection.
- [x] Switch back to running ETL and testing actual outputs.
- [x] Update expected row counts for the fast ETL using pristine local outputs.
- [x] Verify that the failure DB is saved from the builds as expected.
- [x] Run `build-deploy-pudl` GitHub Action on the branch to verify that `dbt` tests pass/fail as expected on full etl


zaneselvans:
- I can run the `dbt` tests from within pytest.
- I've disabled the conditional that keeps the integration tests from running on push for testing.
- Currently running `dbt` tests against `nightly` target to avoid running the ETL in CI
- Only running tests for a single model `out_eia__yearly_generators` -- 4 pass, and 2 fail.
- Currently unable to upload the `pudl_dbt_tests.duckdb` output because it's in the pytest temporary directory.
- I can't know what the temporary directory is from within the GHA workflow file because it's set dynamically by pytest.
- Either need to pass that information out from pytest, or move the output db to a fixed location after the tests run.

zaneselvans:
The mechanics of this all seems to be working now!

## `dbt` tests that fail locally (35)

- Most of these `dbt` test failures are FERC-1 and FERC-714 tables that need to have their Fast ETL row counts updated.
- It would be helpful to show the expected + actual row counts / percentage delta in the logging so we can at a glance whether there's a big problem that needs investigating or a tiny change that we can probably ignore and update the row counts.
- Then there are 4 distribution checks that seem like real novel failures, looking at gas capacity factors in both the annual and monthly generator tables. Seems like those could be real, in that we expect different distributions in recent data compared to all data.

```
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_balance_sheet_assets_sched110_report_year__core_ferc1__yearly_balance_sheet_assets_sched110: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_balance_sheet_liabilities_sched110_report_year__core_ferc1__yearly_balance_sheet_liabilities_sched110: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_depreciation_by_function_sched219_report_year__core_ferc1__yearly_depreciation_by_function_sched219: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_depreciation_changes_sched219_report_year__core_ferc1__yearly_depreciation_changes_sched219: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_energy_dispositions_sched401_report_year__core_ferc1__yearly_energy_dispositions_sched401: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_income_statements_sched114_report_year__core_ferc1__yearly_income_statements_sched114: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_operating_expenses_sched320_report_year__core_ferc1__yearly_operating_expenses_sched320: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_operating_revenues_sched300_report_year__core_ferc1__yearly_operating_revenues_sched300: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_other_regulatory_liabilities_sched278_report_year__core_ferc1__yearly_other_regulatory_liabilities_sched278: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_plant_in_service_sched204_report_year__core_ferc1__yearly_plant_in_service_sched204: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_retained_earnings_sched118_report_year__core_ferc1__yearly_retained_earnings_sched118: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_sales_by_rate_schedules_sched304_report_year__core_ferc1__yearly_sales_by_rate_schedules_sched304: fail
source_check_row_counts_per_partition_pudl_core_ferc1__yearly_utility_plant_summary_sched200_report_year__core_ferc1__yearly_utility_plant_summary_sched200: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_balance_sheet_assets_sched110_report_year__out_ferc1__yearly_balance_sheet_assets_sched110: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_balance_sheet_liabilities_sched110_report_year__out_ferc1__yearly_balance_sheet_liabilities_sched110: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_depreciation_by_function_sched219_report_year__out_ferc1__yearly_depreciation_by_function_sched219: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_depreciation_changes_sched219_report_year__out_ferc1__yearly_depreciation_changes_sched219: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_detailed_balance_sheet_assets_report_year__out_ferc1__yearly_detailed_balance_sheet_assets: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_detailed_balance_sheet_liabilities_report_year__out_ferc1__yearly_detailed_balance_sheet_liabilities: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_detailed_income_statements_report_year__out_ferc1__yearly_detailed_income_statements: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_energy_dispositions_sched401_report_year__out_ferc1__yearly_energy_dispositions_sched401: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_income_statements_sched114_report_year__out_ferc1__yearly_income_statements_sched114: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_operating_expenses_sched320_report_year__out_ferc1__yearly_operating_expenses_sched320: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_operating_revenues_sched300_report_year__out_ferc1__yearly_operating_revenues_sched300: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_other_regulatory_liabilities_sched278_report_year__out_ferc1__yearly_other_regulatory_liabilities_sched278: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_plant_in_service_sched204_report_year__out_ferc1__yearly_plant_in_service_sched204: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_rate_base_report_year__out_ferc1__yearly_rate_base: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_retained_earnings_sched118_report_year__out_ferc1__yearly_retained_earnings_sched118: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_sales_by_rate_schedules_sched304_report_year__out_ferc1__yearly_sales_by_rate_schedules_sched304: fail
source_check_row_counts_per_partition_pudl_out_ferc1__yearly_utility_plant_summary_sched200_report_year__out_ferc1__yearly_utility_plant_summary_sched200: fail
source_check_row_counts_per_partition_pudl_out_ferc714__hourly_planning_area_demand_report_date__out_ferc714__hourly_planning_area_demand: fail
source_expect_column_weighted_quantile_values_to_be_between_pudl_out_eia__monthly_generators_capacity_factor__0_7__0_65__fuel_type_code_pudl_gas_and_report_date_CAST_2015_01_01_AS_DATE_and_capacity_factor_0_0__capacity_mw: fail
source_expect_column_weighted_quantile_values_to_be_between_pudl_out_eia__monthly_generators_capacity_factor__0_95__0_95__fuel_type_code_pudl_gas_and_report_date_CAST_2015_01_01_AS_DATE_and_capacity_factor_0_0__capacity_mw: fail
source_expect_column_weighted_quantile_values_to_be_between_pudl_out_eia__yearly_generators_capacity_factor__0_7__0_65__fuel_type_code_pudl_gas_and_report_date_CAST_2015_01_01_AS_DATE_and_capacity_factor_0_0__capacity_mw: fail
source_expect_column_weighted_quantile_values_to_be_between_pudl_out_eia__yearly_generators_capacity_factor__0_95__0_95__fuel_type_code_pudl_gas_and_report_date_CAST_2015_01_01_AS_DATE_and_capacity_factor_0_0__capacity_mw: fail

```

## Additional `dbt` tests that fail in CI (39)
These are all SEC 10-K failures that are happening because of the non-standard way those tables are brought in. We need to integrate those tables the normal way.
```
source_check_row_counts_per_partition_pudl_core_sec10k__quarterly_company_information_report_date__core_sec10k__quarterly_company_information: error
source_check_row_counts_per_partition_pudl_core_sec10k__quarterly_exhibit_21_company_ownership_report_date__core_sec10k__quarterly_exhibit_21_company_ownership: error
source_check_row_counts_per_partition_pudl_core_sec10k__quarterly_filings_report_date__core_sec10k__quarterly_filings: error
source_check_row_counts_per_partition_pudl_out_sec10k__parents_and_subsidiaries_report_date__out_sec10k__parents_and_subsidiaries: error
```

zaneselvans:
In trying to update the row counts I found a number of different issues.

## plant parts rows without years?

Looking at the row counts for the fast tests, the plant parts association table has a row with no partition value and an expected 3 rows. That seems weird? Are there rows with no report year?
```
out_pudl__yearly_assn_eia_ferc1_plant_parts,2020,1890
out_pudl__yearly_assn_eia_ferc1_plant_parts,2021,1968
out_pudl__yearly_assn_eia_ferc1_plant_parts,2023,2114
out_pudl__yearly_assn_eia_ferc1_plant_parts,,3
```

## Dagster UI + CI ETL produce different FERC outputs

The `ferc1` row count failures were identical in CI and when I ran `make pytest-integration` locally. To generate the new row counts I re-materialized all my local ferc1 output tables in Dagster under the `etl_fast` job. Then when I ran the helper script, there were no changes, except to the (I think) recently updated or added `out_ferc1__yearly_detailed_income_statements` table -- the other ~30 tables didn't have any change in row counts.  Somehow the outputs from `etl_fast` in the Dagster UI is different from what's being generated by the integration test ETL.

While investigating this, I noticed that the row count expectations for the FERC 1 tables in the fast ETL contained different years than the fast ETL settings for FERC 1.  The row counts have 2020-2023, but the FERC 1 fast ETL settings have 2019, 2020, 2021, 2023 (no 2022)

Maybe there were some stale ETL settings in the Launchpad of whoever generated the fast ETL row counts? I thought we had changed that to always refresh directly from the settings file, but when I went to try and re-run my fast FERC to SQLite ETL, my settings did not reflect the settings in the YAML file on disk and I had to actively clobber them.

dbt/seeds/etl_fast_row_counts.csv
```diff
@@ -971,15 +971,15 @@ out_ferc1__yearly_transmission_lines_sched422,2020,26407
...
+out_ferc1__yearly_utility_plant_summary_sched200,2021,6668
 out_ferc1__yearly_utility_plant_summary_sched200,2023,6768
 out_ferc714__hourly_estimated_state_demand,2020,447699
 out_ferc714__hourly_estimated_state_demand,2021,322
 out_ferc714__hourly_estimated_state_demand,2023,446478
 out_ferc714__hourly_estimated_state_demand,2024,323
 out_ferc714__hourly_planning_area_demand,2020,834384
-out_ferc714__hourly_planning_area_demand,2022,784216
-out_ferc714__hourly_planning_area_demand,2023,779627
+out_ferc714__hourly_planning_area_demand,2022,17544
```
zaneselvans, line 981: This is a very large change. What is going on here?

dbt/seeds/etl_fast_row_counts.csv
```diff
@@ -1000,10 +1000,10 @@ out_gridpathratoolkit__hourly_available_capacity_factor,2018,227760
 out_gridpathratoolkit__hourly_available_capacity_factor,2019,227760
 out_gridpathratoolkit__hourly_available_capacity_factor,2020,228384
 out_gridpathratoolkit__hourly_available_capacity_factor,2021,234
-out_pudl__yearly_assn_eia_ferc1_plant_parts,,3
 out_pudl__yearly_assn_eia_ferc1_plant_parts,2020,1890
 out_pudl__yearly_assn_eia_ferc1_plant_parts,2021,1968
 out_pudl__yearly_assn_eia_ferc1_plant_parts,2023,2114
+out_pudl__yearly_assn_eia_ferc1_plant_parts,,3
```
zaneselvans, line 1006: Why does this row with no year exist?

devtools/dbt_helper.py
```diff
@@ -198,8 +198,10 @@ def generate_row_counts(
     new_row_counts = duckdb.sql(row_count_query).df().astype({"partition": str})
     new_row_counts["table_name"] = table_name
 
-    all_row_counts = pd.concat([row_counts_df, new_row_counts]).drop_duplicates(
-        subset=["partition", "table_name"], keep="last"
+    all_row_counts = (
+        pd.concat([row_counts_df, new_row_counts])
+        .drop_duplicates(subset=["partition", "table_name"], keep="last")
+        .sort_values(["table_name", "partition"])
```
zaneselvans, line 204: Keep this file sorted to minimize diffs.

src/pudl/package_data/settings/etl_fast.yml
```diff
@@ -11,15 +11,15 @@ ferc_to_sqlite_settings:
   ferc2_dbf_to_sqlite_settings:
     years: [2019, 2020]
   ferc2_xbrl_to_sqlite_settings:
-    years: [2021, 2022]
+    years: [2021, 2023]
```
zaneselvans, line 14: Keep these years synchronized with the FERC 1 & FERC 714 so that when we're testing them we actually know if the most recent year works.

pyproject.toml
```diff
@@ -292,6 +292,9 @@ filterwarnings = [
     "ignore:pkg_resources is deprecated as an API.:DeprecationWarning",
     "ignore:Subclassing validator classes is not intended to be part of their public API.:DeprecationWarning",
     "ignore:Subclassing validator classes:DeprecationWarning:tableschema",
+    "ignore:datetime.datetime.utcnow() is deprecated:DeprecationWarning:dbt_common",
+    "ignore:datetime.datetime.utcnow() is deprecated:DeprecationWarning:dbt",
+    "ignore:Python 3.14 will, by default, filter extracted tar archives:DeprecationWarning:dbt_common",
```
zaneselvans, line 297: The warnings in our test logging output are getting out of control and making it hard to see wtf is actually going on.  We need to address various deprecation warnings #3418 and also silence the ones that aren't our problem.

dbt/package-lock.yml
```diff
@@ -1,8 +1,8 @@
...
+  - package: metaplane/dbt_expectations
+    version: 0.10.8
   - package: dbt-labs/dbt_utils
     version: 1.3.0
-  - package: calogica/dbt_date
-    version: 0.10.1
-sha1_hash: 29571f46f50e6393ca399c3db7361c22657f2d6b
+  - package: godatadriven/dbt_date
+    version: 0.11.0
+sha1_hash: 984baac0716f78c8f3a7d03a7ff41ba9f609ae21
```
zaneselvans, line 8: Oh great, both the calogica packages got picked up by someone.

test/integration/dbt_test.py
```diff
@@ -0,0 +1,101 @@
...
+                "1",
+                "--target",
+                dbt_target,
+            ]
+        )
+        test_result: dbtRunnerResult = dbt.invoke(
+            [
+                "test",
+                "--threads",
+                "1",
```
zaneselvans, line 80: It would be nice to run this with more threads, but I was getting segfaults.

zaneselvans:
@jdangerx Assuming this row count update works, there's just a handful of distribution checks (which are likely real changes) and then the SEC 10-K tables that need to be addressed to get everything passing in CI.  I guess we should still get it merged in ASAP (Friday) with the dbt test marked xfail so we don't break CI or the nightly builds, and then move on to fixing those remaining issues so we aren't digging the hole any deeper.  I will kick off a branch build now so we can see how it breaks the nightly builds and whether we can grab the output database.

jdangerx (changes requested):
Mostly looks good! A little cruft from your multi-threaded experiments but nothing big. 

One thought I had is that we should pull out *specifically* the natural gas + SEC 10K tests into a different `pytest` test and `xfail` that. Then we can keep the general tests running and eXpect success, to stop further failing tests from cropping up.

> What documentation do you think we should update to go along with these changes? Probably something under development / running the tests? We'll need to be much better about folks updating all the relevant test expectations both for full & fast outputs, otherwise we're gonna get a lot of broken nightly builds.

The act of running the tests remains the same - we'll have to write about how to update the test expectations in the [data validation section](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation).

Could we automate a check for "hey, you touched this table, certain validation tests/dbt schemas should also have been touched," and then run it as part of `pre-commit`?

In any case I don't think fixing the problem of "people will forget to / not know how to deal with this new CI" is blocking this PR. Everyone is aware that the data validation stuff is happening and in flux. Nobody will be surprised by this being weird for a bit, and through our collective struggle we'll find out a good path forward.

Here's an accumulated TODO list for after merge:
* [ ] #4129 
* [ ] #4130 
* [ ] #4065 
* [ ] #4126 
* [ ] #4125 
* [ ] #4131
* [ ] #4119
* [ ] #4132 
* [ ] add something to the row counts generator that alerts you when there's weird row count changes - so we don't have to rely on :eyes:
* [ ] get better/more useful output when a dbt test fails

test/integration/dbt_test.py
```diff
@@ -0,0 +1,104 @@
...
+from dbt.cli.main import dbtRunner, dbtRunnerResult
+from pudl.io_managers import PudlMixedFormatIOManager
+
+logger = logging.getLogger(__name__)
+
+# Ensure that httpfs is installed before doing any multi-threaded dbt operations
+# This prevents errors where multiple threads attempt to install the extension, and
+# all but the first one finds that it's mysteriously already been installed. Only
+# important if we are running dbt using multiple threads.
+duckdb.execute("FORCE INSTALL httpfs")
```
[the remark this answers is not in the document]
zaneselvans: I left it in because it fixes one of the two issues that came up with multi-threaded execution, and it would be nice not to have to figure it out again if/when we get multi-threaded execution working, which would be really nice (and is really just totally broken for everyone? Seems kind of amazing.) Agree that it should be part of the overall setup.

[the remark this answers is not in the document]
jdangerx: Sure, let's have it be part of the overall setup and leave it in.

test/integration/dbt_test.py
```diff
@@ -0,0 +1,104 @@
...
+    The dependency on pudl_io_manager is necessary because it ensures that the dbt
+    tests don't run until after the ETL has completed and the Parquet files are
+    available.
+
+    See https://docs.getdbt.com/reference/programmatic-invocations/ for more details on
+    how to invoke dbt programmatically.
+    """
+    # Identify whether we're running the full or fast ETL, and set the dbt target
+    # appropriately (since we have different test expectations in the two cases)
+    if request.config.getoption("--etl-settings"):
```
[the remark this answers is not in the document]
zaneselvans: Almost anything that cares about the contents of the data needs to know which version it's going to be looking at. In theory there's much more flexibility in our ETL settings than just full vs. fast, but in practice those are the only two sets of input parameters that we test and use day to day and actually have a reasonable expectation of working, though the specification of each of them evolves over time (making use of that flexibility).

I think that's going to keep being the case for the foreseeable future, so having a less janky way to communicate which data environment we're working in across the board would be great.

Maybe in conjunction with that... having separate fast/full output locations would be really nice, especially if we're going to have to run both of them locally to generate fresh row counts or other data testing parameters.

zaneselvans:
> As in, this flag doesn't actually work?

Can't respond directly to your comment about the anonymous usage phone-home, but... that was my impression.  The DEBUG logging statements that showed up indicating connections back to dbtlabs kept coming up even after I set that parameter.  Wasn't sure if maybe I was doing something wrong though.

dbt/dbt_project.yml
```diff
@@ -7,6 +7,10 @@ version: "1.0.0"
 # This setting configures which "profile" dbt uses for this project.
 profile: "pudl_dbt"
 
+# Opt out of telemetry
+flags:
+  send_anonymous_usage_stats: false
```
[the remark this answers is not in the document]
zaneselvans: Can't respond directly to your comment about the anonymous usage phone-home, but... that was my impression. The DEBUG logging statements that showed up indicating connections back to dbtlabs kept coming up even after I set that parameter. Wasn't sure if maybe I was doing something wrong though.

test/integration/dbt_test.py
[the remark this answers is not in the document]
zaneselvans: Okay, moved into `conftest.py`

dbt/models/output/out_eia__monthly_generators/schema.yml
```diff
@@ -39,6 +39,7 @@ sources:
                   max_value: 0.7
                   row_condition: fuel_type_code_pudl='gas' and report_date>=CAST('2015-01-01' AS DATE) and capacity_factor<>0.0
                   weight_column: capacity_mw
+                  severity: warn
```
zaneselvans, line 42: The tests will still get run, but we'll get a yellow warning rather than a red failure.

dbt/profiles.yml
```diff
@@ -2,18 +2,18 @@ pudl_dbt:
   outputs:
     # Define targets for nightly build outputs, and local full/fast ETL outputs
     nightly:
-      threads: 4
+      threads: 1
```
zaneselvans, line 5: Default to single threaded execution until/unless we can figure out how to work around segfaults.

test/conftest.py
```diff
@@ -49,6 +50,13 @@
     "gen_fuel_by_generator_eia923",
 ]
 
+# In general we run our tests and some subprocesses using more than one thread, and
+# sometimes we access remove HTTPS / S3 resources. When this happens it's possible
+# for DuckDB to get confused about whether the httpfs extension is installed and error
+# out if one processes is trying to install it after another one already has. Doing
+# this forced installation during setup avoids that issue.
+duckdb.execute("FORCE INSTALL httpfs")
+
```
zaneselvans, line 59: Moved here so it applies to all tests, and made comment more generic.

test/integration/dbt_test.py
```diff
@@ -0,0 +1,114 @@
...
+                "--threads",
+                "1",
+                "--store-failures",
+                "--target",
+                dbt_target,
+            ]
+            + [
+                f"--exclude {x}"
+                for x in SEC10K_EXCLUDE
+                if os.getenv("GITHUB_ACTIONS", False)
```
zaneselvans, line 96: Exclude these tables only when we are running in GitHub Actions CI, because they do exist and should pass their tests in the nightly builds.

zaneselvans:
## Brittle test updates

@jdangerx adding to your list, another potential followup issue to consider is figuring out a less fragile way to define the regression testing expectations -- as it is right now, the row counts (or whatever) are going to be derived from whatever junk data people happen to have laying around in `PUDL_OUTPUT` which as we know can be a very heterogenous mix, depending on whatever assets have been materialized recently and could represent outputs from many different branches / versions of PUDL.  I suspect that this will become a source of frustration as the number of data tests grows.  We'll catch fast ETL issues in the merge queue, but full ETL issues will break the nightly builds.

But maybe this is just going to be annoying during the transition period when we're running these tests in CI, rather than as asset checks?  In which case getting the asset checks up and running seems like a pretty high priority.

## Less xfail

Agree that it would be great to just xfail the actually xfailing `dbt` tests! I thought about that but didn't dig in deeply.  It looks like there's a `--exclude` option that behaves the same way as `--select` so I will break out the two sets of expected failures and exclude them from the main tests.  There's also the possibility of setting some specific tests to `WARN` instead of `ERROR` on failure.

## Schema Drift

> Could we automate a check for "hey, you touched this table, certain validation tests/dbt schemas should also have been touched," and then run it as part of pre-commit?

We have talked about adding a unit test that fails any time the PUDL table schemas are not mirrored by the corresponding `dbt` `schema.yml` files but I'm not currently seeing it written up anywhere.  This seems like an inframundo task.

test/conftest.py
```diff
@@ -49,6 +50,13 @@
     "gen_fuel_by_generator_eia923",
 ]
 
+# In general we run our tests and some subprocesses using more than one thread, and
+# sometimes we access remove HTTPS / S3 resources. When this happens it's possible
```
jdangerx, line 54:
```suggestion
# sometimes we access remote HTTPS / S3 resources. When this happens it's possible
```

test/integration/dbt_test.py
```diff
@@ -0,0 +1,111 @@
...
+from contextlib import chdir
+from pathlib import Path
+
+from dbt.cli.main import dbtRunner, dbtRunnerResult
+from pudl.io_managers import PudlMixedFormatIOManager
+
+logger = logging.getLogger(__name__)
+
+
+# These tables need to be excluded from our GitHub CI until they are being broght into
```
jdangerx, line 13:
```suggestion
# These tables need to be excluded from our GitHub CI until they are being brought into
```

jdangerx (approved):
Couple typo fixes if ya want em. :shipit:

zaneselvans:
It turns out that `dbt build` also calls `dbt test` (and `dbt seed`) and I had only excluded the SEC 10-K tables from my invocation of `dbt test` so the `dbt build` failed.

Updated the test to call only `dbt build` with the exclusion of SEC 10-K tables on GitHub.

zaneselvans:
For reasons I don't understand the previous attempt to merge got None as the result from `dbtRunner` when the tests apparently passed, and I wasn't able to reproduce it locally. Blargh.

zaneselvans:
@zschira  My god I have already spent multiple hours dealing with the special case that the non-standard access to the SEC 10-K tables requires. We've got to get that into the normal Zenodo archives.

Issue #4086, opened 2025-02-25
Restructure SEC company ownership tables


katie-lamb:
Conduct restructuring on the SEC company ownership tables extracted from Exhibit 21 attachments based on discussions with the team.

**core_sec10k__quarterly_exhibit_21_company_ownership**
This is a minimal (skinny) core table that represents the ownership information pulled from Ex. 21 attachments.

* The CIK of the parent company for each of the subsidiaries listed in this table is the CIK associated with the filename that the information is pulled from (associations between CIK and filename in the core_sec10k__quarterly_filings table)
* Parent company CIK (and maybe company name) should be included in this table
* This is a complete time series of ownership information
* There should be an ID for the subsidiary that we create. 
  * There can be a separate association table which links 10-K filers with a CIK to their identity as a subsidiary. This could be in a standalone association table or potentially integrated into the quarterly company information table.
  * We know that company name evolves over time, so this relationship between "ex21 subsidiary ID" and "CIK" may also need to be time dependent?  Since if the company name changes, the ex21 subsidiary ID will change (as it's based on name)

**out_sec10k__parents_and_subsidiaries**
Should this table actually be called out_sec10k__quarterly_exhibit_21_company_ownership to indicate that it flows from core_sec10k__quarterly_exhibit_21_company_ownership and not core_sec10k__quarterly_company_information? Or is the name fine and more clear as is?

This is a wide(r), legible version of the core table. 

* This table has company information for the subsidiary and parent merged onto the core_sec10k__quarterly_exhibit_21_company_ownership table.
* This table minimally has sub ID, sub name, parent ID (CIK), parent name, fraction owned. Or it could be fully denormalized with all the company information merged on.
* It could have an EIA utility ID for the parent and the subsidiary
* Much of the subsidiary information is null (when that sub doesn’t file a 10-K)
* This table gives a complete time series of ownership information.

#### Tasks

**core_sec10k__quarterly_exhibit_21_company_ownership**
- [x] The CIK of the parent company for each of the subsidiaries listed in this table is the CIK associated with the filename that the information is pulled from (associations between CIK and filename in the core_sec10k__quarterly_filings table)
- [x] Include parent company CIK (and maybe parent company name) in this table
- [x] Ensure this is a complete time series of ownership information
- [x] Create an ID for each subsidiary. Does this ID change if the subsidiary in turn files a 10-K and has a CIK that we eventually match it to?

**out_sec10k__parents_and_subsidiaries**
[...]

PR #4134, opened 2025-03-17
Renormalize SEC 10-K company info tables

zaneselvans:
# Overview

Refactors the normalization and cleaning of company information & name changelog tables to retain more information and produce better normalized / more intuitive outputs.

# Testing

- [x] Materialized all assets up and downstream of SEC 10-K `+key:*sec10k*+`
- [x] Ran full ETL locally including new assets.
- [x] Ran fast ETL locally including new assets
- [x] Ran `dbt test` against full local ETL outputs
- [x] Ran `make pytest-coverage` locally (failing w/ SQLite errors)
- [x] Ran a branch deployment

# Before Requesting Review
- [x] Migrate existing reshaping code from notebook into assets.
- [x] Migrate data cleaning functions into those assets from old assets and address my own previous PR comments.
- [x] Integrate company info data cleaning / standardization into `core_sec10k__quarterly_company_information` asset.
- [x] Update company info resource definition to reflect current columns / PK
- [x] Break out an intermediate `_core_sec10k__changelog_company_name` table that does just the data cleaning.
- [x] Move asset definitions from `pudl_models` module into `transform/sec10k.py` and `output/sec10k.py` modules.
- [x] Switch to doing Parquet output temporarily for tables with working schemas.
- [x] Add a `source_url` field to company info table that links directly to the SEC files into the output tables.
- [x] Update the master filings table to be consistent with the company info changes
- [x] Merge `report_date` and other filing fields associated with `filename_sec10k` into the company info output table.
- [x] Fix upstream `core_sec10k__parents_and_subsidiaries`
- [x] Move other asset definitions into extract, transform, and output modules.
- [x] Associate all the SEC 10-K assets with raw/core/output asset groups
- [x] Remove `pudl_models` and references to it.
- [x] Remove duplicate instances of reading the parent & subsidiaries data from GCS.
- [x] Create the skinny SEC / EIA association table in only one place and re-use it in the output table.
- [x] `out_sec10k__quarterly_company_information`
  - [x] Update alembic schema
  - [x] Materialize the output table
  - [x] Inspect output table for sanity 🤡 
  - [x] Get the schema check working
- [x] Debug schema check in `core_sec10k__quarterly_exhibit_21_company_ownership`
- [x] `core_sec10k__parents_and_subsidiaries`
  - [x] Fix schema / check failures
  - [x] Factor out reusable EIN/TIN cleaning function
  - [x] Factor out reusable SIC id/name cleaning function
- [x] Remove zombie PK from `core_sec10k__quarterly_company_information`
- [x] Remove zombie PK from `core_sec10k__changelog_company_name`
- [x] Start writing above tables to SQLite
- [x] Lower case company name and sec10k type in filings table.
- [x] Do a self-review of the PR.
- [x] Documentation
  - [x] Update / Expand company info table descriptions.
  - [x] Update / Expand company name changelog table descriptions.

# During / After Code Review

## Before Merging
- [x] Ask @katie-lamb why the assertion about unique CIK in the SEC/EIA association table is failing? Is this concerning?
- [x] Update dbt tests
  - [x] Remove deleted tables from dbt tests
  - [x] Add new tables to dbt tests
  - [x] Set new row-count expectations
- [x] Debug huge discrepancy in number of parent-subsidiary records per year for 2017-2023.
- [x] **temporarily** pragma no cover the SEC 10-K functions to avoid coverage failure
- [x] Run a full ETL locally and run the dbt tests against that output
- [x] Consolidate the 10+ alembic migrations into one.
- [x] Remove superfluous assert unique for crosswalk table -- now covered in dbt tests
- [x] Add an assert notna() to protect against filling in NA company names inappropriately. 
- [x] Debug SQLite error when loading SEC 10-K in integration tests locally ([see this discussion](https://github.com/sqlalchemy/sqlalchemy/discussions/12498))
- [x] Add ENUM constraints back in for SEC 10-K fields.
- [x] Run `make pytest-integration` locally and check that dbt (and other!) tests pass
- [x] Update release notes
- [x] @zschira Archive SEC 10-K outputs from DeltaLake in Zenodo #4065 See also #4159
- [x] Switch to using input data from [Zenodo archiver](https://github.com/catalyst-cooperative/pudl-archiver/pull/626)
- [x] Allow processing of only a subset of years.
- [x] Remove special cases that resulted from `USE_PUDL_MODELS` 
- [x] Turn test coverage back on for SEC 10-K modules
- [x] Disable unsatisfiable FK constraints (sadly, all of them)
- [x] Verify that `make pytest-coverage` passes locally.
- [x] Verify that [integration tests pass in CI on GitHub](https://github.com/catalyst-cooperative/pudl/actions/runs/14289155534/job/40048113952)
- [x] Verify that branch deployment succeeds

## Next Steps

- [ ] @zschira fix the way `block_count` is incremented upstream and generate new company info outputs.
- [ ] @zschira Re-run the upstream processing to capture Q1 data for 2018-2022 and use new info block counts.
- [ ] @katie-lamb to rebase her PR #4086 of of this branch and integrate changes.
- [ ] @zschira publish a new version of outputs to the DeltaLake bucket.
- [ ] @zaneselvans Create an SEC 10-K data source documentation page.

### Addressed:
- Ask @zschira about the meaning of the filer and block counts in the raw data. Why is it the case that only **most** records can be concatenated based on filename, filer count, report date, and company info block? Why are a modest proportion of addresses and name changes not lining up with anything? Why is it the case that things only line up when the filer and block counts are the same?
  - It turns out that the info block counter was not being reset in each new filer, so these two counters were getting out of sync, potentially creating misalignments between the different blocks.
- Is there a potential bug in the fillna() step of the company name changelog output table?
  - Yep! There were a handful of NA old company names. Set them to the empty string to avoid filling inappropriately

### Outstanding
- ~Why are the SEC 10-K ENUM constraints failing on NA values?~ (mysteriously started working)
- Why can't I get group-specific (like, all-SEC table) field definitions picked up in `PUDL_PACKAGE`?


zaneselvans:
# Company Info Exploration Notes

## Company data
- 4 of the 5 tables that can be pulled out of this data seem to pertain to companies -- static or slowly changing dimensions.
- `company_data` and `filing_values` have a 1:1 mapping and can be merged based on their index without losing any information.
- There are 386 instances in which `business_address` does not have an index value that corresponds to company + filing data.
- There are 22612 instances in which `mail_address` does not have an index value that corresponds to company + filing data.
- All orphaned addresses have different values for `filer_count` and `company_information_block_count`
- All `company_data` and `filing_values` records have matching `filer_count` and `company_information_block_count` values.
- We can build a nice 400K row table with modest numbers of nulls by left-merging the address tables onto the (company data + filing values) table 
- This means about 23K addresses are not included in the final table. However, it isn't clear what company or filing those addresses ought to pertain to.
- The only clear identifier(s) for the orphaned addresses are the CIK in the filename, but that will generally pertain to the main filer, which may have a different address.
- A quick visual inspection suggests that many of the orphaned addresses are additional instances or forms of addresses already included in the table.
- What is an appropriate natural primary key for this table?
- The `report_date` field is not independent -- it could be derived from the date associated with the filing's filename / file ID.
- The `filer_count` and `company_information_block_count` columns are always equal after this merge, so only one integer column can be kept.
- The filename + one of the counter columns would result in a unique PK. What would that mean conceptually? Master filing + Nth company filing identified in that master filing?
- There's a directly reported CIK in the company data, but also a "master filer" CIK that can be extracted from the filename.
- Having both the master filer and the company CIKs in the same table might be useful, and could indicate whether the company is a subsidiary.
- The filename is mostly for provenance -- where did this data come from.
- If the reported data fields (address, TIN, name, fiscal year, etc.) are truly identical, then we don't lose any information about the companies by deduplicating them.

## Filing data
- There's also filing-level data that's associatd with the company data.
- On cursory inspection, it seems like a lot of the duplicate values that show up when trying to create a natural primary key if we concatenate the filing data with the company data are actually due to the filing data.
- If we separate the filing data out we may get a granular filing history that's separate from the less variable company data?
- The same SEC filing number can show up many times because amendments and the filings they amend share the same filing number.

## Company naming history
- There's a kind of "sub-table" that's associated with the company data, which describes the history of the company names.
- The whole history seems to be reported in each filing, so there's lots of duplication.
- Alone, this table only has the "former" names and the dates of changes away from those former names.
- Being able to merge in the report date, filename, current company name and CIK from the company table will make this table more legible / useful.
- With all of that information there will be a ton of duplication, but also probably lots of useful ways to deduplicate it in a subsequent step.
- Maybe this is a raw or intermediate table that gets handed off for subsequent processing.

src/pudl/transform/sec10k.py
```diff
@@ -0,0 +1,471 @@
...
+    There is a 1-to-1 correspondence between the company_data and filing_values blocks,
+    and virtually all business_address blocks (99.9%), but a substantial number (~5%) of
+    mail_address blocks do not share index values with any company_data or filing_values
+    blocks, and are ultimately dropped. The former_company blocks contain a large amount
+    of duplicative data, but can be associated with a central_index_key value for later
+    association with an individual SEC 10-K filer.
+
+    Further processing that standardizes the contents of these tables is deferred to
+    separate downstream core assets.
+    """
```
zaneselvans, line 192: Not sure if some of this detail should go into the table description. It seemed a bit too far in the weeds.

src/pudl/transform/sec10k.py
```diff
@@ -0,0 +1,471 @@
...
+            how="left",
+            validate="one_to_one",
+        )
+        .merge(  # 22,612 mailing addresses orphaned here.
+            mail_address.sort_index(),
+            left_index=True,
+            right_index=True,
+            how="left",
+            validate="one_to_one",
+        )
```
zaneselvans, line 241: Curious if @zschira understands why there would be extra records in the two address blocks that don't line up with the company info & filing values blocks.  All the records that end up merging are the ones where the filer count and block count are the same, and the ones that don't merge are those where the count values are different.

src/pudl/transform/sec10k.py
```diff
@@ -0,0 +1,471 @@
...
+        core_sec10k__parents_and_subsidiaries.loc[
+            :, ["central_index_key", "utility_id_eia"]
+        ]
+        .drop_duplicates(subset="central_index_key")
+        .dropna()
+    )
+    # TODO: this assertion is failing, so I added the drop_duplicates() back in above,
+    # but should ask Katie about why and if it's expected / important...
+    # Verify that each CIK is matched to only one utility:
+    # assert sec_eia_matches["central_index_key"].nunique() == len(sec_eia_matches)
```
zaneselvans, line 470: @katie-lamb left a comment in the previous version of this asset that indicated each CIK should only appear once in this table, but that expectation doesn't seem to be met. Do we know why? Is it a problem?

src/pudl/output/sec10k.py
```diff
@@ -0,0 +1,82 @@
...
+    it, by shifting the old name column by one row within each central_index_key group,
+    and then fills in the last new company name value with the current company name.
+    """
+    name_changelog = core_sec10k__changelog_company_name.sort_values(
+        ["central_index_key", "name_change_date"]
+    ).assign(
+        company_name_new=lambda x: x.groupby("central_index_key")["company_name_old"]
+        .shift(-1)
+        .fillna(x["company_name"])
+    )
```
zaneselvans, line 77: Might there be an issue here in which there are NA values that *don't* result from the shift, and are getting filled in with the current company name incorrectly?

zaneselvans:
When I run `make pytest-integration` locally, it fails with a bunch of errors from SQLite about trying to write data to the database from the `core_sec10k__quarterly_company_information` table. However, I'm able to run both the full and fast ETL from within the Dagster UI without any issues. (it doesn't fail in CI on GitHub, which I suspect is because those tables aren't processed on GitHub).

src/pudl/metadata/fields.py
```diff
@@ -2645,7 +2649,11 @@
     },
     "location_of_incorporation": {
         "type": "string",
-        "description": "Cleaned location of incorporation of the company.",
+        "description": (
+            "Location of the company's incorporation. This can be a full US state "
+            "name, a state abbreviation, the name of a foreign country, etc. Not yet "
+            "standardized / cleaned."
+        ),
```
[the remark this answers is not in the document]
katie-lamb: Yeah, in the context of Ex. 21 filings this is just what the model was directly pulled/parsed from the Ex. 21 PDF (if a location is even included). There's no standardization in reporting requirements. In the fuzzy match between filers and subsidiaries I do some clean up on this location column that we could just apply to the core ownership table (or just the output ownership table). But not super necessary since I believe this is just a column used in Ex. 21 subs.

src/pudl/transform/sec10k.py
zaneselvans: The previous comment was a little misleading -- duplicates need to be dropped because we're reconstructing an association table from an un-normalized table, which has many records which refer to the same companies.  In addition to that, most of the records have NA values in one or both of the `central_index_key` and `utility_id_eia` columns, and those were also duplicates.

I've clarified the expectations and am asserting that both columns are independently unique. This could probably go into a `dbt` test instead.  They're already required to be jointly unique and non-null because they are the PK for the table.

src/pudl/transform/sec10k.py
zaneselvans: It turns out there is an issue upstream with the way that the `block_count` is assigned -- the counter doesn't currently reset when you enter another block, so the two counters get out of sync if there's a filer that doesn't have all of the blocks (e.g. no mailing address). @zschira is going to fix it and regenerate the outputs upstream.

src/pudl/transform/sec10k.py
zaneselvans: Hopefully the `block_count` fix upstream in the Mozilla repo will obviate most of this because all the blocks will just line up nice and clean.

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
zaneselvans: I think we should be thinking about these tables both as a way to show users what data is feeding into the matching processes (retroactively) and also as if they were the tables we're going to want to feed into the record linkage / matching in the future if/when it's pulled into the PUDL repo.

I think we probably want the messy original version of the data that appears here in the `raw` table, and whatever cleanup you've done applied to the `core` table -- in which case we can revert the description.  It just wasn't matching up with the data in the actual table.

src/pudl/etl/__init__.py
```diff
@@ -118,8 +118,16 @@
...
+# https://github.com/catalyst-cooperative/pudl/issues/4065
 if os.getenv("USE_PUDL_MODELS"):
-    default_assets += load_assets_from_modules([pudl.analysis.pudl_models])
+    default_assets += load_assets_from_modules(
+        [
+            pudl.extract.sec10k,
+            pudl.transform.sec10k,
+            pudl.output.sec10k,
+        ]
+    )
```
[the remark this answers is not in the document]
zaneselvans: Is there some way that this change in how the modules are selected could be resulting in the SQLite loading problem? I think probably not, because the tables are getting processed -- so Dagster has obviously picked them up.

But maybe somehow in CI / pytest they aren't being added to the database schema, so we're trying to write to a table that doesn't exist?  But it does exist in the normal Dagster UI setup?

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -1987,6 +1986,69 @@ core_pudl__codes_datasources,,11
...
+core_sec10k__parents_and_subsidiaries,2014,37862
+core_sec10k__parents_and_subsidiaries,2015,40311
+core_sec10k__parents_and_subsidiaries,2016,44119
+core_sec10k__parents_and_subsidiaries,2017,123606
+core_sec10k__parents_and_subsidiaries,2018,3605
+core_sec10k__parents_and_subsidiaries,2019,9594
+core_sec10k__parents_and_subsidiaries,2020,7737
+core_sec10k__parents_and_subsidiaries,2021,6687
+core_sec10k__parents_and_subsidiaries,2022,5573
+core_sec10k__parents_and_subsidiaries,2023,196880
```
zaneselvans, line 2020: @katie-lamb something seems fishy with the row counts here. Is there a reason why we'd expect it to vary this much from year to year?

dbt/models/sec10k/core_sec10k__assn_sec10k_filers_and_eia_utilities/schema.yml
```diff
@@ -0,0 +1,16 @@
...
+          - check_row_counts_per_partition:
+              table_name: core_sec10k__assn_sec10k_filers_and_eia_utilities
+              partition_column: null
+        columns:
+          - name: central_index_key
+            tests:
+              - unique
+          - name: utility_id_eia
+            tests:
+              - unique
```
zaneselvans, line 16: I added the per-column unique check here manually. They're already required to be non-null because they're the PK.

src/pudl/transform/sec10k.py
```diff
@@ -0,0 +1,494 @@
...
+    """
+    sec_eia_assn = (
+        core_sec10k__parents_and_subsidiaries.loc[
+            :, ["central_index_key", "utility_id_eia"]
+        ]
+        .dropna()
+        .drop_duplicates(subset=["central_index_key", "utility_id_eia"])
+    )
+    assert sec_eia_assn["central_index_key"].is_unique
+    assert sec_eia_assn["utility_id_eia"].is_unique
```
zaneselvans, line 493: Duplicative with the checks in `dbt`. Where should we put this kind of thing?

dbt/seeds/etl_full_row_counts.csv
katie-lamb: The coverage of Ex. 21 attachments significantly drops off after 2017. AI thinks this is due to a reporting requirement changed in 2017 (the SEC's FAST act) which reduced the reporting requirements for 10-k filers and made it so they only need to report subsidiaries of "material importance" rather than all subsidiaries (see [here](https://www.dodd-frank.com/2016/11/sec-issues-report-on-modernization-and-simplification-of-regulation-s-k-required-by-fast-act/?utm_source=chatgpt.com) and [here](https://www.sec.gov/newsroom/press-releases/2016-101) which don't explicitly name this, but get at it). This seems plausible. 

I looked at the raw filings, and the drop off in the number of Ex. 21 attachments is far less than the drop off in row counts (which represents the number of subsidiaries reported in total per year across all filers). This seems good, although I agree still fishy. We could check some of the Ex. 21 attachments post-2017 and see if the attachments are more often formats that the machine learning model can't parse. That seems quite plausible. I can poke around.

Screenshot 2025-04-01 at 10 55 40 PM

src/pudl/metadata/fields.py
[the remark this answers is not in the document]
katie-lamb: Okay, I can either apply this clean up in the ownership PR (which moves this name cleaning in from the mozilla repo) or I can move it in in this PR. The changes I'm thinking of apply cleaning to location of incorporation with a mapping from SEC state codes to full strings and cleaning to company names to expand legal terms and do some other standardization.

Does it make sense to apply company name cleaning to all core tables? That would mean that in the company information table the company names will look closer to what's used for record linkage than what's actually reported as the name to the SEC, e.g. "PGE Corp" becomes "pge corporation"

dbt/seeds/etl_full_row_counts.csv
zaneselvans: Just back of the envelope, the excess 2017 and 2023 records, if spread across all of the intervening years, would be a similar number of records per year as the previous period, so maybe the 2018-2022 records are getting associated with the wrong reporting periods somehow?

katie-lamb: Which table are you using to generate the last plot from the above comment? I'm confused why it doesn't align with the plot I made with the raw quarterly filings of the number of filings with Ex. 21 attachments across time.

katie-lamb: Ah okay mystery solved. I looked back at the [metadata in GCS for the Ex. 21 extraction runs](https://console.cloud.google.com/storage/browser/sec10k-outputs/v2/ex21_extraction_metadata?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&authuser=2&invt=Abts7A&project=catalyst-cooperative-mozilla) and it seems like there must have been an issue because the model never ran on Q1 data for 2018 - 2022 (Q1 has by far the most data). It seems like we did scrape and archive these Q1 filings because the basic 10k (company information w/o ownership) model extraction ran on those years. @zschira I remember you ran the model in batches of years, do you think somehow those Q1 filings were just omitted by accident? Can we run it on these quarters only?

zaneselvans: @katie-lamb Updated the code snippets above to include reading the data out of dagster.

Glad you found the source of the weirdness! It seems like we have a number of issues we need to address upstream before re-running that processing and we should make sure we've made the necessary changes before kicking it off.

Any idea what's up with the apparently ~linear increase in the number of Exhibit 21 records over time?

zschira: Ah yeah there were a couple quarters that were consistently timing out while trying to run, and I think since we were only using the latest quarter for record linkage we didn't up finishing those quarters, but it shouldn't be hard to launch the VM and re-run. It just might take some debugging to get the full quarter to complete. I'll write up an issue on what data needs to be re-created and link here.

katie-lamb: @zaneselvans I might be getting a little confused since I've been re-working the structure of the ownership tables and the old vs new schema is getting muddled in my head. But this is my understanding of those tables in this PR:
* the first plot that you made is showing the total number of parents + subsidiary companies per report year, since that core table as it is in this PR is a concatenation of filers (with an ID column for their parent company if they're also a subsidiary) and subsidiaries (who don't file). 
* the second plot is the number of parent to subsidiary relationships, not the number of Ex. 21 attachments to 10-k filings. I believe that `core_sec10k__quarterly_exhibit_21_company_ownership` in this PR is like a big concatenation of what's been extracted by the model from the PDFs of the Ex. 21's (subsidiary name, location, fraction owned).

Does it make sense that the number of parent-subsidiary relationships would scale linearly 🤔 ? It still seems weird, but maybe less weird than if the number of Ex. 21 attachments are scaling linearly. We could look at some of the highest subsidiary company and parent company name value counts and see if some of these are obviously being counted too many times or something? It might be easier to take a look at this in the next ownership PR, as those tables become easier to use and understand with the new schema

zaneselvans: @katie-lamb I haven't dug into the contents of either of these tables or really modified them -- they should have about the same structure and contents as whatever was getting handed off before, just with a new name.

In both of these plots all I'm doing is counting the number of records in the tables associated with each report date, which I would naively expect to be more or less consistent from year to year with a little bit of random fluctuation, but it seems like that's not true in at least 2 ways:

- the big 2018-2022 gap which you've figured out
- the roughly linear increase in Q1 ownership records over time.

If it makes more sense to look at the second weirdness in the context of the ownership table PR we can do that.

I'm still confused about only using the latest quarter for record linkage.

katie-lamb:
Round up of unresolved questions:

* [here](https://github.com/catalyst-cooperative/pudl/pull/4134#discussion_r2024128663) we talk about whether to move the location and company name cleaning into the core tables, so they align better with what's used for matching or fuzzy matching, or if they should be kept closer to the name registered/filed with the SEC, i.e. do we want LLC to always be expanded to Limited Liability Company in the core tables?
* [here](https://github.com/catalyst-cooperative/pudl/pull/4134#discussion_r2021997437) is a question about a duplicative check with `dbt` checks - where does this go
* [here](https://github.com/catalyst-cooperative/pudl/pull/4134#discussion_r2025665195) is a question about the linear relationship of the number of parent to subsidiary relationships over time, despite the number of Exhibit 21 filings staying relatively constant over time

dbt/seeds/etl_full_row_counts.csv
katie-lamb: The record linkage used the latest quarter because I wanted to be able to run/test it locally and we didn't have time/budget to figure out the infrastructure for switching backends in `splink`. In an ideal/future scenario we'd block on `report_date` and do a linkage on the whole dataset

src/pudl/transform/sec10k.py
```diff
@@ -0,0 +1,494 @@
...
+        _core_sec10k__changelog_company_name.assign(
+            name_change_date=lambda x: pd.to_datetime(x["date_of_name_change"])
+        )
+        .rename(
+            columns={
+                "former_conformed_name": "company_name_old",
+                "company_conformed_name": "company_name",
+            }
+        )
+        .drop_duplicates(
```
zschira, line 389: Out of curiosity, why do we end up with duplicates here?

zschira (approved):
A couple small comments, but nothing blocking

zaneselvans:
- For uniformity I added a last minute output table for the quarterly filings, which includes the source URL.
- I think the test coverage and dbt schemas / row-counts are all working.

## Stumped:
- **Blocking:** SQLite is still barfing on load in the integration tests, although it's fine in Dagster.
- **Non-Blocking** Don't understand why I can't add the ENUM constraints on columns with NA values.
- **Non-Blocking** Don't understand why I can't add SEC field namespace specific field definitions.

zaneselvans:
@katie-lamb I think having the company name standardization applied to the core tables would resolve some of the messiness that shows up in the name changelog -- right now the same companies have slightly different names in different changelogs reported in different filings, which results in some spurious duplication.

src/pudl/output/sec10k.py
```diff
@@ -0,0 +1,108 @@
...
+        + ".txt"
+    )
+    return company_info
+
+
+@dg.asset(
+    io_manager_key="pudl_io_manager",
+    group_name="out_sec10k",
+)
+def out_sec10k__changelog_company_name(
```
katie-lamb, line 81: i feel a little confused what's happening here - `company_name_old` and `company_name_new` indicate a name change on the `name_change_date`, and then `company_name` indicates the current name of the company (in the most recent filing)?

zaneselvans: `company_name` is ultimately dropped -- it's only here to provide the **current** company name associated with the CIK so we can fill in the most recent value for  `company_name_new` since only the former name is included in the changelog data.

katie-lamb: Ah got it, I was looking at a pickled version of the table which had `company_name` in it.

katie-lamb:
@zaneselvans I'm still tempted to keep the `core` table company name cleaning and location of incorporation cleaning in the next ownership PR (just because it's already there and in my mind that feels easier than porting it over to this PR). That would involve the following:

`company_name` cleaning for:
* `_core_sec10k__quarterly_company_information`
* `_core_sec10k__changelog_company_name`
* `core_sec10k__quarterly_exhibit_21_company_ownership`

`location_of_incorporation` (or state of incorporation) cleaning for:
* `_core_sec10k__quarterly_company_information`
* `core_sec10k__quarterly_exhibit_21_company_ownership`

All downstream out and core tables would also have the changes. What do you think about pushing it to the next PR? I guess it might make row counts change slightly?

katie-lamb (approved):
Left a comment about pushing the name and location cleaning to the next ownership PR, but it's non-blocking.

zaneselvans:
@katie-lamb Oh yeah I just meant eventually it would be nice to have the clean names here -- don't think we should do it right now.  But if you've already pulled it into the ownership PR then we can apply it more broadly.

zaneselvans:
# SQLite pytest failure on new tables

Both the fast and the full ETL running from within the Dagster UI seem to behave as expected, but when I attempt to run `make pytest-integration` it fails while trying to write to the new `core_sec10k__quarterly_company_information` table.

I cracked open the two `pudl.sqlite` databases (one in the temporary test directory and one in `$PUDL_OUTPUT` to look at the table schemas and see if anything was different and wow, something is different:

## Schema from Dagster outputs:

```sql
CREATE TABLE IF NOT EXISTS "core_sec10k__quarterly_company_information" (
	filename_sec10k TEXT NOT NULL,
	central_index_key TEXT,
	phone_number TEXT,
	company_name TEXT,
	film_number TEXT,
	fiscal_year_end TEXT,
	sec_act TEXT,
	filing_number_sec TEXT,
	industry_id_sic TEXT,
	industry_name_sic TEXT,
	taxpayer_id_irs TEXT,
	filer_count INTEGER NOT NULL,
	incorporation_state TEXT,
	sec10k_type TEXT,
	business_street_address TEXT,
	business_street_address_2 TEXT,
	business_city TEXT,
	business_state TEXT,
	business_zip_code TEXT,
	business_zip_code_4 TEXT,
	business_postal_code TEXT,
	mail_street_address TEXT,
	mail_street_address_2 TEXT,
	mail_city TEXT,
	mail_state TEXT,
	mail_zip_code TEXT,
	mail_zip_code_4 TEXT,
	mail_postal_code TEXT,
	CONSTRAINT pk_core_sec10k__quarterly_company_information PRIMARY KEY (filename_sec10k, filer_count)
);
```

## Schema from pytest outputs:
```sql
CREATE TABLE core_sec10k__quarterly_company_information (
	filename_sec10k TEXT NOT NULL CONSTRAINT "ck_core_sec10k__quarterly_company_information_`869797815423838214`" CHECK (TYPEOF("filename_sec10k") = 'text'),
	filer_count INTEGER NOT NULL CONSTRAINT "ck_core_sec10k__quarterly_company_information_`645929632982996875`" CHECK (TYPEOF("filer_count") = 'integer'),
	central_index_key TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-8131591961600470204`" CHECK ("central_index_key" IS NULL OR TYPEOF("central_index_key") = 'text'),
	company_name TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-6055054790339972579`" CHECK ("company_name" IS NULL OR TYPEOF("company_name") = 'text'),
	fiscal_year_end TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`6883342265760783397`" CHECK ("fiscal_year_end" REGEXP '^(?:(?NULL[1-9]|1[0-2])(?NULL[1-9]|1\d|2\d|3[01])|(?NULL[13-9]|1[0-2])(?NULL|30)|(?NULL[13578]|1[02])31)$') CONSTRAINT "ck_core_sec10k__quarterly_company_information_`5865496341343288976`" CHECK ("fiscal_year_end" IS NULL OR TYPEOF("fiscal_year_end") = 'text'),
	taxpayer_id_irs TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`543403320458096829`" CHECK ("taxpayer_id_irs" REGEXP '^\d{2}-\d{7}$') CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-4102673625268701631`" CHECK ("taxpayer_id_irs" IS NULL OR TYPEOF("taxpayer_id_irs") = 'text'),
	incorporation_state TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`4859554518323869125`" CHECK ("incorporation_state" IS NULL OR TYPEOF("incorporation_state") = 'text'),
	industry_name_sic TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`3959947840676262577`" CHECK ("industry_name_sic" IS NULL OR TYPEOF("industry_name_sic") = 'text'),
	industry_id_sic TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-1049020856102920516`" CHECK ("industry_id_sic" IS NULL OR TYPEOF("industry_id_sic") = 'text') CONSTRAINT "ck_core_sec10k__quarterly_company_information_`393160964873686241`" CHECK ("industry_id_sic" REGEXP '^\d{4}$'),
	film_number TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-2112028239114759727`" CHECK ("film_number" IS NULL OR TYPEOF("film_number") = 'text'),
	sec10k_type TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-2407677542113752064`" CHECK ("sec10k_type" IS NULL OR TYPEOF("sec10k_type") = 'text'),
	sec_act TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-7600126519029964726`" CHECK ("sec_act" IS NULL OR TYPEOF("sec_act") = 'text'),
	filing_number_sec TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-5371354177211245036`" CHECK ("filing_number_sec" IS NULL OR TYPEOF("filing_number_sec") = 'text'),
	phone_number TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`7246694621579180064`" CHECK ("phone_number" IS NULL OR TYPEOF("phone_number") = 'text'),
	business_street_address TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`5484274367470430808`" CHECK ("business_street_address" IS NULL OR TYPEOF("business_street_address") = 'text'),
	business_street_address_2 TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-4784933923516920717`" CHECK ("business_street_address_2" IS NULL OR TYPEOF("business_street_address_2") = 'text'),
	business_city TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`9188667654242910404`" CHECK ("business_city" IS NULL OR TYPEOF("business_city") = 'text'),
	business_state TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`5039280021878742990`" CHECK ("business_state" IS NULL OR TYPEOF("business_state") = 'text'),
	business_zip_code TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`5925569707734702120`" CHECK ("business_zip_code" IS NULL OR TYPEOF("business_zip_code") = 'text') CONSTRAINT "ck_core_sec10k__quarterly_company_information_`9167138883428811174`" CHECK ("business_zip_code" REGEXP '^\d{5}$'),
	business_zip_code_4 TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`5787026919065795578`" CHECK ("business_zip_code_4" IS NULL OR TYPEOF("business_zip_code_4") = 'text') CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-5663903969516670354`" CHECK ("business_zip_code_4" REGEXP '^\d{4}$'),
	business_postal_code TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-6043389851641362179`" CHECK ("business_postal_code" IS NULL OR TYPEOF("business_postal_code") = 'text'),
	mail_street_address TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-426568826900999257`" CHECK ("mail_street_address" IS NULL OR TYPEOF("mail_street_address") = 'text'),
	mail_street_address_2 TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-1474752677648676221`" CHECK ("mail_street_address_2" IS NULL OR TYPEOF("mail_street_address_2") = 'text'),
	mail_city TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-3464743352250393159`" CHECK ("mail_city" IS NULL OR TYPEOF("mail_city") = 'text'),
	mail_state TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-5711654700478405267`" CHECK ("mail_state" IS NULL OR TYPEOF("mail_state") = 'text'),
	mail_zip_code TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-7793121904295891811`" CHECK ("mail_zip_code" IS NULL OR TYPEOF("mail_zip_code") = 'text') CONSTRAINT "ck_core_sec10k__quarterly_company_information_`7446180905356083089`" CHECK ("mail_zip_code" REGEXP '^\d{5}$'),
	mail_zip_code_4 TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-1794365427123600825`" CHECK ("mail_zip_code_4" IS NULL OR TYPEOF("mail_zip_code_4") = 'text') CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-4660163769868745038`" CHECK ("mail_zip_code_4" REGEXP '^\d{4}$'),
	mail_postal_code TEXT CONSTRAINT "ck_core_sec10k__quarterly_company_information_`-2654227477147437306`" CHECK ("mail_postal_code" IS NULL OR TYPEOF("mail_postal_code") = 'text'),
	CONSTRAINT pk_core_sec10k__quarterly_company_information PRIMARY KEY (filename_sec10k, filer_count)
);
```

## Whut?

- The pytest `CREATE` statement is actually more like what I expect to see -- we programmatically define a bunch of SQL constraints based on our field definitions, and those look more or less like the ones we're defining.
- The lack of of `IS NULL or TYPEOF()` constraints in the Dagster outputs could also explain why my attempts to add ENUM constraints on columns that contain NA values was failing.
- I have no idea what is up with the back-tick quoted random numbers in the constraint names. It seems like those are probably supposed to ensure that all the constraints are named and unique, but the fact that they're in backticks and some of the numbers are negative seems suspicious.
- So, what's different about how these two systems create the database, and how can we make them both look more like what's happening in pytest currently?
- Given all of the Alembic migrations that have been in flux on the branch I'm working on, I'm pretty sure I just removed the PUDL DB and re-created it with `alembic upgrade head` before populating it using Dagster. So maybe Alembic isn't actually getting all of the constraints that we've specified?  Or maybe it is unwilling to work with these weird backtick negative numbers?
- How is the database created in pytest?
- If we start without any database, how is it created in Dagster (I tried this recently and it just fails because the DB isn't there -- it doesn't seem to know how to create it).
- How is the database created in the nightly builds, and which outcome do we get there / in the DB that we're distributing?
- None of this says anything about why the SQLite insertion is failing, but I have to imagine it's failing because there's some values in the data that don't conform to one of the constraints that's been specified.

zaneselvans:
Given that Alembic can't deal with the check constraints, as long as we're creating the database with Alembic, which we seem to be doing, then it seems like I can temporarily avoid this problem by just fixing whatever the constraint violation is.

However, the pandera dataframe asset checks aren't detecting anything, and when I create them in a notebook and try to validate the table that comes out of Dagster, it passes just fine.  I tried adding some bad data to a column and re-ran it and it failed. So I am feeling pretty stumped.

```python
company_info = get_table("core_sec10k__quarterly_company_information")
schema = PUDL_PACKAGE.get_resource("core_sec10k__quarterly_company_information").schema.to_pandera()
schema.validate(company_info)  # all is well
```

```python
broken = company_info.copy()
broken["taxpayer_id_irs"] = broken["taxpayer_id_irs"].str.replace("992", "XXX")
schema.validate(broken)  # fails as expected
```

And the pandera schems **seems** to have all the checks I expect to be in there
```python
for name, col in schema.columns.items():
    print(f"{name}: {col.checks}")
```
```
filename_sec10k: []
filer_count: []
central_index_key: []
company_name: []
fiscal_year_end: [<Check str_matches: str_matches('re.compile('^(?:(?:0[1-9]|1[0-2])(?:0[1-9]|1\\d|2\\d|3[01])|(?:0[13-9]|1[0-2])(?:29|30)|(?:0[13578]|1[02])31)$')')>]
taxpayer_id_irs: [<Check str_matches: str_matches('re.compile('^\\d{2}-\\d{7}$')')>]
incorporation_state: []
industry_name_sic: []
industry_id_sic: [<Check str_matches: str_matches('re.compile('^\\d{4}$')')>]
film_number: []
sec10k_type: []
sec_act: []
filing_number_sec: []
phone_number: []
business_street_address: []
business_street_address_2: []
business_city: []
business_state: []
business_zip_code: [<Check str_matches: str_matches('re.compile('^\\d{5}$')')>]
business_zip_code_4: [<Check str_matches: str_matches('re.compile('^\\d{4}$')')>]
business_postal_code: []
mail_street_address: []
mail_street_address_2: []
mail_city: []
mail_state: []
mail_zip_code: [<Check str_matches: str_matches('re.compile('^\\d{5}$')')>]
mail_zip_code_4: [<Check str_matches: str_matches('re.compile('^\\d{4}$')')>]
mail_postal_code: []
```

zaneselvans:
The culprit was the fiscal year-end REGEX check -- which works fine in Pandera, and fine in Python, and fine as a test in dbt, but seems to be causing problems with SQLite / SQLAlchemy, so I've removed it from our field constraints, added it as a dbt test, and opened [this discussion](https://github.com/sqlalchemy/sqlalchemy/discussions/12498) in the SQLAlchemy repo.

zaneselvans:
Pretty much everything here seems to be working except...
- `make pytest-integration` can't pass locally because of FK constraint failures when only limited years are processed.
- `make pytest-integration` can't pass in CI because `dbt` tests fail on SEC 10-K tables since they don't exist in CI.

zaneselvans:
Okay weirdly **all** of the FK constraints between the `utility_id_eia` columns that appear in the SEC 10-K tables and the utility entity and SCD tables (where there's a `report_date`) are failing. For the SCD tables, this makes sense, since the SEC 10-K data is quarterly, but the SCD table is annual, so 3/4 of the dates won't line up.

But even 100% of the `core_sec10k__assn_sec10k_filers_and_eia_utilities` FK constraints are failing, which is nothing but IDs, so there's got to be like a dtype problem or something more general.

PR #4135, opened 2025-03-18
feat: add metrics dashboard

jdangerx:
# Overview

Adds the infrastructure required for the marimo notebook server

## What did you change?

* move PUDL viewer stuff into a separate file for organization

* add a bunch of secrets
* add SA for deployment: WIF, artifact registry, cloud run usage
* add SA for cloud run service: cloud sql user, secret accessor
* connect cloud run service to cloud sql instance
* add read-only SQL user so that we can use that from the cloud run service (actual privileges handled within postgres shell)
* allow public to access cloud run service

# Testing

Ran `terraform apply` and the infrastructure exists.

If you run `terraform plan` now you should only get 2 changes - the annotations should change on the `pudl-usage-metrics-dashboard` service and the `pudl-viewer` service.


terraform/main.tf
```diff
@@ -563,249 +567,3 @@ resource "google_storage_bucket_iam_member" "nrel_finito_inputs_archiver_gcs_iam
   role   = each.key
   member = "serviceAccount:${google_service_account.nrel_finito_inputs_gha.email}"
 }
-
```
jdangerx, line 566: This all just got moved into a separate file.

zschira (approved):
Looks good! I'm glad we're moving things into separate files, definitely makes it much less unwieldy to read

jdangerx:
Sigh. docs build failed:
```
WARNING: failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://scikit-learn.org/stable/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 503 Server Error: Service Unavailable for url: https://scikit-learn.org/stable/objects.inv
```

Looks like GH pages is currently acting up. Will try again later.

PR #4145, opened 2025-03-21
Add devtools script to check for schema drift against dbt

### LINKED ISSUE #4125: Test for schema drift between PUDL and `dbt` metadata definitions.
# Overview

With our adoption `dbt` we now now two separate places where our column and table names are defined. We need to ensure that they don't get out of sync.

The first is the Resource and Field definitions under `pudl.metadata`. The second is all the `schema.yml` files under `dbt/models`  Initially the `dbt` model definitions are generally derived from the metadata using the `dbt_helper` script, but subsequent changes to the PUDL metadata will not propagate to the `schema.yml` files.  Mismatches **also** don't automatically result in failures.

To prevent the two schemas diverging slowly over time, we need to create a unit test that verifies that the two schemas describe identical sets of tables and columns.  This is a "unit" test in that it doesn't depend on any data being processed, and can run quickly in development & CI.

Until this check is in place, it's likely that our schemas will be divergent.

## Potential Issues

- Because we use `dbt` to create some intermediate models that contain derived values that we need for data testing, it might look like there are additional tables in `dbt` that don't correspond to anything in PUDL.
- The SEC 10-K tables only get defined when the `$USE_PUDL_MODELS` envvar is set (see #4065) and so if that variable isn't set, it might look like those tables are missing from PUDL, but they show up in `dbt`.
- Is it actually the case that every single table defined in the PUDL DB will also show up in the `dbt` `schema.yml` files? Or is it that *only* tables which show up in the PUDL DB schema should show up in the `dbt` schema, and **when** a table appears in both places, it **must** have the same set of columns in both?

## Implementation Possibilities

- It's easy to create a SQLAlchemy metadata object from the PUDL Metadata classes.  If it's easy to create an analogous object from the `dbt` `schema.yml` files, maybe we can just compare the table and column names from those two metadata objects.

# Scope

## In Scope

- Defining a test under `test/unit` that fails if the table names defined in the PUDL metadata are different from the PUDL sources that show up in `dbt`
- Defining a test under `test/unit` that fails if there's a mismatch between the sets of columns that show up in the tables defined in the PUDL metadata and the PUDL sources within `dbt`

## Out of Scope

- Changing how we store our database metadata -- there's a ton of functionality that hangs off of the Pydantic metadata classes that we use to define the database, tables, data sources, fields, etc. which is used all across the PUDL codebase and the `pudl-archiver` repo.
- Automating updates to the `dbt` `schema.yml` files so that they correspond to whatever the PUDL DB structure looks like.
- Checking dtypes or other column-level details match.

# Success Criteria
- ???

krivard:
# Overview

May close #4125.

This PR brings `dbt_helper` into the primary codebase and adds a unit test comparing the pudl and dbt schemas. The test fails under the following conditions:

- There is a table in PUDL that does not exist in dbt
- There is a field in PUDL that does not exist in dbt
- There is a table in dbt that does not exist in PUDL
- There is a field in dbt that does not exist in PUDL

Sample output on error:
```
E       AssertionError: 
E         dbt schema drift detected:
E         
E         dbt is missing the following tables:
E         	core_pudl__codes_data_maturities
E         
E         dbt is missing the following fields:
E         	_core_eia860__cooling_equipment
E         		pond_operating_date
E         		field_i_added_to_pudl_but_not_dbt
E         
E         pudl is missing the following tables:
E         	out_vcerare__hourly_available_capacity_factor
E         
E         pudl is missing the following fields:
E         	_core_eia860__cooling_equipment
E         		field_i_deleted_from_pudl_but_not_dbt
```

## Documentation

- [ ] ~~Update relevant Data Source jinja templates (see `docs/data_sources/templates`).~~
- [ ] ~~Update relevant table or source description metadata (see `src/metadata`).~~
- [ ] ~~Review and update any other aspects of the documentation that might be affected by this PR.~~

# Testing

Ran locally; no output.

Locally modified test to drop some tables/fields and add others, and got a pudlsworth of schema drift failure outputs as above.

## To-do list

From review

- [x] Explicitly check that all PUDL tables exist (instead of just throwing a FileNotFound)
- [x] Also check that all dbt sources exist in PUDL, so that a pre-commit check will find the problem. (don't use all models for this; that would also include the intermediate validation models that are not intended to exist in PUDL)
- [x] Move `dbt_helper` into PUDL at `pudl.scripts`, and add hookups in `pyproject.toml`
- [x] Move this script into unit tests so that it runs as part of existing CI and pre-commits

From template

- [x] If updating analyses or data processing functions: make sure to update or write data validation tests (e.g. `test_minmax_rows()`).
- [ ] ~~For minor ETL changes or data additions, once `make pytest-coverage` passes, make sure you have a fresh full PUDL DB downloaded locally, materialize new/changed assets and all their downstream assets and [run relevant data validation tests](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `pytest` and `--live-dbs`.~~
- [ ] ~~For bigger ETL or data changes run the full ETL locally and then [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `make pytest-validate`.~~
- [ ] ~~Alternatively, run the `build-deploy-pudl` GitHub Action manually.~~


zaneselvans:
- I think there would be an advantage in checking for extra tables & columns in `dbt` that don't exist in PUDL in this test because they could be caught in the `pre-commit` hooks rather than only showing up when someone attempts to merge into `main`, which would be more ergonomic.
- I suspect that we *do* expect there to be some tables that exist in `dbt` which don't exist in PUDL -- namely any of the models / seed tables that we generate for testing purposes, so if we ever have tests defined on those I guess we probably need to have a way of excluding them from this test?
- It seems like there's no explicit test for PUDL tables which are missing from `dbt` -- it's just that we would expect it to fail because we'd be trying to read from a path that doesn't exist?
- Creating a `set()` of tables from each source up front and identifying any extra / missing tables that we can add to an error message, and then iterating only over the tables that do appear in both sources to check the columns (maybe in the same way) would let check everything that can be checked, and report back at the end.
- I'm guessing that you pulled this out into `devtools` because you wanted to get access to the functions defined in the `dbt_helper.py` script, but I think we probably want to move in the other direction -- there's a messy proliferation of scripts outside of the codebase, and we can't import or use anything outside of it in the tests, or test code that lives outside the codebase with pytest. So instead we could move the `dbt_helper.py` under e.g. `src/pudl/scripts` and create an entrypoint for it in `pyproject.toml` so it's available as a command if PUDL is installed.
- I think it's preferable to put this somewhere it can run as part of our unit tests since then it will automatically get run as a `pre-commit` hook, and also automatically get run in CI on GitHub (if someone doesn't have `pre-commit` installed) without needing to set up another separate pre-commit command, and a separate job in our CI workflow to call it.
- The existing `dbt_test.py` module already reaches outside of the test directory to access the `dbt` files, so it should be workable to point the `_get_model_path` and `get_data_source` etc. functions at the right path.
- Side note: I realize now that my suggestion of using the SQLAlchemy metadata that the PUDL Package can create wouldn't work, since we have a number of tables that don't get written to the DB and aren't part of the SQLite schema.

krivard:
Hmm well that's certainly working as intended

dbt/models/ferc714/out_ferc714__hourly_planning_area_demand/schema.yml
```diff
@@ -14,4 +14,6 @@ sources:
           - name: report_date
           - name: datetime_utc
           - name: timezone
-          - name: demand_mwh
+          - name: demand_reported_mwh
+          - name: demand_imputed_pudl_mwh
+          - name: demand_imputed_pudl_mwh_imputation_code
```
krivard, line 19: These are fixing schema drift detected by the test in this very PR!

dbt/models/pudl/core_pudl__codes_imputation_reasons/schema.yml
```diff
@@ -0,0 +1,12 @@
...
+  - name: pudl
+    tables:
+      - name: core_pudl__codes_imputation_reasons
+        data_tests:
+          - check_row_counts_per_partition:
+              table_name: core_pudl__codes_imputation_reasons
+              partition_column: null
+        columns:
+          - name: code
+          - name: description
```
krivard, line 12: More schema drift fixes!

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -1983,6 +1983,7 @@ core_pudl__assn_ferc714_xbrl_pudl_respondents,,118
 core_pudl__assn_utilities_plants,,18790
 core_pudl__codes_data_maturities,,4
 core_pudl__codes_datasources,,11
+core_pudl__codes_imputation_reasons,,9
```
krivard, line 1986: schema drift fix

pyproject.toml
```diff
@@ -138,6 +138,7 @@ pudl_check_fks = "pudl.etl.check_foreign_keys:pudl_check_fks"
 pudl_datastore = "pudl.workspace.datastore:pudl_datastore"
 pudl_etl = "pudl.etl.cli:pudl_etl"
 pudl_service_territories = "pudl.analysis.service_territory:pudl_service_territories"
+dbt_helper = "pudl.scripts.dbt_helper:dbt_helper"
```
krivard, line 141: Include dbt_helper as a console script

src/pudl/scripts/dbt_helper.py
```diff
@@ -470,7 +471,9 @@ def migrate_tests(table_name: str, test_config_name: str, model_name: str | None
     _write_dbt_yaml_config(schema_path, schema)
 
 
-@click.group()
+@click.group(
+    context_settings={"help_option_names": ["-h", "--help"]},
+)
```
krivard, line 476: Since dbt_helper is a real console script now, it has to pass our console script tests, which means accepting both "-h" and "--help" which it didn't before

test/unit/dbt_helper_test.py
```diff
@@ -0,0 +1,188 @@
...
+    actual = _generate_quantile_bounds_test(test_config)
+    assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "row_condition,expected",
+    [
+        ("'0000-00-00'", "CAST('0000-00-00' AS DATE)"),
+        ("x == 0", "x = 0"),
+        ("x != 0", "x <> 0"),
```
krivard, line 92: These each fit on one line so didn't use a namedtuple for them

test/unit/dbt_helper_test.py
```diff
@@ -0,0 +1,188 @@
...
+
+@dataclass
+class DbtSchemaMocks:
+    field: unittest.mock.Mock
+    resource: unittest.mock.Mock
+    pudl_package: unittest.mock.Mock
+    table_name: str
+    partition_column: str
+    schema: DbtSchema
+    yaml: str
```
krivard, line 108: This one is easier to understand with type annotations so used a dataclass

krivard (commented):
self-review

zaneselvans (approved):
This functionality all looks good, and it found some new drift after I merged in `main` locally.

It's not blocking but one thing I would change is the way it reports the drift. The current messages make it sound like the schemas from PUDL and dbt are both equally valid -- in one case, PUDL is missing some fields, and in the other case dbt is missing some (other) fields -- but for now the PUDL schema is the authoritative source, and in both cases it's the dbt schema that needs to be updated to conform to what's found in the PUDL metadata. So I think it would be clearer to tell the user "dbt has these extra unexpeted tables/fields" and `dbt is missing these expected tables/fields which exist in PUDL"

test/integration/dbt_test.py
```diff
@@ -87,3 +89,17 @@ def test_dbt(
...
+
+@pytest.mark.script_launch_mode("inprocess")
+def test_dbt_helper(script_runner):
+    ret = script_runner.run(
+        [
+            "dbt_helper",
+            "add-tables",
+            "--etl-fast",
+            "all",
+        ],
```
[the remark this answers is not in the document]
krivard: I tried it with `--use-local-tables` but it tried to find a parquet file in a random temp directory and failed with FileNotFound, which is not the failure case we're looking for here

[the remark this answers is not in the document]
zaneselvans: And that fixture is `pudl_io_manager` which is defined in `test/conf.py`

PR #4152, opened 2025-03-27
Metadata materials

krivard:
# Overview

Not intended for merging, just discussion

### Sample metadata migrations

* Script for helping put existing metadata description into template format
* Unfilled template with 19 samples
* Filled template with 19 samples

[Samples are annotated in this google sheet](https://docs.google.com/spreadsheets/d/1d9CrtIwVVJbLiotSmn3djQXeSOdEN_V8t4DNvyWYUck/edit?gid=0#gid=0) and were drawn as follows:
- Draw 1 from each group by (layer, asset-type extracted from table name)
- Draw 3 from each group by (layer)
- Draw 1 from each group by (layer, quartile of len(description))

### DIFF
--- a/scratch/sample_metadata.py
+++ b/scratch/sample_metadata.py
@@ -1,0 +1,857 @@
+"""Sample migrated metadata."""
+
+metadata = {
+    "_core_eia860__fgd_equipment": {
+        "original_description": """
+Information about flue gas desulfurization equipment at generation facilities, from EIA-860 Schedule 6E. Note: This table has been cleaned, but not harvested with other EIA 923 or 860 data. The same variables present in this table may show up in other _core tables in other years. Once this table has been harvested, it will be removed from the PUDL database.
+""",
+        "description": """
+# [if timeseries] Time series of {short_description}.
+# [if assn] Links {source1} to {source2}.
+# [if codes] Explains codes for {purpose}.
+# [if entity/scd] Entity/SCD table for {short_description}.
+[else] Catalogs available flue gas desulfurization equipment at generation facilities.
+
+Derived from EIA-860 Schedule 6E.
+
+Usage Warnings
+^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


scratch/sample_metadata.py
```diff
@@ -0,0 +1,857 @@
...
+        "description": """
+[if timeseries] Time series of {short_description}.
+[if assn] Links {source1} to {source2}.
+[if codes] Explains codes for {purpose}.
+[if entity/scd] Entity/SCD table for FERC 714 respondents, annotated with the county FIPS IDs for their service territories.
+[else] {Concise verb/noun indicating type} {short_description}.
+
+This table has no primary key. Each row represents a respondent's information for a particular report date.
+
+Derived from ferc714.
```
aesharpe, line 833: This is a good example of a table that I'm like what the heck is this for. Why do we need this table that's randomly annotated with the county FIPS ID? Did we make it for something specific?

aesharpe (commented):
Only looked at the `samepl_metadata.py` file.

Some thoughts: 

- I mention this a few times throughout, but we should **clarify the difference between the "Usage Warnings" and "Additional Details"** section. If it's too hard to distinguish them, maybe we should combine them.
- Would be good to add as many **implications** as we can throughout the descriptions. E.g.: a usage warning and what that might mean about how you should or shouldn't use the data. Same goes for the "Additional Details" section.
- I wonder if we could create a **dictionary of standard usage warnings** somewhere that we could reference. Would look really clean to have all similar warnings use the same wording.
- Sometimes the **line between an `assn` and `entity/scd`** table feels burry. What is our definition there?
- Do we want to include **links to data source pages** or is that overkill?
- There's some information that sometimes gets included in the "data is derived from" section that feels like it might be a usage warning? It's unclear where we should include information about how a table was created/imputed etc. especially if it's derived from more than one source.

krivard:
Superceded!

PR #4154, opened 2025-03-27
feat: dbt tests in dagster asset checks

### LINKED ISSUE #4132: Run DBT tests for assets that are related in Dagster
# Overview

Up until now we've been getting by with running `dbt build --select "source:your_cool_asset_here"`.

It would be nice if we could say something like `dbt build --select "+your_cool_asset"` to get all of the assets that are upstream of `your_cool_asset` in Dagster, sort of like using `dagster materialize --selection`.

This *seems* like it would need to be an extension to `dbt_helper` - it would translate the dagster selection into a DBT selection, then run the right DBT commands.

# Success criteria

* [ ] we can run all the DBT tests for a dagster asset & its upstream/downstream dependencies with a single invocation of `dbt_helper validate --select +key:"your_dagster_asset"+`

# Next steps

* See if we can clearly translate from Dagster selections to DBT selections.

jdangerx:
# Overview

Closes #4132 .

For all the checks we define in DBT, run them as asset checks in Dagster.

## What did you change?

* Added a factory that creates Dagster asset checks based on the `manifest.json`. `dbt` doesn't like running concurrently so we use `filelock` to manage that.
* Added a configuration to speed up multi-process execution.
* Some small copy/style changes from linting.

## Documentation

# Testing

* ran `etl-fast` and only ran into the same rowcounts issues as I saw on `main` - still need to debug why that's the case but seems unrelated to these changes.
* ran `build-deploy-pudl` on this branch

## To-do list

- [ ] rip out dbt validation tests
- [ ] fix CI issues with `manifest.json` not existing


src/pudl/etl/__init__.py
```diff
@@ -20,8 +21,17 @@
     load_assets_from_modules,
 )
 from dagster._core.definitions.cacheable_assets import CacheableAssetsDefinition
+from dagster_dbt import DbtCliResource
 
 import pudl
+from pudl.etl import (
```
jdangerx, line 26: Changed these from relative imports since I was getting some complaints from the Dagster multi-process executor.

src/pudl/helpers.py
```diff
@@ -2045,11 +2050,12 @@ def get_dagster_execution_config(
             particular tags. This is helpful for applying concurrency limits to
             highly concurrent and memory intensive portions of the ETL like CEMS.
 
-            Dagster description: If a value is set, the limit is applied to only
-            that key-value pair. If no value is set, the limit is applied across
-            all values of that key. If the value is set to a dict with
-            `applyLimitPerUniqueValue: true`, the limit will apply to the number
-            of unique values for that key. Note that these limits are per run, not global.
+            Dagster description: If a value is set, the limit is applied to
```
jdangerx, line 2053: Something about the line-wrapping changed here is all.

src/pudl/helpers.py
```diff
@@ -2062,12 +2068,18 @@ def get_dagster_execution_config(
                 },
             },
         }
+
+    start_method_config = {}
+    if "forkserver" in multiprocessing.get_all_start_methods():
+        start_method_config = {"forkserver": {"preload_modules": ["pudl"]}}
```
jdangerx, line 2074: This saves like 9s per process which matters when we have hundreds of checks that only take a couple seconds to actually run.

src/pudl/settings.py
```diff
@@ -536,6 +536,8 @@ def check_eia_dependencies(cls, data: dict[str, Any]) -> dict[str, Any]:
 class DatasetsSettings(FrozenBaseModel):
     """An immutable pydantic model to validate PUDL Dataset settings."""
 
+    etl_type: Literal["fast", "full", "adhoc"] = "adhoc"
```
jdangerx, line 539: @zaneselvans we talked about adding some way of identifying what 'type' of job we're running. here it is.

zaneselvans:
- I ran the full ETL locally including all of the new dbt asset checks, and it completed as expected, taking 1hr49min, which is about the same time it took before without the dbt asset checks.
- I kicked off the `pytest` CI check to see how it affects the tests / make sure all the asset checks run there.
- If it's all working then we should go ahead and remove `test/integration/dbt_test.py::test_dbt()` which is currently running all of the `dbt` tests separately at the end of the integration tests.

zaneselvans:
- There's [an error](https://app.readthedocs.org/projects/catalystcoop-pudl/builds/27912484/) resulting from trying to access the `dbt` directory relative to the installed python package directory on RTD. I think this doesn't work because for the RTD build the PUDL package is actually installed alongside all of the other python packages, and not just in editable mode from the cloned repository.
- It looks like there may be a real asset check failure related to the `core_eia860__assn_boiler_generator` table that's preventing the fast ETL from completing in CI?
- It took the integration tests more than 2.5 hours to get to that failure, so I'm wondering if the asset checks are somehow adding a lot of additional processing time, at least with the single-threaded in-process Dagster executor that's used in the CI.

zaneselvans:
One (admittedly drastic) way around the RTD build failure might be to build the docs and deploy them using GitHub Pages. This would let us have all the docs under a nice uniform index like...

```
https://docs.catalyst.coop  # index of all projects with docs
https://docs.catalyst.coop/pudl # PUDL documentation
https://docs.catalyst.coop/pudl-archiver  # PUDL Archiver documentation
https://docs.catalyst.coop/handbook # Catalyst Cooperative member handbook
https://docs.catalyst.coop/ferc-xbrl-extractor # etc. etc.
```

jdangerx:
closed in favor of #4261

PR #4162, opened 2025-04-01
EIA 930 BA level imputation

### LINKED ISSUE #4138: Impute and test BA level EIA-930 demand data
# Overview

- The core output of this project is a set of realistic, complete BA-level electricity demand time series.

# Success Criteria

- We have new imputed EIA-930 demand values being written out to Parquet in `out_eia930__hourly_operations` 
- We know how aggregations of BA subregion demand outputs (produced in #4124) compare to the directly reported BA demand (which is available over a longer historical period and covers all BAs, not just the ones with subregional reporting).
- We know how our imputed EIA-930 BA demand time series compare to the analogous previously published outputs that we extracted in #4136

# Next steps
* [ ] ...

zschira:
# Overview

This PR applies our timeseries imputation to EIA 930 BA level demand data.

Closes #4138

## Documentation

# Testing

- There are plots added to the notebook, `notebooks/work-in-progress/timeseries-comparison.ipynb` to demonstrate how well the timeseries imputation is working.
- Ran the full ETL locally
- Ran `dbt build` against the local ETL outputs.
- Ran a manual `workflow_dispatch` of pytest and it passed.
- Ran `pudl_check_fks` locally and it passed.


src/pudl/output/eia930.py
```diff
@@ -8,16 +8,42 @@
...
-    "ERCO": "America/Chicago",
-    "ISNE": "America/New_York",
-    "MISO": "America/Chicago",
-    "NYIS": "America/New_York",
-    "PJM": "America/New_York",
-    "PNM": "America/Denver",
-    "SWPP": "America/Chicago",
-}
+
+def _add_timezone(
```
zschira, line 12: Switched to getting timezones from the `core_eia__codes_balancing_authorities` table.

src/pudl/output/eia930.py
```diff
@@ -26,30 +52,44 @@
...
+
+imputed_subregion_demand_assets = impute_timeseries_asset_factory(
+    input_asset_name="_out_eia930__hourly_subregion_demand",
+    output_asset_name="out_eia930__hourly_subregion_demand",
+    years_from_context=_years_from_context,
     value_col="demand_reported_mwh",
     imputed_value_col="demand_imputed_pudl_mwh",
     id_col="combined_subregion_ba_id",
-    settings=ImputeTimeseriesSettings(method_overrides={2019: "tnn"}),
+    settings=ImputeTimeseriesSettings(method_overrides={2019: "tnn", 2025: "tnn"}),
```
zschira, line 84: Now that I implemented the padding before imputation, we can include 2025 data, but 2025 was failing in the same way as 2019.

src/pudl/output/eia930.py
```diff
@@ -8,16 +8,42 @@
...
+        core_eia930__hourly_operations,
+        core_eia__codes_balancing_authorities,
+    )
+    core_eia930__hourly_operations = core_eia930__hourly_operations.rename(
+        columns={
+            "demand_imputed_mwh": "demand_imputed_eia_mwh",
+            "interchange_imputed_mwh": "interchange_imputed_eia_mwh",
+            "net_generation_imputed_mwh": "net_generation_imputed_eia_mwh",
+        }
+    )
```
zaneselvans, line 42: Does this duplicate the renaming that's being done below in the definition of `core_eia930__hourly_operations_assets`?

src/pudl/output/eia930.py
```diff
@@ -8,16 +8,42 @@
...
+    # TODO: BA code WAUE does not have listed timezone, so dropping these records for now
+    return core_eia930__hourly_operations[
+        core_eia930__hourly_operations.timezone.notnull()
+    ]
```
zaneselvans, line 46: This seems like a blunt-force instrument. If we expect WAUE to be the only entity with an NA time zone right now, it would be safer to assert that all the WAUE records have NA timezone, drop **only** the WAUE records, and then assert that there are no more NA timezones.  That way if the data no longer matches are expectations in the future it breaks and we know it.

src/pudl/transform/eia930.py
```diff
@@ -76,6 +76,17 @@ def core_eia930__hourly_operations_assets(
...
+        columns={"net_generation_imputed_mwh": "net_generation_imputed_eia_mwh"}
+    )
+    operations = operations.rename(
+        columns={
+            "net_generation_imputed_mwh": "net_generation_imputed_eia_mwh",
+            "demand_imputed_mwh": "demand_imputed_eia_mwh",
+            "interchange_imputed_mwh": "interchange_imputed_eia_mwh",
+        }
+    )
+
```
zaneselvans, line 89: Seems like this rename is also done in another downstream asset. Maybe it can be removed there and retained here?

zaneselvans (changes requested):
- Something seems funny with the 2025 row count expectations
- Need to add PR & Issue references for this PR and the last one in the release notes.
- Maybe a duplicative rename that can be removed?
- Could be more conservative in wiping out records w/ NA timezones.

Processing the data myself locally now.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -161,9 +163,19 @@ def utc_dataframe_to_aligned(
...
 def pivot_aligned_timeseries_dataframe(
     aligned_df: DataFrame[AlignedTimeseriesDataFrame],
+    periods: int,
 ) -> DataFrame[TimeseriesMatrix]:
-    """Pivot aligned timeseries dataframe into timeseries matrix."""
-    return aligned_df.pivot(index="datetime", columns="id_col", values="value_col")
+    """Pivot aligned timeseries dataframe into timeseries matrix and pad if needed."""
+    matrix = aligned_df.pivot(index="datetime", columns="id_col", values="value_col")
+
+    # Pad matrix with any missing hours from timeseries
```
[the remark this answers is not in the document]
zaneselvans: Added.

docs/release_notes.rst
```diff
@@ -18,6 +18,8 @@ EIA 930
 ~~~~~~~
 * Added the table :ref:`out_eia930__hourly_subregion_demand`, which
   contains an imputed version of subregion demand.
+* Added the table :ref:`out_eia930__hourly_operations`, which
+  contains an imputed version of BA level demand.
```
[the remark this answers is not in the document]
zaneselvans: Added.

src/pudl/output/eia930.py
zaneselvans: Looks like these were duplicates. I kept the upstream one in the transform module and removed this one.

src/pudl/transform/eia930.py
zaneselvans: Kept this one, removed duplicative rename downstream.

src/pudl/output/eia930.py
zaneselvans: I went ahead and implemented these more conservative measures.

zaneselvans (approved):
Only lingering concern is the difference in the  ratios between the row counts for 2024 vs. 2025 in the fast vs. full ETL checks, which seems suspicious.

zaneselvans:
@zschira I poked at the row counts a little bit while updating them for `dbt` and the reason the ratios are so different is the FULL ETL includes a little bit of 2025 data directly (there's a 2025H1 partition).  And the reason that the 2024 fast row counts are so small compared to the full row counts is that there's only a half-year of data that's been included (one partition).  So I think this is good to go.

zaneselvans:
I'll run the full ETL locally and then run the dbt tests on those outputs while running the CI via a `workflow_dispatch` here to make sure the CI is happy.

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -2493,6 +2493,17 @@ out_eia923__yearly_generation_fuel_combined,2020,14846
...
+out_eia930__hourly_operations,2016,579744
+out_eia930__hourly_operations,2017,578160
+out_eia930__hourly_operations,2018,575286
+out_eia930__hourly_operations,2019,569400
+out_eia930__hourly_operations,2020,562350
+out_eia930__hourly_operations,2021,278446
+out_eia930__hourly_operations,2022,548483
+out_eia930__hourly_operations,2023,541664
+out_eia930__hourly_operations,2024,535824
+out_eia930__hourly_operations,2025,58828
```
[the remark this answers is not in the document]
zaneselvans: There's a little bit of actual 2025 data now -- just a month or two, and that's why it's showing up here more than in the fast ETL.

The fast ETL 2024 row count is only half of the year, which is why it's so much smaller than 2024 in the full ETL.

PR #4163, opened 2025-04-02
Get table name based metadata for table description

cmgosnell:
# Overview

Closes #XXXX.

## Documentation

# Testing

## To-do list

### DIFF
--- a/src/pudl/metadata/classes.py
+++ b/src/pudl/metadata/classes.py
@@ -1868,6 +1868,106 @@ class Resource(PudlMeta):
 # ---- Package ---- #
 
 
+class MetaFromResourceName(PudlMeta):
+    """Class to extract standard metadata to add to Resource.description."""
+
+    name: SnakeCase
+    """Resource name (aka table name)."""
+
+    layer_map: dict = {
+        "_core": (
+            "Processing Stage: Intermediate cleaned but not tidied/normalized. "
+            "These tables are generally published in this semi-processed stage temporarily."
+        ),
+        "core": "Processing Stage: Cleaned, tidied/normalized.",
+        "_out": "Processing Stage: Intermediate output table.",
+        "out": (
+            "Processing Stage: Wide/denormalized (helpful values for easy of use have been "
+            "merged back into these tables). Possible calculations or imputations."
+        ),
+    }
+    layer_string: str = "|".join(layer_map.keys())
+
+    # TODO: add link to https://catalystcoop-pudl.readthedocs.io/en/latest/data_sources/{datasource_name}.html
+    # if we have a data_sources page for it.
+    datasource_map: dict = {
+        datasource_name: f"Data Source: {SOURCES[datasource_name]['title']}"
+        for datasource_name in SOURCES
+    } | {
+        "eia": "Data Source: EIA -- Mix of multiple EIA Forms",
+        "epa": "Data Source: EPA -- Mix of multiple EPA sources.",
+    }
+
+    datasource_strings: str = "|".join(datasource_map.keys())
+
+    time_detail_map: dict = {
+        "yearly": "Time Detail: Annual time series.",
+        "monthly": "Time Detail: Monthly time series.",
+        "hourly": "Time Detail: Hourly time series.",
+    }
+    time_string: str = "|".join(time_detail_map.keys())
+
+    misc_map: dict = {
+        "assn": "Association assets provide connections between entities.",
+        "codes": (
+            "Code tables contain more verbose descriptions of categorical "
+            "codes typically manually compiled from source data dictionaries."
+        ),
+        "entity": ("Entity tables contain static information about entities."),
+        "scd": (
+            "Slowly changing dimension tables describe attributes of entities "
+            "that rarely change."
+        ),
+    }
+    misc_string: str = "|".join(misc_map.keys())
+    table_name_pattern: str = rf"^({layer_string})_({datasource_strings})__({time_string}|)(?:_|)({misc_string}|)(?:_|)(?:_|)(.*)$"
+
+    @classmethod
+    def match(cls):
+        """Return the regex match for the table name."""
+        return re.match(cls.table_name_pattern, cls.name)
+
+    @model_validator(mode="after")
+    def table_name_check(self: Self):
+        """Check the expected pattern of the table name."""
+        if not self.match:
+            raise ValueError(
+                f"Table name not formatted as expected. Table name found: {self.name}.\nExpected table name pattern: {self.table_name_pattern}"
+            )
+
+    def description_layer(self) -> str:
+        """Return a layer description from the resource name."""
+        layer: str = self.match.group(1)
+        layer_description = self.layer_map[layer]
+        return layer_description
+
+    def description_datasource(self) -> str:
+        """Return a description of the datasource from the table name."""
+        datasource: str = self.match.group(2)
+        title = DataSource.from_id(datasource).title
+        description = f"Data Source: {title}"
+        return description
+
+    def description_misc(self) -> str:
+        """Return a description of the miscellaneous from the table name."""
+        misc_opt: str | None = self.match.group(4)
+        description = self.misc_map.get(misc_opt, pd.NA)
+        return description
+
+    def description_time(self) -> str:
+        """Return a description of the time-dimension from table name."""
+        # TODO: ?maybe? we could add the date column into the description.
+        # we'd need to look into the table's columns via its metadata
+        time_opt: str | None = self.match.group(3)
+        description = self.time_detail_map.get(time_opt, pd.NA)
+        return description
+
+    def slug(self) -> str:
+        """Table name slug."""
+        slug: str = self.match.group(5)
+        return slug
+
+
 class Package(PudlMeta):
     """Tabular data package.
 
--- a/src/pudl/metadata/classes.py
+++ b/src/pudl/metadata/classes.py
@@ -1875,14 +1875,21 @@ class MetaFromResourceName(PudlMeta):
     """Resource name (aka table name)."""
 
     layer_map: dict = {
+        "raw": (
+            "Data has been extracted from original format, columns have been renamed for "
+            "consistency and multiple reporting periods have been concatenated but no "
+            "transformations or cleaning have been applied."
+        ),
         "_core": (
-            "Processing Stage: Intermediate cleaned but not tidied/normalized. "
+            "Intermediate cleaned but not tidied/normalized. "
             "These tables are generally published in this semi-processed stage temporarily."
         ),
-        "core": "Processing Stage: Cleaned, tidied/normalized.",
-        "_out": "Processing Stage: Intermediate output table.",
+        "core": (
+            "This layer contains tables that typically break denormalized raw tables into well-modeled tables that serve as building blocks for downstream wide tables and analyses."
+        ),
+        "_out": "Intermediate output table.",
         "out": (
-            "Processing Stage: Wide/denormalized (helpful values for easy of use have been "
+            "Wide/denormalized (helpful values for easy of use have been "
             "merged back into these tables). Possible calculations or imputations."
         ),
     }
@@ -1891,11 +1898,14 @@ class MetaFromResourceName(PudlMeta):
     # TODO: add link to https://catalystcoop-pudl.readthedocs.io/en/latest/data_sources/{datasource_name}.html
     # if we have a data_sources page for it.
     datasource_map: dict = {
-        datasource_name: f"Data Source: {SOURCES[datasource_name]['title']}"
+        datasource_name: SOURCES[datasource_name]["title"]
         for datasource_name in SOURCES
     } | {
-        "eia": "Data Source: EIA -- Mix of multiple EIA Forms",
-        "epa": "Data Source: EPA -- Mix of multiple EPA sources.",
+        "eia": "EIA -- Mix of multiple EIA Forms",
+        "epa": "EPA -- Mix of multiple EPA sources.",
+        "ferc": "FERC -- Mix of multiple FERC Forms.",
+        "ferc1_xbrl": "FERC 1 XBRL -- Post-2021 years of Annual Report of Major Electric Utilities.",
+        "ferc1_dbf": "FERC 1 DBF -- Pre-2021 years of Annual Report of Major Electric Utilities.",
     }
 
     datasource_strings: str = "|".join(datasource_map.keys())
@@ -1908,24 +1918,26 @@ class MetaFromResourceName(PudlMeta):
     time_string: str = "|".join(time_detail_map.keys())
 
     misc_map: dict = {
-        "assn": "Association assets provide connections between entities.",
+        "assn": "Table Type: Association table that provide connections between entities.",
         "codes": (
-            "Code tables contain more verbose descriptions of categorical "
+            "Table Type: Code table which contain more verbose descriptions of categorical "
             "codes typically manually compiled from source data dictionaries."
         ),
-        "entity": ("Entity tables contain static information about entities."),
+        "entity": (
+            "Table Type: Entity tables which contain static information about entities."
+        ),
         "scd": (
-            "Slowly changing dimension tables describe attributes of entities "
+            "Table Type: Slowly changing dimension (SCD) table which describe attributes of entities "
             "that rarely change."
         ),
     }
     misc_string: str = "|".join(misc_map.keys())
     table_name_pattern: str = rf"^({layer_string})_({datasource_strings})__({time_string}|)(?:_|)({misc_string}|)(?:_|)(?:_|)(.*)$"
 
-    @classmethod
-    def match(cls):
+    @property
+    def match(self):
         """Return the regex match for the table name."""
-        return re.match(cls.table_name_pattern, cls.name)
+        return re.match(self.table_name_pattern, self.name)
 
     @model_validator(mode="after")
     def table_name_check(self: Self):
@@ -1934,24 +1946,24 @@ class MetaFromResourceName(PudlMeta):
             raise ValueError(
                 f"Table name not formatted as expected. Table name found: {self.name}.\nExpected table name pattern: {self.table_name_pattern}"
             )
+        return self
 
     def description_layer(self) -> str:
         """Return a layer description from the resource name."""
         layer: str = self.match.group(1)
-        layer_description = self.layer_map[layer]
+        layer_description = f"Processing Stage: {self.layer_map[layer]}"
         return layer_description
 
     def description_datasource(self) -> str:
         """Return a description of the datasource from the table name."""
         datasource: str = self.match.group(2)
-        title = DataSource.from_id(datasource).title
-        description = f"Data Source: {title}"
+        description = f"Data Source: {self.datasource_map[datasource]}"
         return description
 
     def description_misc(self) -> str:
         """Return a description of the miscellaneous from the table name."""
         misc_opt: str | None = self.match.group(4)
-        description = self.misc_map.get(misc_opt, pd.NA)
+        description = self.misc_map.get(misc_opt, None)
         return description
 
     def description_time(self) -> str:
@@ -1959,11 +1971,14 @@ class MetaFromResourceName(PudlMeta):
         # TODO: ?maybe? we could add the date column into the description.
         # we'd need to look into the table's columns via its metadata
         time_opt: str | None = self.match.group(3)
-        description = self.time_detail_map.get(time_opt, pd.NA)
+        description = self.time_detail_map.get(time_opt, None)
         return description
 
     def slug(self) -> str:
-        """Table name slug."""
+        """Table name slug.
+
+        Possible use case: extract what is being associated from an assn table.
+        """
         slug: str = self.match.group(5)
         return slug
 


krivard:
Hmm it's not finding the table description rst template. But only during the readthedocs build; the other approaches seem okay

zaneselvans:
@krivard I'm not sure if this is the source of your problem, but one thing I think that's different about RTD is it installs the package for real, not using `--editable` so if you have code in the PUDL package that's referring to files that aren't deployed as part of the package (i.e. anything outside of `src/pudl/package_data` if it's not code) then the installed package won't have access to them -- the package knows nothing about the rest of the files in the checked out repository, even though the sphinx build is obviously access the RST files and other files that are outside of the set of files which are distributed by the package.

This issue also came up when @jdangerx was messing around with integrating `dbt` into the asset checks -- when the PUDL package was installed and imported, it was failing because that (now abandoned) PR included code that referred directly to files in the checked-out repo, which the installed package knows nothing about.

These issues make me think about fully committing to "PUDL as an application not a library" and no longer supporting a use-case in which the package is installed without everything else in the repo accessible to it, which would mean no longer doing PyPI and conda distributions of `catalystcoop.pudl`  I think it might also mean that we need to switch off of RTD for the docs build, which would let us deploy the docs here on GitHub using GitHub Pages, and host all of our docs under a nice clean domain (which is already set up but not really being used): https://docs.catalyst.coop/ -- each repo would have it's own subdirectory based on its name, e.g.:
* https://docs.catalyst.coop/pudl
* https://docs.catalyst.coop/pudl-archiver
* https://docs.catalyst.coop/ferc-xbrl-extractor
* https://docs.catalyst.coop/pudl-examples

krivard:
@zaneselvans yep! that's what's going wrong here.

in theory [9dbda65](https://github.com/catalyst-cooperative/pudl/pull/4163/commits/9dbda6500073e81cb811937bdfe9748db093eb20) told PUDL to build the resources using the RTD docs path and not the `__file__.parent.parent.[...]` docs path, but clearly didn't quite get there 🫤 

looks like it's actually building the resources twice, once at import (`from pudl.metadata import PUDL_PACKAGE`) and once at `data_dictionary_metadata_to_rst`, and it's the first one (where I can't tell it where to find the templates) that's failing. some more poking...

src/pudl/metadata/classes.py
```diff
@@ -1876,6 +1964,209 @@ def encode(self, df: pd.DataFrame) -> pd.DataFrame:
...
+        ),
+        "_core": (
+            "Data has been cleaned but not tidied/normalized. Published only "
+            "temporarily and may be removed without notice."
+        ),
+        "core": (
+            "Data has been cleaned and organized into well-modeled tables that serve as "
+            "building blocks for downstream wide tables and analyses."
+        ),
+        "_out": "Intermediate output table.",
```
zaneselvans: Note that all the `_out` tables have been removed now -- they only existed to serve ☠️ PudlTabl ☠️

krivard:
VICTORY

src/pudl/metadata/warnings.py
```diff
@@ -0,0 +1,23 @@
...
+    "no_leap_year": "Date column disregards leap years to comply with Actual/365 (Fixed) standard.",
+    "irregular_years": "Some years use a slightly different data definition.",
+    "known_discrepancies": "Contains known calculation discrepancies.",
+    "free_text": "Contains columns which may appear categorical, but are actually free text.",
+    "early_release": "May contain early release data.",
+    "aggregation_hazard": "Some columns are subtotals; use caution when choosing columns to aggregate.",
+    "scale_hazard": "Extremely large table; do not attempt to open with Excel.",  # TODO: set a threshold
+    "outliers": "Outliers present.",
+    "missing_years": "Some years are missing from the data record.",
+    "ferc_is_hard": "FERC data is notoriously difficult to extract cleanly; see TODO for details.",
```
krivard, line 22: this definitely needs to be fixed before merge

src/pudl/scripts/generate_metadata_template.py
```diff
@@ -0,0 +1,40 @@
+"""Tiny CLI for building default table description templates."""
```
krivard: oh huh i forgot this was here. this was before the preview tool turned out so nice. keep it as a lightweight tool or ditch it?

src/pudl/metadata/classes.py
```diff
@@ -958,11 +961,7 @@ class DataSource(PudlMeta):
 
     def get_resource_ids(self) -> list[str]:
         """Compile list of resource IDs associated with this data source."""
-        # Temporary check to use eia861.RESOURCE_METADATA directly
-        # eia861 is not currently included in the general RESOURCE_METADATA dict
```
krivard, line 962: (yes it is)

src/pudl/metadata/classes.py
```diff
@@ -1269,6 +1282,18 @@ class Resource(PudlMeta):
...
+        Literal[
+            "quarterly",
+            "yearly",
+            "monthly",
+            "hourly",
+        ]
+        | None
+    ) = None
+    layer: Literal["raw", "_core", "core", "_out", "out", "test"] | None = None
+    usage_warnings: list[String] = []  # could be list[Literal]
```
krivard:
```suggestion
    usage_warnings: list[String] = []
```

src/pudl/metadata/classes.py
```diff
@@ -1876,6 +2016,209 @@ def encode(self, df: pd.DataFrame) -> pd.DataFrame:
...
+        ),
+        "test": (
+            "Only used in unit and integration testing; not intended for public "
+            "consumption."
+        ),
+    }
+    layer_string: str = "|".join(layer_map.keys())
+
+    # TODO: add link to https://catalystcoop-pudl.readthedocs.io/en/latest/data_sources/{datasource_name}.html
+    # if we have a data_sources page for it.
```
krivard: this should be a post-PR followup

src/pudl/metadata/classes.py
```diff
@@ -1876,6 +2016,209 @@ def encode(self, df: pd.DataFrame) -> pd.DataFrame:
...
+        return self.tabletype_map.get(self.tabletype, "")
+
+    def description_tabletype_prompt(self) -> str:
+        """Return a description prompt for the table type."""
+        return self.tabletype_prompt_map.get(self.tabletype, "")
+
+    def description_time(self) -> str:
+        """Return a description of the time-dimension from table name."""
+        # TODO: ?maybe? we could add the date column into the description.
+        # we'd need to look into the table's columns via its metadata
```
krivard: post-PR followup

src/pudl/metadata/warnings.py
```diff
@@ -0,0 +1,23 @@
+"""Standard usage warnings to reference in dynamic table descriptions."""
```
krivard, line 1: Most TODOs in this file can be fixed in post-PR followups

test/unit/metadata_test.py
```diff
@@ -194,16 +195,67 @@ def test_frictionless_data_package_non_empty():
...
+            (properties.tabletype or properties.meta.get("table_type")) != "timeseries"
+            or (properties.time or properties.meta.get("timeseries_resolution"))
+        ),
+    }
+    for override, has_value in name_parse.items():
+        assert has_value, (
+            f"""Table {resource_name} could not be parsed as layer_source__tabletype_slug and no overrides were set in the table metadata. Rename {resource_name} or set the following override keys: {override}"""
+        )
+    # todo: layer-based checks
+    # todo: asset_type-based checks
```
krivard, line 275: these can be added in post-PR followups

src/pudl/metadata/classes.py
```diff
@@ -1133,7 +1132,21 @@ class CodeDataFrame(pr.DataFrameModel):
...
     # resources would have one...
     title: str | None = None
-    description: str
+    # this is a legacy field until we get all the metadata migrated
+    description: str | None = None
+    # extended metadata
+    # these are the new sections; all are optional and are presented alongside basic info
+    description_summary: str | None = None
+    description_layer: str | None = None
+    description_datasource: str | None = None
```
e-belfer: Do we want this (and some of the other fields like summary) to remain optional, or is the idea that once this all gets migrated we'll enforce these fields as mandatory?

src/pudl/metadata/classes.py
```diff
@@ -1269,6 +1282,18 @@ class Resource(PudlMeta):
...
+    timeseries_resolution: (
+        Literal[
+            "quarterly",
+            "yearly",
+            "monthly",
+            "hourly",
+        ]
+        | None
+    ) = None
+    layer: Literal["raw", "_core", "core", "_out", "out", "test"] | None = None
```
e-belfer: Same question as above - should we allow this to be `None`? Is this a temporary phase to enable migration?

src/pudl/metadata/classes.py
```diff
@@ -1358,6 +1383,117 @@ def _check_harvest_primary_key(cls, value, info: ValidationInfo):
...
+            resource_id: the resource name
+            obj: a PudlResourceDescriptor model dump of the resource
+
+        Returns: a list of all automatically-detected and manually-provided warnings for the requested resource.
+        """
+        usage_warnings = []
+        if "usage_warnings" in obj and obj["usage_warnings"]:
+            usage_warnings = list(obj["usage_warnings"])
+        if len(obj["sources"]) > 1:
+            usage_warnings.append("multiple_inputs")
```
e-belfer: Out of scope but for future purposes: Would this be a good place to amend the warning to specify what these actual datasets are? This would probably make this more informative.

src/pudl/metadata/classes.py
```diff
@@ -1876,6 +2016,209 @@ def encode(self, df: pd.DataFrame) -> pd.DataFrame:
...
+    # TODO: add link to https://catalystcoop-pudl.readthedocs.io/en/latest/data_sources/{datasource_name}.html
+    # if we have a data_sources page for it.
+    datasource_map: dict = {
+        datasource_name: SourceLabels(
+            SOURCES[datasource_name]["label"], SOURCES[datasource_name]["title"]
+        )
+        for datasource_name in SOURCES
+    } | {
+        "eia": SourceLabels("EIA", "EIA -- Mix of multiple EIA Forms"),
+        "epa": SourceLabels("EPA", "EPA -- Mix of multiple EPA sources."),
```
e-belfer: Is this accurate? I _think_ EPA data we publish is almost always coming from CAMD, if not always?

src/pudl/metadata/classes.py
```diff
@@ -1876,6 +2016,209 @@ def encode(self, df: pd.DataFrame) -> pd.DataFrame:
...
+        "ferc": SourceLabels("FERC", "FERC -- Mix of multiple FERC Forms."),
```
e-belfer: Are there actual cases of this in our data?

src/pudl/metadata/classes.py
```diff
@@ -1876,6 +2016,209 @@ def encode(self, df: pd.DataFrame) -> pd.DataFrame:
...
+class MetaFromResourceName(PudlMeta):
+    """Class to extract standard metadata to add to Resource.description."""
+
+    name: SnakeCase
+    """Resource name (aka table name)."""
+
+    seed: dict | None = None
+    """Seed metadata; the Resource dict, if known."""
+
+    layer_map: dict = {
```
e-belfer: In the same way that you do for the first two, it'd be helpful to have docstrings for each of these dictionaries explaining what they contain.

src/pudl/metadata/classes.py
```diff
@@ -1876,6 +2016,209 @@ def encode(self, df: pd.DataFrame) -> pd.DataFrame:
...
+
+    time_detail_map: dict = {
+        "quarterly": "Quarterly",
+        "yearly": "Annual",
+        "monthly": "Monthly",
+        "hourly": "Hourly",
+    }
+    time_string: str = "|".join(time_detail_map.keys())
+
+    tabletype_map: dict = {
```
e-belfer: As noted above, would be super helpful to have docstrings here as it's not at all obvious what is going on.

src/pudl/metadata/classes.py
```diff
@@ -1876,6 +2016,209 @@ def encode(self, df: pd.DataFrame) -> pd.DataFrame:
...
+        "assn": ("Association table", "providing connections between"),
+        "codes": ("Code table containing descriptions of categorical codes", "for"),
+        "entity": ("Entity table containing static information", "about"),
+        "scd": ("Slowly changing dimension (SCD) table", "describing attributes of"),
+        "timeseries": ("time series", "of"),
```
e-belfer: Any reason this one isn't capitalized like the other ones?

src/pudl/metadata/templates/resource_description.rst.jinja
```diff
@@ -0,0 +1,28 @@
+{% set tabletype_parts = resource.description_tabletype() %}
+{{ resource.description_time() }} {{ tabletype_parts.0 }}
+{% if resource.meta.description_summary %}{{ tabletype_parts.1 }} {{ resource.meta.description_summary }}{% endif %}
+
+Processing:
+  {{ resource.description_layer() }} {{ resource.meta.description_layer|default("", true) }}
```
e-belfer: For my own understanding: This is saying - take what we've automatically extracted and add what's been manually specified, if anything.

src/pudl/metadata/classes.py
```diff
@@ -1358,6 +1383,117 @@ def _check_harvest_primary_key(cls, value, info: ValidationInfo):
...
+        if obj["layer"] is None:
+            obj["layer"] = meta_from_name.layer
+        obj["usage_warnings"] = Resource._usage_warnings(resource_id, obj)
+        meta_from_name.meta["usage_warnings"] = [
+            uw["text"] if isinstance(uw, dict) else USAGE_WARNINGS[uw]
+            for uw in obj["usage_warnings"]
+        ]
+
+        # Render description
+        # First, move any description text from unmigrated tables into Additional Details
```
e-belfer: This is temporary code, we should change this after migration is done.

src/pudl/metadata/classes.py
```diff
@@ -1358,6 +1383,117 @@ def _check_harvest_primary_key(cls, value, info: ValidationInfo):
...
+            .get_template("resource_description.rst.jinja")
+            .render(
+                resource=meta_from_name,
+                warnings=USAGE_WARNINGS,
+            )
+        ).strip()
+
+        if cleanup:
+            # drop the structured metadata keys; all their text has been sorted into
+            # the description field
```
e-belfer: Question - why aren't the rest of the fields getting deleted? They seem to be getting tossed out anyways when we're writing the datapackage.

src/pudl/scripts/generate_metadata_template.py
e-belfer: Check - does this still work?

docs/templates/resource.rst.jinja
```diff
@@ -2,18 +2,10 @@
 {{ resource.name }}
 -------------------------------------------------------------------------------
 
-{{ resource.description | wordwrap(78) if resource.description else 'No table description available.' }}
+{{ resource.description|safe if resource.description else 'No table description available.' }}
```
e-belfer, line 5: |safe means - handle escapable html characters (e.g., smart quotes, line breaks).

src/pudl/metadata/classes.py
```diff
@@ -1358,6 +1383,117 @@ def _check_harvest_primary_key(cls, value, info: ValidationInfo):
...
+            obj: a PudlResourceDescriptor model dump of the resource
+            cleanup: [default True] whether to remove the description_* section
+                keys. This may only be set to False if you do not plan to create a
+                Resource object from ``obj``, e.g., if you are the metadata
+                preview wizard and only calling this function to build the
+                description RST.
+        """
+        meta_from_name = MetaFromResourceName(name=resource_id, seed=obj)
+        if obj["table_type"] is None:
+            obj["table_type"] = meta_from_name.tabletype or None
```
e-belfer: The additional `or None` here - can we just make meta_from_name.tabletype be None in this case instead?

krivard:
Fixing tests now

src/pudl/metadata/classes.py
```diff
@@ -1130,10 +1128,92 @@ class CodeDataFrame(pr.DataFrameModel):
...
+        description_summary: str | None = None
+        """A brief (~one-line) description of the contents of this resource.
+        If None or otherwise left unset, will be left blank.
+
+        If filled, should support whichever of the following scenarios is most appropriate for this resource:
+
+        * the table_type is set or can be automatically detected: this value should complete the sentence corresponding to pudl.metadata.descriptions.table_type_fragments[table_type]
+        * the table_type is None/unset _and_ is not named according to a standard table type listed in pudl.metadata.descriptions.table_type_fragments: this value should be a complete sentence summarizing the contents of this resource at a similar level of detail.
+        """
+        description_layer: str | None = None
```
e-belfer: Non-blocking: Can we rename to `description_layer_addl` and `description_source_addl`, or something similar? These docstrings are incredibly helpful, but I think it'd be useful to shout-out in the key name when something is additive rather than an override.

src/pudl/metadata/classes.py
```diff
@@ -1130,10 +1128,92 @@ class CodeDataFrame(pr.DataFrameModel):
...
+        """Unusual details about this resource's level of processing that don't fall into the normal definition of raw/core/_core/out/_out/etc.
+        If None or otherwise left unset, will be left blank.
+        This should only be set in truly obscure situations. If set, should be a complete sentence."""
+        description_source: str | None = None
+        """A brief refinement on the source data for this table, such as indicating the Schedule or other section number.
+        If None or otherwise left unset, will be left blank.
+        If set, should make sense when displayed directly after the title of a datasource (see pudl.metadata.descriptions.source_descriptions); parentheticals work best here."""
+        description_primary_key: str | None = None
+        """For resources with no primary key, a brief summary of what each row contains, and perhaps why a primary key doesn't make sense for this table.
```
e-belfer, line 1205: Can this accidentally override the PK if there is one set for the table? Would be helpful to clarify this behavior here, and guard against overrides if we don't want to allow them.

src/pudl/metadata/classes.py
```diff
@@ -1130,10 +1128,92 @@ class CodeDataFrame(pr.DataFrameModel):
...
+        If None or otherwise left unset, will be left blank.
+        This should only be set if the resource has no natural primary key. If set, should be a complete sentence or two."""
+        description_details: str | None = None
+        """All other information about this resource's construction and intended use, including guidelines and recommendations for best results.
+        If None or otherwise left unset, will be left blank; hides the Additional Details section entirely.
+
+        Q3 2025 Migration Mode variance: if PudlResourceDescriptor.description is a string, it gets moved here so you can see the old description content in the Additional Details section of the preview.
```
e-belfer: Nit: Would just add a TODO: Remove once migration complete to this note so we don't forget to!

src/pudl/metadata/classes.py
```diff
@@ -1130,10 +1128,92 @@ class CodeDataFrame(pr.DataFrameModel):
...
+
+        May also include more-detailed explanations of listed usage warnings."""
+
     # TODO (daz) 2024-02-09: with a name like "title" you might imagine all
     # resources would have one...
     title: str | None = None
-    description: str
+    # the str type is legacy support and should be removed once we get all the metadata migrated
```
e-belfer:
```suggestion
    # TODO: the str type is legacy support and should be removed once we get all the metadata migrated
```

src/pudl/metadata/classes.py
```diff
@@ -1433,6 +1515,12 @@ def dict_from_resource_descriptor(  # noqa: C901
         if "foreign_key_rules" in schema:
             del schema["foreign_key_rules"]
 
+        # overwrite description components with rendered text block
```
e-belfer, line 1525: Is this only necessary during the migration phase (and thus should we mark it for removal)?

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+    "entity": TableTypeFragments("Entity table containing static information", "about"),
+    "scd": TableTypeFragments(
+        "Slowly changing dimension (SCD) table", "describing attributes of"
+    ),
+    "timeseries": TableTypeFragments("time series", "of"),
+}
+"""Standard descriptive text to appear in the Summary (first line) of resource descriptions.
+
+These are split into fragments to permit some resources to do without any additional illustration beyond the basic table type.
+Such resources should not provide description.description_summary, and the system will apply only the first fragment here.
```
e-belfer: Non-blocking: An example might be helpful here to clarify

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+    "monthly": "Monthly",
+    "hourly": "Hourly",
+}
+"""Additional standard descriptive text to appear in the Summary (first line) of resource descriptions, for timeseries resources."""
+
+
+# [jul 2025 kmm] This is a mirror of pudl.metadata.PudlMeta, which can't be included here because it creates a circular import.
+# Collecting all the description mechanisms in one file seemed more important than a few lines of duplication, but if
+# PudlMeta changes, we may wish to apply the same change here.
+class DescriptionMeta(BaseModel):
```
e-belfer, line 96: This is a super generic model configuration, so I think it's fine to duplicate here.

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+      one of the values from the dictionaries above, or a more complex string composed of multiple
+      pieces of standardized text and manually-provided annotations.
+    """
+
+    type: str
+    description: str
+
+
+# [jul 2025 kmm] it feels like this should already exist somewhere but all I found were terrible Python Golf one-liners
+def first_non_none(*args):
```
e-belfer, line 124: Non-blocking: Accordingly, we might want to move this into helpers.py and import it from there? Seems broadly generically useful.

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+        # this is a little gross but calling 5 methods with the same signature individually makes me sad
+        for trait in ["summary", "layer", "source", "primary_key", "details"]:
+            getattr(self, f"_create_{trait}_description")(settings, defaults)
+        self._assemble_usage_warnings(settings, defaults)
+
+    def _create_summary_description(self, settings, defaults):
+        """Compute the summary component (first line) of the resource description.
+
+        The summary is standardized based on table type, and if the table type is timeseries, the timeseries resolution.
+        Any illustration text, if present, is included after the standard text for the timeseries resolution and table type.
```
e-belfer: What do you mean by illustration text? Would be helpful to refer to the exact relevant key.

src/pudl/metadata/descriptions.py
e-belfer: Would be helpful to emphasize in this method that the order of the args imparts a preference, since that's a key part of what you're using it to do.

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+        If the timeseries resolution and table type aren't set manually and can't be auto detected, fall back to
+        the description_summary key and use it as a complete sentence instead of just a predicate fragment.
+        """
+        timeseries_resolution = first_non_none(
+            settings.get("timeseries_resolution"), defaults.timeseries_resolution
+        )
+        table_type = first_non_none(settings.get("table_type"), defaults.table_type)
+        summary_annotation = settings.get("description_summary")
```
e-belfer: As much as possible, keeping the var names the same will reduce mental overhead and help us keep track of what's going on here. See my note above about renaming this to be something more like `description_summary_addl`.

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+        # permit a none tabletype for weirdos like core_ferc714__respondent_id
+        type_fragments = table_type_fragments.get(table_type, NONE_TABLETYPE_FRAGMENTS)
+
+        components = [
```
e-belfer, line 213: What are these components for? An inline comment and a clearer var name might help here.

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+
+        if summary_annotation is not None:
+            components.extend(
+                [
+                    type_fragments.conjunction,
+                    summary_annotation,
+                ]
+            )
+
+        self.summary = ResourceTrait(
```
e-belfer, line 228: Not particularly obvious to me what is going on here, more docs would be helpful.

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+            type=f"{table_type}[{timeseries_resolution}]",
+            description=" ".join(c for c in components if c is not None),
+        )
+
+    def _create_x_description(self, attr, lookup, settings, defaults):
```
e-belfer: Do these want type hints?

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+            components.extend(
+                [
+                    type_fragments.conjunction,
+                    summary_annotation,
+                ]
+            )
+
+        self.summary = ResourceTrait(
+            type=f"{table_type}[{timeseries_resolution}]",
+            description=" ".join(c for c in components if c is not None),
```
e-belfer, line 231: Do we expect more than one manually specified component per summary?

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+        """
+        has_primary_key = (
+            "primary_key" in settings.get("schema", {})
+            and len(settings["schema"]["primary_key"]) > 0
+        )
+        components = [
+            ", ".join(settings["schema"]["primary_key"])
+            if has_primary_key
+            else "This table has no primary key.",
+            settings.get("description_primary_key"),
```
e-belfer: In this case, naming this `description_primary_key_addl` would clarify that it is additive (I believe?) rather than overwriting the auto-generated field.

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,376 @@
...
+        """
+        # TODO: remove this handling after migration is complete
+        # temporary description->description_details passthrough for legacy description text
+        if isinstance(settings.get("description"), str):
+            settings["description_details"] = settings["description"]
+        # end TODO: remove
+
+        self.details = ResourceTrait(
+            type=str(settings.get("description_details") is not None),
+            description=first_non_none(settings.get("description_details"), ""),
```
e-belfer: Non-blocking nit: In this case I think if if not None else "" would be more legible, but that's just my take.

src/pudl/metadata/resources/epacems.py
```diff
@@ -4,7 +4,9 @@
 
 RESOURCE_METADATA: dict[str, dict[str, Any]] = {
     "core_epacems__hourly_emissions": {
-        "description": "Hourly emissions and plant operational data reported via Continuous Emissions Monitoring Systems as required by 40 CFR Part 75.",
+        "description": {
+            "description_summary": "emissions and plant operational data as required by 40 CFR Part 75.",
```
e-belfer: Is this intentional? I would imagine we'd want to just handle this while doing our migration?

src/pudl/metadata/classes.py
```diff
@@ -943,6 +944,7 @@ class DataSource(PudlMeta):
 
     name: SnakeCase
     title: String | None = None
+    label: String | None = None
```
e-belfer: Can you leave some commentary here about the distinction between these two fields (title and label)? I would typically think of label as something more like "eia176" than "EIA 176".

src/pudl/metadata/sources.py
```diff
@@ -701,6 +721,7 @@
     },
     "ferceqr": {
         "title": "FERC Form 920 -- Electric Quarterly Report (EQR)",
+        "label": "FERC Form 920",
```
e-belfer: Do we want to mention EQR here at all? It's the name it's more commonly known by. Noted above that I'm unclear on what the intended use of 'label' is

src/pudl/metadata/warnings.py
```diff
@@ -0,0 +1,23 @@
...
+    "multiple_inputs": "Contains information from multiple raw inputs.",
+    "derived_values": "Contains columns derived from inputs and not originally present in sources.",
+    "imputed_values": "Contains rows where missing values were imputed.",
+    "estimated_values": "Contains estimated values.",  # TODO: what do we mean here
+    "incomplete_id_coverage": "Not all IDs are present.",  # TODO: do we want to set a coverage threshold and only apply this when we don't meet it?
+    "incomplete_value_coverage": "?",  # TODO: do we mean high rates of missingness? do we want to set a threshold?
+    "redacted_values": "Some values have been redacted.",  # eg 88888
+    "mixed_aggregations": "Some entries contain aggregates that do not match the table type.",  # eg 99999
+    "month_as_date": "Date column arbitrarily uses the first of the month.",
+    "no_leap_year": "Date column disregards leap years to comply with Actual/365 (Fixed) standard.",
```
e-belfer, line 13: The casing in this one is particularly weird.

src/pudl/metadata/warnings.py
```diff
@@ -0,0 +1,23 @@
...
+    "irregular_years": "Some years use a slightly different data definition.",
+    "known_discrepancies": "Contains known calculation discrepancies.",
+    "free_text": "Contains columns which may appear categorical, but are actually free text.",
+    "early_release": "May contain early release data.",
+    "aggregation_hazard": "Some columns are subtotals; use caution when choosing columns to aggregate.",
+    "scale_hazard": "Extremely large table; do not attempt to open with Excel.",  # TODO: set a threshold
```
e-belfer, line 19: If we're mostly giving CSVs to people with a built-in subset/download limit, when would we try to use this?

src/pudl/metadata/warnings.py
e-belfer: Is there a follow-up issue for usage warning tasks?

test/integration/console_scripts_test.py
```diff
@@ -100,3 +100,23 @@ def test_pudl_service_territories(
     gdf = gpd.read_parquet(out_path)
     assert set(gdf.columns) == expected_cols
     assert not gdf.empty
+
+
+@pytest.mark.parametrize(
+    "resource_id",
+    [
+        # One migrated table
+        "core_epacems__hourly_emissions",
```
e-belfer, line 109: Is this table actually migrated? It seems like you created a dictionary but haven't set up the components yet.

test/unit/metadata_test.py
```diff
@@ -194,16 +195,76 @@ def test_frictionless_data_package_non_empty():
...
         assert expected_resource["schema"]["fields"] == [
             f.name for f in resource.schema.fields
         ]
         assert (
             expected_resource["schema"].get("primary_key", [])
             == resource.schema.primary_key
         )
+
+
+description_compliant_tables = ["core_epacems__hourly_emissions"]
```
e-belfer: I think we could probably use `mocker` to do this more elegantly with a fake table and fake responses, which would make this more properly a unit test and prevent failures when this table description changes in the future. This could be a follow-up, as noted below. @jdangerx or @zschira might have helpful thoughts here.

src/pudl/metadata/classes.py
```diff
@@ -1130,10 +1128,92 @@ class CodeDataFrame(pr.DataFrameModel):
...
+                "quarterly",
+                "yearly",
+                "monthly",
+                "hourly",
+            ]
+            | None
+        ) = None
+        """If this resource has table_type timeseries, indicates the temporal resolution, otherwise None.
+        If table_type is timeseries and this value is None or otherwise left unset, will be filled in with a default resolution parsed from the resource id string."""
+        layer: Literal["raw", "_core", "core", "_out", "out", "test"] | None = None
```
e-belfer: Non-blocking: Seems like `_out` no longer exists - do we want to remove it as an option altogether?

src/pudl/scripts/generate_metadata_template.py
e-belfer: Yes, this works! I ran `resource_description -n out_eia860__yearly_emissions_control_equipment`

I think it's worth keeping around for now, we might find it's more lightweight for development than running the wizard.

src/pudl/metadata/classes.py
krivard: whoops, that's an abandoned leftover. removing

src/pudl/metadata/classes.py
krivard: all the description_ keys are additional really -- would having that suffix be useful on all of them?

src/pudl/metadata/classes.py
krivard: No; if there is a PK for the table and description_primary_key is set, description_primary_key will appear after the comma-delimited list of pk columns.

it's possible we'll find a use for that in the 2nd-pass migration, so i'd like to leave it for now and follow up with exclusions once we know whether or not we need it

src/pudl/metadata/classes.py
krivard: no; steady-state, structured table description metadata is stored in the python files as a dict on the description key, and is replaced with the rendered description string at build time. the rendered description is never committed to the repository.

src/pudl/metadata/descriptions.py
krivard: Sortof?

The list of components has up to four elements:

* timeseries resolution: might be from a manually-set timeseries resolution
* table type fragment 1: might be from a manually-set table type
* table type fragment 2: might be from a manually-set table type
* description_summary: manually set

src/pudl/metadata/resources/epacems.py
krivard: We need one migrated table to exercise the tests, otherwise coveragebot will throw a fit.

I just picked the least complicated source, and have marked it complete in the migration epic.

src/pudl/metadata/warnings.py
krivard: by before merge, I mean not as a follow-up task; I'll see if someone has an appropriate link on slack

test/integration/console_scripts_test.py
krivard: fully migrated as far as the first pass is concerned -- no new info, just moving existing info into its new home

test/unit/metadata_test.py
krivard: intention here is to check the structure and content of the metadata dictionaries themselves, not the build process -- we want there to be some system that screams if the metadata for a table does not meet some minimum standard. a mock would not accomplish this.

as written, part of migrating a table would be adding it to this list so that it gets tested. once the migration is complete, we drop this list and switch below to `sorted(PUDL_RESOURCES.keys())`.

it's likely this goal is better suited to one of our validation systems, but writing it as a pytest (especially as a "only enforce this for tables we've migrated") was closer to my skillset.

src/pudl/metadata/warnings.py
krivard: any time someone who converted a parquet to csv (with any number of free online tools for doing so) and tried to open it with Excel would have regrets

src/pudl/metadata/warnings.py
krivard: @aesharpe is this the name of the standard or did I meticulously copy a capitalization typo?

src/pudl/metadata/classes.py
krivard: yes; these are the two `epa` tables:

- core_epa__assn_eia_epacamd
- core_epa__assn_eia_epacamd_subplant_ids

src/pudl/metadata/classes.py
e-belfer: I think the `additional_` sufficiently clarifies this for me, thanks! I agree, we should leave this behavior as is for now.

src/pudl/metadata/classes.py
krivard: Yes:

- core_ferc__codes_accounts

src/pudl/metadata/descriptions.py
```diff
@@ -162,57 +182,85 @@ def __init__(self, resource_id: str, settings: dict):
             getattr(self, f"_create_{trait}_description")(settings, defaults)
         self._assemble_usage_warnings(settings, defaults)
 
-    def _create_summary_description(self, settings, defaults):
+    def _create_summary_description(self, settings, defaults: "ResourceNameComponents"):
```
e-belfer, line 190: I would think this doesn't want to be in quotes?

src/pudl/metadata/descriptions.py
e-belfer: Is this just for our internal tracking? I get what the code is doing but I'm missing a why.

src/pudl/metadata/descriptions.py
```diff
@@ -162,57 +182,85 @@ def __init__(self, resource_id: str, settings: dict):
...
             description=" ".join(c for c in components if c is not None),
         )
 
-    def _create_x_description(self, attr, lookup, settings, defaults):
+    def _create_x_description(
+        self,
+        attr: str,
+        lookup: dict[str, str],
+        settings: dict,
+        defaults: "ResourceNameComponents",
```
e-belfer, line 239: Same question here, if this is a class I don't think you need to quote it to typehint?

src/pudl/metadata/classes.py
krivard: (we solved this offline)

test/unit/metadata_test.py
e-belfer: I think it's a great goal but to me that then isn't a unit test! Perhaps this wants to live in test/integration/ until we think about if/how we want to refactor it?

src/pudl/metadata/descriptions.py
krivard: Pylance gets upset without them

src/pudl/metadata/descriptions.py
krivard: resolved offline

src/pudl/metadata/classes.py
```diff
@@ -1130,10 +1127,97 @@ class CodeDataFrame(pr.DataFrameModel):
...
+            Literal[
+                "quarterly",
+                "yearly",
+                "monthly",
+                "hourly",
+            ]
+            | None
+        ) = None
+        """If this resource has :attr:`~PudlResourceDescriptor.PudlDescriptionComponents.table_type_code` timeseries, indicates the temporal resolution, otherwise None.
+        If :attr:`PudlDescriptionComponents.table_type_code` is timeseries and this value is None or otherwise left unset, will be filled in with a default resolution parsed from the resource id string."""
```
jdangerx: tiny nit: the `~` above is to shorten the link to "table_type_code", right? should we shorten it here too? or maybe show it in full the first time someone sees it and then the short version in later references?

src/pudl/metadata/classes.py
```diff
@@ -1130,10 +1127,97 @@ class CodeDataFrame(pr.DataFrameModel):
...
+        * ferc_is_hard
+
+        Any items provided here will be listed before the automatically detected warnings.
+
+        If None or otherwise left unset, will be filled in with auto warnings only. If no auto warnings apply, hides the Usage Warnings section entirely."""
+        additional_summary_text: str | None = None
+        """A brief (~one-line) description of the contents of this resource.
+        If None or otherwise left unset, will be left blank.
+
+        If filled, should support whichever of the following scenarios is most appropriate for this resource:
```
jdangerx, line 1191: non-blocking: hmm, this is a funny smell - should this be two different fields if they apply in two different scenarios?  `table_type_extra_fragment` vs. `additional_summary_text`?

offline discussion: see above. slightly funky, but we can live with it. code is for doing things not for being pretty on its own.

src/pudl/metadata/classes.py
```diff
@@ -1130,10 +1127,97 @@ class CodeDataFrame(pr.DataFrameModel):
...
+        If :attr:`PudlDescriptionComponents.table_type_code` is timeseries and this value is None or otherwise left unset, will be filled in with a default resolution parsed from the resource id string."""
+        layer_code: Literal["raw", "_core", "core", "out", "test"] | None = None
+        """Indicates the degree of processing applied to the data in this resource.
+        If None or otherwise left unset, will be filled in with a default layer parsed from the resource id string."""
+        source_code: str | None = None
+        """Indicates the source we wish to display for this resource; distinct from :attr:`PudlResourceDescriptor.source_ids` because here we want the majority source (or grouped source if truly mixed) and not a complete list of all sources used for this resource.
+        If set, should be a known data source shortcode like "eia923" or one of the grouped shortcodes from :data:`~pudl.metadata.descriptions.source_descriptions`.
+        If None or otherwise left unset, will be filled in with a default source parsed from the resource id string."""
+        usage_warnings: list[str | dict] | None = None
```
jdangerx, line 1160: non-blocking: funny smell here - should we have "stock usage warnings" and "custom usage warnings"?

offline discussion: splitting this into two things makes it harder for someone to order these usage warnings. we can live with the slight code funkiness, especially since the documentation is pretty explicit about the two types of things that can live in here.

src/pudl/metadata/classes.py
```diff
@@ -1130,10 +1127,97 @@ class CodeDataFrame(pr.DataFrameModel):
...
+        If None or otherwise left unset, will be left blank.
+        This should only be set in truly obscure situations. If set, should be a complete sentence."""
+        additional_source_text: str | None = None
+        """A brief refinement on the source data for this table, such as indicating the Schedule or other section number.
+        If None or otherwise left unset, will be left blank.
+        If set, should make sense when displayed directly after the title of a datasource (see :data:`~pudl.metadata.descriptions.source_descriptions`); parentheticals work best here."""
+        additional_primary_key_text: str | None = None
+        """For resources with no primary key, a brief summary of what each row contains, and perhaps why a primary key doesn't make sense for this table.
+        If None or otherwise left unset, will be left blank.
+        This should only be set if the resource has no natural primary key. If set, should be a complete sentence or two."""
```
jdangerx: Based on comments above about

> No; if there is a PK for the table and description_primary_key is set, description_primary_key will appear after the comma-delimited list of pk columns.

We should change this advice:

```suggestion
        If there is a primary key for the table, description_primary_key will appear after the comma-delimited list of PK columns. Should be a complete sentence or two."""
```

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,437 @@
...
+table_type_fragments: dict[str, TableTypeFragments] = {
+    "assn": TableTypeFragments("Association table", "providing connections between"),
+    "codes": TableTypeFragments(
+        "Code table containing descriptions of categorical codes", "for"
+    ),
+    "entity": TableTypeFragments("Entity table containing static information", "about"),
+    "scd": TableTypeFragments(
+        "Slowly changing dimension (SCD) table", "describing attributes of"
+    ),
+    "timeseries": TableTypeFragments("time series", "of"),
```
jdangerx, line 61: nit/question: should this guy be capitalized?
```suggestion
    "timeseries": TableTypeFragments("Time series", "of"),
```

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,437 @@
...
+timeseries_resolution_fragments: dict = {
+    "quarterly": "Quarterly",
+    "yearly": "Annual",
+    "monthly": "Monthly",
+    "hourly": "Hourly",
+}
+"""More standard descriptive text to appear in the Summary (first line) of resource descriptions, for timeseries resources."""
+
+
+# [jul 2025 kmm] This is a mirror of pudl.metadata.PudlMeta, which can't be included here because it creates a circular import.
```
jdangerx, line 93: non-blocking: we could try pulling `PudlMeta` out into a new `metadata.core` module that gets imported in multiple spots. Would be nice to start breaking down that `metadata.classes` mega-module, though doing that whole task is obviously out of scope for this PR.

offline decision: out of scope for this PR.

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+        self.resource_id = resource_id
+        defaults = ResourceNameComponents(name=resource_id)
+
+        # make a copy and move all the description keys up a level
+        settings = dict(settings)
+        if isinstance(settings.get("description"), dict):
+            settings = settings | settings["description"]
+
+        # this is a little gross but calling 5 methods with the same signature individually makes me sad
```
jdangerx, line 185: idk if this is too clever by half but you could have some sort of "component registry" here...

```python
def component(func):
    func.__is_component = True
    return func


class Foo:
    def __init__(self):
        self.components = [
            func
            for name in dir(self)
            if callable(func := getattr(self, name))
            and getattr(func, "__is_component", False)
        ]
        for component in self.components:
            print(component()) # or, like, actually call the function with the settings + defaults

    @component
    def bar(self):
        return "bar"

    @component
    def baz(self):
        return "baz"

    def quux(self):
        return "quux"


Foo() # prints bar and baz but not quux
```

offline decision: add to refactor notes, probably OOS for this PR

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,437 @@
...
+            resource_id: a snake-case string uniquely identifying the resource; aka the table name.
+            settings: a dictionary of resource metadata, usually obtained as a :class:`~pudl.metadata.classes.PudlResourceDescriptor` model dump.
+        """
+        self.resource_id = resource_id
+        defaults = ResourceNameComponents(name=resource_id)
+
+        # make a copy and move all the description keys up a level
+        settings = dict(settings)
+        if isinstance(settings.get("description"), dict):
+            settings = settings | settings["description"]
```
jdangerx, line 183: I *think* you don't run into original-value-mutation issues here but I could be wrong:

```python
>>> settings = {"a": 1, "b": {"x": 24, "y": 25}}
>>> other_settings = settings | settings["b"]
>>> other_settings
{'a': 1, 'b': {'x': 24, 'y': 25}, 'x': 24, 'y': 25}
>>> settings
{'a': 1, 'b': {'x': 24, 'y': 25}}
```

But also this double-stores the description values which is a little annoying... I guess it's more annoying to deal with pulling keys out of different places in the actual handler functions though?

src/pudl/metadata/descriptions.py
```diff
@@ -0,0 +1,437 @@
...
+
+    This class computes and stores "final" description components from all available inputs.
+    There are six final description components:
+
+    * summary
+    * layer
+    * source
+    * primary_key
+    * details
+    * usage_warnings
```
jdangerx, line 161: This could be pulled out into its own little `ResolvedResourceDescription` namedtuple or something, that `build()` spits out.

Speaking of which, right now the actual "building" of the description happens in `__init__`, not `build()`, which is a little funky - usually the builder makes a built thing which is what the caller actually cares about. So maybe we have `build()` spit out the `ResolvedResourceDescription` and then there's `ResolvedResourceDescription.render_template()` and `ResolvedResourceDescription.summarize()` (or... `.__str__()`?).

Then the interface would be `builder.build().render_template()` or somesuch, which would require a few downstream changes, but the benefit would be that it's more obvious where to make changes when adding a component down the line.


offline decision: add to refactor notes but OOS for this PR

test/unit/metadata_test.py
```diff
@@ -194,16 +195,84 @@ def test_frictionless_data_package_non_empty():
...
         assert expected_resource["schema"]["fields"] == [
             f.name for f in resource.schema.fields
         ]
         assert (
             expected_resource["schema"].get("primary_key", [])
             == resource.schema.primary_key
         )
+
+
+description_compliant_tables = [
```
jdangerx: discussed offline: let's key off of "is `resource_dict["description"]` a dict?" and not have to manually maintain this list.

test/unit/metadata_test.py
```diff
@@ -194,16 +195,84 @@ def test_frictionless_data_package_non_empty():
...
+]
+# TODO: flip this to true after we do the second pass to set description_primary_key
+# everywhere that needs it
+CHECK_DESCRIPTION_PRIMARY_KEYS = False
+
+
+@pytest.mark.parametrize(
+    "resource_id", sorted(description_compliant_tables)
+)  # someday: sorted(PUDL_RESOURCES.keys()))
+def test_description_compliance(resource_id):
```
jdangerx, line 250: offline: decided to keep this in unit tests since we want to run it as a pre-commit hook, and the things it tests aren't the outputs of any of our code. just our human brains

jdangerx (approved):
Couple small fixes + some non-blocking structural changes. Approved since I don't feel like I need to re-review once the small fixes go through :cowboy_hat_face:

src/pudl/metadata/descriptions.py
krivard: No; timeseries resource summaries are of the form

```
[timeseries-resolution] [tabletype-subject] [tabletype-conjunction] [additional-summary-text]
```

So the capital letter goes with the timeseries resolution, not here in the subject fragment.

Adding to the refactor list: explore whether applying title-casing after fragments have been glued together is (1) possible and (2) clearer than carefully capitalizing the fragments themselves.

src/pudl/metadata/descriptions.py
krivard: oh weird, your explanation made sense so I tried dropping the copy step and pydantic immediately screamed about extra inputs. so something is clearly propagating back 👻 

leaving the copy in there for a future optimizer to puzzle out

PR #4168, opened 2025-04-03
fix: up fly app timeout

jdangerx:
# Overview

Nightly builds keep failing datasette step with a timeout. The app eventually starts up but only after the timeout is triggered. 

## What did you change?

15 minutes -> 25 minutes.

## Documentation

Don't think there's anything to change here.

# Testing

Just looked in the documentation for a cap on the wait timeout and couldn't find one.

I could have downloaded the nightly build full outputs from GCS and run the publish command, but that requires a bunch of disk space cleanup and that seemed like a distraction, whereas the risk of 'push-and-wait' seems pretty low.

### DIFF
--- a/devtools/datasette/fly/production.toml
+++ b/devtools/datasette/fly/production.toml
@@ -31,4 +31,4 @@ primary_region = "bos"
     timeout = 2000
 
 [deploy]
-wait_timeout = "15m"
+wait_timeout = "25m"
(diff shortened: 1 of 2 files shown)


zaneselvans (approved):

zaneselvans:
Went ahead and merged this just so we can see if it fixes the problem tonight.

PR #4170, opened 2025-04-04
Make sec10k metadata consistent with archiver

zschira:
# Overview

Update SEC10k `sources` metadata to be consistent with archiver, and add `Sec10kSettings` class.

### DIFF
--- a/src/pudl/settings.py
+++ b/src/pudl/settings.py
@@ -181,7 +181,7 @@ class Sec10kSettings(GenericDatasetSettings):
 
     data_source: ClassVar[DataSource] = DataSource.from_id("sec10k")
 
-    years: list[int] = range(1993, 2025)
+    years: list[int] = range(1993, 2024)
     """The list of years to validate."""
 
     tables: list[str] = data_source.working_partitions["tables"]
--- a/src/pudl/settings.py
+++ b/src/pudl/settings.py
@@ -177,7 +177,7 @@ class PhmsaGasSettings(GenericDatasetSettings):
 
 
 class Sec10kSettings(GenericDatasetSettings):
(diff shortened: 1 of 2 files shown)


src/pudl/settings.py
```diff
@@ -176,6 +176,17 @@ class PhmsaGasSettings(GenericDatasetSettings):
     """The list of years to validate."""
 
 
+class Sec10kSettings(GenericDatasetSettings):
+    """An immutable pydantic model to validate PHMSA settings."""
+
+    data_source: ClassVar[DataSource] = DataSource.from_id("sec10k")
+
+    years: list[int] = range(1993, 2025)
```
zaneselvans, line 184: Doesn't the data only go through 2023 right now?

```suggestion
    years: list[int] = range(1993, 2024)
```

src/pudl/settings.py
```diff
@@ -176,6 +176,17 @@ class PhmsaGasSettings(GenericDatasetSettings):
     """The list of years to validate."""
 
 
+class Sec10kSettings(GenericDatasetSettings):
+    """An immutable pydantic model to validate PHMSA settings."""
```
zaneselvans, line 180:
```suggestion
    """An immutable Pydantic model to validate SEC 10-K settings."""
```

zaneselvans (changes requested):
I could be wrong but I think the range of years is wrong. I left suggested changes

src/pudl/settings.py
zschira: Oh yeah we haven't archived the 2024 data yet.

zaneselvans (approved):

PR #4183, opened 2025-04-07
Remove obsolete vs_self, vs_historical, & check_unique_rows tests

### LINKED ISSUE #4182: Remove data validations that we have decided not to migrate
# Overview

A number of data validations we're currently running are relatively low value, and complicated to migrate into the `dbt` testing framework, and aren't worth turning into their own asset checks, so we are removing them.  These include:

* `vs_self` & `vs_historical` hard to migrate, mostly exist to support exploring what reasonable bounds would be to check for the weighted quantile distribution checks.
* `check_unique_rows` -- only checks that PKs are unique. Pandera & SQLite already check this.

## What I did:

* I removed the functions implementing these checks from `pudl/validate.py`
* Removed all tests from under `test/validate` that relied on these functions.
* Updated all of the weighted quantile notebooks not to use `vs_self` or `vs_historical` (deleted related cells)
* For `check_unique_rows` I verified that every test was checking the set of columns that already defined the primary key for the table it was running on.  In 2 instances, this wasn't the case, but the columns being checked were an ad-hoc combination including plant `capacity_mw` in the FERC small plants tables, which are a trash fire and now use a psuedo-key.

zaneselvans:
# Overview

This PR removes a number of data validations that we're deprecating or that are duplicative, including:

- `check_unique_rows`
- `vs_self` (a statistical distribution check)
- `vs_historical` (another statistical distribution check)

Closes #4182

## What did you change?

* I removed the functions implementing these checks from `pudl/validate.py`
* Removed all tests from under `test/validate` that relied on these functions.
* Updated all of the weighted quantile notebooks not to use `vs_self` or `vs_historical` (deleted related cells)
* For `check_unique_rows` I verified that every test was checking the set of columns that already defined the primary key for the table it was running on.  In 2 instances, this wasn't the case, but the columns being checked were an ad-hoc combination including plant `capacity_mw` in the FERC small plants tables, which are a trash fire and now use a psuedo-key.

## Testing

- I ran all cells in all of the `test/validate/notebooks/*.ipynb` notebooks and removed offending cells. All but 2 now run successfully, and the remaining 2 failures are due to the reorganization of a columns in a table, not the removal of these tests.
- I ran `dbt build` against the two models that I updated with `unique_combination_of_columns` tests.
- Currently running the data validations locally...

# To-do list

- [x] Add 2 non-PK unique combination of columns tests to dbt
- [x] Run dbt tests against those two tables and make sure they pass
- [x] Run the data validations locally to make sure nothing blows up.

### DIFF
--- a/src/pudl/validate.py
+++ b/src/pudl/validate.py
@@ -365,91 +365,6 @@ def vs_bounds(
             )
 
 
-def vs_self(
-    df, data_col, weight_col, query="", title="", low_q=0.05, mid_q=0.5, hi_q=0.95
-):
-    """Test a distribution against its own historical range.
-
-    This is a special case of the :func:`pudl.validate.vs_historical` function, in which
-    both the ``orig_df`` and ``test_df`` are the same. Mostly it helps ensure that the
-    test itself is valid for the given distribution.
-    """
-    if weight_col is None or weight_col == "":
-        df["ones"] = 1.0
-        weight_col = "ones"
-    vs_historical(
-        df,
-        df,
-        data_col,
-        weight_col,
-        query=query,
-        low_q=low_q,
-        mid_q=mid_q,
-        hi_q=hi_q,
-        title=title,
-    )
-
-
-def vs_historical(  # noqa: C901
-    orig_df,
-    test_df,
-    data_col,
-    weight_col,
-    query="",
-    low_q=0.05,
-    mid_q=0.5,
-    hi_q=0.95,
-    title="",
-):
-    """Validate aggregated distributions against original data."""
-    if query != "":
-        orig_df = orig_df.copy().query(query)
-        test_df = test_df.copy().query(query)
-    if title != "":
-        logger.info(title)
-    if weight_col is None or weight_col == "":
-        orig_df["ones"] = 1.0
-        test_df["ones"] = 1.0
-        weight_col = "ones"
-    if low_q:
-        low_range = historical_distribution(orig_df, data_col, weight_col, low_q)
-        low_test = weighted_quantile(test_df[data_col], test_df[weight_col], low_q)
-        logger.info(f"{data_col} ({low_q:.0%}): {low_test:.6} >= {min(low_range):.6}")
-        if low_test < min(low_range):
-            raise ValueError(
-                f"Lower value {low_test} below lower limit {min(low_range)} "
-                f"in validation of {data_col}"
-            )
-
-    if mid_q:
-        mid_range = historical_distribution(orig_df, data_col, weight_col, mid_q)
-        mid_test = weighted_quantile(test_df[data_col], test_df[weight_col], mid_q)
-        logger.info(
-            f"{data_col} ({mid_q:.0%}): {min(mid_range):.6} <= {mid_test:.6} "
-            f"<= {max(mid_range):.6}"
-        )
-        if mid_test < min(mid_range):
-            raise ValueError(
-                f"Middle value {mid_test} below lower limit {min(mid_range)} "
-                f"in validation of {data_col}"
-            )
-        if mid_test > max(mid_range):
-            raise ValueError(
-                f"Middle value {mid_test} above upper limit {max(mid_range)} "
-                f"in validation of {data_col}"
-            )
-
-    if hi_q:
-        hi_range = historical_distribution(orig_df, data_col, weight_col, hi_q)
-        hi_test = weighted_quantile(test_df[data_col], test_df[weight_col], hi_q)
-        logger.info(f"{data_col} ({hi_q:.0%}): {hi_test:.6} <= {max(hi_range):.6}.")
-        if hi_test > max(hi_range):
-            raise ValueError(
-                f"Upper value {hi_test} above upper limit {max(hi_range)} "
-                f"in validation of {data_col}"
-            )
-
-
 def bounds_histogram(
     df, data_col, weight_col, query, low_q, hi_q, low_bound, hi_bound, title=""
 ):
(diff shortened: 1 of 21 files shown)


test/validate/eia_test.py
```diff
@@ -40,94 +39,3 @@ def test_no_null_cols_eia(pudl_out_eia, live_dbs, cols, df_name):
...
         pudl_out_eia.__getattribute__(df_name)(), cols=cols, df_name=df_name
     )
-
-
-@pytest.mark.parametrize(
-    "df_name,unique_subset",
-    [
-        (
-            "bf_eia923",
-            Resource.from_id("core_eia923__monthly_boiler_fuel").schema.primary_key,
```
zaneselvans, line 50: These are all being generated from the PK constraints, and so can definitely be removed.

test/validate/fbp_ferc1_test.py
```diff
@@ -34,20 +34,3 @@ def test_vs_bounds(pudl_out_ferc1, live_dbs, cases):
...
-    historical data.
-    """
-    if not live_dbs:
-        pytest.skip("Data validation only works with a live PUDL DB.")
-    else:
-        fbp_ferc1 = pudl_out_ferc1.fbp_ferc1()
-        for f in ["gas", "oil", "coal", "waste", "nuclear"]:
-            fbp_ferc1[f"{f}_cost_per_mmbtu"] = (
-                fbp_ferc1[f"{f}_fraction_cost"] * fbp_ferc1["fuel_cost"]
-            ) / (fbp_ferc1[f"{f}_fraction_mmbtu"] * fbp_ferc1["fuel_mmbtu"])
```
zaneselvans, line 51: I wonder if this calculation is why we have unweighted quantile checks for the FBP table?

test/validate/ferc1_test.py
```diff
@@ -73,52 +73,6 @@ def test_no_null_cols_ferc1(live_dbs, asset_value_loader, cols, asset_key):
...
-    [
-        (
-            "_out_ferc1__yearly_plants_utilities",
-            ["utility_id_ferc1", "plant_name_ferc1"],
-        ),
-        (
-            "out_ferc1__yearly_steam_plants_fuel_by_plant_sched402",
-            ["report_year", "utility_id_ferc1", "plant_name_ferc1"],
-        ),
-        (
```
zaneselvans, line 87: Verified that these are PKs

test/validate/ferc1_test.py
```diff
@@ -73,52 +73,6 @@ def test_no_null_cols_ferc1(live_dbs, asset_value_loader, cols, asset_key):
...
-                "report_year",
-                "utility_id_ferc1",
-                "plant_name_ferc1",
-                "capacity_mw",  # Why does having capacity here make sense???
-            ],
-        ),
-        (
-            "out_ferc1__yearly_plant_in_service_sched204",
-            ["report_year", "utility_id_ferc1", "ferc_account_label"],
-        ),
```
zaneselvans, line 108: Verified that this is an existing PK.

test/validate/ferc1_test.py
```diff
@@ -73,52 +73,6 @@ def test_no_null_cols_ferc1(live_dbs, asset_value_loader, cols, asset_key):
...
-        (
-            "out_ferc1__yearly_pumped_storage_plants_sched408",
-            [
-                "report_year",
-                "utility_id_ferc1",
-                "plant_name_ferc1",
-                "capacity_mw",  # Why does having capacity here make sense???
-            ],
-        ),
```
zaneselvans, line 104: Inclusion of `capacity_mw` here makes these kinda fishy -- I think it's probably a vestige of an older version of these tables? Not entirely sure. Current PK for these tables is a pseudo key `record_id` that just says where the data came from in the FERC 1 mess.

It might be reasonable to turn each of these into non-PK multi-column uniqueness tests within the `dbt` schemas.

test/validate/mcoe_test.py
```diff
@@ -91,27 +91,6 @@ def test_no_null_rows_mcoe(pudl_out_mcoe, live_dbs, df_name, thresh):
...
 
 
-@pytest.mark.parametrize(
-    "df_name,unique_subset",
-    [
-        ("hr_by_unit", ["report_date", "plant_id_eia", "unit_id_pudl"]),
-        ("hr_by_gen", ["report_date", "plant_id_eia", "generator_id"]),
-        ("fuel_cost", ["report_date", "plant_id_eia", "generator_id"]),
-        ("capacity_factor", ["report_date", "plant_id_eia", "generator_id"]),
-        ("mcoe", ["report_date", "plant_id_eia", "generator_id"]),
```
zaneselvans, line 101: Verified that these are all PKs for the `PudlTabl` supporting underscore tables that are referenced by the short names here.

test/validate/ferc1_test.py
zaneselvans: I went ahead and added them, since the test has been passing and it seems like a real, though heuristic, expectation.

src/pudl/validate.py
```diff
@@ -624,30 +515,6 @@ def plot_vs_bounds(df, validation_cases):
...
-def plot_vs_self(df, validation_cases):
-    """Validate a bunch of distributions against themselves."""
-    for args in validation_cases:
-        try:
-            vs_self(df, **args)
-        except ValueError as e:
-            logger.error(str(e))
-            warnings.warn("ERROR: Validation Failed", stacklevel=1)
-
-        historical_histogram(df, test_df=None, **args)
```
krivard, line 636: with these changes it looks like historical_histogram isn't called anywhere else, so we can get rid of it, too

test/validate/fbp_ferc1_test.py
krivard: Kinda? we're covering that using a validation model: 
https://github.com/catalyst-cooperative/pudl/blob/71e9059f256c7e96766a2bffcd3a070630e774bc/dbt/models/ferc1/out_ferc1__yearly_steam_plants_fuel_by_plant_sched402/schema.yml#L44

the model does the cost per mmbtu calculation, so the quantile check is unweighted

krivard (approved):
Good to merge once we drop historical_histogram, unless you have reason to keep it.

Is the ability to interactively check something similar to historical/self in a notebook something that's nice to have? if so maybe file a ticket for someone to figure out how to do this with dbt.

src/pudl/validate.py
zaneselvans: You know, I removed this initially, and then I put it back because it seems like we'll need to debug some failing `vs_bounds` tests or better understand the historical annual variability in the distributions that we're checking in order to select bounds that can pass in both the full ETL and fast ETL scenarios, and it seemed like it would be useful in that context.

test/validate/fbp_ferc1_test.py
zaneselvans: It just seemed weird and possibly wrong to me (even though I'm the one that made the old test) that it would *not* be weighted, but seeing that the cost per MMBTU is being constructed from these already aggregated values, rather than a bunch of individual records with very different weights, it seemed like maybe there was no other choice.

krivard: sounds plausible!

PR #4190, opened 2025-04-09
Aggregated eia930 demand

### LINKED ISSUE #4139: Create and test aggregated EIA-930 demand outputs
# Overview

- In addition to BA subregions and the full BAs, Tyler Ruggles and Alicia Wongel published aggregations of the BA demand up to the level of the interconnect (East/West/ERCOT) which are useful in other modeling applications.
- We'll want to produce similar outputs in a table of aggregated values that's separate from `out_eia930__hourly_operations`

# Success Criteria

- We've produced aggregate outputs that are comparable replacements for the aggregated EIA-930 outputs published by Tyler Ruggles and Alicia Wongel

# Next steps

* [x] Define mappings between individual BAs and the interconnects / CONUS (everything)
* [x] Create a new table/asset that aggregates the BA demand up to the level of the interconnects & CONUS.
* [ ] Use tools developed in #4137 to compare our aggregated outputs with those previously published.

zschira:
# Overview

Closes #4139.

This PR adds an output table containing aggregated EIA930 demand to the levels of BA, interconnect, and contiguous US.

## Documentation

# Testing

## To-do list


src/pudl/metadata/codes.py
```diff
@@ -1810,104 +1810,104 @@
     "core_eia__codes_balancing_authorities": {
```
zschira, line 1810: I ended up adding the mapping of region->interconnect to `core_eia__codes_balancing_authorities` . It's not the most strictly well normalized place to add this info, but it seemed fairly logical/straightforward

src/pudl/metadata/fields.py
```diff
@@ -1144,6 +1144,11 @@
         "description": "Electricity demand (energy) within a given timeframe.",
         "unit": "MWh",
     },
+    "demand_agg_mwh": {
```
zschira, line 1147: Does this naming make sense? I know we use the `agg` tag in some column names, but not sure if we have formal conventions for when to do this.

src/pudl/metadata/resources/eia930.py
```diff
@@ -192,6 +192,31 @@
...
+            """EXPERIMENTAL / WORK-IN-PROGRESS, 2025-04-04.
+
+This aggregates imputed demand from the ``out_eia930__hourly_operations`` table.
+It aggregates demand to the level of EIA regions, interconnects, and the contiguous United States."""
+        ),
+        "schema": {
+            "fields": [
+                "datetime_utc",
+                "code",
+                "aggregation_group",
```
zschira, line 206: I think this field was used for `gridpathtoolkit`, and it seemed like it made sense here. Need to override the field metadata if we want to use this column name though.

zschira (commented):
@zaneselvans I'm working on an updated example notebook with some explorations of the aggregated data, but had a couple small questions about the implementation that you might have thoughts on.

src/pudl/metadata/codes.py
zaneselvans: Yeah these coding tables aren't perfectly normalized, but they're small and useful and I think it's okay.

src/pudl/metadata/fields.py
zaneselvans: I don't think we need a separate column name here -- the variable that's being reported is still just `demand_mwh` (or maybe `demand_imputed_pudl_mwh`) -- the fact that it's aggregated up doesn't change the nature of the variable.  I think we should convey the nature of the geographic aggregation in the table level descriptions.

src/pudl/metadata/resources/eia930.py
zaneselvans: We have infrastructure for resource (table) specific and data source specific column definitions, down at the bottom of `fields.py` which will override the generic field type/descriptions/constraints.  So if this will show up with different descriptions in the two diferent data sources, probably we should define the fields for the data sources, rather than globally.

However, I had trouble making this work in the SEC 10-K PR.  But it's supposed to work! And it does work for some other data sources, so I think there was just a bit of plumbing that I was missing.

src/pudl/metadata/resources/eia930.py
```diff
@@ -192,6 +192,31 @@
...
+                "demand_agg_mwh",
+            ],
+            "primary_key": [
+                "datetime_utc",
+                "code",
```
zaneselvans, line 211: What does code mean here? That's awfully generic. Need to explain the table description, or use a more descriptive column label.

src/pudl/metadata/resources/eia930.py
zaneselvans: What values do we expect to find in `aggregation_group`? `conus`, `eastern`, `western`, `ercot`, region codes? individual BA codes? Should explain that in the table description.

zaneselvans: Maybe I'm not understanding the different roles of `aggregation_group` and `code` in this context.

src/pudl/metadata/fields.py
```diff
@@ -2520,6 +2525,11 @@
         "description": "Original reported energy interchange between adjacent balancing authorities.",
         "unit": "MWh",
     },
+    "interconnect_code_eia": {
+        "type": "string",
+        "description": "EIA interconnect code.",
+        "constraints": {"enum": {"EASTERN", "WESTERN", "ERCOT"}},
+    },
```
zaneselvans, line 2532: In most cases outside of canonical abbreviations (e.g. for states) we've defaulted to using `snake_case` for code values, since otherwise you end up with a mix of cases so it seems a little funny for just these to be all caps.

src/pudl/metadata/resources/eia930.py
zaneselvans: In the GridPath RA Toolkit table, I think `aggregation_group` corresponds to the meaning that `code` has here -- it's the identifier of the individual group, not a category that describes what kind of group it is. Maybe it would make sense here to have

* `aggregation_group` -- the ID of the group, like `ERCOT` or `CONUS`
* `aggregation_level` -- the granularity of aggregation, e.g. `conus`, `interconnect`, `region_eia`

src/pudl/metadata/fields.py
zschira: Makes sense, I went with `demand_imputed_pudl_mwh` so it's still clear this is imputed data.

src/pudl/metadata/fields.py
zschira: Fixed

src/pudl/metadata/resources/eia930.py
zschira: I think the `aggregation_group/level` combo makes sense. Fixed

zschira:
Comparisons between our aggregated demand and the baseline from Tyler/Alicia can be found [here](https://www.kaggle.com/code/zachschira/eia930-imputation-sanity-checks/notebook#Compare-aggregated-demand). On first glance our version looks better, but it's hard to say if that is genuinely the case.

src/pudl/metadata/fields.py
```diff
@@ -6617,6 +6629,11 @@
             "description": "A manually assigned PUDL utility ID for the owner company that is responsible for the day-to-day operations of the generator, not the operator utility. May not be stable over time."
         }
     },
+    "out_eia930__hourly_demand": {
+        "aggregation_group": {
+            "description": "Label indentifying a group of aggregated deamnd."
```
zaneselvans, line 6634:
```suggestion
            "description": "Label identifying an group of balancing authorities to be used in aggregating demand E.g. a region of the US or a whole interconnect."
```

src/pudl/metadata/resources/eia930.py
```diff
@@ -192,6 +192,33 @@
         "etl_group": "eia930",
         "create_database_schema": False,
     },
+    "out_eia930__hourly_demand": {
```
zaneselvans, line 195: I think the name of this table should make it clear that this isn't the generic place to go for all demand -- it's just the aggregations. Also, in the baseline data, is there no aggregation of any of the other unimputed variables for use alongside the demand? Is there a good reason not to also provide aggregated (though not yet imputed) net generation and demand forecasts?

```suggestion
    "out_eia930__hourly_aggregated_demand": {
```

zaneselvans (changes requested):
Some minor naming suggestions and questions inline.

Looking at the comparison charts, our results and theirs seem really different, like unreasonably different. I suspect that there's a problem in the way we're extracting their aggregated values, since it seems like there are duplicate values in every hourly timestep which should never be the case, and that seems to be what's resulting in the hourly serration in the baseline time series. I think we need to figure that out before we can actually compare results. My intuition is to look for a bad merge somewhere that ends up duplicating all the timestamps?

But I agree that our hourly aggregated values look cool and normal.

image

image

image

src/pudl/metadata/fields.py
```diff
@@ -6617,6 +6629,11 @@
             "description": "A manually assigned PUDL utility ID for the owner company that is responsible for the day-to-day operations of the generator, not the operator utility. May not be stable over time."
         }
     },
+    "out_eia930__hourly_demand": {
+        "aggregation_group": {
+            "description": "Label identifying an group of balancing authorities to be used in aggregating demand E.g. a region of the US or a whole interconnect."
```
zaneselvans, line 6634:
```suggestion
            "description": "Label identifying a group of balancing authorities to be used in aggregating demand E.g. a region of the US or a whole interconnect."
```

src/pudl/metadata/resources/eia930.py
```diff
@@ -192,6 +192,33 @@
...
+        "schema": {
+            "fields": [
+                "datetime_utc",
+                "aggregation_level",
+                "aggregation_group",
+                "demand_imputed_pudl_mwh",
+            ],
+            "primary_key": [
+                "datetime_utc",
+                "aggregation_level",
```
zaneselvans, line 213: Does `aggregation_level` really need to be part of the primary key?

zaneselvans (approved):
Looking at the updated comparison notebook I see that our demand curves are different from the baseline even for timesteps in which no imputation has been done.  Do we know why that is? It seems particularly clear during the hours of lowest daily demand.

I could imagine it's because they're adding up sub-region demand to get BA demand, and we're taking the BA demand directly. Are there other potential explanations?

zaneselvans:
It looks like the table name didn't get updated to `out_eia930__hourly_aggregated_demand` in the `schema.yml` and I'm surprised that this didn't result in a unit test failure since I would have thought the dbt schema and the PUDL schema would have been found to be in conflict with each other.

src/pudl/metadata/resources/eia930.py
zaneselvans: If the uniqueness of `aggregation_group` values across different `aggregation_level`'s is incidental then it probably makes sense to keep it in the PK.

PR #4212, opened 2025-04-24
Update metadata for eia_bulk_elec to eiaapi

e-belfer:
# Overview

In https://github.com/catalyst-cooperative/pudl-archiver/pull/637 we are changing the `eia_bulk_electricity` archiver to archive _all_ EIA bulk resources. The first step to doing so is to update the metadata in PUDL.

## What did you change?
* Updated the description to capture descriptions of all of the datasets in the EIA bulk repository
* Updated the name to `eia_bulk` throughout PUDL, **except** for the asset group name which I've kept as eia_bulk_elec (is this confusing)?
* Updated the working partition

Out of scope: Actually changing the ETL to pass the correct partition at the extract step - we will do this when we update the data after the archive is made!

## Documentation

# Testing

## To-do list

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -71,6 +71,11 @@ Quality of Life Improvements
 * We now publish a `Frictionless data package
   <https://datapackage.org/standard/data-package/>`__ describing our Parquet
   outputs, with the name ``pudl_datapackage.json``. See :issue:`4069` and :pr:`4070`.
+* We renamed ``eia_bulk_elec`` to ``eiaapi`` to conform to our dataset naming protocols
+  and reflect the expansion of the EIA Bulk API archive to include all datasets
+  published through the EIA API, not just the bulk electricity data. See `this PUDL
+  archiver issue <https://github.com/catalyst-cooperative/pudl-archiver/issues/628>`__
+  and PR :pr:`4212`.
 
 New Tests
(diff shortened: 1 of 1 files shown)


krivard (approved):
lgtm!

krivard:
Oh that IndexError is odd. Let me take a closer look

krivard (approved):
Looks good! Keeping _elec in the asset names feels long-term wise and worth any short-term confusion, especially since there are at least two people who know about it.

zaneselvans:
For now the assets that are derived from this data source are specific to the Electricity data API, and I imagine that if we were going to pull in data from the other bulk API JSON dumps, they would probably have their own assets and asset groups, so I think changing the data source names up to the point where there's a partition specific (Electricity) asset, which refers to that subset of the larger data source makes sense.

The extract/transform for the EIA fuel price data is also a little nonstandard. We go pretty directly into generating a table that's useful for filling the fuel receipts and costs missing data, which might be weird/confusing.

Separately... it has always bugged me that this data source alone had underscores in its ID -- I'd be partial to going from `eia_bulk_elec` to just `eiabulk` to keep it in line with all the others. Another string we've used to identify data coming from this source is `eiaapi` (when we use the fuel prices in the fuel receipts & costs table).

PR #4213, opened 2025-04-24
Fix aggregated eia930 imputed demand row counts

zschira:
# Overview

Fixes row counts for new EIA 930 aggregated imputed demand in dbt row counts test.


krivard (approved):
👍

zschira:
Thanks!

PR #4219, opened 2025-04-27
Use correct target in dbt_helper integration test

### LINKED ISSUE #4218: Nightly Build Failure 2025-04-26
# Overview

The nightly builds are unable to update what commit the `nightly` tag points at, because we run the `dbt_helper` script to test that it works and (unexpectedly!) it ends up modifying the Fast ETL row counts CSV. So, why is it finding different row counts? How is it different?

And then, I think our expectation is that if this update to the `nightly` tag fails, then we shouldn't go ahead with deploying the build outputs. But it is deploying them.

# Relevant logs

```
+ echo 'Updating nightly branch to point at nightly-2025-04-26.'
Updating nightly branch to point at nightly-2025-04-26.
+ git fetch --force --tags origin nightly-2025-04-26
From https://github.com/catalyst-cooperative/pudl
 * tag                   nightly-2025-04-26 -> FETCH_HEAD
+ git fetch origin nightly:nightly
From https://github.com/catalyst-cooperative/pudl
 * [new branch]          nightly    -> nightly
+ git checkout nightly
error: Your local changes to the following files would be overwritten by checkout:
	dbt/seeds/etl_fast_row_counts.csv
Please commit your changes or stash them before you switch branches.
Aborting
```

zaneselvans:
# Overview

We have an integration test for the `dbt_helper` script which gets run in the nightly builds, but it had the `etl-fast` target hard-coded, so it was incorrectly updating row counts based on the Full ETL outputs, causing the nighty tag update to fail.

- This PR updates it to use whatever the correct dbt target is, based on the ETL settings file (full vs. fast) as we do for the other dbt tests.
- In the course of debugging this I also saw that we were echoing the PUDL Bot's PAT to our logs, so I changed the build script to avoid doing that.
- I revoked the old PAT and created a new one.
- There were also still some SEC 10-K tables being excluded from the integration tests, which is no longer appropriate, since those tables are now brought in using Zenodo just like everything else.

Closes #4218 

# Testing

## To-do list

- [x] Run the integration tests in CI using `workflow_dispatch` to see if they still pass
- [ ] Run the `build-deploy-pudl` GitHub Action manually.

### DIFF
--- a/test/integration/dbt_test.py
+++ b/test/integration/dbt_test.py
@@ -12,21 +12,30 @@ from pudl.io_managers import PudlMixedFormatIOManager
 logger = logging.getLogger(__name__)
 
 
-# These tables need to be excluded from our GitHub CI until they are being brought into
-# PUDL in the normal way via a Zenodo archive. Currently the row count checks don't
-# fail -- they result in an error (since the tables don't exist at all.)
-SEC10K_EXCLUDE = [
-    "core_sec10k__quarterly_company_information",
-    "core_sec10k__quarterly_exhibit_21_company_ownership",
-    "core_sec10k__quarterly_filings",
-    "out_sec10k__parents_and_subsidiaries",
-]
+@pytest.fixture(scope="module")
+def dbt_target(test_dir: Path, request) -> str:
+    """Fixture defining the dbt target based on the full/fast ETL spec."""
+    # Identify whether we're running the full or fast ETL, and set the dbt target
+    # appropriately (since we have different test expectations in the two cases)
+    if request.config.getoption("--etl-settings"):
+        etl_settings_yml = Path(request.config.getoption("--etl-settings"))
+    else:
+        etl_settings_yml = Path(
+            test_dir.parent / "src/pudl/package_data/settings/etl_fast.yml"
+        )
+    if etl_settings_yml.name == "etl_full.yml":
+        dbt_target = "etl-full"
+    elif etl_settings_yml.name == "etl_fast.yml":
+        dbt_target = "etl-fast"
+    else:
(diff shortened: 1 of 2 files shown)


test/integration/dbt_test.py
```diff
@@ -12,21 +12,30 @@
...
 
-# These tables need to be excluded from our GitHub CI until they are being brought into
-# PUDL in the normal way via a Zenodo archive. Currently the row count checks don't
-# fail -- they result in an error (since the tables don't exist at all.)
-SEC10K_EXCLUDE = [
-    "core_sec10k__quarterly_company_information",
-    "core_sec10k__quarterly_exhibit_21_company_ownership",
-    "core_sec10k__quarterly_filings",
-    "out_sec10k__parents_and_subsidiaries",
-]
```
zaneselvans, line 23: These should all be available as normal tables now.

test/integration/dbt_test.py
```diff
@@ -12,21 +12,30 @@
...
+@pytest.fixture(scope="module")
+def dbt_target(test_dir: Path, request) -> str:
```
zaneselvans, line 16: Made a fixture since we use it twice.

test/integration/dbt_test.py
```diff
@@ -103,7 +97,7 @@ def test_dbt_helper(pudl_io_manager: PudlMixedFormatIOManager, script_runner):
         [
             "dbt_helper",
             "add-tables",
-            "--etl-fast",
+            f"--{dbt_target}",
```
zaneselvans, line 100: I think it would be more intuitive if we used the same idiom with this script as dbt does, e.g. `--target etl-fast`

pudlbot (approved):

PR #4223, opened 2025-04-28
Make dbt target selection explicit in dbt_helper arguments.

### LINKED ISSUE #4218: Nightly Build Failure 2025-04-26

zaneselvans:
# Overview

- Change the `dbt_helper` script to explicitly select the dbt target on the command line, either `--target etl-full` (the default) or `--target etl-fast`.
- This is the same as the CLI for dbt and makes parsing the targets more uniform in functions that need this data / the script tests.

Closes #4218 

## Documentation

- [x] Updated the help message on `dbt_helper`
- [x] Updated our dbt README.md

## To-do list

- [x] Run unit tests
- [x] Run [integration tests in CI](https://github.com/catalyst-cooperative/pudl/actions/runs/14715798193/job/41298768063) on GitHub
- [ ] Run branch deployment this afternoon.


dbt/README.md
```diff
@@ -13,7 +13,7 @@ section below for examples using these profiles.
 To setup the `dbt` project, simply install the PUDL `conda` environment as normal,
 then run the following commands from this directory.
 
-```
+```bash
```
zaneselvans, line 16: Adding languages to the code fencing since my linter was complaining about it, and it allows syntax highlighting.

src/pudl/scripts/dbt_helper.py
```diff
@@ -353,11 +354,11 @@ class TableUpdateArgs:
...
-    "--etl-fast",
-    default=False,
-    is_flag=True,
-    type=bool,
-    help="Update row counts for fast ETL counts.",
+    "--target",
+    default="etl-full",
+    type=click.Choice(["etl-full", "etl-fast"]),
+    show_default=True,
+    help="What dbt target should be used as the source of new row counts.",
```
zaneselvans, line 361: Rather than a flag that's used to specify the non-default option, make it an argument with a limited set of acceptable values, making the target explicit for easier parsing elsewhere (e.g. in the tests) This argument also mirrors the `--target` option in dbt.

test/integration/dbt_test.py
```diff
@@ -97,7 +97,8 @@ def test_dbt_helper(
         [
             "dbt_helper",
             "add-tables",
-            f"--{dbt_target}",
+            "--target",
+            dbt_target,
```
zaneselvans, line 101: The logic here is simpler and more legible with an arg that explicitly sets the target (full/fast).

src/pudl/scripts/dbt_helper.py
marianneke: Agreed, much more practical, and easier to change should we ever have more targets

dbt/README.md
```diff
@@ -186,13 +186,13 @@ end). This should show you the years and the expected and found row counts. If t
...
-```
-python devtools/dbt_helper.py add-tables {table_name} --row-counts-only --clobber --use-local-tables
+```bash
+python devtools/dbt_helper.py add-tables {table_name} --target etl-full --row-counts-only --clobber --use-local-tables
 ```
 
 This will tell the helper script to overwrite the existing row counts with new row
 counts from the table in your local `PUDL_OUTPUT` stash. If you want to update the
-`etl-fast` row counts, simply append the option `--etl-fast` to the above command.
+`etl-fast` row counts, use `--target etl-fast` instead of `--target etl-full`.
```
marianneke, line 195:
```suggestion
`etl-fast` row counts, use `--target etl-fast` instead of the default `--target etl-full`.
```

marianneke (approved):
lgtm! (tiny readme suggestion)

PR #4245, opened 2025-05-07
Remove deprecated fbp_ferc1 vs_bounds data validations.

### LINKED ISSUE #4106: Migrate vs_bounds tests to dbt
Many tests use the method `vs_bounds` defined in `validate.py`. These tests compute a quantile (either weighted or standard) and compare the value to some configurable bounds. These tests can be migrated into the new framework using the `dbt_helper` script.

However, it's unclear how valuable or valid the unweighted quantile checks that are being applied to the FERC tables are. Conceptually, these should be weighted just like the other validation tests. Why aren't they?

- @marianneke has updated the SQL based weighted quantile calculation to be continuous in #4090. 
- However there are some tests that got migrated and made more lax using the discrete distribution, which need to be reverted.
- The rest of the weighted quantile tests should be migratable using the new continuous distribution macro.

zaneselvans:
# Overview

We forgot to remove one data validation test that was still referring to the deleted `vs_bounds` parameters in `pudl.validate` and that caused the builds to fail on 2025-05-07. This removes the deprecated validation, which was the only validation specific to the ferc1 fuel by plant table.

Closes #4106


krivard (approved):
🚀

PR #4247, opened 2025-05-07
Eia930 imputation comparison

### LINKED ISSUE #4205: Test alternative setups for EIA 930 imputation
### Overview

Our timeseries imputation implementation works by aligning many correlated timeseries' which have had missing/anomalous values flagged for imputation, and using the all of these timeseries' for reference to impute these values. While the exact mechanism for the imputation is somewhat of a black box to us, we believe that including good/complete timeseries data when imputing messier/less complete data would help to improve the imputation for this data. As such, we've theorized that including EIA 930 BA level demand data with subregion level data could improve the subregion imputation results, as the smaller subregions are often messier than the larger BA's. 

### Success Criteria

We choose a final setup for our subregion/BA level imputations backed by an empirical analysis for comparing different timeseries imputations. The exact methods for comparing these imputations will be developed/tracked by #4137.

### Tasks
* [ ] Rework imputation to use both BA/subregion level demand in a single big imputation
* [ ] Compare results
* [ ] Tweak imputation implementation further and compare if necessary

zschira:
# Overview

Closes #4205.

This PR tests different arrangements for EIA 930 imputation and compares them. Specifically, it tests performing imputation on BA/subregion demand separately, and combining all timeseries' from both into a single imputation run.

To test these different imputation arrangements, this PR also adds infrastructure to test imputation and compute a quantitative score of imputation. This works by finding months with a high ratio of values which were flagged for imputation, and using this pattern of flagging to drop values from otherwise good months of data. With this we are able to test imputation in realistic scenarios where we have good baseline reported data which we can compare to. To do this comparison, it will then compute Mean Squared Error. We could theoretically compute a separate metric, but MSE seems good enough to compare the relative performance of different methods.

As it stands now, this PR SHOULD NOT be merged. It contains both versions of the imputation to make comparison easy. Once we feel good about the comparisons, and have decided one version is decidedly better, we should remove the alternate version before merging.

## What did you change?
I added the combined imputation assets, called `out_eia930__hourly_subregion_demand_from_combined_imputation` and `out_eia930__hourly_operations_from_combined_imputation`, as well as the simulation/scoring infrastructure.

## Documentation

# Testing

## To-do list

- [ ]


src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -179,6 +180,26 @@ def pivot_aligned_timeseries_dataframe(
     return matrix.reindex(all_hours)
 
 
+@pa.check_types
```
zschira, line 183: If we want we could potentially make this an asset check and do the simulation/scoring in production, which could be nice to keep track of. That being said, adding in the simulation does theoretically alter the imputation at least slightly, as it adds new timeseries' when performing imputation.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1581,147 @@ def filter_missing_values(
...
+    )
+
+    # Use reference month to get hours which should be flagged in simulation month
+    simulation_data = _get_simulation_data(
+        df, pd.concat([bad_months.reset_index(), good_months.reset_index()], axis=1)
+    )
+
+    # Add new columns to timeseries/flag matricies with simulated data
+    # These columns are identical to original columns with simulated flagged values Nulled
+    for simulation_id in simulation_data.simulation_id_col.unique():
```
zaneselvans, line 1705: More narrative explanation of what is going on in this block would be helpful. Probably both inline comments, and a docstring somewhere that lays out the whole system of simulating nulls for testing imputation methods.

zaneselvans (changes requested):
I left a few substantive questions about how our choices of what to impute and the integration of the simulated data into the overall imputation process affects things. But in addition to that I just feel like this needs more high-level narrative documentation that outlines what the simulation process is trying to do. I tried to highlight the places that felt especially confusing. And probably some of my substantive questions wouldn't be necessary if I had a better mental model for what's going on.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1581,147 @@ def filter_missing_values(
     return df
 
 
+@dataclass
+class SimulateFlagsSettings:
+    """Define settings used to simulate flagged values for scoring imputation."""
+
+    num_sections: int = 30
+    """The number of sections of data to simulate."""
```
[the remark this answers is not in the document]
zschira: Yeah it is hard-coded to a month, so I've changed this.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1581,147 @@ def filter_missing_values(
...
+    min_flag_rate: float = 0.1
+    """Min ratio of bad points in a section of data to be used for reference."""
+    max_flag_rate: float = 0.5
+    """Min ratio of bad points in a section of data to be used for reference."""
```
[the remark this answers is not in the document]
zschira: Fixed

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1581,147 @@ def filter_missing_values(
...
+
+class SimulationDataFrame(pa.DataFrameModel):
+    """Collect months of data which will be used to simulate flagged values.
+
+    Each row in this dataframe will contain a reference ID and month, and a simulation
+    ID and month. The reference ID/month points to a month of reported data where
+    a high ratio of values were flagged for imputation, and the simulation ID/month
+    points to a month of reported data with no flagged values. The flagged values
+    in the reference month will be used to simulate flagged values in the simulation
+    month.
```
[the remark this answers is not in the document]
zschira: This is a good question, and I don't feel like I necessarily have a definitive answer. My thinking was that this approach was a good balance since it will use timeseries' with lots of imputed values for reference months but use timeseries' with generally "good" data to validate against. This way we should still be testing scenarios where there's lots of missing values at least for the simulated months, even if the rest of the timeseries is pretty "good".

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1581,147 @@ def filter_missing_values(
...
+    simulation_id_col: Series[Any]
+    simulation_month: Series[pa.dtypes.DateTime]
+
+
+@pa.check_types
+def _get_simulation_data(
+    df: DataFrame[AlignedTimeseriesDataFrame],
+    simulation_df: DataFrame[SimulationDataFrame],
+):
+    """Transform flagged pattern from reference periods to simulation periods."""
```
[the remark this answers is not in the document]
zschira: I tried to add some more detail here. Let me know if you have any questions still.

src/pudl/output/eia930.py
```diff
@@ -74,14 +75,90 @@ def _years_from_context(context) -> list[int]:
...
+    id_col="generic_id",
+    settings=ImputeTimeseriesSettings(
+        simulate_nulls_settings=SimulateFlagsSettings(),
+    ),
+    output_io_manager_key="io_manager",
+)
+
+
+@asset(io_manager_key="parquet_io_manager")
+def out_eia930__hourly_operations_from_combined_imputation(
```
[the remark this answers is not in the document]
zschira: I actually hadn't been materializing these because the simulation/MSE computation is done upstream in the assets created by the asset factory. I've switched these to use the default asset manager for now though.

src/pudl/output/eia930.py
```diff
@@ -74,14 +75,90 @@ def _years_from_context(context) -> list[int]:
...
+    return _out_eia930__hourly_operations.merge(
+        _out_eia930__combined_imputed_demand[
+            [
+                "datetime_utc",
+                "generic_id",
+                "demand_imputed_pudl_mwh",
+                "demand_imputed_pudl_mwh_imputation_code",
+            ]
+        ],
+        left_on=["datetime_utc", "combined_subregion_ba_id"],
```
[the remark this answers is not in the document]
zschira: Oh yeah, again I haven't been creating these at this point. I'll fix if we end up going with this imputation arrangement (although I think we probably won't)

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1664,6 +1827,9 @@ def _flag_timeseries_matrix(
     ) -> tuple[pd.DataFrame, pd.DataFrame]:
         """Flag/Null anomalous and missing values."""
         matrix, flags = flag_ruggles(matrix)
+        if (simulate_nulls_settings := settings.simulate_nulls_settings) is not None:
```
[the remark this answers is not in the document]
zschira: Added to the docstring, and I added a comment to the `ImputeTimeseriesSettings` class about the simulation settings.

src/pudl/output/eia930.py
```diff
@@ -74,14 +75,90 @@ def _years_from_context(context) -> list[int]:
...
+imputed_combined_demand_assets = impute_timeseries_asset_factory(
+    input_asset_name="_out_eia930__combined_demand",
+    output_asset_name="_out_eia930__combined_imputed_demand",
+    years_from_context=_years_from_context,
+    value_col="demand_reported_mwh",
+    imputed_value_col="demand_imputed_pudl_mwh",
+    id_col="generic_id",
+    settings=ImputeTimeseriesSettings(
+        simulate_nulls_settings=SimulateFlagsSettings(),
+    ),
```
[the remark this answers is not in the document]
zschira: Yeah if `simulate_flags_settings` (I changed the name for consistency) is `None`, then no simulation will happen. This is basically just using default imputation settings, and enabling simulation with the default simulation settings.

src/pudl/analysis/timeseries_cleaning.py
zschira: Added some more comments

src/pudl/analysis/timeseries_cleaning.py
zschira: > I'm not sure I understand what you mean here. Can you say more about how the testing would impact the actual imputation? Is it just that we're adding a whole new simulated time series into the collection of time series that are used in the imputation process? So in the case where we run this kind of validation in production as an asset check we'd always be feeding these extra time series in, checking that the MSE is below some threshold, and then dropping all of the _SIMULATED records before we get to an asset that we actually publish? So in effect we'd be overweighting the importance of whatever time series we've duplicated to test with?

Yeah exactly, we're adding these extra timeseries' before doing imputation, which then get used during the imputation, so the timeseries' we choose for the simulation are essentially "double counted". We then drop these when we create the output assets, so we could theoretically do simulation in production without the simulated data actually showing up in published data,. From what I've seen so far doing the simulation has a minimal impact on the results for the rest of the non-simulated data, but I'll try and get some plots and analysis to demonstrate this.

> What determines which time series end up getting duplicated for testing? Is it just the random sampling we do of heavily imputed / unimputed months?

Yeah this is correct, it will find heavily imputed months (or more accurately, months where the imputation rate is between configurable bounds), and months with no imputed values, and then randomly samples `num_months` months from these sets.

> Are the whole multi-year timeseries being duplicated? or just the year that the simulated month appears in? Or just the individual months? The blocks that we impute are whole years, right?

Right now it's duplicating the whole multi-year timeseries`, although I think it would be pretty straightforward to switch this to duplicating years, because we are imputing in year-long blocks. I think duplicating just the simulated months would be hard because we need year long blocks to line up with the rest of the matrix.

> When we're producing the simulated time series, are we shifting the data in time? Or just shifting the pattern of nulls in time? If we're shifting the data, then it seems like we could me messing up the real correlations between the different time series.

We're just shifting the pattern of Nulls, not any actual data.

zschira (commented):
## Overview of Changes
I've modified the overall simulation flow, so now all simulation related stuff is done in a set of parallel assets that have no impact on the main imputation. The results now make a lot more sense, and don't seem to be so obviously 'overfit'. I'll summarize below what I've seen so far.

I also switched to using `mean_absolute_percentage_error` as the comparison metric, given that there are substantial magnitude differences between BA's and subregions, so percent error is more informative.

## Summary of results
With this new setup, the results seem much more meaningful. Given that there's some random sampling, the simulation is not deterministic, so I've ran it several times for the combined impuation, BA imputation, and subregion imputation and have found pretty consistent results, but I think it would still would be worthwhile to do more test runs before choosing a final setup.

Results:

- Combined imputation: between 1-2% error
- BA imputation: between 2-3%
- Subregion imputation: consistently ~1% error

This results so far are slightly counter to what we'd expected, given that isolated subregion impuation is actually outperforming the combined and isolated BA imputation. Not sure what exactly to make of it. One take away could be that all setups are performing pretty well, so maybe we should just do the combined imputation to avoid having to do extra cpu-intensive imputation stuff.

I've also included a notebook called `notebooks/work-in-progress/imputation_comparison.ipynb` that plots some of the simulation months so we can see what's going on. Given that which months/entities are chosen is non-deterministic, the notebook will dynamically plot the 3 simulated months with the highest error. These plots show more realistic results than the previous setup.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1561,162 @@ def filter_missing_values(
...
+        how="inner",
+    )
+
+    # Filter to hours where values where flagged (NULL) in reference month
+    flagged = simulation_df[simulation_df["value_col"].isnull()]
+
+    # Transform hours in reference month to hours in simulation month
+    flagged["simulation_datetime"] = (
+        flagged["datetime"] - flagged["reference_month"] + flagged["simulation_month"]
+    )
```
zaneselvans, line 1625: It's not intuitive to me what it means to add and subtract datetimes like this.

Both `simulation_month` and `reference_month` are month-starting datetimes, right?

It seems like `simulation_month` - `reference_month` is meant to be a timedelta -- a shift (an amount of time, rather than a point in time) that can be applied to some other datetime. Is that right?

Why isn't it ever a problem that the time delta is always based on month-start datetimes, when months have different numbers of hours in them, and the time delta is being used to modify hourly timestamps?

src/pudl/output/eia930.py
```diff
@@ -74,14 +75,54 @@ def _years_from_context(context) -> list[int]:
...
     value_col="demand_reported_mwh",
     imputed_value_col="demand_imputed_pudl_mwh",
     id_col="combined_subregion_ba_id",
-    settings=ImputeTimeseriesSettings(method_overrides={2019: "tnn", 2025: "tnn"}),
+    real_id_cols=[
+        "balancing_authority_code_eia",
+        "balancing_authority_subregion_code_eia",
+    ],
+    settings=ImputeTimeseriesSettings(
+        method_overrides={2019: "tnn", 2025: "tnn"},
```
zaneselvans, line 123: We are likely to forget to update these hard-coded years. Is there some way to instead apply `tnn` to the first and last years, whatever they might be?

zaneselvans (changes requested):
I'm still feeling fuzzy on how this is working, I think partly because of the nomenclature (e.g. "flagged" and "imputed" rather than just "nulled").

I keep looking for the place where you set the `value_col` to `NA` based on an observed pattern of null values and not finding it.  I think because you're entirely manipulating the indices and leaving the data "in place?"

Given that we're not duplicating the whole time series now, I'm wondering if it makes sense to only include the years with simulated null values in them in the imputation as I think we're doing now. Doesn't that mean we're potentially dropping a bunch of time series data that would otherwise have been fed into the imputation (and which is fed into the production imputation)?

zschira:
> In the "Combined imputation" case, what is the MAPE for just the imputed BA time series, and what is the MAPE for just the imputed Subregion time series? If the imputed time series in the other cases are only one or the other type, it seems like we need to break them out separately in order to compare their performance. Otherwise t seems like the 1-2% error could just be the average of the 1% error and 2-3% errors, couldn't it?

@zaneselvans this is a good question, I was computing the MAPE for a combined set of BA/subregion data. I've updated this now, so when we do the combined imputation, it will generate simulation data for BAs/subregions separately and score them separately. The results I've found seem to confirm your suspicion here:

- MAPE for BA's in combined imputation: 2-3%
- MAPE for subregion's in combined imputation: ~1%
- MAPE for standalone BA imputation: 2-3%
- MAPE for standalone subregion imputation: ~1%

> I keep looking for the place where you set the value_col to NA based on an observed pattern of null values and not finding it. I think because you're entirely manipulating the indices and leaving the data "in place?"

This happens in `simulate_flags`, but we never actually modifying the `value_col`, because the operation is performed on the pivoted timeseries matrix. This is the one that is indexed by `datetime`, with one column per BA/subregion, so at this point we don't have a `value_col`.

> Given that we're not duplicating the whole time series now, I'm wondering if it makes sense to only include the years with simulated null values in them in the imputation as I think we're doing now. Doesn't that mean we're potentially dropping a bunch of time series data that would otherwise have been fed into the imputation (and which is fed into the production imputation)?

When we actually do imputation we're only ever looking at one year at a time, so as long as the rest of the year is present, then it shouldn't impact the results of the imputation.

zaneselvans:
> * MAPE for BA's in combined imputation: 2-3%
> * MAPE for subregions in combined imputation: ~1%
> * MAPE for standalone BA imputation: 2-3%
> * MAPE for standalone subregion imputation: ~1%

So this means that the quality of the imputations aren't particularly affected by whether they have all of the time series or just a restricted subset available?  In which case we can just do the simpler / less computationally intensive thing of imputing the BAs and subregions independently?

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1561,162 @@ def filter_missing_values(
...
+        ] = None
+        flag_matrix.loc[
+            simulation_data[simulation_data["simulation_id_col"] == simulation_id][
+                "simulation_datetime"
+            ],
+            simulation_id,
+        ] = "simulated"
+    return (
+        timeseries_matrix[timeseries_matrix.index.year.isin(simulated_years)],
+        flag_matrix[flag_matrix.index.year.isin(simulated_years)],
```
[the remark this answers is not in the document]
zaneselvans: @zschira this is the bit where I'm worried that if we're only keeping a subset of the years of data, we're losing information from those time series that can be used to inform the imputation in other years of data. But maybe I'm not understanding how the selection of a subset of years flows through to other parts of the process?

zaneselvans:
>> Given that we're not duplicating the whole time series now, I'm wondering if it makes sense to only include the years with simulated null values in them in the imputation as I think we're doing now. Doesn't that mean we're potentially dropping a bunch of time series data that would otherwise have been fed into the imputation (and which is fed into the production imputation)?

> When we actually do imputation we're only ever looking at one year at a time, so as long as the rest of the year is present, then it shouldn't impact the results of the imputation.

Left [a comment](https://github.com/catalyst-cooperative/pudl/pull/4247/files#r2098670885) on the section I was concerned about here.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1561,162 @@ def filter_missing_values(
...
+    # These new columns mimic real timeseries' with simulated months that have
+    # flagged values which weren't flagged in the original data
+    for simulation_id in simulation_data.simulation_id_col.unique():
+        # Apply simulated flags to each matrix
+        timeseries_matrix.loc[
+            simulation_data[simulation_data["simulation_id_col"] == simulation_id][
+                "simulation_datetime"
+            ],
+            simulation_id,
+        ] = None
```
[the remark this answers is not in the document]
zschira: `pd.NA` makes more sense. Fixed.

src/pudl/analysis/timeseries_cleaning.py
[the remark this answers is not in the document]
zschira: If you look at `impute_flagged_values`, it actually filters down to a single year at a time before feeding data into the actual imputation, so when we're actually doing the imputation these other years aren't available for imputation.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1561,162 @@ def filter_missing_values(
...
+    flag_matrix: DataFrame[TimeseriesMatrix],
+) -> DataFrame[TimeseriesMatrix]:
+    """Simulate flagged values for scoring imputation.
+
+    Find months of data with high rate of flagged values, and uses these sections
+    as a reference to flag values in otherwise good sections of data. This allows us
+    to impute data in a realistic scenario where we have good reported data, which
+    we can compare to in order to compute quantitative metrics to validate the
+    quality of our imputation.
+    """
```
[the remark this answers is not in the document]
zschira: Added

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1561,162 @@ def filter_missing_values(
...
+
+    # Get all years with simulated data to filter timeseries
+    simulated_years = good_months["simulation_month"].dt.year.unique()
+
+    # Use reference month to get hours which should be flagged in simulation month
+    simulation_data = _get_simulation_data(
+        df,
+        pd.concat(
+            [bad_months.reset_index(), good_months.reset_index()], axis="columns"
+        ),
```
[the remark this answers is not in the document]
zschira: This is intentional, it's just taking two sets of months and matching one "good" month to one "bad" month, but which specifically get matched is not important.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1561,162 @@ def filter_missing_values(
...
+    """
+
+    reference_id_col: Series[Any]
+    reference_month: Series[pa.dtypes.DateTime]
+    simulation_id_col: Series[Any]
+    simulation_month: Series[pa.dtypes.DateTime]
+
+
+@pa.check_types
+def _get_simulation_data(
```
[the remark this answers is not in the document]
zschira: Changed the name since what this is actually "getting" is datetimes that we want to NULL for simulation purposes.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -162,15 +164,16 @@ def utc_dataframe_to_aligned(
...
     """Pivot aligned timeseries dataframe into timeseries matrix and pad if needed.
 
     Padding finds the complete list of hours from the start of the first day
     present in the timeseries to the end of the last, and then fills any missing hours
     with NULLs.
     """
-    matrix = aligned_df.pivot(index="datetime", columns="id_col", values="value_col")
+    matrix = aligned_df.pivot(index="datetime", columns="id_col", values=values_col)
 
     # Pad matrix with any missing hours from timeseries
```
zaneselvans, line 178: If I understand this correctly, it's potentially changing the number of rows in the dataframe -- if it does end up filling in hours at the start or end, or gaps in the middle.  So what happens with the reindex at the end?  Is that what causes NULL values to be added, in association with any new index value that wasn't present in the original?  If so, a comment explaning that would be helpful

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -162,15 +164,16 @@ def utc_dataframe_to_aligned(
 @pa.check_types
 def pivot_aligned_timeseries_dataframe(
     aligned_df: DataFrame[AlignedTimeseriesDataFrame],
-    periods: int,
+    periods: int = 24,
```
zaneselvans, line 167: Periods isn't being used in the function anywhere, and the padding seems to have hard-coded 24, so this parameter seems like it should be removed?  Or if other numbers of periods are possible, then the padding should depend on the `periods` parameter.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1569,200 @@ def filter_missing_values(
...
+) -> pd.DataFrame:
+    """Helper function to melt imputed timeseries matrix and merge back on input asset."""
+    # Melt imputed matrix/flag matrix to long format
+    imputed_df = melt_imputed_timeseries_matrix(matrix, flags)
+
+    # Merge back on core table
+    return aligned_df.merge(
+        imputed_df.rename(columns={"value_col": "imputed_value_col"}),
+        on=["id_col", "datetime"],
+        how="left",
```
zaneselvans, line 1617: Please add a merge validation for readability and error checking.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1569,200 @@ def filter_missing_values(
...
+    # In the ``datetime`` column.
+    simulation_df["period"] = simulation_df["reference_month"].dt.to_period("M")
+    imputed_df["period"] = imputed_df["datetime"].dt.to_period("M")
+
+    # Merge on month and ID to get a DataFrame with all hours in the reference months
+    simulation_df = imputed_df.merge(
+        simulation_df,
+        left_on=["id_col", "period"],
+        right_on=["reference_id_col", "period"],
+        how="inner",
```
zaneselvans, line 1654: Please add a merge validation for legibility and error checking.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1613,12 +1819,32 @@ def impute_timeseries_asset_factory(
...
+            If specified, these columns will be used instead during scoring.
+        settings: Configurable options for imputation (see :class:`ImputeTimeseriesSettings`).
     """
-    timeseries_matrix_asset = f"_{output_asset_name}_timeseries_matrix"
-    aligned_input_asset = f"_{input_asset_name}_aligned"
-    cleaned_timeseries_matrix_asset = f"_{output_asset_name}_cleaned_timeseries_matrix"
-    flags_asset = f"_{output_asset_name}_timeseries_matrix_flags"
-    imputed_asset = f"_{output_asset_name}_imputed_matrix"
+    timeseries_matrix_asset = re.sub(
+        r"^__", "_", f"_{output_asset_name}_timeseries_matrix"
```
zaneselvans, line 1836: Can you add a comment explaining what's up with all the regex substititions here? I think I know, but it's confusing at first.

src/pudl/output/eia930.py
```diff
@@ -74,14 +75,59 @@ def _years_from_context(context) -> list[int]:
...
+imputed_combined_demand_assets = impute_timeseries_asset_factory(
+    input_asset_name="_out_eia930__combined_demand",
+    output_asset_name="_out_eia930__combined_imputed_demand",
+    years_from_context=_years_from_context,
+    value_col="demand_reported_mwh",
+    imputed_value_col="demand_imputed_pudl_mwh",
+    id_col="generic_id",
+    simulation_group_col="simulation_group",
+    settings=ImputeTimeseriesSettings(
+        simulate_flags_settings=SimulateFlagsSettings(),
```
zaneselvans, line 114: Ah, I guess it looks like we don't have the issue of the starting/ending years needing to use the `tnn` method in the combined (and BA) cases, so maybe my comment below is irrelevant -- if we're going to ditch the separately imputed assets, which seems like the right thing to do.

zaneselvans: Or rather, the assets will still exist, but they'll just be a simple splitting out of this combined imputed demand asset into BA and subregion outputs.

zaneselvans (changes requested):
- I tweaked some docstrings and comments to try and make them clearer (to me) -- let me know if I screwed up the meanings!
- I made a few more minor doc/comment requests.
- One minor code change request -- please add `how` and `validate` arguments to the many merges that you're doing -- they help the reader understand what's going on and can also catch a lot of unexpected merge behavior, which we've been caught by so many times in the past.

src/pudl/metadata/dfs.py
```diff
@@ -24,12 +24,14 @@ class ImputationReasonCodes(Enum):
...
         "Indicates value is significantly different from nearest unflagged value."
     )
+    SIMULATED = "Used for scoring imputation using simulated data. SHOULD NOT APPEAR IN PRODUCTION DATA."
 
 
 IMPUTATION_REASON_CODES = pd.DataFrame(
     [
         {"code": code.name.lower(), "description": code.value}
         for code in ImputationReasonCodes
+        if code != ImputationReasonCodes.SIMULATED
```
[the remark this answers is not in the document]
zschira: I was going to add an override for the simulated table, but I think `dbt` tests is a good idea! Added now.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1641,9 +1867,19 @@ def _prepare_timeseries_matrix(
         """
         # Convert from datetime_utc to local datetime
         aligned_df = utc_dataframe_to_aligned(
-            input_df.rename(columns={value_col: "value_col", id_col: "id_col"})
+            input_df.rename(
+                columns={
+                    value_col: "value_col",
+                    id_col: "id_col",
+                    simulation_group_col: "simulation_group",
```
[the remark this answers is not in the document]
zschira: From my testing pandas just doesn't do anything if `simulation_group_col` is `None`. I think it tries to rename a column called `None`, but it doesn't exist so nothing happens.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1664,66 +1900,182 @@ def _flag_timeseries_matrix(
...
+
+        This takes the real output asset and the simulated output asset, and will
+        compute a metric comparing the imputed simulated data to the real data. The
+        metric used is ``mean_absolute_percentage_error`` as percent error is more robust
+        to magnitude changes in the underlying data than total error.
+        """
+        mape_dict = {}
+        for group_name, gdf in simulated_df.groupby("simulation_group"):
+            # Get just rows where we simultated NULLS
+            simulated_gdf = gdf[gdf["flags"] == "simulated"]
```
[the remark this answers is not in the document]
zschira: At this point the flag column just contains strings, so a direct comparison to the `enum` would fail.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1664,66 +1900,182 @@ def _flag_timeseries_matrix(
...
+            simulated_flags_asset: AssetOut(key=simulated_flags_asset),
+        },
+        name=f"_{output_asset_name}_simulate_flag_timeseries_matrix",
+    )
+    def _simulate_flags(
+        imputed_df: pd.DataFrame,
+        matrix: pd.DataFrame,
+        flags: pd.DataFrame,
+    ) -> tuple[pd.DataFrame, pd.DataFrame]:
+        """Simulate flag values to impute for scoring imputation."""
```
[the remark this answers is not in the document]
zschira: Added some more description. I also added more to the top-level docstring of the factory function.

src/pudl/analysis/timeseries_cleaning.py
zschira: Yeah the `reindex` is what actually adds the NULL's. I've added a comment explaining.

src/pudl/analysis/timeseries_cleaning.py
zschira: I've removed `periods` from here. There's a couple other places `periods` gets referenced, so I made it a configurable parameter, but it's not something I've ever messed with, so it might not be necessary at all.

src/pudl/analysis/timeseries_cleaning.py
zschira: Done.

src/pudl/analysis/timeseries_cleaning.py
zschira: Done.

src/pudl/analysis/timeseries_cleaning.py
zschira: Added, it's basically just making sure these assets only have one underscore

src/pudl/output/eia930.py
zschira: Yeah I'm not sure that the `tnn` method had anything to do with the first/last year and that may have just been a coincidence. I'm going to test using the `tnn` method on everything though, which would ultimately be simpler.

dbt/models/eia930/out_eia930__hourly_operations/schema.yml
```diff
@@ -20,5 +20,8 @@ sources:
           - name: demand_adjusted_mwh
           - name: demand_imputed_pudl_mwh
           - name: demand_imputed_pudl_mwh_imputation_code
+            tests:
```
zschira, line 23: Added `dbt` tests for "simulated" flags.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1579,17 +1580,21 @@ class SimulateFlagsSettings:
     """Min ratio of bad points in a section of data to be used for reference."""
     max_flag_rate: float = 0.5
     """Max ratio of bad points in a section of data to be used for reference."""
+    output_io_manager_key: str = "io_manager"
```
zschira, line 1583: Allow the io-manager for the simulated asset to be configurable so we can output for visualization purposes.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1649,10 +1656,11 @@ def _add_simulated_flag_col(
         left_on=["id_col", "period"],
         right_on=["reference_id_col", "period"],
         how="inner",
+        validate="many_to_one",
     )
 
     # Filter to hours where values were flagged (NULL) in reference month
-    flagged = simulation_df[simulation_df["value_col"].isnull()]
+    flagged = simulation_df[simulation_df["flags"].notnull()]
```
zschira, line 1663: Switched to checking for flags directly, but doesn't actually change the results.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -2047,35 +2098,47 @@ def _score_imputation(
...
+            combined_df = simulated_gdf.merge(
+                imputed_df,
+                how="inner",
+                on=["datetime_utc", "id_col"],
+                validate="one_to_one",
+            )
 
             # Compute metric
             mape_dict[group_name] = mean_absolute_percentage_error(
-                combined_df["value_col_x"],
```
zschira: Was accidentally comparing reported values for a second which produces 0 MAPE, but fixed that now.

zschira (commented):
@zaneselvans there were a bunch of little things I had to fix while getting the final asset structure working, but this should all be ready for a final review now.

src/pudl/output/eia930.py
```diff
@@ -74,25 +75,93 @@ def _years_from_context(context) -> list[int]:
...
+        on=[
+            "datetime_utc",
+            "balancing_authority_code_eia",
+            "balancing_authority_subregion_code_eia",
+        ],
+        validate="one_to_one",
+        how="left",
+    )
+
+    return subregion_demand, ba_demand
```
zaneselvans, line 164: These should be returned as a tuple of explicit `dagster.Output` objects, to explicitly link the value with the asset name, so we aren't depending on the brittle ordering of values to keep them straight.  And now I see that there are a bunch of other multi-assets in the timeseries cleaning module that also need this.

zaneselvans (changes requested):
- **Blocking:** I'm getting a couple of failed asset materializations. I left comments where the errors are happening.
- **Blocking:** We need to enforce some maximum allowable MAPE don't we? What's reasonable? 5%?  Have you seen what the MAPE values are in the fast ETL?
- **Blocking:** ~All multi-asset return values should be tuples of `dagster.Output` objects that explicitly associate the asset name with the returned value to avoid reliance on the brittle ordering of values and outs listed in the decorator.~ (I went ahead and did this)
- **Question:** I noticed that in the EIA-930 sub-DAG both the simulation and production assets depend on `_out_eia930__combined_imputed_demand_imputed` which is fairly long running, and this means that the two imputation processes can't run in parallel. Why is this? Is it that there's no way to access the flagged values other than through the outputs of the imputation?

src/pudl/output/eia930.py
zaneselvans: I went ahead and added all these explicit `Output` constructs since it was easy.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1664,66 +1921,224 @@ def _flag_timeseries_matrix(
...
+            # Compute metric
+            mape_dict[group_name] = mean_absolute_percentage_error(
+                combined_df["imputed_value_col_x"],
+                combined_df["imputed_value_col_y"],
+            )
+            logger.info(
+                f"Imputed simulated NULLS for group {group_name} with mean percent error: {mape_dict[group_name]}"
+            )
+
+        return mape_dict
```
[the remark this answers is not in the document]
zschira: Ah yeah, I've added in a basic `asset_check` now and a configurable threshold to the settings.

src/pudl/output/eia930.py
zschira: Thanks, I also removed the type annotations as that was causing a mismatch

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1560,6 +1570,205 @@ def filter_missing_values(
...
+    flagged = flagged[
+        flagged["simulation_datetime"].dt.to_period("M")
+        == flagged["simulation_month"].dt.to_period("M")
+    ]
+
+    # Apend column with simulated flags to imputed dataframe
+    imputed_df["simulated_flags"] = False
+    imputed_df = imputed_df.set_index(["datetime", "id_col"])
+    imputed_df.loc[
+        (flagged["simulation_datetime"], flagged["simulation_id_col"]),
```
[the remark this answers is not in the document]
zschira: Hmm, I haven't seen this issue, I wonder if it could have been an upstream asset that was out of sync or something?

I also plan to run a series of test runs to try to catch any potential transient errors with the simulation stuff before merging.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -2149,6 +2159,16 @@ def _score_imputation(
 
         return mape_dict
 
+    @asset_check(
+        asset=imputation_score_asset, name=f"{imputation_score_asset}_asset_check"
+    )
+    def _check_score(mape: float):
+        return AssetCheckResult(
+            passed=mape > settings.simulate_flags_settings.mape_threshold,
```
zaneselvans, line 2167: It seems like this check is backwards. We want the measured mape to be less than the threshold don't we?

src/pudl/output/eia930.py
zaneselvans: Ah yeah, whoops. Thanks for catching that.

src/pudl/analysis/timeseries_cleaning.py
zschira: Fixed now.

src/pudl/analysis/timeseries_cleaning.py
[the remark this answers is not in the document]
zaneselvans: Hmm. I tried to rematerialize everything upstream and downstream of all the eia930 assets. I'll try again. I got this same error both times I attempted.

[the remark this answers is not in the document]
zaneselvans: The asset selector I used was

```
+key:"**eia930**"+
```

zaneselvans:
Another thing I just thought of -- with the random selection of years that end up being involved in the imputations, do the simulation tables have consistent expected row counts?  If not, then whenever anybody does `dbt_helper udpate-tables --row-counts all` whatever row counts happen to be present will get codified as the expectation, even if it's not stable.

zaneselvans:
I've kicked of a CI run here to see if the fast ETL also works.

zschira:
> Another thing I just thought of -- with the random selection of years that end up being involved in the imputations, do the simulation tables have consistent expected row counts? If not, then whenever anybody does dbt_helper udpate-tables --row-counts all whatever row counts happen to be present will get codified as the expectation, even if it's not stable.

That's a good point, I removed the row count test from the simulated table, but we definitely don't want that getting overwritten. It might make sense to add a set of tables to the `dbt_helper` script that it will skip for all row-count related stuff for.

zaneselvans:
Ah! Okay, if the row count check has been removed from the table then at least it's in a good state for now. Ideally the update all would not *add* that test to a table that has been explicitly told not to have it.

zschira:
> Oh it would be good to add something to the release notes about the inclusion of these diagnostic imputation tables.

Added

> If you have time before you leave, a separate PR adding a new documentation page that talks about the imputation process as applied to both EIA-930 and FERC-714 and linking to it from the data source documentation pages for EIA-930 and FERC-714 would be good, while the information is still fresh in your head.

Yeah I'll make sure to get a PR ready before I leave.

> Also, did you happen to try validating the imputation on the FERC-714 data? If so, how was its MAPE?

I did, and the MAPE came out to ~3.5%. Might be nice to enable the simulation for 714 at some point to keep track during builds as well, but I feel like we should get this working in builds first.

zaneselvans:
Seems like we might need to mess around with the tests to reduce memory usage.

zschira:
> Seems like we might need to mess around with the tests to reduce memory usage.

Yeah I'm a bit confused why this change is causing it to break. Seems like the ETL is finishing fine, and I haven't added any new tests. Maybe there's something timing related that's breaking?

zaneselvans:
IIRC, with the in-process executor (which is what the integration tests use) Dagster stores any asset that's using the default IO manager in memory, rather than writing it out to disk as a pickled dataframe, and we've definitely added some big new assets using the default IO manager with these changes, so I suspect that's what tipping us over the memory available in the runner.

zaneselvans:
Hmm, in the most recent run it didn't get killed, it had a failure in the MAPE calculation due to a lack of values.  I had no trouble running it locally in the fast ETL though.

```
Stack Trace:
  File "/home/runner/micromamba/envs/pudl-dev/lib/python3.12/site-packages/dagster/_core/execution/plan/utils.py", line 56, in op_execution_error_boundary
    yield
  File "/home/runner/micromamba/envs/pudl-dev/lib/python3.12/site-packages/dagster/_utils/__init__.py", line 392, in iterate_with_context
    next_output = next(iterator)
                  ^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/pudl-dev/lib/python3.12/site-packages/dagster/_core/execution/plan/compute_generator.py", line 129, in _coerce_op_compute_fn_to_iterator
    result = invoke_compute_fn(
             ^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/pudl-dev/lib/python3.12/site-packages/dagster/_core/execution/plan/compute_generator.py", line 117, in invoke_compute_fn
                                                                    ^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/pudl/pudl/src/pudl/analysis/timeseries_cleaning.py", line 2158, in _score_imputation
    mape_dict[group_name] = mean_absolute_percentage_error(
  File "/home/runner/micromamba/envs/pudl-dev/lib/python3.12/site-packages/sklearn/utils/_param_validation.py", line 216, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/pudl-dev/lib/python3.12/site-packages/sklearn/metrics/_regression.py", line 473, in mean_absolute_percentage_error
    _check_reg_targets_with_floating_dtype(
  File "/home/runner/micromamba/envs/pudl-dev/lib/python3.12/site-packages/sklearn/metrics/_regression.py", line 198, in _check_reg_targets_with_floating_dtype
    y_type, y_true, y_pred, multioutput = _check_reg_targets(
                                          ^^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/pudl-dev/lib/python3.12/site-packages/sklearn/metrics/_regression.py", line 105, in _check_reg_targets
    y_true = check_array(y_true, ensure_2d=False, dtype=dtype)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/pudl-dev/lib/python3.12/site-packages/sklearn/utils/validation.py", line 1130, in check_array
    raise ValueError(

2025-05-24 04:20:45 +0000 - dagster - ERROR - etl_fast - cc991a98-6ac8-4c14-812c-31cecaab948f - 2936 - RUN_FAILURE - Execution of run for "etl_fast" failed. An exception was thrown during execution.

ValueError: Found array with 0 sample(s) (shape=(0,)) while a minimum of 1 is required.
```

zaneselvans:
Welp, maybe we strip out that test as well.

zschira:
@zaneselvans yeah seems like it. I'm happy to approve that PR, but you said it needs the `migrate_no_null_rows` changes merged in as well?

Definitely seems like we need to invest some resources in improving CI right now. It would make development so much smoother if we could get the total runtime/memory usage down somewhat substantially.

zaneselvans:
@zschira let me get @cmgosnell to review the other PR that has just the no null cols changes -- there's something a little weird in there about the MCOE tests that I think she is familiar with that I wanted her eyeballs on. I will remove those changes from my branch that is trying to merge into this PR, and we can merge them down from `main`.

zaneselvans:
OMFG fiiiiiiiinally. Now onto breaking the nightly builds 🤣

PR #4153, opened 2025-03-27
fix: rowcounts check now fails on missing/extra partitions

jdangerx:
# Overview

## What problem does this address?

I was making sure my asset checks (#4132, #4154) did the right thing for ETLs both fast and full, and found that the row-counts checks don't fail if you're completely missing a year.

i.e., my fast results had correct data for 2020 and 2023, but no data for the other years. If I ran it expecting `etl-full`, it *should* have failed due to missing all those years but instead succeeded.

## What did you change?

Switch to a FULL OUTER join instead of INNER. Add null checks.

## Documentation

Doesn't seem like this needs to change. Brings it more in line with expectations.

# Testing

How did you make sure this worked? How can a reviewer verify this?

Kept updating the expected rows CSVs:
* missing a year in expected CSV: should fail with one mismatch
* adding a year in expected CSV: should fail with one mismatch
* changing a row count in expected CSV: should fail with one mismatch

(`dbt build --select pudl_dbt.pudl.core_eia860__assn_boiler_cooling` was the model I was using...)

Tried to write a unit test to cover all these conditions, but ran into issues not being able to call a `test` like a `macro`.

I think we'd have to set up some sort of test environment for DBT where we have models whose only purpose is to call data tests, then populate that environment via `pytest`, and then also remember to strip everything down when the test is over.

Alternatively we could try to render the test SQLs via Jinja, if we can get ahold of the DBT Jinja context, and then run the SQL directly in `pytest`.

Both options seemed like rabbit holes so I'm punting for now.

## To-do list

- [ ] If updating analyses or data processing functions: make sure to update row count expectations in `dbt` tests.
- [ ] Run `make pytest-coverage` locally to ensure that the merge queue will accept your PR.
- [ ] Review the PR yourself and call out any questions or issues you have.
- [ ] For minor ETL changes or data additions, once `make pytest-coverage` passes, make sure you have a fresh full PUDL DB downloaded locally, materialize new/changed assets and all their downstream assets and [run relevant data validation tests](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `pytest` and `--live-dbs`.
- [ ] For bigger ETL or data changes run the full ETL locally and then [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `make pytest-validate`.
- [ ] Alternatively, run the `build-deploy-pudl` GitHub Action manually.

PR #4249, opened 2025-05-08
Summarize row count diffs

zschira:
# Overview

This PR adds a dbt helper command to summarize row count diffs after running dbt tests.

### DIFF
--- a/src/pudl/scripts/dbt_helper.py
+++ b/src/pudl/scripts/dbt_helper.py
@@ -558,7 +558,7 @@ def migrate_tests(table_name: str, test_config_name: str, model_name: str | None
     is_flag=True,
     help="Report exact year by year row count changes, otherwise report total percent change.",
 )
-def summarize_row_count_failures(verbose: bool = False):
+def summarize_row_count_diffs(verbose: bool = False):
     """Load all row count failures from duckdb file and summarize.
 
     After running dbt tests, this command can be used to summarize row count changes.
@@ -567,7 +567,7 @@ def summarize_row_count_failures(verbose: bool = False):
     percent change in row counts.
 
     Example usage:
-        dbt_helper summarize-row-count-failures
+        dbt_helper summarize-row-count-diffs
     """
     # Load failures
     with Path("./dbt/target/run_results.json").open() as f:
@@ -592,11 +592,15 @@ def summarize_row_count_failures(verbose: bool = False):
     # Loop through failures and output results
     extract_table_name_pattern = r"table_name = '(\w+)'"
     for failure in failures:
+        # Extract table name from json
         table_name = re.search(
             extract_table_name_pattern, failure["compiled_code"]
         ).group(1)
+
+        # Get row count diffs from duckdb
         row_counts_df = db.sql(f"SELECT * FROM {failure['relation_name']}").df()  # noqa: S608
 
+        # Write full or summarized results
         if verbose:
             logger.warning(
                 f"Row count failures found for table: {table_name}\n{row_counts_df}"
@@ -651,7 +655,7 @@ def dbt_helper():
 
 dbt_helper.add_command(add_tables)
 dbt_helper.add_command(migrate_tests)
-dbt_helper.add_command(summarize_row_count_failures)
+dbt_helper.add_command(summarize_row_count_diffs)
 
 
 if __name__ == "__main__":
--- a/src/pudl/scripts/dbt_helper.py
+++ b/src/pudl/scripts/dbt_helper.py
@@ -1,5 +1,6 @@
 """A basic CLI to autogenerate dbt data test configurations."""
 
+import json
 import re
 from collections import defaultdict, namedtuple
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


aesharpe (changes requested):
I like the concept here, but I got an error when I tried to test it. I made sure there were row count discrepancies between by df and the row count csvs--ran `dbt build --select test_name:check_row_counts_per_partition --target=etl-full` to confirm--but when when I ran `python src/pudl/scripts/dbt_helper.py summarize-row-count-diffs` I got the following error: 

```
  File "/Users/austensharpe/Desktop/Repos/pudl/src/pudl/scripts/dbt_helper.py", line 680, in <module>
    dbt_helper()
  File "/Users/austensharpe/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1161, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austensharpe/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/austensharpe/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/austensharpe/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1443, in invoke
  File "/Users/austensharpe/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 788, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austensharpe/Desktop/Repos/pudl/src/pudl/scripts/dbt_helper.py", line 618, in summarize_row_count_diffs
    row_counts_df = db.sql(f"SELECT * FROM {failure['relation_name']}").df()  # noqa: S608
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
duckdb.duckdb.CatalogException: Catalog Error: Table with name None does not exist!
Did you mean "pg_index"?
```

src/pudl/scripts/dbt_helper.py
```diff
@@ -550,6 +556,88 @@ def migrate_tests(table_name: str, test_config_name: str, model_name: str | None
...
+def summarize_row_count_diffs(table: str, verbose: bool = False):
+    """Load all row count failures from duckdb file and summarize.
+
+    After running dbt tests, this command can be used to summarize row count changes.
+    It uses a json file dbt saves at 'dbt/target/run_results.json'. The verbose option
+    will output the exact changes per partition, while the default will output the total
+    percent change in row counts.
+
+    Example usage:
+        dbt_helper summarize-row-count-diffs
```
aesharpe, line 582: This is much better than copying and pasting the SQL queries to see how the data have changed! But I think it would be even more helpful to be able to see all the data at once--row count percent change _and_ the actual row numbers. Sometimes the percent change doesn't tell the whole story.

Dazhong's PR #4153 updates the row count SQL queries to show all these fields: 

```
    '{{ table_name }}' as table_name,
    expected.partition as expected_partition,
    observed.partition as observed_partition,
    expected_count,
    observed_count,
    (observed_count - expected_count) / expected_count as diff_relative_to_expected,
```
But it's annoying because you have to run each SQL query individually to see them. If this `dbt_helper` function could basically compile these all together into one table and just print that out it would be amazing. 

Other small tweaks could be:
- Rounding the percentage values to a few sig figs

zschira: I've tested this with Dazhong's changes merged in, and it will now print all of the fields in the updated test, so if you use the `--verbose` option you'll get the percent/total diff for each partition. Without `--verbose` it will still just prints the percent change across all partitions.

zschira:
@aesharpe I haven't been able to reproduce the error you encountered, but I have tested with all changes from `main` merged into this branch and it is running for me without issue.

dbt/README.md
```diff
@@ -62,6 +62,23 @@ To see all options for this command run:
...
+#### `summarize-row-count-diffs`
+When integrating new data, we often have to update row counts for many tables at once.
+In these cases, `dbt` does not provide any easy way to quickly see all failures from
+the `check_row_counts_per_partition` test, so we've added this command to grab all
+the failures from the most recent run of `dbt` tests, and summarize the results.
+
+The basic usage of this command is:
+
+```bash
+dbt_helper summarize-row-count-diffs
```
aesharpe, line 74: I would add the `--verbose` flag because I can't really imagine using this helper function without it.

dbt/README.md
```diff
@@ -167,16 +184,16 @@ dbt build --select source:pudl.{table_name} --store-failures --target={etl-fast|
...
 The output of this command should show you a `sql` query you can use to see partitions
-where the row count test failed. To see these, you can do:
+where the row count test failed. While you can copy and paste these queries into
+`duckdb`, this can be quite tedious if many tables have changed, so we've also
+provided a command to [summarize all failures](#summarize-row-count-diffs). Basic usage looks like:
 
 ```bash
-duckdb {PUDL_OUTPUT}/pudl_dbt_tests.duckdb
+dbt_helper summarize-row-count-diffs
 ```
```
aesharpe, line 193: I would add the `--verbose` flag here because that is the most useful output.

zaneselvans: If it's the most useful output and we expect it to be the main way that this subcommand is used, maybe it shouldn't require a `--verbose` flag at all.  Is there a good use case for the non-verbose output?

zaneselvans:
@zschira has this been superseded by the `dbt_helper validate` subcommand that @jdangerx added, which shows the partitions that had bad row counts and expected/actual counts and % deviation?

zaneselvans:
I'm pretty sure this has been superseded by the dbt validate script, so I'm closing. But we can reopen if I'm wrong.

PR #4251, opened 2025-05-08
Process FERC-714 data for 2024

zaneselvans:
# Overview

Integrates the 2024 data from FERC Form 714.


zaneselvans:
Unlike FERC Forms 1, 2, 6, & 60 which had a reporting deadline of April 18th this year, the [FERC-714 reporting deadline is June 1st](https://www.ferc.gov/industries-data/electric/general-information/electric-industry-forms/form-no-714-annual-electric/viewer). As a result there's only a handful of respondents in the data so far (about 10% of the total). So it doesn't make sense to do this integration until Q3.

PR #4253, opened 2025-05-09
Skip MCOE null cols check that is crashing pytest worker.

zaneselvans:
For some reason the MCOE null columns check for fuel_cost is crashing the pytest worker in the nightly builds, even though it passes locally. Skip that validation for the moment to prevent this from blocking quarterly updates / nightly builds.


cmgosnell (approved):
do ittt

Issue #4216, opened 2025-04-25
dbt "nightly" target uses local parquet files for intermediate models


krivard:
### Describe the bug

Our dbt system has three targets:

- etl_full, which looks at the local PUDL_OUTPUT parquet files and uses rowcount tables computed for the full ETL
- etl_fast, which looks at the local PUDL_OUTPUT parquet files and uses rowcount tables computed for the _fast_ ETL
- nightly, which (is supposed to) look at the parquet files on S3 (and uses full ETL counts)

We test two kinds of models in dbt:

- PUDL's parquet table outputs
- "intermediate" models, encoded as views that query one or more of PUDL's parquet table outputs

The `nightly` target correctly queries the parquet files on S3 for tests of PUDL table outputs.

**When testing intermediate models, the `nightly` target queries local parquet files**.

This appears to happen due to several intersecting factors:

- the intermediate models are stored as unmaterialized views in pudl_dbt_tests.duckdb
- the parquet paths used in dbt's unmaterialized views use absolute paths to the parquet files they query
- the nightly copy of pudl_dbt_tests.duckdb thus has views pointing to files on the nightly filesystem, which obviously don't apply to the user's local filesystem
- the `nightly` target explicitly selects the user's local copy of pudl_dbt_tests.duckdb from $PUDL_OUTPUT, presumably to get around the absolute paths in the nightly copy
- thus the unmaterialized views for `nightly` use the absolute paths for the user's local $PUDL_OUTPUT, not the nightly parquets

### Bug Severity

- **Medium-High**: There is no way to locally check intermediate models against nightlies per se. Lukewarm workarounds:
  - I can generate full ETL results locally and run against those as a facsimile, but the degree to which they will exactly match nightlies (without wiping and rebuilding my environment) is undetermined
  - I could mirror nightlies directly into my local PUDL_OUTPUT filesystem where my local pudl_dbt_test.duckdb expects them to be, but (1) that seems mean to dagster, and (2) if there's stuff in there I want to keep, I'll have to figure out how to swap files in and out
  - I could reach into either my local or a mirrored copy of pudl_dbt_test.duckdb and edit the parquet paths used in the unmaterialized views, but that feels like major overkill

### To Reproduce
The problem is most obvious when you point PUDL_OUTPUT somewhere unexpected.

If you point PUDL_OUTPUT at somewhere that does not exist, dbt can't test nightlies at all:
```
[...]

PR #4259, opened 2025-05-12
Disable row counts for etl-fast and remove nightly dbt target

zaneselvans:
# Overview

- To ensure that we get our `v2025.5.0` data release out on time, this PR short-circuits the row count checks in our Fast ETL outputs, as there's some work required to make sure they behave consistently in local and CI environments, and that's creating friction in the development process right now.
- It also removes the `nightly` target for dbt, since that was primarily useful in the context of our initial migration of the validation tests, which is now complete.

# Testing

I ran the dbt tests locally with both `etl-full` and `etl-fast` targets, with full ETL outputs in my `PUDL_OUTPUT` directory, and the fast tests did not complain about row count violations, while the full tests did (I have some leftover FERC-714 tables that are from another PR).

## To-do list

### DIFF
--- a/dbt/README.md
+++ b/dbt/README.md
@@ -2,11 +2,10 @@
 
 This directory contains an initial setup of a `dbt` project meant to write [data
 tests](https://docs.getdbt.com/docs/build/data-tests) for PUDL data. The project is
-setup with profiles that allow you to select running tests on `nightly` builds,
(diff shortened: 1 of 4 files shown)


krivard (approved):
Glorious. This also fixes #4216 btw!

I dropped the last remaining `nightly` from the README; should be good to merge whenever

Issue #4231, opened 2025-05-04
Missing data in `out_eia__monthly_generators`


leostimpfle:
### Describe the bug
For some generators `total_mmbtu` is empty in `out_eia__monthly_generators` although `fuel_consumed_mmbtu` is non-empty `core_eia923__monthly_generation_fuel` (consistent with the raw data reported under `Total Fuel Consumed (MMBtu)` in [EIA 923](https://www.eia.gov/electricity/data/eia923/).

Strangely, `net_generation_mwh` appears to be consistently recorded.

### Bug Severity
**High:** This bug is preventing me from using PUDL.

### To Reproduce

See [this gist](https://gist.github.com/leostimpfle/c6a781fdd3a29987553b0488a5882101).

### Expected behavior
I would expect `total_mmbtu` in `out_eia__monthly_generators` to match the values reported under `Total Fuel Consumed (MMBtu)` in [EIA 923](https://www.eia.gov/electricity/data/eia923/) (consistent with the values recorded under `fuel_consumed_mmbtu` in `core_eia923__monthly_generation_fuel`).

cmgosnell:
hey @leostimpfle ! 👋🏻  Thanks for pointing this out and providing the detailed explanation!  We are focused early this week on getting new quarterly data integrated into PUDL, but we will look into this and get back to you later this week.

krivard:
Results of @aesharpe and my initial poking:

- `core_eia923__monthly_generation_fuel` is upstream of `out_eia__monthly_generators`, but plant 56329 disappears downstream of `core_eia923__monthly_generation_fuel` and is added back in later
- culprit is `core_eia860__assn_boiler_generator`, which does not contain any entries for plant 56329 -- this causes all entries for the plant to be dropped when we compute heat rate by unit

Next steps:

- Figure out where 56329 comes from when it's added back in later -- likely EIA 860 but worth being sure
- Figure out why 56329 does not appear in the boiler generator association table, and if there's anything we can do about that
- Determine how many other plants/generators are affected

krivard:
Plant 56329 is located in Hawai'i. It is not included in:

- `core_eia860__assn_boiler_generator`
- `raw_eia860__boiler_generator_assn`
- `raw_eia860__boiler_info`

It _is_ included in:

- `raw_eia860__plant`
- `raw_eia860__generator_existing`
- `raw_eia860m__generator_existing`

for reports as far back as 2005.

[...]

PR #4266, opened 2025-05-14
add min_count=1 into groupby sum for heat rate

cmgosnell:
# Overview

as @krivard  was poking at unit-based heat rate discrepencies for #4231 and i noticed a smol guy in the mcoe module which is just that when only null records are being summed by default that produces a 0 when we like that kinda thing to be a null.

## What did you change?
added `min_count=1`.. also checked for other groupby sum's and it looks like this is the only bare one in this module at least.

## Documentation

this feel so minor that I would like to not add a line in the release notes... but I'd be happy to if it seems better to include it.

# Testing

- [x] re-ran all of the `out_derived_gen_attributes` assets from the dagster ui
- [x] compared the nightly and new outputs

```python
tbl_name = "out_eia__monthly_generators"
night = pd.read_parquet(f"s3://pudl.catalyst.coop/nightly/{tbl_name}.parquet")
new = defs.load_asset_value(tbl_name)

cols = [
    "net_generation_mwh",
    "fuel_consumed_for_electricity_mmbtu",
    # "total_mmbtu"
]
for col in cols:
    print(f"""
new:\n{new[new[col].isin([np.nan, 0])][col].value_counts(dropna=False)}\n
nightly:\n{night[night[col].isin([np.nan, 0])][col].value_counts(dropna=False)}
""") 

>>
new:
net_generation_mwh
0.0    33558
NaN     4942
Name: count, dtype: int64

nightly:
net_generation_mwh
0.0    38500
Name: count, dtype: int64


new:
fuel_consumed_for_electricity_mmbtu
0.0    48856
NaN     4001
Name: count, dtype: int64

nightly:
fuel_consumed_for_electricity_mmbtu
0.0    52857
Name: count, dtype: int64
```

- [x] then re-ran `dbt build` & `make pytest-coverage` locally

## To-do list


krivard (approved):
👍 nice checks too

zaneselvans:
Any reason not to go ahead and merge this bb?

zaneselvans:
Our GitHub runners are still getting killed because of... some resource constraint. It seems to happen when the FERC DBF data is being downloaded to check for refyear schema compatibility with all past years of data.  Since we don't expect this data structure to change any more, and it's never failed before... I went ahead and disabled it when it runs on GHA -- it'll still run locally and in the nightly builds (which is the only place it really tests all the historical years)

PR #4271, opened 2025-05-15
Remove memory intensive output tests obviated by dbt row counts.

zaneselvans:
# Overview

GHA CI is failing due to what appears to be memory usage, during output tests that read a bunch of data out of SQLite into dataframes. These tests are just verifying that the soon to be removed `PudlTabl` object still works, and there's actually data in the DB... which is largely done through the row count checks now.  Trying to rip this stuff out now so we can get the CI back to working (and running faster) and get the release out this week!!!

This issue first came up last night with the PR that brought in a new quarter's worth of EIA-860/923 data. @aesharpe re-ran the CI with some small change and it passed... but apparently this is a stochastic failure and it's not going away without some direct mitigation.


test/integration/output_test.py
```diff
@@ -44,29 +43,7 @@ def test_nuclear_fraction(fast_out, df_name, expected_nuke_fraction, tolerance):
...
-)
-def test_ferc1_outputs(fast_out, df_name):
-    """Check that FERC 1 output functions work."""
-    logger.info(f"Running fast_out.{df_name}()")
-    df = fast_out.__getattribute__(df_name)()
-    logger.info(f"Found {len(df)} rows in {df_name}")
-    assert not df.empty
-
-
+@pytest.skip(reason="Memory intensive, GHA CI failing. Migrate to dbt ASAP.")
```
zaneselvans, line 46: I'm skipping rather than deleting these tests since I think we might want to continue verifying the 1:1 or 12:1 ratios between these tables, but that can and should be migrated to dbt.  Leaving them here for reference.

cmgosnell (approved):
✂️

PR #4274, opened 2025-05-16
Disable Datasette deployment in nightly builds

zaneselvans:
We haven't been able to successfully redeploy Datasette to Fly.io since March 28th (6 weeks...) which feels like a functional deprecation, even though we haven't gotten the new user portal to feature parity yet.

The nightly builds currently spend ~45 minutes attempting to redeploy Datasette whenever they run, to no effect.

This PR comments out the Datasette deployment actions in the nightly build script, so that we can stop making those pointless attempts, unless / until we actually go in and fix whatever the problem is. Or, maybe more likely, until we get the PUDL Portal to feature parity and start deploying it to https://data.catalyst.coop, usurping Datasette completely.

## Documentation

Should update our data access documentation and various other links to remove this access mode?  I'm assuming that the site will come back up post-deployment attempt as it has before (presumably reverting to the old machine at Fly.io).  If that's not correct then... maybe Datasette has fully deprecated itself for us and we need to remove it from all of our docs.

## To-do list

- [ ] Verify that https://data.catalyst.coop comes back up (or make it come back up with flyctl...)
- [ ] If it's down for good, update documentation to direct users to the new PUDL Viewer instead
- [ ] If it's down for good, update deployment location of PUDL Viewer to be https://data.catalyst.coop
- [ ] If it's down for good, set up an automatic redirect from https://viewer.catalyst.coop to https://data.catalyst.coop
- [ ] If it's down for good, remove Datasette / Fly.io specific code / scripts from PUDL repo
- [ ] If it's down for good, shut down our Fly.io account.

### DIFF
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -143,7 +143,6 @@ function notify_slack() {
     message+="SAVE_OUTPUTS_SUCCESS: $SAVE_OUTPUTS_SUCCESS\n"
     message+="UPDATE_NIGHTLY_SUCCESS: $UPDATE_NIGHTLY_SUCCESS\n"
     message+="UPDATE_STABLE_SUCCESS: $UPDATE_STABLE_SUCCESS\n"
-    message+="DATASETTE_SUCCESS: $DATASETTE_SUCCESS\n"
     message+="CLEAN_UP_OUTPUTS_SUCCESS: $CLEAN_UP_OUTPUTS_SUCCESS\n"
     message+="DISTRIBUTION_BUCKET_SUCCESS: $DISTRIBUTION_BUCKET_SUCCESS\n"
     message+="GCS_TEMPORARY_HOLD_SUCCESS: $GCS_TEMPORARY_HOLD_SUCCESS \n"
@@ -210,7 +209,6 @@ ETL_SUCCESS=0
 SAVE_OUTPUTS_SUCCESS=0
 UPDATE_NIGHTLY_SUCCESS=0
 UPDATE_STABLE_SUCCESS=0
-DATASETTE_SUCCESS=0
 WRITE_DATAPACKAGE_SUCCESS=0
 CLEAN_UP_OUTPUTS_SUCCESS=0
 DISTRIBUTION_BUCKET_SUCCESS=0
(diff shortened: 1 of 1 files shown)


docker/gcp_pudl_etl.sh
```diff
@@ -210,7 +211,7 @@ ETL_SUCCESS=0
 SAVE_OUTPUTS_SUCCESS=0
 UPDATE_NIGHTLY_SUCCESS=0
 UPDATE_STABLE_SUCCESS=0
-DATASETTE_SUCCESS=0
+# DATASETTE_SUCCESS=0
```
jdangerx, line 214: totally non-blocking: I sort of think we should just delete the datasette code instead of commenting it out. If we want to turn it back on we can revert this commit. Would be kind of cleaner to do that if this PR was *only* disabling datasette instead of disabling datasette *and* bash style changes but :shrug:

docker/gcp_pudl_etl.sh
```diff
@@ -337,16 +338,16 @@ gcloud storage --quiet cp "$LOGFILE" "$PUDL_GCS_OUTPUT"
...
-      $CLEAN_UP_OUTPUTS_SUCCESS == 0 && \
-      $DISTRIBUTION_BUCKET_SUCCESS == 0 && \
-      $GCS_TEMPORARY_HOLD_SUCCESS == 0 && \
-      $ZENODO_SUCCESS == 0
-]]; then
+if [[ $ETL_SUCCESS == 0 &&
+    $SAVE_OUTPUTS_SUCCESS == 0 &&
+    $UPDATE_NIGHTLY_SUCCESS == 0 &&
+    $UPDATE_STABLE_SUCCESS == 0 &&
+
```
jdangerx, line 345: Is this awkward empty line + the post-conditional comment so that we remember to put datasette back in? If so I think we should probably explain the intent of this somewhat strange looking block with an expanded comment in line 340.

jdangerx (approved):
Seems fine, some slight readability/leaving-notes-for-our-future-selves things but I don't need to re-review for that.

docker/gcp_pudl_etl.sh
zaneselvans: Oh. I didn't even realize my bash linter had done all that. Let me revert.

zaneselvans:
I got rid of all the accidental linting changes, and made sure the datasette disabling stuff was in a simple commit all by itself.

PR #4280, opened 2025-05-20
Dbt add schema diff and update

marianneke:
# Overview

Closes #XXXX.

We need an elegant way to handle updates to dbt schemas which allow us to 

## What did you change?
- [x] Add method to DbtSchema class to allow reading existing schemas from yaml
- [x] Add functions to evaluate the difference between old and new schemas and print them in a report
- [x] Add update flag to all update functions to allow dbt schema to be overwritten if no information would be deleted (i.e. there is no information in the existing schema that is not present in the new schema)
- [x] Use this update flag also to allow the row counts to be updated (behavior identical to clobber in the row count case)

## Documentation

# Testing

## To-do list

### DIFF
--- a/src/pudl/scripts/dbt_helper.py
+++ b/src/pudl/scripts/dbt_helper.py
@@ -9,6 +9,7 @@ import click
 import duckdb
 import pandas as pd
 import yaml
+from enum import Enum
 from pydantic import BaseModel
 
 from pudl.logging_helpers import configure_root_logger, get_logger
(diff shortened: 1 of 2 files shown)


src/pudl/scripts/dbt_helper.py
```diff
@@ -167,6 +167,131 @@ def from_yaml(cls, schema_path: Path) -> "DbtSchema":
...
+
+
+def diff_dict_keys(field, old, new):
+    """Return a diff of dictionary keys as a nested dictionary."""
+    old_keys, new_keys = set((old or {}).keys()), set((new or {}).keys())
+    added, removed = list(new_keys - old_keys), list(old_keys - new_keys)
+    return {field: {"added": added, "removed": removed}} if added or removed else {}
+
+
+def diff_dbt_column(o, n):
```
jdangerx, line 189: Let's name these `old` and `new`, I was confused by the 1-letter names.

It would be nice, in general, to have some type hints here - I find them very helpful when thinking about what a function does.

jdangerx (changes requested):
I like adding this new flag! A soft clobber that only clobbers if it's non-destructive is nice. 

Some blocking comments:

* How should we test this? Seems like we should add unit tests for the cli behavior in `test/unit/dbt_helper_test.py` - we can test what happens with different schemas / desired updates / rowcounts / etc. Let me know if you want help coming up with test cases.
* Dict diffing seems like it should probably just live in a separate module and be tested separately from the CLI behavior as well.
  * What do you think about using a library for diffing like `deepdiff`? I've never used it, so IDK if it's worth pulling in that dependency vs. maintaining our own diffing + diff-printing code.
  * If we decide to keep our own diffing code I think it would be helpful to define the shape of a diff using e.g. a `dataclass` so that we can know what the different functions take in and spit out.

Non-blocking comment:

* It would be nice to have `--update` and `--clobber` be mutually exclusive, so we don't run into confusing behavior if both are set.

jdangerx (approved):
Thanks for adding the tests + renaming variables!

src/pudl/scripts/dbt_helper.py
```diff
@@ -166,6 +167,60 @@ def from_yaml(cls, schema_path: Path) -> "DbtSchema":
         with schema_path.open("r") as schema_yaml:
             return cls.model_validate(yaml.safe_load(schema_yaml))
 
+    @classmethod
+    def to_yaml(cls, schema_path: Path):
+        """Write DbtSchema object to YAML file."""
+        with schema_path.open("w") as schema_file:
+            yaml.dump(
```
jdangerx, line 174: This is just moving the old `_write_dbt_schema` function to the class, yes? Wouldn't happen to fix the [`prettier` formatting issue](https://github.com/catalyst-cooperative/pudl/issues/4119) for free?

src/pudl/scripts/dbt_helper.py
```diff
@@ -166,6 +167,60 @@ def from_yaml(cls, schema_path: Path) -> "DbtSchema":
...
+        "iterable_item_removed",
+        "attribute_deleted",
+    }
+
+    return any(key in diff and diff[key] for key in change_keys)
+
+
+def _print_schema_diff(diff: DeepDiff, old_schema: DbtSchema, new_schema: DbtSchema):
+    """Print old and new YAML, and summary of schema changes."""
+    print("\n======================")
```
jdangerx: non-blocking: Since this is intended only to happen interactively, using `print` should be fine here, though we could use the `logger` to print out these lines if we want to configure all the logging together...

also non-blocking: you could probably do this in one `print` call:

```python

print(
    "\n".join([
        "========",
        "Old YAML:",
        yaml.dump(...),
        ...
        _schema_diff_summary(diff) # returns a string now instead of printing out
    ])
)
```

test/unit/dbt_helper_test.py
```diff
@@ -504,3 +507,133 @@ def test_dbt_schema__add_column_tests(mocker, blank_schema):
         mocker.sentinel.second_model_test
         in two_model_tests.models[0].columns[0].data_tests
     )
+
+
+@pytest.mark.parametrize(
+    "diff,expected",
+    [
+        ({"dictionary_item_added": {"root['description']"}}, False),  # Add only
```
jdangerx: nit: these comments are helpful! if you want to include them in the parametrized test name you can use [`pytest.param`](https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids) or any of the other options listed at that link.

test/unit/dbt_helper_test.py
```diff
@@ -504,3 +507,133 @@ def test_dbt_schema__add_column_tests(mocker, blank_schema):
...
+            DbtTable(name="model2", description="added", columns=[]),
+        ],
+    )
+
+    diff = DeepDiff(
+        old_schema.model_dump(exclude_none=True),
+        new_schema.model_dump(exclude_none=True),
+        ignore_order=True,
+    )
+    _print_schema_diff_summary(diff)
```
jdangerx: nit: if we change this function to return a string, and then print out the string at the call site when necessary, then we won't need to capture stdout here which would be nice!

src/pudl/scripts/dbt_helper.py
```diff
@@ -259,40 +315,56 @@ def update_row_counts(
...
+        diff = DeepDiff(
+            old_schema.model_dump(exclude_none=True),
+            new_schema.model_dump(exclude_none=True),
+            ignore_order=True,
+            verbose_level=2,
+            view="tree",
+        )
+
+        if schema_has_removals_or_modifications(diff):
+            print(
```
jdangerx: This seems like a great candidate for `logger.warning()` instead of `print()`.

jdangerx (approved):
There are some small things I would personally change, but I'm working on letting other people do things their way when it doesn't make a big difference so these aren't blocking.

* refactoring a bit to have the different `_print_*` functions return the strings they *would* print, and then calling `print()` on their outputs - this makes them easier to compose with each other and test.
* also using the `logger.warning` and `logger.info` functions to replace the `print()` calls so it's easy to control the loglines all together. But I can see an argument for just calling `print()` when you're in an interactive-only environment.
* giving names to the parametrized pytest cases

You do also need to [update the conda-lock files](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/dev_setup.html#updating-the-pudl-development-environment) for mergeability - let me know if you need help with that!

src/pudl/scripts/dbt_helper.py
marianneke: I opted for a solution in between: Every function that is called, such as yaml.dump or getting the diff summary, goes into its own logger.info. This way we can isolate potential issues.

PR #4291, opened 2025-05-22
Reintegrate 88888 99999

### LINKED ISSUE #808: Re-integrate 88888 and 99999 data to eia861
Found and re-purposing this old issue. It was closed, but the 88888 and 99999 values were still getting dropped 🤷‍♀ .

# Description
Many years ago, when we first integrated the EIA 861 data, we came across rows with `88888` and `99999` in the `utility_id_eia` field for several tables. 

**What are these rows?**
- `88888` = Information for utilities who's identifying information has been redacted for privacy reasons. Each row represents an individual utility. 

- `99999` = State-level adjustment values added by EIA that account for capacity lost in conversion from DC to AC. These values are negative, and do not represent reported values from an individual utility.

**In what tables do these rows appear?**
- `88888`
     - **Demand Side Management** (one utility from 2010 - 2012)
     - **Sales** (one utility from 2008 - 2023)
     - **Utility Data** (nine utilities from Texas from 2009 - 2023)
     - **Operational Data Revenue** (nine utilities from Texas from 2008 - 2023)
     - **Balancing Authority** (one utility from 2010 - 2011, two in 2012)
     - **Dynamic Pricing** (one utility from 2020 - 2023)
     
- `99999`
     - **Sales** (one per year/state/ba_code/customer_class/service_type/business_model from 2001-2023)
          - A state, service type, and balancing authority-level adjustment is made for non-respondents and for customer-sited respondents.
     - **Non Net Metering** (one per year/state/ba_code from 2018-2023)
          - We make a state-level adjustment for missing PV capacity and to convert state total capacity to AC units for those respondents who report data in DC units; we use 0.8256 as a conversion factor to change DC to AC. For other energy sources, we have not established imputation procedures
     - **Net Metering** (one per year/state/ba_code from 2018-2023)
          - We make a state-level adjustment for missing PV capacity and to convert state total capacity to AC units for those respondents who report data in DC units; we use 0.8256 as a conversion factor to change DC to AC. For other energy sources, we have not established imputation procedures
     - **Advanced Metering Infrastructure** (one per year/state/ba_code from 2020-2023)
          - Unclear how these 99999 values are calculated for `non_amr_ami` (Number of non-AMR/AMI meters)....
     - **Balancing Authority** (we can remove these because there is no useful data here. I think they get introduced via PUDL, not the raw data).

**We removed most of these values from the tables they appeared in because:** 
- `88888` values were not unique primary keys
- `99999` values were not the same type of data as the rest of the table (a utility)

In the interest of preserving as much of the original data as possible, we think it's appropriate to re-integrate these values where they appear in the raw data. There are several caveats and re-integration methods to consider.

# Possible Solutions

`99999`: Because these values don't complicate the primary key, it's pretty easy to just re-integrate them when there are values that aren't NA. We don't have to build any work-arounds.

`88888`: These values are more complicated because they interfere with the primary key. However, there are only two tables with more than one `88888` value per year. 
- **Option 1**: Add another primary key column to distinguish `88888` values.
     - PRO: preserve data granularity
     - CON: add a column that is mostly NA
- **Option 2**: Aggregate `88888` values by the primary key.
     - PRO: ease of use by users, no primary key issues
     - CON: lose data granularity, need to explain aggregation somewhere.
- **Option 3**: Rename `88888` values to distinguish them
     - PRO: solves primary key issue
     - CON: can't be consistent year over year with rename, not as searchable.

We met as a team and decided that **Option 2** is the best for now.
     
# To-Do

**Re-integrate `99999`**
- [ ] Remove all references to removing `99999` rows from eia861
- [ ] Remove `99999` values from tables where all rows are NA -- this is a product of our normalization process and table splitting.
- [ ] Add usage warning in table description explaining `99999` rows.

**Re-integrate `88888`**
- [ ] Remove all references to removing `88888` rows from eia861
- [ ] Write a function to aggregate `88888` rows by the primary key and integrate it into the relevant tables where this is expected. Make sure primary key would check would fail on tables otherwise (so we can see if this happens in the future).
- [ ] Add a usage warning to the table description explaining the aggregated rows.

aesharpe:
# Overview

Closes #808

Adds rows where `utility_id_eia` is `99999` or `88888` back into eia923 and eia861 (see issue linked above for more detail).

## What did you change?

- [x] Remove documentation that talks about getting rid of those values
- [x] Remove lines of code that remove these values from 923 and 861
- [x] Write a function that combines certain 88888 values when necessary
- [x] Write a unit test that checks the combination of 88888 values
- [x] Update data source documentation page
- [ ] Update row count validation tests

**Possibilities**

- **Generic Function:** talked about making a more generic helper function for combining rows like 88888, but I didn't want to spend a bunch of time making something too generic / work with a lot of edge cases. For the time being, it's only applicable to 88888/99999 rows, but I'm sure we could adapt it in the future if needed.
- **Investigate `88888/99999` `plant_id_eia`:** There are some of these that could maybe be added back in? But that could also be more complicated and OOS for this. We should at least mention them in the docs.

## Documentation

# Testing

- Materialize 861 and 923 assets in dagster
- Run `pytest test/unit/transform/eia861.py`
## To-do list

**Final Run Through**


src/pudl/transform/eia861.py
```diff
@@ -1020,6 +1022,55 @@ def _harvest_associations(dfs: list[pd.DataFrame], cols: list[str]) -> pd.DataFr
...
+                len(non_num_cols) == 0
+                or ((group[non_num_cols] == group[non_num_cols].iloc[0]).all()).all()
+            ):
+                group = group.set_index(idx_cols)
+                non_num_group = group[non_num_cols].iloc[[0]]
+                num_group = group.drop(columns=non_num_cols).groupby(idx_cols).sum()
+                return pd.concat([non_num_group, num_group], axis=1).reset_index()
+            # Exclude rows with 88888 utility_id_eia that can't be combined due to different values in
+            # non-numeric columns.
+            return None
```
krivard, line 1061: is there a rough idea of how many rows we have to exclude? /would that be important for users to know?

krivard (commented):
The hard parts look right! Tagged a couple of potential optimizations, an order-of-operations question, and a couple potential things to document

src/pudl/transform/eia861.py
```diff
@@ -1020,6 +1022,55 @@ def _harvest_associations(dfs: list[pd.DataFrame], cols: list[str]) -> pd.DataFr
...
+                return group
+            # If there are no non-numeric columns, or all non-numeric columns have the same value,
+            # Sum the numeric columns and return the result.
+            if (
+                len(non_num_cols) == 0
+                or ((group[non_num_cols] == group[non_num_cols].iloc[0]).all()).all()
+            ):
+                group = group.set_index(idx_cols)
+                non_num_group = group[non_num_cols].iloc[[0]]
+                num_group = group.drop(columns=non_num_cols).groupby(idx_cols).sum()
```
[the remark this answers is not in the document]
aesharpe: Sadly it is needed for formatting purposes. If you just run `.sum()` it outputs a series with just the sums and does not retain the index columns (which we want to keep!) If you have a better idea of how to keep those I would love to know because I hate the unnecessary groupby too.

src/pudl/transform/eia861.py
aesharpe: That's a good point. Might want to add a check that there aren't more than a certain number getting dropped.

src/pudl/transform/eia861.py
```diff
@@ -1800,10 +1850,8 @@ def core_eia861__yearly_dynamic_pricing(
         "variable_peak_pricing",
     ]
 
-    raw_dp = _pre_process(
-        raw_eia861__dynamic_pricing.query("utility_id_eia not in [88888]").assign(
-            short_form=lambda x: _make_yn_bool(x.short_form)
-        )
+    raw_dp = _pre_process(raw_eia861__dynamic_pricing, idx_cols).assign(
+        short_form=lambda x: _make_yn_bool(x.short_form)
```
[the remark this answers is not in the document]
aesharpe: Hmmm, I'm not sure, but I will change it back just in case!

src/pudl/transform/eia861.py
```diff
@@ -1020,6 +1023,62 @@ def _harvest_associations(dfs: list[pd.DataFrame], cols: list[str]) -> pd.DataFr
...
+    agg_utils_88888 = utils_88888.groupby(idx_cols, group_keys=False).apply(
+        custom_group_agg
+    )
+    recombined_df = pd.concat(
+        [df[df["utility_id_eia"] != 88888], agg_utils_88888], ignore_index=True
+    )
+    # Make sure that the number of rows altered stays somewhat small. We don't expect
+    # there to be a lot of dropped or combined 88888 rows.
+    if (len(df) - len(recombined_df)) > 15:
+        ForkedPdb().set_trace()
```
aesharpe, line 1072: For testing, will remove later

src/pudl/transform/eia861.py
```diff
@@ -1020,6 +1023,62 @@ def _harvest_associations(dfs: list[pd.DataFrame], cols: list[str]) -> pd.DataFr
...
+        raise AssertionError(
+            f"Number of 88888 rows has changed by more than expected: {len(df) - len(recombined_df)}!"
+        )
+    # Check to make sure that the idx_cols are actually valid primary key cols:
+    return recombined_df
+
+    return df
```
aesharpe, line 1079: Fails right now -- investigating.

src/pudl/transform/eia861.py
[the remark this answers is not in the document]
krivard: oh wow.

i found a way but it is definitely unreadable and we should keep the groupby!!

if you are curious it is 

`num_group = pd.DataFrame(group.drop(columns=non_num_cols).sum()).transpose().set_index(group.index.take([0]))`

src/pudl/transform/eia861.py
```diff
@@ -528,13 +528,14 @@
     "NY": "NPCC",
     "NEW": "NPCC",
     "YORK": "NPCC",
+    "MISE": "MISO",
```
jdangerx, line 531: ha. good catch

src/pudl/transform/eia861.py
```diff
@@ -661,6 +664,11 @@ def _tidy_class_dfs(
     reshaped data after the fact, broadcasting their values across all of the records
     that they pertain to.
 
+    This function also checks for duplicate primary key values in the reshaped data, and
+    consolidates them by summing the data values. This is necessary because the EIA-861
+    data is not always clean, and sometimes contains duplicate records that are
+    identical except for the values in the class columns.
```
jdangerx: clarification for me: what are the 'class columns'?

src/pudl/transform/eia861.py
```diff
@@ -850,20 +856,12 @@ def clean_nerc(df: pd.DataFrame, idx_cols: list[str]) -> pd.DataFrame:
...
         A DataFrame with correct and clean NERC regions.
     """
-    idx_no_nerc = idx_cols.copy()
-    if "nerc_region" in idx_no_nerc:
-        idx_no_nerc.remove("nerc_region")
-
-    # Split raw df into primary keys plus nerc region and other value cols
-    nerc_df = df[idx_cols].copy()
-    other_df = df.drop(columns="nerc_region").set_index(idx_no_nerc)
-
```
jdangerx, line 866: I don't see any change in the intended behavior of the function... was this code just hanging out doing nothing useful before?

src/pudl/transform/eia861.py
```diff
@@ -1020,6 +1014,86 @@ def _harvest_associations(dfs: list[pd.DataFrame], cols: list[str]) -> pd.DataFr
...
+    values are known withholdings whereas the duplicate primary key rows are unknown data
+    quality issues.
+
+    Args:
+        df: The dataframe to process.
+        idx_cols: The index (primary key) columns of the input dataframe. They must
+            identify unique records in the input data.
+
+    Returns:
+        A dataframe with the combined or dropped 88888 utiltiy_id_eia values and unique pks.
```
jdangerx:
```suggestion
        A dataframe with the combined or dropped 88888 utility_id_eia values and unique pks.
```

src/pudl/transform/eia861.py
```diff
@@ -1020,6 +1014,86 @@ def _harvest_associations(dfs: list[pd.DataFrame], cols: list[str]) -> pd.DataFr
     return assn
 
 
+def _combine_88888_values(df: pd.DataFrame, idx_cols: list[str]) -> pd.DataFrame:
+    """Combine 88888 utility_id_values.
+
+    This function sums rows with a utility_id_eia of 88888 into a single row by
+    primary key. It drops rows with a utility_id_eia of 88888 if there are non-numeric
+    columns with different values that are impossible to combine. E.g.: boolean
+    columns where one value is Y and the other is N. This results in a small loss of
```
jdangerx, line 1023: non-blocking: you mention a few times that the loss of data is small - might be good to quantify that in the docs so people know what to expect.

src/pudl/transform/eia861.py
```diff
@@ -1020,6 +1014,86 @@ def _harvest_associations(dfs: list[pd.DataFrame], cols: list[str]) -> pd.DataFr
...
+    Returns:
+        A dataframe with the combined or dropped 88888 utiltiy_id_eia values and unique pks.
+    """
+    if 88888 not in df.utility_id_eia.unique():
+        return df
+    non_num_cols = (
+        df.set_index(idx_cols).select_dtypes(exclude="number").columns.tolist()
+    )
+
+    def custom_group_agg(group):
```
jdangerx: nit: `custom_group_agg` doesn't really tell me much about what this function does... something like `sum_only_numeric_values` might be a little clearer

src/pudl/transform/eia861.py
```diff
@@ -1020,6 +1014,86 @@ def _harvest_associations(dfs: list[pd.DataFrame], cols: list[str]) -> pd.DataFr
...
+    utils_88888 = df[df["utility_id_eia"] == 88888]
+    agg_utils_88888 = utils_88888.groupby(
+        idx_cols, group_keys=False, dropna=False
+    ).apply(custom_group_agg)
+    recombined_df = pd.concat(
+        [df[df["utility_id_eia"] != 88888], agg_utils_88888], ignore_index=True
+    )
+    # Make sure that the number of rows altered stays somewhat small. We don't expect
+    # there to be a lot of dropped or combined 88888 rows.
+    if (len(df) - len(recombined_df)) > 35:
```
jdangerx: blocking: Might be good for this to be a ratio instead of a flat number - that'll be a much more meaningful definition of 'somewhat small.'

src/pudl/transform/eia861.py
```diff
@@ -2176,7 +2220,6 @@ def core_operational_data_eia861(raw_eia861__operational_data: pd.DataFrame):
 
     Transformations include:
 
-    * Remove rows with utility ids 88888.
```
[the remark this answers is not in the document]
jdangerx: Agree, worth putting here as well as in pre_process docstring.

src/pudl/transform/eia861.py
```diff
@@ -2333,7 +2403,6 @@ def core_utility_data_eia861(raw_eia861__utility_data: pd.DataFrame):
...
```
jdangerx, line 2336: Same as above - probably something like "Aggregate values for rows with utility id 88888 (anonymized) - see _combine_88888_values and _pre_process for details."

src/pudl/transform/eia861.py
```diff
@@ -2420,9 +2483,7 @@ def core_utility_data_eia861(raw_eia861__utility_data: pd.DataFrame):
 
     # Transform RTO table
     transformed_ud_rto = tidy_ud_rto.assign(
-        rto_operation=lambda x: (
-            x.rto_operation.fillna(False).replace({"N": False, "Y": True})
-        ),
+        rto_operation=lambda x: (_make_yn_bool(x.rto_operation.fillna(False))),
```
jdangerx, line 2516: blocking: Do we no longer need to clean the "Y" and "N" values?

test/unit/transform/eia861.py
```diff
@@ -0,0 +1,107 @@
...
+        (actual_2, expected_2),
+        (actual_3, expected_3),
+        (actual_4, expected_4),
+    ],
+)
+def test__combine_88888_values(actual, expected):
+    """Test that combine_88888 correctly combines data from multiple sources."""
+    idx_cols = ["report_date", "utility_id_eia", "state"]
+    actual_test = eia861._combine_88888_values(actual, idx_cols)
+    actual_test.to_pickle("/Users/austensharpe/Desktop/actual_test.pkl")
```
jdangerx: blocking: Do we need to pickle this df or was this leftover from debugging? If we do still need it, we should probably not pickle it to your personal desktop :wink:

test/unit/transform/eia861.py
```diff
@@ -0,0 +1,107 @@
...
+
+from io import StringIO
+
+import pandas as pd
+import pytest
+
+from pudl.metadata.fields import apply_pudl_dtypes
+from pudl.transform import eia861
+
+# Test that rows with the same primary key are combined correctly
```
jdangerx, line 11: Nice work with setting up the parametrized tests and adding comments to explain each df!

test/unit/transform/eia861.py
```diff
@@ -0,0 +1,107 @@
...
+2019-01-01,88888,pd.NA,ERCOT,1100
+"""
+    ),
+).pipe(apply_pudl_dtypes, group="eia")
+
+
+@pytest.mark.parametrize(
+    "actual,expected",
+    [
+        (actual_1, expected_1),
```
jdangerx: nit: 1, 2, 3, and 4 are not the *most* useful differentiators here, though obviously the comments are helpful. If you want to make this clearer, you can use [pytest's "test case ID" functionality](https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids) for parametrized tests.

That'd look like

```python
@pytest.mark.parametrize(
    "actual,expected",
    [
        pytest.param(actual_1, expected_1, id="same_pk_does_combine"),
        ...
    ],
)
def test__combine...
```


Up to you. The test output will be slightly nicer.

test/unit/transform/eia861.py
```diff
@@ -0,0 +1,107 @@
...
+        (actual_2, expected_2),
+        (actual_3, expected_3),
+        (actual_4, expected_4),
+    ],
+)
+def test__combine_88888_values(actual, expected):
+    """Test that combine_88888 correctly combines data from multiple sources."""
+    idx_cols = ["report_date", "utility_id_eia", "state"]
+    actual_test = eia861._combine_88888_values(actual, idx_cols)
```
jdangerx: nit: the typical vocabulary word here is "observed" vs. "expected" value. And "actual" would probably be the "input" or "raw" data :shrug:

jdangerx (changes requested):
Looks pretty good - the docs updates help a lot with clarity about what the hell we're doing with 88888. Just a few blocking questions but nothing structural - let me know if anything is confusing!

src/pudl/transform/eia861.py
aesharpe: It's defined earlier in the doc string, but I think this latter half of the sentence was created by co-pilot, and I'm not sure I trust it, so I'm removing it.

src/pudl/transform/eia861.py
aesharpe: Yeah, basically it was an early Austen attempt at split-apply-combine....I thought that by performing the transforms on only a subset of columns and then sticking them back together with the other columns would save time....but it was silly and it was causing problems and not really doing anything useful.... 🤦

src/pudl/transform/eia861.py
aesharpe: This `.replace({"N": False, "Y": True})` was doing the same thing as the `_make_yn_bool` function, so I just added that in there instead.

test/unit/transform/eia861.py
aesharpe: This is a great suggestion!

test/unit/transform/eia861.py
aesharpe: hahahahah good catch, it's leftover from my testing

test/unit/transform/eia861.py
aesharpe: Does that mean that the we can use `actual` for the input data and should just change `actual_test` to be `observed`?

zaneselvans:
I'm testing out the new "update the row counts for me" functionality by running a deployment on this branch.  It should take about 2 hours to run, and the build should fail (due to the discrepancy in row count expectations) but should also produce a new `etl_full_row_counts.csv` that can be downloaded do your local repo with...

```bash
gcloud storage cp gs://builds.catalyst.coop/2025-06-23-2324-17a492713-reintegrate-88888-99999/etl_full_row_counts.csv dbt/seeds/
```

At which point `git diff` will show you what changes it found.

test/unit/transform/eia861.py
```diff
@@ -0,0 +1,106 @@
...
+from pudl.metadata.fields import apply_pudl_dtypes
+from pudl.transform import eia861
+
+# Test that rows with the same primary key are combined correctly
+actual_1 = pd.read_csv(
+    StringIO(
+        """report_date,utility_id_eia,state,ba_code,value
+2019-01-01,88888,TX,ERCOT,100
+2019-01-01,88888,TX,ERCOT,300
+2019-01-01,88888,pd.NA,ERCOT,800
```
zaneselvans: Sorry for the drive-by comment but is the literal string "pd.NA" what we want here? Or should it just be a blank entry, which will be turned into an actual pandas null value by virtue of the `apply_pudl_dtypes`?

aesharpe: I think blank here is fine, that's a good point. I'll convert it.

aesharpe:
Hoookayyy 

I think (hope) I addressed all the comments. 
Some of the row changes are relatively large due to primary keys (i.e., things like customer class and fuel type effectively quadrupling or octupling(?) the number of rows for a given utility). This is all within the realm of expected value changes!!

src/pudl/transform/eia861.py
```diff
@@ -1028,47 +1020,91 @@ def _combine_88888_values(df: pd.DataFrame, idx_cols: list[str]) -> pd.DataFrame
...
-        )
-        recombined_df = pd.concat(
-            [df[df["utility_id_eia"] != 88888], agg_utils_88888], ignore_index=True
+    if 88888 not in df.utility_id_eia.unique():
+        return df
+    non_num_cols = (
+        df.set_index(idx_cols).select_dtypes(exclude="number").columns.tolist()
+    )
+
+    def sum_numeric_values_when_strings_match(group):
```
krivard: nice rename 👍

src/pudl/transform/eia861.py
```diff
@@ -1028,47 +1020,91 @@ def _combine_88888_values(df: pd.DataFrame, idx_cols: list[str]) -> pd.DataFrame
...
+            na_cols = idx_has_na[idx_has_na].index.tolist()
+            mod_idx_cols = [x for x in idx_cols if x not in na_cols]
+            # Now we can safely sum the numeric columns, recombine with the non-numeric
+            # columns, and return the result.
+            group = group.set_index(mod_idx_cols)
+            non_num_group = group[non_num_cols + na_cols].iloc[[0]]
+            num_group = (
+                group.drop(
+                    columns=non_num_cols + na_cols
+                )  # STATE is not part of non_num cols so it is getting blanked instead of NA-ed
```
krivard, line 1062: If this is a debugging leftover you can just yeet it, but if it needs to stay: what makes STATE special? none of the other columns are being mentioned by name. are there special considerations the caller needs to keep in mind related to state data?

src/pudl/transform/eia861.py
```diff
@@ -1028,47 +1020,91 @@ def _combine_88888_values(df: pd.DataFrame, idx_cols: list[str]) -> pd.DataFrame
...
+                .groupby(mod_idx_cols, dropna=False)
+                .sum()
+            )
+            return pd.concat([non_num_group, num_group], axis="columns").reset_index()
+        # When there are rows that can't be combined due to discrepancies in the non-numeric
+        # columns, we drop the duplicates and return the first row.
+        # This is a bit of a hack, but it ensures that we don't end up with multiple rows
+        # with the same primary key, which would cause problems later on.
+        # There are very few instances of this anyways. On the order of 1-2 per idx group.
+        # Right now there the following tables have the following dupes. Note that most of
```
krivard, line 1070:
```suggestion
        # As of August 2025 there are the following dupes. Note that most of
```

src/pudl/transform/eia861.py
```diff
@@ -1028,47 +1020,91 @@ def _combine_88888_values(df: pd.DataFrame, idx_cols: list[str]) -> pd.DataFrame
...
+        [df[df["utility_id_eia"] != 88888], agg_utils_88888], ignore_index=True
+    )
+    # We don't expect there to be a lot of dropped or combined 88888 rows so we'll
+    # monitor the percent decrease in rows. Need the len_diff < -1 because the unit
+    # tests only test a few rows and therefore the difference in rows is a
+    # much higher percentage than 0.001.
+    len_diff = len(recombined_df) - len(df)
+    if (len_diff < -1) and (len_diff / len(df) < -0.001):
+        raise AssertionError(
+            f"Number of 88888 rows has changed by more than expected: {len_diff}!"
```
krivard, line 1094: Since the test is based on a fraction of the total rowcount and not just on the absolute change in rowcount, we want the error message to include the expected change in rowcount along with the actual value:

```suggestion
            f"Expected reduction in 88888 rows less than {len(df)*0.001} but found: {-len_diff}!"
```

src/pudl/transform/eia861.py
```diff
@@ -2378,18 +2440,21 @@ def core_utility_data_eia861(raw_eia861__utility_data: pd.DataFrame):
     * Tidy subset of the data by NERC region.
     * Tidy subset of the data by RTO.
     * Convert Y/N columns to booleans.
+    * Combine 88888 utility ID EIA rows with duplicate primary keys.
+    * Remove 88888 utility ID EIA rows with duplicate primary keys containing distinct,
+      non-consolidateable qualitative information. This is a small number of records and
+      represents less than a 0.001 decrease in rows. See _combine_88888_values and
+      _pre_process for details.
```
krivard: On the one hand it's nice to have this information to hand everywhere it has an impact; on the other, if the decrease fraction changes, then (1) we'll have to update it in a whole bunch of places and (2) if we forget any of them, then it'll say different fractions in different places, and that will be very confusing.

if you're willing to commit to keeping all these values updated and in sync, then 🫡 keep it as is; otherwise consider making the following change everywhere this text appears:

```suggestion
    * Combine 88888 utility ID EIA rows with duplicate primary keys.
    * Remove 88888 utility ID EIA rows with duplicate primary keys containing distinct,
      non-consolidateable qualitative information. See _combine_88888_values and
      _pre_process for details.
```

and keep the fraction only in the `_combine_88888_values` docstring.

test/unit/transform/eia861.py
```diff
@@ -0,0 +1,106 @@
...
+).pipe(apply_pudl_dtypes, group="eia")
+
+
+@pytest.mark.parametrize(
+    "actual,expected",
+    [
+        pytest.param(actual_1, expected_1, id="same_pk_does_combine"),
+        pytest.param(actual_2, expected_2, id="no_88888_does_not_combine"),
+        pytest.param(actual_3, expected_3, id="non_numeric_does_not_combine"),
+        pytest.param(actual_4, expected_4, id="na_index_col_does_not_duplicate"),
```
krivard, line 99: these param ids are great -- i didn't know you could do that!

docs/templates/eia861_child.rst.jinja
```diff
@@ -68,7 +68,12 @@ The Freedom of Information Act (FOIA), 5 U.S.C. §552, the Department of Energy
 10 C.F.R. §1004.11, implementing the FOIA, and the Trade ﻿Secrets Act, 18 U.S.C. §1905 allows
 qualifying respondents to restrict access to their data. According to sources at the EIA,
 approximately 3 respondents have used this as a means to keep their utility-level data proprietary.
-These entries appear under the utility id ``88888`` in the data.
+These entries appear under the utility id ``88888`` in the data. Having multiple utilities with the
+same ``utility_id_eia`` breaks the primary key rules for several tables. We handle this by summing
+values with the same primary key and non-numeric column values. When there are two rows with the
+same primary key and different non-numeric column values, we keep one and drop the other. This results
+in a slight loss of data (on the order of a few rows) but preserves the primary key structure of the
```
krivard: ehhhh in my dialect 'a few' means like 3-8 and there are more than that

```suggestion
in a slight loss of data (less than one row in a thousand) but preserves the primary key structure of the
```

src/pudl/transform/eia861.py
```diff
@@ -1028,47 +1020,91 @@ def _combine_88888_values(df: pd.DataFrame, idx_cols: list[str]) -> pd.DataFrame
...
+        if (
+            len(non_num_cols) == 0
+            or (group[non_num_cols] == group[non_num_cols].iloc[0]).all().all()
+        ):
+            # Special case: if there are NA values in the index columns we can't
+            # combine two dfs with the same index without creating two rows.
+            # This messes up the results
+            # , so we drop the NA values from the index
+            # when combining the newly aggregated groups. This does not affect the
+            # actual primary key, it's just a slightly hacky way to combine the rows correctly.
```
krivard, line 1052: maybe hacky but it's also super clever and having the comment explain it is 🏆

src/pudl/transform/eia861.py
aesharpe: This was a debugging leftover, nice catch!

src/pudl/transform/eia861.py
aesharpe: This is a great suggestion!

test/unit/transform/eia861.py
aesharpe: Either @zaneselvans or @jdangerx showed me and I'm so thankful

src/pudl/transform/eia861.py
```diff
@@ -528,26 +528,31 @@
...
     * Standardize common NA values found in EIA spreadsheets.
     * Drop the ``early_release`` column, which only contains non-null values when the
       data is an early release, and we extract this information from the filenames, as
       it's uniform across the whole dataset.
     * Convert report_year column to report_date.
+    * Aggregate values for rows with utility id 88888 (anonymized) - see _combine_88888_values
+      for details.
     """
-    return (
+    prep_df = (
```
jdangerx, line 549: nit: why assign this value to `prep_df` instead of returning it like before? easier to throw a breakpoint in there?

src/pudl/transform/eia861.py
```diff
@@ -1020,6 +1014,96 @@ def _harvest_associations(dfs: list[pd.DataFrame], cols: list[str]) -> pd.DataFr
...
+        # with the same primary key, which would cause problems later on.
+        # There are very few instances of this anyways. On the order of 1-2 per idx group.
+        # As of August 2025 there are the following dupes. Note that most of
+        # the dupes are the same utility in different years.
+        # - BA table: 1
+        # - OD table: 16
+        # - Sales table: 32
+        # - UD table: 15
+        # - DP table: 8
+        no_dupes = group.drop_duplicates(subset=idx_cols)
```
jdangerx, line 1085: nit: might be good to explicitly pass in [`keep="first"`](https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.drop_duplicates.html) so that people know *which* data gets dropped...

I'm not sure there's really a meaningful way to sort the data so that there's a meaningful difference between "keep the first" and "randomly pick one to keep" though. So :shrug:

jdangerx (approved):
Looking good! The tests are clear & give me confidence that this will work, at least for the cases we've thought of already :wink:

PR #4296, opened 2025-05-25
Remove no_null_rows tests

### LINKED ISSUE #4104: Migrate no_null_rows validation tests to dbt
In some of our wide, denormalized output tables where we've merged together lots of information, it is sometimes expected that we will have a lot of columns with null values, but only up to a point. The `no_null_rows` validation checks that all rows have at least some minimum proportion of the columns in the row has a non-null value. It's currently applied in the `validate/mcoe_test.py` and `integration/output.py` test modules.

Converting this to a `dbt` test will require summing up the number of null vs. non-null columns in each row and then selecting any rows with more nulls than are allowed by the threshold.

zaneselvans:
In an attempt to cut down memory usage in our CI, I went hunting for tests that could be migrated or deprecated and found that only the MCOE tables were being checked with pudl.validate.no_null_rows().

I experimented with trying to create a `dbt` macro / SQL query that could do this check, but it ended up seeming pretty complex, and we only use this check in one validation test, it's a one-liner and fast with pandas, and we actually use the same function inline as a merge validator elsewhere in the codebase, so I moved those tests into asset checks on the MCOE assets.

In the process I found that there was some PudlTabl cruft in the integration and validation tests that looked like it was also unnecessarily reading a large dataframe out of pudl.sqlite, which I've tried to clean up a bit here.

Also found that the AEO table asset checks were not marked as blocking... which seemed wrong?  I see that there are also still row-count checks being done in those asset checks, which should probably be removed since they'll duplicate the dbt row count checks.

Closes #4104 

# Testing

- Ran the integration tests in GitHub CI with a `workflow_dispatch` run to make sure changes to `PudlTabl` didn't break anything.
- Materialized the EIA AEO assets to make sure blocking asset checks are passing.

## To-do list

- [x] Run CI on GHA


src/pudl/analysis/mcoe.py
```diff
@@ -233,6 +242,57 @@ def mcoe_generators_asset(
...
+
+def mcoe_asset_check_factory(spec: McoeCheckSpec) -> AssetChecksDefinition:
+    """Turn a MCOE check spec into an AssetChecksDefinition."""
+
+    @asset_check(asset=spec.asset, blocking=spec.blocking)
+    def excessively_null_rows(df: pd.DataFrame) -> AssetCheckResult:
+        """Check that the MCOE dataframe has no excessively null rows."""
+        excessively_null_rows = (
+            df.isna().sum(axis="columns") / len(df.columns) > spec.max_null_fraction
+        )
```
zaneselvans, line 282: In poking around at this, I found that we're actually using `pv.no_null_rows` as a validation step in some big merges in e.g. the Plant Parts EIA module, as well as below in this one).  At the moment this core checking code is duplicated, since the other validation function is set up to be usable with `df.pipe()` which isn't really compatible with the use case here. But maybe the core logic can be factored out and serve both purposes without duplication.

src/pudl/analysis/mcoe.py
```diff
@@ -647,7 +707,7 @@ def mcoe(
         .pipe(
             pudl.validate.no_null_rows,
             df_name="fuel_cost + capacity_factor",
-            thresh=0.9,
+            max_null_fraction=0.9,
         )
```
zaneselvans, line 711: Example of current usage of `no_null_rows` as an inline merge validation.

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -147,33 +147,13 @@
...
-    pudl_out = pudl.output.pudltabl.PudlTabl(pudl_engine,freq='YS')
-
-Then make the table via pudl_out:
-
-.. code-block:: python
-
-    plant_parts_eia = pudl_out.plant_parts_eia()
-
-
-OR make the table via objects in this module:
```
zaneselvans, line 171: Removed because this is, I think, no longer accurate given that `PudlTabl` is just a wrapper around the PUDL DB now.

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -147,33 +147,13 @@
...
+For debugging and development purposes, you can build the plant parts using the classes
+from this module directly in a notebook or script like this:
 
 .. code-block:: python
 
     gens_mega = MakeMegaGenTbl().execute(mcoe, own_eia860)
-    parts_compiler = MakePlantParts(pudl_out)
+    parts_compiler = MakePlantParts()
```
zaneselvans, line 156: Doesn't actually depend on `pudl_out` at this point.  You have to pass dataframes to the appropriate methods.

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -590,11 +570,11 @@ def __init__(self):
         # for all of the plant parts
         self.id_cols_list = make_id_cols_list()
 
-    def execute(self, gens_mega, plants_eia860, utils_eia860):
+    def execute(self, gens_mega, plants_eia860, utils_eia860) -> pd.DataFrame:
```
zaneselvans, line 573: This and the other changes below are just type annotations and reflowing docstrings.

src/pudl/output/pudltabl.py
```diff
@@ -127,9 +126,6 @@ def __init__(
         self.fill_tech_desc = fill_tech_desc  # only for eia860 table.
         self.unit_ids = unit_ids
 
-        # Used to persist the output tables. Returns None if they don't exist.
-        self._dfs = defaultdict(lambda: None)
-
```
zaneselvans, line 132: We don't actually cache dataframes in PudlTabl any more. Not sure why this was sticking around.

src/pudl/transform/eiaaeo.py
```diff
@@ -626,7 +626,7 @@ class AeoCheckSpec:
 def make_check(spec: AeoCheckSpec) -> AssetChecksDefinition:
     """Turn the AeoCheckSpec into an actual Dagster asset check."""
 
-    @asset_check(asset=spec.asset)
+    @asset_check(asset=spec.asset, blocking=True)
```
zaneselvans, line 629: @jdangerx Is there a reason that these checks weren't marked as blocking before?

src/pudl/validate.py
```diff
@@ -95,32 +95,39 @@ def check_date_freq(df1: pd.DataFrame, df2: pd.DataFrame, mult: int) -> None:
         )
 
 
-def no_null_rows(df, cols="all", df_name="", thresh=0.9):
-    """Check for rows filled with NA values indicating bad merges.
+def no_null_rows(
+    df: pd.DataFrame,
+    cols="all",
+    df_name: str = "",
+    max_null_fraction: float = 0.9,
```
zaneselvans, line 102: Changed the name of `thresh` parameter to be more descriptive.

test/integration/output_test.py
```diff
@@ -81,19 +81,3 @@ def test_eia_outputs(fast_out, df1_name, df2_name, mult, kwargs):
...
     pv.check_date_freq(df1, df2, mult)
-
-
-@pytest.mark.parametrize(
-    "df_name,thresh",
-    [
-        ("mcoe_generators", 0.9),
-    ],
-)
-def test_null_rows(fast_out, df_name, thresh):
```
zaneselvans, line 92: Migrated into an asset check.

test/validate/mcoe_test.py
```diff
@@ -73,26 +56,6 @@ def test_no_null_cols_mcoe(pudl_out_mcoe, live_dbs, df_name):
...
         "winter_capacity_estimate",
     ]
-    df = pudl_out_mcoe.__getattribute__(df_name)()
+    df = pudl_out_eia.__getattribute__(df_name)()
     cols = [col for col in df.columns if col not in deprecated_cols]
     pv.no_null_cols(df, cols=cols, df_name=df_name)
-
-
-@pytest.mark.parametrize("df_name,thresh", [("mcoe", 0.8)])
-def test_no_null_rows_mcoe(pudl_out_mcoe, live_dbs, df_name, thresh):
```
zaneselvans, line 82: Migrated into asset check.

test/validate/mcoe_test.py
```diff
@@ -19,23 +19,6 @@
...
-    rate, fuel costs, and capacity factors. However, for the purposes of testing, we
-    don't want to lose those values -- that's the kind of thing we're looking for.  So
-    here we override those defaults, causing the MCOE output dataframe to be cached with
-    all the nasty details, so it is available for the rest of the tests that look at
-    MCOE results in this module
-    """
-    if live_dbs and pudl_out_eia.freq is not None:
-        logger.info("Reading MCOE data (with generator attributes) out of the PUDL DB.")
-        _ = pudl_out_eia.mcoe_generators()
-    return pudl_out_eia
```
zaneselvans, line 36: When we switched to `PudlTabl` just being a wrapper around the database, I think this became obsolete, and potentially led to an unintentional change in behavior below.  Right now `PudlTabl` can grab a couple of different versions of the MCOE aka generators tables. This creation of a new `pudl_out_mcoe` fixture was meant to populate a cached dataframe in PudlTabl I think, and change what data was available via `PudlTabl.mcoe()` -- but that data is available directly now via `PudlTabl.mcoe_generators()`  So, below the test now looks at `mcoe_generators` directly.

src/pudl/analysis/mcoe.py
cmgosnell: why isn't using `pv.no_null_rows` compatible here? we use it below with `pipe` but that doesn't mean it needs to be used that way.

zaneselvans: Because it will `raise` an exception if it finds null rows, rather than allowing us to construct an `AssetCheckResult` to return.

But actually duh, I could just try/except and catch that exception.

test/validate/mcoe_test.py
```diff
@@ -48,14 +31,14 @@ def pudl_out_mcoe(pudl_out_eia, live_dbs):
         "hr_by_gen",
         # "fuel_cost",
         "capacity_factor",
-        "mcoe",
+        "mcoe_generators",
```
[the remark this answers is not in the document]
zaneselvans: Yeah, I think that's right.

test/validate/mcoe_test.py
cmgosnell: can you add a new issue that would make a test that does the thing that this old documentatation says its trying to do?

src/pudl/analysis/plant_parts_eia.py
zaneselvans: Okay I updated these method calls to reflect what we're actually doing now, but I'm pretty sure there are other aspects of the documentation in here that have not been updated to reflect the dagsterization of all this.  Maybe that's something we could fix if we were to create a Methodology section in the docs.

```py
    gens_mega = MakeMegaGenTbl().execute(
        mcoe=out_eia__yearly_generators,
        own_eia860=out_eia860__yearly_ownership,
    )
    plant_parts_eia = MakePlantParts().execute(
        gens_mega=out_eia__yearly_generators_by_ownership,
        plants_eia860=out_eia__yearly_plants,
        utils_eia860=out_eia__yearly_utilities,
    )
```

cmgosnell (approved):
this looks good to me!! thanks for adding #4300 and for making the `pv` test usable in this context.

PR #4298, opened 2025-05-26
Skip `ferc1_eia_train_test` in GHA CI to cut memory usage

zaneselvans:
# Overview

- I used this branch to experiment with some ways of reducing memory usage in our GHA CI.
- I didn't find any clever trick -- in the end I just set the `test/integration/ferc1_eia_train_test.py` module to be skipped when running on GitHub. This seems to pretty reliably result in the tests being able to run. But it's apparently not a 100% guarantee!  I has at least one CI run that still got killed during the ETL. So I think we are teetering right on the edge of being able to run on the free runners.
- It turns out this means a big chunk of the FERC1 to EIA record linkage code doesn't get run in the tests any more, so I had to reduce our expected test coverage from 93% to 92% to get it to pass.
- This is not a great solution, but I'm hoping that there are other integration tests that can be migrated into dbt or asset checks and become much more memory efficient.
- I'm also hopeful that when we rip out `PudlTabl` and have all of the integration & validation tests pull data from the Parquet outputs, there will be some big performance / memory efficiency improvements, which will allow us to turn this test back on. But we can't wait for all that to happen before the EIA-930 PR gets merged in.

# Testing

- Ran the tests locally and they still all pass (and include the disabled module)
- Ran the CI on GHA with a `workflow_dispatch` repeatedly, until getting to some combination of changes that lets them pass reliably.


zaneselvans:
>  iirc this is the hyperparameter tuning

Aaah, that makes sense.

katie-lamb (approved):
This doesn't seem concerning to me. IIRC training the model doesn't work well on the fast ETL because there aren't enough years of data to train a well-performing model. So we freeze the weights from a training run on all the data, and use them in the fast ETL instead of re-training.

I'll check if we do end up training in the full ETL, or if we should re-train the model now that there are new years of data. 

Also, these tests don't actually test the training process, they mostly test some training data setup/formatting before training and the overrides that happen after the training. I'll check to see that this whole setup and override process is supposed to be skipped during the fast ETL.

PR #4307, opened 2025-05-30
Switch PudlTabl from reading SQLite to reading Parquet

zaneselvans:
# Overview

To try and speed up our CI and potentially reduce memory usage, this change adds a parameter to the PudlTabl output object allowing the user to select whether the class should read from SQLite or Parquet outputs. It uses the same method of reading from Parquet as our existing `ParquetIOManager` in Dagster.

The new `table_source="parquet"` parameter is now used in all the CI tests that actually read data.

This is a stopgap to try and speed things up and work around resource constraints until we can remove `PudlTabl` completely. If reading from Parquet instead of SQLite makes CI / builds faster and/or avoids memory constraints on GitHub runners, then we can switch the logic over to fully relying on Parquet and remove the SQLite reading logic.

🎉 It works! I turned all the memory constrained tests back on.

## To-do list

- [x] Run `make pytest-validation` locally and see if it works / how long it takes (4 min, down from 11 min)
- [x] Run `make pytest-integration` locally and see if it works / how long it takes (50 min, same as normal)
- [x] If those work, kick off GHA CI with `workflow_dispatch` and see how long it takes
- [x] If the CI works, re-enable all the tests that were turned off b/c of memory constraints
- [x] If CI passes w/o running out of memory, kick off a branch deployment and see how long it takes.
- [x] Have someone else review it.
- [x] Remove logic from `PudlTabl` that reads SQLite -- now only read from Parquet
- [x] Revert target test coverage to 93%


src/pudl/output/pudltabl.py
```diff
@@ -249,29 +261,113 @@ def _register_output_methods(self: Self):
...
         for table_name, method_name in table_method_map_yearly_only.items():
             self.__dict__[method_name] = partial(
-                self._get_table_from_db,
+                self.get_table,
                 table_name=table_name,
                 allowed_freqs=["YS"],
             )
 
-    def _get_table_from_db(
+    def _get_table_from_parquet(  # noqa: C901
```
zaneselvans, line 291: This method mirrors both `_get_table_from_db()` and the `ParquetIOManager` to try and speed up & reduce the memory intensity of reading PUDL Tables in our CI & testing.

test/validate/ferc1_test.py
```diff
@@ -1,8 +1,4 @@
-"""Validate post-ETL FERC Form 1 data and the associated derived outputs.
-
-These tests depend on a FERC Form 1 specific PudlTabl output object, which is a
-parameterized fixture that has session scope.
-"""
```
zaneselvans, line 5: Removed PudlTabl cruft -- no longer true/relevant.

src/pudl/output/pudltabl.py
```diff
@@ -249,29 +261,113 @@ def _register_output_methods(self: Self):
...
+        if date_col is not None:
+            if start_filter_value is not None:
+                filters.append((date_col, ">=", start_filter_value))
+            if end_filter_value is not None:
+                filters.append((date_col, "<=", end_filter_value))
+
+        # Read with filters and optimizations
+        read_kwargs = {
+            "source": parquet_path,
+            "schema": res.to_pyarrow(),
```
zaneselvans, line 352: Oh whoops this should just use `pyarrow_schema` defined above.

test/conftest.py
```diff
@@ -168,7 +168,7 @@ def pudl_out_ferc1(live_dbs: bool, pudl_engine: sa.Engine, request) -> PudlTabl:
     """Define parameterized PudlTabl output object fixture for FERC 1 tests."""
     if not live_dbs:
         pytest.skip("Validation tests only work with a live PUDL DB.")
-    return PudlTabl(pudl_engine=pudl_engine, freq=request.param)
+    return PudlTabl(pudl_engine=pudl_engine, freq=request.param, table_source="parquet")
```
zaneselvans, line 171: Redirect all the PudlTabl objects used in our tests (the only place it's still actually used) to read data from Parquet outputs instead of SQLite.  There are a number of these objects defined in other fixtures for various different testing purposes.  I tried to change them all.

test/integration/ferc1_eia_train_test.py
```diff
@@ -30,12 +29,6 @@
 
 logger = logging.getLogger(__name__)
 
-if os.getenv("GITHUB_ACTIONS", False):
-    pytest.skip(
-        reason="As of 2025-05-25 this test is causing the GitHub runner to run out of memory.",
-        allow_module_level=True,
-    )
-
```
zaneselvans, line 38: Reading from Parquet instead seems to make this unnecessary.

test/validate/mcoe_test.py
```diff
@@ -29,7 +29,7 @@
     [
         "hr_by_unit",
         "hr_by_gen",
-        # "fuel_cost",
```
zaneselvans, line 32: Works fine if we're reading from Parquet.

test/conftest.py
zaneselvans: If we switch to `PudlTabl` reading from Parquet all the time, I'll remove these.

src/pudl/io_managers.py
```diff
@@ -333,7 +333,12 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
         table_name = get_table_name_from_context(context)
         parquet_path = PudlPaths().parquet_path(table_name)
         res = Resource.from_id(table_name)
-        df = pq.read_table(source=parquet_path, schema=res.to_pyarrow()).to_pandas()
+        df = pq.read_table(
+            source=parquet_path,
+            schema=res.to_pyarrow(),
+            use_threads=True,  # Enable multi-threading for faster reads
```
jdangerx: afaict this [defaults to true](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.read_table.html) - having it explicit here is great but I'm wondering if you needed to set this for it to actually work?

src/pudl/output/pudltabl.py
```diff
@@ -249,29 +261,113 @@ def _register_output_methods(self: Self):
...
+            raise ValueError(
+                f"{table_name} needs one of these frequencies {allowed_freqs}, "
+                f"but got {self.freq}"
+            )
+        table_name = self._agg_table_name(table_name)
+        logger.warning(
+            "PudlTabl is deprecated and will be removed from the pudl package "
+            "once known users have migrated to accessing the data directly from "
+            "pudl.sqlite. To access the data returned by this method, "
+            f"use the {table_name} table in the pudl.sqlite database."
```
jdangerx: Suppose we could update this to be parquet-specific...

```suggestion
            "PudlTabl is deprecated and will be removed from the pudl package "
            "once known users have migrated to accessing the data directly from "
            "parquet files. To access the data returned by this method, "
            f"use the {table_name}.parquet file in the pudl outputs."
```

src/pudl/output/pudltabl.py
```diff
@@ -249,29 +261,113 @@ def _register_output_methods(self: Self):
...
+                start_filter_value = pd.to_datetime(self.start_date).year
+            if self.end_date is not None:
+                end_filter_value = pd.to_datetime(self.end_date).year
+
+        # Add date filters if applicable
+        if date_col is not None:
+            if start_filter_value is not None:
+                filters.append((date_col, ">=", start_filter_value))
+            if end_filter_value is not None:
+                filters.append((date_col, "<=", end_filter_value))
```
jdangerx: non-blocking, but it could be nice to pull this out into a separate `build_pq_date_filters(schema: pa.Schema, start_date, end_date) -> list[tuple]` function. Then you can throw a few unit tests at it and also this function gets a bit more readable.

src/pudl/output/pudltabl.py
```diff
@@ -249,29 +261,113 @@ def _register_output_methods(self: Self):
...
+        read_kwargs = {
+            "source": parquet_path,
+            "schema": pyarrow_schema,
+            "use_threads": True,  # Enable multi-threading for faster reads
+            "memory_map": True,  # Use memory mapping for better memory efficiency
+        }
+
+        # Only add filters if we have any
+        if filters:
+            read_kwargs["filters"] = filters
```
jdangerx: Annoying that they don't treat `filters=[]` as "there are no filters, move along." But oh well, I guess that's how it is. If you do end up pulling the filters out into its own function you could have it spit out `None` if there are no filters available, which would let you avoid this annoying `if`.

src/pudl/output/pudltabl.py
```diff
@@ -249,29 +261,113 @@ def _register_output_methods(self: Self):
...
+
+        # Read only the columns we need based on the resource schema
+        columns_to_read = res.get_field_names()
+        if columns_to_read:
```
jdangerx: It looks like `get_field_names()` can only return a non-truthy value if it's an empty list:
```
    def get_field_names(self) -> list[str]:
        """Return a list of all the field names in the resource schema."""
        return [field.name for field in self.schema.fields]
```

Double-checking: when we have an empty list for `self.schema.fields` we want to read *all* the parquet columns?

If not, we might be able to get away with just shoving this into the original `read_kwargs` above and avoiding this if statement.

If so, we can use the walrus operator to verrry slightly clean this up:

```suggestion
        if columns_to_read := res.get_field_names():
```

src/pudl/io_managers.py
zaneselvans: Ah, you're right. No I didn't realize it defaulted to True -- I was just blindly adding keyword args that promised to make things faster and less memory intensive. I think this doesn't actually end up helping much, since like 90% of the time the `get_table` function takes to run is spent imposing the PUDL dtypes (on the big monthly generators table it takes 3 seconds to read the parquet file, but 40 seconds to read it and enforce the schema. But that's still much faster than the 250 seconds it took to read from SQLite and enforce the schema.

Pandas 2.3 will make pyarrow a mandatory dependency, and it looks like the release is immanent.  Once that's in place, we might want to look at switching to using the `pyarrow` dtypes across the board, or at least making whatever changes are necessary to ensure that every column in the Parquet outputs produces a reasonable dtype when it's read into Pandas -- right now the datetime columns are getting interpreted as `object` by default for reasons I don't understand. And there might be other undesirable conversions. But if they all had the right dtypes natively, there would be no need for a lot of the work that's done by `enforce_schema()` and our reading/writing would be much faster. And also end users could just `df.read_parquet(dtype_backend="pyarrow")` and avoid needing to mess around with dtypes manually, which seems like it would be much more ergonomic.

zaneselvans:
@jdangerx My plan was to remove the SQLite infrastructure/docs if this seems like it works and is a good (if temporary) change, but I was putting off making those changes until I'd actually seen that the new Parquet setup worked and didn't break any tests, and somebody else had given it a 👍🏼 . It seems like this is all good and solving some resource constraints, so I'll go ahead and make those changes.

src/pudl/io_managers.py
zaneselvans: This might also fix the problem we're having in the PUDL Viewer where Null dates are interpreted as 0 which means `1970-01-01T00:00:00`

src/pudl/output/pudltabl.py
zaneselvans: Hmm. If `self.schema.fields == []` then... there can't be any data in the Parquet file and something is terribly wrong. This might just be an oops on my part. We actually want to read all the columns all the time -- PudlTabl only grabs whole tables (columnwise)

src/pudl/output/pudltabl.py
```diff
@@ -252,66 +223,91 @@ def _register_output_methods(self: Self):
...
+            return None
+
+        # Build filters
+        filters = []
+        if self.start_date is not None:
+            filters.append((date_col, ">=", convert_date(self.start_date)))
+        if self.end_date is not None:
+            filters.append((date_col, "<=", convert_date(self.end_date)))
+
+        return filters if filters else None
```
zaneselvans, line 296: Pulled it out and simplified the logic a bit.

src/pudl/output/pudltabl.py
```diff
@@ -252,66 +223,91 @@ def _register_output_methods(self: Self):
...
+            use_threads=True,  # Enable multi-threading for faster reads
+            memory_map=True,  # Use memory mapping for better memory efficiency
+            filters=self._build_parquet_date_filters(
+                start_date=self.start_date,
+                end_date=self.end_date,
+                pyarrow_schema=pyarrow_schema,
+            ),
+        ).to_pandas()
+        # Enforce the expected PUDL dtypes and other constraints on the DataFrame:
+        return res.enforce_schema(df)
```
zaneselvans: Much simpler!

src/pudl/output/pudltabl.py
```diff
@@ -44,62 +45,34 @@ class PudlTabl:
...
-        pudl_engine: sa.Engine,
         freq: Literal["YS", "MS", None] = None,
         start_date: str | date | datetime | pd.Timestamp = None,
         end_date: str | date | datetime | pd.Timestamp = None,
-        fill_fuel_cost: bool = False,
-        roll_fuel_cost: bool = False,
         fill_net_gen: bool = False,
-        fill_tech_desc: bool = True,
-        unit_ids: bool = False,
     ) -> Self:
```
zaneselvans, line 51: Removed a number of arguments that are totally unused here and elsewhere.  We don't have them anywhere in our codebase, and hopefully everyone else who might have relied on this at some point is off by now! We've had it deprecated for more than a year.

src/pudl/output/pudltabl.py
zaneselvans: I think this was a brain fart -- I was thinking of the one special case of a generation fuel table below where we select only a subset of the columns, but it's handled down there and it's only a single special case, so I've removed this no-op.

test/unit/output/pudltabl_test.py
```diff
@@ -0,0 +1,130 @@
+"""Unit tests for PudlTabl class methods."""
```
zaneselvans: One thing LLMs are great at is scaffolding unit tests.

src/pudl/output/pudltabl.py
```diff
@@ -403,19 +351,10 @@ def gen_eia923(self: Self, update: bool = False) -> pd.DataFrame:
...
         return gen_df
-
-    ###########################################################################
-    # GLUE OUTPUTS
-    ###########################################################################
-    def epacamd_eia(self: Self) -> pd.DataFrame:
-        """Read the EPACAMD-EIA Crosswalk from the PUDL DB."""
-        return pd.read_sql("core_epa__assn_eia_epacamd", self.pudl_engine).pipe(
-            apply_pudl_dtypes, group="glue"
-        )
```
zaneselvans, line 421: Not sure why this was implemented as a special case, since it seems to be covered by the normal output methods. I rolled it up into the method registration above since it was the last reference to `pudl_engine`.

src/pudl/output/pudltabl.py
```diff
@@ -44,87 +45,52 @@ class PudlTabl:
...
+        self.start_date = (
+            min(pudl.helpers.get_working_dates_by_datasource("ferc1"))
+            if start_date is None
+            else pd.to_datetime(start_date)
+        )
+        self.end_date = (
+            max(pudl.helpers.get_working_dates_by_datasource("eia"))
+            if end_date is None
+            else pd.to_datetime(end_date)
+        )
```
zaneselvans, line 92: Condensed and simplified.

src/pudl/output/pudltabl.py
```diff
@@ -210,6 +176,8 @@ def _register_output_methods(self: Self):
             "out_eia923__AGG_fuel_receipts_costs": "frc_eia923",
             "out_eia923__AGG_generation": "gen_original_eia923",
             "out_eia923__AGG_generation_fuel_combined": "gf_eia923",
+            # EPA CAMD to EIA crosswalk
+            "core_epa__assn_eia_epacamd": "epacamd_eia",
```
zaneselvans, line 179: Added this here to remove the last ref to `self.pudl_engine` and since it didn't actually seem to be a special case (see below).

src/pudl/output/pudltabl.py
zaneselvans: Yeah this was a mess and ruff was complaining about it being too complex too. Much simplified now.

test/unit/output/pudltabl_test.py
```diff
@@ -0,0 +1,130 @@
...
+from datetime import date, datetime
+
+import pandas as pd
+import pyarrow as pa
+import pytest
+
+from pudl.output.pudltabl import PudlTabl
+
+
+class TestPudlTablAggTableName:
```
jdangerx: nit: this seems really easy to parameterize.

But then again, it's even easier to just leave as is. hence, nit.

test/unit/output/pudltabl_test.py
```diff
@@ -0,0 +1,130 @@
...
+class TestPudlTablEdgeCases:
+    """Test edge cases and error conditions."""
+
+    def test_empty_schema(self):
+        """Test behavior with empty PyArrow schema."""
+        pudl_tabl = PudlTabl()
+        empty_schema = pa.schema([])
+
+        filters = pudl_tabl._build_parquet_date_filters(empty_schema)
+        assert filters is None
```
jdangerx: Did you want to test `columns` behavior in this case too?

test/unit/output/pudltabl_test.py
```diff
@@ -0,0 +1,130 @@
...
+    def test_invalid_frequency_raises_error(self):
+        """Test that invalid frequencies raise ValueError."""
+        with pytest.raises(ValueError, match="freq must be one of None, 'MS', or 'YS'"):
+            PudlTabl(freq="QS")  # Quarterly not supported
+
+        with pytest.raises(ValueError, match="freq must be one of None, 'MS', or 'YS'"):
+            PudlTabl(freq="D")  # Daily not supported
+
+        with pytest.raises(ValueError, match="freq must be one of None, 'MS', or 'YS'"):
+            PudlTabl(freq="invalid")
```
jdangerx: This is a lot of code to test that "we do, in fact, check for membership in this list of values at initialization." Which is fine, I guess, but kind of brittle and obscures the more interesting parts of the frequency behavior.

I'm more interested in checking "if I set `freq` to monthly, and try to get a yearly-only table, does that throw an error?" and the related constellation of behavior there - what do you think?

test/unit/output/pudltabl_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def test_date_conversion(self):
```
jdangerx: Similarly to above, this seems to mostly test some pretty trivial code: "does `pd.to_datetime` work?" - here are some other related things that might be more useful to test:

* "do start/end default to the right dates?" (though that might live in a pudl helpers test instead of pudltabl test)
* given start/end dates (or a lack of them), do we get the expected filter behavior?

test/unit/output/pudltabl_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def test_invalid_date_format_raises_error(self):
+        """Test that invalid date formats raise appropriate errors."""
+        with pytest.raises((ValueError, TypeError)):
+            PudlTabl(start_date="not-a-date")
+
+        with pytest.raises((ValueError, TypeError)):
+            PudlTabl(end_date="2020-13-45")  # Invalid month/day
+
+    def test_fill_net_gen_parameter(self):
```
jdangerx: Another instance of "this test seems like it *could* be testing something useful but isn't yet". Currently it's just "does the parameter get set?", but we could be testing "does `fill_net_gen` affect the behavior of `gen_eia923` as expected?"

jdangerx (changes requested):
Nice refactoring/cleanup! And a good start on the tests :)

The tests (as scaffolded by LLM) could probably use a bit more consideration in terms of "what is this actually testing? what is the interesting behavior that we need to make sure is actually happening?"

zaneselvans:
Given that we're trying to blow away `PudlTabl` entirely as soon as possible, and we've stripped it out of every application outside of a few remaining tests, I'm kind of inclined to... just get rid of the test module. I think what we would have wanted to test here if anything is that the switch from reading from SQLite to Parquet and associated change in filtering logic doesn't actually impact the contents of the outputs, which would require having both the old and new logic in place, and would make this into an integration, rather than unit test, since it would require access to the data, not just the class.

However, the fact that none of the remaining tests that use `PudlTabl` are complaining about anything after the switch gives at least a little bit of an indication that nothing got egregiously broken. There are currently zero instances of non-default start/end dates being used in the tests. If I did find some weird behavior in `PudlTabl` at this point, I think that would just make me want to prioritize putting time into ripping it out rather than fixing the behavior.

zaneselvans:
Found another test that was disabled across the board for memory usage issues and am poking at making it just read from Parquet directly -- and only the columns it needs. As a prototype of removing PudlTabl dependence directly.

src/pudl/glue/ferc1_eia.py
zaneselvans, line 1: No substantive changes in this file -- just type annotations and formatting.

src/pudl/helpers.py
```diff
@@ -2263,3 +2265,59 @@ def retry(
...
+        filters=filters,
+        use_threads=True,
+        memory_map=True,
+    ).to_pandas()
+
+    # Only enforce schema if we're reading all columns
+    if columns is None:
+        return resource.enforce_schema(df)
+    # For specific columns, apply PUDL dtypes for the columns we have
+    return apply_pudl_dtypes(df, group=resource.field_namespace)
```
zaneselvans, line 2323: The need for this functionality was showing up in a couple of places, so I consolidated it. PudlTabl and the PudlParquetIOManager both use this now, as do the tests that need to load data directly.

src/pudl/output/pudltabl.py
```diff
@@ -252,66 +219,81 @@ def _register_output_methods(self: Self):
...
+            f"data returned by this method, read from the {table_name} table in the "
+            f"pudl.sqlite database or access the {table_name}.parquet file in the PUDL "
+            "outputs."
         )
+        res = Resource.from_id(table_name)
+        pyarrow_schema: pa.Schema = res.to_pyarrow()
+
+        # Build date filters and use the helper function
+        filters = self._build_parquet_date_filters(pyarrow_schema=pyarrow_schema)
+        return get_parquet_table(table_name, filters=filters)
```
zaneselvans, line 269: Consolidated reusable logic from this method over in get_parquet_table()

test/conftest.py
```diff
@@ -163,55 +162,6 @@ def pudl_etl_parameters(etl_settings: EtlSettings) -> DatasetsSettings:
     return etl_settings.datasets
 
 
-@pytest.fixture(scope="session", params=["YS"], ids=["ferc1_annual"])
-def pudl_out_ferc1(live_dbs: bool, pudl_engine: sa.Engine, request) -> PudlTabl:
-    """Define parameterized PudlTabl output object fixture for FERC 1 tests."""
-    if not live_dbs:
-        pytest.skip("Validation tests only work with a live PUDL DB.")
-    return PudlTabl(pudl_engine=pudl_engine, freq=request.param)
-
```
zaneselvans, line 172: Got rid of all the PudlTabl fixtures here -- two of them were already not being used.  Now the tests use `pudl.helpers.get_parquet_table()` to read data directly

test/integration/ferc1_eia_train_test.py
```diff
@@ -27,37 +27,13 @@
...
-def plant_parts_eia(fast_out_annual):
-    """The plant_parts_eia output table."""
-    return fast_out_annual.plant_parts_eia().reset_index()
-
 
 @pytest.fixture(scope="module")
-def eia_ferc1(fast_out_annual):
-    """The eia_ferc1 output table."""
-    # the pudl_out name here is the old, non-alphabetized ordering.
-    return fast_out_annual.ferc1_eia()
```
zaneselvans: Now I read these dataframes from parquet directly, inside the test functions. This results in some duplicate reads, since the same data is used in more than one test, but they're fast.

test/integration/glue_test.py
zaneselvans, line 1: No real changes in here -- just type annotations.

test/integration/output_test.py
```diff
@@ -2,82 +2,90 @@
...
-def fast_out(pudl_engine):
-    """A PUDL output object for use in CI."""
-    return pudl.output.pudltabl.PudlTabl(
-        pudl_engine,
-        freq="MS",
-        fill_fuel_cost=True,
-        roll_fuel_cost=True,
-        fill_net_gen=True,
-        fill_tech_desc=True,
-    )
```
zaneselvans, line 23: Removed in favor of reading the Parquet data directly.

test/integration/output_test.py
```diff
@@ -2,82 +2,90 @@
...
+        ("out_eia923__monthly_boiler_fuel", 12 / 1),
+        ("out_eia923__monthly_fuel_receipts_costs", 12 / 1),
+        ("out_eia923__monthly_generation", 12 / 1),
+        ("out_eia923__monthly_generation_fuel_by_generator_energy_source", 12 / 1),
+        ("out_eia923__monthly_generation_fuel_by_generator", 12 / 1),
+        ("out_eia923__monthly_generation_fuel_combined", 12 / 1),
+        ("_out_eia__monthly_heat_rate_by_unit", 12 / 1),
+        ("_out_eia__monthly_heat_rate_by_generator", 12 / 1),
+        ("_out_eia__monthly_fuel_cost_by_generator", 12 / 1),
+        ("_out_eia__monthly_capacity_factor_by_generator", 12 / 1),
```
zaneselvans, line 63: - PudlTabl method names replaced with the table names they would have mapped to.
- Removed the df1 parameter since it was always the same. Ditto the kwargs.

test/integration/output_test.py
```diff
@@ -2,82 +2,90 @@
...
-        ("gens_eia860", "hr_by_gen", 12 / 1, {}),
-        ("gens_eia860", "fuel_cost", 12 / 1, {}),
-        ("gens_eia860", "capacity_factor", 12 / 1, {}),
-        pytest.param(
-            "gens_eia860",
-            "mcoe_generators",
-            12 / 1,
-            {"all_gens": False},
-            marks=pytest.mark.xfail(reason="MCOE has time coverage issues."),
-        ),
```
zaneselvans, line 72: A resource-intenstive test that we expected to fail. Removed.

test/integration/output_test.py
```diff
@@ -2,82 +2,90 @@
...
+
+    # Get columns needed for date frequency checking
+    # Include data_maturity if it exists in schema for proper incremental_ytd filtering
+    def get_date_freq_columns(table_name: str) -> list[str]:
+        """Get columns needed for date frequency checking."""
+        columns = ["report_date"]
+        resource = Resource.from_id(table_name)
+        if "data_maturity" in resource.get_field_names():
+            columns.append("data_maturity")
+        return columns
```
zaneselvans, line 82: This lets us read just 1-2 columns from the Parquet files, rather than the whole tables, which dramatically reduces memory usage and read times, since the frequency checker just needs the date index and data maturity columns. I guess I could have done the filtering here too, rather than over in check_date_freq.

test/unit/analysis/allocate_gen_fuel_test.py
```diff
@@ -102,141 +101,75 @@ def test_distribute_annually_reported_data_to_months_if_annual():
...
-    ):
-        self._gens_eia860 = gens_eia860
-        self._gen_eia923 = gen_eia923
-        self._gen_original_eia923 = gen_original_eia923
-        self._generation_fuel_eia923 = generation_fuel_eia923
-        self._plants_eia860 = plants_eia860
-        self._boiler_fuel_eia923 = boiler_fuel_eia923
-        self._boiler_generator_assn_eia860 = boiler_generator_assn_eia860
-
-        self.freq = freq
```
zaneselvans, line 129: Removed the mocking, since really the only thing this is doing is feeding dataframes in to the tests below.  Now they're defined as constants and passed in directly through the test function parameterizations.

test/validate/eia_test.py
```diff
@@ -3,39 +3,59 @@
...
+        "_out_eia__yearly_generators",
+        "out_eia860__yearly_ownership",
+        "out_eia__yearly_plants",
+        "_out_eia__plants_utilities",
+        "out_eia__yearly_utilities",
+        "core_eia860__scd_emissions_control_equipment",
+        "core_eia860__assn_yearly_boiler_emissions_control_equipment",
+        "out_eia860__yearly_emissions_control_equipment",
+        "core_eia860__assn_boiler_stack_flue",
+        "core_eia860__assn_boiler_cooling",
```
zaneselvans, line 49: Replace the PudlTabl method names with explicit table names. I'm pretty sure this list covers all the parameterized (yearly, monthly, unaggregated) tables we were testing before. But also, this is a stopgap until we get the no null cols test working in Dagster asset checks.

test/validate/mcoe_test.py
```diff
@@ -13,33 +13,46 @@
...
+        "_out_eia__yearly_capacity_factor_by_generator",
+        "_out_eia__yearly_derived_generator_attributes",
+        "out_eia__yearly_generators",
+        # Monthly tables
+        "_out_eia__monthly_heat_rate_by_unit",
+        "_out_eia__monthly_heat_rate_by_generator",
+        "_out_eia__monthly_fuel_cost_by_generator",
+        "_out_eia__monthly_capacity_factor_by_generator",
+        "_out_eia__monthly_derived_generator_attributes",
+        "out_eia__monthly_generators",
```
zaneselvans, line 45: Again, swapping in table names for PudlTabl methods.

test/validate/plant_parts_eia_test.py
```diff
@@ -22,61 +24,58 @@
...
-def test_run_aggregations(pudl_out_eia, live_dbs):
+def test_run_aggregations(live_dbs: bool, pudl_engine: sa.Engine) -> None:
     """Run a test of the aggregated columns.
 
     This test will used the plant_parts_eia, re-run groubys and check similarity.
     """
     if not live_dbs:
         pytest.skip("Data validation only works with a live PUDL DB.")
-    if pudl_out_eia.freq == "YS":  # Annual only.
-        logger.info("Testing ownership fractions for owned records.")
```
zaneselvans, line 33: Most of the lines changed below are just because I flipped the nature of this check, so it bails early.

test/validate/plant_parts_eia_test.py
```diff
@@ -22,61 +24,58 @@
...
-            if part_name == "plant_match_ferc1":
-                test_merge = prep_test_merge(
-                    part_name,
-                    plant_parts_eia,
-                    plant_parts_eia.loc[
-                        plant_parts_eia.plant_part != "plant_match_ferc1"
-                    ],
+    mcoe = get_parquet_table(table_name="out_eia__yearly_generators")
+    own_eia860 = get_parquet_table(table_name="out_eia860__yearly_ownership")
+    plant_parts_eia = get_parquet_table(table_name="out_eia__yearly_plant_parts")
```
zaneselvans, line 39: Read these tables directly rather than through PudlTabl.

zaneselvans (commented):
Okay I.... went on a bit of a PudlTabl extermination rampage, as an experiment in using Claude Sonnet 4.0 for refactoring with agent mode. It was pretty trippy! At least for this kind of tedious shuffling of stuff around, it worked very well.

The only remaining dependency on PudlTabl that I'm aware of is in the FERC1-EIA glue / manual plant & utility mapping utility functions, which also interact with a test that exercises that code. That seemed a little more involved so I'm punting it for the moment.

src/pudl/analysis/record_linkage/eia_ferc1_train.py
```diff
@@ -516,6 +515,20 @@ def validate_override_fixes(
...
+        len(
+            ferc_override_dups := only_overrides[
+                only_overrides["record_id_ferc1"].duplicated(keep=False)
+            ]
+        )
+        == 0
+    ), (
+        f"Found record_id_ferc1 duplicates: \
+    {ferc_override_dups.record_id_ferc1.unique()}"
+    )
```
zaneselvans, line 530: Discovered that we were not checking this, because we were getting an XPASS in the tests.

src/pudl/analysis/record_linkage/eia_ferc1_train.py
```diff
@@ -426,25 +426,25 @@ def check_if_already_in_training(training_data, validated_connections):
...
-    eia_ferc1,
-    training_data,
-    expect_override_overrides=False,
-    allow_mismatched_utilities=True,
+    validated_connections: pd.DataFrame,
+    ppe: pd.DataFrame,
+    eia_ferc1: pd.DataFrame,
+    training_data: pd.DataFrame,
+    expect_override_overrides: bool = False,
+    allow_mismatched_utilities: bool = True,
```
zaneselvans, line 434: Just adding type annotations and cleaning up the docstring.

test/integration/ferc1_eia_train_test.py
```diff
@@ -118,12 +123,20 @@ def test_validate_override_fixes(
...
+    # Get data tables with only the columns needed by validate_override_fixes
+    # to reduce memory usage during testing
+    plant_parts_eia = get_parquet_table(
+        "out_eia__yearly_plant_parts",
+        columns=["record_id_eia", "utility_id_pudl", "report_year"],
+    )
+    eia_ferc1 = get_parquet_table(
+        "out_pudl__yearly_assn_eia_ferc1_plant_parts",
+        columns=["record_id_ferc1", "report_date"],
+    )
```
zaneselvans, line 139: This test was still *occasionally* but not reliably causing the runner to get killed.

Reduce memory usage by reading only the columns that are actually referenced in the validate_override_fixes function.

test/integration/ferc1_eia_train_test.py
```diff
@@ -140,22 +153,23 @@ def test_validate_override_fixes(
...
-    )
+    with expectation:
+        validate_override_fixes(
+            validated_connections=test_df,
+            ppe=plant_parts_eia,
+            eia_ferc1=eia_ferc1,
+            training_data=eia_ferc1_training_data_restricted,
+            expect_override_overrides=False,
+            allow_mismatched_utilities=False,
+        )
```
zaneselvans, line 164: Use pytest.raises() and the context manager to communicate more specific test outcomes, rather than marking xfail -- these aren't failures, they're tests that verify an exception is raised appropriately.

test/integration/ferc1_eia_train_test.py
```diff
@@ -140,22 +153,23 @@ def test_validate_override_fixes(
...
 
 
 def test_generate_all_override_spreadsheets(
     pudl_engine: sa.Engine,  # Required to ensure that the data is available.
 ):
     """Test the genation of the override spreadsheet for mapping FERC-EIA records."""
     # Get data tables directly
-    plant_parts_eia = get_parquet_table("out_eia__yearly_plant_parts").reset_index()
+    plant_parts_eia = get_parquet_table("out_eia__yearly_plant_parts")
```
zaneselvans: No point in resetting the index on a dataframe we just read out of Parquet -- this is a vestige from a bygone era.

test/integration/ferc1_eia_train_test.py
```diff
@@ -105,47 +82,62 @@ def eia_ferc1_training_data():
...
         pytest.param(
             ["True"],
             [2020],
             ["299_2020_plant_total_14354"],  # EIA id already in training data
             ["f1_steam_2020_12_134_2_2"],  # FERC1 id NOT in training data
             [246],
-            marks=pytest.mark.xfail(reason="EIA id already in training data."),
+            pytest.raises(
+                AssertionError,
+                match=r"The following EIA records area already in the training data",
```
jdangerx:
```suggestion
                match=r"The following EIA records are already in the training data",
```

Looks like the other instances of this string are in `pudl.analysis.record_linkage.eia_ferc1_train.py:589` and `:598`.

test/integration/ferc1_eia_train_test.py
```diff
@@ -105,47 +82,62 @@ def eia_ferc1_training_data():
...
         pytest.param(
             ["True"],
             [2020],
             ["294_2020_plant_owned_14354"],  # EIA id NOT in training data
             ["f1_steam_2020_12_134_3_1"],  # FERC1 id already in training data
             [246],
-            marks=pytest.mark.xfail(reason="FERC1 id already in training data."),
+            pytest.raises(
+                AssertionError,
+                match=r"The following FERC 1 records area already in the training data",
```
jdangerx:
```suggestion
                match=r"The following FERC 1 records are already in the training data",
```

test/integration/ferc1_eia_train_test.py
jdangerx: Nice parametrization of the context managers! Didn't know about nullcontext.

PR #4313, opened 2025-06-07
Migrate from deprecated SourceAsset to new AssetSpec

### LINKED ISSUE #3977: Switch from using the deprecated `SourceAsset` to `AssetSpec`
Switch from using the deprecated [`SourceAsset`](https://docs.dagster.io/_apidocs/assets#dagster.SourceAsset) to [`AssetSpec`](https://docs.dagster.io/concepts/assets/external-assets). 

The FERC Form 1 and Form 714 raw data are extracted into SQLite files before the main ETL begins. To bring the raw SQLite tables into dagster, we create `SourceAsset`s for all the tables. To create a `SourceAsset` you just need to initialize it with a table name and an io manager key.

Dagster is removing `SourceAsset`s in 2.0.0 so we need to start using `AssetSpec`s! This is what the `SourceAsset` docs say:

> Use AssetSpec instead. If using the SourceAsset io_manager_key property, use AssetSpec(...).with_io_manager_key(...)..

Here is an example of how to make this change:
https://github.com/catalyst-cooperative/pudl/blob/9dd1ab9fe483c58541547703b6bd85c922a2aaa5/src/pudl/extract/ferc1.py#L360-L366

to 

```python
raw_ferc1_dbf_assets = [
    AssetSpec(
        key=AssetKey(f"raw_ferc1_dbf__{table_name}"),
    ).with_io_manager_key(""ferc1_dbf_sqlite_io_manager")
    for table_name in dbf_table_names
]
```

```[tasklist]
### Tasks
- [ ] Replace SourceAsset with AssetSpec for FERC 714 and extraction
- [ ] Incorporate AssetSpec into `get_asset_keys()`
```

zaneselvans:
# Overview

Update our code to use the new `AssetSpec` class, instead of the deprecated `SourceAsset`

Closes #3977 

# Testing

- Rematerialized FERC-714 and FERC Form 1 core assets (which have the source assets affected by this PR upstream) locally without any problems.
- Ran [integration tests in CI on GitHub](https://github.com/catalyst-cooperative/pudl/actions/runs/15510289818/job/43670345264)

### DIFF
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -2,8 +2,6 @@
 
 import importlib.resources
 import itertools
-import os
-import warnings
 
 import pandera as pr
 from dagster import (
@@ -11,9 +9,8 @@ from dagster import (
     AssetChecksDefinition,
     AssetKey,
     AssetsDefinition,
-    AssetSelection,
+    AssetSpec,
     Definitions,
-    SourceAsset,
     asset_check,
     define_asset_job,
     load_asset_checks_from_modules,
@@ -119,6 +116,7 @@ default_assets = list(
         load_assets_from_modules(
             modules,
             group_name=group_name,
+            include_specs=True,
         )
         for group_name, modules in all_asset_modules.items()
     )
@@ -170,7 +168,7 @@ def asset_check_from_schema(
 
 
 def _get_keys_from_assets(
-    asset_def: AssetsDefinition | SourceAsset | CacheableAssetsDefinition,
+    asset_def: AssetsDefinition | AssetSpec | CacheableAssetsDefinition,
 ) -> list[AssetKey]:
     """Get a list of asset keys.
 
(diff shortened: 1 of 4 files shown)


src/pudl/etl/__init__.py
```diff
@@ -2,18 +2,15 @@
 
 import importlib.resources
 import itertools
-import os
-import warnings
```
zaneselvans, line 6: Removed some derelict unused imports.

src/pudl/etl/__init__.py
```diff
@@ -119,6 +116,7 @@
         load_assets_from_modules(
             modules,
             group_name=group_name,
+            include_specs=True,
```
zaneselvans, line 119: This was the magic sauce that was missing before -- tells it to pick up specs as well as full-blown assets.

zaneselvans:
Merged these changes into #4314

PR #4315, opened 2025-06-09
feat: capture DBT output and get failure context

jdangerx:
Added a function that runs `dbt build`, captures the output, and uses the output to run the duckdb queries that will give us failure output.

Next steps:

* handle weighted quantile tests differently.
* handle errors as well as failures.
* wrap this function in a CLI command
* use this function in integration tests

# Overview

Closes #XXXX.

## Documentation

# Testing

## To-do list

### DIFF
--- a/src/pudl/dbt_wrapper.py
+++ b/src/pudl/dbt_wrapper.py
@@ -1,0 +1,74 @@
+"""Wrap DBT invocations so we can get custom behavior."""
+
+from contextlib import chdir
+from typing import NamedTuple, cast
+
+import duckdb
+import pandas as pd
+from dbt.artifacts.schemas.results import TestStatus
+from dbt.artifacts.schemas.run import RunExecutionResult
+from dbt.cli.main import dbtRunner, dbtRunnerResult
+from dbt.contracts.graph.nodes import GenericTestNode
+
+from pudl.workspace.setup import PUDL_ROOT_PATH, PudlPaths
+
+
+def get_failed_nodes(results: RunExecutionResult) -> list[GenericTestNode]:
+    """Get test node output from tests that failed."""
+    return [res.node for res in results if res.status == TestStatus.Fail]
+
+
+class NodeContext(NamedTuple):
+    """Associate a node's *name* with information describing what went wrong."""
+
+    # TODO 2025-06-09: consider adding whether this failed or errored, and a
+    # free-form text description field
+    name: str
+    context: pd.DataFrame
+
+    def pretty_print(self):
+        """Nice output for logging to stdout."""
+        return f"{self.name}:\n\n{self.context}"
+
+
+def get_context_for_nodes(nodes: list[GenericTestNode]) -> list[NodeContext]:
+    """Post-process node output to figure out failure context.
+
+    For most tests, just run the compiled code in the node.
+
+    (TODO 2025-06-09) for weighted quantile tests, run the "debug quantiles" DBT
+    operation.
+    """
+    contexts = []
+    duckdb_path = PudlPaths().output_file("pudl_dbt_tests.duckdb")
+    with duckdb.connect(duckdb_path) as con:
+        for node in nodes:
+            con.execute(node.compiled_code)
+            node_df = con.fetchdf()
+            contexts.append(NodeContext(name=node.name, context=node_df))
+    return contexts
+
+
+def run_build(model_selection):
+    """Run the DBT build and get failure information back.
+
+    * run the DBT build using our selection, returning test failures
+    * get extra context for each test failure
+    * print out test failure context
+    """
+    dbt_target = "etl-full"
+    cli_args = ["--target", dbt_target, "--select", model_selection]
+    dbt = dbtRunner()
+    dbt_dir = PUDL_ROOT_PATH / "dbt"
+
+    # 2025-06-06 TODO: use the --project-dir flag instead of chdiring all the time?
+    with chdir(dbt_dir):
+        _ = dbt.invoke(["deps"])
+        _ = dbt.invoke(["seed"])
+        build_output: dbtRunnerResult = dbt.invoke(["build"] + cli_args)
+        build_results = cast(RunExecutionResult, build_output.result)
+
+    failed_nodes = get_failed_nodes(build_results)
+    failure_contexts = get_context_for_nodes(failed_nodes)
+
+    print("\n\n====\n\n".join(c.pretty_print() for c in failure_contexts))
--- a/src/pudl/workspace/setup.py
+++ b/src/pudl/workspace/setup.py
@@ -13,6 +13,8 @@ logger = pudl.logging_helpers.get_logger(__name__)
 
 PotentialDirectoryPath = DirectoryPath | NewPath
 
+PUDL_ROOT_PATH = Path(__file__).parent.parent.parent.parent
+
 
 class PudlPaths(BaseSettings):
     """These settings provide access to various PUDL directories.


jdangerx:
fyi @krivard - no need to review at this point, just a heads up on where I'm at here.

src/pudl/dbt_wrapper.py
```diff
@@ -0,0 +1,82 @@
...
+from pudl.workspace.setup import PUDL_ROOT_PATH, PudlPaths
+
+
+def get_failed_nodes(results: RunExecutionResult) -> list[GenericTestNode]:
+    """Get test node output from tests that failed."""
+    return [res.node for res in results if res.status == TestStatus.Fail]
+
+
+class NodeContext(NamedTuple):
+    """Associate a node's *name* with information describing what went wrong."""
```
jdangerx, line 22: Figured this was a little more explicit than using `dict[str, pd.DataFrame]`.

src/pudl/dbt_wrapper.py
```diff
@@ -0,0 +1,82 @@
...
+    (TODO 2025-06-09) for weighted quantile tests, run the "debug quantiles" DBT
+    operation.
+    """
+    contexts = []
+    duckdb_path = PudlPaths().output_file("pudl_dbt_tests.duckdb")
+    with duckdb.connect(duckdb_path) as con:
+        for node in nodes:
+            con.execute(node.compiled_code)
+            node_df = con.fetchdf()
+            contexts.append(NodeContext(name=node.name, context=node_df))
```
jdangerx, line 46: So in theory, when we want to deal with making better weighted quantile output, this would be where we have different handlers based on different node names.

test/integration/dbt_test.py
```diff
@@ -51,40 +48,16 @@ def test_dbt(
...
-    assert test_result.success
+    if not test_result.success:
+        raise AssertionError(
+            f"failure contexts:\n{test_result.format_failure_contexts()}"
+        )
 
 
 @pytest.mark.script_launch_mode("inprocess")
-def test_dbt_helper(
+def test_update_tables(
```
jdangerx, line 68: Renamed to be a little more precise as I accidentally ran this when I ran `pytest ... -k test_dbt`

src/pudl/scripts/dbt_helper.py
```diff
@@ -499,25 +500,53 @@ def update_tables(
...
     """Script for auto-generating dbt configuration and migrating existing tests.
 
-    This CLI currently provides one sub-command: ``update-tables`` which can update or
-    create a dbt table (model) schema.yml file under the ``dbt/models`` repo. These
-    configuration files tell dbt about the structure of the table and what data tests
-    are specified for it. It also adds a (required) row count test by default. The
-    script can also generate or update the expected row counts for existing tables,
-    assuming they have been materialized to parquet files and are sitting in your
-    $PUDL_OUT directory.
+    This CLI currently provides the following sub-commands:
```
jdangerx, line 544: It sort of seems like this "here are the two sub-commands" documentation gets repeated a few times and it might be nice to reuse somehow.

src/pudl/scripts/dbt_helper.py
```diff
@@ -499,25 +500,53 @@ def update_tables(
...
+    "--target",
+    default="etl-full",
+    type=click.Choice(["etl-full", "etl-fast"]),
+    help="DBT target - etl-full or etl-fast.",
+)
+def validate(select: str = "*", target: str = "etl-full") -> None:
+    """Validate a selection of DBT models / sources.
+
+    Wraps the ``dbt build`` command line which doesn't give you the actual data
+    that was returned from the test query.
```
krivard:
```suggestion
    Wraps the ``dbt build`` command line so we can annotate the result with the 
    actual data that was returned from the test query.
```

src/pudl/workspace/setup.py
```diff
@@ -13,6 +13,8 @@
 
 PotentialDirectoryPath = DirectoryPath | NewPath
 
+PUDL_ROOT_PATH = Path(__file__).parent.parent.parent.parent
```
krivard, line 16: oh heyyyyyyyy i may lift this for use in the metadata wizard, ty 🏅

src/pudl/scripts/dbt_helper.py
```diff
@@ -499,25 +500,53 @@ def update_tables(
...
 
+@click.command()
+@click.option(
+    "--select", default="*", help="DBT selector for the asset(s) you want to validate."
+)
+@click.option(
+    "--target",
+    default="etl-full",
+    type=click.Choice(["etl-full", "etl-fast"]),
+    help="DBT target - etl-full or etl-fast.",
```
krivard:
```suggestion
    help="DBT target - etl-full (default) or etl-fast.",
```

docs/release_notes.rst
```diff
@@ -36,6 +36,10 @@ VCE RARE
 Quality of Life Improvements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+* We've added a new sub-command to ``dbt_helper`` - ``dbt_helper validate``.
+  This lets you run validation tests for a selection of DBT models and also
```
krivard, line 40: It looks like "models" and "nodes" are used kinda interchangeably in this PR. Do we need to worry about situations where those two concepts are distinct, or does building a mental model where they are similar enough to stand in for one another get us valuable flexibility?

krivard (commented):
driveby!

this looks great! some docs nits and question about how picky we should be or not-be about vocabulary

docs/release_notes.rst
jdangerx: Ah. Yes. We should say "node" because Sources are not Models but both are Nodes. Good catch.

src/pudl/scripts/dbt_helper.py
zschira: Yeah I think consolidating the docs, especially as this is changing frequently, would be really nice. Not sure the best place to do that though, as it's useful here to show up in `--help` and in the readme.

test/integration/dbt_test.py
```diff
@@ -51,40 +48,16 @@ def test_dbt(
     See https://docs.getdbt.com/reference/programmatic-invocations/ for more details on
     how to invoke dbt programmatically.
     """
+    test_result = build_with_context(model_selection="*", dbt_target=dbt_target)
```
zschira: It's nice to be able to use this here.

dbt/README.md
```diff
@@ -62,6 +62,20 @@ To see all options for this command run:
...
+#### `validate`
+
+If you want to check if a materialized asset passes the validation tests defined in DBT, you can use `dbt_helper validate`. You can pass in the `--select` parameter to filter to a specific set of validation tests, and `--target` to set the DBT target.
+
+This will run the associated tests, then print out the test query output so you can see what went wrong.
+
+Example usage:
+
+```bash
+dbt_helper validate --select "source:pudl_dbt.pudl.out_eia__yearly_generators"
```
zschira, line 74: Not blocking, just musing: I wonder if it would make sense to abstract away the `dbt` select syntax here, given that it's kinda verbose/hard to remember, and in practice I'm not sure if we're typically using the full functionality, or just trying to run all tests for a table.

It could be nice to just pass a list of tables to run tests for, and the command could figure out how to execute all the source tests, and maybe any intermediate models and tests on those models that depend on the source tables.

jdangerx:
Hm, integration tests failing with:

```
E           AssertionError: failure contexts:                                                                                                                                                             
E           test_row_counts_extra_partition:
E                                                 
E              observed_mismatch_count  expected_mismatch_count                                      
E           0                        0                        1                                      
E           =====                                 
E           test_row_counts_missing_partition:
E                                                 
E              observed_mismatch_count  expected_mismatch_count                                      
E           0                        0                        1                                      
E           =====                                 
E           test_row_counts_wrong_partition:
E                                                 
E              observed_mismatch_count  expected_mismatch_count                                      
E           0                        0                        1     
```

It's the unit tests for the row_counts test that are failing. Confusing but I can debug on Monday - next steps are probably to try running *just* that test in the test_dbt integration test by tweaking the selection - and trying with/without `--live-dbs` to see how that goes...

dbt/README.md
jdangerx: I agree, there's machinery in dagster-dbt to create a mapping between dagster assets + dbt nodes based on the manifest.json - we'd need to customize it a bit but should be OK. I figured it would be a good follow-up PR.

The steps would be something like:

* make a mapping of dagster asset -> dbt node
* take a *dagster*-style asset selection string instead of a dbt selection string ("+out_eia__yearly_generators" or something)
* resolve the asset selection via dagster
* then figure out what dbt nodes those assets correspond to and pass *that* to the `dbt build --select` flag

Doesn't seem too crazy and would be a nice QOL improvement since Dagster is sort of the core mental model for our ETL.

e-belfer: +1, this kind of "generate all assets downstream of this asset/group of assets" and "validate all those assets" workflow would go a long way in speeding up testing and development on the data integration side.

src/pudl/scripts/dbt_helper.py
```diff
@@ -499,25 +500,53 @@ def update_tables(
...
+    type=click.Choice(["etl-full", "etl-fast"]),
+    help="DBT target - etl-full (default) or etl-fast.",
+)
+def validate(select: str = "*", target: str = "etl-full") -> None:
+    """Validate a selection of DBT nodes.
+
+    Wraps the ``dbt build`` command line so we can annotate the result with the
+    actual data that was returned from the test query.
+    """
+    test_result = build_with_context(model_selection=select, dbt_target=target)
```
e-belfer: Playing around with this on my branch, this wants to be `node_selection` not `model_selection`

jdangerx: ugh. I thought I did "rename-symbol" via the LSP but I guess it didn't catch this symbol somehow. wack

PR #4239, opened 2025-05-05
Integrate 2014-2018 VCE RARE data

e-belfer:
# Overview

Closes #4226, #4008 and #4009. Possibly closes #3914.

## What problem does this address?
- Add 2014-2018 VCE RARE data to PUDL
- Tackle some loose ends from the last release

Also, right now there's a bug with West Virginia state names!
`wetzel_west_virginia` becomes `wetzel_west` and `virginia` when this code runs:

```
.assign(
            place_name=lambda x: x.county_state_names.str.extract(
                rf"([a-z_]+)_({state_pattern})$"
            )[0].astype("category")
        )
```

## What did you change?
- Add 2014-2018 VCE RARE data to PUDL
- Make lat and lon dataset specific
- Rename `county_or_lake_name` to `place_name`
- Update dbt tests
- Fix bug with West Virginia state name split
- Handle nulls in 2015 data
- Handle values in excess of 110% for solar capacity in 2016

Out of scope: See https://github.com/catalyst-cooperative/pudl/tree/vce-rare-census
- Make the county names conform to corresponding FIPS codes from the latest Census PEP data
- Undo the "st" to "saint" lake transformation, as this abbreviation is now widely used in all place names due to the use of census standard names

## Documentation

Make sure to update relevant aspects of the documentation:

- [x] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html): reference the PR and related issues.
- [x] Update relevant Data Source jinja templates (see `docs/data_sources/templates`).
- [x] Update relevant table or source description metadata (see `src/metadata`).
- [x] Review and update any other aspects of the documentation that might be affected by this PR.

# Testing

How did you make sure this worked? How can a reviewer verify this?

## To-do list

- [x] If updating analyses or data processing functions: make sure to update row count expectations in `dbt` tests.
- [ ] Run `make pytest-coverage` locally to ensure that the merge queue will accept your PR.
- [x] Review the PR yourself and call out any questions or issues you have.
- [x] For minor ETL changes or data additions, once `make pytest-coverage` passes, make sure you have a fresh full PUDL DB downloaded locally, materialize new/changed assets and all their downstream assets and [run relevant data validation tests](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `pytest` and `--live-dbs`.
[...]

PR #4319, opened 2025-06-11
Map VCE RARE `place_name` to latest Census PEP vintage

### LINKED ISSUE #3914: VCE RARE post-release loose ends
# Overview

Issues that we need to address but which didn't make it into the first released version of the VCE RARE data.

```[tasklist]
* [ ] Refactor transform to reduce peak memory use so we can use the 8CPU/64GB VM in nightly builds
* [ ] Fix the expected row-count asset check so that it can pass in the fast ETL
* [ ] Fix the spelling of "Lake Hurron" and tell VCE about it so they can fix it upstream.
* [ ] Expand "Lake St. Claire" to "Lake Saint Clare" since every other "Saint" is spelled out.
```

e-belfer:
# Overview

Closes #3914 

In #4239, we ran into memory problems and deprioritized fixing the last loose end of #3914. Namely, that the place names don't 100% correspond to the Census PEP vintage data we use in PUDL - see for example, FIPS ID 46102 with place name Shannon County, rather than Oglala Lakota.

## What did you change?

- Take the latest year of Census PEP data, and use it to override the `place_name` value in `out_vcerare__hourly_available_capacity_factor`.
- Update docs
- Add a dbt test to ensure this county in particular is renamed as expected

## Documentation

# Testing

## To-do list

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -30,7 +30,8 @@ Bug Fixes
 VCE RARE
 ~~~~~~~~
 * Standardized ``place_name`` using data from the latest Census PEP vintage,
-  found in ``_core_censuspep__yearly_geocodes``. See :issue:`3914` and PR :pr:`4319`.
+  found in ``_core_censuspep__yearly_geocodes``. See issue :issue:`3914` and PR
+  :pr:`4319`.
 
 Quality of Life Improvements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/pudl/transform/vcerare.py
+++ b/src/pudl/transform/vcerare.py
@@ -300,7 +300,10 @@ def _standardize_census_names(vce_fips_df: pd.DataFrame, census_pep_data: pd.Dat
         ["place_name", "area_name"],
     ].rename(columns={"place_name": "vce_place_name", "area_name": "census_place_name"})
     logger.debug(f"Updating the following place names:\n{log_df}")
-    assert len(log_df) <= 74
+    # Identified 74 replacements in 2025-06, expect this shouldn't change much.
+    # If it does, manually inspect the debug log above and make sure name changes
+    # are reasonable and expected.
+    assert len(log_df) <= 74, f"Expected 74 replacements, found {len(log_df)}"
 
     names_df = (
         names_df.drop(columns=["place_name", "state_census"])
@@ -430,8 +433,9 @@ def one_year_hourly_available_capacity_factor(
         f"Transforming the VCE RARE hourly available capacity factor tables for {year}."
     )
     # Clean up the FIPS table and update state_county names to match Census data
-    fips_df = _prep_lat_long_fips_df(raw_vcerare__lat_lon_fips)
-    fips_df_census = _standardize_census_names(fips_df, census_pep_data)
+    fips_df_census = _prep_lat_long_fips_df(raw_vcerare__lat_lon_fips).pipe(
+        _standardize_census_names, census_pep_data
+    )
 
     # Apply the same transforms to all the capacity factor tables. This is slower
     # than doing it to a concatenated table but less memory intensive because
--- a/test/unit/transform/vcerare_test.py
+++ b/test/unit/transform/vcerare_test.py
@@ -34,7 +34,7 @@ def test_standardize_census_names():
                 "example_bad_arizona",
                 "incorrectly_named_west_virginia",
                 "real_fake_idaho",
-                "another_massachusetts",
+                "another_massachusetts",  # Check what happens when there's no corresponding Census data
             ],
             "lat county": [1.1, 1.2, 1.2, 1.3, 1.4, 1.5],
             "long county": [0.0, 0.1, 0.1, 0.2, 0.3, 0.4],


docs/templates/vcerare_child.rst.jinja
```diff
@@ -60,13 +60,8 @@ areas as well).
...
 * The place name for Lake Huron is misspelled "Hurron" in the original data.
-* The place name for Lake Saint Claire in the original data uses the abbreviation "St."
-  including the period character -- this is the only abbreviation and only non
-  alphabetical or underscore character in any of the place names.
-* In 2015 Shannon County, South Dakota was renamed Oglala Lakota County and assigned a
-  new FIPS code. The VCE data includes the new FIPS code (46102) but the old name, while
-  the Census DP1 dataset included with PUDL is from 2010, and so uses both the old name
-  and the old FIPS code (46113).
+* We update the ``place_name`` column using the latest Census PEP vintage data, to
+  ensure consistency of place names between this dataset and other PUDL data.
```
krivard, line 64: How often do we update the Census PEP vintage we're using? is that process automated or manual?

//mostly worried this description will go out of date

src/pudl/transform/vcerare.py
```diff
@@ -230,6 +231,92 @@ def _check_for_valid_counties(
...
+    )  # Clean up the place name
+
+    # Log differences, but only show ones that don't have a difference of "_"
+    # to make this actually informative when debugging
+    log_df = names_df.loc[
+        names_df.place_name.str.replace("_", " ") != names_df.area_name,
+        ["place_name", "area_name"],
+    ].rename(columns={"place_name": "vce_place_name", "area_name": "census_place_name"})
+    logger.debug(f"Updating the following place names:\n{log_df}")
+    assert len(log_df) <= 74
```
krivard: is this assert a debugging leftover or will it stay in? if it stays, add a comment saying where the number came from and how to decide if it needs to be updated

src/pudl/transform/vcerare.py
```diff
@@ -230,6 +231,92 @@ def _check_for_valid_counties(
...
+
+    names_df = (
+        names_df.drop(columns=["place_name", "state_census"])
+        .rename(
+            columns={
+                "area_name": "place_name",
+            }
+            # This gets converted to a string in the merge, so we reconvert to categorical
```
krivard, line 314: do we need to worry about the new categorical matching the old one in any way? are the category ids from the old df kicking around anywhere else?

src/pudl/transform/vcerare.py
```diff
@@ -300,7 +389,9 @@ def _combine_cap_fac_with_fips_df(
     )
     combined_df = pd.merge(
         cap_fac_df, fips_df, on="county_state_names", how="left", validate="m:1"
-    )
+    ).drop(columns="county_state_names")
+    # We need the _vcerare columns earlier when we check that no counties are missing, but
+    # at this point they are extraneous
```
krivard, line 397: 👍 good clarifier

src/pudl/transform/vcerare.py
```diff
@@ -340,8 +429,10 @@ def one_year_hourly_available_capacity_factor(
     logger.info(
         f"Transforming the VCE RARE hourly available capacity factor tables for {year}."
     )
-    # Clean up the FIPS table
+    # Clean up the FIPS table and update state_county names to match Census data
     fips_df = _prep_lat_long_fips_df(raw_vcerare__lat_lon_fips)
+    fips_df_census = _standardize_census_names(fips_df, census_pep_data)
```
krivard: do we need both dfs?

```suggestion
    fips_df = _standardize_census_names(fips_df, census_pep_data)
```

src/pudl/transform/vcerare.py
```diff
@@ -230,6 +231,92 @@ def _check_for_valid_counties(
...
+    ]
+    names_df = pd.concat([names_df, lakes_and_cities])
+
+    # Where there is no county data, fill in with VCE data
+    names_df["area_name"] = names_df["area_name"].fillna(
+        names_df["place_name"].astype(str)
+    )
+    names_df["area_name"] = names_df["area_name"].str.replace(
+        "_", " "
+    )  # Clean up the place name
```
krivard, line 294: for the record this (separate, merge, concat, fill nulls, fix spelling) is a really clean and clever way to solve the problem, nice work!

src/pudl/transform/vcerare.py
```diff
@@ -230,6 +231,92 @@ def _check_for_valid_counties(
...
+        .str.replace("parish", "")
+        .str.strip()
+    )
+
+    # Drop lakes and two cities we're going to remove from our dataset later
+    vce_fips_df_sub = vce_fips_df.loc[
+        ~vce_fips_df.county_state_names.isin(
+            ["bedford_city_virginia", "clifton_forge_city_virginia"]
+        )
+    ].dropna(subset="county_id_fips")
```
krivard, line 271: the back of my brain wants there to be a nice way to specify this as a single selector so we only have to compute it once, and just invert it when we want to pull out the opposite set below, but i haven't been able to think of one so i think the back of my brain is out of luck

docs/release_notes.rst
```diff
@@ -27,6 +27,11 @@ Expanded Data Coverage
 Bug Fixes
 ^^^^^^^^^
 
+VCE RARE
+~~~~~~~~
+* Standardized ``place_name`` using data from the latest Census PEP vintage,
+  found in ``_core_censuspep__yearly_geocodes``. See :issue:`3914` and PR :pr:`4319`.
```
krivard: optional
```suggestion
  found in ``_core_censuspep__yearly_geocodes``. See issue :issue:`3914` and PR :pr:`4319`.
```

src/pudl/transform/vcerare.py
```diff
@@ -389,6 +482,23 @@ def out_vcerare__hourly_available_capacity_factor(
     def _get_year(df, year):
         return df.loc[df["report_year"] == year]
 
+    # Get census vintage to conform the data to.
+    assert int(_core_censuspep__yearly_geocodes.report_year.max()) >= int(
+        raw_vcerare__fixed_solar_pv_lat_upv["report_year"].max()
+    )  # Check these are in sync
```
krivard, line 492: if this fails, does that mean we need to update our copy of census pep? or would something else be wrong?

src/pudl/transform/vcerare.py
```diff
@@ -389,6 +482,23 @@ def out_vcerare__hourly_available_capacity_factor(
...
+
+    # Only keep latest census data relating to the state-county level
+    # See: https://www.census.gov/programs-surveys/geography/technical-documentation/naming-convention/cartographic-boundary-file/carto-boundary-summary-level.html
```
krivard, line 495: 👍 yes good reference

test/unit/transform/vcerare_test.py
```diff
@@ -0,0 +1,80 @@
...
+    vce_fips_test_df = pd.DataFrame(
+        {
+            "FIPS": ["01001", "04002", "04003", "54003", "16004", "25005"],
+            "county_state_names": [
+                "saint_johns_alabama",
+                "saint_johns_arizona",
+                "example_bad_arizona",
+                "incorrectly_named_west_virginia",
+                "real_fake_idaho",
+                "another_massachusetts",
```
krivard: I love the line annotations on the census test df, maybe do similar here?

docs/templates/vcerare_child.rst.jinja
e-belfer: Should be annual, the same cadence as VCE RARE.

src/pudl/transform/vcerare.py
e-belfer: Mostly trying to prevent runaway code renames if anything gets weird, but will add some clarifying comments here.

src/pudl/transform/vcerare.py
e-belfer: The column they are in gets dropped, and these don't need to match - we're filling from this column where we don't have census data, otherwise we want to override these values.

src/pudl/transform/vcerare.py
e-belfer: Ah no, this is a holdover from when I was briefly using both. Good catch, I'll streamline.

src/pudl/transform/vcerare.py
e-belfer: Yes, or that the census data is newer than the VCE RARE data. I think if these get out of sync in a non-fixable way we'll want to make a call about what to do, but I'd like to leave the assertion as is for now.

test/unit/transform/vcerare_test.py
e-belfer: Done for the last value!

PR #4332, opened 2025-06-19
Integrate 2025 Q1 EIA 860M and 860 and 923 2024 early release data

### LINKED ISSUE #4323: Integrate EIA 860 2024 Early Release data

Based on the [Existing Data Updates Docs](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html)

- [x] [Obtain fresh data](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#obtain-fresh-data)
- [x] [Map the structure of the new data](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#map-the-structure-of-the-new-data)
- [x] [Test data extraction](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#test-data-extraction)
- [x] [Update table and column transformations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#update-table-column-transformations)
- [x] [Update the PUDL db schema](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#update-the-pudl-db-schema)
- [x] [Connect datasets](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#connect-datasets)
- [ ] [Update the output routines](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#update-the-output-routines-and-run-full-tests)
- [ ] [Run the ETL](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#run-the-etl)
- [ ] [Run and update data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#run-and-update-data-validations)
- [ ] [Update the documentation](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#update-the-documentation)

Hijinks:

- Cranberry Point Energy Storage LLC (Plant ID 62844) has a literal value for the proposed storage unit generator ID of "NA". This is true in 2023 and 2024. I updated the "expect to drop" number to 2 and left a note about which plant ID this is.
- Fixed one bad `planned_energy_source_code` value.

e-belfer:
# Overview

Closes #4323, #4324, #4325.

- Bring 2024 early release data in for EIA 923 and 860
- Bring in April 860M data
- Bring in March 2025 EIA 923 data

## What did you change?
- Handled a large increase in the number of bunk generator IDs harvested from the direct support table by migrating to dropping all rows with entirely null generator values, and raising an error if any of these are not in an expected list to avoid silently/unexpectedly dropping gens.
- Matched a known set of direct_support generator IDs to their actual generator IDs by capitalizing them - the differences in capitalization were marking a larger number of generators as non-existent than are actually the case.
- Updated settings, DOIs, working partitions, and expected row counts
- Mapped new EIA plants
- Handled test failures (see below)

## Test failures

Failure in test source_expect_quantile_constraints_pudl_out_eia__monthly_generators_unit_heat_rate_mmbtu_per_mwh___quantile_0_5_min_value_7_0_max_value_7_5____quantile_0_05_min_value_6_4____quantile_0_95_max_value_13_0___fuel_type_code_pudl_gas_and_report_date_CAST_2015_01_01_AS_DATE___net_generation_mwh (models/output/out_eia__monthly_generators/schema.yml)

* 0.05 quantile is not met, dropped from 6.4 to 6.39

Failure in test source_expect_complete_valid_ownership_pudl_out_eia860__yearly_ownership_3 (models/eia860/out_eia860__yearly_ownership/schema.yml)

```
report_date │ plant_id_eia │ generator_id │ fraction_owned │ toy_join │      pct_missing      │
│    date     │    int32     │   varchar    │     double     │  int32   │        double         │
├─────────────┼──────────────┼──────────────┼────────────────┼──────────┼───────────────────────┤
│ 2024-01-01  │        56032 │ 1            │            2.0 │        1 │ 0.0019389238972370335 │
│ 2022-01-01  │          341 │ CT5          │            1.5 │        1 │ 0.0019389238972370335 │
│ 2023-01-01  │          341 │ CT5          │            1.5 │        1 │ 0.0019389238972370335 │
│ 2024-01-01  │          341 │ CT5          │            1.5 │        1 │ 0.0019389238972370335 │
│ 2021-01-01  │          341 │ CT5          │            1.5 │        1 │ 0.0019389238972370335
```

Known failures as of June 2025:
* 341 CT5 has 1.5 total ownership fraction for 2021, 2022, 2023, 2024 (one new year of the same error already present)
* 56032 generator 1 has a 2.0 total ownership fraction in 2024 (new error)

## Documentation

# Testing

## To-do list

- [x] Debug excessively null rows failure in fast ETL only in MCOE
- [x] Check foreign keys
- [x] Run `dbt build` on full ETL
- [x] Run the `build-deploy-pudl` GitHub Action manually. (https://console.cloud.google.com/storage/browser/builds.catalyst.coop/2025-06-24-0100-a63a8986d-860-860m-q-update)


docs/dev/existing_data_updates.rst
```diff
@@ -431,6 +431,13 @@ settings file.
 **8.2)** Once the fast ETL works, run the full ETL using the ``etl_full.yml`` settings
 to populate complete FERC 1 & PUDL DBs and EPA CEMS Parquet files.
 
+* Run crashing? Open the launchpad for the job.
+* Note that running the full ETL can take many hours on a standard laptop. If you are
+  only updating one data source, consider downloading the latest nightly build outputs
```
cmgosnell: this can be oos for this pr but it would be good to enumerate how. point to [data access page](https://catalystcoop-pudl.readthedocs.io/en/latest/data_access.html#cloud-storage)?... but honestly a think i've been confused about downloading all of the nightly builds is if you need to download all the parquet files and the sqlite for just the former. it'd just be good to have it spelled out specifically for consistent development.

dbt/models/output/out_eia__monthly_generators/schema.yml
```diff
@@ -102,11 +102,11 @@ sources:
                       min_value: 7.0
                       max_value: 7.5
                     - quantile: 0.05
-                      min_value: 6.4
+                      min_value: 6.39
                     - quantile: 0.95
                       max_value: 13.0
                   weight_column: net_generation_mwh
-                  description: Historical note, EIA natural gas reporting really only becomes usable in 2015.
+                  description: Quantile 0.05 dropped from 1.4 to 1.39 in June 2025. Keep an eye on it, and investigate if it gets worse. Historical note, EIA natural gas reporting really only becomes usable in 2015.
```
cmgosnell: just getting this straight bc i haven't really seen these tests at all yet... this means the heat rate for the least efficient gas plants dipped slightly. which is to say with this new round of data had plants that were even less efficient.

also i think your note here should be `dropped from 6.4 to 6.39`? unless im misunderstanding something

src/pudl/helpers.py
```diff
@@ -1602,11 +1602,13 @@ def drop_all_null_records_with_multiindex(
     """
     # ensure there isn't more than the expected number of nulls before dropping
     df = df.set_index(idx_cols)
-    assert df.loc[idx_records].isnull().all().all(), (
-        "Non-null data found where no data was expected:",
-        f"{df.loc[idx_records].dropna(axis='columns', how='all')}",
-    )  # Make sure all values in all rows and columns here are null
-    return df.drop(idx_records).reset_index()
+    if any(record in df.index for record in idx_records):  # Handle fast ETL
```
cmgosnell: nit! (it just took me a second to understand)
```suggestion
    # Skip if df has no idx_records - for fast ETL.
    if any(record in df.index for record in idx_records):  
```

src/pudl/output/eia.py
```diff
@@ -5,7 +5,6 @@
 from dagster import Field, asset
 
 import pudl
-from pudl.helpers import drop_all_null_records_with_multiindex
```
cmgosnell, line 8: hm are we no longer using this helper function at all? maybe we should delete it altogether.

src/pudl/output/eia.py
```diff
@@ -145,20 +144,71 @@ def _out_eia__yearly_generators(
...
+    ]  # Columns that will have values
+    all_null_gens = out_df[
+        out_df[[col for col in out_df.columns if col not in non_null_cols]]
+        .isnull()
+        .all(axis=1)
+    ]
+    all_null_gens_ids = [
+        (int(row.plant_id_eia), row.generator_id) for row in all_null_gens.itertuples()
+    ]
+    unexpected_gens = set(all_null_gens_ids) - set(empty_generator_ids)
```
cmgosnell: non-blocking but you could make `empty_generator_ids` a pd.Index and use the `all_null_gens.index.difference(empty_generator_ids)` so you don't have to do the itertuples situation. i think it would be easier to read but totally not blocking this looks like it does what you want!

src/pudl/output/eia.py
```diff
@@ -145,20 +144,71 @@ def _out_eia__yearly_generators(
...
+        (9170, "3093"),
+        (34516, "SOL1"),
+        (56951, "VESTA"),
+        (57991, "PV2"),
+        (58644, "ALL"),
+        (62355, "2WPSO"),
+        (62760, "SONRI"),
+        (62975, "SYNLB"),
+        (63210, "SAINT"),
+        (69312, "EcheB"),
```
cmgosnell: should this guy be removed? bc its weird cased and it's CAPITALIZED friend is down below?

src/pudl/transform/eia860.py
```diff
@@ -416,6 +416,16 @@ def _core_eia860__generators_energy_storage(
...
+    # capitalize the direct_support generator IDs
+    gen_support_columns = [
+        col
+        for col in storage_df.columns
+        if col.startswith("generator_id_direct_support_")
+    ]
+    storage_df.loc[:, gen_support_columns] = storage_df.loc[
+        :, gen_support_columns
+    ].apply(lambda x: x.str.upper())
+
```
cmgosnell: oh my... are they uncapitalized in their raw form? this is a yellow flag bc imu we generally do our darndest not to touch the generator IDs bc EIA generally does a good job of not being dinguses with their IDs.

At the very least I think you should add more documentation here about why your doing this (even just copy past what you noted in the pr comment)

src/pudl/output/eia.py
e-belfer: Yes, let's do that.

src/pudl/transform/eia860.py
e-belfer: The ones that were causing problems were all title case, whereas the data coming directly from the generator ID reporting for the same generator was all-caps.

docs/dev/existing_data_updates.rst
e-belfer: I'm going to just remove this, actually. I think we owe this entire doc section a once-over and an update, and enough other things are in flight that I will hold off on this minor update. But yes, you need both SQLite and Parquet outputs from the builds.

dbt/models/output/out_eia__monthly_generators/schema.yml
e-belfer: Yes, exactly that - got the numbers wrong and will update. If this merits more serious investigation I could also write up an issue specifically to look into this.

src/pudl/output/eia.py
e-belfer: Ooh, yes good catch!

src/pudl/transform/eia860.py
e-belfer: Manually reviewing the generator IDs in `core_eia__entity_generators`, there are a very small number of cases where they are not title-case, and this is true up through the present. So yes, it'd be incorrect to assume that all directly reported generator IDs are upper-case. Thoughts on what makes sense to do here?

['West',
 'East',
 'Kato',
 'House',
 'unk',
 'South',
 'Kenec',
 'GenB',
 'GenA',
 'ge']

e-belfer: The alternative is that these IDs get added to my drop-list and not harvested into the core list, which maybe seems preferable?

zaneselvans: The generator IDs show up in other datasets too so if we touch them, it'll probably cause problems with anyone trying to link them. The only case in which we've altered reported generator IDs in the past is when a single generator has a varying number of leading zeroes in its IDs across years or tables. Like  we treat `0001` and `01` and `1` as the same ID. But there are only like 10 of these, and if we don't do it, then those generator records can't be linked within the EIA data.

dbt/seeds/etl_fast_row_counts.csv
cmgosnell, line 1: wait are we updating the fast-ies at all right now? i think we are not running them at all. not bad to update them or anything...

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -2618,14 +2649,14 @@ out_eia__yearly_assn_plant_parts_plant_gen,2014,374328
...
-out_eia__yearly_assn_plant_parts_plant_gen,2024,492294
-out_eia__yearly_assn_plant_parts_plant_gen,2025,498246
+out_eia__yearly_assn_plant_parts_plant_gen,2018,421549
+out_eia__yearly_assn_plant_parts_plant_gen,2019,436454
+out_eia__yearly_assn_plant_parts_plant_gen,2020,450106
+out_eia__yearly_assn_plant_parts_plant_gen,2021,465441
+out_eia__yearly_assn_plant_parts_plant_gen,2022,479997
+out_eia__yearly_assn_plant_parts_plant_gen,2023,494226
+out_eia__yearly_assn_plant_parts_plant_gen,2024,509681
+out_eia__yearly_assn_plant_parts_plant_gen,2025,499520
```
cmgosnell, line 2410: you think this is from the capitalization maybe? bc all the gens get fed into this

dbt/seeds/etl_full_row_counts.csv
cmgosnell, line 1: i'm a little surprised that there are lower counts for some of the 2023 tables. it looks like it originates from  `core_eia860__scd_generators`. you think its the capitalization changing? or is there something else that happened that i'm missing that would cause that?

src/pudl/transform/eia860.py
cmgosnell: we decided to manually fix a small handful of these support generators that have the same ID as the `generator_id` except for capitalization.

cmgosnell (approved):
i know you still need to re-check the row counts with the changes to the capitalization but im going to approve this pending those checks.

it'd be nice to know why the 2023 generators when down (at least partially due to capitalization)

dbt/seeds/etl_full_row_counts.csv
e-belfer: These changes have now been reverted, seems like the capitalization was causing some kind of problem and it's good we dropped it!

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -3150,11 +3171,11 @@ out_ferc1__yearly_purchased_power_and_exchanges_sched326,2021,6899
 out_ferc1__yearly_purchased_power_and_exchanges_sched326,2022,7052
 out_ferc1__yearly_purchased_power_and_exchanges_sched326,2023,7066
 out_ferc1__yearly_purchased_power_and_exchanges_sched326,2024,7065
-out_ferc1__yearly_rate_base,1994,14707
-out_ferc1__yearly_rate_base,1995,14387
-out_ferc1__yearly_rate_base,1996,14768
-out_ferc1__yearly_rate_base,1997,14515
-out_ferc1__yearly_rate_base,1998,14298
+out_ferc1__yearly_rate_base,1994,14718
```
e-belfer, line 3174: I'm confused about these changes, but they are coming directly from the nightly builds. Is it possible this table has non-stochastic behaviors? This seems to be more or less but not 100% reverting to the row counts prior to https://github.com/catalyst-cooperative/pudl/commit/7a0ee031da2f2e1b55f10e72e2e2a08b990a8168#diff-626b2ee249fc60a4b456621e31860291ef7296bf2fbec3c617f184e3ec93cd02

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -2115,7 +2131,8 @@ out_eia923__monthly_generation_fuel_by_generator,2019,294289
 out_eia923__monthly_generation_fuel_by_generator,2020,302100
 out_eia923__monthly_generation_fuel_by_generator,2021,317800
 out_eia923__monthly_generation_fuel_by_generator,2022,327729
-out_eia923__monthly_generation_fuel_by_generator,2023,337632
+out_eia923__monthly_generation_fuel_by_generator,2023,337536
```
e-belfer, line 2134: Everything that is downstream of `_out_eia__yearly_generators()` will see some modest drop in number here due to us getting rid of the null generators.

dbt/seeds/etl_full_row_counts.csv
cmgosnell: (we chatted about this already but for 👀 ) this table is no longer being row-count checked because it seems like this tables row counts seem unstable.

dbt/seeds/etl_full_row_counts.csv
e-belfer: The data in 2018 contains proposed plant parts, which are still changing. The old records have a handful of plant parts with NaN generator_operating_year_plant_gen. These records now have updated `generator_operating_year_plant_gen` years, which add new plant parts.

In 2018, as an example:
* we gain two rows `62313_2023_2018_plant_operating_year_owned_61834_proposed` and `62313_2023_2018_plant_operating_year_total_61834_proposed` where the plant gets an operating date in the new data.
* Same for 62520 and 62523, which get an operating date of 2025

PR #4335, opened 2025-06-19
chore: fix 'legacy alias' for ruff pre-commit check

jdangerx:
## Overview

We're supposed to use `ruff-check` instead of `ruff` now, I guess. Also, dropping the `--exit-non-zero-on-fix` didn't seem to change the behavior when there was a linting error that got auto-fixed. So we might as well remove it.

# Testing

I added an extra newline to the import block of `transform/eia930.py` then tried to commit it. `ruff` complained and the commit didn't go through. I had to manually add the fix from `ruff` before re-committing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,8 +31,8 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.11.13
     hooks:
-      - id: ruff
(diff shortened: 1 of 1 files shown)


zaneselvans (approved):

zaneselvans:
Oh this had been annoying me too. Thank you for nabbing it!

PR #4336, opened 2025-06-20
Add extra flag to timeseries imputation for bad years of data

zschira:
# Overview

This PR adds a new imputation flag, `BAD_YEAR`. This flag indicates that an entire year of data was dropped for a specific respondent because there was too much missing data. This should ensure that anytime the `imputed` column isn't equal to the `reported` column, there will be a corresponding `flag` for the value.

### DIFF
--- a/src/pudl/analysis/timeseries_cleaning.py
+++ b/src/pudl/analysis/timeseries_cleaning.py
@@ -1223,9 +1223,59 @@ def flag_anomalous_region(
     return ts.flag(mask, ImputationReasonCodes.ANOMALOUS_REGION)
 
 
+def flag_bad_years(
+    ts: FlaggedTimeseries,
+    min_data: int = 100,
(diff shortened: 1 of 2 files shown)


src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -1530,63 +1586,6 @@ def impute_flagged_values(
     return pd.concat(results)
 
 
-@pa.check_types
```
zschira, line 1533: Previously, this method did two things:
1. Find respondent/year combos that are missing lots of data and drop those years
2. Drop any respondents entirely that don't have any data across all years

I've split this into a new `flag_bad_years` function that does `1.`, plus properly assigns flags to all values in these years, and dropping blank respondents is now just moved into the top level `_flag_timeseries_matrix` function.

Issue #4126, opened 2025-03-14
Add developer documentation covering dbt


zaneselvans:

krivard:
[I've roughed out the places in the documentation that we might need to update](https://docs.google.com/document/d/1Rx4k-_HDG2I2Rb9ogzeYJOtVDy4mQXsMRjSZ2H9NNNg/edit?tab=t.1xxpe8lo0vcz#heading=h.avfl47cb688y) in a new tab of the original data validation design doc. It's kindof a lot, but that's mostly because the old validation tests were also a big part of everyday processes so everything that touched them needs a look.

See if that makes sense and we can start assigning parts out.

jdangerx:
@zaneselvans I've updated the description with the questions I've heard bouncing around. Some questions:

* does it make sense to split the questions into a few chunks so we can parallelize work?
* sounds like people mostly don't read the `dbt/README.md` - should we move the still-relevant content out of it and delete the file?

(both my answers are a light 'yes' but open to being convinced otherwise)

zaneselvans:
I would definitely like to migrate the README content into our docs / RST. That was the first thing I was going to do. Then we'll have a container and can create headers for the topics and split them up. There's a significant amount of content. Some of which we might want to hold off on fleshing out completely until we get the `dbt_helper` fixes merged in.

PR #4341, opened 2025-06-23
Test saving observed dbt row counts in nightly builds

zaneselvans:
# Overview

Copy the observed full ETL row counts to the nightly build outputs so they can be generated and used without needing to to run the Full ETL locally.

## To-do list

- [x] Run a branch deployment. Check that observed row counts are available on on `gs://builds.catalyst.coop`.
- [x] Undo the test change I made to the row counts file.
- [x] Update developer docs to offer this option for grabbing authoritative nightly build row count expectations.

### DIFF
--- a/docs/dev/testing.rst
+++ b/docs/dev/testing.rst
@@ -146,8 +146,17 @@ database:
 The data validation cases that pertain to the contents of the data tables are
 currently stored as part of the :mod:`pudl.validate` module.
 
-The expected number of records in each output table is stored in the validation
-test modules under ``test/validate`` as pytest parameterizations.
+To catch unexpected changes to the data, we keep track of the expected number of rows in
+each data table we distribute. These expectations are stored in
+``dbt/seeds/etl_full_row_counts.csv`` and they can be updated using the ``dbt_helper``
+script. If you can't run the full ETL locally, the nightly builds / branch deployments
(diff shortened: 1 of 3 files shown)


zaneselvans:
Cool!  It seems to work, though when it finds that there are changes to the row counts, that results in a build "failure" but so long as we know that, it does seem to provide the correct outputs.

docs/dev/testing.rst
```diff
@@ -146,8 +146,17 @@ database:
 The data validation cases that pertain to the contents of the data tables are
 currently stored as part of the :mod:`pudl.validate` module.
 
-The expected number of records in each output table is stored in the validation
-test modules under ``test/validate`` as pytest parameterizations.
+To catch unexpected changes to the data, we keep track of the expected number of rows in
```
e-belfer, line 149: This should also get added to the "existing data updates" docs, but if you want to take note of this in the bigger "update dbt docs" issue we could just handle this all at once there.

zaneselvans: I was trying to avoid that bigger revision here -- there's a bunch of stale stuff we need to update for the new dbt situation. Which I think makes sense to tackle with the dbt dev docs issue post ☠️ PudlTabl ☠️

e-belfer: #4126 Tagging here for posterity.

PR #4348, opened 2025-06-25
Migrate `check_date_freq` monthly vs. annual checks into dbt

### LINKED ISSUE #4286: Migrate monthly vs. annual aggregation checks from `test_eia_outputs`
# Overview

- The `test_eia_outputs` output test mostly checked that various tables could be read out of `PudlTabl` and that kind of testing has now been removed.
- However, it also checks that there's the expected ratio between annual time periods and monthly time periods in yearly/monthly tables.
- This test has been useful in the past for catching errors resulting from merging dataframes with different time frequencies together (which is sometimes necessary) improperly
- It seems like it's probably an easy kind of thing to check in SQL, so why not migrate it into `dbt`?
- There are also probably additional tables that should be checked this way now.

zaneselvans:
# Overview

- Migrates the tests from `test/integration/output_test.py::test_eia_outputs` into a `dbt` test.
- Allows us to remove `pudl.validate.check_date_freq`

Closes #4286 

# Testing

To check whether this is working on your local outputs, you can run:

```bash
dbt build --select "test_name:expect_date_frequency_ratio"
```

The pytest version can be run with:
```bash
pytest --live-dbs test/integration/output_test.py::test_eia_outputs
```

## To-do list

- [x] Debug failure on `out_eia923__monthly_generation_fuel_combined`
- [x] Debug failure on `out_eia923__monthly_fuel_receipts_costs`
- [x] Make sure it works [on the fast ETL](https://github.com/catalyst-cooperative/pudl/actions/runs/15868880011/job/44740889830) as well as the full ETL
- [ ] Apply the test to some additional tables?


src/pudl/validate.py
```diff
@@ -76,8 +76,10 @@ def check_date_freq(df1: pd.DataFrame, df2: pd.DataFrame, mult: int) -> None:
         raise ValueError("Missing report_date column in one or both input DataFrames")
 
     # Remove ytd values that mess up ratio assumptions
-    if "data_maturity" in df2:
-        df2 = df2[df2["data_maturity"] != "incremental_ytd"].copy()
+    if "data_maturity" in df1.columns:
+        df1 = df1.loc[df1["data_maturity"] != "incremental_ytd"]
+    if "data_maturity" in df2.columns:
+        df2 = df2.loc[df2["data_maturity"] != "incremental_ytd"]
```
zaneselvans: I made the SQL version check both dataframes since that seems more robust in the event that we're using a different dataframe to compare against at some point, and wanted the python version to be the same for debugging purposes.

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
```diff
@@ -0,0 +1,52 @@
...
+    multiplier,
+    date_column='report_date'
+) %}
+
+WITH model_years AS (
+    SELECT
+        EXTRACT(YEAR FROM {{ date_column }}) as year,
+        COUNT(DISTINCT {{ date_column }}) as date_count
+    FROM {{ model }}
+    WHERE {{ date_column }} IS NOT NULL
```
zaneselvans, line 41: This `IS NOT NULL` check shouldn't really be necessary, since `report_date` is part of the PK and can't be NULL. Maybe remove?  I don't think we would use this check on a dataframe where the date column wasn't part of the PK.

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
```diff
@@ -0,0 +1,52 @@
...
+    {% if 'data_maturity' in adapter.get_columns_in_relation(model) %}
+        AND (data_maturity != 'incremental_ytd' OR data_maturity IS NULL)
```
zaneselvans: Does it make sense to include `OR data_maturity IS NULL` here? If it were `NULL` wouldn't it also not equal `incremental_ytd`?

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
```diff
@@ -0,0 +1,52 @@
...
+    {% endif %}
+    GROUP BY EXTRACT(YEAR FROM {{ date_column }})
```
zaneselvans, line 46: With the `GROUP BY` hanging out down here all alone, how does it know to apply the `COUNT(DISTINCT())` across all values of `date_column` within the group?

zaneselvans (commented):
This gives the right answer on 9 of 11 cases, but two tables are getting unexpected results and I don't know why.

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
zaneselvans: Ah, there's only a single row for each year, which has the distinct date count. And then those are all summed in the totals.

zaneselvans:
There was an issue both with needing to filter out `monthly_update` data as well as `incremental_ytd` data, and then the conditionals I was using to try and read the columns associated with the model didn't work, because we have sources that are stored in parquet, not a database that the DuckDB adapter can read DB metadata from.

zaneselvans:
I think I've fixed the initial issues and this is ready for review.

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
krivard: Even if it's redundant I'd rather keep it -- SQL treats nulls & NAs differently from Python (and different SQLs also treat them differently), and the prospect of tracking down a problem where Python says it's fine but the particular SQL we're using says it's not, sounds no fun at all

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
```diff
@@ -0,0 +1,85 @@
...
+-- - compare_model: The reference model to compare against (typically lower frequency, e.g. annual)
+-- - multiplier: Expected ratio (e.g. 12 for monthly:annual comparison)
+-- - date_column: Date column name (default: 'report_date')
+-- - model_has_data_maturity: Filter provisional data from model (default: true)
+-- - compare_model_has_data_maturity: Filter provisional data from compare_model (default: true)
+--
+-- Data Filtering:
+-- - Excludes NULL dates
+-- - Excludes provisional records (data_maturity = 'incremental_ytd' or 'monthly_update')
+--   when the respective has_data_maturity parameter is true
```
krivard, line 24: Same as the other PR; copy/move some of this content to generic_tests/schema.yml so it can be with its friends

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
```diff
@@ -0,0 +1,85 @@
...
+
+{% test expect_date_frequency_ratio(
+    model,
+    compare_model,
+    multiplier,
+    date_column='report_date',
+    model_has_data_maturity=true,
+    compare_model_has_data_maturity=true
```
krivard, line 32: optional: I might rename these as
```suggestion
    exclude_provisional_model_data=true,
    exclude_provisional_compare_model_data=true
```

so it's about specifying what to do rather that what table features are available

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
```diff
@@ -0,0 +1,85 @@
...
+SELECT
+    model_total,
+    compare_total,
+    {{ multiplier }} as expected_multiplier,
+    'Expected ' || (compare_total * {{ multiplier }}) ||
+    ' = ' || compare_total || ' * ' || {{ multiplier }} ||
+    ', but found ' || model_total
+    as failure_reason
+FROM totals
+WHERE model_total != compare_total * {{ multiplier }}
```
krivard, line 83: I like this; iirc elsewhere we test ratios by dividing and so have to check the denominator isn't zero. This is very tidy

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
zaneselvans: Went ahead and copied them both over in the other PR and it's in now!

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
zaneselvans: I'm not sure about this rename. This was a hack around a more elegant solution that we can't use because our tables are stored in Parquet, rather than in DuckDB (or some other DB).

If our data was in a DB, we could use dbt's `adapter` object which has some useful methods for flow control like:

```jinja
{% if 'data_maturity' in adapter.get_columns_in_relation(compare_model) %}
```

I feel like the intention is really to always exclude the provisional data because it creates issues when the most recent year isn't complete. But we can't always do that because not every table has the data maturity.

But then again, it does literally determine whether we exclude the provisional data or not.  But it can only work on some tables.

And then there's also the `early_release` data maturity, which is annual (so we don't want to exclude it) but also provisional (so the rename could be confusing).

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
zaneselvans: I think I went with this because I was assuming / requiring the ratios to be integers.  But I guess it would also work with floats and zeroes.

zaneselvans:
> Sidenote this PR identifies the solution to the two-sided two-model dbt test paradox, the entire reason why margay allowed singular tests in the initial migration 🌟

Ah cool! tbh the syntax here was suggested by an LLM but it made sense and did the right thing, after I got some errors while trying to just use the table name about it not being a defined model.

dbt/tests/data_tests/generic_tests/expect_date_frequency_ratio.sql
krivard: Fair argument. In that case, maybe the parameter description could be something like

> "Tell the test that the data_maturity column exists in the model (default: true). If this column is available, it will be used to ignore incomplete years of data."

PR #4349, opened 2025-06-25
Replace FERC-714 hour consistency test with dbt test.

### LINKED ISSUE #4284: Migrate `test_year_in_ferc714_hourly_planning_area_demand` test
# Overview

- The `test_year_in_ferc714_hourly_planning_area_demand` test checks that the overwhelming majority of `report_date` years and timestamp years in two different columns in the FERC-714 hourly table are the same -- a few hours at the beginning/end of each year end up being associated with different years because of timezones / UTC but this should only be like 0.1% of all records.
- This should be a pretty easy test to do in SQL and would probably be much faster, so why not migrate it to `dbt`?

zaneselvans:
# Overview

- Convert the test that was checking for consistency between the years found in `report_date` and `datetime_utc` for the FERC-714 hourly demand data into a dbt test that uses SQL
- Apply the same test to both the `core_` and `out_` versions of the table.

Closes #4284 

## Testing

I ran all (2) instances of this test with:
```bash
dbt build --select "test_name:expect_consistent_years"
```

I reduced the tolerance in one of the test specifications to where it ought to fail, and ran the tests again, and it did fail.

The old pytest validation can be run (on another branch) with:

```bash
pytest --live-dbs test/validate/service_territory_test.py
```

## To-do list

- [ ] Make sure [the test works on fast](https://github.com/catalyst-cooperative/pudl/actions/runs/15869750010/job/44743441948) as well as full ETL outputs.


dbt/tests/data_tests/generic_tests/expect_temporal_year_consistency.sql
```diff
@@ -0,0 +1,74 @@
...
+-- 1. Extract years from both temporal columns
+-- 2. Count total records and records with mismatched years
+-- 3. Calculate the mismatch ratio
+-- 4. Fail if the ratio exceeds the specified threshold
+--
+-- Parameters:
+-- - model: The model to test
+-- - datetime_column: High-frequency datetime column (e.g., 'datetime_utc')
+-- - date_column: Reporting period date/year column (e.g., 'report_date')
+-- - max_mismatch_ratio: Maximum allowed fraction of mismatched records (default: 0.001 = 0.1%)
```
krivard: Migrate or copy some of this info to `dbt/test/data_tests/generic_tests/schema.yml` so it's consistent with the documentation for the other generic tests

dbt/tests/data_tests/generic_tests/expect_temporal_year_consistency.sql
```diff
@@ -0,0 +1,74 @@
...
+SELECT
+    total_records,
+    mismatched_records,
+    mismatch_ratio,
+    {{ max_mismatch_ratio }} as max_allowed_ratio,
+    'Found ' || ROUND(mismatch_ratio * 100, 5) || '% (' || mismatched_records ||
+    ' of ' || total_records || ') mismatched years between ' ||
+    '{{ datetime_column }}' || ' and ' || '{{ date_column }}' ||
+    ', exceeds threshold of ' || ROUND({{ max_mismatch_ratio }} * 100, 5) || '%'
+    as failure_reason
```
krivard: nice!

dbt/tests/data_tests/generic_tests/expect_temporal_year_consistency.sql
```diff
@@ -0,0 +1,74 @@
+-- Generic test: expect_temporal_year_consistency
```
krivard:
```suggestion
-- Generic test: expect_consistent_years
```

Also: should this file be named `expect_consistent_years.sql` to match the name of the test macro? that would make it easier to find in the filesystem.

krivard (changes requested):
Two fixes: 
- Looks like the name of the macro changed late in the game and wasn't updated everywhere
- Missing macro documentation in `generic_tests/schema.yml` - this can either be moved or copied from what you already have

dbt/tests/data_tests/generic_tests/expect_temporal_year_consistency.sql
zaneselvans: Ah whoops.  I didn't know that's where documentation was happening. I'll translate it over. Does this end up getting displayed anywhere?

dbt/tests/data_tests/generic_tests/expect_temporal_year_consistency.sql
zaneselvans: Ah whoops yeah I renamed it late and did not search-and-replace effectively.

dbt/tests/data_tests/generic_tests/expect_temporal_year_consistency.sql
zaneselvans: This is all from an LLM -- just cribbing patterns from other examples of dbt tests on the interwebs. But great to make use of those patterns!

dbt/tests/data_tests/generic_tests/expect_temporal_year_consistency.sql
krivard: I don't think we're using dbt's documentation builders yet, but if we ever do, the schema.yml is where it will look for macro docs. it'll also double check it can find all your macros, which is a bonus

zaneselvans: Okay I went ahead and ported the information over.

zaneselvans:
@krivard I went ahead and added both new data test docs to the file since I was in there and these two PRs hang off of each other.

krivard (approved):
🚢 Good fix!

PR #4353, opened 2025-06-25
Remove superset terraform config and Docker files

zaneselvans:
# Overview

Undeploy superset, given that we are going with Marimo for our usage metrics dashboards, and the Eel Hole for our publicly facing data access

- I removed all the elements of `main.tf` that had superset in their names, and ran `terraform validate` and it passed.
- I removed all the files (mostly related to a Docker image spec) under the `superset` directory.

## To-do list

- [x] Grep around for any other docs or references to Superset.
- [x] Verify whether service account 345950277072 and its dependents are still required, or can be deleted.
- [x] Actually run `terraform apply` and deal with the consequences.
- [x] Verify that the cloud run and CloudSQL instances have been shut down.
- [x] Verify that our other existing services are undamaged.


terraform/main.tf
```diff
@@ -144,315 +144,12 @@ resource "google_sql_database" "mlflow_postgresql_database" {
...
-}
-
-resource "google_cloud_run_v2_service_iam_member" "cloudbuild_superset" {
-  location = google_cloud_run_v2_service.pudl-superset.location
-  name     = google_cloud_run_v2_service.pudl-superset.name
-  role     = "roles/run.admin"
-  member   = "serviceAccount:345950277072@cloudbuild.gserviceaccount.com"
-}
-
 data "google_compute_default_service_account" "google_compute_default_service_account_data" {
```
jdangerx: I believe we can delete this and the ability for cloudbuild to impersonate the default compute SA - the only thing we have running in cloud build is the superset build.

jdangerx (approved):
:shipit:

PR #4316, opened 2025-06-10
Remove the long-deprecated PudlTabl output management class

zaneselvans:
# Overview

This PR will remove all uses of and references to the PudlTabl output class, and the class definition itself.

- [x] Update stale docstrings that mention `pudl_out` or `PudlTabl` and its methods and instead refer to the modern PUDL table names directly.
- [ ] Refactor `pudl.glue.ferc1_eia` to no longer use `PudlTabl`
- [ ] Refactor `test/integration/glue_tests.py` to no longer use `PudlTabl`
- [ ] Update notebooks under `test/validate/notebooks/` to read from Parquet instead of using `PudlTabl`
- [ ] Refactor `pudl.output.epacems` to no longer use PudlTabl`
- [ ] Remove all reference to the PudlTabl outputs from our documentation
- [ ] Check that the docs on generating new plant & utility IDs for manual mapping are still correct
- [ ] Delete `pudl.output.pudltabl` module ⚔️ 

Closes #3215 

## Before merging...

- [ ] Run `make pytest-coverage` locally to ensure that the merge queue will accept your PR.
- [ ] Run a manual branch deployment.
- [ ] Review the PR yourself and call out any questions or issues you have.

PR #4354, opened 2025-06-25
Remove deprecated dbt tables from dbt data test specs

zaneselvans:
# Overview

There were a couple of references to the `_out_eia__` tables that we removed with the death of `PudlTabl` in #4316 in the singular dbt test definitions... which somehow didn't result in test failures!

- I've replaced those references with the normal `out_eia__` tables instead.
- this PR also removes a vestigial make target (`make dagster`) that was out of date, and should be replaced with the `dagster dev` command directly -- it used to make sense to use make here, because we needed to supply the dagster environments to the command, but now those are in `pyproject.toml` and get picked up automatically.

## Testing

- I ran all of the dbt tests locally and these singular tests didn't generate warnings.

### DIFF
--- a/docs/dev/run_the_etl.rst
+++ b/docs/dev/run_the_etl.rst
@@ -182,13 +182,14 @@ Once ``DAGSTER_HOME`` is set, launch the dagster UI by running:
 
 .. code-block:: console
 
-    $ make dagster
+    $ dagster dev
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


jdangerx (approved):
Verified that 16 `*validate*` tests run and pass on this branch, and only 14 run on `main` (with two warnings about removed schemas). :ship:

Issue #4333, opened 2025-06-19
Improve ergonomics of updating expected row counts


zaneselvans:
# Overview

Updating the expected row counts for dbt can be frustrating. Some issues that have come up:

## ~Needing to run the full ETL locally~

- With #4341 we started saving the row counts to the build outputs whenever we run a deployment, so this is no longer an issue

## Using non-standard partition columns

- We have some default columns that are used to partition tables so we have more granular row count expectations
- These are not always the right partitioning columns, and so we have the option of specifying other columns in the tests.
- However, when adding a new table and row count expectations for the first time, the script doesn't allow you to specify what non-standard columns you would like to use. So you have to add it, and then edit the test spec in `schema.yml` and then go back and regenerate the expectations. Having a direct

## Removing obsolete partition values

(maybe this is/was already fixed, by @jdangerx's PR making row count checks more strict?) 

- When updating row counts, if there's an old partition value that's no longer relevant (e.g. because the partitioning column has been changed) the script will add new records to the row count CSV, but doesn't remove the obsolete records.

e-belfer:
Row counts are not updated for datasets without partitions - e.g., `core_eia__entity_generators` when running with `all` and `--update`.

e-belfer:
When runnning `dbt_helper update-tables --row-counts --update core_pudl__assn_eia_pudl_plants` on #4332, the command added a new row rather than updating the existing one.

zaneselvans:
When running

```
dbt_helper update-tables --row-counts all
```

the script exits with success and no errors, but does not actually update any **existing** row counts.  However it will add row counts for missing tables. My intuition was that if I ask it to update the row counts and it can't (because I didn't use `--clobber`) that it would fail and give me an error.

zaneselvans:
The help message explaining the `--update` option only mentions its effects on the schemas, and says nothing about how it will affect the row counts.

e-belfer:
> When running
> 
> ```
> dbt_helper update-tables --row-counts all
> ```
> 
> the script exits with success and no errors, but does not actually update any **existing** row counts. However it will add row counts for missing tables. My intuition was that if I ask it to update the row counts and it can't (because I didn't use `--clobber`) that it would fail and give me an error.

It _should_:

> if table_name in existing["table_name"].to_numpy() and not (clobber or update):
>         return UpdateResult(
>             success=False,
>             message=f"Row counts for {table_name} already exist (run with clobber or update to overwrite).",
>         )

[...]

PR #4359, opened 2025-06-26
Update changed unpartitioned row-counts manually & sort row count file.

zaneselvans:
# Overview

After merging ~50,000 lines of changes since the last nightly build passed... nobody is surprised there are a handful of issues.

- Remove `report_date` from the the row counts partition of one of the SEC 10-K Exhibit 21 tables, since we removed date dependence. This caused a dbt test error.
- Update row counts for a handful of unpartitioned tables that the `dbt_helper` script is currently failing to update. See #4333  

## To Do
- [x] Run a branch deployment and make sure that this actually fixes all of the issues...
- [x] Create #4364


dbt/seeds/etl_full_row_counts.csv
```diff
@@ -3553,6 +3523,37 @@ out_pudl__yearly_assn_eia_ferc1_plant_parts,2023,2114
...
+out_sec10k__parents_and_subsidiaries,2015,218300
+out_sec10k__parents_and_subsidiaries,2016,232626
+out_sec10k__parents_and_subsidiaries,2017,231337
+out_sec10k__parents_and_subsidiaries,2018,33907
+out_sec10k__parents_and_subsidiaries,2019,32582
+out_sec10k__parents_and_subsidiaries,2020,30415
+out_sec10k__parents_and_subsidiaries,2021,29681
+out_sec10k__parents_and_subsidiaries,2022,29297
+out_sec10k__parents_and_subsidiaries,2023,212135
+out_sec10k__parents_and_subsidiaries,,16781
```
e-belfer, line 3556: What is this extra row with no partition?

zaneselvans: I assume there are a number of records in here for which the `report_year` value is NA but let me check...

```sql
select count(*) from '../pudl-output/parquet/out_sec10k__parents_and_subsidiaries.parquet' where report_date is NULL;
```
```
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│    16781     │
└──────────────┘
```

e-belfer: Is that desired behavior for this table, or a problem?

zaneselvans: `report_date` isn't part of the PK for this table so it's allowed. I looked at `filing_date` and it has the same number of NULL values. I think these dates come from the quarterly filing data that's linked to this table through the `filename_sec10k` field.  But it does seem weird.

Samping the rows that have `report_date IS NULL` I see that 100% of the parent and subsidiary columns are null in virtually all of them, which makes me suspect a bad merge (like a left merge onto the quarterly filings table?)

e-belfer: Sounds like it's worth writing up an issue looking into this more, but for now that we should update the row counts to reflect the data we are publishing? Otherwise we'll need to separate out the SEC10k code changes from everything else that's currently in main.

zaneselvans: So... I think something is wrong here and we should investigate in #4360 but the row count output reflects the current data reality, so I think it's good to merge for the moment.

zaneselvans:
We have a strange outcome here.

## Different local / nightly row counts (new badness)

The dbt row count check has failed, claiming that there’s 1 fewer rows in a table than I see locally.  I downloaded the file and it does in fact have one fewer rows. Let me regenerate again locally...

```
E           AssertionError: failure contexts:
E           source_check_row_counts_per_partition_pudl__None__core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities:
E           
E                                                     table_name expected_partition observed_partition  expected_count  observed_count  diff_relative_to_expected
E           0  core_sec10k__assn_exhibit_21_subsidiaries_and_...                                                  5546            5545                   -0.00018
```

## `dbt_helper` blind to unpartitioned tables (known issue)

However the fresh `etl_full_row_counts.csv` that has been produced by `dbt_helper` after the nightly builds does not reflect this because `dbt_helper` can't update unpartitioned tables.

## Spurious row updates (known behavior but it's actually a problem)

Also, the new row counts **do** have changes -- for the FERC 1 rate base table, which (for reasons I don't understand) apparently has an unstable number of rows?  This will cause the nightly builds to fail because it results in a dirty repo when it attempts to update the `nightly` tag (and would also break the `stable` tag or a versioned release tagging process).

If there are going to be tables with ill-defined row counts, we need to be able to turn off the row count test for them entirely, not just set it to `warn` -- meaning that `dbt_helper` does not produce row counts for tables without the row count test specified, which it currently does.  Otherwise we can't use the script to effectively detect whether there are changes.

zaneselvans:
Rematerialized all assets upstream & downstream of `core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities` locally and I still 5546 rows.

Loading them both to see wtf this one row diff is. Also rebuilding my local conda environment just in case.

The one row that I get locally that doesn't show up in the nightly build is:

```py
[('0000034088_exxon mobil_', 6088)]
```

Unsurprisingly, updating my `conda` environment doesn't change the outcome.

zaneselvans:
* The row counts in here should match the nightly build outputs now.
* At least for me, one of the SEC 10-K tables consistently has an additional row in the local ETL.
* For the moment I have set that row count test to `severity: warn` so we can merge this w/o breaking things.
* However, I still expect nightly builds to fail with these changes b/c `dbt_helper` creates row counts for tables that don't have the row-count test specified (e.g. the mysteriously variable FERC 1 rate base tables).
* I'm going to run a branch deployment just to confirm that everything else is still working and then merge.
* I will work on the `dbt_helper` bug(s) in #4364

PR #4362, opened 2025-06-26
Add pre-commit hook to keep row count files sorted.

zaneselvans:
# Overview

As noted in comments on #4333 if someone manually edits our row count CSVs rather than using `dbt_helper` this can result in them no longer being sorted, which will cause spurious diffs in the nightly builds or any time someone else runs `dbt_helper` to update the row counts.

This PR adds a git pre-commit hook that forces these CSVs to be sorted before they can be checked in to the repo.

## Testing

- I moved a row in the file and ran the pre-commit hook, and it found a change, and re-sorted the file, failing the pre-commit hook.
- The hook ran on pre-commit-ci and successfully found naturally occurring unsorted entries and sorted them.

## To-do list

- [x] Check whether this hook can be run effectively by `pre-commit-ci` or if it needs to be excluded.


dbt/seeds/etl_fast_row_counts.csv
zaneselvans, line 1: The hook runs just fine in pre-commit-ci and successfully found and fixed these unsorted entries.

dbt/seeds/etl_full_row_counts.csv
zaneselvans, line 1: The hook runs just fine in pre-commit-ci and successfully found and fixed these unsorted entries.

katie-lamb (approved):
In the process of learning about dbt I thought I'd quickly review this PR - I tested it by changing a row in the row counts files and had the pre-commit resort it to be correct. Looks good to me.

Issue #4364, opened 2025-06-27
Fix row-count update bugs in `dbt_helper`


zaneselvans:
# Overview

A number of actual bugs have been raised in the "ergonomics" focused issue #4333 and I'm going to move them here and make them urgent. It turns out this process is finicky and we need robust unit tests to make sure that it keeps doing all the things we need it to do as the script evolves...

## Success Criteria

All of the following need to work either when table row counts are updated specifically, or as part of "all"

* [ ] Row counts for **unpartitioned** tables are **created** if they don't exist.
* [ ] Row counts for **unpartitioned** tables are **updated** if they do exist.
* [ ] Updating row counts on an existing **unpartitioned** table does not result in duplicate entries.

zaneselvans:
## Refactoring notes

* Now that we have all of our tables in, it doesn't make sense for `dbt_helper` to focus on pre-populating tests with naive defaults (using `report_year`, `report_date`, and `datetime_utc` as the only possible partition columns)
* If we want to have an "add-tables" subcommand that does this scaffolding that's great, but it seems like the expected behavior when we're adding vs. updating tables is different.
* We should be able to use any column (or none) as a partition, rather than hard-coding a handful of temporal columns with a hard-coded prioritization (e.g. we should partition the EPA CEMS by `year`)
* We should be able to add a row count test by hand, specifying the partition column, and then have `dbt_helper` use that information when creating the row counts -- either the first time or during an update. This would let us change the partition column easily.
* The functions that calculate row counts need to be decoupled somehow from file paths in order to be easier to test. Or we need to use temporary files / test data stored on disk as the inputs.
* Because we need to use DuckDB to query the tables (speed, memory) we can't use dataframes as the inputs to these functions in `dbt_helper`
* However, DuckDB is happy to query pandas dataframes just as well as Parquet files, so maybe we can use dataframes in the tests, and Parquet in the real application?

zaneselvans:
WTF now it's refusing to pick up the new row counts for my local imputed demand tables, which were built on a different branch. But the tests (correctly) fail. 😠

marianneke:
The --clobber behavior of silently not updating seems really counterintuitive of how I would understand clobber to work. Let's specify what the desired behavior for clobber and maybe update could be, if you think there is a desirable behavior we can assign to the latter.

marianneke:
[...]

PR #4369, opened 2025-06-30
Clobber updated nightly row-counts before updating nightly tag

zaneselvans:
# Overview

The real change in here is just a couple of lines (which I will highlight) that check-out the original row counts before attempting to update the nightly or stable tag.  This is a short-term fix to get around the fact that a few tables have unstable row counts, and we have no way of excluding them from the row counts that are automatically generated at the moment.  See #4364 and #4367

However... in the process of making this update my VS Code ShellCheck plugin reformatted the whole script. And also somehow this is now being applied by NeoVim.  I've found the ShellCheck linting to be really helpful in the past for avoiding common shell programming errors / pitfalls & enforcing best practices so I went ahead and added it as a pre-commit hook so we can all have the same formatting & catch the same common issues.

# Testing

- Ran shell-check locally on the nightly build script and dealt with errors / warnings.
- Ran it again and got it to pass.
- Can only check the nightly build tag updating process by running a nightly build on main, so this hasn't been tested yet.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
       - id: end-of-file-fixer
 
   ####################################################################################
-  # Formatters: hooks that re-write Python & documentation files
+  # Formatters: hooks that re-write code & documentation files
   ####################################################################################
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.12.1
@@ -40,6 +40,11 @@ repos:
     hooks:
       - id: prettier
         types_or: [yaml]
+
+  - repo: https://github.com/koalaman/shellcheck-precommit
+    rev: v0.10.0
+    hooks:
+      - id: shellcheck
 
   ####################################################################################
   # Linters: hooks that check but don't alter files
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -23,54 +23,54 @@ function initialize_postgres() {
     # 3. tell it to actually fail when we mess up, instead of continuing blithely
     # 4. create a *dagster* user, whose creds correspond with those in docker/dagster.yaml
     # 5. make a database for dagster, which is owned by the dagster user
-    pg_ctlcluster "$PG_VERSION" dagster start && \
-    createdb -h127.0.0.1 -p5433 && \
-    psql -v "ON_ERROR_STOP=1" -h127.0.0.1 -p5433 && \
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -118,5 +123,5 @@ ci:
   autoupdate_branch: main
   autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
   autoupdate_schedule: weekly
-  skip: [unit-tests, nb-output-clear, conda-lock]
+  skip: [unit-tests, nb-output-clear, conda-lock, shellcheck]
```
zaneselvans, line 126: Skipping because the first attempted run failed due to the lack of a `docker` executable -- not sure what's going on here, as it seems like docker-based checks are common, and it **seemed** like the pre-commit hook repo for this check is set up right for docker but... I dunno man.

docker/gcp_pudl_etl.sh
zaneselvans, line 1: Almost everything in here is just reformatting by shellcheck. I will highlight the actual changes.

docker/gcp_pudl_etl.sh
```diff
@@ -161,44 +161,48 @@ function merge_tag_into_branch() {
...
+        git config user.name "pudlbot" &&
+        set +x &&
+        echo "Setting authenticated git remote URL using PAT" &&
+        git remote set-url origin "https://pudlbot:$PUDL_BOT_PAT@github.com/catalyst-cooperative/pudl.git" &&
+        set -x &&
+        echo "Updating $BRANCH branch to point at $TAG." &&
+        # Check out the original row counts so the working tree is clean.
+        # This is a temporary hack around the unstable row-counts in some tables.
+        # TODO: fix this for real in issue #4364 / PR #4367
+        git checkout -- dbt/seeds/ &&
```
zaneselvans, line 175: This is the substantive change -- check out the original version of the row count seed CSVs so that we don't have a dirty working tree when we try to update the `nightly` / `stable` tags.  Need to fix this for real in #4364 / #4367

jdangerx: Ah and this function only gets called after the outputs get saved to GCS so we still get the updated seeds if we need em.

jdangerx (approved):
Sure will let us push a tag!

I wonder if adding a `git diff dbt/seeds` would make it a little easier to see if we're accidentally missing any row-counts changes, though I understand that any effort making this flow better seems better used actually dealing with the root causes in #4364 .

PR #4380, opened 2025-07-01
Tweak nuke fraction test

zaneselvans:
# Overview

A tweak to @katie-lamb's nuclear fraction test migration -- I was making a suggestion, and tested it, and it got a little more complex than I expected so I made this PR rather than push to her branch.

### DIFF
--- a/dbt/tests/data_tests/generic_tests/expect_column_fraction_with_condition.sql
+++ b/dbt/tests/data_tests/generic_tests/expect_column_fraction_with_condition.sql
@@ -1,20 +1,30 @@
 {% test expect_column_fraction_with_condition(
     model,
     column_name,
-    column_condition,
-    expected_fraction,
-    tol=0.02
+    row_condition,
+    min_fraction,
+    max_fraction
 ) %}
 
+WITH fraction_check AS (
+    SELECT
+        SUM(CASE WHEN {{ row_condition }} THEN {{ column_name }} ELSE 0 END) * 1.0 / SUM({{ column_name }}) AS actual_fraction,
+        SUM(CASE WHEN {{ row_condition }} THEN {{ column_name }} ELSE 0 END) AS condition_sum,
+        SUM({{ column_name }}) AS total_sum,
+    FROM {{ model }}
+)
+
 SELECT
-    SUM(
-        CASE
-            WHEN {{ column_condition }}
-            THEN {{ column_name }} ELSE 0
-        END
-    )* 1.0
-    / SUM({{ column_name }}) AS actual_fraction
-FROM {{ model }}
-HAVING abs(actual_fraction - {{ expected_fraction }}) > {{ tol }}
+    'Test failed: actual_fraction of {{ column_name }} where ' ||
+    '{{ row_condition | replace("'", "''")}}' ||
+    ' is outside of acceptable range.'
+    AS failure_message,
+    ROUND(actual_fraction, 4) as actual_fraction,
+    {{ min_fraction }} AS min_fraction,
+    {{ max_fraction }} AS max_fraction,
+    ROUND(condition_sum, 4) AS condition_sum,
+    ROUND(total_sum, 4) AS total_sum
+FROM fraction_check
+WHERE actual_fraction < {{ min_fraction }} OR actual_fraction > {{ max_fraction }}
 
 {% endtest %}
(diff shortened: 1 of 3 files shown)


dbt/tests/data_tests/generic_tests/expect_column_fraction_with_condition.sql
```diff
@@ -1,20 +1,30 @@
 {% test expect_column_fraction_with_condition(
     model,
     column_name,
-    column_condition,
-    expected_fraction,
-    tol=0.02
+    row_condition,
+    min_fraction,
+    max_fraction
```
zaneselvans, line 6: For most of the other tests where there's some range we've gone with explicitly setting a min/max rather than a central value with a tolerance -- it's just a little more flexible/explicit -- so I made that tweak here.

There's already a convention in `dbt` (or maybe it's `dbt_expectations`) that this kind of selector is called named `row_condition` rather than `column_condition`, so using the latter seemed kind of confusing to me. That being said, the behavior here is slightly different than the existing `row_condition` default meaning -- in the other tests it preemptively limits what rows are sent into the test at all, so maybe we should pick another name entirely.

dbt/tests/data_tests/generic_tests/expect_column_fraction_with_condition.sql
```diff
@@ -1,20 +1,30 @@
...
 ) %}
 
+WITH fraction_check AS (
+    SELECT
+        SUM(CASE WHEN {{ row_condition }} THEN {{ column_name }} ELSE 0 END) * 1.0 / SUM({{ column_name }}) AS actual_fraction,
+        SUM(CASE WHEN {{ row_condition }} THEN {{ column_name }} ELSE 0 END) AS condition_sum,
+        SUM({{ column_name }}) AS total_sum,
+    FROM {{ model }}
+)
+
```
zaneselvans, line 16: In order to build useful failure messages I've found it really helpful to make these little temporary tables (aka "Common Table Expressions" or CTEs) and then select from them.

dbt/tests/data_tests/generic_tests/expect_column_fraction_with_condition.sql
```diff
@@ -1,20 +1,30 @@
...
+    '{{ row_condition | replace("'", "''")}}' ||
+    ' is outside of acceptable range.'
+    AS failure_message,
+    ROUND(actual_fraction, 4) as actual_fraction,
+    {{ min_fraction }} AS min_fraction,
+    {{ max_fraction }} AS max_fraction,
+    ROUND(condition_sum, 4) AS condition_sum,
+    ROUND(total_sum, 4) AS total_sum
+FROM fraction_check
+WHERE actual_fraction < {{ min_fraction }} OR actual_fraction > {{ max_fraction }}
```
zaneselvans, line 28: For debugging purposes it's nice if the failing outputs have all the information required to understand what's going on -- rather than needing to cobble it together from the name of the test that failed / the `schema.yml` file, etc. It would be easy to have more than one test of this variety applied to the same table, checking different columns, or different categories within a particular column (adding in gas or coal or solar fractions...) so adding the row condition and column_name parameters in here is nice.

I was trying to do this in a suggestion on your PR but then had to debug escaping the single-quotes in the `row_condition` and so just made this PR onto your PR instead.

dbt/tests/data_tests/generic_tests/expect_column_fraction_with_condition.sql
zaneselvans: (though I realize the original test you were migrating had this symmetric check)

dbt/tests/data_tests/generic_tests/expect_column_fraction_with_condition.sql
zaneselvans: This produces output like 
|                                                    failure_message                                                     | actual_fraction | min_fraction | max_fraction |  condition_sum   |    total_sum     |
|------------------------------------------------------------------------------------------------------------------------|----------------:|-------------:|-------------:|-----------------:|-----------------:|
| Test failed: actual_fraction of net_generation_mwh where fuel_type_code_pudl='nuclear' is outside of acceptable range. | 0.1949          | 0.08         | 0.12         | 19152669036.0512 | 98282638887.2146 |

dbt/tests/data_tests/generic_tests/expect_column_fraction_with_condition.sql
katie-lamb: Yeah I saw that `dbt_expectations` has the `row_condition` convention, but also felt like this case didn't exactly mirror the behavior used there. I then must have got confused about intuiting what a `column_condition` would be. `row_condition` seems fine and I think is intuitive enough, but maybe `row_filter` is a better name just to be unambiguous about it being a little different than `dbt_expectations`?

katie-lamb (approved):
I made one suggestion about changing the `row_condition` variable name to `row_filter`. The test failure output looks good to me.

PR #4382, opened 2025-07-03
Add no-null-cols dbt test to all tables

### LINKED ISSUE #4105: Migrate not_null validation tests to dbt

zaneselvans:
# Overview

- Implements and exhaustive check for completely null columns across all distributed PUDL tables implemented in `dbt`
- Removes the remaining `no_null_cols` tests under `test/validate`
- Renames `id_dc_coupled_tightly` to `is_dc_coupled_tightly` (typo)
- Consolidates `operation_switch` columns with existing `can_switch_when_operating` that means the same thing
- Removes several superseded Dagster asset checks that were checking for all-null columns.
- Converts several Y/N columns to `boolean` which were previously getting set to NULL in the multifuel generator tables.
- Changes how we import pandera to be pandas specific, since the library now supports many types of dataframes.
- Excludes some columns from being checked for NULL in the entity tables, since there's no way to block on year.
- Adds exhaustive special cases for identifying which NULL column-years are acceptable, but only the tables that needed any special cases for the Fast ETL to pass -- there are surely many more for other tables and columns, if we want to add them.

Closes #4105

## Remaining Test Design Questions
- Should we apply this kind of testing for expected/unexpected null column-year blocks across all tables?
- Is it worth pulling this test into a macro so we can unit test it?
- The test currently checks that columns have no entirely null years for some expected subset of all years. But it does not make the complementary assertion -- that in the other years, the columns **are** NULL, as expected. This would be easy to add to the existing test, without changing any parameters.

# Testing
- Ran all instances of the new test on local full ETL outputs (and found some minor issues with:
```bash
dbt build --select "test_name:expect_columns_not_all_null"
```
- Ran `make pytest-integration` to identify columns that need special cases because they only exist in older years.
- Wrote a script to find all the fully null column-years in a table so we can catch the special cases that aren't triggered by the integration tests.

## To-do list
- [x] Rename `id_dc_coupled_tightly` to `is_dc_coupled_tightly`
- [x] Add missing boolean columns to those that get fixed, so the Y/N values aren't nulled out.
- [x] Fix pandera imports to be pandas specific to avoid deprecation warnings.
- [x] Rename `conditional_columns` to `row_conditions` in test params.
- [x] Do a self-review of the PR to highlight issues / ask questions.
- [x] Add test description fields to explain special cases / excluded columns.
- [x] Catalog all the special cases that look suspicious for further investigation.

### Full ETL
- [x] Investigate all-null `core_nrelatb__yearly_projected_cost_performance.model_tax_credit_case_nrelatb`
- [x] Investigate all-null `core_nrelatb__yearly_projected_financial_cases_by_scenario.model_tax_credit_case_nrelatb`
- [x] Investigate 3 all-null columns in `core_eia860__scd_generators_multifuel` (boolean cols with only Y/N values)
- [x] Investigate why `is_dc_coupled_tightly` is entirely null (changing the name seems to have fixed it)
- [x] Do a `workflow_dispatch` run of `build-deploy-pudl` to make sure that the tests pass in the nightly builds.

### Fast ETL
- [x] Run the fast ETL to see what special cases need to be accommodated in CI (discontinued columns)
- [x] `_core_eia860__cooling_equipment`
- [x] `_core_eia860__fgd_equipment`
- [x] `_core_eia923__cooling_system_information`
- [x] `_core_eia923__fgd_operation_maintenance`
- [x] `core_eia860__assn_boiler_generator`
- [x] `core_eia860__assn_boiler_stack_flue`
- [x] `core_eia860__scd_generators_solar`
- [x] `core_eia860__scd_generators_wind`
- [x] `out_eia923__fuel_receipts_costs`
- [x] `core_eia860__scd_plants`
- [x] `core_eia860__scd_utilities`
- [x] `core_eia860__scd_generators_multifuel`
- [x] `core_eia923__entity_coalmine`
- [x] `core_eia__entity_boilers`
- [x] `core_eia__entity_generators`
- [x] `out_eia__monthly_generators`
- [x] `out_eia__yearly_boilers`
- [x] `out_eia__yearly_generators`
- [x] Run `make pytest-integration` make sure that we actually specified all the exceptions correctly.


src/pudl/metadata/fields.py
```diff
@@ -5925,7 +5925,7 @@
             "still charge from the grid)."
         ),
     },
-    "id_dc_coupled_tightly": {
+    "is_dc_coupled_tightly": {
```
zaneselvans, line 5924: Given the other similar columns, this sure seemed like a typo, and it just happened to also be all null (apparently not related to the typo) so I updated the name. Still not sure why it's all null.

dbt/models/eia860/core_eia860__scd_generators_energy_storage/schema.yml
```diff
@@ -31,7 +35,7 @@ sources:
           - name: stored_excess_wind_and_solar_generation
           - name: is_ac_coupled
           - name: is_dc_coupled
-          - name: id_dc_coupled_tightly
+          - name: is_dc_coupled_tightly
```
zaneselvans, line 42: Rename due to apparent typo.

dbt/models/eia860/core_eia860__scd_generators_energy_storage/schema.yml
```diff
@@ -4,6 +4,10 @@ sources:
     tables:
       - name: core_eia860__scd_generators_energy_storage
         data_tests:
+          - expect_columns_not_all_null:
+              exclude_columns:
+                - storage_technology_code_3
+                - storage_technology_code_4
```
zaneselvans, line 14: These seem to be placeholders for a future in which there could be more than 2 energy storage techs deployed in a place, so it seems okay that they're all null for now. But it would be good if we could automatically catch when they are no longer all null in the future. Where is the right place to do that and how?

dbt/models/eia860/_core_eia860__fgd_equipment/schema.yml
```diff
@@ -4,6 +4,22 @@ sources:
     tables:
       - name: _core_eia860__fgd_equipment
         data_tests:
+          - expect_columns_not_all_null:
+              exclude_columns:
+                - sorbent_type_3
+                - sorbent_type_4
+                - so2_equipment_type_4
```
zaneselvans, line 14: These appear to be placeholders that allow a larger number of sorbent / equipment types to be specified than anybody is actually using right now.

dbt/models/eia860/_core_eia860__cooling_equipment/schema.yml
```diff
@@ -4,6 +4,15 @@ sources:
     tables:
       - name: _core_eia860__cooling_equipment
         data_tests:
+          - expect_columns_not_all_null:
+              exclude_columns:
+                - tower_type_3
+                - tower_type_4
+                - cooling_type_4
```
zaneselvans, line 14: Appear to be placeholders for a larger number of these kinds of things than anyone is actually using right now.  But would be nice if we could automatically detect if/when someone does actually report a value in one of them so we can make this more stringent / migrate it to a conditional column rather than an exclusion.

I guess we could also just use the conditional columns to exclude all years up to the present, and each year that would result in a new failure in the new year, unless someone has started using the column.  Maybe that would be more elegant / simpler.

dbt/models/eia860/_core_eia860__fgd_equipment/schema.yml
```diff
@@ -4,6 +4,22 @@ sources:
...
+                county: EXTRACT(year FROM report_date) <= 2010
+                county_id_fips: EXTRACT(year FROM report_date) <= 2010
+                fgd_operating_date: EXTRACT(year FROM report_date) <= 2012
+                fgd_operational_status_code: EXTRACT(year FROM report_date) <= 2012
+                sludge_pond: EXTRACT(year FROM report_date) >= 2016
+                fgd_manufacturer: EXTRACT(year FROM report_date) <= 2012
+                fgd_manufacturer_code: EXTRACT(year FROM report_date) <= 2012
+                steam_plant_type_code: EXTRACT(year FROM report_date) >= 2014
+                plant_summer_capacity_mw: EXTRACT(year FROM report_date) <= 2010
+                water_source: EXTRACT(year FROM report_date) <= 2010
```
zaneselvans, line 25: In general, these conditions seem to take the form of "new column appeared in year X" or "old column was deprecated in year Y" -- the very few cases where there's a smattering of all null columns across a couple of years but not others I've added to the `exclude_columns` argument, because it seems like there are a few columns that are so rarely used that there are in fact many years in which there are zero values, even though they haven't been discontinued / added.

Alternatively, those flaky years that aren't a new or discontinued column could be specified explicitly with `IS IN / NOT IN (list, of, years)` -- which would be better?

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
```diff
@@ -0,0 +1,63 @@
+{% test expect_columns_not_all_null(
+    model,
+    exclude_columns=[],
+    conditional_columns={}
+) %}
+
+-- Get all column names using DuckDB's DESCRIBE
```
krivard: Optional backoff:

`DESCRIBE` has broad-ish support (DuckDB but also Oracle and MySQL, and postgres if you squint. Not SQL Server or T-SQL though)

so this is still technically brittle to our choice of backend, but not quite as brittle as being limited to DuckDB only.

If you wanted to specify this was DuckDB's version of DESCRIBE bc we're using DuckDB's `column_name` instead of e.g. MySQL's `Field` from the result, that's fine, I just twigged on "wait how obscure is this? oh. not that obscure"

```suggestion
-- Get all column names using DESCRIBE
```

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
```diff
@@ -0,0 +1,63 @@
...
+{% set get_columns_query %}
+    SELECT column_name
+    FROM (DESCRIBE {{ model }})
+    WHERE column_name NOT IN ({{ "'" + exclude_columns | join("', '") + "'" if exclude_columns else "''" }})
```
krivard: Does this not work?

```suggestion
    WHERE column_name NOT IN ('{{ exclude_columns | join("', '") }}')
```

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
```diff
@@ -0,0 +1,63 @@
...
+{% endset %}
+
+{% if execute %}
+    {% set columns_result = run_query(get_columns_query) %}
+    {% set column_names = columns_result.columns[0].values() %}
+{% else %}
+    {% set column_names = [] %}
+{% endif %}
+
+WITH column_null_checks AS (
```
krivard: Could you move this down to L53? the `WITH` context isn't necessary while building `checks`.

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
krivard: or maybe
```suggestion
-- DESCRIBE implementations vary; this assumes DuckDB's
```

krivard (approved):
Looks reasonable. I read the test carefully, and only glanced through the models -- if you want more detailed eyes on the models lmk and I can take another pass.

* Tests are a macro that automatically runs the resulting SQL and then reasons about the output. They can't be unit tested because there's no way to turn off the reasoning step.
* If we want unit tests for this, you could really just move the entire body into a macro. Then use the bare macro version for unit testing purposes, and just run the macro inside the generic test for dbt testing purposes. See [weighted_quantile (macro)](https://github.com/catalyst-cooperative/pudl/blob/main/dbt/macros/weighted_quantile.sql), [expect_column_weighted_quantile_values_to_be_between (generic test)](https://github.com/catalyst-cooperative/pudl/blob/main/dbt/tests/data_tests/generic_tests/expect_column_weighted_quantile_values_to_be_between.sql), and [test_weighted_quantile (unit test)](https://github.com/catalyst-cooperative/pudl/blob/main/dbt/tests/unit_tests/test_weighted_quantile.sql) for an example.
* I'm in favor of keeping the blank columns and excluding them from the check. Mostly because figuring out we needed to add those columns back in seems like it would be a long and terrible journey.
* i'm with you on fixing the `id`->`is` typo

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
zaneselvans: I think so long as we stay compatible with Postgres and its descendants I will feel good about this.  I suspect we have other things with this level of brittleness floating around too.

dbt/models/eia860/_core_eia860__cooling_equipment/schema.yml
zaneselvans: @krivard ☝🏼 this seemed like it might be a good way to handle the "nobody has used this column **yet**" cases such that we wouldn't forever exclude a column that hasn't been used.  But also there's a good chance that these 3rd and 4th tier categories will just end up being the kind of column that are super sparsely used and often have years with no values, in which case allowing them to be fully NULL seems fine.

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
zaneselvans: Yep that seems to work just the same.

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
zaneselvans: Whoops. Cut-and-pasting.

dbt/models/eia923/out_eia923__fuel_receipts_costs/schema.yml
```diff
@@ -4,6 +4,14 @@ sources:
...
       - name: out_eia923__fuel_receipts_costs
         data_tests:
+          - expect_columns_not_all_null:
+              conditional_columns:
+                mercury_content_ppm: EXTRACT(year FROM report_date) NOT IN (2008, 2009, 2010, 2011, 2013)
+                natural_gas_delivery_contract_type_code: EXTRACT(year FROM report_date) >= 2014
+                moisture_content_pct: EXTRACT(year FROM report_date) >= 2014
+                chlorine_content_ppm: EXTRACT(year FROM report_date) >= 2014
+                mine_id_msha: EXTRACT(year FROM report_date) = 2025
+                coalmine_county_id_fips: EXTRACT(year FROM report_date) <= 2010
```
zaneselvans, line 14: These two look fishy to me. Is the MHSA mine ID really a brand new thing that only exists in 2025 data? Or is this something weird from our machinations with the coalmine table?  And similarly with the FIPS codes. I thought we were adding them based on the county name -- in which case, why would we only have them for a couple of the earliest years?

dbt/models/output/core_eia860__assn_boiler_generator/schema.yml
```diff
@@ -4,6 +4,11 @@ sources:
     tables:
       - name: core_eia860__assn_boiler_generator
         data_tests:
+          - expect_columns_not_all_null:
+              conditional_columns:
+                boiler_generator_assn_type_code: EXTRACT(year FROM report_date) BETWEEN 2009 AND 2012
+                steam_plant_type_code: EXTRACT(year FROM report_date) >= 2014
+                data_maturity: EXTRACT(year FROM report_date) >= 2009
```
zaneselvans, line 11: Why do we have no `data_maturity` in earlier years?

dbt/models/output/core_eia860__assn_boiler_generator/schema.yml
```diff
@@ -4,6 +4,11 @@ sources:
     tables:
       - name: core_eia860__assn_boiler_generator
         data_tests:
+          - expect_columns_not_all_null:
+              conditional_columns:
+                boiler_generator_assn_type_code: EXTRACT(year FROM report_date) BETWEEN 2009 AND 2012
```
zaneselvans, line 9: Why would we ONLY have a BGA association type code for 2009-2012?  Shouldn't they be getting assigned across the board (e.g. the original codes)

dbt/models/eia860/core_eia860__assn_boiler_stack_flue/schema.yml
```diff
@@ -4,6 +4,11 @@ sources:
     tables:
       - name: core_eia860__assn_boiler_stack_flue
         data_tests:
+          - expect_columns_not_all_null:
+              row_conditions:
+                stack_id_eia: EXTRACT(year FROM report_date) <= 2012
+                flue_id_eia: EXTRACT(year FROM report_date) <= 2012
+                stack_flue_id_eia: EXTRACT(year FROM report_date) >= 2013
```
zaneselvans, line 11: Not being familiar with this table, it looks like the way these IDs are reported changed in 2013, and I wonder if there's a single format that we can use that's uniform across all years?

dbt/models/ferc1/out_ferc1__yearly_all_plants/schema.yml
```diff
@@ -4,6 +4,17 @@ sources:
...
+              exclude_columns:
+                - capex_annual_addition_rolling
+                - capex_annual_per_mw_rolling
+                - capex_annual_per_mwh_rolling
+              row_conditions:
+                asset_retirement_cost: report_year >= 2003
+                capex_annual_addition: report_year >= 1995
+                capex_annual_per_kw: report_year >= 1995
+                capex_annual_per_mw: report_year >= 1995
+                capex_annual_per_mwh: report_year >= 1995
```
zaneselvans, line 22: I'm a little surprised that these columns are 100% null in the early years. Want to double check that's expected.

dbt/models/ferc1/out_ferc1__yearly_steam_plants_sched402/schema.yml
```diff
@@ -4,6 +4,17 @@ sources:
...
+              exclude_columns:
+                - capex_annual_addition_rolling
+                - capex_annual_per_mw_rolling
+                - capex_annual_per_mwh_rolling
+              row_conditions:
+                asset_retirement_cost: report_year >= 2003
+                capex_annual_addition: report_year >= 1995
+                capex_annual_per_kw: report_year >= 1995
+                capex_annual_per_mw: report_year >= 1995
+                capex_annual_per_mwh: report_year >= 1995
```
zaneselvans, line 22: Somewhat surprised that these are 100% in the earliest years. Want to check if that's real.

dbt/models/output/core_eia860__scd_boilers/schema.yml
```diff
@@ -4,6 +4,95 @@ sources:
     tables:
       - name: core_eia860__scd_boilers
         data_tests:
+          - expect_columns_not_all_null:
+              row_conditions:
```
zaneselvans, line 9: This table has a large number of row conditions that appear to exclude the most recent partial year (2025) -- are we actually getting any 2025 boiler information?

dbt/models/output/core_eia860__scd_generators/schema.yml
```diff
@@ -4,6 +4,62 @@ sources:
     tables:
       - name: core_eia860__scd_generators
         data_tests:
+          - expect_columns_not_all_null:
+              row_conditions:
```
zaneselvans, line 9: A lot of these row conditions are just excluding 2025 -- for which we only have the subset of generator attributes that are reported in the EIA-860M.  Is there a more compact / elegant way to handle this?  And also not make it annoying to update?

dbt/models/output/core_eia860__scd_plants/schema.yml
```diff
@@ -4,6 +4,42 @@ sources:
     tables:
       - name: core_eia860__scd_plants
         data_tests:
+          - expect_columns_not_all_null:
+              row_conditions:
```
zaneselvans, line 9: Another one with a lot of cases of null 2025 data.

dbt/models/output/core_eia860__scd_utilities/schema.yml
```diff
@@ -4,6 +4,30 @@ sources:
     tables:
       - name: core_eia860__scd_utilities
         data_tests:
+          - expect_columns_not_all_null:
+              row_conditions:
```
zaneselvans, line 9: Again, lots of 2025 exclusions. Seems like it could be annoying to update, and also it won't start failing / alert us when it needs to be updated. So it'll likely get stale and just fail to test more recent data as it fills in, unless we do something to force that.

dbt/models/output/core_eia__entity_generators/schema.yml
```diff
@@ -26,5 +29,5 @@ sources:
           - name: rto_iso_location_wholesale_reporting_id
           - name: associated_combined_heat_power
           - name: original_planned_generator_operating_date
-          - name: operating_switch
+          - name: can_switch_when_operating
```
zaneselvans, line 37: We had 2 different column names for the same variable and I consolidated them.

dbt/models/output/out_eia__monthly_generators/schema.yml
```diff
@@ -4,6 +4,71 @@ sources:
...
+                planned_modifications: EXTRACT(year FROM report_date) BETWEEN 2007 AND 2012
+                planned_net_summer_capacity_derate_mw: EXTRACT(year FROM report_date) >= 2007
+                planned_net_summer_capacity_uprate_mw: EXTRACT(year FROM report_date) >= 2007
+                planned_net_winter_capacity_derate_mw: EXTRACT(year FROM report_date) BETWEEN 2007 AND 2024
+                planned_net_winter_capacity_uprate_mw: EXTRACT(year FROM report_date) BETWEEN 2007 AND 2024
+                planned_new_capacity_mw: EXTRACT(year FROM report_date) BETWEEN 2013 AND 2024
+                planned_new_prime_mover_code: EXTRACT(year FROM report_date) BETWEEN 2007 AND 2024
+                planned_repower_date: EXTRACT(year FROM report_date) BETWEEN 2007 AND 2024
+                planned_uprate_date: EXTRACT(year FROM report_date) >= 2007
+                reactive_power_output_mvar: EXTRACT(year FROM report_date) IN (2004, 2005, 2006, 2016, 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024)
```
zaneselvans: Odd that reactive power was reported in very early years, and then again starting in 2016 to present.

dbt/models/output/out_eia__monthly_generators/schema.yml
```diff
@@ -4,6 +4,71 @@ sources:
     tables:
       - name: out_eia__monthly_generators
         data_tests:
+          - expect_columns_not_all_null:
+              row_conditions:
```
zaneselvans, line 9: Lots of conditions excluding the partial 2025 year for which we only have EIA-860M data.

dbt/models/output/out_eia__monthly_generators/schema.yml
```diff
@@ -179,7 +244,7 @@ sources:
           - name: can_burn_multiple_fuels
           - name: nameplate_power_factor
           - name: net_capacity_mwdc
-          - name: operating_switch
+          - name: can_switch_when_operating
```
zaneselvans, line 240: We had two column names for the same variable, and I consolidated them.

zaneselvans: I'm surprised there isn't a special case above for `can_switch_when_operating` -- it's only reported for a single year for generators in general (but much more common for the multi-fuel generators)

dbt/models/output/out_eia__yearly_assn_plant_parts_plant_gen/schema.yml
```diff
@@ -4,6 +4,10 @@ sources:
     tables:
       - name: out_eia__yearly_assn_plant_parts_plant_gen
         data_tests:
+          - expect_columns_not_all_null:
+              row_conditions:
+                unit_id_pudl_plant_gen: EXTRACT(year FROM report_date) BETWEEN 2008 AND 2024
+                ferc1_generator_agg_id_plant_gen: EXTRACT(year FROM report_date) BETWEEN 2005 AND 2019
```
zaneselvans: @cmgosnell does it make sense that these columns would be 100% null outside of these date ranges?

dbt/tests/data_tests/generic_tests/schema.yml
```diff
@@ -182,7 +197,7 @@ macros:
       - name: max_discrepancy_rate
         type: float
         description: Maximum acceptable fraction of records which can be counted as a discrepancy before the test should fail.
-  - name: expect_columns_are_close
+  - name: test_expect_columns_are_close
```
zaneselvans, line 212: Lack of `test_` prefix resulted in some warnings due to name mismatch.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -56,7 +56,7 @@
 from pudl.logging_helpers import get_logger
 from pudl.metadata.dfs import ImputationReasonCodes
 
-logger = get_logger(__file__)
+logger = get_logger(__name__)
```
zaneselvans, line 59: Loggers need the module chain, not the file path.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -39,7 +39,7 @@
 
 import numpy as np
 import pandas as pd
-import pandera as pa
+import pandera.pandas as pa
```
zaneselvans, line 42: Updated all of our pandera imports since they have reorganized to have dataframe-library specific sub-packages.

src/pudl/metadata/fields.py
```diff
@@ -5752,7 +5748,7 @@
             "production tax credit (PTC) and investment tax credit (ITC). For more detail, see: "
             "https://atb.nrel.gov/electricity/2024/financial_cases_&_methods"
         ),
-        "constraints": {"enum": ["Market", "R&D"]},
+        "constraints": {"enum": ["ITC", "PTC + ITC", "PTC"]},
```
zaneselvans, line 5751: Fix to NREL ATB null columns.

src/pudl/transform/eia860.py
```diff
@@ -507,15 +508,18 @@ def _core_eia860__generators_multifuel(
...
         "can_switch_oil_gas",
         "can_switch_when_operating",
+        "has_factors_that_limit_switching",
+        "has_storage_limits",
+        "has_air_permit_limits",
+        "has_other_factors_that_limit_switching",
+        "can_cofire_oil_and_gas",
+        "can_cofire_100_oil",
+        "can_fuel_switch",
+        "has_regulatory_limits",
```
zaneselvans, line 522: Not all boolean columns were listed here, so the Y/N values were getting set to NULL when the dtype was enforced, leading to entirely NULL columns.

src/pudl/transform/eia860.py
```diff
@@ -1312,30 +1316,6 @@ def _core_eia860__cooling_equipment(
     return apply_pudl_dtypes(ce_df, group="eia", strict=True).pipe(PUDL_PACKAGE.encode)
 
 
-@asset_check(asset=_core_eia860__cooling_equipment, blocking=True)
-def cooling_equipment_null_cols(cooling_equipment):
```
zaneselvans, line 1316: Superseded by dbt `expect_columns_not_all_null` test

src/pudl/transform/eia860.py
```diff
@@ -1463,28 +1443,6 @@ def _core_eia860__fgd_equipment(
     return PUDL_PACKAGE.encode(fgd_df).pipe(apply_pudl_dtypes, strict=False)
 
 
-@asset_check(asset=_core_eia860__fgd_equipment, blocking=True)
-def fgd_equipment_null_check(fgd):
```
zaneselvans, line 1467: Superseded by dbt `expect_columns_not_all_null` test

src/pudl/transform/eia923.py
```diff
@@ -1295,21 +1294,6 @@ def _core_eia923__cooling_system_information(
     )
 
 
-@asset_check(asset=_core_eia923__cooling_system_information, blocking=True)
-def cooling_system_information_null_check(csi):
```
zaneselvans, line 1299: Superseded by dbt `expect_columns_not_all_null` test

src/pudl/transform/eia923.py
```diff
@@ -1408,28 +1392,6 @@ def _core_eia923__fgd_operation_maintenance(
     )
 
 
-@asset_check(asset=_core_eia923__fgd_operation_maintenance, blocking=True)
-def fgd_operation_maintenance_null_check(fgd):
```
zaneselvans, line 1412: Superseded by dbt `expect_columns_not_all_null` test

src/pudl/validate.py
```diff
@@ -64,23 +64,22 @@ def no_null_rows(
...
+    df: pd.DataFrame, cols: list[str] | str = "all", df_name: str = ""
 ) -> pd.DataFrame:
     """Check that a dataframe has no all-NaN columns.
 
-    Occasionally in the concatenation / merging of dataframes we get a label
-    wrong, and it results in a fully NaN column... which should probably never
-    actually happen. This is a quick verification.
+    Occasionally in the concatenation / merging of dataframes we get a label wrong, and
+    it results in a fully NaN column... which should probably never actually happen.
+    This function can be used to guard against such errors.
```
zaneselvans: Only docstring and type annotation clean up in this file.

dbt/models/output/out_eia__yearly_boilers/schema.yml
```diff
@@ -4,6 +4,106 @@ sources:
...
+                a single year and are effectively foreward/back filled because they are
+                part of the boiler entity table. This means they're fully populated if
+                that year of data is included, and entirely null if it's not.
+              exclude_columns:
+                - mercury_control_existing_strategy_4
+                - mercury_control_existing_strategy_5
+                - mercury_control_existing_strategy_6
+                - boiler_manufacturer
+                - boiler_manufacturer_code
+              row_conditions:
```
zaneselvans, line 22: Another table with lots of conditions excluding 2025 -- still surprised that we have a bunch 2025 boiler information?

dbt/models/output/out_eia__yearly_plant_parts/schema.yml
```diff
@@ -4,6 +4,18 @@ sources:
     tables:
       - name: out_eia__yearly_plant_parts
         data_tests:
+          - expect_columns_not_all_null:
+              row_conditions:
+                unit_id_pudl: EXTRACT(year FROM report_date) BETWEEN 2008 AND 2024
+                ferc1_generator_agg_id: EXTRACT(year FROM report_date) BETWEEN 2005 AND 2019
```
zaneselvans, line 11: Why would `ferc1_generator_agg_id` only be present between 2005-2019?

zaneselvans:
@krivard since you mentioned taking a look at it, this is the notebook jank I've been using to compile the exhaustive `row_conditions` for the tables that weren't passing in CI

```python
# Create a python function which, given the name of a PUDL table , use duckdb to query it directly and for each column in the
# table, identify years where the column is entirely null. Use that information to construct the
# appropriate conditional_column argument for the `expect_columns_not_all_null` test in dbt, selecting
# only the years in which we expect the column to have non-null values.
def get_null_years(table_name: str, verbose: bool = False) -> dict[str, str]:
    """
    Identify years where each column in the specified PUDL table is entirely null.

    Args:
        table_name (str): The name of the PUDL table to analyze.
        verbose (bool): If True, print detailed information about the analysis.

    Returns:
        dict[str, str]: A dictionary mapping column names to SQL conditions for dbt tests.
    """
    import duckdb

    pq_path = Path(os.environ["PUDL_OUTPUT"]) / f"parquet/{table_name}.parquet"

    if not pq_path.exists():
        raise FileNotFoundError(f"Parquet file not found: {pq_path}")

    if verbose:
        print(f"Analyzing table: {table_name}")
        print(f"File path: {pq_path}")

    # Connect to DuckDB and work directly with the parquet file
    conn = duckdb.connect()

    try:
        # Get column names from the parquet file
        columns_query = f"DESCRIBE '{pq_path}'"
        columns_result = conn.execute(columns_query).fetchall()
        column_names = [row[0] for row in columns_result if row[0] != "report_date"]

        if verbose:
            print(f"Found {len(column_names)} columns to analyze (excluding report_date)")

        # Get available years in the dataset
        years_query = f"SELECT DISTINCT EXTRACT(year FROM report_date) AS year FROM '{pq_path}' ORDER BY year"
        available_years = [row[0] for row in conn.execute(years_query).fetchall()]

        if verbose:
            print(f"Available years: {available_years}")

        # Initialize a dictionary to hold conditions for each column
        conditional_columns = {}

        # Iterate over each column to find years where it is entirely null
        for column in column_names:
            # Find years where this column is entirely null
            # We do this by comparing the count of non-null values to the total count for each year
            query = f"""
            WITH year_stats AS (
                SELECT
                    CAST(EXTRACT(year FROM report_date) AS INTEGER) AS year,
                    COUNT(*) AS total_rows,
                    COUNT({column}) AS non_null_rows
                FROM '{pq_path}'
                GROUP BY EXTRACT(year FROM report_date)
            )
            SELECT year
            FROM year_stats
            WHERE non_null_rows = 0
            ORDER BY year
            """

            result = conn.execute(query).fetchall()
            if verbose:
                print(f"{len(result)} null years found for column '{column}'")
            null_years = [int(row[0]) for row in result]

            if verbose and null_years:
                print(f"Column '{column}' is entirely null in years: {null_years}")

            if null_years:
                # Generate compact condition based on the pattern of null years
                condition = generate_compact_year_condition(null_years, available_years)
                conditional_columns[column] = condition

    finally:
        conn.close()

    if verbose:
        print(f"Found {len(conditional_columns)} columns with entirely null years")

    return conditional_columns


def generate_compact_year_condition(null_years: list[int], available_years: list[int]) -> str:
    """
    Generate a compact SQL condition that excludes null years and includes only years with data.

    Args:
        null_years: List of years where the column is entirely null
        available_years: List of all years present in the dataset

    Returns:
        A compact SQL condition string
    """
    # Get the years where data exists (inverse of null_years)
    data_years = sorted([year for year in available_years if year not in null_years])

    if not data_years:
        # This shouldn't happen, but just in case
        return f"EXTRACT(year FROM report_date) NOT IN ({', '.join(map(str, null_years))})"

    # Check if data years form a contiguous range
    if len(data_years) == (data_years[-1] - data_years[0] + 1):
        # Contiguous range - use compact notation
        if len(data_years) == 1:
            return f"EXTRACT(year FROM report_date) = {data_years[0]}"
        elif data_years[0] == min(available_years):
            # Data starts at beginning and goes to some end year
            return f"EXTRACT(year FROM report_date) <= {data_years[-1]}"
        elif data_years[-1] == max(available_years):
            # Data starts at some year and goes to the end
            return f"EXTRACT(year FROM report_date) >= {data_years[0]}"
        else:
            # Data is in the middle - use BETWEEN
            return f"EXTRACT(year FROM report_date) BETWEEN {data_years[0]} AND {data_years[-1]}"
    else:
        # Non-contiguous ranges - check if we can use simple >= or <= conditions
        # by comparing the number of years to include vs exclude
        if len(data_years) <= len(null_years):
            # Fewer data years than null years - use IN
            return f"EXTRACT(year FROM report_date) IN ({', '.join(map(str, data_years))})"
        else:
            # More data years than null years - use NOT IN
            return f"EXTRACT(year FROM report_date) NOT IN ({', '.join(map(str, null_years))})"

# Test with a single table first to debug
tables = [
    #"core_eia860__scd_generators_multifuel"
    #"out_eia__monthly_generators",
    #"out_eia__yearly_generators",
    "out_eia__yearly_boilers",
    #"core_eia923__entity_coalmine",
    #"core_eia__entity_generators",
    #"core_eia__entity_boilers",
]

for table_name in tables:
    result = get_null_years(table_name, verbose=False)
    print(f"\nResult for {table_name}:")
    print(f"Number of columns with null years: {len(result)}")
    print("\nColumns and their conditions:")
    for col, condition in result.items():
        print(f"  {col}: {condition}")
```

krivard:
@zaneselvans 

> this is the notebook jank I've been using to compile the exhaustive row_conditions for the tables that weren't passing in CI

i feel better knowing it's only 7 tables and not scores of them. is it just lucky that all the tables not passing CI had `report_date`?

the code looks okay. some of the _comments_ are real misleading but i take it from "jank" that this is not intended to be reused later once we've forgetten how it works, so that's probably fine

zaneselvans:
@krivard welll, it was definitely more tables, and some of them didn't have `report_date` (there were a couple of FERC tables).

I definitely wasn't planning on re-using it when I started out, but if we did want to check for null blocks more extensively, or if the number of conditions in here that are going to have to be updated every year to deal with the fact that the EIA-860M is null for many values, then we may need to make it a real thing. What do you think?

### Tables with columns-specific `row_conditions`:
```
_core_eia860__cooling_equipment
_core_eia860__fgd_equipment
core_eia860__assn_boiler_stack_flue
core_eia860__scd_generators_multifuel
core_eia860__scd_generators_solar
core_eia860__scd_generators_wind
_core_eia923__cooling_system_information
_core_eia923__fgd_operation_maintenance
out_eia923__fuel_receipts_costs
out_ferc1__yearly_all_plants
out_ferc1__yearly_steam_plants_sched402
core_eia860__assn_boiler_generator
core_eia860__scd_boilers
core_eia860__scd_generators
core_eia860__scd_plants
core_eia860__scd_utilities
out_eia__monthly_generators
out_eia__yearly_assn_plant_parts_plant_gen
out_eia__yearly_boilers
out_eia__yearly_generators
out_eia__yearly_plant_parts
out_eia__yearly_plants
out_eia__yearly_utilities
```

### Tables with excluded columns:
```
_core_eia860__cooling_equipment
_core_eia860__fgd_equipment
core_eia860__scd_generators_energy_storage
core_eia923__entity_coalmine
out_ferc1__yearly_all_plants
out_ferc1__yearly_steam_plants_sched402
core_eia__entity_boilers
core_eia__entity_generators
out_eia__yearly_boilers
out_pudl__yearly_assn_eia_ferc1_plant_parts
```

dbt/models/eia860/core_eia860__scd_generators_multifuel/schema.yml
```diff
@@ -4,7 +4,40 @@ sources:
     tables:
       - name: core_eia860__scd_generators_multifuel
         data_tests:
-          - expect_columns_not_all_null
+          - expect_columns_not_all_null:
+              row_conditions:
```
krivard, line 8: Do we have any guesses as to why this table is so spotty?

dbt/models/eia923/out_eia923__fuel_receipts_costs/schema.yml
```diff
@@ -4,7 +4,14 @@ sources:
...
       - name: out_eia923__fuel_receipts_costs
         data_tests:
-          - expect_columns_not_all_null
+          - expect_columns_not_all_null:
+              row_conditions:
+                mercury_content_ppm: EXTRACT(year FROM report_date) NOT IN (2008, 2009, 2010, 2011, 2013)
+                natural_gas_delivery_contract_type_code: EXTRACT(year FROM report_date) >= 2014
+                moisture_content_pct: EXTRACT(year FROM report_date) >= 2014
+                chlorine_content_ppm: EXTRACT(year FROM report_date) >= 2014
+                mine_id_msha: EXTRACT(year FROM report_date) = 2025
```
krivard, line 13: I suspect we'll have a better time if we make all the 2025 entries open-ended
```suggestion
                mine_id_msha: EXTRACT(year FROM report_date) >= 2025
```

dbt/models/output/core_eia860__assn_boiler_generator/schema.yml
krivard: There's a massive change in the file naming scheme for eia860 for files before 2013, but it does feel like we should be using a default final for stuff before that

dbt/models/output/core_eia860__scd_boilers/schema.yml
krivard: ...if it turns out to be "not yet; we don't expect the rest of the 2025 data until later this year" I do not like the idea of having to apply and revert these every year, and/or having to remember to update the end date to year $NOW-1 every year

dbt/models/output/core_eia860__scd_plants/schema.yml
krivard: 3+ cases feels sufficient to add special handling. maybe an "exclude_current_year" flag, and it takes a list of columns

dbt/models/output/core_eia__entity_generators/schema.yml
krivard: Blocking: that needs to go in the release notes!

dbt/models/output/out_eia__yearly_boilers/schema.yml
```diff
@@ -4,7 +4,106 @@ sources:
     tables:
       - name: out_eia__yearly_boilers
         data_tests:
-          - expect_columns_not_all_null
+          - expect_columns_not_all_null:
+              description: >
+                Excluded mercury control strategies are very infrequently used, since
+                few boilers have more than three. It's fine if they're null. The boiler
+                manufacturer and code are static attributes which were only reported in
+                a single year and are effectively foreward/back filled because they are
```
krivard:
```suggestion
                a single year and are effectively forward/back filled because they are
```
(would be cool if `typos` could be configured to *just* check description fields in schema yml)

dbt/models/ferc1/out_ferc1__yearly_all_plants/schema.yml
```diff
@@ -4,7 +4,22 @@ sources:
     tables:
       - name: out_ferc1__yearly_all_plants
         data_tests:
-          - expect_columns_not_all_null
+          - expect_columns_not_all_null:
+              description: >
+                Exactly which columns contain rolling averages will depend on what years
```
krivard: presumably the set of columns with rolling averages doesn't change
```suggestion
                Exactly which years have rolling averages will depend on what years
```

dbt/models/output/out_pudl__yearly_assn_eia_ferc1_plant_parts/schema.yml
```diff
@@ -4,7 +4,16 @@ sources:
     tables:
       - name: out_pudl__yearly_assn_eia_ferc1_plant_parts
         data_tests:
-          - expect_columns_not_all_null
+          - expect_columns_not_all_null:
+              description: >
+                Exactly which columns contain rolling averages will depend on what years
```
krivard: presumably the set of columns with rolling averages doesn't change
```suggestion
                Exactly which years have rolling averages will depend on what years
```

migrations/versions/d0c5180c12f0_.py
```diff
@@ -0,0 +1,24 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    pass
+
+
+def downgrade() -> None:
+    pass
```
krivard, line 24: if you're pals with alembic and can easily drop this empty migration, please do, but if it'll be an immense pain it can stay

migrations/versions/d9eb718658b0_consolidate_operating_swtich_and_can_.py
```diff
@@ -0,0 +1,50 @@
...
+down_revision = 'd0c5180c12f0'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('core_eia__entity_generators', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('can_switch_when_operating', sa.Boolean(), nullable=True, comment='Indicates whether a fuel switching generator can switch fuels while operating.'))
+        batch_op.drop_column('operating_switch')
```
krivard, line 23: this looks less like consolidating two existing columns and more like replacing an old column with a brand new one. if it does the right thing under the hood, great, just flagging in case it's not doing what you expected it to do

src/pudl/analysis/mcoe.py
```diff
@@ -712,11 +712,12 @@ def mcoe(
...
         # Make sure the merge worked!
         .pipe(
-            pudl.validate.no_null_rows,
+            pv.no_null_rows,
             df_name="fuel_cost + capacity_factor",
             max_null_fraction=0.9,
         )
-        .pipe(pudl.validate.no_null_cols, df_name="fuel_cost + capacity_factor")
+        # This is the only remaining use of this function. Remove? Rename?
+        .pipe(pv.no_null_cols, df_name="fuel_cost + capacity_factor")
```
krivard: not equipped to answer this; lmk if you want feedback on it and I can dig for the necessary context

src/pudl/analysis/timeseries_cleaning.py
krivard: many blessings upon you; that future warning was annoying

migrations/versions/52a6b00f2f0a_fix_nrel_tax_credit_case_enum.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('core_nrelatb__yearly_projected_financial_cases_by_scenario', schema=None) as batch_op:
+        batch_op.alter_column('model_tax_credit_case_nrelatb',
+               existing_type=sa.Enum('ITC', 'PTC + ITC', 'PTC'),
+               type_=sa.VARCHAR(length=6),
```
krivard, line 41: flagging that this downgrade resets to VARCHAR and not the previous enum `{"enum": ["Market", "R&D"]}`

I suspect this column was created before enums got propagated into the db via alembic, so i don't know what the right call is here. probably just live with it 😞

src/pudl/transform/eia860.py
krivard: ooh that is a very nice catch!

i'm confused about the change in ordering though; is there a system here or is it just differently-random now?

krivard (commented):
* release notes needed for column changes
* probably need to resolve the "except current year" tagging and/or workflow before merge
* alembic did multiple funky things; ptasl
* I like it when a table with row conditions has a description saying why the conditions are needed but recognize doing this everywhere might be a huge burden. It does make tables with many many row conditions significantly less overwhelming & significantly more legible, but idk -- do what you feel is maintainable and we'll make it work

dbt/models/output/core_eia860__assn_boiler_generator/schema.yml
zaneselvans: IIRC we are entirely responsible for the `data_maturity` value -- and we assign it based on metadata associated with the Zenodo archive / the filenames, which is why I'm surprised here.

migrations/versions/d9eb718658b0_consolidate_operating_swtich_and_can_.py
zaneselvans: Ah, by "consolidating" meant that we had 2 different column names in the past for the same semantic column (`operating_switch` and `can_switch_when_operating`) and I am renaming all of the `operating_switch` instances to `can_switch_when_operating`.

src/pudl/transform/eia860.py
zaneselvans: It makes no sense -- it's just the ordering of an exhaustive list that I was running through somewhere else and having it be the same made it easier to make sure I didn't miss anything.

migrations/versions/d0c5180c12f0_.py
zaneselvans: I **think** this is how it's supposed to be. This is an alembic schema merge -- there were two different branches of the schema that needed to be brought together. So what it's encoding isn't changes to the schema, but the fact that the new `HEAD` has two distinct ancestors.

dbt/models/output/out_pudl__yearly_assn_eia_ferc1_plant_parts/schema.yml
zaneselvans: Fixed.

dbt/models/ferc1/out_ferc1__yearly_all_plants/schema.yml
zaneselvans: Fixed.

dbt/models/output/out_eia__yearly_boilers/schema.yml
zaneselvans: Fixed!

zaneselvans:
Added release notes highlighting the column names and oopsies all-null columns that are being fixed by this PR.

migrations/versions/52a6b00f2f0a_fix_nrel_tax_credit_case_enum.py
zaneselvans: Hmm, I don't understand why it would have done this... But also, the old Enum resulted in nulling out these columns entirely, so don't think we lose anything in the downgrade here. And also we rarely downgrade so... hopefully no damage done.

zaneselvans:
@e-belfer @krivard Okay, it seems like there are enough tables with a bunch of exceptions that only serve to avoid complaining about columns which are null b/c of the most recent partial year from EIA-860M that it'll be annoying to do updates if we don't handle them programmatically somehow. A few options:

### Column-level exceptions
* Keep the specific expectations, and commit to doing the manual updates (brittle and labor intensive).
* Keep the specific expectations, and make a real script for doing the "please enumerate all the null blocks" work. This would be the strictest expectation, but also probably brittle and with have high maintenance overhead (though a different kind from the first option).

### Table-level exceptions (my preference I think)
* Create a table-level exception that we could apply just to the tables which are heavily impacted by the null values which result from EIA-860M not reporting all the EIA-860 columns.
* This could be dynamic (e.g. by querying `EXTRACT(year FROM MAX(report_date))` in the EIA-860M changelog table or something.) and prevent us from needing to do annual updates, but would still prevent us from checking for null values in columns that we *do* expect the EIA-860M to populate in the most recent year. But catching those can only really be done with an annoyingly large number of column-level exceptions.
* This could also be done with a simple "Ignore all-null columns in 2025" but we would need to remember to update it.

### Blanket exception
Create a blanket exception for the most recent (partial) year (currently 2025) in all tables. This would be simple and get rid of a significant number of the special cases, but would also ignore some column-years where we shouldn't find 100% null values (e.g. in the EIA-930 and EPA CEMS tables)

src/pudl/analysis/mcoe.py
zaneselvans: My gut says this is superfluous -- we'll be applying the no-null-cols check to the resulting table(s) so we'll catch an error if there is one. We won't catch it **during** the ETL. But we would catch it in the merge queue if not by running the integration tests locally.

zaneselvans: So... Imma remove it. But @cmgosnell or @e-belfer if you have feels on this, I can put it back.

zaneselvans:
Okay now I have a special case that allows us to ignore the null values that result from the EIA-860M partial years, but I think it's not quite right... since there will be parts of the year during which the max EIA-860M year, and the current EIA-860 year are actually the same... so the test needs to allow nulls in the max EIA-860M year only when it's greater than the max EIA-860 year.... right?

Or wait, maybe that's never actually the case...
- In January 2025, the most recent EIA-860M data will be from 2024. But the most recent annual EIA-860 data will be from 2023.
- In March to May 2025, the most recent EIA-860M data would probably be from 2025, but the most recent EIA-860 data will actually still be from 2023... so we need to allow null column-years in 2024 **and** 2025.
- Around June or July, 2025, we'll get Early Release 2024 data, at which point the only allowable EIA-860M null block will become 2025.
- This will continue to be the case in late 2025, when there's final 2024 data from EIA-860 available, and the EIA-860M data is still 2025.

So we need to be able to accommodate either 1 or 2 years of NULL column values resulting from EIA-860M running ahead of EIA-860. So rather than defining `current_year` in the test, we need to define a list of 1 or 2 `eia860m_only_years` in which to be lenient about NULL columns.

But to do that I need to be able to authoritatively differentiate between EIA-860 and EIA-860M years, based only on the contents of the PUDL DB... what's the right way to do that?

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
```diff
@@ -0,0 +1,93 @@
...
+{% endif %}
+
+{% if execute %}
+    {% set columns_result = run_query(get_columns_query) %}
+    {% set column_names = columns_result.columns[0].values() %}
+
+    {# Validate that report_date column exists if ignore_eia860m_nulls is enabled #}
+    {% if ignore_eia860m_nulls and 'report_date' not in column_names %}
+        {{ exceptions.raise_compiler_error("ignore_eia860m_nulls=true requires a 'report_date' column in the table being tested, but 'report_date' was not found in " + model.name) }}
+    {% endif %}
```
zaneselvans, line 38: I'm not thrilled about this dependency, but it works for all the current tables that we need it to, and I **suspect** that almost any table that runs into the problem of having nulls due to EIA-860M not reporting everything in EIA-860 will also have `report_date`

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
```diff
@@ -0,0 +1,93 @@
...
+    SELECT column_name
+    FROM (DESCRIBE {{ model }})
+    WHERE column_name NOT IN ('{{ exclude_columns | join("', '") }}')
+{% endset %}
+
+{% if ignore_eia860m_nulls %}
+    {% set get_last_eia860_year_query %}
+        SELECT EXTRACT(year FROM MAX(report_date)) as last_eia860_year
+        FROM {{ source('pudl', 'core_eia860__scd_ownership') }}
+    {% endset %}
```
zaneselvans, line 19: This ugly ad-hoc solution works, but makes me wonder if it would be helpful if we actually wrote some canonical metadata into the PUDL DB so that we could reference it when needed -- e.g. raw archive partitions that were processed, DOIs of the input archives, hash and ref of the git commit that the DB was built from, etc.

src/pudl/analysis/mcoe.py
```diff
@@ -712,11 +712,10 @@ def mcoe(
         )
         # Make sure the merge worked!
         .pipe(
-            pudl.validate.no_null_rows,
+            pv.no_null_rows,
             df_name="fuel_cost + capacity_factor",
             max_null_fraction=0.9,
         )
-        .pipe(pudl.validate.no_null_cols, df_name="fuel_cost + capacity_factor")
```
zaneselvans, line 719: I went ahead and removed this function, since this is purely defensive, currently passing, and we'll catch the same error with the no-null-cols check (just a bit later in the build process)

src/pudl/validate.py
```diff
@@ -63,38 +63,6 @@ def no_null_rows(
     return df
 
 
-def no_null_cols(
```
zaneselvans, line 66: byeeeee.

dbt/models/output/out_eia__yearly_boilers/schema.yml
zaneselvans: With the updated test and `ignore_eia860m_nulls: true` the conditions here are simplified and shouldn't need to be updated every year.

dbt/models/output/core_eia860__scd_plants/schema.yml
zaneselvans: Well, the new version of the test prevents us from needing to update this every year, though there are still quite a few `row_conditions` -- but they contain real information about our expectations of the data

dbt/models/output/core_eia860__scd_boilers/schema.yml
zaneselvans: Now the test programmatically excludes any years which exist only in the 860M, and not the 860.

zaneselvans (commented):
If this version of the test and the simplified `row_conditions` seems good enough, I think I'm inclined to merge it in, and split out the remaining data weirdness that the test has highlighted for investigation in a followup issue. What do you think @krivard?

zaneselvans:
I went ahead and cleaned up my notebook jank and turned it into a real script. I suspect it needs some unit tests in order to keep meeting 93% coverage though.

dbt/models/output/core_eia860__scd_boilers/schema.yml
krivard: 🏆

dbt/models/output/out_eia__yearly_generators/schema.yml
```diff
@@ -27,46 +26,40 @@ sources:
...
+                planned_new_capacity_mw: EXTRACT(year FROM report_date) >= 2013
+                planned_new_prime_mover_code: EXTRACT(year FROM report_date) >= 2007
+                planned_repower_date: EXTRACT(year FROM report_date) >= 2007
                 planned_uprate_date: EXTRACT(year FROM report_date) >= 2007
-                reactive_power_output_mvar: EXTRACT(year FROM report_date) IN (2004, 2005, 2006, 2016, 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024)
-                startup_source_code_1: EXTRACT(year FROM report_date) BETWEEN 2009 AND 2024
-                startup_source_code_2: EXTRACT(year FROM report_date) BETWEEN 2009 AND 2024
-                startup_source_code_3: EXTRACT(year FROM report_date) BETWEEN 2009 AND 2024
-                startup_source_code_4: EXTRACT(year FROM report_date) BETWEEN 2009 AND 2024
+                reactive_power_output_mvar: EXTRACT(year FROM report_date) IN (2004, 2005, 2006) OR EXTRACT(year FROM report_date) >= 2016
```
krivard, line 49: nice split 👍

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
krivard: and it'll call for help if it runs into trouble. this is a good approach until we find we need something better

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
```diff
@@ -11,42 +12,71 @@
...
+        {% set exclusion_note = "" %}
+    {% endif %}
+
+    {# Set failure reason and row_condition display #}
+    {% if column_name in row_conditions %}
+        {% set failure_reason = "Conditional check failed: " + row_conditions[column_name] + exclusion_note %}
+        {% set row_condition_display = row_conditions[column_name] | replace("'", "''") %}
+    {% else %}
+        {% set failure_reason = "Column is entirely NULL" + exclusion_note %}
+        {% set row_condition_display = "N/A (entire table)" if not exclusion_note else "EXTRACT(year FROM report_date) <= " + last_eia860_year|string %}
```
krivard: optional refinement: put `"EXTRACT(year FROM report_date) <= " + last_eia860_year|string` in a variable and use it both here and in `where_condition` above so we don't have to sync them up manually. Definitely hoping they never have to change, but if they do..

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
krivard: i'm always a fan of recording provenance. oos for this pr though

dbt/tests/data_tests/generic_tests/schema.yml
```diff
@@ -69,7 +69,9 @@ macros:
     description: >
       Check that all columns in a table are not entirely NULL. Can exclude specific
       columns and apply conditional checks (e.g., only check for NULLs when certain
-      conditions are met).  Uses DuckDB's DESCRIBE to dynamically discover column names.
+      conditions are met). Uses DuckDB's DESCRIBE to dynamically discover column names.
+      When ignore_eia860m_nulls is enabled, automatically excludes the current year
+      from null checks to accommodate EIA-860M data limitations.
```
krivard: May as well say what it actually does, so we don't muddy future debugging journeys
```suggestion
      When ignore_eia860m_nulls is enabled, automatically excludes years which are
      only available in EIA-860M.
```

docs/release_notes.rst
```diff
@@ -18,23 +18,44 @@ Quality of Life Improvements
...
+issue :issue:`4105` and PR :pr:`4382`. As a result of more broadly applying this check,
+we found and fixed a few data quality and column naming issues resulting in minor
+changes to the database schema:
+
+* ``id_dc_coupled_tightly`` was renamed to ``is_dc_coupled_tightly`` (typo).
+* ``switch_operating`` was consolidated with the existing
+  ``can_switch_when_operating`` column found in the multi-fuel generator tables.
+* The ``model_tax_credit_case_nrelatb`` column had its allowable enumerated values
+  corrected, resulting in real non-null contents. See PR :pr:`4384`.
+* Three previously entirely null ``boolean`` columns in the mutlifuel generator table
```
```suggestion
* Three previously entirely null ``boolean`` columns in the multifuel generator table
```

src/pudl/scripts/pudl_null_cols.py
```diff
@@ -0,0 +1,228 @@
...
+    """Find years where a specific column is entirely null."""
+    pq_path = PudlPaths().parquet_path(table_name)
+    year_expr = date_column_to_year_expr(date_column)
+    where_clause = f"WHERE {year_expr} <= {max_year}" if max_year else ""
+    sql = f"""
+    SELECT CAST({year_expr} AS INTEGER) AS year
+    FROM '{pq_path}'
+    {where_clause}
+    GROUP BY {year_expr}
+    HAVING COUNT({column}) = 0
```
krivard, line 50: 👍 this is a better version of the query than before

src/pudl/scripts/pudl_null_cols.py
```diff
@@ -0,0 +1,228 @@
...
+        if data_years[0] == min(available_years):
+            # Data starts at beginning - prefer <= for compatibility
+            return f"{year_expr} <= {data_years[-1]}"
+        if data_years[-1] == max(available_years):
+            # Data goes to the end of our filtered range - use >=
+            return f"{year_expr} >= {data_years[0]}"
+        # Data is in a finite middle range - use BETWEEN
+        return f"{year_expr} BETWEEN {data_years[0]} AND {data_years[-1]}"
+
+    # If we don't have a contiguous range, use the more compact of IN or NOT IN:
```
krivard, line 153: 👍 good comment fix

src/pudl/scripts/pudl_null_cols.py
```diff
@@ -0,0 +1,228 @@
...
+        exclusion
+
+    """
+    # Get the years where data exists (inverse of null_years)
+    data_years = sorted([year for year in available_years if year not in null_years])
+    year_expr = date_column_to_year_expr(date_column)
+
+    if not data_years:
+        logger.warning("No data found! Column should be excluded or debugged.")
+        return "FALSE -- Column has no data. Exclude it explicitly or debug."
```
krivard: Should this get surfaced in a docstring or --help text, or can we assume it'll come up at code review time? Something like "Generates row_condition FALSE if no data is found, which will exclude the column from null checks if left as-is"

test/unit/scripts/test_pudl_null_cols.py
```diff
@@ -0,0 +1,322 @@
...
+        ("report_year", "report_year"),
+        ("year", "year"),
+        ("fiscal_year_end", "fiscal_year_end"),
+        ("report_date", "EXTRACT(year FROM report_date)"),
+        ("start_date", "EXTRACT(year FROM start_date)"),
+        ("timestamp", "EXTRACT(year FROM timestamp)"),
+    ],
+)
+def test_date_column_to_year_expr(input_column, expected_output):
+    """Test the date_column_to_year_expr function with various inputs."""
```
krivard:
```suggestion
    """Test that date_column_to_year_expr uses EXTRACT when appropriate."""
```

docstrings should probably say something not already in the function signature

test/unit/scripts/test_pudl_null_cols.py
```diff
@@ -0,0 +1,322 @@
...
+            [2020, 2021, 2022],
+            "report_year",
+            "report_year = 2021",
+        ),
+    ],
+)
+def test_compact_row_condition(
+    null_years, available_years, date_column, expected_output
+):
+    """Test the compact_row_condition function with various scenarios."""
```
krivard:
```suggestion
    """Test edge cases for compact_row_condition."""
```

test/unit/scripts/test_pudl_null_cols.py
```diff
@@ -0,0 +1,322 @@
...
+@pytest.mark.parametrize(
+    "column,expected_null_years",
+    [
+        ("column_a", [2021]),  # null in 2021
+        ("column_b", [2020, 2022]),  # null in 2020, 2022
+        ("column_c", []),  # no null years
+    ],
+)
+def test_get_null_years(mocker, parquet_data, column, expected_null_years):
+    """Test get_null_years function with different columns."""
```
krivard:
```suggestion
    """Test that get_null_years really does exclude null years."""
```

test/unit/scripts/test_pudl_null_cols.py
```diff
@@ -0,0 +1,322 @@
...
+    mock_pudl_paths.return_value = mock_paths
+
+    null_years = get_null_years(
+        table_name="test_table", column=column, date_column="report_date"
+    )
+    assert null_years == expected_null_years
+
+
+def test_get_null_years_with_max_year(mocker, parquet_data):
+    """Test get_null_years function with max_year constraint."""
```
krivard:
```suggestion
    """Test that get_null_years obeys the max_year constraint."""
```

test/unit/scripts/test_pudl_null_cols.py
```diff
@@ -0,0 +1,322 @@
...
+        table_name="test_table",
+        column="column_b",
+        date_column="report_date",
+        max_year=2021,
+    )
+    assert null_years == [2020]  # 2022 is excluded by max_year
+
+
+def test_get_available_years(mocker, parquet_data):
+    """Test get_available_years function."""
```
krivard:
```suggestion
    """Test that get_available_years really does list all years present in the date column."""
```

test/unit/scripts/test_pudl_null_cols.py
```diff
@@ -0,0 +1,322 @@
...
+    mock_pudl_paths.return_value = mock_paths
+
+    available_years = get_available_years(
+        table_name="test_table", date_column="report_date"
+    )
+    assert available_years == [2020, 2021, 2022]
+
+
+def test_get_available_years_with_max_year(mocker, parquet_data):
+    """Test get_available_years function with max_year constraint."""
```
krivard:
```suggestion
    """Test that get_available_years obeys the max_year constraint."""
```

test/unit/scripts/test_pudl_null_cols.py
```diff
@@ -0,0 +1,322 @@
...
+    mock_pudl_paths.return_value = mock_paths
+
+    available_years = get_available_years(
+        table_name="test_table", date_column="report_date", max_year=2021
+    )
+    assert available_years == [2020, 2021]
+
+
+def test_infer_row_conditions(mocker):
+    """Test the infer_row_conditions function."""
```
krivard, line 228: What is this test exercising that's not already covered by `test_compact_row_condition`?

test/unit/scripts/test_pudl_null_cols.py
```diff
@@ -0,0 +1,322 @@
...
+    result = runner.invoke(
+        main, ["test_table", "--ignore-eia860m", "--max-year", "2021"]
+    )
+
+    assert result.exit_code != 0
+    assert "mutually exclusive" in result.output
+
+
+def test_main_yaml_output(mocker):
+    """Test that CLI outputs valid YAML."""
```
krivard:
```suggestion
    """Test that CLI converts row conditions to YAML syntax."""
```

(we're not running a validator here)

dbt/models/eia923/out_eia923__fuel_receipts_costs/schema.yml
zaneselvans: Something is weird with this one in particular -- it's in my to-investigate list.

dbt/tests/data_tests/generic_tests/expect_columns_not_all_null.sql
zaneselvans: Good catch.

src/pudl/scripts/pudl_null_cols.py
zaneselvans: I actually learned something!

src/pudl/scripts/pudl_null_cols.py
zaneselvans: Ah, this is half-baked. What I wanted to do was require the user to either debug the entirely null column (rendering it no longer entirely null) or explicitly exclude it from the test, in effect saying "Yeah, I know it's all null or irregularly null and that's fine." So maybe this should have been TRUE and not FALSE...

test/unit/scripts/test_pudl_null_cols.py
zaneselvans: The idea is that it's running through the other logic that's in the `infer_row_conditions` function, but the truth is I don't really understand whether that's happening, given all the mocking.

src/pudl/scripts/pudl_null_cols.py
zaneselvans: Because this is AND-ed with another condition, neither TRUE nor FALSE gives the desired behavior but I can cause the test to fail with some bad syntax like ERROR...

zaneselvans: Maybe I'm overthinking this -- I'm trying to avoid the scenario where there's a fully null column, and this generates a condition that then... doesn't check that column, and that's overlooked, when really what needs to happen is a human needs to take a look at it, which should really happen with all of these generated conditions, but especially in the case of any truly all-NULL column, since those are very likely to indicate a real error.

zaneselvans: Thinking about it more, I think causing the tests to error out in this case is fine -- we're *generating* test conditions here, which need to be reviewed, and this is a flag for extra human 👀

zaneselvans: And I think it makes more sense to write it out as a condition which will generate an error than to raise an error in the script, since where this is happening there's no context on what column it is that is causing the problem.

PR #4387, opened 2025-07-03
feat: translate dagster asset selections to dbt node selections

### LINKED ISSUE #4132: Run DBT tests for assets that are related in Dagster

jdangerx:
# Overview

Closes #4132 . For real this time.

We work on dagster assets but have to run validation tests with dbt nodes. dbt has no idea what the dependencies in dagster are so it's hard to select "all downstream assets" or somesuch. Plus the dbt names are much longer.

## What did you change?

* add a flag to dbt_helper validate to pass in dagster selections instead
* add stub translation logic to `dbt_wrapper.py`
* add stub unit tests

## Documentation

# Testing

## To-do list

- [x] make a fake manifest for testing - maybe make the yamls, and then write them / run dbt parse in a temp dir so we can make sure it's a real manifest.
  file structure:

  ```
  your_dbt_project/
  ├── dbt_project.yml            # Required: Project configuration
  ├── models/                    # Required: Directory for models (can be empty)
  └── models/sources.yml         # Defines your sources
  ```

  source should only include cell and animal. dna is like a 'raw' asset that doesn't get persisted. maybe i should just rename those to raw/core/out haha.
- [x] grab the source selector out of the `manifest.json`'s `sources` map using `f"source:{package_name}.{schema}.{name}"`. make a map from the dbt source's name to the selector. then for each asset, try to grab a selector.
- [x] some dagster assets won't have a corresponding dbt source, log out the tables you'll be checking and which you'll be skipping.

### DIFF
--- a/dbt/README.md
+++ b/dbt/README.md
@@ -64,17 +64,22 @@ dbt_helper update-tables --help
 
 #### `validate`
 
-If you want to check if a materialized asset passes the validation tests defined in DBT, you can use `dbt_helper validate`. You can pass in the `--select` parameter to filter to a specific set of validation tests, and `--target` to set the DBT target.
+If you want to check if a materialized asset passes the validation tests defined in DBT, you can use `dbt_helper validate`.
 
-This will run the associated tests, then print out the test query output so you can see what went wrong.
+This understands how to translate dagster asset selections into DBT node selections, and does some extra legwork to make the test outputs more informative.
+
+See `dbt_helper validate --help` for usage details.
 
 Example usage:
 
 ```bash
-dbt_helper validate --select "source:pudl_dbt.pudl.out_eia__yearly_generators"
+dbt_helper validate --asset-select "key:out_eia__yearly_generators" # for just this asset
+dbt_helper validate --asset-select "+key:out_eia__yearly_generators" # for this asset as well as all upstream assets
+dbt_helper validate --asset-select "+key:out_eia__yearly_generators" --exclude "*check_row_counts*"  # skip rowcounts
+dbt_helper validate --select "source:pudl_dbt.pudl.out_eia__yearly_generators" # if you want to select a DBT node in particular
 ```
 
-See [official selection syntax documentation](https://docs.getdbt.com/reference/node-selection/syntax) for details.
+See [dbt selection syntax documentation](https://docs.getdbt.com/reference/node-selection/syntax) and [Dagster selection syntax documentation](https://docs.dagster.io/guides/build/assets/asset-selection-syntax/reference) to see all the possibilities.
 
 ## Adding tests
 
--- a/src/pudl/dbt_wrapper.py
+++ b/src/pudl/dbt_wrapper.py
@@ -13,6 +13,7 @@ from dbt.artifacts.schemas.run import RunExecutionResult
 from dbt.cli.main import dbtRunner, dbtRunnerResult
 from dbt.contracts.graph.nodes import GenericTestNode
 
+import pudl.etl
 from pudl.logging_helpers import get_logger
 from pudl.workspace.setup import PUDL_ROOT_PATH, PudlPaths
 
@@ -147,7 +148,6 @@ def build_with_context(
         success=build_output.success,
         failure_contexts=compiled_sql_contexts + weighted_quantile_contexts,
     )
-    return BuildResult(success=build_output.success, failure_contexts=failure_contexts)
 
 
 def dagster_to_dbt_selection(
@@ -155,13 +155,15 @@ def dagster_to_dbt_selection(
 ) -> str:
     """Translate dagster asset selection to db node selection.
 
+    We use the dbt manifest to determine which sources are defined in dbt so
+    that we can map them to dagster assets. So, we need to generate a fresh dbt
+    manifest via ``dbt parse`` whenever we run this function.
+
     * turn asset selection into asset keys
     * turn asset keys into node names
     * turn node names into selection string
     """
     if defs is None:
-        import pudl.etl
-
         defs = pudl.etl.defs
 
     asset_keys = dg.AssetSelection.from_string(selection).resolve(
@@ -180,7 +182,7 @@ def dagster_to_dbt_selection(
 
     # all dagster assets are treated as sources so we only have to look here.
     dbt_node_selectors = [
-        f"source:{s['package_name']}.{s['source_name']}.{s['name']}"
+        f"source:{s['source_name']}.{s['name']}"
         for s in manifest["sources"].values()
         if s["name"] in asset_names
     ]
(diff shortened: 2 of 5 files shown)


dbt/README.md
```diff
@@ -64,17 +64,22 @@ dbt_helper update-tables --help
 
 #### `validate`
 
-If you want to check if a materialized asset passes the validation tests defined in DBT, you can use `dbt_helper validate`. You can pass in the `--select` parameter to filter to a specific set of validation tests, and `--target` to set the DBT target.
+If you want to check if a materialized asset passes the validation tests defined in DBT, you can use `dbt_helper validate`.
 
-This will run the associated tests, then print out the test query output so you can see what went wrong.
+This understands how to translate dagster asset selections into DBT node selections, and does some extra legwork to make the test outputs more informative.
+
+See `dbt_helper validate --help` for usage details.
```
jdangerx, line 71: Decided to mostly just leave a pointer here - `dbt_helper validate --help` is (a) closer to the code so easier to maintain and (b) closer to the terminal where people are already typing `dbt_helper validate whatever`.

src/pudl/dbt_wrapper.py
```diff
@@ -86,3 +88,40 @@ def build_with_context(
...
+        defs = pudl.etl.defs
+
+    asset_keys = dg.AssetSelection.from_string(selection).resolve(
+        defs.resolve_asset_graph()
+    )
+    asset_names = {asset_key.to_user_string() for asset_key in asset_keys}
+
+    if manifest is None:
+        manifest_path = PUDL_ROOT_PATH / "dbt" / "target" / "manifest.json"
+        with chdir(PUDL_ROOT_PATH / "dbt"):
```
jdangerx, line 176: It can be a little slow to run `dbt parse` every time, but if we want to make sure the manifest is up-to-date and we're not dropping any tests we have to do it. Or, we can go and check to see if the manifest is fresher than all the schemas, but that seems fragile so I decided against it for now.

src/pudl/scripts/dbt_helper.py
```diff
@@ -502,32 +503,97 @@ def update_tables(
...
+        if asset_select is not None:
+            raise click.UsageError(
+                "You can't pass --select and --asset-select at the same time."
+            )
+        node_selection = select
+    else:
+        if asset_select is not None:
+            node_selection = dagster_to_dbt_selection(asset_select)
+        else:
+            node_selection = "*"
```
jdangerx, line 587: I manually tested all four of these cases, but for some reason when using the `click.testing.CliRunner()` in tests they kept complaining about 

```
             finally:                                                                                                                         
                 sys.stdout.flush()                                                                                             
 >               stdout = outstreams[0].getvalue()                                                                          
                          ^^^^^^^^^^^^^^^^^^^^^^^^                                                                    
 E               ValueError: I/O operation on closed file.                                                                       
                                                                                                                                         
 ../../micromamba/envs/pudl-dev/lib/python3.12/site-packages/click/testing.py:438: ValueError  
```

I went down all sorts of rabbit holes trying to chase this down but eventually gave up, reasoning that "if it's broken and people are using it, they will notice; if it's broken and people aren't using it, it doesn't matter."

dbt/README.md
e-belfer: Seems wise, no problem on my end with doing that.

src/pudl/dbt_wrapper.py
```diff
@@ -86,3 +88,40 @@ def build_with_context(
...
+    selection: str, defs: dg.Definitions | None = None, manifest=None
+) -> str:
+    """Translate dagster asset selection to db node selection.
+
+    * turn asset selection into asset keys
+    * turn asset keys into node names
+    * turn node names into selection string
+    """
+    if defs is None:
+        import pudl.etl
```
e-belfer: I'm assuming this is a big one to import and thus you're only doing it when needed. Can you confirm my intuition here?

src/pudl/scripts/dbt_helper.py
```diff
@@ -502,32 +503,97 @@ def update_tables(
...
+
+    Usage examples:
+
+    Run all the checks for one asset:
+
+        $ dbt_helper validate --asset-select "key:out_eia__yearly_generators"
+
+    Run the checks for one specific dbt node:
+
+        $ dbt_helper validate --select "source:pudl_dbt.pudl.out_eia__yearly_generators"
```
krivard, line 564: drive-by

are both of these path prefixes needed for it to work with dbt_helper? when just using the dbt command line I omit pudl_dbt like so:

```suggestion
        $ dbt_helper validate --select "source:pudl.out_eia__yearly_generators"
```

src/pudl/scripts/dbt_helper.py
```diff
@@ -502,32 +503,97 @@ def update_tables(
...
-    "--select", default="*", help="DBT selector for the asset(s) you want to validate."
+    "--select",
+    help="DBT selector for the asset(s) you want to validate. Syntax "
+    "documentation at https://docs.getdbt.com/reference/node-selection/syntax",
+)
+@click.option(
+    "--asset-select",
+    "-a",
+    help=(
+        "*DAGSTER* selector for the asset(s) you want to validate."
```
e-belfer: Run `--help` and take a look at the rendering - a bunch of things are missing spaces between words (e.g., Syntaxdocumentation in --exclude, canuse in --asset-select), and this should get fixed!

src/pudl/scripts/dbt_helper.py
```diff
@@ -502,32 +503,97 @@ def update_tables(
...
+    help="DBT selector for the asset(s) you want to exclude from validation. Syntax"
+    "documentation at https://docs.getdbt.com/reference/node-selection/syntax",
 )
 @click.option(
     "--target",
     default="etl-full",
     type=click.Choice(["etl-full", "etl-fast"]),
     help="DBT target - etl-full (default) or etl-fast.",
 )
+@click.option("--dry-run/--wet-run", default=False)
```
e-belfer: Can you add a description for this field, please?

src/pudl/dbt_wrapper.py
jdangerx: That is a good reason to do this import here - but I just tested it and importing `pudl.etl` after `pudl` has already been imported doesn't really take much time (like half a second... which is annoying but trivial in the grand scheme of "`pudl` takes 5-10s to import anyways"). So I moved the import to a less surprising place up top.

PR #4388, opened 2025-07-06
Update conda lockfile for week of 2025-07-06

pudlbot:


dbt/models/eia923/core_eia923__monthly_generation_fuel_nuclear/schema.yml
```diff
@@ -15,7 +15,7 @@ sources:
           - name: fuel_type_code_pudl
             data_tests:
               - dbt_expectations.expect_column_values_to_be_in_set:
-                  value_set: ["nuclear"]
+                  value_set: ["nuclear", "oil"]
```
zaneselvans, line 18: This seems weird, but it's because there's diesel backup power at nuclear plants that gets used rarely.

pyproject.toml
```diff
@@ -403,6 +403,7 @@ BCK = "BCK"
 consit = "consit"
 cll = "cll"
 EDE = "EDE"
+fillin = "fillin"
```
zaneselvans, line 406: Not sure why this wasn't triggering the spell-checker before.

.pre-commit-config.yaml
```diff
@@ -53,7 +53,7 @@ repos:
 
   # Spell check our documentation and code comments
   - repo: https://github.com/crate-ci/typos
-    rev: v1.24.5
+    rev: v1.34.0
```
zaneselvans, line 56: For reasons I don't understand, `pre-commit autoupdate` tried to change this to `v1` rather than a specific version, which `pre-commit` does not support -- the ref has to point at a single, stable version.  Updating manually but not sure what will happen in the future.

zaneselvans (approved):

.pre-commit-config.yaml
zaneselvans: Switched to using a mirror of the `typos` pre-commit hook to avoid the weird [floating v1 tag](https://github.com/crate-ci/typos/issues/390) issue.

PR #4392, opened 2025-07-08
fix the labeling of early release and provisional for eia923

### LINKED ISSUE #4328: Fix error in labeling `data_maturity` EIA 923’s December monthly update
### Describe the bug
We noticed an edge case in the way we label the `data_maturity` on the EIA 923 Monthly update. During our last quarterly update of EIA we integrated monthly EIA 923 data from [this archive](https://zenodo.org/records/15294970).

[Our current pattern](https://github.com/catalyst-cooperative/pudl/blob/b4bb8d7dfe1cd03672f02f3720b4869ea332b4aa/src/pudl/extract/excel.py#L168-L169) for labeling data maturity assumes that if the month is anything other than 12 its a monthly/ytd update.

But that's 

### Bug Severity
 - **Low:** The bug isn't causing me problems, but something's still wrong here. Data is all there but its labeled incorrectly...

cmgosnell:
# Overview

Closes #4328.

- when a ytd december 923 data is released, our old assumption was that that was a final release.. which makes sense bc we made that assumption before we were doing the quarterly updates
- also also the environmental information files were not being labeled as provisional when they are early release because these file names have a space instead of an underscore.... womp womp.

## What did you change?
- i slightly simplified how we check for ytd updates by just checking if there is no "final" in the file name.
- i added the "early release" option for provisional
- i added a unit test (which required adding to our `package_data/test` files

## Documentation

# Testing

## To-do list


src/pudl/extract/excel.py
```diff
@@ -161,17 +160,22 @@ def add_data_maturity(
...
-            if release_month != "12":
-                maturity = "incremental_ytd"
+        # The core EIA923 table we've been pulling from has EIA923_Schedules_2_3_4_5_M_
+        # in the beginning, and the environmental info file usually starts with
+        # EIA923_Schedule_8_Annual_Envir_Info (but w/ an r)... So we use EIA923_Schedule
+        # w/o the s and it seems consistent that the file names either end with
+        # "Early_Release" or "Final" or "Final_Revision". The ytd files seem to just
+        # have early release (w/ or w/o the space)
+        # We could use r"EIA923_Schedule(.*)(\d{4})(.*)(.xlsx|.xls)" as a pattern for
+        # all recent 923 files, but that seems more brittle.
```
cmgosnell, line 172: i think this is an easy way to share this example: https://regex101.com/r/wRL0Bo/1

.... should we do this w/ regex instead so it does indeed break if/when the file naming format changes? more brittle but more noticing when this is not longer conforming to our assumptions?

src/pudl/extract/excel.py
```diff
@@ -161,17 +160,22 @@ def add_data_maturity(
...
+        elif ("early_release" in file_name.lower()) or (
```
e-belfer: One drive-by suggestion for simplicity: `early[ |]release` with regex would handle both cases, and you could also use case=False to handle a variety of cases in the match.

test/unit/extract/excel_test.py
```diff
@@ -105,6 +107,32 @@ def test_extract(self):
         }
         assert expected_boxes == res["boxes"].to_dict()
 
+    def test_add_data_maturity(self):
```
krivard, line 110: yes good

src/pudl/extract/excel.py
krivard: or just look for "early" ?

src/pudl/extract/excel.py
krivard: so I like regexes and I like the idea of something that will definitely scream when there's a breaking change but at the same time the not-regex code is so so simple

I'm reluctant to introduce regexes here unless we need to pull multiple things out of the filename.

(did I also go down a small rabbithole and iterate on your regex? [yes I did](https://regex101.com/r/wxOlL3/3) but I don't think it's the right call for this, yet)

src/pudl/package_data/test/file_map.csv
```diff
@@ -1,4 +1,5 @@
-page,2010,2011,2012
+page,2010,2011,2012,2013
 books,b-file.xlsx,b-file.xlsx,everything.xlsx
 boxes,b-file.xlsx,b-file.xlsx,everything.xlsx
 shoes,s-file.xlsx,s-file.xlsx,everything.xlsx
+time,EIA923_Schedules_2_3_4_5_M_12_2021_Final_Revision.xlsx,EIA923_Schedules_2_3_4_5_M_12_2023_Final.xlsx,EIA923_Schedules_2_3_4_5_M_12_2024_21FEB2025.xlsx,EIA923_Schedule_8_Annual_Envir_Infor_2013_Early Release.xlsx
```
krivard: do you want to add a column for

```
EIA923_Schedules_2_3_4_5_M_12_2024_Early_Release.xlsx
```

so we don't regress on the underscore?

krivard (approved):
love it, no regex needed.

no blocking comments.

src/pudl/package_data/test/file_map.csv
cmgosnell: good idea!

PR #4394, opened 2025-07-09
Fix core_pudl__codes_imputation_reasons row count

zschira:
# Overview
Builds failed because we added a new flag to `core_pudl__codes_imputation_reasons`, so the row counts were off.


zaneselvans (approved):

zaneselvans:
Maybe it would be good to add this table of codes as an upstream dependency for the imputation assets so it gets regenerated when we run all related assets?

PR #4401, opened 2025-07-10
fix: output YAML the way `prettier` likes it

### LINKED ISSUE #4119: Write `dbt` YAML files the way `prettier` likes them

jdangerx:
# Overview

Closes #4119.

The YAML files dumped out by `dbt_helper update-tables` didn't indent the lists the same way prettier wants. This adds a weird little custom class that fixes that.

## What did you change?

- change YAML output format
- update `dbt_helper update-tables` so that it runs without errors
- change `dbt_helper update-tables --update` output to have more readable diff

## Documentation

# Testing

Tried updating a schema and the diff was concise. Also the `prettier` check passed.

```fish
daz@inkwell ~/w/pudl (4119-prettier-fix) [1]> dbt_helper update-tables --schema out_eia__yearly_plants --clobber                                                                               (pudl-dev) 
2025-07-10 16:57:16 [    INFO] catalystcoop./home/daz/work/pudl/src/pudl/scripts/dbt_helper.py:388 Wrote schema config for table out_eia__yearly_plants at dbt/models/output/out_eia__yearly_plants/schema.yml.
daz@inkwell ~/w/pudl (4119-prettier-fix)> git diff dbt                                                                                                                                         (pudl-dev) 
diff --git a/dbt/models/output/out_eia__yearly_plants/schema.yml b/dbt/models/output/out_eia__yearly_plants/schema.yml
index 8c15afada..1b02f2d30 100644
--- a/dbt/models/output/out_eia__yearly_plants/schema.yml
+++ b/dbt/models/output/out_eia__yearly_plants/schema.yml
@@ -7,10 +7,6 @@ sources:
               table_name: out_eia__yearly_plants
               partition_column: report_date
-          - expect_date_frequency_ratio:
-              compare_model: source("pudl", "out_eia__yearly_generators")
-              multiplier: 1
-              date_column: report_date
         columns:
           - name: plant_id_eia
           - name: plant_name_eia
daz@inkwell ~/w/pudl (4119-prettier-fix)> git add dbt                                                                                                                                          (pudl-dev) 
daz@inkwell ~/w/pudl (4119-prettier-fix)> pre-commit run prettier                                                                                                                              (pudl-dev) 
prettier.................................................................Passed
```

Also, tried updating a schema with just `--update` and got a nice diff output.

## To-do list

### DIFF
--- a/src/pudl/logging_helpers.py
+++ b/src/pudl/logging_helpers.py
@@ -22,8 +22,6 @@ def configure_root_logger(logfile: str | None = None, loglevel: str = "INFO"):
     numba_logger.setLevel(logging.WARNING)
 
     logger = logging.getLogger("catalystcoop")
-    logger.propagate = False
-
     log_format = "%(asctime)s [%(levelname)8s] %(name)s:%(lineno)s %(message)s"
     coloredlogs.install(fmt=log_format, level=loglevel, logger=logger)
 
--- a/src/pudl/scripts/dbt_helper.py
+++ b/src/pudl/scripts/dbt_helper.py
@@ -11,7 +11,6 @@ import click
 import duckdb
 import pandas as pd
 import yaml
-from colorama import Fore, Style
 from deepdiff import DeepDiff
 from pydantic import BaseModel
 
@@ -22,6 +21,8 @@ from pudl.workspace.setup import PudlPaths
 
 configure_root_logger()
 logger = get_logger(__name__)
+logger.parent.propagate = False
+
 
 ALL_TABLES = [r.name for r in PUDL_PACKAGE.resources]
 
@@ -232,16 +233,14 @@ def schema_has_removals_or_modifications(diff: DeepDiff) -> bool:
 def _log_schema_diff(old_schema: DbtSchema, new_schema: DbtSchema):
     """Print colored summary of schema changes."""
     summary = _schema_diff_summary(old_schema, new_schema)
-    colored_diff = []
+    click.echo("Proposed schema changes:")
     for line in summary:
         if line.startswith("- "):
-            colored_diff.append(Fore.RED + line + Style.RESET_ALL)
+            click.secho(line, fg="red")
         elif line.startswith("+ "):
-            colored_diff.append(Fore.GREEN + line + Style.RESET_ALL)
+            click.secho(line, fg="green")
         else:
-            colored_diff.append(line)
-
-    logger.info("\n" + "\n".join(colored_diff))
+            click.secho(line)
 
 
 def _schema_diff_summary(old_schema: DbtSchema, new_schema: DbtSchema):
@@ -386,13 +385,15 @@ def update_table_schema(
         )
 
         if schema_has_removals_or_modifications(diff):
-            logger.warning(
-                "\n⚠️ WARNING: Some elements would be deleted by this update! Please update manually instead."
-            )
             _log_schema_diff(old_schema, new_schema)
+            # TODO 2025-07-11: perhaps we can integrate this `git add -p` flow directly
+            # into the --clobber command?
             return UpdateResult(
                 success=False,
-                message=f"DBT configuration for table {table_name} has information that would be deleted. Update manually or run with clobber.",
+                message=f"DBT configuration for table {table_name} has "
+                "information that would be deleted. Update manually, or: "
+                "1) re-run with --clobber, 2) run `git add dbt/models -p` "
+                "and follow the instructions for recovering the deleted info",
             )
 
     model_path.mkdir(parents=True, exist_ok=True)


src/pudl/logging_helpers.py
```diff
@@ -22,6 +22,8 @@ def configure_root_logger(logfile: str | None = None, loglevel: str = "INFO"):
     numba_logger.setLevel(logging.WARNING)
 
     logger = logging.getLogger("catalystcoop")
+    logger.propagate = False
```
jdangerx: This being `True` was causing some irritating double-logging when I ran `dbt_helper` commands.

src/pudl/scripts/dbt_helper.py
```diff
@@ -194,38 +228,34 @@ def schema_has_removals_or_modifications(diff: DeepDiff) -> bool:
...
-        if isinstance(changes, dict):
-            for path, value in changes.items():
-                summary_elements.append(f"  - {path}: {value}")
-        else:
-            for item in changes:
-                summary_elements.append(f"  - {item}")
 
-    return "\n".join(summary_elements)
+def _schema_diff_summary(old_schema: DbtSchema, new_schema: DbtSchema):
+    """Return a summary of schema changes based on YAML output."""
```
jdangerx, line 247: The `deepdiff` output is useful for figuring out what types of changes there are (add, remove, etc) but the string output is tough to read. Also, dumping out two whole YAML files is hard to compare. But we can just compare the YAML files with `difflib` and print out something actually kind of nice.

jdangerx:
@marianneke I noticed in testing this PR out that the update logic basically forces us to manually update all schemas that have custom data tests - probably we would want this to preserve any data tests that already exist? But that seems out of scope for this PR (which already has ballooned to doing two things instead of one).

I guess one issue would be "what if the data test no longer refers to an existing column?" but again that seems out of scope here.

```fish
daz@inkwell ~/w/pudl (4119-prettier-fix)> dbt_helper update-tables --schema out_eia__yearly_plants --update                                                                                    (pudl-dev) 
2025-07-10 17:17:13 [ WARNING] catalystcoop./home/daz/work/pudl/src/pudl/scripts/dbt_helper.py:388 
⚠️ WARNING: Some elements would be deleted by this update! Please update manually instead.
2025-07-10 17:17:13 [    INFO] catalystcoop./home/daz/work/pudl/src/pudl/scripts/dbt_helper.py:243 
--- old_schema
+++ new_schema
@@ -7,10 +7,6 @@
               table_name: out_eia__yearly_plants
               partition_column: report_date
-          - expect_date_frequency_ratio:
-              compare_model: source("pudl", "out_eia__yearly_generators")
-              multiplier: 1
-              date_column: report_date
         columns:
           - name: plant_id_eia
           - name: plant_name_eia
2025-07-10 17:17:13 [   ERROR] catalystcoop./home/daz/work/pudl/src/pudl/scripts/dbt_helper.py:410 DBT configuration for table out_eia__yearly_plants has information that would be deleted. Update manually or run with clobber.
```

src/pudl/logging_helpers.py
marianneke: Could you comment on the change in behavior this will cause? Looking into it briefly, it seems that setting it to `False` is helpful if there is some logging you want to treat in an isolated manner, but could this negatively affect logging behavior in other parts of the code?

Also, would it make more sense to set this to False in the dbt helper module or maybe in all scripts by default (since it makes more sense for that to be isolated)?

test/unit/dbt_helper_test.py
```diff
@@ -612,34 +611,33 @@ def test_complex_schema_diff_output(capsys):
...
-    assert "'description': 'old'" in output, output
-    assert "'description': 'new'" in output, output
-
-    # model2 addition
-    assert "model2" in output, output
-    assert "'description': 'added'" in output, output
+    output = _schema_diff_summary(old_schema, new_schema)
+    expected = """--- old_schema
++++ new_schema
+@@ -1,12 +1,20 @@
```
marianneke, line 617: I like having an explicit form of the expected output 👍

jdangerx:
> @marianneke I noticed in testing this PR out that the update logic basically forces us to manually update all schemas that have custom data tests - probably we would want this to preserve any data tests that already exist? But that seems out of scope for this PR (which already has ballooned to doing two things instead of one).
> 
> I guess one issue would be "what if the data test no longer refers to an existing column?" but again that seems out of scope here.

One workaround I guess is to hit that `--clobber` and then use `git add -p` to keep the stuff you want from the old version.

src/pudl/logging_helpers.py
jdangerx: Good question. I tried running `make pytest-coverage` and the integration test logs were a little funky. I decided that instead of chasing that down and figuring out what exactly was going on I should:

* use `click.secho()` for the colored output - that's the "intended" tool for CLI program output that I just learned about from #4163 
* slap that `logger.parent.propagate = False` in the dbt_helper script only

Then I think we don't have to worry about breaking stuff outside of this script when we merge :)

marianneke:
> @marianneke I noticed in testing this PR out that the update logic basically forces us to manually update all schemas that have custom data tests - probably we would want this to preserve any data tests that already exist? But that seems out of scope for this PR (which already has ballooned to doing two things instead of one).
> 
> I guess one issue would be "what if the data test no longer refers to an existing column?" but again that seems out of scope here.

Agree. More elegant automated handling of updates would definitely be helpful. That can get a little convoluted with edge cases though, especially as you said if columns are renamed but also if parameters in the test are changed, and we need to decide whether to just add tests to the column or replace.

src/pudl/scripts/dbt_helper.py
```diff
@@ -20,10 +21,50 @@
...
 
+def _prettier_yaml_dumps(yaml_contents: dict[str, Any]) -> str:
+    """Dump YAML to string that Prettier likes."""
+
+    class PrettierCompatibleDumper(yaml.Dumper):
+        """Custom Dumper that indents lists like prettier does.
+
+        Default dumper behavior::
+
+            foo:
```
marianneke, line 38: Wait, was the old behavior in DbtSchema.to_yaml` like this? Because then I would have expected to see errors due to incorrect alignment.

marianneke:
Adding note here too for posterity:
The `_prettier_yaml_dumps` functionality you added is quite general so I wonder if that should be moved out of the script and into the core library instead.

src/pudl/scripts/dbt_helper.py
jdangerx: Yes, this was the old behavior - it's "valid YAML" but also `prettier` runs in our pre-commit hooks which fixes all the alignment issues.

PR #4405, opened 2025-07-13
Organize dbt models based on table name to make them easier to find.

zaneselvans:
# Overview

While working on #4105 / #4382 it was a bit frustrating to not be able to find dbt model definitions based on the table name -- especially EIA tables were scattered throughout a number of different subdirectories, since the directory was based on poorly-defined data "sources" metadata.

This PR re-organizes the model directories to reflect the data-source element of the table name. E.g. for `core_eia__scd_generators` it would be `eia` -- it's always the 2nd element of the table name. This makes it easy to find the model if you know the table name.

It also updates the `get_data_source()` logic in `dbt_helper` to reflect this change, so any future models produced with the script will land in the right place.

# Testing

- Ran the `dbt` tests locally after having re-organized the models, and they still passed.
- Ran the unit tests.

## To-do list

- [ ] Run `make pytest-coverage` locally.
- [ ] Run the `build-deploy-pudl` GitHub Action manually.


zaneselvans:
Seems like git handled merging the renamed files just fine, but it wasn't something that GitHub wanted to do automatically.

marianneke (approved):
I would not be the person to opine on ideal naming conventions, but the implementation looks good.

Nit: since the logic now is based on table name, you may want to figure out edge case / data weirdness and the desired behavior in those cases. E.g. could somehow a table name be written wrong, such as `out__bla_bla` which would result in an empty string. Should the code then throw an error ideally?

zaneselvans:
Hmm, I guess that is a potential failure... I suspect the right thing to do is go further upstream and enforce the table naming convention that we've adopted in the `pudl.metadata.classes.Resource` class with a Pydantic validation for `Resource.name`. This might be a good check to work into the new metadata / docs build work. I know @krivard was identifying some outstanding rogue table names.  I've added #4436 as a place to discuss this and come up with a concrete proposal.

PR #4429, opened 2025-07-15
chore: skip integration tests for docs

### LINKED ISSUE #4410: Skip slow and irrelevant checks for docs-only builds
### Overview

If documentation updates take 2h to merge because of slow CI we will not update our docs (see: our docs are not updated often).

### Success Criteria

If your PR only touches files in `docs/**.rst`:

* [ ] we'll run `ci-docs` but not the integration tests.

### Next steps
* [ ] look at [paths-filter](https://github.com/dorny/paths-filter?tab=readme-ov-file#examples) and adapt an example to our use case.

jdangerx:
# Overview

Closes #4410.

When pushing a docs-only change we shouldn't have to wait for our slow CI build.

## What did you change?

Added a job in the pytest workflow that checks to see:

* is this manually dispatched?
* is this a merge-queue trigger AND are there non-docs changes? 

If either of those are true, it sets a flag telling downstream jobs to run the full CI job + unit tests + coverage.

Otherwise, skips those.

## Documentation

# Testing

- [x] remove the 'merge group' filter so we can easily test outside of the queued environment
- [x] make a [docs-only PR](https://github.com/catalyst-cooperative/pudl/pull/4430) to this branch and see if everything is properly skipped
- [x] make a [mixed PR](https://github.com/catalyst-cooperative/pudl/pull/4431) to this branch and see if things get triggered
- [x] restore merge group filter

## To-do list

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -21,8 +21,8 @@ Quality of Life Improvements
   our pre-commit hooks expect, reducing annoying back-and-forth and diffs. See
   :issue:`4119` and :pr:`4401`.
 
-* Stopped running code checks in CI when no code has changed. See :issue:`4410`
-  and :pr:`4429`.
+* Stopped running code checks in CI when only the documentation has changed.
+  See issue :issue:`4410` and PR :pr:`4429`.
 
 Bug Fixes
 ^^^^^^^^^


.github/workflows/pytest.yml
```diff
@@ -14,6 +14,27 @@ env:
...
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          # this will raise a warning, but it works:  https://github.com/dorny/paths-filter/issues/225
+          predicate-quantifier: "every"
+          filters: |
+            code:
+              - '!docs/**.rst'
+              - '!docs/**.md'
+              - '!README.rst'
```
jdangerx: I wanted to be fairly conservative here so we don't accidentally skip the code checks when code has changed - this says that "if any file that has changed is not the README, or a md/rst within the docs folder, that counts as a code change"

zaneselvans: There are fewer of these files and they get updated less frequently but we might also want to include:

docs/**.pdf
docs/**.html

What about docs/**.rst.jinja? If there haven't been changes to any python files, but the Jinja template content has changed, is there a risk of something that needs to get tested not getting tested? Since those templates will get exercised in the docs build.
Maybe need to make sure that this gets another look when the new metadata stuff is merged in, in case there are other things that need to be included / ignored.

.github/workflows/pytest.yml
```diff
@@ -66,6 +87,8 @@ jobs:
 
   ci-unit:
     runs-on: ubuntu-latest
+    needs: change_filter
+    if: ${{ needs.change_filter.outputs.run_code_checks == 'true' }}
```
zaneselvans: Maybe I'm not understanding but it seems like this will also mean that the unit tests won't run until we get into the merge queue, which may result in missing some basic failures until late in the PR process (syntax errors, imports, dependencies, etc). Is that really what we want?

zaneselvans (changes requested):
I'm nervous about not attempting to run the unit tests until we get to the merge queue. Am I understanding correctly that that's what this would do?

And maybe some additional docs-only files we can include.

.github/workflows/pytest.yml
jdangerx: Duh. Yes, good catch. I added this in at the last second and totally forgot about why we even run unit tests in the first place. Taking this out.

.github/workflows/pytest.yml
jdangerx: Yeah, I think `**.rst.jinja` should be OK also:

- if the change is only to `**.rst.jinja` it should get caught in docs build
- if the change is to something else that feeds into a custom macro or something, it will trigger the code checks.

docs/release_notes.rst
```diff
@@ -21,6 +21,9 @@ Quality of Life Improvements
   our pre-commit hooks expect, reducing annoying back-and-forth and diffs. See
   :issue:`4119` and :pr:`4401`.
 
+* Stopped running code checks in CI when no code has changed. See :issue:`4410`
+  and :pr:`4429`.
```
zaneselvans:
```suggestion
* Stopped running code checks in CI when only the documentation has changed.
  See issue :issue:`4410` and PR :pr:`4429`.
```

zaneselvans:
@jdangerx Hold on, I would not have expected this to skip the merge queue & integration tests, because the workflow file was altered, and it seems like it should not have been excluded by the `"!docs/**.rst"` etc patterns. Would you have expected this?

jdangerx:
OK, some strangeness going on here:

1. When I pushed changes to the branch, that triggered a `sync` action. That build skipped integration tests because it's not a merge group event: https://github.com/catalyst-cooperative/pudl/actions/runs/16331284344
2. Then, when the merge group event happened, the `change_filter` job failed because it thought pytest.yaml was malformed: https://github.com/catalyst-cooperative/pudl/actions/runs/16331365834 - this led to the downstream `ci-integration` and `ci-coverage` jobs, which both depend on `change_filter`, getting skipped - and ["skip" is as good as "succeed" as far as branch protections go](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging)
3. Finally, I just kicked off a [workflow dispatch on `main`](https://github.com/catalyst-cooperative/pudl/actions/runs/16345843252) which successfully ran `change_filter` and decided to run `ci-integration` - so that seems to actually be working fine. I cancelled it to save execution time.

So, it *seems* like the `change_filter` failure was maybe a quirk of how the merged workflow was being handled?

Next step is to double-check to make sure:
(a) if integration tests fail, ci-coverage *fails* instead of getting skipped
(b) change_filter does work in merge group events.

Not sure how to test (a) without breaking `ci-integration` intentionally in a PR to `main`, then popping it in the merge queue. Which then might auto-merge once done, if `ci-coverage` is not well-behaved... maybe we can force a different required check (e.g. docker build) also to fail to protect us from breaking `main`.

New plan is to copy branch protection rules from `main` to another test branch and make the PR against that branch.

jdangerx:
OK, found out that this was because `github.event.merge_group` was evaluating to an `Object` in `run_code_checks:` . This worked before because we had the evaluation in a conditional, which coerces the `Object` into a boolean. But when we're just assigning the `github.event.merge_group` to an output, it doesn't coerce the `Object` and instead complains about "the template is not valid. A mapping was not expected". Which, OK, fair. But then because the `change_filter` job fails, the following `ci-integration` and `ci-coverage` jobs both get *skipped* instead of failing. Which lets the PR get merged.

Fix: 
* coerce the merge group object to a boolean
* add change_filter to branch protection

PR #4435, opened 2025-07-15
Dbt helper refactor row counts

### LINKED ISSUE #4364: Fix row-count update bugs in `dbt_helper`

marianneke:
# Overview

Closes #4364.

dbt helper's row count functionality had a few bugs, as well as other behaviors that were not ideal:
- All potential cases were not handled, including where tests are not defined (should not have row counts in csv)
- The source of truth for which partition column to use was inferred from columns present and some simple rules, instead of explicitly read from the schema
- The test suite did not cover the update logic, as well as most helper functions

## What did you change?
- Removed code creating row counts test automatically during initial model setup
    - This required removing this behavior from pytest module as well
    - Kept helper functions to do with constructing row count tests around so they can be used in personal scripts. Should later be added to a CLI method for adding tests conveniently
- Updated row count refresh behavior to remove old row counts whenever new row counts can be written (if partition column is changed, this means removing old partitions)
- Updated row count calculation behavior to allow a SQL expression to be added as `partition_column` is replace by `partition_expr`
    - This included simplifying the logic in dbt helper as well as the row counts macro since users can now add `"EXTRACT(YEAR FROM report_date)"` as the partition column, meaning casting to year does not need to happen in the code anymore, reducing the likelihood of the code diverging between dbt helper and the macro.
    - This also required the modification of all schema.yml files that had a date or timestamp column as the partition column, since those would not be cast as a year anymore by the code.
- Added tests for row count related functions to improve test coverage
- Added schema constructor to be used in the rest of the test suite
- Updated docs describing dbt helper behavior
- Added new behavior to release notes

## Documentation

# Testing

- Added pytest cases for all new functionality (as well as improving coverage of old functionality)

## To-do list


docs/release_notes.rst
```diff
@@ -11,12 +11,22 @@ New Data
 
 Expanded Data Coverage
 ^^^^^^^^^^^^^^^^^^^^^^
+EIA 860M
+~~~~~~~~
+* Added May 2025 EIA 860m data as part of the new monthly release cycle.
```
jdangerx: This probably belongs in a separate PR, yes?

dbt/README.md
```diff
@@ -42,12 +42,15 @@ dbt_helper update-tables --schema {table_name(s)}
 
 will add a file called `dbt/models/{data_source}/{table_name}/schema.yml` for each
 listed table. This yaml file tells `dbt` about the table and its schema. If the
-table already exists and you need to update it, you'll have to add `--clobber`
+table's schema already exists and you need to update it, you'll have to add `--clobber`
```
jdangerx: We should probably explain more about the clobber/update mechanics and what to do about them, but that's out of scope for this PR I think. More something for #4126 .

dbt/README.md
```diff
@@ -42,12 +42,15 @@ dbt_helper update-tables --schema {table_name(s)}
...
 
-It will also specify the `check_row_counts_per_partition` test. This test works by
-comparing expected row counts for partitions within a table (typically distinct
-`report_date` values) stored in `etl_fast_row_counts.csv` and `etl_full_row_counts.csv`
-against the actual row counts in the materialized tables.
+The default schema is very minimal. One can add things such as descriptions or tests.
+For adding tests, the `dbt_helper.py` module contains helper functions currently not
+exposed through the CLI.
+
+The `DbtSchema` class has generic methods `add_source_tests` and `add_column_tests`. These require the user to pass the tests and relevant parameters as dictionaries and write the result to yaml
```
jdangerx: Note for #4126 is to have an example of how to use this.

src/pudl/scripts/dbt_helper.py
```diff
@@ -325,19 +312,49 @@ def _write_row_counts(row_counts: pd.DataFrame, target: str = "etl-full"):
...
+
+    partition_columns = _extract_row_count_partitions(schema)
+    has_test = bool(partition_columns)
     existing = _get_existing_row_counts(target)
-    if table_name in existing["table_name"].to_numpy() and not (clobber or update):
+    has_existing_row_counts = table_name in existing["table_name"].to_numpy()
+    allow_overwrite = clobber or update
+
+    if not has_test:
+        if has_existing_row_counts:
```
jdangerx: nit: maybe a little easier to read?
```suggestion
    if has_existing_row_counts and not has_test:
```

Overall this conditional is pretty gnarly but so long as it's tested thoroughly I'm good with it.

src/pudl/scripts/dbt_helper.py
```diff
@@ -410,8 +424,21 @@ def _log_update_result(result: UpdateResult):
...
+    if table.data_tests:
+        for test in table.data_tests:
+            test_def = test.get("check_row_counts_per_partition")
+            if isinstance(test_def, dict):
+                partitions.append(test_def.get("partition_column"))
+
+    return partitions
+
+
+def _infer_partition_column(table: DbtTable) -> str:
```
jdangerx: Did you still want to remove this logic, or saving that for a separate PR?

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
     )
 
 
+def test__get_existing_row_counts(mocker):
```
jdangerx, line 98: * this tests `_get_existing_row_counts`, which is not part of the public interface of `dbt_helper` - it's just an implementation detail that we're trying to hide away
* this test just says that "we call the mock and, indeed, the return value we passed in is the same as itself" - which is also very closely tied to the implementation of `_get_existing_row_counts` ("reads This Specific CSV") rather than the intended behavior ("given a specific target, grabs the correct row counts").

Because of this I think the test as written is (a) not that informative (when it fails, it tells us the implementation changed, not that the observed behavior doesn't match the intended behavior anymore) and (b) kind of brittle (it fails whenever the implementation changes, regardless of whether the observed behavior is correct or not).

I think we have two options here:
* delete the test because the behavior is covered by testing our actual public interface
* update this test to tell us about whether the function actually fulfills its spec - but since this is a function about "find the right file to read and read it", I think we would need to do a bit more than "mock read_csv to return a dataframe, then check that the dataframe is the same as itself." In this case I think we should have a docstring in the definition of the function itself that describes the *intended* behavior as well, because otherwise when the test fails we won't know if it was the test or the code that needs changing.

I know that this is immediately preceded by a bunch of other tests that have similar issues: `test__get_local_table_path()`, `test__get_model_path()`, and `test__get_row_count_csv_path()`. I would probably clean those up but that's out of scope of this PR.

src/pudl/scripts/dbt_helper.py
```diff
@@ -325,19 +312,49 @@ def _write_row_counts(row_counts: pd.DataFrame, target: str = "etl-full"):
...
         )
 
+    if has_existing_row_counts and not allow_overwrite:
+        return UpdateResult(
+            success=False,
+            message=f"Row counts for {table_name} already exist. Use clobber or update to overwrite.",
+        )
+
+    # At this point: test exists, and either no existing or overwrite allowed
+    partition_column = partition_columns[0]  # TODO: support multiple partitions
```
zaneselvans: Might as well assert what's being said in the comment, just in case something weird has happened.

For now it seems like we should be asserting here that this list has only one element.

src/pudl/scripts/dbt_helper.py
```diff
@@ -410,8 +424,21 @@ def _log_update_result(result: UpdateResult):
...
-    all_columns = [c.name for c in PUDL_PACKAGE.get_resource(table_name).schema.fields]
+def _extract_row_count_partitions(table: DbtTable) -> list[str | None]:
+    """Extract partition columns from check_row_counts_per_partition tests in a DbtTable."""
+    partitions: list[str | None] = []
+
+    if table.data_tests:
+        for test in table.data_tests:
+            test_def = test.get("check_row_counts_per_partition")
+            if isinstance(test_def, dict):
+                partitions.append(test_def.get("partition_column"))
```
zaneselvans: So this is iterating over all tests in `data_tests` and if `check_row_counts_per_partition` is defined it'll get the test definition, otherwise it'll get None?  Is there a scenario where it can get something other than None or a dict? If it did it seems like that would be an error.

src/pudl/scripts/dbt_helper.py
zaneselvans: Since this is no longer being used and we want to require an explicitly chosen `partition_column` I think we probably want to remove it. Is there an argument for keeping it around?

src/pudl/scripts/dbt_helper.py
```diff
@@ -421,6 +448,16 @@ def _infer_partition_column(table_name: str) -> str:
     return None
 
 
+def _get_row_count_test_dict(table_name: str, partition_column: str):
```
zaneselvans: This isn't being used anywhere and the functionality is very simple. Do we want to keep it?  If so, it seems like it's meant to be imported elsewhere so we should remove the leading underscore on the function name.

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+        "pudl.scripts.dbt_helper.pd.read_csv", return_value=mock_df
+    )
+
+    result = _get_existing_row_counts()
+
+    pd.testing.assert_frame_equal(result, mock_df)
+    mock_read_csv.assert_called_once()
+
+
+def test__write_row_counts(mocker):
```
jdangerx: Similar to above - I think we should:
* not test this function directly (we can preface it with a `__` to make it *really* hard to accidentally use...) OR
* write down what "business logic" we expect this code to do as opposed to the steps it should take to get there, and then test for that instead of the implementation.

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+    )
+    mock_to_csv = mocker.patch("pandas.DataFrame.to_csv")
+
+    _write_row_counts(df)
+
+    mock_csv_path.assert_called_once()
+    mock_to_csv.assert_called_once_with("/fake/path.csv", index=False)
+
+
+def test__get_row_count_test_dict():
```
jdangerx: In contrast to the above, I like this! Even though it's testing something that's also not "supposed" to be used elsewhere, it tests that the function "does what it's supposed to" and there's also a docstring that describes the function under test is "supposed to" do.

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+            ],
+            [],
+        ),
+        (
+            None,
+            [],
+        ),
+    ],
+)
+def test__extract_row_count_partitions(data_tests, expected):
```
jdangerx, line 189: I like this test too, again because it's focused on the intent rather than the implementation of the function :)

I wonder - what should happen if we pass in the following `data_tests`?

```
data_tests = [
    {
        "check_row_counts_per_partition": { 
            "table_name": "plants",
            "partition_column": None,
        }
    }
]
```

Based on reading the code I would expect to get `[None]` back, but IDK if that's the intended behavior.

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
-    ).get_resource.return_value = mock_resource
-    mock_resource.schema.fields = [with_name(mocker.Mock(), key)]
-    assert _infer_partition_column("") == key
+def test__infer_partition_column(key):
+    if key is not None:
+        columns = [DbtColumn(name=key)]
+        expected = key
+    else:
+        columns = [DbtColumn(name="not_a_partition_column")]
+        expected = None
```
jdangerx: This seems like we should just chuck the `columns` value in a parameter too. And then set `expected = key`.

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+
 @pytest.mark.parametrize("key", ["report_year", "report_date", "datetime_utc", None])
-def test__infer_partition_column(mocker, key):
-    mock_resource = mocker.Mock()
-    mocker.patch(
-        "pudl.scripts.dbt_helper.PUDL_PACKAGE"
-    ).get_resource.return_value = mock_resource
-    mock_resource.schema.fields = [with_name(mocker.Mock(), key)]
-    assert _infer_partition_column("") == key
+def test__infer_partition_column(key):
```
jdangerx: I like that this test also is testing the interface, not the implementation, but also I think we're missing a piece of the interface here. I might be wrong, but from reading the code it *seems* like we would prefer `report_year` over` report_date` over `datetime_utc` - which we could test here by setting columns to `["report_year"]`, `["report_date", "report_year"]`, etc.

Though I guess we may also just want to rip *all* of this logic out anyways.

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+    )
+    mocker.patch("pudl.scripts.dbt_helper._calculate_row_counts", return_value=new_df)
+    mock_write = mocker.patch("pudl.scripts.dbt_helper._write_row_counts")
+    mock_model_path = mocker.patch("pudl.scripts.dbt_helper._get_model_path")
+    mock_model_path.return_value.__truediv__.return_value = "fake/schema.yml"
+    mocker.patch("pudl.scripts.dbt_helper.DbtSchema.from_yaml", return_value=schema)
+    mocker.patch(
+        "pudl.scripts.dbt_helper._extract_row_count_partitions",
+        return_value=[given["partition_column"]] if given["has_test"] else [],
+    )
```
jdangerx: non-blocking: this pile of mocks is here because `update_row_counts` does a bunch of IO... the function would be easier to test if we had a wrapper that read all the files into memory, passed the values into `update_row_counts`, and then wrote (or didn't write) the output of `update_row_counts` to a different file.

That would probably be a good pattern to follow in the future, and provide easy boundaries to break apart the `dbt_helper` module into smaller submodules, but is *totally* out of scope for this PR.

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+    assert result == expect["result"]
+
+    # Determine whether something should be written
+    has_test = given["has_test"]
+    table_in_existing = given["table_name"] in existing_df["table_name"].to_numpy()
+    allow_update = given["clobber"] or given["update"]
+
+    should_write = (has_test and (not table_in_existing or allow_update)) or (
+        not has_test and table_in_existing and allow_update
+    )
```
jdangerx: This logic feels like it itself needs a test somewhere... or we can just define "should write" within the `expect` and call it a day.

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+                    "partition": ["2020"],
+                    "row_count": [100],
+                }
+            ),
+        },
+    ),
+]
+
+
+@pytest.mark.parametrize("case", COMBINE_ROW_COUNT_TEST_CASES)
```
jdangerx, line 374: Like the test cases, like that `_combine_row_counts` is a nice pure function which makes the test so easy to read :)

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+@pytest.mark.parametrize("case", CALCULATE_ROW_COUNTS_CASES)
+def test__calculate_row_counts_(case, mocker):
+    # Given
+    mocker.patch(
+        "pudl.scripts.dbt_helper._get_local_table_path",
+        return_value=case.given["mocked_path"],
+    )
+    mocker.patch(
+        "pudl.scripts.dbt_helper.duckdb.sql"
+    ).return_value.df.return_value = case.given["mocked_duckdb_df"]
```
jdangerx: Hmm, I feel like the *key* piece of the functionality of `_calculate_row_counts` is "it figures out the right column to aggregate by" - so with this mock we should probably add another assertion to make sure the right SQL is being passed into `duckdb.sql`, instead of just having it spit out the right value uncritically.

test/unit/dbt_helper_test.py
```diff
@@ -204,65 +645,65 @@ def dbt_schema_mocks(mocker):
...
-    actual = DbtSchema.from_table_name(
-        dbt_schema_mocks.table_name,
-        partition_column=dbt_schema_mocks.partition_column,
-    )
-    assert actual == dbt_schema_mocks.schema
-
-
-def test_dbt_schema__from_yaml(mocker, dbt_schema_mocks):
-    with StringIO(dbt_schema_mocks.yaml) as f:
+def test_dbt_schema_loading(dbt_schema_mocks, mocker):
```
jdangerx, line 665: clarification: this test checks to make sure we can round-trip the schemas through yaml and also through our bespoke metadata dictionary?

jdangerx (changes requested):
Glad this is coming along!

Major blocking changes/questions to address:

* did you want to pull out the inference logic here, or in a different PR?
* adding tests is good and some of these were really helpful in making me feel confident in your changes! I think some of the ones you've added are a little bit brittle or uninformative and could use some rework or just straight-up deletion
* whichever underscored/"private-ish" functions you decide to keep testing should probably have docstrings that tell us a bit more about the business logic the function encapsulates - what, as a caller, we need to know that the function does.

Also had some non-blocking musing, clarification questions, etc.

test/unit/dbt_helper_test.py
```diff
@@ -26,6 +35,46 @@
...
+                    }
+                }
+            ]
+        return DbtSchema(
+            version=2,
+            models=None,
+            sources=[
+                DbtSource(
+                    name=source_name,
+                    data_tests=None,
```
zaneselvans: Oh interesting, I didn't realize that you could have `data_tests` at the "source" level too. The "source" here is `pudl` right?  What kinds of data tests would make sense there? Would that just be anything that depends on more than one table?

test/unit/dbt_helper_test.py
zaneselvans, line 189: The case of unpartitioned tables (where there's no partition column and we just want to count all the rows in the table) has been problematic and should be covered. I'm not sure what kind of values we're assigning to the partition column in those cases when we generate the row counts, or when we read in existing row counts.  Is it `None`? The empty string? `pd.NA`, `np.nan`?  We should pick one and make sure it round-trips cleanly, and is being used in all the existing test specs

```py
"check_row_counts_per_partition": {
    "table_name": "plants",
    "partition_column": "",
}
"check_row_counts_per_partition": {
    "table_name": "plants",
    "partition_column": None,
}
```

test/unit/dbt_helper_test.py
zaneselvans: This function is no longer being used and we wanted to remove this functionality entirely. Is there a reason to leave it in / test it? We want to allow any column to be used as the partition column, and for the user to explicitly include it in the test definition.

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+        columns=columns,
+        data_tests=None,
+    )
+
+    assert _infer_partition_column(table) == expected
 
 
 GivenExpect = namedtuple("GivenExpect", ["given", "expect"])
 
+ROW_COUNT_TEST_CASES = [
```
zaneselvans, line 198: One case that's been problematic which we need to check that's not represented here is when there is no value in the partition column. This can happen for two reasons:
- "there's no partition value in this row because the table is unpartitioned" (we're just counting all the rows), and
- "there's no partition value in this row because sometimes the partition column value is null"

So, test cases we might want to add:
- Test exists, with no partition column, and the row counts change.
- Test exists, with no partition column, and the user removes the test. 
- Test exists, but has no partition column initially, and the user adds a partition column -- so the existing row counts would only have a single record for the table, with no value in the partition column, but the new row counts would have multiple rows for the table, with non-null partition values.
- Test exists, with no partition column, the user adds a partition column, and the new row counts also include a record indicating row counts for when the partition column value is null.
- Are there other weird combos?

Might also want to mix in some cases where the partition column isn't a datetime, e.g. "state" with string values "CA", "TX", etc for the partition values since we now expect those to be valid.

src/pudl/scripts/dbt_helper.py
marianneke: I can imagine including this logic in an interactive test creator (suggesting likely partition columns for user to pick from, but they can also pick another). Also, I believe Kathryn has her own scripts for convenience of adding tests. She might like the convenience of being able to import this? But I definitely agree we should not leave this for long

src/pudl/scripts/dbt_helper.py
marianneke: I can reorganize it a bit with the early return failure cases at the top

test/unit/dbt_helper_test.py
```diff
@@ -54,18 +103,377 @@ def test__get_row_count_csv_path():
...
+                },
+                {
+                    "check_row_counts_per_partition": {
+                        "table_name": "plants",
+                        "partition_column": "datetime_utc",
+                    }
+                },
+            ],
+            ["report_date", "datetime_utc"],
+        ),
```
zaneselvans: What's the idea with this test case?  I can imagine two different scenarios where we might want to have multiple partition columns:
- we want to partition using multiple columns in a groupby: e.g. breaking down the EPA CEMS hourly data by *both* state and year, so that each row count record is one state-year of data, and then test all of those partition row counts in one test.
- we want to do two entirely separate partitionings of the table, essentially meaning two separate instances of the test, with a different column (or different combinations of columns if we're allowing the above), in which case the two columns would be different kinds of things, like `report_date` and `state`. But I think this would mean that we need to include the name of the partition column in the row counts CSV, so we can tell which instance of the test each record pertains to. This seems like the scenario that's imagined by this test case, but it's currently invalid in the broader context.

And then also I don't think we would ever be likely to use two date / time columns together in either of these cases.

I feel like having the multiple columns / multiple tests stuff partly implemented in one function, but invalid everywhere else is confusing and invites problems.

test/unit/dbt_helper_test.py
marianneke: Yes, `[None]` would be the intended behavior. It's an efficient way to store all the information in the row count tests:
- The # of elements in the list is the number of row count tests (with empty list being no tests)
- If an element is `None` there is no partition column

zaneselvans (changes requested):
When I try to run either of these:
```
dbt_helper update-tables --row-counts --clobber all
dbt_helper update-tables --row-counts --clobber core_epacems__hourly_emission
```

I get the following error:
```
  File "/Users/zane/miniforge3/envs/pudl-dev/bin/dbt_helper", line 8, in <module>
    sys.exit(dbt_helper())
             ^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1161, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1443, in invoke
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 788, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/scripts/dbt_helper.py", line 557, in update_tables
    update_row_counts(
  File "/Users/zane/code/catalyst/pudl/src/pudl/scripts/dbt_helper.py", line 325, in update_row_counts
    partition_columns = _extract_row_count_partitions(schema)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/scripts/dbt_helper.py", line 431, in _extract_row_count_partitions
    if table.data_tests:
       ^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/pydantic/main.py", line 991, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'DbtSchema' object has no attribute 'data_tests'
```

It looks like `_extract_row_count_partitions()` is being called on a `DbtSchema` rather than the expected `DbtTable`.
 
* I noted some additional test cases that have been problems in the past, esp. when there's no partition column, or the partition value is null.
* There are a couple of helper functions that aren't used any more and should probably be removed? The partition column inference logic in particular I thought we wanted to purge entirely.
* I think allowing multiple partition columns in one small part of the code but having them be invalid / undefined elsewhere (there's 2 different ways I can imagine having multiple columns) is confusing and invites problems. For the moment I think we should just stick with one column, and then implement that all at once when we know how we actually want it to work.

In general I'm struggling to parse what's going on in here though, partly because of the mocking, and partly because of how the inline test data is being stored. I think I would have a hard time adding new test cases and feeling confident that I was really testing what I thought I was testing. Is there an easy way to make the parametrized tests more compact or pull the test data out into something that reads like a real CSV we might be working with?

src/pudl/scripts/dbt_helper.py
marianneke: Adding this as a ValueError when the partition_columns list is constructed above

src/pudl/scripts/dbt_helper.py
marianneke: If it were not a dictionary, then it would be incorrectly formatted in the schema, in which case this code would interpret that as there not being any row counts test defined. Do you think there should be a ValueError raised in that case, or should schema validation happen in some other place?

src/pudl/scripts/dbt_helper.py
marianneke: Only if folks who are importing this module in their own scripts find it convenient. But I guess it's cleaner to just remove and write convenient functions for adding the different types of tests in the CLI

test/unit/dbt_helper_test.py
marianneke: That is true mainly because this function does not do more than read csv (except ensure the partition column is a string). But generally I do feel there is value even in testing quite basic functionality, since the implementation can change

test/unit/dbt_helper_test.py
marianneke: Good point: I will add a more salient multiple test example back in when we start supporting that case, and instead just have the three types of cases we currently support represented: no partition (partition columns should be [None]), partition (partition columns should be ["partition column defined in test"]), no test (partition columns should be empty list).

test/unit/dbt_helper_test.py
marianneke: I was thinking it might be a convenience function when we implement an interactive tool for creating data tests, but it will be easy enough to reimplement if it is useful and actually it's a better idea to not tie ourselves to old logic so decided to yeet it after all

test/unit/dbt_helper_test.py
marianneke: Your second case "there's no partition value in this row because sometimes the partition column value is null" would be the same as the partition column having a constant value, right?

zaneselvans: I'm not sure what you mean by "partition column having a constant value" -- it's "constant" in that like, instead of having "2009" as a year we could have `pd.NA` as a "year" so it's just another value that should appear in the `GROUP BY`. But it's weird case in that, if you're just looking at that row in the CSV, it's the same as the case in which there isn't any partition column at all, which can be confusing / confounding. Especially if NA values are being dealt with inconsistently, or if we're depending on checking for equality between partition values, given that `pd.NA != pd.NA`

src/pudl/scripts/dbt_helper.py
zaneselvans: I think the _underscore suggests not to import it elsewhere, and I think we want to stop doing this kind of automated setting in general now that we've got all the tables in. I think we probably want to get away from just specifying the column entirely, and instead give it an SQL snippet that'll be used in the GROUP BY to identify the value to group on, like `EXTRACT(year FROM datetime_utc)` or `year` or `state` etc. depending on the table.  Like right now if I want to make it partition the EPA CEMS hourly emissions table by the year portion of `operating_datetime_utc` I don't know if that would work -- would it know to try and extract the year? Or would I get a single group for every timestep?

src/pudl/scripts/dbt_helper.py
marianneke: ...decided to raise a ValueError when test is defined incorrectly

test/unit/dbt_helper_test.py
marianneke: Good call-out. There are no tests on the source level, just on the table and column levels under a given source. The way I understood this property is as a transient buffer before tests are applied to the table (which assumes the current pattern of one schema doc per table). However, looking more closely, it looks like the property is not actually used. 
@krivard Since you wrote this piece, can you confirm whether I can take this property out to have the class properties align more closely with the information present in the actual schema yml files?

test/unit/dbt_helper_test.py
marianneke: Yes, exactly

test/unit/dbt_helper_test.py
krivard: It was probably a false path I went down while debugging pydantic validations that never got pruned afterwards -- should be safe to remove. It'll squawk if it's not.

test/unit/dbt_helper_test.py
jdangerx: > But generally I do feel there is value even in testing quite basic functionality, since the implementation can change

I agree it can helpful to know things like "`_get_existing_row_counts` reads the rowcounts from the correct file and makes sure the partition is a string." 

But, that's not what this test does!

Instead, it's saying that "`_get_existing_row_counts` calls `pd.read_csv` and returns whatever it gets back." If the implementation changes we'll have to change this test even if the function still works correctly, which seems annoying. And it doesn't actually show that any of the intended behavior is happening - an implementation that's just

```python
def _get_existing_row_counts(target):
    return pd.read_csv("/dev/null")
```

still passes this test, which seems bad.

What do you think of changing this test to test for the high-level interface the function presents rather than the implementation inside it?

test/unit/dbt_helper_test.py
jdangerx: Yeah, that's nice! And I'm glad we have a test about that case now too!

test/unit/dbt_helper_test.py
jdangerx: Cool! non-blocking question: What prompted this fixture-based approach? My first/naive thought would have been to instantiate `DbtSchema` objects directly and attempt to round-trip in various formats.

test/unit/dbt_helper_test.py
jdangerx: See above note about testing for interface/business logic rather than implementation.

test/unit/dbt_helper_test.py
jdangerx: See above note about testing for interface/business logic rather than implementation.

jdangerx (changes requested):
I agree that it can be valuable to test functions that may not be part of the Official Public Interface of the module, especially in development, but still have some comments about some of the existing tests - I resolved the conversations that I have no more questions on!

test/unit/dbt_helper_test.py
marianneke: If you have only one year of data (with year being the partition column), that would also have the same behavior as no partition, similar to a column that has a null value. That is what I meant.

test/unit/dbt_helper_test.py
marianneke: I updated this test. I _am_ patching `pd.read_csv` still, but now testing the functionality (for example making sure the partition column returns a string)

test/unit/dbt_helper_test.py
marianneke: We can probably simply test its behavior through the `test_update_row_counts` function. Do you agree with removing this test?

test/unit/dbt_helper_test.py
zaneselvans: Oh okay, yes that would be the same -- just one group that constitutes the whole table.

src/pudl/scripts/dbt_helper.py
```diff
@@ -325,19 +311,58 @@ def _write_row_counts(row_counts: pd.DataFrame, target: str = "etl-full"):
...
     target: str = "etl-full",
     clobber: bool = False,
     update: bool = False,
 ) -> UpdateResult:
     """Generate updated row counts per partition and write to csv file within dbt project."""
+    model_path = _get_model_path(table_name, data_source)
+    schema_path = model_path / "schema.yml"
+    schema = DbtSchema.from_yaml(schema_path)
+
+    partition_columns = _extract_row_count_partitions(schema)
```
zaneselvans: `_extract_row_count_partitions()` expectes a `DbtTable` rather than a `DbtSchema` so this call fails -- the `DbtSchema` has no `data_tests` attribute from which the partition column can be extracted.

src/pudl/scripts/dbt_helper.py
```diff
@@ -193,13 +179,13 @@ def add_column_tests(
         return schema
 
     @classmethod
-    def from_table_name(cls, table_name: str, partition_column: str) -> "DbtSchema":
+    def from_table_name(cls, table_name: str) -> "DbtSchema":
         """Construct configuration defining table from PUDL metadata."""
         return cls(
             sources=[
                 DbtSource(
                     version=2,
```
zaneselvans: It doesn't look like `DbtSource` has a `version` attribute, but `DbtSchema` does.

src/pudl/scripts/dbt_helper.py
```diff
@@ -108,23 +108,10 @@ def add_column_tests(self, column_tests: dict[str, list]) -> "DbtSource":
 
         return self.model_copy(update={"columns": list(columns.values())})
```
zaneselvans, line 110: Both `add_source_tests` and `add_column_tests` return `DbtTable` objects (since they're making an updated copy the model they're part of) but they claim to be returning `DbtSource` objects, and those types aren't compatible with each other.

src/pudl/scripts/dbt_helper.py
```diff
@@ -108,23 +108,10 @@ def add_column_tests(self, column_tests: dict[str, list]) -> "DbtSource":
...
-                    "partition_column": partition_column,
-                }
-            }
-        ]
-
     @classmethod
-    def from_table_name(cls, table_name: str, partition_column: str) -> "DbtSchema":
+    def from_table_name(cls, table_name: str) -> "DbtSchema":
         """Construct configuration defining table from PUDL metadata."""
         return cls(
```
zaneselvans: This claims to be returning a `DbtSchema` but is actually returning a `DbtTable`

zaneselvans (changes requested):
I haven't looked at the code changes in detail yet but as I noted in the previous review, it seems like the script can't be used to update row counts due to some confusion around what types / classes.

I changed the partition column for `core_epacems__hourly_emissions`  from null to `year` (since the table has a billion rows and it seems like it would be good to split it up a little) by editing the `schema.yml`.  Then I ran:

```console
dbt_helper update-tables --row-counts --clobber core_epacems__hourly_emission
```

It still results in the following error, since it's attempting to call `_extract_row_count_partitions()`  on a `DbtSchema` when what it expects is a `DbtTable`.

```
  File "/Users/zane/miniforge3/envs/pudl-dev/bin/dbt_helper", line 8, in <module>
    sys.exit(dbt_helper())
             ^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1161, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 1443, in invoke
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/click/core.py", line 788, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/scripts/dbt_helper.py", line 561, in update_tables
    update_row_counts(
  File "/Users/zane/code/catalyst/pudl/src/pudl/scripts/dbt_helper.py", line 324, in update_row_counts
    partition_columns = _extract_row_count_partitions(schema)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl/src/pudl/scripts/dbt_helper.py", line 439, in _extract_row_count_partitions
    if table.data_tests:
       ^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.12/site-packages/pydantic/main.py", line 991, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'DbtSchema' object has no attribute 'data_tests'
```

With type checking turned on I see number of warnings about incompatible types and invalid attributes in the script. It seems like there may have been some confusion about what the `DbtSchema`, `DbtTable` and `DbtSource` classes were referring to, with some incorrect method return types and argument type annotations, some of which I've commented on below.

src/pudl/scripts/dbt_helper.py
marianneke: I made these edits

src/pudl/scripts/dbt_helper.py
marianneke: Made changes to all the class definitions to ensure they have the correct properties

test/unit/dbt_helper_test.py
marianneke: I am reverting this cleanup so I can do it in a less busy PR together with the other cleanup changes that were called out.

src/pudl/scripts/dbt_helper.py
marianneke: Reversed the cleanup and will tackle that in a separate PR that includes the other issues you mentioned

src/pudl/scripts/dbt_helper.py
```diff
@@ -289,20 +275,17 @@ def _calculate_row_counts(
...
-        )
+    if partition_column is None:
+        partition_expr = "''"
+        group_by_clause = ""
     else:
-        row_count_query = (
-            f"SELECT '' as partition, COUNT(*) as row_count FROM '{table_path}'"  # noqa: S608
-        )
+        partition_expr = partition_column
+        group_by_clause = f"GROUP BY {partition_column}"
```
jdangerx: I guess we're setting `partition_expr = partition_column` here, but should this be `f"GROUP BY {partition_expr}"` instead?

src/pudl/scripts/dbt_helper.py
```diff
@@ -325,33 +308,78 @@ def _write_row_counts(row_counts: pd.DataFrame, target: str = "etl-full"):
 
 def update_row_counts(
     table_name: str,
-    partition_column: str = "report_year",
+    data_source: str,
     target: str = "etl-full",
     clobber: bool = False,
     update: bool = False,
```
jdangerx: It looks like right now `clobber` and `update` do the same thing (which is clobber existing counts) for `update_row_counts`... is this what we want?

If yes, we should probably condense these two params into one (or disallow `--update` for the row counts update... this would probably look like pulling `update-tables --schema` and `update-tables --row-counts` into separate click commands (`update-schema` and `update-row-counts`) so they can have different CLI interfaces.

If no, we should probably have `--update` do a little diff of the rowcounts & only overwrite if no existing partitions change counts.

Since this PR's been going for a while, we could also just disallow `--update` for row counts for now, and make this change later if people want that diffing functionality.

jdangerx (approved):
Some concerns about e.g. `--update` vs. `--clobber` behavior, but nothing blocking at this point.

I tried running the script with the following scenarios and they all worked as expected:

* `partition_column: (2 * report_year)`: partition column replaced with 2x year, rowcounts the same
* No `partition_column` at all: one row for the whole table
* `partition_column: null`: one row for the whole table
* `partition_column: utility_id_pudl`: one row per utility ID
* no test defined at all: removed completely

We'll need to update the documentation, too. And I think the test cases that Zane was asking for would still be good to add. But this seems usable and I want to get this in before folks need to do another big round of rowcount updates.

dbt/models/eia/core_eia__yearly_fuel_receipts_costs_aggs/schema.yml
```diff
@@ -7,7 +7,7 @@ sources:
           - expect_columns_not_all_null
           - check_row_counts_per_partition:
               table_name: core_eia__yearly_fuel_receipts_costs_aggs
-              partition_column: report_date
+              partition_column: "YEAR(report_date)"
```
zaneselvans: **Blocking:** Since this is no longer a column, I think it would be clearer to rename the argument (here and everywhere else). Maybe to `partition_expr` which is already what's being used in the macro and some other contexts?

**Blocking:** The documentation for this test (in `dbt/tests/data_tests/generic_tests/schema.yml`) should be updated to reflect the new behavior.

**Non-Blocking:** For the other tests with this kind of expression I've been using `EXTRACT(YEAR FROM report_date)` since it seems to have wider support across different SQL backends. `YEAR(report-date)` works for DuckDB but I think does not work for Postgres or SQLite or BigQuery? If we already have a bunch of other stuff that's DuckDB specific maybe it doesn't matter. I'm not suggesting we try and make everything compatible with Postgres, just if there are simple alternatives that are more general maybe it's worth trying to use them?

src/pudl/scripts/dbt_helper.py
zaneselvans: I think partly this complexity is coming from putting both schema and row-count updates in the same subcommand. Might we want them to be two separate subcommands instead? If we don't anticipate trying to do both things at the same time that might be a cleaner UI.

* `dbt_helper update-schema` (**only** affects schema.yml files)
* `dbt_helper row-counts` (**only** affects `etl_full_row_counts.csv`)

But let's not put that in this PR.

marianneke: Do you think updating counts of existing partitions should not be allowed with `--update`? I would assume that updating counts is allowed, but just removing partitions (e.g. if the partition column is changed) would only be allowed with `--clobber`

zaneselvans: I think the fact that we have both of these options is accidental and artificial. The idea is to have a guard against making potentially destructive changes so that if, say, someone has just spent a bunch of time manually adding new test cases to some `schema.yml` files they don't accidentally blow that work away with a programmatic update.

The row counts should only ever be updated programmatically using the script, so I don't think it's particularly important to have the guard for row counts. Which I think we could implement more cleanly if the schema and row-count updates were not in the same command.

dbt/models/eia/core_eia__yearly_fuel_receipts_costs_aggs/schema.yml
marianneke: All three points addressed

dbt/macros/row_counts_per_partition.sql
```diff
@@ -1,4 +1,4 @@
-{% macro row_counts_per_partition(model, table_name, partition_column, force_row_counts_table=none) %}
+{% macro row_counts_per_partition(model, table_name, partition_expr, force_row_counts_table=none) %}
```
zaneselvans, line 1: Updated this last use of `partition_column`

src/pudl/dbt_wrapper.py
```diff
@@ -19,6 +19,8 @@
 
 logger = get_logger(__name__)
 
+DBT_DIR: Path = PUDL_ROOT_PATH / "dbt"
+
```
zaneselvans, line 23: Created an importable constant since we want access to this in other contexts as well.

src/pudl/scripts/dbt_helper.py
```diff
@@ -148,7 +134,7 @@ def add_source_tests(self, source_tests: list) -> "DbtSource":
             update={"tables": [self.tables[0].add_source_tests(source_tests)]}
         )
 
-    def add_column_tests(self, column_tests: dict[list]) -> "DbtSource":
```
zaneselvans, line 151: Simplified invalid type annotation.

src/pudl/scripts/dbt_helper.py
```diff
@@ -193,13 +179,12 @@ def add_column_tests(
         return schema
 
     @classmethod
-    def from_table_name(cls, table_name: str, partition_column: str) -> "DbtSchema":
+    def from_table_name(cls, table_name: str) -> "DbtSchema":
         """Construct configuration defining table from PUDL metadata."""
         return cls(
             sources=[
                 DbtSource(
-                    version=2,
```
zaneselvans, line 201: Invalid `version` attribute was resulting in errors.  Only the `DbtSchema` has a version.

src/pudl/scripts/dbt_helper.py
```diff
@@ -256,11 +255,11 @@ def _get_local_table_path(table_name):
 
 
 def _get_model_path(table_name: str, data_source: str) -> Path:
-    return Path("./dbt") / "models" / data_source / table_name
+    return DBT_DIR / "models" / data_source / table_name
```
zaneselvans, line 258: It's annoying and confusing to have the script only be runnable from a single directory. Make it agnostic as to where it's executed.

src/pudl/scripts/dbt_helper.py
```diff
@@ -269,16 +268,15 @@ def _get_existing_row_counts() -> pd.DataFrame:
 
 def _calculate_row_counts(
     table_name: str,
-    partition_column: str = "report_year",
```
zaneselvans, line 286: Something is fishy here.

- There's no guarantee that the table we're looking at will contain a `report_year` column. If it doesn't have `report_year` and `partition_expr is None` then we'll attempt to partition by `report_year` and I think we should get an error.
- In the case of a table that has `report_year` if we wanted to leave it unpartitioned, it seems like that would be impossible (since the null value would trigger assignment to `report_year`)
- Given the conditional below it seems like we really want a default value of "''" (since that's what we assign `partition_expr` to when it's None)
- However, when I switch to using that default value, I get errors from `partition_expr: null` -- the result of casting that to `VARCHAR` appears to be "None" and there's no column named "None" in the table so the query fails.
- It seems to work as it is now -- I can repartition `core_epacems__hourly_emissions` using `year` or `state` or `null` and they all behave as expected -- but I don't understand why it doesn't work with a default value of "''" or why it does work with a default value of `report_year`.

src/pudl/scripts/dbt_helper.py
jdangerx: nit: could this be `dict[str, list]`?

src/pudl/scripts/dbt_helper.py
marianneke: @zaneselvans I agree we should remove the default assignment of `report_year`, but the behavior you described is not what happens. See here what happens when you pass a `None` value to a function with a default assignment for that value; it overrides it:
```python
>>> def function_with_default(some_par="a"):
...     print(some_par, type(some_par))
... 
>>> function_with_default()
a <class 'str'>
>>> function_with_default(None)
None <class 'NoneType'>
```

On the default assignment: the function that does the computation checks for `None` and then reassigns it to `"''"`, so that function would not work properly. We can set it to `None` by default if you prefer no partition as the default (which makes sense to me).

a clean way to reduce confusion is to not set `partition_expr = "''"` inside the function, which has led to the confusion, so now I add a new variable `partition_expr_sql` which I simply set to `partition_expr` if not null and `"''"` if it is null. Then I set `partition_expr: str | None = None` as the default value to eliminate confusion there.

zaneselvans: Ah okay, so the "None" it gets when no argument is provided (triggering the default assignment) is different from an explicit "None". That makes the behavior make more sense.

Is there a reason to have a default of None vs. not having a default at all?

src/pudl/scripts/dbt_helper.py
marianneke: Ah, I thought I had removed those. Maybe I accidentally added them back in with a rebase, sorry.

src/pudl/scripts/dbt_helper.py
jdangerx: > However, when I switch to using that default value, I get errors from partition_expr: null -- the result of casting that to VARCHAR appears to be "None" and there's no column named "None" in the table so the query fails.

I can't reproduce this with the following diff:

```diff
diff --git a/dbt/models/ferc1/out_ferc1__yearly_steam_plants_fuel_sched402/schema.yml b/dbt/models/ferc1/out_ferc1__yearly_steam_plants_fuel_sched402/schema.yml
index cc659a2aa..e022554b7 100644
--- a/dbt/models/ferc1/out_ferc1__yearly_steam_plants_fuel_sched402/schema.yml
+++ b/dbt/models/ferc1/out_ferc1__yearly_steam_plants_fuel_sched402/schema.yml
@@ -7,7 +7,7 @@ sources:
           - expect_columns_not_all_null
           - check_row_counts_per_partition:
               table_name: out_ferc1__yearly_steam_plants_fuel_sched402
-              partition_expr: report_year
+              partition_expr: null
         columns:
           - name: report_year
           - name: utility_id_ferc1
diff --git a/src/pudl/scripts/dbt_helper.py b/src/pudl/scripts/dbt_helper.py
index 7cc1a57ab..42672b617 100644
--- a/src/pudl/scripts/dbt_helper.py
+++ b/src/pudl/scripts/dbt_helper.py
@@ -268,7 +268,7 @@ def _get_existing_row_counts() -> pd.DataFrame:
 
 def _calculate_row_counts(
     table_name: str,
-    partition_expr: str,
+    partition_expr: str = "''",
 ) -> pd.DataFrame:
     table_path = _get_local_table_path(table_name)
```

Given:

* `partition_expr: null`, `partition_expr:`,  and just no `partition_expr`  line in the YAML all lead to this function being called with `partition_expr=None`
* This function is *never* called without an explicit `partition_expr` (the only callsite is `_calculate_row_counts(table_name, partition_expr)`)

I think  `partition_expr: str | None` and `partition_expr: str | None = None` are both fine.

jdangerx (approved):
Took another look with Zane's new changes that make the script work again. It does indeed appear to work, and I think the strange behavior in `_calculate_row_counts` has been addressed. So I think this is good to merge.

We should probably add some sort of test that runs the full command against some easy-to-interpret, stable data so we know that things are working as expected even through weird merges & rebases, but that seems out of scope for this PR.

src/pudl/scripts/dbt_helper.py
zaneselvans: Was trying not to dig into the structure but that seems right. Updated it here and below.

zaneselvans (approved):

zaneselvans:
Coverage failed, but because of a problem uploading to CodeCov having to do with gpg. Not sure what changed -- the CodeCov action hasn't changed versions recently. Trying again, with code coverage report printout before upload.

PR #4438, opened 2025-07-16
Enforce more specific PUDL resource name constraints.

### LINKED ISSUE #4436: Enforce PUDL table naming conventions in Pydantic Resource model
### Overview

### Success Criteria

* [ ] ...

### Next steps
* [ ] ...

zaneselvans:
# Overview

- Adds a more specific Annotated string type that uses Pydantic's StringConstraints to enforce our PUDL table naming conventions (at least, more than just being SnakeCase)
- Requires the two-part (layer, data source) portion of the table before the double underscore, and only allows valid data sources, and the `core` `_core` and `out` layers (since those are the only ones we write to the database or Parquet).
- The new type was easy. Tracking down all the tests where we had dummy table names that didn't conform was a little more annoying.
- I'm not sure if this is really *the* way to do it, but it's *a* way to do it. Open to other ideas!

Closes #4436 

# Testing

- Ran the unit tests locally

## To-do list


src/pudl/metadata/classes.py
```diff
@@ -181,11 +183,34 @@ def _get_jinja_environment(template_dir: DirectoryPath = None):
 SnakeCase = Annotated[
     str,
     StringConstraints(
-        min_length=1, strict=True, pattern=r"^[a-z_][a-z0-9_]*(_[a-z0-9]+)*$"
+        min_length=1,
+        strict=True,
+        pattern=r"^[a-z_][a-z0-9_]*(_[a-z0-9]+)*$",
```
zaneselvans, line 188: No changes here -- just forced args onto individual lines.

src/pudl/metadata/classes.py
```diff
@@ -181,11 +183,34 @@ def _get_jinja_environment(template_dir: DirectoryPath = None):
...
 
+PudlResourceName = Annotated[
+    str,
+    StringConstraints(
+        min_length=1,
+        strict=True,
+        pattern=rf"^(_?core_|out_)({'|'.join(ALL_PUDL_SOURCES)})__([a-z0-9_]+)$",
+    ),
+]
+"""A valid PUDL resource name conforming to our naming conventions :class:`str`."""
```
zaneselvans, line 201: This was the lightest weight way I could see to do this. However, it means that the runtime enforcement of type conformance needs to be done with the function below. So maybe it would be better if it were a "real" Pydantic model, if it's not being used to type a field in a Pydantic model (as it is in the `Resource` class and everything that hangs off of that)

src/pudl/metadata/sources.py
```diff
@@ -943,3 +943,5 @@
     },
 }
 """Data source attributes by PUDL identifier."""
+
+ALL_PUDL_SOURCES = sorted(set(SOURCES.keys()).union(["epa", "eia", "ferc"]))
```
zaneselvans, line 947: - We use the `eia` and `ferc` "sources" for tables that are generic to the whole agency (i.e. they come from more than one data source).
- However, we haven't been great about really defining what that means -- there are many tables with multiple "sources" listed which nevertheless have a single form listed in the resource name.
- The `epa` source on the other hand is a rogue, used I think because `epacamd_eia` pertains to two data sources (it's glue...) and is the **only** data source we have that contains an underscore, which would mess up our table naming if we used it as the data source.
- So maybe `epacamd_eia` shouldn't actually be allowed here. Maybe we should rename that source.

src/pudl/scripts/dbt_helper.py
```diff
@@ -257,8 +261,10 @@ def _schema_diff_summary(old_schema: DbtSchema, new_schema: DbtSchema):
     return [line for line in stripped if line != ""]
 
 
-def get_data_source(table_name: str) -> str:
+def get_data_source(table_name: PudlResourceName) -> str:
     """Return the data source element of the table's name."""
+    if not is_valid_pudl_resource_name(table_name):
+        raise ValueError(f"Invalid PUDL resource name: {table_name}")
```
zaneselvans, line 267: Avoids the possibility of a malformed table name resulting in undefined behavior.

zaneselvans (commented):
Most of the changed lines are fixing names of test tables, both in the unit tests and in the docstring examples / tests in `classes.py` Substantiative changes in:
- `classes.py` (not the docstrings)
- `dbt_helper.py`
- `sources.py`

Issue #4432, opened 2025-07-15
In `out_eia923__fuel_receipts_costs` why do we only have MHSA mine IDs for 2025, and only have County FIPS codes for 2009-2010?


zaneselvans:

jdangerx:
@zaneselvans sounds like the success criteria for this would be something like:
* we figure out and fix what's causing the data loss of FIPS/MSHA codes
* we document the weirdness we know about in the source data
Does that sound right?

zaneselvans:
@jdangerx 

* I think debugging what is causing this problem and fixing it is going to be way more trouble than it's worth.
* However, I think we can avoid the problem entirely by skipping the super janky coalmine table normalization process entirely.
* It was always an ad-hoc special case, and while it makes sense from a database normalization point of view hypothetically, in practice the coalmine data is quite messy and incomplete, and the fuel receipts and costs table is unusual in that it truly has no natural primary key, merging data back into it is extra annoying.
* If we stop stripping the coalmine table out of the fuel receipts and costs table at all and just treat those columns as if they were normal data columns in the FRC table, we can avoid all of this complexity and definitely retain all of the coalmine data.
* With all of those coalmine records in the FRC table we can do some cleanup of the values and have the code be simpler and the data more useful than it is now.

## Success Criteria
- Remove the `core_eia923__entity_coalmine` entirely
- Simplify the transform of the `core_eia923__fuel_receipts_costs` table to no longer strip out the coalmine data.
- Instead, clean the coalmine columns in place (standardize the `county_id_fips` codes, coalmine type codes, MSHA mine ids, etc.)
- Remove the step(s) where we merge the coalmine data back into the `out_eia923__*_fuel_receipts_costs` table(s) since the data will already be there, and the thing that's getting merged in will no longer exist.
- Update the `expect_columns_not_all_null` test criteria to reflect the new distribution of nulls, if any.

PR #4442, opened 2025-07-17
Strange nulls

### LINKED ISSUE #4407: Investigate unusual patterns of column nulls
# Overview

While working on #4105 / #4382 we found unusual patterns of null values in some columns, which should be further investigated and either fixed or understood.

## Solved
- [x] Are FERC 1 plant `capex_annual_per_*` columns really expected to be entirely null in 1994?
    - Yes -- these derived fields depend on `capex_annual_addition` which is calculated by diffing adjacent years of total capital additions, so the value is undefined until the 2nd year of data.  
- [x] Did FERC 1 plant `asset_retirement_cost` really only appear in 2003?
     - Yep, that's when it appeared!
- [x] Why no special case for `can_switch_when_operating` in `out_eia__monthly_generators`?
    - The special case of `year >= 2009` is required in the `core_eia860__scd_generators_multifuel` because that table didn't exist in earlier years, but this column has been reported since the beginning in the generators table.
- [x] In `core_eia860__assn_boiler_generator` why is the `boiler_generator_assn_type_code` only available for 2009-2012?
    - Yes, as per spreadsheets and the field documentation -- this is A (actual) vs. T (theoretical). From EIA, not us. 
- [x] @cmgosnell do you have an intuition as to why the `ferc1_generator_agg_id_plant_gen` column in `out_eia__yearly_assn_plant_parts_plant_gen` would only have non-null values between 2005-2019? I'm not finding the column name anywhere in our transforms.
    - This is a manually assigned ID used to group generators that have an unusually complex relationship with their FERC plants. It hasn't been updated since 2020 and it only affects a very small number of records. So it's expected to be null in the recent years. 

Image

## Further Discussion
- [x] Why does `core_eia860__scd_boilers` have so many `row_conditions` that exclude 2025?
    - A few columns that cross over from the EIA-860 generators mean that the table runs through 2025, even though the boilers data only exists for the main EIA-860 years (2024).
    - This makes me wonder if we should take another look at the eia/eia860 tables that are being extended into 2025 by a few eia860m columns and decide whether they really provide any marginal value, or if they're really just NULL for any useful purpose in the years where only EIA-860M data is available -- which seems confusing / unhelpful. Maybe we should limit their extent to just the EIA-860 years instead.
- [x] In `core_eia860__assn_boiler_generator` why is there no `data_maturity` before 2009? We assign it.
    - Weirdly, `data_maturity` is completely NULL in 2008, 0% NULL for 2009-2011, a *tiny* bit NULL in 2012, slightly more null in 2013, and then 40-50% null for 2014 and later. 
    - Looking at the `raw_eia860__boiler_generator_assn` table, it has a fully populated `data_maturity` field, so my guess is that we are not assigning any `data_maturity` to the additional records which we infer. This seems like it could e confusing / misleading to others too... maybe we should have an "inferred" or "estimated" or other code that goes in that column to indicate the nature of the row when we're filling things in and it's not really reported data with a well-defined maturity?
    - I think this is one instance of a wider problem where we propagate `data_maturity` but it gets confused along the way -- either because there's filling, or the records are no longer just traceable to data with a single source.

## Success Criteria

- All of the above patterns of nullness have been investigated.
- If they are due to a bug, the bug has been fixed.
- If they are correct, we understand why the unusual pattern of nullness exists and is expected.
- If any of these strange nulls are more complex and require further investigation or refactoring, they've been split off into an independent sub-issue.

zaneselvans:
# Overview

- Reviewed the unusual patterns of null column-years that were identified in #4105 / #4382 
- Added explanations for the columns that required them, either in field descriptions or the dbt `schema.yml`
- Pulled #4432 out separately because it's a more involved fix.
- Removed the no longer used `not_all_null.sql` dbt test definition.

Closes #4407 

# Testing

- I re-ran the dbt tests locally on my full ETL output just to make sure `not_all_null.sql` wasn't being used somewhere.

## To-do list


dbt/models/eia860/core_eia860__scd_boilers/schema.yml
```diff
@@ -5,6 +5,11 @@ sources:
       - name: core_eia860__scd_boilers
         data_tests:
           - expect_columns_not_all_null:
+              description: >
+                This table has a lot of row_conditions because it has a few columns that
+                are found in the EIA-860M generators table (and so appear in the most),
+                recent monthly data) but most of the columns are specific to boilers,
+                which are not reported in the EIA-860M.
```
zaneselvans: The fact that this table is included and has a lot of row conditions is a bit of an artifact of the fact that the `row_conditions` were generated before I created the `ignore_eia860m_nulls` argument.  I chose to leave them in (even though I think the table would not have any failures in the fast ETL now) because I think we'll want to deploy this test more widely and it encodes real expectations about which column-years should be null.

dbt/models/ferc1/out_ferc1__yearly_all_plants/schema.yml
```diff
@@ -9,7 +9,15 @@ sources:
...
                 of data have been processed, since the moving window will exclude the
                 first couple of years in the range. This means we can't define a simple
-                range of years in which we expect them to be non-null.
+                range of years in which we expect them to be non-null. The
+                capex_annual_addition column is calculated by taking the difference
+                between adjacent capex_total values, so it will always be null in the
+                first year of data... at least in the case of the DBF data -- the XBRL
+                data (2021 and later) includes both the current and previous year's
+                values, so even if we only process a single recent year of data, we will
+                still end up with a non-null value. The capex_annual_per_* colusmn are
```
zaneselvans:
```suggestion
                still end up with a non-null value. The capex_annual_per_* columns are
```

src/pudl/etl/__init__.py
```diff
@@ -16,7 +16,6 @@
     load_asset_checks_from_modules,
     load_assets_from_modules,
 )
-from dagster._core.definitions.cacheable_assets import CacheableAssetsDefinition
```
zaneselvans: - This class definition moved in Dagster in v1.11 causing import errors.
- We're importing it from a non-public part of their API.
- We aren't actually handling the case that it shows up in below.
- So it seemed like removing it was best.

src/pudl/output/eia923.py
```diff
@@ -310,7 +311,7 @@ def _add_fuel_cost_per_mmbtu_source_col(
         the input source. Apply to the original source before any imputations have been
         applied and apply this directly after a new source of fuel cost has been added.
         """
-        if "fuel_cost_per_mmbtu_source" not in frc_df:
+        if "fuel_cost_per_mmbtu_source" not in frc_df.columns:
```
zaneselvans, line 314: Be more explicit for readability.

src/pudl/output/eia923.py
```diff
@@ -169,9 +169,10 @@ def out_eia923__generation_fuel_combined(
     """Denormalize the `generation_fuel_combined_eia923` table.
 
     This asset first combines the :ref:`core_eia923__monthly_generation_fuel` and
-    :ref:`core_eia923__monthly_generation_fuel_nuclear` into a single table with a uniform primary
-    key (consolidating multiple nuclear unit IDs into a single plant record) and then
-    denormalizes it by merging in some addition plant and utility level columns.
+    :ref:`core_eia923__monthly_generation_fuel_nuclear` into a single table with a
+    uniform primary key (consolidating multiple nuclear unit IDs into a single plant
+    record) and then denormalizes it by merging in some addition plant and utility level
+    columns.
```
zaneselvans, line 175: Just reflowing the text.

src/pudl/metadata/fields.py
```diff
@@ -1593,11 +1593,20 @@
     "expense_type": {"type": "string", "description": "The type of expense."},
     "ferc1_generator_agg_id": {
         "type": "integer",
-        "description": "ID dynamically assigned by PUDL to EIA records with multiple matches to a single FERC ID in the FERC-EIA manual matching process.",
+        "description": (
+            "ID dynamically assigned by PUDL to EIA records with multiple "
+            "matches to a single FERC ID in the FERC-EIA manual matching process. "
+            "The ID is manually assigned and has not been updated since 2020."
+        ),
```
zaneselvans: Add color to this and the following descriptions, to explain the weird pattern of null values.

test/integration/glue_test.py
```diff
@@ -195,7 +195,7 @@ def test_for_fk_validation_and_unmapped_ids(
         id_cols,
     )
 
-    if label_df:
+    if label_df is not None:
```
zaneselvans, line 198: This was giving me a warning b/c T/F value of a dataframe is not defined.

src/pudl/transform/eia923.py
```diff
@@ -435,13 +435,14 @@ def _yearly_to_monthly_records(df: pd.DataFrame) -> pd.DataFrame:
 
 
 def _coalmine_cleanup(
-    cmi_df: pd.DataFrame, _core_censuspep__yearly_geocodes
+    cmi_df: pd.DataFrame,
+    _core_censuspep__yearly_geocodes: pd.DataFrame,
```
zaneselvans, line 439: Just a type annotation.

src/pudl/transform/eia923.py
```diff
@@ -435,13 +435,14 @@ def _yearly_to_monthly_records(df: pd.DataFrame) -> pd.DataFrame:
...
 ) -> pd.DataFrame:
     """Clean up the core_eia923__entity_coalmine table.
 
-    This function does most of the core_eia923__entity_coalmine table transformation. It is separate
-    from the coalmine() transform function because of the peculiar way that we are
-    normalizing the :ref:`core_eia923__monthly_fuel_receipts_costs` table.
+    This function does most of the core_eia923__entity_coalmine table transformation. It
+    is separate from the coalmine() transform function because of the peculiar way that
+    we are normalizing the :ref:`core_eia923__monthly_fuel_receipts_costs` table.
```
zaneselvans, line 445: Just reflowing text.

dbt/models/eia860/core_eia860__scd_boilers/schema.yml
```diff
@@ -5,6 +5,11 @@ sources:
       - name: core_eia860__scd_boilers
         data_tests:
           - expect_columns_not_all_null:
+              description: >
+                This table has a lot of row_conditions because it has a few columns that
+                are found in the EIA-860M generators table (and so appear in the most),
```
jdangerx:
```suggestion
                are found in the EIA-860M generators table (and so appear in the most
```

dbt/models/ferc1/out_ferc1__yearly_all_plants/schema.yml
```diff
@@ -9,7 +9,15 @@ sources:
                 Exactly which years have rolling averages will depend on what years
                 of data have been processed, since the moving window will exclude the
                 first couple of years in the range. This means we can't define a simple
-                range of years in which we expect them to be non-null.
+                range of years in which we expect them to be non-null. The
+                capex_annual_addition column is calculated by taking the difference
+                between adjacent capex_total values, so it will always be null in the
+                first year of data... at least in the case of the DBF data -- the XBRL
```
jdangerx, line 15: non-blocking - did you consider using some sort of row condition here like `report_year > MIN(report_year)`? That would capture *some* error cases but not all (report years 2021-2025, 2021 rolling value should be non-null but that's not checked). Or we could use some sort of fancy subquery thing to get the equivalent of the Python `min(min(report_year), 2021)` to capture all this English in code.

src/pudl/metadata/fields.py
jdangerx: curious - What's the impact of us not updating the FERC-EIA manual matching since 2020?

zaneselvans: @cmgosnell says there are a whopping 260 records total affected by these manual mappings across all years, out of (I think) millions on the PPE overall, so whatever the impact is, it's likely very small. This was done with / for RMI when they were being very granular about the FERC-EIA stuff.

src/pudl/etl/__init__.py
zaneselvans: Had to make this change on `main` for this week's dependency updates.

dbt/models/ferc1/out_ferc1__yearly_all_plants/schema.yml
zaneselvans: No, I didn't really, though I could see how that might work. I've been more thinking of this test as describing our expectations about in which years we expect to see what data, and making that contingent on the vagaries of the ETL settings or the funny double-reporting of data in XBRL didn't seem in line with that implicit conceptualization of the test. So I excluded them because given these dependencies there is no stable expectation about the years in which we would expect to see these columns

src/pudl/metadata/fields.py
zaneselvans: I'll update the description to make it clear that this only touches a few 100 records across all years.

dbt/models/ferc1/out_ferc1__yearly_all_plants/schema.yml
zaneselvans: I just gave this a quick try, and it doesn't actually work out of the box, because the `row_condition` is going into the `WHERE` clause, which can't include aggregations, including `MIN(report_year)` so I think we should punt this detail for now.

zaneselvans:
~Whooops, found a bug having to do with the dtype of `report_year` -- it's a string, not a date or year! wtf!~ Sorry paranoia on my part

PR #4447, opened 2025-07-18
Add cites from July newsletter

krivard:
# Overview

## What did you change?

Three new papers cite us! Wildfires, enforcement, and "passing the savings onto not our customers, no"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,9 +71,12 @@ docs-clean:
 	rm -rf docs/data_dictionaries/code_csvs
 	rm -f docs/data_sources/eia*.rst
 	rm -f docs/data_sources/epacems*.rst
+	rm -f docs/data_sources/epacamd*.rst
 	rm -f docs/data_sources/ferc*.rst
 	rm -f docs/data_sources/gridpathratoolkit*.rst
+	rm -f docs/data_sources/nrelatb*.rst
 	rm -f docs/data_sources/phmsagas*.rst
+	rm -f docs/data_sources/vcerare*.rst
 
 # Note that there's some PUDL code which only gets run when we generate the docs, so
 # we want to generate coverage from the docs build.
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -33,6 +33,10 @@ Bug Fixes
 Documentation
 ^^^^^^^^^^^^^
 
+* Migrated table description metadata into new format for:
+
+  * EPA: see issue :issue:`4395` and PR :pr:`4398`.
+
 * Added data source pages for:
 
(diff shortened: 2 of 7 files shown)


zaneselvans:
I continue to be unreasonably surprised by every piece of evidence that people are using our data.

docs/catalyst_cites.bib
```diff
@@ -1,3 +1,44 @@
...
+author = {Gowrisankaran, Gautam and Langer, Ashley and Zhang, Wendan},
+title = {Policy Uncertainty in the Market for Coal Electricity: The Case of Air Toxics Standards},
+journal = {Journal of Political Economy},
+volume = {133},
+number = {6},
+pages = {1757-1795},
+year = {2025},
+doi = {10.1086/734779},
+URL = {https://doi.org/10.1086/734779},
+eprint = {https://doi.org/10.1086/734779},
```
zaneselvans: The eprint field seems to be generating a bad link:

[arXiv:https://doi.org/10.1086/734779](https://arxiv.org/abs/https://doi.org/10.1086/734779)

And actually ctrl-F-ing through the bibliography I see that there are a lot of other bad arXiv links too (and a few good ones)

docs/catalyst_cites.bib
```diff
@@ -1,3 +1,44 @@
...
+
+@article{SINGH2025107475,
+title = {Wires and fire: Wildfire investment and network cost differences across {California}’s power providers},
+journal = {The Electricity Journal},
+volume = {38},
+number = {3},
+pages = {107475},
+year = {2025},
+issn = {1040-6190},
+doi = {https://doi.org/10.1016/j.tej.2025.107475},
```
zaneselvans: I think the DOI field needs to be just the bare DOI, rather than a URL -- this is producing:

[doi:https://doi.org/10.1016/j.tej.2025.107475](https://doi.org/https://doi.org/10.1016/j.tej.2025.107475)

as it tries to construct a DOI URL itself.

zaneselvans (approved):
- I went ahead and fixed the DOI / eprint formatting issues I saw across both of our external citation .bib files
- Also added docs/**.bib to the list of file patterns for which changes don't trigger the integration tests in the merge queue.

zaneselvans:
Huh. @jdangerx I tried to add the `.bib` files to the list of files that can be changed without triggering the integration tests in the merge queue, but it still seems to be running the integration tests.

Oh. Lol. I guess it's running the integration tests because I edited the `pytest.yml` file to add `docs/**.bib` so it would ignore them. Next time!

PR #4448, opened 2025-07-20
Update conda lockfile for week of 2025-07-20

pudlbot:

### DIFF
--- a/src/pudl/etl/__init__.py
+++ b/src/pudl/etl/__init__.py
@@ -16,7 +16,6 @@ from dagster import (
     load_asset_checks_from_modules,
     load_assets_from_modules,
 )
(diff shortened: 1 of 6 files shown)


src/pudl/etl/__init__.py
```diff
@@ -16,7 +16,6 @@
     load_asset_checks_from_modules,
     load_assets_from_modules,
 )
-from dagster._core.definitions.cacheable_assets import CacheableAssetsDefinition
```
zaneselvans, line 19: We're importing this class from a private module and not using it, and where it is within the Dagster library has changed, so I'm removing it for now.

zaneselvans (approved):

PR #4464, opened 2025-07-21
Migrate resource description metadata for eia tables

### LINKED ISSUE #4424: Migrate resource description metadata for eia (51 tables, 1116 words)

krivard:
# Overview

Closes #4424.

Old descriptions

## What did you change?

New structure for all tables `*_eia__*`

Weird bits:
* 33 of the 51 tables affected by this PR list a single data source, but are named for the generic `eia` pseudosource and thus appear with the warning "Contains information from multiple raw inputs". Do we want to specify the actual, single source for these instead? If so, in this PR or in a followup pass? I've tagged them in the spreadsheet for easy followup either way.
* `out_eia__yearly_assn_plant_parts_plant_gen` hits an edge case in the automated system that causes it to be recognized simultaneously as an association table and a timeseries table, which displays as "Annual Association table providing connections between..." (with double capitalization). I think this is fine to leave for now, but I've filed a bug #4476 so we don't lose track of it.
* *Update: we'll handle this in a repo-wide pass later.* There is a mix of styles in how we refer to forms. I started out trying to standardize them but ran out of steam and left the rest however the were specified in the old description text, so long as the sentence still parsed.
  * the EIA-XXX vs the EIA XXX
  * form EIA-XXX vs form EIA XXX
  * EIA-XXX data vs EIA XXX data

## Documentation

# Testing

Wizard previews and datapackage build


src/pudl/metadata/resources/eia.py
```diff
@@ -448,7 +500,10 @@
         "field_namespace": "eia",
     },
     "core_eia__codes_operational_status": {
-        "description": "Codes and metadata pertaining to operational status reported to EIA. Compiled from EIA-860 instructions and EIA-923 file layout spreadsheets.",
+        "description": {
+            "additional_summary_text": "operational status reported to EIA.",
+            "additional_details_text": "Compiled from EIA-860 instructions and EIA-923 file layout spreadsheets.",
```
cmgosnell, line 505: just a note. i am 97% sure this is the case for all of these EIA codes.

src/pudl/metadata/resources/eia.py
```diff
@@ -576,7 +639,9 @@
         "field_namespace": "eia",
     },
     "core_eia__entity_generators": {
-        "description": "Static generator attributes compiled from across the EIA-860 and EIA-923 data.",
+        "description": {
+            "additional_summary_text": "generators compiled from across the EIA-860 and EIA-923.",
+        },
```
cmgosnell, line 646: I think it would be good to add a new general usage warning about these harvested tables. they are the `core_eia__entity_*` tables and the `core_eia860__scd_*` tables.

src/pudl/metadata/resources/eiaapi.py
```diff
@@ -4,9 +4,9 @@
 
 RESOURCE_METADATA: dict[str, dict[str, Any]] = {
     "core_eia__yearly_fuel_receipts_costs_aggs": {
-        "description": (
-            "Aggregated fuel receipts and costs from the EIA bulk electricity data."
-        ),
+        "description": {
+            "additional_summary_text": "aggregated fuel receipts and costs from the EIA bulk electricity data.",
+        },
```
cmgosnell: i think this one would be good to just say this comes from the eiaapi... honeslty not really sure why we decided to not just use the actual data source as the name here

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -4,15 +4,16 @@
...
-Practically speaking, a plant is a collection of generator(s). There are many
-attributes of generators (i.e. prime mover, primary fuel source, technology
-type). We can use these generator attributes to group generator records into
-larger aggregate records which we call "plant parts". A plant part is a record
-which corresponds to a particular collection of generators that all share an
-identical attribute and utility owner. E.g. all of the generators with unit_id=2, or all
-of the generators with coal as their primary fuel source.
+        "description": {
+            "additional_summary_text": "the aggregation of all EIA plant parts, for use in matching to FERC 1.",
+            "additional_details_text": (
```
cmgosnell: its organized duplication but omigosh so much duplication
```suggestion
            "additional_summary_text": "the aggregation of all EIA plant parts, for use in matching to FERC 1.",
            "usage_warnings": ["aggregation_hazard"],
            "additional_details_text": (
```

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -88,7 +91,9 @@
         "field_namespace": "eia",
     },
     "out_eia__yearly_generators_by_ownership": {
-        "description": "A mega table of all EIA generators with ownership integrated.",
+        "description": {
+            "additional_summary_text": "all EIA generators with ownership integrated.",
+        },
```
cmgosnell: oof this is oos for this pr 100% but this is too short for what this table is and has no pk.

cmgosnell: also also
```suggestion
        "description": {
            "additional_summary_text": "all EIA generators with ownership integrated.",
            "usage_warnings": ["aggregation_hazard"],
        },
```

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -130,16 +135,18 @@
...
-        "description": """In order to easily determine what generator records are associated with every
-plant part record, we made this association table. This table associates every plant part
-record (identified as ``record_id_eia``) from the ``out_eia__yearly_plant_parts`` table to the associated
-'plant_gen' records (identified as ``record_id_eia_plant_gen``) from the same table. The plant part
-records have a one to many relationship to the associated 'plant_gen' records. The 'plant_gen' records
-are also included in the plant part records, so for these records there is a one to one relationship
-and the ``record_id_eia`` and the ``record_id_eia_plant_gen`` are the same.
+        "description": {
+            "additional_summary_text": "plant parts (identified as ``record_id_eia``) and  'plant_gen' records (identified as ``record_id_eia_plant_gen``) from the :ref:`out_eia__yearly_plant_parts` table.",
+            "additional_details_text": """
```
cmgosnell: same same same all
```suggestion
            "additional_summary_text": "plant parts (identified as ``record_id_eia``) and  'plant_gen' records (identified as ``record_id_eia_plant_gen``) from the :ref:`out_eia__yearly_plant_parts` table.",
            "usage_warnings": ["aggregation_hazard"],
            "additional_details_text": """
```

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
cmgosnell: also i agree with your note in sheet that "the aggregation" doesn't really make sense.  I think removing "the aggregation of" makes sense to me even though as a summary line not knowing what the heck a plant part is is... a little weird but fine

src/pudl/metadata/resources/mcoe.py
```diff
@@ -6,19 +6,28 @@
...
-estimates."""
-        ),
+estimates.""",
+            "usage_warnings": [
+                "estimated_values",
+                {
+                    "type": "custom",
+                    "description": "Due to coverage problems in other tables, some generators have heat rate estimates but not fuel cost estimates.",
+                },
+            ],
```
cmgosnell: i cribbed this from myself from ferc all plants table.. i don't know if it rings as clear here but there are a lot of generators in this table that for example dont have any fuel cost or heat rate estimates because they are only reported in eia860 not in 923.

```suggestion
            "usage_warnings": [
                "estimated_values",
                {
                    "type": "custom",
                    "description": "Due to coverage problems in other tables, some generators have heat rate estimates but not fuel cost estimates.",
                },
                {
                    "type": "custom",
                    "description": "Not all columns are originally reported in or calculable from the input tables. Expect nulls.",
                    },
            ],
```

src/pudl/metadata/resources/eia.py
krivard: Warning about what?

src/pudl/metadata/resources/eiaapi.py
krivard: the data source name changed recently from eia_bulk_elec to eiaapi 🌈 ⭐

cmgosnell: aaaaaah now it makes senseeee. okay cool good to know and good note in sheeeet

src/pudl/metadata/resources/eia.py
cmgosnell: ah i meant to put this note here as a reminder to come back and write some suggested usage warning for the harvesting process... you seem to be a better copy editor than i am but i will try and please  make it smoother on the ears/brain if you so desire.... something like: "Data has been normalized from several EIA sources. If there are multiple values reported for the same entity, PUDL typically chooses the most consistent value reported". 🤷🏻

cmgosnell (approved):
Imma approve this knowing that I'd really like to include a warning about the eia harvesting process but that it doesn't need to be in this pr and overall this looks heckin grrreaaat

PR #4473, opened 2025-07-22
feat: add IAP to usage metrics dashboard

jdangerx:
# Overview

One part of catalyst-cooperative/pudl-usage-metrics-dashboard#32.

This consolidates the changes I made within the GCP console to turn on [Identity Aware Proxy](https://cloud.google.com/run/docs/securing/identity-aware-proxy-cloud-run) for the usage metrics service.

## What did you change?

Added the limited amount of terraform configuration that was available and wrote comments about what was not available yet. Presumably future versions of the Google provider will allow for more fine-grained configuration in Terraform.

# Testing

Ran `terraform apply` - plan looked good. Went and looked at the [console](https://console.cloud.google.com/run/detail/us-east1/pudl-usage-metrics-dashboard/security?inv=1&invt=Ab3eDA&project=catalyst-cooperative-pudl) and saw that after application, the IAP configuration was still there.

## To-do list


terraform/main.tf
```diff
@@ -110,7 +110,7 @@ resource "google_sql_database_instance" "mlflow_backend_store" {
   region           = "us-central1"
   database_version = "POSTGRES_14"
   settings {
-    tier = "db-f1-micro"
+    tier              = "db-f1-micro"
```
jdangerx, line 113: Just from `terraform fmt`  - suppose we should add that as a pre-commit hook eventually.

terraform/main.tf
```diff
@@ -6,7 +6,7 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = "6.14.1"
+      version = "6.45.0"
```
jdangerx, line 9: Had to upgrade to get the IAP stuff to work at all. You'll need to run `terraform init -upgrade`.

terraform/pudl-usage-metrics-dashboard.tf
```diff
@@ -88,9 +88,16 @@ resource "google_sql_user" "pudl_usage_metrics_dashboard_readonly" {
...
+  provider            = google-beta
   name                = "pudl-usage-metrics-dashboard"
   location            = "us-east1"
   deletion_protection = false
+  # TODO 2025-07-22: latest GCP provider (6.45.0) doesn't let us configure more than just 'enabled' and 'disabled'
+  # I added catalyst-cooperative-pudl-admins@catalyst.coop to the IAP policy here.
+  # I also had to click a button in the console to give a service account `run.invoke` on this service, but that policy doesn't show up in IAM anywhere.
+  launch_stage         = "BETA"
+  iap_enabled          = true
+  invoker_iam_disabled = true
```
jdangerx, line 100: This then disables the whole 'publicly visible to the Internet' thing.

terraform/pudl-usage-metrics-dashboard.tf
jdangerx, line 1: The `pudl_usage_metrics_dashboard_password` secret should be removed once https://github.com/catalyst-cooperative/pudl-usage-metrics-dashboard/pull/35 is merged. But not yet.

terraform/.terraform.lock.hcl
jdangerx, line 1: This is just a bunch of changes due to upgrading the Google provider version.

terraform/pudl-usage-metrics-dashboard.tf
```diff
@@ -151,11 +158,3 @@ resource "google_cloud_run_v2_service" "pudl_usage_metrics_dashboard" {
     }
   }
 }
-
-
-resource "google_cloud_run_v2_service_iam_member" "pudl_usage_metrics_dashboard_internet" {
```
jdangerx, line 156: Now that we're using IAP instead of IAM to manage access to the page, we don't need this block anymore.

zaneselvans (approved):
Ehhh, I can approve but I don't have any idea what IAP vs IAM means. Or really how we were authenticating before or now... it seems like we want to let anybody with a catalyst.coop email address to log in and see the metrics, right?

jdangerx:
> Ehhh, I can approve but I don't have any idea what IAP vs IAM means. Or really how we were authenticating before or now... it seems like we want to let anybody with a catalyst.coop email address to log in and see the metrics, right?

Yeah it's all pretty confusing, here's an attempt at a summary:

The old way (IAM):
* we set up a rule on the service that says "anyone on the internet is able to access this service"
* we put up a password for authentication, so only password-knowers can actually log in and see the dashboards.

The new way (IAP):
* we set up a rule that says "the following users within our organization have access to this service" - I've set it to `catalyst-cooperative-pudl-admins@catalyst.coop` for now. This can apparently only be changed manually right now but it's pretty easy to go to the console and add/remove/edit this rule.
* when we try to access the dashboard, it makes you log in to Google first. If the rule above says so, Google lets you pass through to the dashboard.
* We don't need to have a password on the dashboard itself anymore - though it's still there until catalyst-cooperative/pudl-usage-metrics-dashboard#35 is merged.

Does that clear things up?

zaneselvans:
So IAP is another way of delegating the authentication to Google, which is GCP specific, and depends on permissions groups that we have set up within GCP. Rather than OAuth+auth0, which also delegates authentication to Google (or GitHub, or whoever) but has no idea who anybody is or whether they have permissions -- it's just a way of tracking a particular user.

PR #4386, opened 2025-07-03
feat: improve quantile constraints failure output

jdangerx:
# Overview

Closes #4293 .

## What problem does this address?

Weighted quantiles test output only tells you which quantile failed, not what the actual values are and/or any other context. For that we have a whole other command: `dbt run-operation debug_quantile_constraints` which is a little bit annoying to call.

So... let's call it for the user!

## What did you change?

Added some logic to differentiate between `expect_quantile_constraints` tests and the other tests which we can simply run the compiled code for.

Added a handler for the `expect_quantile_constraints` tests which builds up the `run-operation` command and grabs the output.

Changed `NodeContext.context` to be a string for more flexible presentation. The formatting etc. is now the job of the caller.

## Documentation

Make sure to update relevant aspects of the documentation:

- [ ] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html): reference the PR and related issues.
- [ ] Update relevant Data Source jinja templates (see `docs/data_sources/templates`).
- [ ] Update relevant table or source description metadata (see `src/metadata`).
- [ ] Review and update any other aspects of the documentation that might be affected by this PR.

# Testing

How did you make sure this worked? How can a reviewer verify this?

1. Mess with the expected values for the `out_eia__yearly_generators` table:
```diff
diff --git a/dbt/models/output/out_eia__yearly_generators/schema.yml b/dbt/models/output/out_eia__yearly_generators/schema.yml
index 43aaae293..b03f385d8 100644
--- a/dbt/models/output/out_eia__yearly_generators/schema.yml
+++ b/dbt/models/output/out_eia__yearly_generators/schema.yml
@@ -33,8 +33,8 @@ sources:
                   row_condition: fuel_type_code_pudl='gas' and report_date>=CAST('2015-01-01' AS DATE) and capacity_factor<>0.0
                   constraints:
                     - quantile: 0.65
-                      min_value: 0.4
-                      max_value: 0.7
+                      min_value: 0.6
+                      max_value: 0.6
                     - quantile: 0.15
                       min_value: 0.005
                     - quantile: 0.95
diff --git a/dbt/seeds/etl_full_row_counts.csv b/dbt/seeds/etl_full_row_counts.csv
index 14c1648e7..22e7f8527 100644
--- a/dbt/seeds/etl_full_row_counts.csv
+++ b/dbt/seeds/etl_full_row_counts.csv
@@ -2369,7 +2369,7 @@ out_eia__yearly_boilers,2022,5785
[...]

PR #4474, opened 2025-07-23
Expand and update data validation documentation

### LINKED ISSUE #4126: Add developer documentation covering dbt

zaneselvans:
# Overview

- Migrate content from `dbt/README.md` into our `docs/` build.

Closes #4126


dbt/tests/data_tests/generic_tests/schema.yml
zaneselvans, line 1: Mostly this is reflowing long descriptions.

The only real change was removing the old `test_not_all_null` entry, since it's been replaced and was resulting in a warning.

docs/dev/debugging_quantile_checks.rst
zaneselvans, line 1: Brought this content into the main data_validation.rst file to avoid the one itty-bitty nested page.

docs/dev/testing.rst
```diff
@@ -4,22 +4,20 @@
 Testing PUDL
 ===============================================================================
 
-.. toctree::
-   :hidden:
-
-   debugging_quantile_checks
-
```
zaneselvans, line 11: I consolidated this into the `data_validation` file to avoid a single file at this nesting level.

docs/dev/testing.rst
```diff
@@ -4,22 +4,20 @@
...
 We use `pytest <https://pytest.org>`__ to specify software unit & integration tests,
-and to coordinate data validation tests.  There are several ``pytest`` commands stored
-as targets in the PUDL ``Makefile`` for convenience and to ensure that we're all running
-the tests in similar ways by default.
```
zaneselvans, line 15: Removed references to the pytest validation tests and directed to `data_validation` / dbt docs where appropriate.

docs/dev/testing.rst
```diff
@@ -4,22 +4,20 @@
...
 
-To run the checks that will be required for a PR to be merged into the ``main`` branch
-on GitHub, you can use the following command:
+To run the checks that will be run on a PR by our continuous integration (CI) on GitHub
+before it's merged into the ``main`` branch you can use the following command:
 
 .. code-block:: console
 
-    $ make pytest-coverage
+    $ make ci
```
zaneselvans, line 17: This was slightly inaccurate.

docs/dev/testing.rst
```diff
@@ -63,11 +54,9 @@ by ``pytest`` are prefixed with ``pytest-``
 In addition to running the ``pytest-unit`` and ``pytest-integration`` targets mentioned
 above there are also:
 
-* ``pytest-validate``: The full data validation tests, which run on an already
-  existing PUDL DB.
 * ``pytest-integration-full``: The integration tests, but run on all years of data
-  rather than just the most recent year. This test also assumes you already have a live
-  PUDL DB.
+  rather than just the most recent year. This test assumes you already have the
+  complete outputs from the full PUDL ETL in your ``$PUDL_OUTPUT`` directory.
```
zaneselvans, line 59: Lest anybody be surprised that their computer is busy for the next half hour.

docs/dev/testing.rst
```diff
@@ -77,23 +66,20 @@ above there are also:
...
 * ``nuke``: delete your existing FERC and PUDL databases, rebuild
-  them from scratch, and run all of the tests and and data validations (akin to running
+  them from scratch, and run all of the tests and data validations (akin to running
   the nightly builds) for an extensive check of everything. This will take 3 hours or
   more to complete, and likely fully utilize your computer's CPU and memory.
 * ``install-pudl``: Remove your existing ``pudl-dev`` ``conda`` environment and
   reinstall all dependencies as well as the ``catalystcoop.pudl`` package defined by
   the repository in ``--editable`` mode for development.
 * ``docs-build``: Remove existing PUDL documentation outputs and rebuild from scratch.
-* ``dagster``: start up the Dagster UI (will remain running in your terminal until you
```
zaneselvans, line 95: We removed this a while ago!

docs/dev/testing.rst
```diff
@@ -77,23 +66,20 @@ above there are also:
...
-There are a number of non-test ```make`` targets. To see them all open up the
-``Makefile``.
+There are several non-test ``make`` targets. To see them all open the ``Makefile``.
 
 * ``ferc``: Delete all existing XBRL and DBF derived FERC databases and metadata and
   re-extract them from scratch
-* ``pudl``: Delete your existing ``pudl.sqlite`` and re-run the full ETL from scratch.
-  Assumes that the FERC DBs already exist.
+* ``pudl``: Delete your existing ``pudl.sqlite`` DB and Parquet outputs and re-run the
+  full ETL from scratch. Assumes that the FERC DBs already exist.
```
zaneselvans, line 74: Updated this and some other refs to focus less on `pudl.sqlite` and more on the parquet outputs.

docs/dev/testing.rst
```diff
@@ -120,53 +106,21 @@ datastore instead by using our custom ``--tmp-data`` with ``pytest``:
...
-``$PUDL_OUTPUT/pudl.sqlite``. To run the full data validation against an existing
-database:
-
-.. code-block:: console
-
-    $ make pytest-validate
-
-The data validation cases that pertain to the contents of the data tables are
-currently stored as part of the :mod:`pudl.validate` module.
-
```
zaneselvans, line 148: This content has been removed.

docs/dev/testing.rst
```diff
@@ -120,53 +106,21 @@ datastore instead by using our custom ``--tmp-data`` with ``pytest``:
...
-each data table we distribute. These expectations are stored in
-``dbt/seeds/etl_full_row_counts.csv`` and they can be updated using the ``dbt_helper``
-script. If you can't run the full ETL locally, the nightly builds / branch deployments
-also generate updated row count expectations. So you can kick off the
-``build-deploy-pudl`` GitHub Action using the ``workflow_dispatch`` trigger on your
-branch on GitHub, and then download the updated ``etl_full_row_counts.csv`` file from
-the build outputs that are uploaded to
-``gs://builds.catalyst.coop/<build-id>/etl_full_row_counts.csv`` once the build has
-completed. See the :doc:`nightly_data_builds` documentation for more details on the
-nightly builds.
```
zaneselvans, line 159: Migrated this paragraph into the new `data_validation` page.

docs/dev/testing.rst
```diff
@@ -213,13 +167,11 @@ test suite and keeps the data and tests very clearly separated.
...
 operates on an existing database, and has nothing to do with the construction of that
-database. For example, the output routines:
+database. For example, the EPA CEMS specific tests:
 
 .. code-block:: console
 
-  $ pytest --live-dbs test/integration/output_test.py
-
-We also use this option to run the data validations.
+  $ pytest --live-dbs test/integration/epacems_test.py
```
zaneselvans, line 174: This output test no longer exists so I switched to one that does.

docs/dev/testing.rst
```diff
@@ -244,11 +196,3 @@ data to a temporary datastore that's later deleted automatically:
...
 
    $ pytest --tmp-data test/integration/etl_test.py
-
--------------------------------------------------------------------------------
-When dbt tests fail
--------------------------------------------------------------------------------
-
-If the test that failed is a quantile constraint (starts with
-``expect_quantile_constraints`` or ``source_expect_quantile_constraints``), we have a
-macro that can help debug the problem: :doc:`/dev/debugging_quantile_checks`
```
zaneselvans, line 254: Removed because it's part of the main `data_validation.rst` page now.

docs/dev/testing.rst
```diff
@@ -120,53 +106,21 @@ datastore instead by using our custom ``--tmp-data`` with ``pytest``:
...
 there's no testing environment to set up. Instead, it just uses your Python environment
 which should be the ``pudl-dev`` conda environment discussed in :doc:`/dev/dev_setup`.
 This is convenient if you're debugging something specific or developing new test cases.
 
+If you are working on integration tests, note that most of them require processed PUDL
+outputs. If you try to run a single integration test directly with pytest it will
+likely end up running the fast ETL which will take 45 minutes. If you have processed
+PUDL outputs locally already, you can use ``--live-dbs`` instead. This is only helpful
+if the thing you're testing isn't part of the ETL itself.
+
```
zaneselvans, line 123: Seemed like an important detail to avoid folks accidentally running the fast ETL!

src/pudl/metadata/fields.py
```diff
@@ -6905,9 +6905,9 @@
 
 def get_pudl_dtypes(
     group: str | None = None,
-    field_meta: dict[str, Any] | None = FIELD_METADATA,
-    field_meta_by_group: dict[str, Any] | None = FIELD_METADATA_BY_GROUP,
-    dtype_map: dict[str, Any] | None = FIELD_DTYPES_PANDAS,
+    field_meta: dict[str, Any] = FIELD_METADATA,
+    field_meta_by_group: dict[str, Any] = FIELD_METADATA_BY_GROUP,
+    dtype_map: dict[str, Any] = FIELD_DTYPES_PANDAS,
```
zaneselvans, line 6910: I was getting type annotation warnings here and so far as I can tell, these `| None` annotations can never end up being used, since if you say None, then the default values get used.

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+
+   Modern analytical query engines are extremely fast and memory efficient. By using dbt
+   with `DuckDB <https://duckdb.org/>`_ to query the PUDL `Parquet
+   <https://parquet.apache.org/>`_ outputs we're able to run thousands of validations
+   across hundreds of tables with billions of rows in a minute, instead of the 2-3 hours
+   it used to take our much less extensive validation tests to run. Plus we get to learn
+   SQL.
+
+--------------------------------------------------------------------------------
+Data validation guidelines
```
zaneselvans, line 29: We should come up with something for this section, but that's probably beyond the scope of this PR.

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+.. todo::
+
+   We should chat about the minimum level of data validation that we expect for a table
+
+   * row counts
+   * no entirely null columns
+   * anything else now?
+
+--------------------------------------------------------------------------------
+Example of typical data validation workflow
```
zaneselvans, line 50: I kinda wanted just a narrative / bulleted outline of "What is it like to work with these tests while working on a new table" or something, before diving into all of the details. Do others think that would be helpful?

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+tests you'll want to run the following from within the ``dbt/`` directory in your pudl
+repo:
+
+.. code-block:: bash
+
+    # Install additional dbt-specific dependencies in the dbt project
+    dbt deps
+    # Materializes dbt-specific "seed" tables from CSVs in the seeds/ directory
+    # In our case, these contain the expected row counts for PUDL tables
+    dbt seed
```
zaneselvans: These were both in the old docs, but I'm not sure if they're truly required -- running `dbt build` seems to rematerialize the seeds tables.

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+``$PUDL_OUTPUT/parquet/`` directory. It's important that you ensure the outputs you're
+testing are actually the result of the code on your current branch, otherwise you may
+be surprised when the data test fails in CI or the nightly builds.
+
+.. tip::
+
+   If you want to run the tests against the fast ETL outputs that are generated by the
+   integration tests without wiping out your local full ETL outputs, you can temporarily
+   set the ``$PUDL_OUTPUT`` environment variable to point at the tmpdir created by
+   ``pytest``. This is printed out at the beginning of the ``pytest`` run.
```
zaneselvans: Do others think this is a useful tip as it is? Or will people need more direction to find that temporary directory / re-set envvars?

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+  tables would be defined as `models <https://docs.getdbt.com/docs/build/models>`_ which
+  are somewhat analogous to `Dagster assets
+  <https://docs.dagster.io/guides/build/assets/defining-assets>`_.
+* As a SQL-based tool, dbt generally expects to be querying a database. However, in our
+  case the tables are stored as Apache Parquet files, which we query with SQL via
+  DuckDB. This means some of dbt's functionality is not available. For example, we can't
+  use `the dbt adapter object
+  <https://docs.getdbt.com/reference/dbt-jinja-functions/adapter>`_ in our test
+  definitions because it relies on being able to access the underlying database schema,
+  which doesn't exist (at least, not within the dbt context).
```
zaneselvans: Are there other quirks that folks have run into that they'd like to add to this list? Is this the right place for it?

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+
+.. note::
+
+   There are a handful of data validation tests that have been implemented using
+   `Dagster's asset checks <https://docs.dagster.io/guides/test/asset-checks>`_.
+   Typically these tests weren't well suited to SQL, weren't performance bottlenecks,
+   and had already been implemented in Python. E.g. :func:`pudl.validate.no_null_rows`.
+
+The ``dbt_helper`` script
+~~~~~~~~~~~~~~~~~~~~~~~~~
```
zaneselvans, line 147: Rather than trying to repeat all of the `dbt_helper` documentation here and make it a standalone section, I felt like it was more natural to work it into each of the thematic sections where it had something to offer. I think having the complete, canonical documentation for it be in the module / help messages makes sense.

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+``dbt/seeds/etl_full_row_counts.csv`` and they can be updated using the ``dbt_helper``
+script based on the observed row counts in your local PUDL Parquet outputs. If you can't
+run the full ETL locally, the nightly builds / branch build also generate updated row
+count expectations. After a branch build completes, you can download the updated
+``etl_full_row_counts.csv`` file from the build outputs that are uploaded to
+``gs://builds.catalyst.coop/<build-id>/etl_full_row_counts.csv`` See the
+:doc:`nightly_data_builds` documentation for more details on accessing the nightly build
+outputs. Replace the ``etl_full_row_counts.csv`` in your local PUDL git repo with the
+one you've downloaded and use ``git diff`` to see what has changed. Make sure to review
+the row count changes closely to see if there's anything unexpected.
```
zaneselvans, line 230: Is this enough / the right information to make the branch/nightly build row counts useful to folks?

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+
+--------------------------------------------------------------------------------
+Debugging data validation failures
```
zaneselvans, line 260: Needs to be fleshed out.

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+Some of the tests in the example above like ``expect_columns_not_all_null`` or
+``check_row_counts_per_partition`` are defined by us, and can be found in the SQL
+files with the same name under ``dbt/tests/data_tests/generic_tests/``
+
+Documentation for the tests that we define is in
+``dbt/testse/data_tests/generic_tests/schema.yml``
+
+.. todo::
+
+   * Integrate documentation of our existing generic tests into the docs build.
```
zaneselvans, line 449: Is there a short path to getting our generic test documentation building into our docs somehow? It feels a bit buried right now.

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+        tables:
+          - name: new_table_name
+            data_tests:
+              - expect_columns_not_all_null
+              - check_row_counts_per_partition:
+                  table_name: new_table_name
+                  partition_column: report_date
+
+--------------------------------------------------------------------------------
+Defining new data validation tests
```
zaneselvans, line 558: This section and the subsections under it need to be fleshed out.

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+
+In some cases you may need to modify a table or calculate some derived values before
+you can apply a test. There are two ways to accomplish this. First, you can add the
+table as a ``source`` as described above, then create a SQL file in the ``tests/``
+directory like ``tests/{data_source}/{table_name}.yml``.  From here you can construct a
+SQL query to modify the table and execute a test on the intermediate table you've
+created. ``dbt`` expects a SQL test to be a query that returns 0 rows for a successful
+test. See the ``dbt`` `source function
+<https://docs.getdbt.com/reference/dbt-jinja-functions/source>`_ for guidance on how to
+reference a ``source`` from a SQL file.
```
zaneselvans: This paragraph didn't really make sense to me, or seem like something we were actually doing, but maybe I'm missing something?

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+``check_row_counts_per_partition`` are defined by us, and can be found in the SQL
+files with the same name under ``dbt/tests/data_tests/generic_tests/``
+
+Documentation for the tests that we define is in
+``dbt/testse/data_tests/generic_tests/schema.yml``
+
+.. todo::
+
+   * Integrate documentation of our existing generic tests into the docs build.
+   * Convert all bespoke / singular tests into generic tests.
```
zaneselvans: @krivard I haven't looked into converting these yet. Do you think it's easy enough that we should just do it in this PR so that we neither need to document them, or leave them undocumented?

zaneselvans (commented):
I tried to organize this document so that the level of complexity/involvement of the tasks increases as you get further into it.

The main sections that still need to be fleshed out are:
- defining new SQL tests
- debugging failures

docs/dev/data_validation.rst
krivard: i'm in favor

docs/dev/data_validation.rst
krivard: I'd take them out, or relegate them to an "edge cases / troubleshooting" section

docs/dev/data_validation.rst
krivard: An example of the pytest output would be a good addition here

docs/dev/data_validation.rst
krivard: the quirks don't explain how to run dbt (and probably get in the way actually/make it seem more complicated than it is), but they do explain some of our design decisions, which is helpful for anyone who needs to get beyond the basic mechanics. Maybe they belong in a "why we did it this way" section instead?

docs/dev/data_validation.rst
krivard: this was a good call

docs/dev/data_validation.rst
krivard: At least three options here:

- Extract docs from tests/**/schema.yml
- Extract docs from `static_index.html` generated using `dbt docs generate --static --select path:tests/`
- Extract docs from `manifest.json` generated using `dbt docs generate --static --select path:tests/`

The static index is _very_ pretty when rendered but the source is completely unintelligible so unless we want to throw Playwright at it we probably want one of the other two.

I'd recommend using the manifest, which should be manageable when built using that path selection. It might even be only one line of `jq` to pull out the fields we want for each test, though using python to do that will be more legible.

docs/dev/data_validation.rst
krivard: Should be straightforward; they just need a macro signature, then be listed in schema.yml in one of the two (or n) tables being tested

docs/dev/data_validation.rst
krivard: this seems like it's awkwardly describing singular tests, which we don't [want to] use anymore. but this text has some bonus typos (it would be `{table_name}.sql`, not `.yml`) and with singular tests you don't ever modify (`UPDATE SET`) the table or create one (`CREATE`) so much as query it, perhaps saving a bunch of intermediate steps using `WITH AS`. but that has a similar enough feel to modifying a table/creating an intermediate table for me to think that's what was meant.

but frankly we should rewrite this and the preceding two sections with the guidance we actually want people to follow going forward, which will need some discussion on how to decide when to just use a generic test and when to pull stuff out into a macro so it can be unit-tested in turn.

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+and may run into memory limits. It's also easy to accidentally end up with local outputs
+that are the result of code from multiple different branches, and so may not be
+consistent with each other. If you're only altering a few tables, rematerializing them
+in Dagster and then running the specific dbt tests that apply to them and any tables
+downstream of them should work fine.
+
+Kicking off a branch build
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+But especially when we're doing big quarterly or annual updates, and dozens or hundreds
```
krivard: I see how this was meant to be a transition from the previous section, but I don't think we really need it -- and without it, the sentence will make more sense to someone jumping to this section directly.

```suggestion
When we're doing big quarterly or annual updates, and dozens or hundreds
```

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+of tables are changing simultaneously, it is helpful to be able to run the full ETL from
+scratch, run all of the data validation tests against the outputs, and use the results
+to update the test parameters (especially expected row counts) appropriately. This can
+be done by manually kicking off a PUDL deployment on your branch. To initiate a branch
```
krivard: Consider a paragraph break here for better skimmability
```suggestion
be done by manually kicking off a PUDL deployment on your branch. 

To initiate a branch
```

docs/dev/data_validation.rst
krivard: Branch build rowcounts are definitely useful and this probably gives folks enough information to track them down.

Nightly build rowcounts I'm less sure about. Under what circumstances would the nightly build rowcounts differ from what's in the repo?

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+* Using ``duckdb < path/to/compiled.sql``
+* Using DuckDB's ``.read path/to/compiled.sql`` to play with data interactively.
+* Go through a simpler example before getting into the complicated quantile checks test.
+
+Debugging quantile checks
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. todo::
+
+  This seems quite involved. Can we make it simpler? Improve the test failure output to
```
krivard, line 277: This section probably went out of date with https://github.com/catalyst-cooperative/pudl/pull/4386 and should be updated based on that new functionality.

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,688 @@
...
+which will produce the intermediate table you want to execute tests on. To use this
+approach, simply add a sql file to ``dbt/models/{data_source}/{table_name}/``. Now, add
+a SQL file to this directory named ``validate_{table_name}`` and define your model for
+producing the intermediate table here. Finally, add the model to the ``schema.yml`` file
+and define tests exactly as you would for a ``source`` table. See
+``models/ferc1/out_ferc1__yearly_steam_plants_fuel_by_plant_sched402`` for an example of
+this pattern.
+
+Note: when adding a model, it will be stored as a SQL ``view`` in the file
+``$PUDL_OUTPUT/pudl_dbt_tests.duckdb``.
```
krivard, line 643: This could be a good place to note that while most other compiled tests can be run from an empty duckdb prompt, compiled tests against these intermediate models require `dubkdb $PUDL_OUTPUT/pudl_dbt_tests.duckdb`

docs/dev/testing.rst
krivard: Want to update https://github.com/catalyst-cooperative/pudl/blob/main/.github/pull_request_template.md as well?

krivard (commented):
thankyou for organizing!

krivard:
I'm going to see if I can just do the singular -> generic conversion in under an hour

docs/dev/data_validation.rst
zaneselvans: Well, the row counts are generated by *any* build -- nightly or branch.  In the case of the branch builds, it's probably intentional -- you want to update the row counts before you merge into `main`.  In the case of the nightly builds, they'll probably be more helpful forensically.  Like if you merge a PR and you didn't *think* it was going to affect any of the row counts, so you didn't update them, and then the row counts change, the nightly build will fail. But it'll also generate the new row counts, which you can then hopefully use to fix the problem.

docs/dev/data_validation.rst
zaneselvans: I guess beyond documenting the data tests we've defined for ourselves, it would be useful to document what data tests are being applied to what data for users which I think the dbt docs build also puts out. Do we really want to parse their outputs and reconstruct those docs within Sphinx? Or do we want to make use of their docs outputs?

zaneselvans:
@krivard great, thank you for doing the test conversions!

test/integration/plant_parts_eia_test.py
zaneselvans, line 1: This test didn't blow up the GitHub runner, and also passed the CI, and didn't increase the time it took for the tests to run noticeably, so there doesn't seem to be any reason for it to only be run in the nightly builds w/ full ETL, and it was the last test under `validate`

Moving it here means we can really get rid of all remnants of `validate` in the docs, scripts, code, etc.

Makefile
```diff
@@ -132,12 +132,8 @@ pytest-coverage: coverage-erase docs-build pytest-ci
 
 .PHONY: pytest-integration-full
 pytest-integration-full:
-	pytest ${pytest_args} -n 4 --no-cov --live-dbs --etl-settings ${etl_full_yml} test/integration
-
-.PHONY: pytest-validate
-pytest-validate:
 	pudl_check_fks
-	pytest ${pytest_args} -n 4 --no-cov --live-dbs test/validate
+	pytest ${pytest_args} -n 4 --no-cov --live-dbs --etl-settings ${etl_full_yml} test/integration
```
zaneselvans: Remove the `pytestl-validate` target and move the FK check into the full integration tests target.

Makefile
```diff
@@ -134,21 +134,14 @@ pytest-coverage: coverage-erase docs-build pytest-ci
 pytest-integration-full:
 	pytest ${pytest_args} -n 4 --no-cov --live-dbs --etl-settings ${etl_full_yml} test/integration
 
-.PHONY: pytest-validate
-pytest-validate:
-	pudl_check_fks
-	pytest ${pytest_args} -n 4 --no-cov --live-dbs test/validate
-
```
zaneselvans, line 141: Remove the pytest-validate target because there's no more tests down there!

Makefile
```diff
@@ -134,21 +134,14 @@ pytest-coverage: coverage-erase docs-build pytest-ci
...
 # Run the full ETL, generating new FERC & PUDL SQLite DBs and EPA CEMS Parquet files.
 # Then run the full integration tests and data validations on all years of data.
 # NOTE: This will clobber your existing databases and takes hours to run!!!
 # Backgrounding the data validation and integration tests and using wait allows them to
 # run in parallel.
 .PHONY: nuke
 nuke: coverage-erase docs-build pytest-unit ferc pudl
-	pudl_check_fks
```
zaneselvans, line 149: `pudl_check_fks` isn't needed in here anywhere, because it's called programmatically within the integration tests.

docker/gcp_pudl_etl.sh
```diff
@@ -51,12 +51,6 @@ function run_pudl_etl() {
             --etl-settings "$PUDL_SETTINGS_YML" \
             --live-dbs test/integration test/unit \
             --no-cov &&
-        pytest \
-            -n auto \
-            --gcs-cache-path gs://internal-zenodo-cache.catalyst.coop \
-            --etl-settings "$PUDL_SETTINGS_YML" \
-            --live-dbs test/validate \
-            --no-cov &&
```
zaneselvans, line 59: No more validation tests.

pyproject.toml
```diff
@@ -65,6 +65,7 @@ dependencies = [
     "pytest-console-scripts>=1.4",
     "pytest-cov>=5",
     "pytest-mock>=3.14",
+    "pytest-order>=1.3",
```
zaneselvans, line 68: A plugin that lets us tell Pytest which tests to run first. Mostly for readability of the logs given our unusual use of fixtures to trigger the fast ETL.

pyproject.toml
```diff
@@ -360,10 +361,6 @@ zip = ">=3.0"
 
 [tool.coverage.run]
 source = ["src/pudl/", "test/integration/", "test/unit/"]
-omit = [
-    # Never hit by integration tests:
-    "src/pudl/validate.py",
-]
```
zaneselvans, line 366: Removed the derelict validation functions / plotting routines and added a unit test for the one weighted_quantile test, so now this should have fully coverage / get used in the ETL.

test/integration/ferc_xbrl_extract_test.py
```diff
@@ -28,9 +69,7 @@
         "core_ferc1__yearly_income_statements_sched114",
     ],
 )
-def test_filter_for_freshest_data(
-    ferc1_engine_xbrl, table_name: pd.DataFrame
-) -> pd.DataFrame:
+def test_filter_for_freshest_data(ferc1_engine_xbrl: sa.Engine, table_name: str):
```
zaneselvans, line 72: Changed the return type to None.

test/integration/ferc_xbrl_extract_test.py
```diff
@@ -13,6 +14,46 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.mark.order(1)
+def test_ferc1_xbrl2sqlite(ferc1_engine_xbrl: sa.Engine, ferc1_xbrl_taxonomy_metadata):
```
zaneselvans, line 18: Relocated this test here for to match the pattern of `ferc_dbf_extract_test.py`

The new pytest mark says "run this first" so that the `ferc1_engine_xbrl` fixture output is associated with a test that obviously depends on it.

test/integration/ferc_dbf_extract_test.py
```diff
@@ -15,6 +15,7 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.mark.order(1)
```
zaneselvans, line 18: Make this one of the 1st tests that gets run, so that the fixture executes in association with a test where it makes intuitive sense.

test/integration/etl_test.py
```diff
@@ -36,45 +38,6 @@ def test_pudl_engine(
         check_foreign_keys(pudl_engine)
 
 
-def test_ferc1_xbrl2sqlite(ferc1_engine_xbrl: sa.Engine, ferc1_xbrl_taxonomy_metadata):
```
zaneselvans, line 39: Relocated to the ferc_xbrl_extract test below.

test/integration/etl_test.py
```diff
@@ -7,16 +7,18 @@
...
 from dagster import build_init_resource_context
 
 import pudl
 from pudl.etl.check_foreign_keys import check_foreign_keys
+from pudl.resources import dataset_settings
 
 logger = logging.getLogger(__name__)
 
 
+@pytest.mark.order(2)
```
zaneselvans, line 21: Run this test 2nd, so that it output from the pudl_engine fixture execution shows up with an intuitively obvious test.

test/integration/etl_test.py
```diff
@@ -7,16 +7,18 @@
...
 import logging
 
-import pandas as pd
+import pytest
 import sqlalchemy as sa
 from dagster import build_init_resource_context
 
 import pudl
 from pudl.etl.check_foreign_keys import check_foreign_keys
+from pudl.resources import dataset_settings
```
zaneselvans, line 16: Without this import I was getting some warnings about types / non-existent attributes.

test/integration/dbt_test.py
```diff
@@ -33,6 +33,7 @@ def dbt_target(test_dir: Path, request) -> str:
     return dbt_target
 
 
+@pytest.mark.order(3)
```
zaneselvans, line 36: Run the dbt tests as soon as the FERC1 and PUDL ETL processes (fixtures) have completed.

src/pudl/validate.py
```diff
@@ -153,207 +152,3 @@ def weighted_quantile(data: pd.Series, weights: pd.Series, quantile: float) -> f
         return np.interp(quantile, Pn, df.data)
 
     return np.nan
-
-
-def historical_distribution(
```
zaneselvans, line 158: These functions all have to do with the weighted quantile tests, and those tests are different enough now that we'll need to re-imagine how we visualize them for debugging, and extract the parameters from the `dbt` tests, so I think we ought to go ahead and remove these.

src/pudl/validate.py
```diff
@@ -24,7 +23,7 @@ def __init__(self, message: str, null_rows: pd.DataFrame):
 
 def no_null_rows(
     df: pd.DataFrame,
-    cols="all",
+    cols: list[str] | str = "all",
```
zaneselvans, line 26: cleaning up type annotations here a little.

src/pudl/validate.py
zaneselvans: Plus that means we can add this module to our test coverage.

.codecov.yml
```diff
@@ -11,9 +11,6 @@ coverage:
         # sometimes (e.g. zenodo archive downloading)
         threshold: 0.1%
 
-ignore:
-  - "src/pudl/validate.py"
-
```
zaneselvans, line 16: I don't understand why codecov doesn't just use the coverage configuration in pyproject.toml, but it doesn't.

docs/dev/data_validation.rst
zaneselvans: You did it! :tada:

docs/dev/data_validation.rst
zaneselvans: I'm going to remove it.

docs/dev/data_validation.rst
zaneselvans: This section still needs some work and that detail seems related to the debugging failures section which is also not written yet, so I'm going to hold off for the moment.

docs/dev/data_validation.rst
krivard: I'll take it

docs/dev/data_validation.rst
zaneselvans: I can work on the vanilla version of this but would love some help on the macros, unit tests, and intermediate tables parts.

test/unit/validate_test.py
zaneselvans, line 1: - These are lightly edited LLM generated tests.
- Still no test for `pv.no_null_rows()` though. I need to play with it a bit more.

zaneselvans:
@krivard  Deferring further big docs edits to subsequent PRs so we can get all the other changes & the existing docs merged in.

dbt/tests/data_tests/generic_tests/schema.yml
```diff
@@ -205,13 +229,19 @@ macros:
...
+          value.
       - name: max_discrepancy_rate
         type: float
         description: Maximum acceptable fraction of records which can be counted as a discrepancy before the test should fail.
   - name: test_expect_columns_are_close
     description: >
-      Check that two float columns are close using the ``numpy.isclose`` equation for comparing columns. Generally this is used with the ``row_condition`` argument, as it's uncommon that there are two colums which are supposed to be entirely equal, but there may be a subset of rows which are supposed to be equal.
+      Check that two float columns are close using the ``numpy.isclose`` equation for
+      comparing columns. Generally this is used with the ``row_condition`` argument, as
+      it's uncommon that there are two colums which are supposed to be entirely equal,
```
krivard: since we're here
```suggestion
      Check that two float columns are close using the same criteria as 
      ``numpy.isclose``. Generally this is used with the ``row_condition`` argument, as
      it's uncommon that there are two columns which are supposed to be entirely equal,
```

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,694 @@
...
+PUDL Specific Design Choices
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Our usage of dbt is slightly unusual, since we rely on Dagster to coordinate our data
+pipeline, and are only using dbt for data validation. Some quirks of our setup to be
+aware of:
+
+* From dbt's point of view, the PUDL tables are
+  `sources <https://docs.getdbt.com/docs/build/sources>`_ -- external tables about which
+  it knows very little other than the table and column names. It assumes the tables will
```
krivard: will will
```suggestion
  it knows very little other than the table and column names. It assumes the tables
```

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,694 @@
...
+  <https://docs.dagster.io/guides/build/assets/defining-assets>`_.
+* As a SQL-based tool, dbt generally expects to be querying a database. However, in our
+  case the tables are stored as Apache Parquet files, which we query with SQL via
+  DuckDB. This means some of dbt's functionality is not available. For example, we can't
+  use `the dbt adapter object
+  <https://docs.getdbt.com/reference/dbt-jinja-functions/adapter>`_ in our test
+  definitions because it relies on being able to access the underlying database schema,
+  which doesn't exist (at least, not within the dbt context).
+* One exception to this is any intermediate tables that are defined as dbt models (see
+  below). These will be created as materliazed views in a DuckDB database at
```
krivard:
```suggestion
  below). These will be created as materialized views in a DuckDB database at
```

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,694 @@
...
+* investigate further in a Python notebook
+* fix a bug, re-run the pipeline, and repeat the check
+* adjust the quantile constraints (& consider leaving a dated note for followup in
+  case it gets worse)
+
+--------------------------------------------------------------------------------
+Applying pre-defined validations to existing data
+--------------------------------------------------------------------------------
+
+Applyng an existing generic test to an existing table should be as easy as editing
```
krivard:
```suggestion
Applying an existing generic test to an existing table should be as easy as editing
```

docs/dev/data_validation.rst
krivard: Followup: #4492

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,694 @@
...
+``schema.yml`` file and define tests exactly as you would for a ``source`` table. See
+``models/ferc1/out_ferc1__yearly_steam_plants_fuel_by_plant_sched402`` for an example of
+this pattern.
+
+Note: when adding a model, it will be stored as a SQL ``view`` in the file
+``$PUDL_OUTPUT/pudl_dbt_tests.duckdb``.
+
+--------------------------------------------------------------------------------
+Unmigrated Data Validation Docs (cannibalize)
+--------------------------------------------------------------------------------
```
krivard, line 647: Drop this section now or later?

src/pudl/validate.py
```diff
@@ -48,16 +47,16 @@ def no_null_rows(
...
     if cols == "all":
-        cols = df.columns
+        cols = list(df.columns)
 
     null_rows = df[cols].isna().sum(axis="columns") / len(cols) > max_null_fraction
     if null_rows.any():
         raise ExcessiveNullRowsError(
             message=(
-                f"Found {null_rows.sum(axis='rows')} excessively null rows in {df_name}.\n"
+                f"Found {len(null_rows)} excessively null rows in {df_name}.\n"
```
krivard: It looks like this change switches from printing out the number of rows failing our excessively-null test, to printing out the total number of rows, null or not. Check my read?

```
df: r rows, c columns
df[cols]: r rows, |cols| columns
df[cols].isna(): r rows, |cols| columns, 1 if na, 0 otherwise
df[cols].isna().sum(axis="columns"): r rows, 1 column, number of null columns in each row
df[cols].isna().sum(axis="columns") / len(cols): r rows, 1 column, percentage of null columns in each row
df[cols].isna().sum(axis="columns") / len(cols) > max_null_fraction: r rows, 1 column, 1 if excessively null, 0 otherwise
```

```suggestion
                f"Found {null_rows.sum()} excessively null rows in {df_name}.\n"
```

docs/dev/data_validation.rst
```diff
@@ -0,0 +1,694 @@
...
+run the full ETL locally, the nightly builds / branch build also generate updated row
+count expectations. After a branch build completes, you can download the updated
+``etl_full_row_counts.csv`` file from the build outputs that are uploaded to
+``gs://builds.catalyst.coop/<build-id>/etl_full_row_counts.csv`` See the
+:doc:`nightly_data_builds` documentation for more details on accessing the nightly build
+outputs. Replace the ``etl_full_row_counts.csv`` in your local PUDL git repo with the
+one you've downloaded and use ``git diff`` to see what has changed. Make sure to review
+the row count changes closely to see if there's anything unexpected.
+
+PUDL Specific Design Choices
```
zaneselvans, line 234: @krivard I pulled the quirks down here as a postscript rather than a prologue since that seems like less of a distraction.

test/unit/validate_test.py
```diff
@@ -0,0 +1,175 @@
...
+        ([1, 2, 3, 4, 5], [0, 0, 0, 1, 0], 0.5, 4),
+        # Uniform weights - should behave like regular quantile
+        ([1, 2, 3, 4, 5], [1, 1, 1, 1, 1], 0.5, 3),
+        ([1, 2, 3, 4], [1, 1, 1, 1], 0.5, 2.5),
+        # Edge cases for quantiles
+        ([1, 2, 3, 4, 5], [1, 1, 1, 1, 1], 0.0, 1),
+        ([1, 2, 3, 4, 5], [1, 1, 1, 1, 1], 1.0, 5),
+        # Non-uniform weights
+        ([1, 2, 3], [1, 2, 1], 0.5, 2),
+        ([10, 20, 30], [1, 3, 1], 0.5, 20),
```
krivard: Do these cases actually exercise non-uniform weights?

* 0.5 quantile on uniform [1, 2, 3] is also 2
* 0.5 quantile on uniform [10, 20, 30] is also 20

Consider examples that would differ from their uniformly-weighted cousins:

```suggestion
        ([1, 2, 4], [1, 2, 1], 0.5, 2),
        ([10, 20, 40], [1, 3, 1], 0.5, 20),
```

test/unit/validate_test.py
```diff
@@ -0,0 +1,175 @@
...
+        # Two data points
+        ([1, 3], [1, 1], 0.5, 2),
+        ([1, 3], [1, 3], 0.5, 2.5),
+        ([1, 3], [3, 1], 0.5, 1.5),
+        # Repeated values (tests the groupby logic)
+        ([1, 1, 2, 2], [1, 1, 1, 1], 0.5, 1.5),
+        ([1, 1, 3, 3], [1, 1, 1, 1], 0.5, 2),
+        # Zero weights mixed with non-zero
+        ([1, 2, 3, 4, 5], [0, 1, 0, 0, 0], 0.5, 2),
+        ([1, 2, 3, 4, 5], [0, 0, 0, 0, 1], 0.5, 5),
```
krivard: These are already covered by Basic above
```suggestion
```

test/unit/validate_test.py
```diff
@@ -0,0 +1,175 @@
...
+        ([1, np.inf, 3], [1, 1, 1], 0.5, 2.0),
+        ([1, -np.inf, 3], [1, 1, 1], 0.5, 2.0),
+        # NaN weights - function filters out rows with NaN weights
+        ([1, 2, 3], [1, np.nan, 1], 0.5, 2.0),
+        # Mix of NaN and inf - function filters out problematic rows
+        ([1, np.nan, np.inf, 4], [1, 1, 1, 1], 0.5, 2.5),
+    ],
+)
+def test_weighted_quantile_with_invalid_data(data, weights, quantile, expected):
+    """Test that weighted_quantile handles invalid data by filtering it out."""
```
krivard: the data is not invalid (invalid data generates an error)

```suggestion
def test_weighted_quantile_filtering(data, weights, quantile, expected):
    """Test that weighted_quantile filters out problematic rows of data and weights."""
```

test/unit/validate_test.py
```diff
@@ -0,0 +1,175 @@
...
+        ([1, 2, 3, 4], [1, 1, 1, 1], 0.5, 2.5),
+        # Edge cases for quantiles
+        ([1, 2, 3, 4, 5], [1, 1, 1, 1, 1], 0.0, 1),
+        ([1, 2, 3, 4, 5], [1, 1, 1, 1, 1], 1.0, 5),
+        # Non-uniform weights
+        ([1, 2, 3], [1, 2, 1], 0.5, 2),
+        ([10, 20, 30], [1, 3, 1], 0.5, 20),
+        # Different quantiles with same data
+        ([1, 2, 3, 4, 5], [1, 1, 1, 1, 1], 0.25, 1.75),
+        ([1, 2, 3, 4, 5], [1, 1, 1, 1, 1], 0.75, 4.25),
```
krivard: This is covered by "Edge cases" above since `weighted_quantiles` doesn't special-case 0 or 1 quantiles
```suggestion
```

test/unit/validate_test.py
```diff
@@ -0,0 +1,175 @@
...
+    column, threshold, n_outliers_allowed, expected_pass
+):
+    """Test the group_mean_continuity_check function with various scenarios.
+
+    Uses a test dataframe with different column patterns:
+    - stable_metric: Values around 100 with minimal variation
+    - gradual_growth: Steady 10% growth per year
+    - sudden_jump: Large 5x jump from 2022 to 2023
+    - volatile_metric: Random fluctuations around 100
+    - negative_change: Consistent 10% decline per year
```
krivard: I'm not sure of the implementation details here; is the % change computed just at the beginning of the sequence or at every data point? bc if it's the latter then for a consistent decline of 100 in absolute terms, the % change increases from 10% to ~15% by the time it reaches the end. Does that matter?

test/unit/validate_test.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@pytest.mark.parametrize(
+    "column,threshold,n_outliers_allowed,expected_pass",
+    [
+        # Test cases that should PASS
+        ("stable_metric", 0.1, 0, True),  # Very stable values, low threshold
+        ("gradual_growth", 0.2, 0, True),  # Gradual 10% growth per year, 20% threshold
+        ("sudden_jump", 5.0, 0, True),  # Large jump but threshold accommodates it
+        ("volatile_metric", 0.3, 1, True),  # Volatile but allowing 1 outlier
```
krivard: Is this exercising the outlier if volatile_metric never differs by more than 20%?

krivard (approved):
Some typos, one question about a change that seems to compute a substantially different value than before, and some test revisions, but nothing I feel the need to gatekeep on fixes

docs/dev/data_validation.rst
zaneselvans: Huh, why isn't the spell checker catching these?

docs/dev/data_validation.rst
zaneselvans: It seemed like raw materials that someone might want to pull into some of the unfinished sections, or at least topics / cases to make sure the new version covered (or superceded) so I was leaving it? It's slightly unclean but this is the dev docs and (hopefully) we're going to finish the docs this/next week...

docs/dev/data_validation.rst
zaneselvans: Moved it to the bottom of the section so it's less distracting, and referenced it from the "defining new tests" section.

test/unit/validate_test.py
zaneselvans: Whoops you're right. So long as this is checking the median value, and the only non-uniform weight is the on the middle (median) element, this will always pass. Need to put the non-uniform weight on a non-middle value I think.

test/unit/validate_test.py
zaneselvans: I think you're right. I've bumped one of the changes to be more than 30% and dropped the threshold to 20% and thankfully it's still passing.

PR #4483, opened 2025-07-27
Update conda lockfile for week of 2025-07-27

pudlbot:


zaneselvans (approved):

zaneselvans:
I was waiting on Dagster 1.11.3 since it looks like it will bring in Python 3.13 for us, but there are some issues. So going ahead and merging.

zaneselvans:
Alas the SQLAlchemy.org SSL certificate seems to have expired so the intersphinx build is failing. No rush so I'll hold off for the moment. Poked Mike on Mastodon.

PR #4491, opened 2025-07-28
remove row count tests for etl-fast

marianneke:
# Overview

We want to not run row count tests when they are likely to fail spuriously, such as for an etl-fast run. In general, we may want to think about which tests we think are relevant for certain targets, and use something like [test configs](https://docs.getdbt.com/reference/data-test-configs) to ensure we define explicitly when each test should run.

The current PR only addresses the first issue: ensuring the test query is bypassed when the target is etl-fast, and removing the update process for the row counts seed table for etl-fast in dbt helper.

### Open Questions
DBT has the option to add tags to tests, as a part of a [config block](https://docs.getdbt.com/reference/data-test-configs). How do we feel about having dbt test tags of the form `target:etl-full` that we can try to parse in our test to make it more clear from the schema.yml when the test is expected to run?

## What did you change?
- Remove references to etl-fast from dbt helper
- Modify test macro for row count checks (not generic macro computing row counts, but the dbt test) to pass if the target is not etl-full

## Documentation

# Testing

## To-do list


src/pudl/scripts/dbt_helper.py
zaneselvans, line 1: From the point of view of the `dbt_helper` script, it seems like we've eliminated the concept of the dbt target -- there's no option to specify it, and the full row counts csv is hard coded.

But dbt will still know internally what the target is, and the row count check behavior depends on it.

dbt/tests/data_tests/generic_tests/check_row_counts_per_partition.sql
```diff
@@ -1,6 +1,19 @@
...
+    'SKIPPED' as observed_partition,
+    NULL as expected_count,
+    NULL as observed_count,
+    NULL as diff_relative_to_expected
+  WHERE FALSE
+{% else %}
+  -- Run the real test logic
+  -- note 2025-03-28: logic is in a macro so that we can test it.
+  {{ row_counts_per_partition(model, table_name, partition_column) }}
+{% endif %}
```
zaneselvans: It feels a little bit dishonest (though it's convenient) to have the row count checks all pass happily when run against the fast outputs. Is there any way to return "SKIP" or some other result that's neither passing nor failing?

I feel like it might be better to adopt the pattern that we have in the integration tests, of intentionally excluding the row-count tests when we're running dbt tests against the fast ETL outputs.

My sense is that the overwhelming majority of the time when we're running tests against the fast outputs, it's going to be in the context of running the integration tests, either with pytest locally or in CI on GitHub.

I think our intention now is that all of the tests (except row counts) should pass meaningfully against both the fast and full ETL outputs. If that's true, would it be simpler and more self-consistent to get rid of the `etl-fast` dbt target altogether and require users to use `--exclude` / `--select` to avoid the row counts / run only the tests of interest if they are running against fast ETL outputs?

When do we expect folks to run the data tests against fast outputs manually? I think the main case is probably when there's a test that passes against the full ETL outputs, but is failing against the fast ETL outputs, and so the logic or parameters of the test need to be tweaked to work against both, in which case the user will probably be focused on running that single test that's being developed or tweaked, and not trying to run everything.

It looks like we could also set `enabled:` to be conditional on the value of `target.name` in the test specification:

```yaml
version: 2
sources:
  - name: pudl
    tables:
      - name: core_epacems__hourly_emissions
        data_tests:
          - expect_columns_not_all_null
              enabled: "{{ target.name == 'etl-full' }}"
              table_name: core_epacems__hourly_emissions
              partition_column: null
```

It feels a little duplicative to have to add this to every schema.yml, but we could do it with a quick script.  Conceptually it feels like it reflects what we really want -- "if we're not running against the full ETL, don't run this test at all", rather than "If we're not running against the full ETL this test passes." and this fact would be highlighted in the most readable place (the YAML file) rather than down inside the test definition.

src/pudl/scripts/dbt_helper.py
```diff
@@ -318,20 +316,19 @@ def _combine_row_counts(existing: pd.DataFrame, new: pd.DataFrame) -> pd.DataFra
     )
 
 
-def _write_row_counts(row_counts: pd.DataFrame, target: str = "etl-full"):
-    csv_path = _get_row_count_csv_path(target)
+def _write_row_counts(row_counts: pd.DataFrame):
+    csv_path = _get_row_count_csv_path()
```
zaneselvans, line 320: One failure mode I could imagine is someone accidentally updating row counts based on the fast ETL outputs. Is there any easy way to avoid that?

dbt/tests/data_tests/generic_tests/check_row_counts_per_partition.sql
krivard: I definitely prefer dumping this test into SKIPPED (which I'm pretty sure is what `enabled=false` will do) instead of a spurious PASS.

We could set enabled in the test definition instead:

https://docs.getdbt.com/reference/resource-configs/enabled

à la

```
{% test <testname>() %}

{{ config(
  enabled=(target.name == 'etl-full')
) }}

select ...

{% endtest %}
```

I don't think we'd lose much visibility-wise by having it in the test definition instead of schema.yml. Most of the user groups I can think of who would be reading schema.yml are equally capable and likely to read the test definition for a custom test like this, if they're truly chasing some funny behavior.

zaneselvans: Oh, I agree that's a nice combination of low-touch and conceptually consistent.  Just tested locally by adding it to the test definition (not the macro) and it works great.

test/integration/dbt_test.py
```diff
@@ -81,17 +76,16 @@ def test_update_tables(
...
     available.
     """
+    args = [
+        "dbt_helper",
+        "update-tables",
+        # "--schema",  # Uncomment when we have schema-preserving updates
+        "all",
+    ]
+    if dbt_target == "etl-full":
+        args.append("--row-counts")
```
zaneselvans, line 86: Only run with the `--row-counts` flag if we're using the `full-etl` outputs -- otherwise we'll get a bunch of spurious row count updates in CI.

test/integration/dbt_test.py
```diff
@@ -37,30 +37,25 @@ def dbt_target(test_dir: Path, request) -> str:
...
+    Note that the row count checks will automatically be disabled unless dbt_target is
+    'etl-full'. See the ``check_row_counts_per_partition.sql` generic test.
     """
-    # 2025-06-24 skip rowcount tests for fast ETL, since that seems to behave
-    # differently in CI vs. locally.
-    #
-    # see https://github.com/catalyst-cooperative/pudl/issues/4275
-    node_exclusion = None
-    if dbt_target == "etl-fast":
-        node_exclusion = "*check_row_counts_per_partition*"
```
zaneselvans, line 59: Row count checks are now disabled within the test itself, based on the target.

src/pudl/scripts/dbt_helper.py
```diff
@@ -616,8 +595,8 @@ def validate(
 
     build_params = {
         "node_selection": node_selection,
-        "dbt_target": target,
         "node_exclusion": exclude,
+        "dbt_target": "etl-full",
```
zaneselvans, line 599: This is a required argument -- the script crashes without it, and dbt does need to know what target it is running against. I think we want it hard coded to be `etl-full` -- but if there's ever anything else that varies by target other than the row counts, we'll need to remember this.

docs/dev/data_validation.rst
```diff
@@ -669,12 +669,10 @@ run the command:
...
 .. code-block:: bash
 
-    dbt_helper update-tables --target etl-full --row-counts --clobber {table_name}
+    dbt_helper update-tables --row-counts --clobber {table_name}
 
 This will tell the helper script to overwrite the existing row counts with new row
-counts from the table in your local ``PUDL_OUTPUT`` stash. If you want to update the
-``etl-fast`` row counts, use ``--target etl-fast`` instead of the default ``--target
-etl-full``.
+counts from the table in your local ``PUDL_OUTPUT`` stash.
```
zaneselvans, line 675: `--target` flag no longer exists, so remove it.

dbt/tests/data_tests/generic_tests/check_row_counts_per_partition.sql
```diff
@@ -1,5 +1,9 @@
 {% test check_row_counts_per_partition(model, table_name, partition_column) %}
 
+{{ config(
+  enabled=(target.name == 'etl-full')
+) }}
+
```
zaneselvans, line 6: Disable the test rather than passing if we aren't on `etl-full`.

dbt/tests/data_tests/generic_tests/check_row_counts_per_partition.sql
zaneselvans: I've made this change and updated some other things to align with it. I think the PR might be okay as it is now?

zaneselvans (approved):
@krivard I made edits to switch to setting `enabled=true` based on the target like you suggested. I played with running it locally with both full and fast targets in dbt directly and it seemed to do the right thing.

There were also a few lingering references to `--target` and I needed to hard-code `dbt_target="etl-full"` in a function call in the `dbt_helper` script where it is a required argument (the script was crashing otherwise).

Do you want to 👀 again? Should we merge?

zaneselvans:
okay YOLO

PR #4497, opened 2025-07-29
Fix ferc714 misalign

### LINKED ISSUE #4439: Misalignment of FERC 714 IDs in '...respondents...' tables
### Describe the bug
The respondent_id_ferc714 column does not seem consistent between the hourly planning area demand data and the respondent ID crosswalk tables.

### Bug Severity
High

### To Reproduce
Let's say I want to pull California ISO hourly load data from "out_ferc714__hourly_planning_area_demand" with the knowledge of the EIA BA Code of CISO. The data comes indexed by respondent_id_ferc714, which seems to be derived from the old csv respondent_id and newer xbrl entity_id. However, when trying to merge the hourly data with the respondent_id_ferc714 column in "out_ferc714__respondents_with_fips," the only respondent ID associated with the CISO EIA BA Code is 125, with the respondent name of Public Service Company of New Mexico.

Image

### Expected behavior
I would expect the respondent_id_ferc714 in "out_ferc714__respondents_with_fips" to align with ID 24, which in "core_ferc714__respondent_id" is assigned to "California Independent System Operator Corporation". It seems like the old _csv ID for CISO was 125, which may be where the error is coming from. 

Image

### Software Environment?
I've been remote querying the data from the AWS location using pd.read_parquet. All of these issues have been found without adjusting or filtering the tables at all.

### Additional context
As a bonus potential issue with the respondent tables, I have a question regarding identification of Pacificorp West and East. In "core_ferc714__respondent_id," there are 3 respondent IDs associated with the name Pacificorp. I had assumed the ID, name combination was unique. In this case, the duplicated name for 3 IDs makes it confusing if and where the Pacificorp split is found in the FERC 714 data.

Image

krivard:
# Overview

Closes #4439 .

A bad EIA code in one of the FERC filings was being mis-repaired on the wrong row, creating a spurious mapping between a respondent in New Mexico and the California ISO.

## What did you change?

Switched a combined EIA_CODE_FIXES entry from using the csv ID to using the PUDL-assigned FERC714 id, so that it repairs the correct row when building `core_ferc714__respondent_id`.

Checked all the other entries in EIA_CODE_FIXES to make sure they weren't also pointing to the wrong ID, and they were not.

Surprised this wasn't a bad join! but it turns out we're not joining on respondent id anywhere in the vicinity of when the bad mapping is introduced.

## Documentation

# Testing

Materialized all downstream assets from `core_ferc714__respondent_id` and double checked that `out_ferc714__respondents_with_fips` no longer places respondent 125 ("Public Service Company of New Mexico") with the CISO balancing authority (in California).

Eyeball-checked the remaining entries in EIA_CODE_FIXES when selected by ferc id, and they seem to match their comments, with the exception of three:

```
        295: 40229,  # Old Dominion Electric Cooperative (missing)
        312: 59435,  # NaturEner Glacier Wind (missing)
        329: 39347,  # East Texas Electricity Cooperative (missing)
```

These unfortunates list an EIA id that does not or no longer exists.

I attempted to disable them in EIA_CODE_FIXES, but that failed PK checks. With them in there, we get NAs in the balancing authority columns, but so long as that's not hurting anything, that seems the best way forward.

## To-do list


src/pudl/transform/ferc714.py
```diff
@@ -212,14 +212,16 @@ class and any generic 714 transforms are defined as internal module functions.
...
         303: 14725,  # PJM Interconnection Illinois Hub (missing)
         304: 14725,  # PJM Interconnection Northern Illinois Hub (missing)
         305: 14725,  # PJM Interconnection Dominion Hub (missing)
         306: 14725,  # PJM Interconnection AEP-Dayton Hub (missing)
         309: 12427,  # Michigan Power Pool / Power Coordination Center (missing)
+        # [2025-07-29 kmm] the following EIA IDs don't / no longer exist :(
+        # but if I disable them here, the PK checks fail, so here they remain
+        295: 40229,  # Old Dominion Electric Cooperative (missing)
         312: 59435,  # NaturEner Glacier Wind (missing)
         329: 39347,  # East Texas Electricity Cooperative (missing)
```
zaneselvans, line 226: There's ambiguity in the FERC-714 reporting as to whether these IDs refer to Utility or Balancing Authority IDs, and not every ID can be used to identify both types of entities. Usually the same ID refers to the same entity in both sets, but sometimes they refer to different entities in the two sets, and sometimes they exist in only one set or the other.  All of these IDs show up in `core_eia__entity_utilities`, though 59435 has a different name in there ("Innovative Solar 37, LLC" -- but they do change names sometimes).

So I suspect that these IDs aren't missing in a different / worse way than the others in this list, but maybe I'm not understanding what you found?

zaneselvans (approved):
Thanks for doing this spelunking. Just one non-blocking clarification about the "missing" IDs.

PR #4499, opened 2025-07-29
docs: adding new data validation test types

jdangerx:
# Overview

See #4126.

We didn't have good docs on "what if I need to add a new type of test that isn't already provided for me?"

## What did you change?

* turn this section into a [how-to guide](https://diataxis.fr/start-here/#how-to-guides) instead of a mix of that, [explanatory guide](https://diataxis.fr/start-here/#explanation), and [reference guide](https://diataxis.fr/start-here/#reference)
* add content

## Documentation

# Testing

1. `make docs-build`
2. running `python -m http.server` from `docs/_build/html`
3. click around in `localhost:8000`

## To-do list

### DIFF
--- a/docs/dev/data_validation.rst
+++ b/docs/dev/data_validation.rst
@@ -22,8 +22,7 @@ define our own `custom data tests
    with `DuckDB <https://duckdb.org/>`__ to query the PUDL `Parquet
    <https://parquet.apache.org/>`__ outputs we're able to run thousands of validations
    across hundreds of tables with billions of rows in a minute, instead of the 2-3 hours
-   it used to take our much less extensive validation tests to run. Plus we get to learn
-   SQL.
+   it used to take our much less extensive validation tests to run.
 
 --------------------------------------------------------------------------------
 Data validation guidelines
@@ -442,7 +441,7 @@ Some of the tests in the example above like ``expect_columns_not_all_null`` or
 files with the same name under ``dbt/tests/data_tests/generic_tests/``
 
 Documentation for the tests that we define is in
-``dbt/testse/data_tests/generic_tests/schema.yml``
+``dbt/tests/data_tests/generic_tests/schema.yml``
 
 .. todo::
 
@@ -558,24 +557,56 @@ building on the above example would look like:
 Defining new data validation tests
 --------------------------------------------------------------------------------
 
-.. note::
 
-  For comprehensive coverage, see `the dbt documentation
-  <https://docs.getdbt.com/best-practices/writing-custom-generic-tests>`__
+Sometimes you will want to test a property that can't be expressed
+using the existing dbt tests like ``check_row_counts_per_partition`` (in
+``dbt/tests/data_tests/generic_tests``) or the tests in `dbt_expectations
+<https://hub.getdbt.com/metaplane/dbt_expectations/latest/>`__ or `dbt_utils
+<https://hub.getdbt.com/dbt-labs/dbt_utils/latest/>`__.
 
-In dbt a data test is a ``SELECT`` statement written in SQL that's designed to return
-no results when the test passes. When the test fails, the results should be helpful in
-diagnosing the reason for the failure. In simple tests that check some per-row criteria,
-the results might just be all the rows that didn't meet the specified criteria. For
-tests that check some property of groups, multiple columns, or the table as a whole, it
-can be helpful to construct a result set that provides a summary of what failed.
+In those cases you'll need to define a new *type* of data validation test using
+dbt!
 
-Generic dbt data tests are often parametrized, meaning they take arguments other than
-the table being tested. These arguments and other information that is available from dbt
-can be used along with Jinja templates to dynamically construct complex SQL statements,
-leading to more generalizable, reusable tests.
+Writing tests in dbt means they'll be located next to all the other data
+validation we're defining in the dbt schemas, which is nice. They also tend to
+be quite performant.
 
-If you're not already familiar with SQL, some useful resources:
+In a few rare cases you may need to write the check with access to all of
+the tools within Python. In those cases, you can use `Dagster's asset checks
+<https://docs.dagster.io/guides/test/asset-checks>`__, but in general we prefer
+using dbt tests.
+
+How do I write a new dbt test?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A dbt test is a templated SQL query that runs on your output data to look for
+problems in the data. The query should be designed to return no rows if there
+are no problems with the data. If the query returns any rows at all, then the
+test will fail.
+
+The test will need to live as a templated piece of SQL within
+``pudl/dbt/tests/data_tests/generic_tests``. dbt has `official docs
+<https://docs.getdbt.com/best-practices/writing-custom-generic-tests>`__ for
+doing this, but the core steps are:
+
+1. Check to see whether the test you need is already provided by `dbt-utils
+   <https://hub.getdbt.com/dbt-labs/dbt_utils/latest/>`__ or `dbt-expectations
+   <https://github.com/metaplane/dbt-expectations>`__.
+2. Make a file called ``pudl/dbt/tests/data_tests/generic_tests/your_test.sql``.
+3. Add ``{% test your_test(some_test_params...) %}`` to the top of the file
+   and ``{% endtest %}`` to the end. By default, if a test is defined at the
+   **table** level, it will receive the ``model`` parameter; if it's defined
+   at the **column** level, it will receive both ``model`` and ``column``
+   parameters; and you can add more custom parameters in the test signature
+   which will be read out of the schema YAML.
+4. Write a SQL ``SELECT`` statement that returns any data that would fail your
+   test, as well as useful debugging information. See our existing tests in
+   ``dbt/tests/data_tests/generic_tests`` to see some common patterns.
+   ``dbt/tests/data_tests/generic_tests/expect_consistent_years.sql`` may be
+   of particular use as a simple example that returns useful debugging context
+   along with the failing rows.
+
+If you're not already familiar with SQL, here are some useful resources:
 
 * `Interactive Mode SQL Tutorial <https://mode.com/sql-tutorial>`__
 * `Greg Wilson's Querynomicon <https://third-bit.com/sql/>`__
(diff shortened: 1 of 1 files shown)


docs/dev/data_validation.rst
```diff
@@ -22,8 +22,7 @@ define our own `custom data tests
    with `DuckDB <https://duckdb.org/>`__ to query the PUDL `Parquet
    <https://parquet.apache.org/>`__ outputs we're able to run thousands of validations
    across hundreds of tables with billions of rows in a minute, instead of the 2-3 hours
-   it used to take our much less extensive validation tests to run. Plus we get to learn
-   SQL.
+   it used to take our much less extensive validation tests to run.
```
jdangerx, line 25: While it's fun to learn SQL it's not really relevant to the user... I feel like a killjoy but I think we don't need that in the docs :cry:

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
-tests that check some property of groups, multiple columns, or the table as a whole, it
-can be helpful to construct a result set that provides a summary of what failed.
-
-Generic dbt data tests are often parametrized, meaning they take arguments other than
-the table being tested. These arguments and other information that is available from dbt
-can be used along with Jinja templates to dynamically construct complex SQL statements,
-leading to more generalizable, reusable tests.
-
-If you're not already familiar with SQL, some useful resources:
+Sometimes you will want to test a property that can't be expressed using the
```
jdangerx: I wanted to use [semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) but the linter fought me on line-length stuff, which was pretty annoying, so I went back to "editor's default hard-wrapping command"

docs/dev/data_validation.rst
zaneselvans: 😢

docs/dev/data_validation.rst
```diff
@@ -442,7 +441,7 @@ Some of the tests in the example above like ``expect_columns_not_all_null`` or
 files with the same name under ``dbt/tests/data_tests/generic_tests/``
 
 Documentation for the tests that we define is in
-``dbt/testse/data_tests/generic_tests/schema.yml``
+``dbt/tests/data_tests/generic_tests/schema.yml``
```
zaneselvans, line 444: Well at least it wasn't testes.

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
-can be helpful to construct a result set that provides a summary of what failed.
-
-Generic dbt data tests are often parametrized, meaning they take arguments other than
-the table being tested. These arguments and other information that is available from dbt
-can be used along with Jinja templates to dynamically construct complex SQL statements,
-leading to more generalizable, reusable tests.
-
-If you're not already familiar with SQL, some useful resources:
+Sometimes you will want to test a property that can't be expressed using the
+existing DBT tests like ``check_row_counts_per_partition`` or the tests in
```
zaneselvans: I know this is a little pedantic and peer-pressurey but dbt isn't capitalized by dbt Labs or by the developer community around the tool, and I think we should follow that convention. To me seeing it capitalized says "Oh, they aren't that familiar with the tool." in the same way using CalTech always marked someone as an outsider because everyone at the school used Caltech. Maybe I'm being neurotic.

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
+``dbt_expectations`` or ``dbt_utils``.
+
+In those cases, you can use `Dagster's asset checks
+<https://docs.dagster.io/guides/test/asset-checks>`__ for a pure Python
+approach, or you can define a new test within DBT.
```
zaneselvans: I don't think we want people to use asset checks for data validation unless the test can't be done well within dbt/SQL, so I would mention them as a side-note later on, or express that preference and the justification for it clearly here. Like if someone made a PR and they implemented an asset check that could be done easily in SQL I would ask for a change.

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
+be quite performant. But, they're a little less flexible than the asset checks,
+so sometimes you will just have to write the check in Python.
+
+How do I write a new DBT test?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A DBT test is a templated SQL query that runs on your output data to look for
+problems in the data. If the query returns no rows, then the test passes. If
+the query returns any rows at all, then there are problems with the data and the
+test fails.
```
zaneselvans: Edited to make it clearer that you have to design the test to behave this way -- it's not an intrinsic property of the data.

```suggestion
A dbt test is a templated SQL query that runs on your output data to look for
problems in the data. The query should be designed to return no rows if there
are no problems with the data. If the query returns any rows at all, then the
test will be considered failing.
```

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+Testing the Tests
+~~~~~~~~~~~~~~~~~
 
-.. todo::
+OK, now you have a new test, which *seems* to be working.
+How can we check to make sure it's doing what we want?
 
-  Flesh out this section :user:`jdangerx`
+DBT has robust macro testing tools, and tests are basically macros,
```
zaneselvans: Do we describe what a macro is anywhere above here? (tbh I don't feel like I totally know what a macro is)

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
   setup that may affect what functionality is available when writing new tests.
 
-Defining Macros
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+Testing the Tests
+~~~~~~~~~~~~~~~~~
 
-.. todo::
+OK, now you have a new test, which *seems* to be working.
+How can we check to make sure it's doing what we want?
```
zaneselvans, line 626: I think we should set some expectation about when we want the tests to be tested -- right now we're only testing a  small subset of the tests we have defined, and for a lot of them that seems like it's probably okay. What kinds of red flags suggest that it's worth the extra work to create a macro and test it?

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
-<https://docs.getdbt.com/docs/build/models>`__ that materializes the intermediate table
-you want to execute tests on.  Defining a new model means adding a SQL file to
-``dbt/models/{data_source}/{table_name}/`` containing a ``SELECT`` statement. The
-results of that ``SELECT`` will constitute the new table.
+  WITH test_row_counts AS (
+      SELECT * FROM (VALUES
+          ('test_table', 2022, 1),
+          ('test_table', 2023, 1),
+      ) AS t(table_name, partition, row_count)
+  ),
```
zaneselvans, line 670: Can we fence these blocks and indicate they are SQL to get highlighting? Or is that just a GitHub thing? Right now they're getting syntax highlighting, but it's not for SQL.

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
 
 * `Interactive Mode SQL Tutorial <https://mode.com/sql-tutorial>`__
 * `Greg Wilson's Querynomicon <https://third-bit.com/sql/>`__
 * `Interactive DuckDB SQL Tutorial <https://motherduckdb.github.io/sql-tutorial/>`__
 * `DuckDB SQL Introduction <https://duckdb.org/docs/stable/sql/introduction.html>`__
 * `SQL for Data Scientists <https://www.oreilly.com/library/view/sql-for-data/9781119669364/>`__ (book)
 
-Many LLMs are also good at converting detailed natural language descriptions of a query
-into SQL, but you'll still need to have your own understanding of SQL to ensure that
-it's really doing what you intended, and to keep the queries readable and concise.
```
zaneselvans, line 588: No ❤️ for the clankers. 🤖🗑️

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
+data to cover many different use cases of the macro.
+
+If the test is particularly weird and hard to get good debug
+info for, you can add custom debug handlers for your test type in
+:func:`pudl.dbt_wrapper.build_with_context`, which gives you access to the full
+power of Python.
+
+
+Creating intermediate tables for a test
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
zaneselvans, line 747: Having messed around with SQL a bit more now, I wonder if we shouldn't mention that in many cases it may be simpler / more appropriate to create an ephemeral table using the CTE pattern `WITH` / `AS`:

```sql
WITH temporary_table AS (
    SELECT stuff AND things AS stuff_and_things
    FROM source(pudl.really_real_table)
)
SELECT stuff_and_things FROM temporary_table
```

I kinda wonder if we should switch our existing intermediary tables over to this setup? Are there cases in which we really do need to create these intermediate tables?  I could see them being useful if they were consumed by more than one test, but I don't think that's the case right now is it? They seem to add significant complexity.

docs/dev/data_validation.rst
zaneselvans: I like the idea of semantic linefeeds, especially for prose that's stored in git, but I've always found long lines really hard to read. Do you have a good editor setup for soft-wrapping them and keeping the navigation up and down ergonomic?

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,192 @@ building on the above example would look like:
...
+   <https://github.com/metaplane/dbt-expectations>`__.
+2. Make a file called ``pudl/dbt/tests/data_tests/generic_tests/your_test.sql``.
+3. Add ``{% test your_test(some_test_params...) %}`` to the top of the file
+   and ``{% endtest %}`` to the end. By default, if a test is defined at the
+   **table** level, it will receive the ``model`` parameter; if it's defined
+   at the **column** level, it will receive both ``model`` and ``column``
+   parameters; and you can add more custom parameters in the test signature
+   which will be read out of the schema YAML.
+4. Write a SQL ``SELECT`` statement that returns any data that would fail your
+   test, as well as useful debugging information
```
zaneselvans: We have a lot more detail in the testing the tests section below. I think it would be clarifying to pull one of our existing generic test definitions in here as a concrete example. One that includes parameters other than `model` and `column` and provides useful debugging output in the failure. Maybe `expect_consistent_years.sql`?

More detail on what "useful debugging information" means would also be helpful. We've had some tests that just return a single boolean value, which isn't super helpful.

docs/dev/data_validation.rst
jdangerx: It's nice to conform to people's style guides! I just personally find this specific style guide very distracting but I'm sure I'll get used to it.

docs/dev/data_validation.rst
jdangerx: Re: example, I think we can probably just give folks a pointer to "hey, there are some examples here." I don't think we should be in the business of explaining how SQL statements work in this documentation, really... if it's hard for someone to read the examples then it's probably most instructive to go look at the various SQL resources we linked and/or ask somebody (or... some-bot-y) to explain.

Re: expanding on "useful debugging information": I sort of think that if you're reading "how do I write a whole new test type" you've probably already fought with other tests & had to debug them. Plus we (will) have a whole debugging section elsewhere.

docs/dev/data_validation.rst
jdangerx: I think it depends on what we're trying to do... for example in the `out_ferc1__yearly_steam_plants_sched402` tests we have a ton of `expect_quantile_constraints` tests defined on various calculated columns. I don't think we'd want to make a custom test that applies only to the `out_ferc1__yearly_steam_plants_sched402` table in this case... instead the current intermediate table approach seems pretty clean.

I guess we *could* make all tests be table-level and then we could do some fancy templating & CTE work to allow us to define column-level tests on arbitrary SQL expressions, but that seems like maybe too clever by half?

zaneselvans: Ah you're right we do have a gajillion different subsetting tests that are defined on that one in particular, so it does seem like it makes sense, rather than recalculating over and over.  But still I think it's important to explain when that approach is justified vs. when an ephemeral CTE is appropriate.

docs/dev/data_validation.rst
zaneselvans: I'm good with just pointing people at the examples we have down under `tests/data_tests/generic_tests` rather than inlining some SQL here.

On setting the test up to be easily debugged vs. how do I actually do the debugging, they feel like separate but complementary topics.  Like if you're debugging and all the test author has left you with is a single row in the result set that contains one column with the value "False" you have no choice but to go spelunking and pick apart the query that defines the test to get some more legible intermediary outputs, or play with the table in Python. Whereas if the test is designed to give a useful summary of what caused the test failure, you may be able to address the problem without needing to do that extra work, and the test author is going to have the context required to compile that summary while they're developing the test, whereas the test failure debugger may be a different person months later and lack that context, and have to spend time reconstructing it.

E.g. `check_column_correlation` returns a single number, a correlation. But it doesn't say what the minimum required correlation was for the test to pass was, or what columns were being checked for correlation.

I think this is more of an issue in the tests that are checking some ensemble property of the whole table or groups of records and multiple columns, rather than just row-by-row tests where getting the the rows that fail the test returned to you is informative.

Maybe this is soluble by just pointing folks at another example -- some test that does provide legible summary information, to highlight the fact that it's possible and ergonomic, and you don't have to just return the raw "result" of your check. Maybe `expect_consistent_years`?

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,209 @@ building on the above example would look like:
...
-
-Generic dbt data tests are often parametrized, meaning they take arguments other than
-the table being tested. These arguments and other information that is available from dbt
-can be used along with Jinja templates to dynamically construct complex SQL statements,
-leading to more generalizable, reusable tests.
-
-If you're not already familiar with SQL, some useful resources:
+Sometimes you will want to test a property that can't be expressed using the
+existing dbt tests like ``check_row_counts_per_partition`` or the tests in
+``dbt_expectations`` or ``dbt_utils``.
```
cmgosnell: where are the existing tests? (aka how can you check to see if a test you want already exists?)

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,209 @@ building on the above example would look like:
...
+
+How do I write a new dbt test?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A dbt test is a templated SQL query that runs on your output data to look for
+problems in the data. The query should be designed to return no rows if there
+are no problems with the data. If the query returns any rows at all, then the
+test will fail.
+
+The test lives as a templated piece of SQL within
```
cmgosnell: grammar future test nit
```suggestion
The test will need to live as a templated piece of SQL within
```

docs/dev/data_validation.rst
cmgosnell: ah this is answered below.

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,209 @@ building on the above example would look like:
...
+1. Pull the test logic out into a macro
+2. Use the test as a *very* thin wrapper around the logic macro
+3. Test the logic macro
 
-Testing the Tests
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+First, we pull the test logic out into a macro (let's call it ``your_macro()``):
 
-.. todo::
+1. Copy the test file from above to ``pudl/dbt/macros/your_macro.sql``
```
cmgosnell: wait for real it needs to be totally copied?! what if you update the test file and not this macro file? EDIT: ah okay i see below... the content in the test is replaced by calling the marco okay okay. also it says it above in step 2 okay okay okay

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,209 @@ building on the above example would look like:
...
+  expected_mismatch_counts as (
+      SELECT * FROM (VALUES
+          ('test_table', 0),
+      ) AS t(table_name, num_mismatches)
+  ),
+
+We're saying here that ``expected_mismatch_counts`` is 0 - there are *no*
+partitions where we expect there to be a mismatch. Next, we call the macro:
+
+.. code-block:: jinja
```
cmgosnell, line 701: this one supposed to be jinja instead of sql like all the others?

docs/dev/data_validation.rst
```diff
@@ -558,85 +557,209 @@ building on the above example would look like:
...
+      FROM ({{
+          row_counts_per_partition('test_table', 'test_table', 'report_year', force_row_counts_table='test_row_counts')
+      }})) as observed_mismatch_count,
+      num_mismatches AS expected_mismatch_count,
+      FROM expected_mismatch_counts
+  )
+
+This one is a bit more complicated.
+
+Let's start from the macro call ``{{ row_counts_per_partition(...) }}``. This
```
cmgosnell, line 714: this is prolly basic but this implies that any sql file within `dbt/tests` have access to all the `dbt/macros`?

cmgosnell (commented):
I learned a lot while reading through this! I am a good person to test this on imo bc i am a complete noob with this system and syntax. I definitely got a little lost in the explanation of the testing the tests section... but I feel like I would have a much better starting place than before.

As a noob to both dbt and sql it would be helpful for me to understand how to best pick apart and debug these multi-stage processes. that one looked real simple all in all - I'm just less familiar with what each component should return. I appreciated the step by step break down and I liked the pattern of "say the high level thing up top and then explain it in more depth later. I also liked the pointers to existing files that are good examples.

docs/dev/data_validation.rst
jdangerx: Yeah, otherwise the docs build blows up because the `{{}}` isn't SQL syntax.

docs/dev/data_validation.rst
jdangerx: Yes!

cmgosnell (approved):
🎉

PR #4500, opened 2025-07-30
Extract PHMSA Gas Distribution 1970-1989

### LINKED ISSUE #3290: Extract PHMSA distribution and transmission data from 1970-1989
From 1970 - 1989, PHMSA distribution and transmission data is reported in one `.xls` spreadsheet per dataset, with one tab containing multiple years of data. This data was not included in the original round of raw data extraction because of data particularities. These will need to be addressed in order to extract this data:

- Multi-row column headers with repeated names for the final row of columns. This will make column mapping substantially more work, even though the columns available are roughly similar to 1990 data, with slightly fewer fields. 
- Tabs correspond to years rather than pages, with no breakdown between table parts. Each tab has a different version of the form. This will probably require some level of adaptation of our existing extraction infrastructure, which expects one sheet per year with tabs sorted by form section.
- We'll probably want to make a second Extraction class, which grabs the correct spreadsheet by year, takes a multiindex in header on read-in and concatenates several rows of information to make a single definitive column header, and then uses that column header in the rename. We'll have to figure out how to map one year (1970) to a tab containing multiple years, and how to avoid extracting the same spreadsheet 10 times if we are trying to grab data from 1970-1979.

Potentially, we will want to extract this data into separate `raw_phmsagas__transmission_1970_1979`, `raw_phmsagas__transmission_1980_1981` etc. tables and then split and concatenate them to the other tables during processing.

cmgosnell:
# Overview

Closes #3290.

old years not extracted

## What did you change?
- [x]  map four tabs of old years in csv's
- [x] fix weird duplicate column names for 1984
- [x] make sure the multi-year tabs get their `report_year` or date column properly cleaned
- [x] manage the partitions/settings files

## Documentation

# Testing

## To-do list


src/pudl/extract/phmsagas.py
```diff
@@ -26,6 +28,68 @@ def __init__(self, *args, **kwargs):
...
+
+        The first iteration of the duplicate og column name gets renamed to
+        the second name... in the df all of the main_* comes first, then
+        the services_* comes afterward. Because of this we have 20 instances
+        of duplicate pudl-renamed columns. So we need to rename the first ones
+        (the main_* ones). We need to find the locations of all the should be
+        main_* columns in the df & all the map of the services_* cols and the
+        main_* column names from the mapping sheet when they have the same mapped
+        og name.
+        """
```
cmgosnell: this is a kinda ridiculous solution to a problem that i didn't think could happen but did (read_excel naming multiple columns with the same name). i could certain just hand map these 20 columns but i'd be nervous about renaming them incorrectly that way. alt option would be to institute the solution @e-belfer  came up with originally which would be to build an extractor functionality that would read in multi-indexes... i don't actually think that would work because none of the multi-index headers are using merged cells so i don't think it would read it properly but that is a total unconfirmed guess.

src/pudl/extract/phmsagas.py
```diff
@@ -26,6 +28,68 @@ def __init__(self, *args, **kwargs):
         self.cols_added = []
         super().__init__(*args, **kwargs)
 
+    def remove_duplicate_column_name(self, df: pd.DataFrame, page: str) -> pd.DataFrame:
+        """Rename the duplicate column names in 1984.
+
+        On ingest of the should be a multi-header excel sheet, 20 of the columns
```
e-belfer: I am not able to reproduce this problem when I'm reading in the '84 spreadsheet using `pd.read_excel()` on the latest spreadsheet. Is it possible that something in our renaming is causing this problem?

e-belfer:
@cmgosnell Are you planning on handling the partitions in the extraction at all (e.g., if I try to extract just 1983, extracting the 1982 partition)?

src/pudl/extract/extractor.py
```diff
@@ -407,6 +407,8 @@ def partitions_from_settings(context) -> DynamicOutput:
         )
         partition = partition[0]
         parts = getattr(data_settings, partition)  # Get the actual values
+        if name == "phmsagas":  # phmsa has old years with multiple years in each tab
+            parts = data_settings.extraction_years
```
cmgosnell, line 411: what the comment says... ella and i debated how to handle this many-year-in-one-tab difference...

alt: could have read the full tab for every year then filtered by year... unnecessary resource use.
alt: could have fed the settings yml files/classes only the first years of the tabs.. misleading/possibly silently wrong for future selves

option i went w/ here: enable the settings files/classes to know the full range of years:
* why? bc that's real, good for future data source page, good for future use of settings in core/out
* how? added an extra property of phmsagas's settings class that is used here and only here.

src/pudl/extract/phmsagas.py
cmgosnell: eb realized this was caused by trailing spaces which were immediately getting removed in `pudl.helpers.simplify_columns`. bc of this realization I developed a new method which is a version of load_source

src/pudl/extract/extractor.py
```diff
@@ -199,7 +199,7 @@ def validate(self, df: pd.DataFrame, page: str, **partition: PartitionSelection)
             missing_raw_cols = set(expected_cols).difference(df.columns)
             if extra_raw_cols:
                 raise ValueError(
-                    f"{page}/{partition_selection}: Extra columns found in extracted table:"
+                    f"{page}/{partition_selection}: Columns found in raw extracted table that are unmapped:"
```
aesharpe, line 202: This is much clearer :)

src/pudl/extract/extractor.py
aesharpe: This feels like a reasonable work-around!

src/pudl/extract/phmsagas.py
```diff
@@ -26,16 +26,54 @@ def __init__(self, *args, **kwargs):
...
+        Old-ish years (1990-2009) of PHMSA data have one Excel tab in the raw data, while
+        newer data has multiple tabs. To extract data into tables that follow the newer
+        data format without duplicating the older data, we need to split older pages into
+        multiple tables by column. To prevent each table from containing all columns from
+        these older years, filter by the list of columns specified for the page, with a
         warning.
+
+        The oldest years (before 1990) contain multiple years in one tab. The records
+        contain a report_year column but some of them are reported at a two digit year
+        (ex: 87 for 1987). We convert these into four digit years.
```
aesharpe, line 67: This is a bit confusing to me. I took a look at the newer years (`phmsagas_gas_distribution_2010_present`) and it looks like each of the files only has one tab? The only files that appear to have more than one tab are in the `annual_gas_distribution_1970_1989.xls` file.

It's also unclear why you would want to split up conglomerated data (this is what is sounds like you're doing with the old-ish data to get it into the new format) when we are just going to smoosh it back together again. Is this because of how we process PHMSA data overall?

I know this is just a re-write of an old description.

The last bit about containing multiple years makes sense.

aesharpe (approved):
This looks good to me! Just added some questions about documentation.

src/pudl/extract/phmsagas.py
cmgosnell: yea imma be honest it took me a minute to understand this chunk but yea as you note the only thing i changed here was the years and line breaks. but if i understand correctly this was done because the newer data is broken up more granularly than this 1990-2009 so for each of the newest pages we pull a tab with a gazillion columns and break it out into newer format by selecting down the columns.

It's very possible that we would do this some different way or not do this at all if we decided to fully integrate phmsa.

PR #4508, opened 2025-08-01
Hotfix update row counts

zaneselvans:
# Overview

- I forgot to run `dbt_helper update-tables --clobber all` before merging #4435 yesterday, and there were some derelict row counts that needed to be removed. This would have made the nightly builds fail if they hadn't encountered a badzipfile.
- To make it clearer what was actually changing, I wanted to update the logging output to differentiate between successful updates where nothing is updated, and where row counts are actually changed.
- There's also one table that for reasons we don't understand gets different row counts (by 1) in the nightly builds vs. running locally. We had it set to "Warn" but I think this discrepancy means that if you update the row counts locally, you'll create an error in the nightly builds, since the actual and expected row counts there will not match, and the row counts file will get updated. So I removed the row count test. We should put a "row counts are close to this number" test there instead (it's a built-in -- we can apply it to the variable-length rate base table too)
- While attempting to output different logs for no-change vs. successful row-count updates, I found that we had differing NA values between the old and new row counts (np.nan vs. None) which meant some fiddling with the dataframes to make them comparable. At first I thought this meant we were also losing the row counts associated with a null partition that's mixed in with partitions that have values, but in the end I think this was just a transient issue with the dataframes I was using to check for changes vs. no changes.

# Testing

* I ran `dbt_helper update-tables --clobber all` locally and the only tables changed were the one that should have no row counts, and the one with different row counts in the nightly builds vs. local ETL.
* I ran `dbt build --select "test_name:*row_counts*" with the new row counts (and my fresh recent ETL outputs) and all passed.
* However, this change has broken the `dbt_helper` unit tests, and I'm not sure how to update the test cases.
* I'm also not sure how to add a test case that exercises the case of mixed null / non-null partition values.

## To-do list

- [ ] Fix the unit tests.
- [ ] Add a new unit test for the mixed null / non-null partition values.
- [ ] Add a unit test for the "successful no-op" behavior.


dbt/seeds/etl_full_row_counts.csv
```diff
@@ -3143,37 +3142,6 @@ out_ferc1__yearly_purchased_power_and_exchanges_sched326,2021,6899
...
-out_ferc1__yearly_rate_base,2015,13145
-out_ferc1__yearly_rate_base,2016,13212
-out_ferc1__yearly_rate_base,2017,13311
-out_ferc1__yearly_rate_base,2018,13367
-out_ferc1__yearly_rate_base,2019,13597
-out_ferc1__yearly_rate_base,2020,13703
-out_ferc1__yearly_rate_base,2021,15219
-out_ferc1__yearly_rate_base,2022,15077
-out_ferc1__yearly_rate_base,2023,15385
-out_ferc1__yearly_rate_base,2024,15196
```
zaneselvans, line 3176: Because these row counts vary a little bit randomly, we removed the test from the table. But forgot to remove the actual row counts.

We should add a test that the overall row counts are approximately what we expect them to be. This test is available from dbt_expectations I think.

dbt/models/sec10k/core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities/schema.yml
```diff
@@ -5,10 +5,6 @@ sources:
       - name: core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities
         data_tests:
           - expect_columns_not_all_null
-          - check_row_counts_per_partition:
-              table_name: core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities
-              partition_expr: null
-              severity: warn
```
zaneselvans, line 11: The row counts for this table vary by +/- 1 between the local full ETL and the nightly builds. We don't know why. It was set to warn rather than fail, which avoids dbt test failures resulting from this discrepancy, but updating all row counts locally vs. in the nightly builds will result in changes to the row count file, which will result in changes in the checked out repo in the nightly builds that will block updates to the `nightly` or other tags.

src/pudl/scripts/dbt_helper.py
```diff
@@ -331,8 +331,8 @@ def update_row_counts(
 
     if not has_test and not has_existing_row_counts:
         return UpdateResult(
-            success=False,
-            message=f"No row count test defined for {table_name}, nothing to update.",
+            success=True,
+            message=f"No row count test defined for {table_name}, and no row counts found, nothing to update.",
```
zaneselvans, line 343: The case of "there are neither row counts nor tests" seems like a success to me -- the row counts and the table data tests are consistent with each other.  This happens when we run use `all` rather than enumerating specific tables.

src/pudl/scripts/dbt_helper.py
```diff
@@ -365,13 +366,37 @@ def update_row_counts(
...
+        .sort_values(row_count_idx)
+        .reset_index(drop=True)
+    )
+
+    try:
+        pd.testing.assert_frame_equal(std_old, std_new, check_like=True)
+        return UpdateResult(
+            success=True,
+            message=f"Row counts for {table_name} are unchanged.",
+        )
```
zaneselvans: * Differentiating between no-op success and actual update success is helpful for identifying where changes are happening for debugging purposes.
* Is there any potential failure mode here by not attempting to update the file when we see there is no change?  Previously we would try to update it no matter what.

src/pudl/scripts/dbt_helper.py
```diff
@@ -365,13 +366,37 @@ def update_row_counts(
...
+        .sort_values(row_count_idx)
+        .reset_index(drop=True)
     )
+    std_new = (
+        new.astype("string")
+        .assign(partition=lambda df: df["partition"].replace("None", pd.NA))
+        .fillna(value="")
+        .sort_values(row_count_idx)
+        .reset_index(drop=True)
+    )
```
zaneselvans: The old and new row count dataframes comapre as different even when the contents are really the same, because of different NA values (`np.nan` vs. `None`) and the resulting different data types. I standardized them to do the comparison below, but maybe this expectation of comparability should be integrated further upstream?

marianneke: Just to clarify: is this due to the case of all values of the partition column being null in the DB and is the value the string `"None"`?

I think we should move this upstream for sure: probably in the SQL query we can coalesce the value of the partition expression with `''` so that null is always replaced with the empty string, which is what we want because we expect all values in the output df to be strings.

zaneselvans: Not all values of the partition column, just some of them. Like, where the partition column is nullable, and mostly has real values, but also has a few missing values, and we want to keep track of the number of records associated with missing values in the partition column. The table where this is showing up right now is `out_pudl__yearly_assn_eia_ferc1_plant_parts` which has 788 rows with no value for the year. And what value you end up getting in the `partition` column for that row where there's a NULL value in the partition column depends on exactly how you read it in. E.g. here we get `np.nan`, and the `partition` column has been cast by Pandas to be a `float` because the default integery type is non-nullable:

```python
from pudl.dbt_wrapper import DBT_DIR
row_counts = pd.read_csv(DBT_DIR / "seeds/etl_full_row_counts.csv")
row_counts.loc[row_counts["table_name"] == "out_pudl__yearly_assn_eia_ferc1_plant_parts"]
```
|      | table_name                                  |   partition |   row_count |
|-----:|:--------------------------------------------|------------:|------------:|
| 3487 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        1994 |        2439 |
| 3488 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        1995 |        2451 |
| 3489 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        1996 |        2417 |
| 3490 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        1997 |        2294 |
| 3491 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        1998 |        2281 |
| 3492 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        1999 |        2151 |
| 3493 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2000 |        2096 |
| 3494 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2001 |        1928 |
| 3495 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2002 |        1845 |
| 3496 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2003 |        1732 |
| 3497 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2004 |        1686 |
| 3498 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2005 |        1789 |
| 3499 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2006 |        1722 |
| 3500 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2007 |        1735 |
| 3501 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2008 |        1727 |
| 3502 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2009 |        1661 |
| 3503 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2010 |        1689 |
| 3504 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2011 |        1741 |
| 3505 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2012 |        1786 |
| 3506 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2013 |        1751 |
| 3507 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2014 |        1774 |
| 3508 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2015 |        1786 |
| 3509 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2016 |        1805 |
| 3510 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2017 |        1816 |
| 3511 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2018 |        1819 |
| 3512 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2019 |        1819 |
| 3513 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2020 |        1890 |
| 3514 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2021 |        1968 |
| 3515 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2022 |        2024 |
| 3516 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2023 |        2114 |
| 3517 | out_pudl__yearly_assn_eia_ferc1_plant_parts |        2024 |        2189 |
| 3518 | out_pudl__yearly_assn_eia_ferc1_plant_parts |         nan |         788 |

src/pudl/scripts/dbt_helper.py
marianneke: I do like the separate case of no-op change, but I prefer adding an early escape if block to this try except approach. for example, adding a helper function

```
def dataframes_equal(df1, df2):
    return (
        df1.sort_values(by=list(df1.columns)).reset_index(drop=True)
        .equals(
            df2.sort_values(by=list(df2.columns)).reset_index(drop=True)
        )
    )
```
and replacing the try block with
```
if dataframes_equal(old, new):
    return ...
```

(the situation of having "None" in the `new` dataframe should be resolved in the SQL query in compute new rows)

src/pudl/scripts/dbt_helper.py
marianneke: OK, makes sense. This code standardizing the old and new dfs can be removed in favor of simply changing the SQL query in `_calculate_row_counts`:

replace
```
CAST({partition_expr_sql} AS VARCHAR) AS partition,
```
with 
```
CAST(COALESCE({partition_expr_sql}, '') AS VARCHAR) AS partition,
```
and the missing values should be replaced with an empty string.

src/pudl/scripts/dbt_helper.py
marianneke: Agreed

marianneke (dismissed):
I have a couple of changes to propose:

1. I prefer keeping to the existing pattern of having early return blocks in the update function if some condition is met, rather than having the write action inside an exception block. it just superficially reads as confusing when there is an except block for something that is totally fine to happen (like having records to write)
2. The case of null values in the data should indeed be prevented upstream, namely in the SQL query in the function that calculates the new counts. That is much easier to fix than in some comparison code inside a function that should specialize in deciding when to update things.
3. Also, I want to keep the logic inside the update function simpler, and if we want to compare two dataframes to see if the data inside them is the same (irrespective of index or order), that should be a helper function.

src/pudl/scripts/dbt_helper.py
```diff
@@ -281,13 +284,18 @@ def _calculate_row_counts(
 
     row_count_query = f"""
 SELECT
-    CAST({partition_expr_sql} AS VARCHAR) AS partition,
+    CAST(COALESCE(CAST({partition_expr_sql} AS VARCHAR), '') AS VARCHAR) AS partition,
```
zaneselvans, line 287: Unfortunately this change along wasn't enough to make the old and new dataframes comparable. I moved the additional munging up into the functions that read the old/new row counts though, rather than having it in the update function.

src/pudl/scripts/dbt_helper.py
```diff
@@ -263,7 +263,10 @@ def _get_row_count_csv_path() -> Path:
 
 
 def _get_existing_row_counts() -> pd.DataFrame:
-    return pd.read_csv(_get_row_count_csv_path(), dtype={"partition": str})
+    return pd.read_csv(
+        _get_row_count_csv_path(),
+        dtype={"partition": "string", "table_name": "string"},
+    ).fillna(value="")
```
zaneselvans, line 269: Needed to do the `fillna(value="")` for the old row counts separate from the SQL query that constructs the new ones. Switched to Pandas nullable strings, since we use them everywhere else and they're a real type, not `object`

src/pudl/scripts/dbt_helper.py
```diff
@@ -281,13 +284,18 @@ def _calculate_row_counts(
...
 
-    new_row_counts = duckdb.sql(row_count_query).df().astype({"partition": str})
-    new_row_counts["table_name"] = table_name
+    new_row_counts = (
+        duckdb.sql(row_count_query)
+        .df()
+        .assign(table_name=table_name)
+        .astype({"partition": "string", "table_name": "string"})
+        .loc[:, ["table_name", "partition", "row_count"]]
+    )
```
zaneselvans, line 298: Munge the results here to ensure that they are comparable to the old row counts.

src/pudl/scripts/dbt_helper.py
```diff
@@ -365,12 +374,26 @@ def update_row_counts(
 
     partition_expr = partition_expressions[0]  # TODO: support multiple partitions
     new = _calculate_row_counts(table_name, partition_expr)
+
+    # Make old and new row counts comparable so we can detect changes
+    row_count_idx = ["table_name", "partition"]
+    if (
+        old.sort_values(by=row_count_idx)
+        .reset_index(drop=True)
+        .equals(new.sort_values(by=row_count_idx).reset_index(drop=True))
```
zaneselvans, line 383: `df.equals()` has less flexibility in terms of what you ask it to check than `assert_frame_equal` so switching to this method means more dataframe munging upstream, which is why I used that testing method before. If we don't like this, I'm inclined to go back to `assert_frame_equal` with all its bells and whistles rather than creating a helper function that reimplements that flexibility.

dbt/models/ferc1/out_ferc1__yearly_rate_base/schema.yml
```diff
@@ -3,6 +3,13 @@ sources:
   - name: pudl
     tables:
       - name: out_ferc1__yearly_rate_base
+        data_tests:
+          - dbt_expectations.expect_table_row_count_to_be_between:
+              min_value: 433000
+              max_value: 434000
+          - expect_columns_not_all_null:
+              description: ferc_account is entirely null. Seems like an error.
+              severity: warn
```
zaneselvans, line 14: I think we noticed this before and put it on the todo list, but I wanted to make it stand out.

dbt/models/ferc1/out_ferc1__yearly_rate_base/schema.yml
```diff
@@ -3,6 +3,13 @@ sources:
   - name: pudl
     tables:
       - name: out_ferc1__yearly_rate_base
+        data_tests:
+          - dbt_expectations.expect_table_row_count_to_be_between:
+              min_value: 433000
+              max_value: 434000
```
zaneselvans: The row count for this table jumps around (we don't really understand why) but this at least should alert us if something big happens.  Not 100% sure these are the right bounds. I guess we'll see how much it changes and can adjust!

dbt/models/sec10k/core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities/schema.yml
```diff
@@ -5,10 +5,10 @@ sources:
...
+          - dbt_expectations.expect_table_row_count_to_be_between:
+              description: Nightly builds and local ETL yield slightly different output
+              min_value: 5545
+              max_value: 5546
```
zaneselvans: Have only seen a single row of variation here. Put some guardrails on it.

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -1829,7 +1829,6 @@ core_pudl__codes_imputation_reasons,,12
 core_pudl__codes_subdivisions,,69
 core_pudl__entity_plants_pudl,,20320
 core_pudl__entity_utilities_pudl,,17061
-core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities,,5545
```
zaneselvans, line 1832: Disabled row count checks for this table since local vs. nightly builds have different outputs. But added guardrails (see above).

test/unit/dbt_helper_test.py
```diff
@@ -1046,3 +1080,187 @@ def test_complex_schema_diff_output():
...
+            "partition_expr: year",
+            pd.read_csv(
+                StringIO(
+                    "year, fake_data\n"
+                    "2020, 1.23456\n"
+                    "2021, 2.34567\n"
+                    "2021, 4.56789\n"
+                    "2022, 3.14159\n"
+                )
+            ),
```
zaneselvans, line 1099: - I turned these into CSVs and switched to more intuitive names / values to help readability.
- Added some sigfigs to the values so ruff doesn't try and put everything on one line.

pudlbot (approved):
🤔 🤖 ✅

PR #4525, opened 2025-08-06
add DbtSchema method to preserve tests and metadata from old schema

### LINKED ISSUE #4466: dbt helper update-tables should not delete existing tests
### Overview

When running `dbt_helper update-tables --schema <> --update` we get output that deletes all the data tests, whether they're defined on the table or the column.

Here's an example where I add the `acid_gas_control` column to `out_ferc1__yearly_steam_plants_fuel_sched402` with `dbt_helper update-tables --schema out_ferc1__yearly_steam_plants_fuel_sched402 --update`:

```diff
--- old_schema
+++ new_schema
@@ -3,12 +3,8 @@
   - name: pudl
     tables:
       - name: out_ferc1__yearly_steam_plants_fuel_sched402
-        data_tests:
-          - expect_columns_not_all_null
-          - check_row_counts_per_partition:
-              table_name: out_ferc1__yearly_steam_plants_fuel_sched402
-              partition_column: report_year
         columns:
+          - name: acid_gas_control
           - name: report_year
           - name: utility_id_ferc1
           - name: utility_id_pudl
@@ -19,83 +15,10 @@
           - name: fuel_consumed_total_cost
           - name: fuel_consumed_units
           - name: fuel_cost_per_mmbtu
-            data_tests:
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='coal'
-                  constraints:
-                    - quantile: 0.05
-                      min_value: 0.75
-                    - quantile: 0.95
-                      max_value: 4.2
-                  weight_column: fuel_consumed_mmbtu
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='oil'
-                  constraints:
-                    - quantile: 0.05
-                      min_value: 5.0
-                    - quantile: 0.95
-                      max_value: 25.0
-                  weight_column: fuel_consumed_mmbtu
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='gas'
-                  constraints:
-                    - quantile: 0.05
-                      min_value: 1.8
-                    - quantile: 0.95
-                      max_value: 12.0
-                  weight_column: fuel_consumed_mmbtu
           - name: fuel_cost_per_unit_burned
-            data_tests:
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='coal'
-                  constraints:
-                    - quantile: 0.1
-                      min_value: 7.0
-                    - quantile: 0.95
-                      max_value: 100.0
-                  weight_column: fuel_consumed_units
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='oil'
-                  constraints:
-                    - quantile: 0.05
-                      min_value: 25.0
-                    - quantile: 0.95
-                      max_value: 140.0
-                  weight_column: fuel_consumed_units
-                  description: Demoted to warning April 2025 when quantile 0.05 dropped from 25.0 to 1.6. Investigate in issue 4189.
-                  severity: warn
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='gas'
-                  constraints:
-                    - quantile: 0.05
-                      min_value: 1.6
-                    - quantile: 0.95
-                      max_value: 12.0
-                  weight_column: fuel_consumed_units
-                  description: Quantile 0.05 may have dropped from 2.0 to 1.63 (etl-full) in April 2025 (had been marked xfail with tracking issue 2073).
           - name: fuel_cost_per_unit_delivered
           - name: fuel_mmbtu_per_unit
-            data_tests:
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='oil'
-                  constraints:
-                    - quantile: 0.05
-                      min_value: 5.5
-                    - quantile: 0.95
-                      max_value: 6.8
-                  weight_column: fuel_consumed_units
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='gas'
-                  constraints:
-                    - quantile: 0.05
-                      min_value: 0.98
-                    - quantile: 0.95
-                      max_value: 1.096
-                  weight_column: fuel_consumed_units
-                  description: Quantile 0.95 rose from 1.08 to 1.096 April 2025. Keep an eye on it, and investigate if it gets worse.
           - name: fuel_type_code_pudl
           - name: fuel_units
           - name: record_id
-            data_tests:
-              - dbt_expectations.expect_column_values_to_be_unique
```

### Success Criteria

* [ ] table-level data tests are never touched by `dbt_helper update-tables --update`
* [ ] if a column that contains a `data-test` would be deleted by `--update`, tell the user to manually update that column and then re-run `dbt_helper update-tables --schema --update`
* [ ] `--clobber` still clobbers everything

### Next steps
* [ ] take a look at the `diff` object generated by DeepDiff in `dbt_helper.update_table_schema` to see if that gives us the information we need to update or not-update
* [ ] potentially, refactor the `update_table_schema` function to be pure so it's easier to test
* [ ] write tests for the success criteria

marianneke:
# Overview

Closes #4466.

PUDL metadata does not include tests, descriptions, dbt tags and other info one can add to dbt schemas, so when we update dbt schemas based off the PUDL metadata columns, currently all extra information such as tests etc is lost. This makes the update schema function quite unusable, as this means most schema updates have to be done manually.

## What did you change?
This PR adds a method to DbtSchema class to port over metadata from one schema to another. Assuming you have an old schema in dbt (call it `old_schema`) and are updating it using new column information from PUDL metadata (`new_schema`), the behavior is as follows:
- The starting point is `new_schema`, meaning all tests, description, tags, columns etc have the values set in `new_schema`
- If a piece of table-level data in `new_schema`, such as row count test or table description, is not set, port it over from `old_schema`
- For each column in `new_schema`, if tests, tags or descriptions are not set, port them over from `old_schema`

## Documentation

# Testing

## To-do list

### DIFF
--- a/src/pudl/scripts/dbt_helper.py
+++ b/src/pudl/scripts/dbt_helper.py
@@ -201,6 +201,76 @@ class DbtSchema(BaseModel):
             yaml_output = _prettier_yaml_dumps(self.model_dump(exclude_none=True))
             schema_file.write(yaml_output)
 
+    def merge_metadata_from(self, old_schema: "DbtSchema") -> "DbtSchema":
+        """Merge metadata from an old schema into this one, preferring new values where present."""
+
+        def merge_tables(new_tables, old_tables):
+            old_table_map = {t.name: t for t in old_tables or []}
+            merged = []
+
+            for new_table in new_tables or []:
+                old_table = old_table_map.get(new_table.name)
+                if not old_table:
+                    merged.append(new_table)
+                    continue
+
+                old_columns = {c.name: c for c in old_table.columns or []}
+                merged_columns = [
+                    (
+                        col.model_copy(
+                            update={
+                                "description": col.description
+                                or old_columns.get(col.name, {}).description,
+                                "tests": col.tests
+                                or old_columns.get(col.name, {}).tests,
+                                "tags": col.tags or old_columns.get(col.name, {}).tags,
+                            }
+                        )
+                        if col.name in old_columns
+                        else col
+                    )
+                    for col in new_table.columns or []
+                ]
+
+                merged_table = new_table.model_copy(
+                    update={
+                        "description": new_table.description or old_table.description,
+                        "tests": new_table.tests or old_table.tests,
+                        "tags": new_table.tags or old_table.tags,
+                        "meta": new_table.meta or old_table.meta,
+                        "columns": merged_columns,
+                    }
+                )
+                merged.append(merged_table)
+
+            return merged
+
+        # Merge sources
+        old_sources = {s.name: s for s in old_schema.sources or []}
+        new_sources = []
+        for new_source in self.sources or []:
+            old_source = old_sources.get(new_source.name)
+            merged_tables = merge_tables(
+                new_source.tables, old_source.tables if old_source else []
+            )
+            merged_source = new_source.model_copy(update={"tables": merged_tables})
+            new_sources.append(merged_source)
+
+        # Merge models (if any)
+        if self.models:
+            old_models = {m.name: m for m in old_schema.models or []}
+            new_models = []
+            for new_model in self.models:
+                old_model = old_models.get(new_model.name)
+                merged_model = merge_tables(
+                    [new_model], [old_model] if old_model else []
+                )[0]
+                new_models.append(merged_model)
+        else:
+            new_models = None
+
+        return self.model_copy(update={"sources": new_sources, "models": new_models})
+
 
 def schema_has_removals_or_modifications(diff: DeepDiff) -> bool:
     """Check if the DeepDiff includes any removals or modifications."""
@@ -413,11 +483,16 @@ def update_table_schema(
             message=f"DBT configuration already exists for table {table_name} and clobber or update is not set.",
         )
 
-    new_schema = DbtSchema.from_table_name(table_name)
+    generated_schema = DbtSchema.from_table_name(table_name)
+    # Default: use generated schema
+    new_schema = generated_schema
 
     if model_path.exists() and update:
         # Load existing schema
         old_schema = DbtSchema.from_yaml(schema_path)
+        # Replace new schema with the generated schema but copying missing metadata
+        # from the old schema
+        new_schema = generated_schema.merge_metadata_from(old_schema)
 
         # Generate the diff report
         diff = DeepDiff(


src/pudl/scripts/dbt_helper.py
```diff
@@ -201,6 +201,76 @@ def to_yaml(self, schema_path: Path):
             yaml_output = _prettier_yaml_dumps(self.model_dump(exclude_none=True))
             schema_file.write(yaml_output)
 
+    def merge_metadata_from(self, old_schema: "DbtSchema") -> "DbtSchema":
+        """Merge metadata from an old schema into this one, preferring new values where present."""
+
+        def merge_tables(new_tables, old_tables):
+            old_table_map = {t.name: t for t in old_tables or []}
```
zaneselvans: Can you explain how this idiom of `or []` in a dictionary comprehension (and other contexts below) works in a comment? it's not something we've used before and doesn't seem intuitive.

src/pudl/scripts/dbt_helper.py
```diff
@@ -201,6 +201,76 @@ def to_yaml(self, schema_path: Path):
...
+        for new_source in self.sources or []:
+            old_source = old_sources.get(new_source.name)
+            merged_tables = merge_tables(
+                new_source.tables, old_source.tables if old_source else []
+            )
+            merged_source = new_source.model_copy(update={"tables": merged_tables})
+            new_sources.append(merged_source)
+
+        # Merge models (if any)
+        if self.models:
```
jdangerx: I suppose we can't *define* any models in `pudl.metadata.resources` so maybe we only need `self.model_copy(update={"sources": new_sources})` and can get rid of this whole block.

jdangerx (changes requested):
I went ahead and tested out the behavior...

### adding a new column

I added `"acid_gas_control"` to `out_eia__yearly_plants` definition in `pudl/metadata/resources/eia.py` and ran `dbt_helper update-tables --schema out_eia__yearly_plants`.

* :heavy_check_mark: with no other flags: this failed with "configuration already exists" - kind of annoying but as expected
* :heavy_check_mark: `--update`: added the column to `schema.yml`, did not delete tests!
* :heavy_check_mark: `--clobber`: added the column **and** deleted the table-level tests. Doesn't seem very useful but also as expected...

### deleting a column with no tests defined

I removed the `"has_net_metering"` field from `out_eia__yearly_plants`.

* :heavy_check_mark: no flag: failed with "configuration already exists"
* :x: `--update`: retains the tests in the diff, but because a column is being removed, refuses to actually write the changes out:
```
Proposed schema changes:
--- old_schema
+++ new_schema
@@ -77,7 +77,6 @@
           - name: natural_gas_pipeline_name_2
           - name: natural_gas_pipeline_name_3
           - name: nerc_region
-          - name: has_net_metering
           - name: pipeline_notes
           - name: primary_purpose_id_naics
           - name: regulatory_status_code
2025-08-12 16:56:02 [   ERROR] catalystcoop.pudl.scripts.dbt_helper:531 DBT configuration for table out_eia__yearly_plants has information that would be deleted. Update manually, or: 1) re-run with --clobber, 2) run `git add dbt/models -p` and follow the instructions for recovering the deleted info
```
* :heavy_check_mark: `--clobber`: blows everything away and re-initializes the schema, as above.

## remove a column that *has* tests defined on it

I removed `capacity_factor` from the `out_eia__{freq}_generators` resource defined in `pudl/metadata/resources/mcoe.py`

* :heavy_check_mark: no flag: fails with 'configuration already exists'
* :heavy_check_mark: `--update`: attempts to delete the column and its associated tests; then identifies that this would be destructive and doesn't actually do it (though with a slightly misleading error message):
```
Proposed schema changes:
--- old_schema
+++ new_schema
@@ -87,32 +87,6 @@
           - name: capacity_mw
           - name: fuel_type_code_pudl
           - name: planned_generator_retirement_date
-          - name: capacity_factor
-            data_tests:
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='gas' and report_date>=CAST('2015-01-01' AS DATE) and capacity_factor<>0.0
-                  constraints:
-                    - quantile: 0.65
-                      min_value: 0.4
-                      max_value: 0.7
-                    - quantile: 0.15
-                      min_value: 0.005
-                    - quantile: 0.95
-                      max_value: 0.95
-                  weight_column: capacity_mw
-                  description: Historical note, EIA natural gas reporting really only becomes usable in 2015.
-              - expect_quantile_constraints:
-                  row_condition: fuel_type_code_pudl='coal' and capacity_factor<>0.0
-                  constraints:
-                    - quantile: 0.6
-                      min_value: 0.444
-                      max_value: 0.9
-                    - quantile: 0.1
-                      min_value: 0.04
-                    - quantile: 0.95
-                      max_value: 0.95
-                  weight_column: capacity_mw
-                  description: Quantile 0.6 dropped from 0.5 to 0.444 in April 2025. Keep an eye on it, and investigate if it gets worse.
           - name: fuel_cost_per_mmbtu_source
           - name: fuel_cost_per_mmbtu
             data_tests:
2025-08-12 17:19:49 [   ERROR] catalystcoop.pudl.scripts.dbt_helper:531 DBT configuration for table out_eia__yearly_generators has information that would be deleted. Update manually, or: 1) re-run with --clobber, 2) run `git add dbt/models -p` and follow the instructions for recovering the deleted info
```
* :heavy_check_mark: `--clobber`: as expected, deletes table and column level tests



Overall:

* blocking: I would like to see tests that test these scenarios!
  * The existing test tests whether the merged metadata is correct but doesn't then test "what do we do about the merged metadata vs. the existing metadata" which is actually core logic that needs testing
  * I think following the pattern of `test_update_table_row_counts_update`, where we directly call the command using `CliRunner` and patch out any globals we're using, might be the clearest way to express these scenarios and test that "the command does what we expect"

* blocking: If `--update` should only refuse to write out the new schema if human-generated content would be deleted, we should update the error message.

* non-blocking: the original script had "initializing new tables" as its main use case, but in the future the main use case will be "updating tables" - does it make sense for `--update` to be the default behavior?

src/pudl/scripts/dbt_helper.py
```diff
@@ -201,6 +201,76 @@ def to_yaml(self, schema_path: Path):
...
+                    for col in new_table.columns or []
+                ]
+
+                merged_table = new_table.model_copy(
+                    update={
+                        "description": new_table.description or old_table.description,
+                        "data_tests": new_table.data_tests or old_table.data_tests,
+                        "tags": new_table.tags or old_table.tags,
+                        "meta": new_table.meta or old_table.meta,
+                        "columns": merged_columns,
```
zaneselvans: I'm a little nervous about the specific enumeration of all these properties. Do we know we've covered all the things we can ever encounter? If we were to manually add a new field that's compatible with dbt but not with our `DbtSchema` or `DbtModel` etc. Pydantic classes, then we'd get a validation error when reading in the existing schema and have to fix things, right? So it wouldn't happen silently and accidentally remove something.

src/pudl/scripts/dbt_helper.py
zaneselvans: I guess you could arrange to try and merge an old model with another old model -- reading both of them out of `schema.yml` files, rather than generating one from the PUDL metadata. But I don't think that's a scenario we're trying to support. So maybe we can move the `if self.models` up to the top and have it return an error early.

zaneselvans (commented):
I agree with Dazhong's suggestion that we get rid of the `--update` flag since this is already called `update-tables` and that's what we're going to be doing almost all the time when using this subcommand.

And also the request for an end-to-end test that mirrors the pattern of `test_update_table_row_counts_update` so we can compare against what actually ends up getting written.

src/pudl/scripts/dbt_helper.py
jdangerx: Fair - we *currently* don't ever call `merge_metadata_from` on any schemas that we read from YAML, but that doesn't mean we can't in the future.

I would personally make `merge_metadata_from` only take the sources as a parameter, instead of the whole schema, so that we can't accidentally pass in models at all. But checking at the top achieves a similar effect with less refactoring :)

src/pudl/scripts/dbt_helper.py
marianneke: If DBT updates their classes, then we would have to do the same in our classes if it otherwise breaks something, though these properties can only be properties that are defined in the classes as we have defined them.

zaneselvans: The scenario I'm imagining that would be nice to avoid is that we discover a new property that we want to use in dbt, and start adding that information to the `schema.yml` files manually, and then later attempt to do a schema update with the script, and it wipes out that manually added information silently. But I think what would happen is we'd get a validation failure when attempting to construct the `DbtSchema` object from the `schema.yml` input, since it would include additional fields which aren't defined in the Pydantic class. Does that seem right?

src/pudl/scripts/dbt_helper.py
marianneke: It's replacing any "falsey" value of old_tables (most likely case would be if the old schema has no tables defined, therefore might have old_tables = None) with an empty list, in order for the dict comprehension not to fail in that case, but to simply return an empty map, which is the expected behavior in case there is nothing to map.

src/pudl/scripts/dbt_helper.py
marianneke: Yes, if we add properties directly to the schema.yml and try to load a `DbtSchema.from_yaml` from that schema, it should throw a ValidationError. However, this only protects us from adding details to the schemas that are not implemented in the classes. We would still need to have a way to ensure any new info added to the classes are also added to the method that transfers them over. I will see if there is some elegant way to ensure all metadata is preserved, because the alternative is that we have to always add any new facts to both the class and to this `merge_metadata_from` method.

marianneke: I have now made the logic for merging schemas robust to changes in our definitions of classes in dbt helper. This means that if we want to introduce new types of schema fields, all we have to do is add them to the relevant DbtColumn, DbtTable, DbtSource or DbtSchema classes, and the merge metadata logic will still work as expected.

src/pudl/scripts/dbt_helper.py
marianneke: I have removed support for DBT models in the schema merging logic and added a ValueError in case someone tries to merge a schema with models defined.

marianneke:
Requesting re-review @zaneselvans @jdangerx 
- Removed support for merging schemas in case a DBT model is present (only source is supported). It returns ValueError if that happens
- Generalized schema merging logic to work with whatever the definition of the different DBT classes in dbt helper are, meaning if we add a new type of field, e.g. config, it merges that too
- Removed all logic related to `--update` flag, and made the old `--update` behavior the default behavior for schema updates (with row counts, update behaved exactly as clobber, so just removed it)
- Modified behavior of schema updates when simple column removals happen: previously these caused a warning and refusal to auto-update, now these do auto-update. Only when removing metadata such as descriptions or tests the auto-update is refused
- Added tests for schema updates, testing cases discussed in PR

docs/release_notes.rst
```diff
@@ -97,6 +97,12 @@ Quality of Life Improvements
   the logic for updating row counts now depends on whether a test has been defined in
   the dbt schema, whether any existing row counts for that table are present in the seed
   table, as well as user provided settings such as ``--clobber``.
+* When running ``dbt_helper update-tables`` without the ``clobber`` flag, existing
```
jdangerx: nit: this lives in a new release now, wasn't in 2025.8.0

jdangerx:
It would also be good to update the `docs/dev/data_validation_quickstart.rst` with the new usage instructions ~once #4552 gets merged.~ since #4552 was merged.

docs/release_notes.rst
zaneselvans: Moved it up to the top section.

zaneselvans:
Updated the data validation quickstart. Gonna merge if the integration tests pass.

PR #4528, opened 2025-08-07
Add ferc1 util xbrl dbf id to out

### LINKED ISSUE #4365: Merge FERC 1 DBF + XBRL utility IDs into output tables
# Overview

* As came up in #4363 the multiple layers of utility / respondent IDs that exist to connect DBF + XBRL Utility IDs to an all-time FERC Utility ID, and FERC Utility IDs to EIA Utility IDs via our PUDL IDs is pretty complex, and should really have some high-level user-facing documentation.
* It would be useful for users to have direct access to the DBF/XBRL Utility IDs in the FERC 1 output tables, because those IDs are stable long-term, and may appear in other datasets.

## Success Criteria

* [ ] All `out_ferc1__` tables that contain `utility_id_ferc1` also contain the associated `utility_id_ferc1_xbrl` and `utility_id_ferc1_dbf`.
* [ ] We have added a Utility ID association/crosswalk/mapping page to the Methodology section of our docs that explains to users at a high level how we connect old & new FERC Utility IDs, and FERC & EIA Utility IDs (via PUDL IDs) and how the different kinds of IDs should be used.

## Next Steps
* Find the function that we use to denormalize the FERC 1 output tables right now, and modify it to merge in the XBRL & DBF IDs.
* If that ID-merging functionality isn't currently generic, split it out into its own function so the logic can live in just one place.
* Refactor the other `out_ferc1__` assets to use this generic denormalization function.

aesharpe:
# Overview

Closes #4365
Lots of files changed because each dbt table schema is it's own file!

Adds `utility_id_ferc1_dbf` and `utility_id_ferc1_xbrl` columns to all ferc1 output tables.

## What did you change?

- Merge ferc1 output tables with `pudl.glue.ferc1_eia.get_utility_map_ferc1()` table at three different high-level places: 
   - ferc1_output_asset_factory (all the financial tables)
   - _out_ferc1__yearly_plants_utilities (gets merged with all the plant tables)
   - exploded_table_asset_factory (all the detailed tables)
- Update table metadata
   - In doing this I moved the `report_year` column to the top when it wasn't
- Update dbt table metadata
   - In doing this I moved the `report_year` column to the top when it wasn't
- Add new migration

## Documentation

# Testing

- Materialize `out_ferc1` (and `_out_ferc1`) assets in dagster and take a look at them in the `inspect_assests.ipynb`

## To-do list


docs/release_notes.rst
```diff
@@ -51,6 +51,9 @@ Quality of Life Improvements
 * Stopped running code checks in CI when only the documentation has changed.
   See issue :issue:`4410` and PR :pr:`4429`.
 
+* Added ``utility_id_ferc1_dbf`` and ``utility_id_ferc1_xbrl`` columns into all ferc1
+  output tables. See :issue:`4365` and PR :pr:`5428`.
```
zaneselvans: Should get rid of the blank lines separating these bullet points above too.

```suggestion
  output tables. See :issue:`4365` and PR :pr:`4528`.
```

migrations/versions/341e3c844e85_add_xbrl_dbf_util_id_cols_to_ferc1_out_.py
zaneselvans, line 1: Because the schema has also changed on `main` you'll probably want to do...
```console
git merge main
alembic history
alembic merge heads
alembic upgrade head
```
And then commit the new migration it has made for you.

src/pudl/output/ferc1.py
```diff
@@ -247,13 +247,20 @@ def _create_output_asset(
...
+            .merge(
+                kwargs["core_pudl__assn_ferc1_pudl_utilities"],
+                on="utility_id_ferc1",
+                how="left",
+                validate="many_to_one",
+            )
+            .merge(
+                pudl.glue.ferc1_eia.get_utility_map_ferc1(),
+                on="utility_id_ferc1",
+                how="left",
```
zaneselvans: I can see why you wanted to use this function, since it happens to conveniently pull the information you need out of a CSV, but doing it this way creates an implicit dependency outside of Dagster's asset graph, which entangles this asset with the implementation details of how these association tables are stored and retrieved.

Instead we can create an explicit dependency within the asset graph, using the upstream assets that contain these utility IDs in well-normalized association tables. Namely:

* `core_pudl__assn_ferc1_dbf_pudl_utilities`
* `core_pudl__assn_ferc1_xbrl_pudl_utilities`

To convey that explicit dependency with upstream assets and keep this asset from being dependent on the mechanics of how those assets are created, you can add those two assets as upstream dependencies in the `ferc1_output_asset_factory` definition alongside `core_pudl__assn_ferc1_pudl_utilities` and then left-merge them each onto the denormalized dataframe that's being constructed here using `utility_id_ferc1` in the same way that `core_pudl_assn_ferc1_pudl_utilities` is being merged on., which might look something like:

```suggestion
            .merge(
                kwargs["core_pudl__assn_ferc1_dbf_pudl_utilities"],
                on="utility_id_ferc1",
                how="left",
                validate="many_to_one",
            )
            .merge(
                kwargs["core_pudl__assn_ferc1_xbrl_pudl_utilities"],
                on="utility_id_ferc1",
                how="left",
                validate="many_to_one",
```

The same pattern of adding the explicit upstream assets as dependencies and merging them in individually can be applied below as well. Don't forget the `validate="many_to_one".

zaneselvans (changes requested):
Rather than using that glue function, which entangles these assets with some messy implementation details and creates dependencies that won't show up in the asset graph, I think we should pull in the association table assets as explicit inputs. See suggested edits.

src/pudl/output/ferc1.py
aesharpe: Thank you, great point! I'll update these

aesharpe: @zaneselvans what makes a table part of the `kwargs` dictionary?

src/pudl/output/ferc1.py
```diff
@@ -995,9 +1021,16 @@ def exploded_tables_asset(
...
+                kwargs["core_pudl__assn_ferc1_dbf_pudl_utilities"],
                 on="utility_id_ferc1",
                 how="left",
+                validate="many_to_one",
+            )
+            .merge(
+                kwargs["core_pudl__assn_ferc1_xbrl_pudl_utilities"],
+                on="utility_id_ferc1",
+                how="left",
+                validate="many_to_one",
```
zaneselvans: Unless you add these assets to the inputs of the `exploded_tables_asset` they won't be available for use here.

zaneselvans (approved):

zaneselvans:
- I rematerialized all my `out_ferc1` assets locally and didn't get any errors.
- I cracked open a couple of output tables and they included the new CSV/XBRL IDs which looked fine. As expected both of them contain a number of NA values for utilities that don't exist in both reporting epochs.
- I ran `dbt build` locally and those tests passed.
- If there are integration tests failures resulting from this change, the merge queue / integration tests should catch them, but I'd be a little surprised if that happened.

👍🏼

PR #4529, opened 2025-08-08
Update FERC-714 row counts changed by ID mapping fix.

zaneselvans:
# Overview

- Update row counts resulting from minor ID mapping change in #4439.
- Simplify `sort-row-counts` pre-commit hook to only look at full etl row counts, since we removed the fast ones.

# Testing

- Ran the ETL locally and used `dbt_helper` to generate new row counts for all tables.
- Compared those results to the ones from the nightly builds and found them to be identical.
- Re-ran all dbt tests locally with the new row counts and they pass.


dbt/seeds/etl_full_row_counts.csv
```diff
@@ -3443,24 +3443,24 @@ out_ferc714__hourly_planning_area_demand,2020,834384
...
-out_ferc714__respondents_with_fips,2015,8908
-out_ferc714__respondents_with_fips,2016,8906
-out_ferc714__respondents_with_fips,2017,8908
-out_ferc714__respondents_with_fips,2018,8924
-out_ferc714__respondents_with_fips,2019,8991
-out_ferc714__respondents_with_fips,2020,8995
-out_ferc714__respondents_with_fips,2021,8943
-out_ferc714__respondents_with_fips,2022,8953
-out_ferc714__respondents_with_fips,2023,8964
+out_ferc714__respondents_with_fips,2006,9242
```
krivard, line 3446: wait ~80 records affected by 1 id fix? is that reasonable?

or is this PR fixing row counts from other 714 changes as well?

zaneselvans: I think this is only changes to the row counts from the mapping change.

In this table there is one record for every combination of count FIPS ID in a respondent's territory, and the associated respondent, So if we've created a new respondent by no longer lumping it in with CAISO, we should expect to see a number of new rows, and that number will change over time as the associated counties change (the counties are pulled from the EIA861 service territories).

So I'm not surprised that these row counts changed, although I also did not investigate them.

PR #4530, opened 2025-08-08
Integrate 2024 FERC-714 data

### LINKED ISSUE #4409: Integrate 2024 FERC-714 data

Based on the [Existing Data Updates Docs](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html)

- [x] [Obtain fresh data](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#obtain-fresh-data)
- [x] [Map the structure of the new data](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#map-the-structure-of-the-new-data)
- [x] [Test data extraction](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#test-data-extraction)
- [x] [Update table and column transformations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#update-table-column-transformations)
- [ ] [Update the PUDL db schema](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#update-the-pudl-db-schema)
- [ ] [Connect datasets](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#connect-datasets)
- [ ] [Update the output routines](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#update-the-output-routines-and-run-full-tests)
- [x] [Run the ETL](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#run-the-etl)
- [x] [Run and update data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#run-and-update-data-validations)
- [x] [Update the documentation](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/existing_data_updates.html#update-the-documentation)

krivard:
# Overview

Closes #4409.

## What did you change?

* Updated DOI, etl*.yml, and data source to include new year of data
* Updated `pudl.transform.ferc714` with new fixes for 2024 data
  * One respondent temporarily held out due to missing forecast year / forecast year field having misc other text in it; I have a ticket in with FERC so hopefully that gets fixed at the source
  * New utc offsets
  * New gaps
  * New duplicate datetimes
* Added empty-tensor handling to part of the imputation code that was crashing the pipeline for me (did not investigate why or how)
* Updated the Existing Data Updates guide for 714, which still referenced CSV mappings

## Documentation

# Testing

## To-do list


krivard:
Oh rats -- this will need new row counts from the misalignment fixes in #4497.

The code changes were merged in with [f224be4](https://github.com/catalyst-cooperative/pudl/pull/4530/commits/f224be444ce64daea57e70b01ae3bf3d2f7b01dc), and I've kicked off a branch build here which should produce a correct rowcounts file: https://github.com/catalyst-cooperative/pudl/actions/runs/16838114955

If it's ready before I have to quit for the day I'll pull it in, otherwise some other kind soul will need to take over from here.

zaneselvans:
I'll give the row count update another shot!

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -3461,6 +3464,7 @@ out_ferc714__respondents_with_fips,2020,9032
 out_ferc714__respondents_with_fips,2021,8976
 out_ferc714__respondents_with_fips,2022,8987
 out_ferc714__respondents_with_fips,2023,8998
+out_ferc714__respondents_with_fips,2024,218
```
zaneselvans, line 3467: This is expected to be small, since without the EIA-861 data for 2024, there are no 2024 service territories.

docs/dev/existing_data_updates.rst
```diff
@@ -343,6 +334,28 @@ above. Continue to iterate and debug until assets generate successfully.
...
+  value should be.
+* :py:const:`pudl.transform.ferc714.DISCONTINUOUS_DATES` - Update this if you see
+  ``AssertionError: We expect there to be fewer than XX gaps in the xbrl time series
+  but we found these YY gaps:``, after investigating the new gaps and confirming
+  they occur on reasonable dates (usually around daylight saving time start or end).
+* :py:const:`pudl.transform.ferc714.DUPLICATED_DATETIMES` - Update this if you see
+  ``AssertionError: Found YY duplicate UTC datetimes, but we expected XX or less``,
+  after investigating the new duplicates and confirming they occur on reasonable
+  dates (usually when a respondent changes timezones, whether due to daylight saving
+  time or otherwise).
```
zaneselvans: Editing just because imprecise definitions of "timezone" cause so much data suffering.

```suggestion
* :py:const:`pudl.transform.ferc714.DUPLICATED_DATETIMES` - Update this if you see
  ``AssertionError: Found YY duplicate UTC datetimes, but we expected XX or less``,
  after investigating the new duplicates and confirming they occur on reasonable
  dates (usually when a respondent changes their UTC offset, whether due to daylight
  savings time or otherwise).
```

src/pudl/transform/ferc714.py
```diff
@@ -1142,7 +1191,15 @@ def spot_fix_forecast_years_xbrl(df):
...
-        df = df.astype({"forecast_year": "Int64"})
+        # [2024-aug kmm]
+        # there was some mixup and C004245 put text in their forecast year fields;
+        # excluding those records for now
+        text_in_year_mask = (
+            (df.respondent_id_ferc714_xbrl == "C004245")
+            & (df.report_year == 2024)
+            & (df.net_demand_forecast_mwh.isna())
+        )
+        df = df.loc[~text_in_year_mask].astype({"forecast_year": "Int64"})
```
zaneselvans: If they revise this filing in the future and fix the bad data, will we still end up excluding it? Or is there a way to gracefully pull it in if it gets fixed? Or maybe we want to raise an error so we can remove this special case?

zaneselvans (approved):
I updated the row counts and made a couple of suggested wording edits.

Not blocking, but a real question about the text-in-year fix: Is it set up such that it will gracefully accept corrected data if it arrives? Or maybe we want to get an error if there are no such problems in the future so we can remove the special case? @cmgosnell do you have thoughts?

src/pudl/transform/ferc714.py
zaneselvans: I updated this to raise an error if the data gets fixed upstream so we don't end up with a zombie spot fix. Thanks for filing a ticket with FERC @krivard how did you do that?

cmgosnell: i think the assert you added is a good thing here to catch if/when this becomes no longer necessary

zaneselvans: @cmgosnell Okay then I think I'll go ahead and add it to the merge queue.

zaneselvans:
For some reason in the fast ETL only, it's finding a couple of dupes in the `core_ferc714__respondent_id` table.

One of them is the Sacramento Municipal Utility District, and the reason seems to be that there are two values of `eia_code` associated with the respondent -- NA in the XBRL data and 16534 in the CSV data. My guess is that the EIA code shows up in some years and not others of the XBRL data, and the condensation of the XBRL data is losing the previously reported ID?  Either than or they revised their past filings and now it's missing entirely?  This problem doesn't come up in the full ETL.

I can get rid of the error by adding a spot-fix to assign the `eia_code` for the XBRL data, but that seems a little sketchy given that the ID does seem to exist in some XBRL data (it's fine in the full ETL).

The other supposedly non-unique record is "Wabash Valley Power Association, Inc." with `respondent_id_ferc714 == 159`.  Wabash starts out with a two values for `respondent_identification_code`: None and 40211 in the XBRL data. In the notebook that resolves successfully to 40211, but in Dagster ETL and CI it ends up having multiple values at the point where we are checking the table schema. I can also make this work with a spot fix

If I run `make pytest-integration` I get this duplicate and I've been able to reproduce it in Dagster by reprocessing the FERC-714 XBRL to SQLite and raw CSV data, but when I do the `RespondentId.run()` steps in a notebook I don't get any duplicates.

If I add (ill advised?) XBRL spot fixes for *both* of the IDs:

```
        "C011399": 16534,  # Sacramento Municipal Utility District (SMUD)
        "C000045": 40211,  # Wabash Valley Power Association, Inc.
```

Have we not had to deal with the case of having a mix of correct and NA `eia_codes` already?

`ensure_eia_code_uniqueness()` seems like it only checks for uniqueness but doesn't take any actions to ensure it, and doesn't raise an error if you have one numeric value along with NA values.

If that's the case, should we be doing `fillna()` in the `spot_fix_eia_codes()` rather than overwriting any values that are present?  If we did that then we would have no null values, and if the spot fix wasn't the same as the value that's present in some records, we'd get an error from `ensure_eia_code_uniqueness()` because there would be multiple non-null values.

Command to run just the FERC-714 xbrl to SQLite:

```console
ferc_to_sqlite --dataset-only ferc714_xbrl src/pudl/package_data/settings/etl_fast.yml
```

src/pudl/metadata/classes.py
```diff
@@ -1789,9 +1789,11 @@ def enforce_schema(self, df: pd.DataFrame) -> pd.DataFrame:
 
         df = self.format_df(df)
         pk = self.schema.primary_key
-        if pk and not (dupes := df[df.duplicated(subset=pk)]).empty:
+        if pk and not (dupes := df[df.duplicated(subset=pk, keep=False)]).empty:
```
zaneselvans, line 1792: I set keep=False here because it's more informative to see all of the duplicate records when the PKs are not unique -- or at least a sampling of them, when there are too many dupes to display.

src/pudl/transform/ferc714.py
```diff
@@ -251,6 +251,8 @@ class and any generic 714 transforms are defined as internal module functions.
         "C002447": 7004,  # Buckeye Power: was null or the entity_id
         "C001526": 14369,  # Avangrid Renewables: was null or the entity_id
         "C001132": 15248,  # PGE. Bad id was 43. New one lines up w/ CSV  and is EIA util
+        "C011399": 16534,  # Sacramento Municipal Utility District (SMUD)
+        "C000045": 40211,  # Wabash Valley Power Association, Inc.
```
zaneselvans: It feels wrong to be setting these with the code fixes, since they do exist in some cases, and so far as I can tell, the system is already supposed to deal gracefully with null values and automatically pick the values that are present. And the process seems to work as expected in the full ETL, but is breaking in the fast ETL.

But setting these does make everything work.

cmgosnell: perhaps this means these guys exist in the full years or data but not in the fast years.

zaneselvans: They both exist in the fast years, but in they come up as a mix of correct and null values, and those are the "duplicates" that are appearing in the output and causing the PK uniqueness check to fail -- we get one row for the correct value, and one for the NA value.

Where the two values come from varies between the two cases.  For SMUD there's a good value in the CSVs and no value in the XBRL.  For Wabash there's both good and null values coming from the XBRL.

zaneselvans: I pulled out some more of the respondent IDs from the middle of their processing and it seemed like there were other examples of IDs with a mix of valid and NA values, which aren't causing problem.s

|     |   respondent_id_ferc714 | respondent_id_ferc714_xbrl   | eia_code   |   eia_code_count |
|----:|------------------------:|:-----------------------------|:-----------|-----------------:|
| 115 |                       3 | C011389                      | 317        |                1 |
| 195 |                       3 | C011389                      | <NA>       |              nan |
| 167 |                      76 | C000029                      | 13434      |                1 |
|  65 |                      76 | C000029                      | <NA>       |              nan |
| 135 |                     106 | C001789                      | 12825      |                1 |
|   1 |                     106 | C001789                      | <NA>       |              nan |
|  87 |                     114 | C002422                      | 14610      |                1 |
|  60 |                     114 | C002422                      | <NA>       |              nan |
| 118 |                     120 | C003851                      | 15143      |                1 |
|  76 |                     120 | C003851                      | <NA>       |              nan |
| 123 |                     152 | C004480                      | 18642      |                1 |
|  78 |                     152 | C004480                      | <NA>       |              nan |
|   4 |                     159 | C000045                      | 40211      |                1 |
| 156 |                     159 | C000045                      | <NA>       |              nan |
| 145 |                     161 | C011370                      | 28503      |                1 |
|  20 |                     161 | C011370                      | <NA>       |              nan |
| 120 |                     167 | C001188                      | 20910      |                1 |
|   3 |                     167 | C001188                      | <NA>       |              nan |
| 121 |                     174 | C003849                      | 1307       |                1 |
|  63 |                     174 | C003849                      | <NA>       |              nan |
|  88 |                     195 | C011374                      | 17867      |                1 |
|  52 |                     195 | C011374                      | <NA>       |              nan |
| 173 |                     206 | C002869                      | 57995      |                1 |
|  53 |                     206 | C002869                      | <NA>       |              nan |
| 197 |                     216 | C011420                      | 17543      |                1 |
| 196 |                     216 | C011420                      | <NA>       |              nan |
| 151 |                     217 | C011454                      | 25471      |                1 |
|  31 |                     217 | C011454                      | <NA>       |              nan |

zaneselvans:
OKay, the divergence between the notebook and CI/Dagster behavior is due to the fact that:
- In the fast ETL we convert 2021 AND 2024 data from XBRL to SQLite -- so both of those years are in the FERC-714 DB I've been working form.
- HOWEVER, we only process the 2024 data further in the PUDL ETL.
- Dagster and the CI are explicitly using the fast ETL dataset settings, so when they use the IOManager to grab data, they only get the 2024.
- In my notebook, the default dataset settings (Full ETL) are being used, so when *I* pull the raw source asset, I get both 2021 and 2024 data, which has a different set of available `eia_code` values so it's not surprising that the two environments behave differently.

This explains some of the weird behaviors -- for CID C000045 with only data from 2024 it can get no valid `eia_code` from the XBRL data, so it's providing NA.

This means that both of the strange behaviors above are the same behavior -- they're both cases in which we have a correct `eia_code` coming from the CSV data, and no `eia_code` coming from the XBRL data, and currently there seems to be no process that attempts to fill in the XBRL gaps with the CSV IDs so we end up with the following situation:

Here our two problem IDs have either a `respondent_id_ferc714_xbrl` **or** an `eia_code` but never both. When we call `condense_into_one_source_table` and drop_duplicates on `["respondent_id_ferc714, "eia_code"]` we end up keeping both the NA and available versions, which results in the duplication.

```
     respondent_id_ferc714 respondent_id_ferc714_xbrl  eia_code
215                      3                    C011389       317
274                    132                    C011399      <NA>
131                    132                       <NA>     16534
290                    159                    C000045      <NA>
158                    159                       <NA>     40211
```

I've added another step to fill NA with whatever unique code is present, if any, before we do this drop duplicates and it seems to fix the problem. Running the integration tests locally now just to make sure, but take a look and see what you think @cmgosnell

src/pudl/transform/ferc714.py
```diff
@@ -605,6 +651,16 @@ def condense_into_one_source_table(df):
...
+    @staticmethod
+    def fill_missing_eia_codes(df: pd.DataFrame) -> pd.DataFrame:
+        """Fill missing eia_code values with unique non-null value per respondent."""
+        # Fill with first valid value. We don't have to worry about whether the eia_code
+        # is unique because that has already been verified
+        df["eia_code"] = df.groupby("respondent_id_ferc714")["eia_code"].transform(
+            "first"
+        )
+        return df
+
```
zaneselvans, line 663: @cmgosnell I have half a mind to combine this with the logic in the ensure EIA code is unique function, since it only works if the EIA codes are already unique, and because it would make the "uniqueness" that's being ensured also encompass NA values, which it doesn't right now.

cmgosnell: I think this is a good solution for this problem but imo combining it into the uniqueness check would be mixing purposes. and iirc the mixed null and valid codes in the uniqueness check is a good thing - there are instances of this but we fill in nulls so in that step we should expect to see nulls and other non-null unique values. I guess i can imagine doing this lil step and then the uniqueness check w/o removing the nulls as along as its explicitly stated.

zaneselvans:
Yay it passed locally.

src/pudl/transform/ferc714.py
krivard: @zaneselvans I sent email to ferconlinesupport@ferc.gov -- so either that's hooked up automatically to their ticket system or someone at FERC flagged it as ticketworthy

PR #4538, opened 2025-08-12
Update EIA 923 w/ 2025 Q3

### LINKED ISSUE #4516: EIA 923 Q3 2025 Update

cmgosnell:
# Overview

Closes #4516.

## Documentation

# Testing

## To-do list


dbt/seeds/etl_full_row_counts.csv
```diff
@@ -819,7 +819,7 @@ core_eia923__monthly_boiler_fuel,2021,106812
 core_eia923__monthly_boiler_fuel,2022,103536
 core_eia923__monthly_boiler_fuel,2023,102180
 core_eia923__monthly_boiler_fuel,2024,90999
-core_eia923__monthly_boiler_fuel,2025,72307
+core_eia923__monthly_boiler_fuel,2025,72226
```
cmgosnell, line 822: these went down slightly. I looked into it. it seems like all instances of these outputs that were removed were rows that were completely null or 0... so I'm not super concerned about this. but this does flow down into some of the other tables.

```python
pudl_paths = pudl.workspace.setup.PudlPaths()
pudl_engine=sa.create_engine(pudl_paths.pudl_db)
metadata = sa.MetaData()
metadata.reflect(pudl_engine)

tbl = "core_eia923__monthly_boiler_fuel"
new = defs.load_asset_value(tbl)
night = nightly_df = pd.read_parquet(f"s3://pudl.catalyst.coop/nightly/{tbl}.parquet")

pks = [c.name for c in metadata.tables[tbl].primary_key]
new = new.set_index(pks)
night = night.set_index(pks)

(
    night
    .loc[night.index.difference(new.index)]
    .drop(columns=["fuel_type_code_pudl","data_maturity"])
    .replace(to_replace=0, value=np.nan)
    .isnull().all()
)
```
got me:
```python
fuel_consumed_units    True
fuel_mmbtu_per_unit    True
sulfur_content_pct     True
ash_content_pct        True
dtype: bool
```

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -867,7 +867,7 @@ core_eia923__monthly_generation,2021,45156
 core_eia923__monthly_generation,2022,44280
 core_eia923__monthly_generation,2023,43944
 core_eia923__monthly_generation,2024,40260
-core_eia923__monthly_generation,2025,31224
+core_eia923__monthly_generation,2025,31116
```
cmgosnell: huh actually something weird is happening here... when i checked for non zero or null values I found one plant!

```python
tbl = "core_eia923__monthly_generation"
new = defs.load_asset_value(tbl)
night = nightly_df = pd.read_parquet(f"s3://pudl.catalyst.coop/nightly/{tbl}.parquet")

pks = [c.name for c in metadata.tables[tbl].primary_key]
new = new.set_index(pks)
night = night.set_index(pks)

lost_in_the_night = (
    night
    .loc[night.index.difference(new.index)]
    .drop(columns=["data_maturity"])
    .replace(to_replace=0, value=np.nan)
)
    
(lost_in_the_night[~lost_in_the_night.net_generation_mwh.isnull()])
```
I get this one plant:
image

which when i investigate has **two rows**  in the new ytd file... which is weird and wrong. i assume we must be dropping dupes somewhere otherwsie this tables' pks would scream

image

cmgosnell: okay i figured it out... we've been dropping all the duplicates like this not just the null ones which is what we were trying to do... but now i need to rerun all the row counts :-/

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -2481,14 +2481,14 @@ out_eia__yearly_plant_parts,2014,198298
...
-out_eia__yearly_plant_parts,2024,311673
-out_eia__yearly_plant_parts,2025,301715
+out_eia__yearly_plant_parts,2018,236381
+out_eia__yearly_plant_parts,2019,249885
+out_eia__yearly_plant_parts,2020,262115
+out_eia__yearly_plant_parts,2021,274261
+out_eia__yearly_plant_parts,2022,286160
+out_eia__yearly_plant_parts,2023,298310
+out_eia__yearly_plant_parts,2024,311883
+out_eia__yearly_plant_parts,2025,303501
```
cmgosnell: it might be good to write down somewhere why this is happening... if i understand correctly this ppe table is influenced by the presence of new boiler generator assn and unit ids. im a lil confused why it only goes back through 2018 though i thought 2015 was the change over year for bga's...

src/pudl/transform/eia923.py
cmgosnell, line 1: dang.... hm this ended up being kind of a lot... while im re-running things to update row counts i kinda wanna pull this de-duping out into its own little function that i can write some unit tests for.

cmgosnell:
hey @aesharpe  or @zaneselvans ! i've updated this deduping and am not running everything downstream (which includes cems 😞 ). i've updated the row counts for just this table for now which effect almost every year... and this table feeds into many tables so i expect this will change a lot of row counts. but while i wait if you wanna give this a wip review that would be helpful! (tryna get this merged in today)

src/pudl/transform/eia923.py
```diff
@@ -1000,14 +1000,56 @@ def _core_eia923__generation(raw_eia923__generator: pd.DataFrame) -> pd.DataFram
...
     # have duplicate records for a given generator/date combo. However, in all
     # cases one of them has no data (net_generation_mwh) associated with it,
     # so it's pretty clear which one to drop.
     unique_subset = ["report_date", "plant_id_eia", "generator_id"]
+    # reset the index omigosh we are using the index to drop records it needs
+    # to be actually unique
+    gen_df = gen_df.reset_index(drop=True)
     dupes = gen_df[gen_df.duplicated(subset=unique_subset, keep=False)]
-    gen_df = gen_df.drop(dupes.net_generation_mwh.isna().index)
+    drop_em = dupes[dupes.net_generation_mwh.isna() | (dupes.net_generation_mwh == 0)]
```
zaneselvans, line 1018: At this point we've already pivoted the table so that every row is a month instead of having monthly columns. So it seems possible that we'll only drop part of a year of data because it was only NA/zero in some of the months? Is that potentially an issue?

src/pudl/transform/eia923.py
```diff
@@ -1000,14 +1000,56 @@ def _core_eia923__generation(raw_eia923__generator: pd.DataFrame) -> pd.DataFram
...
+    # very much don’t expect to be the primary key for this table)
+    # we are going to find them... make sure they are the plant we expect... then
+    # aggregate them and effectively drop their prime mover code bc that is
+    # only column that differs in these records. We don't expect this table
+    # to vary by pm code.... also generators definitionally shouldn't have
+    # two prime_mover_codes
+    still_dupe_mask = gen_df.duplicated(subset=unique_subset, keep=False)
+    still_dupes = gen_df[still_dupe_mask]
+    assert all(still_dupes.plant_id_eia.unique() == 3405)
+    assert all(still_dupes.report_date.dt.year.unique() == [2012, 2013])
```
zaneselvans: Is this ordering guaranteed? You could use `sorted()` to make it not matter. Or make both sides into a `set`

```suggestion
    assert all(sorted(still_dupes.report_date.dt.year.unique()) == [2012, 2013])
```

src/pudl/transform/eia923.py
```diff
@@ -1000,14 +1000,56 @@ def _core_eia923__generation(raw_eia923__generator: pd.DataFrame) -> pd.DataFram
...
+        col
+        for col in still_dupes
+        if col not in unique_subset + ["net_generation_mwh", "prime_mover_code"]
+    ]
+    deduped = pd.merge(
+        still_dupes_gb[["net_generation_mwh"]].sum(),
+        still_dupes_gb[first_cols].first(),
+        left_index=True,
+        right_index=True,
+    ).reset_index()
```
zaneselvans: This feels a little contorted. Would this be equivalent?

```suggestion
    agg_cols = [
        col for col in still_dupes.columns 
        if col not in unique_subset + ["net_generation_mwh", "prime_mover_code"]
    ]
        
    deduped = (
        still_dupes
        .groupby(unique_subset)
        .agg({
            "net_generation_mwh": "sum",
            **{col: "first" for col in agg_cols}
        })
        .reset_index()
    )
```

src/pudl/transform/eia923.py
```diff
@@ -1000,14 +1000,56 @@ def _core_eia923__generation(raw_eia923__generator: pd.DataFrame) -> pd.DataFram
...
+
+    gen_df = pd.concat([gen_df[~still_dupe_mask], deduped])
```
zaneselvans: I don't know if it matters anywhere downstream but might use `ignore_index=True` here to prevent duplicate indices in the concatenated df.

```suggestion
    gen_df = pd.concat([gen_df[~still_dupe_mask], deduped], ignore_index=True)
```

src/pudl/transform/eia923.py
zaneselvans: I think at least these year/plant specific assertions need to be inside a conditional so that it doesn't break the fast ETL, which won't include these years of data.

Like `if not still_dupes.empty:`

dbt/seeds/etl_full_row_counts.csv
zaneselvans, line 1: I pushed updates to the row counts on the branch you're merging into after you kicked off your own ETL locally so watch out for weirdness.

zaneselvans (changes requested):
Some questions/suggestions and one blocking issue: I think the plant & year specific assertions need to be inside a conditional so they don't fail the fast ETL.

src/pudl/transform/eia923.py
cmgosnell: yea this would probably be better. i have just experienced unexpected behavior with `groupby.agg` enough times that I don't trust to enough

src/pudl/transform/eia923.py
cmgosnell: ah good point

src/pudl/transform/eia923.py
cmgosnell: yes at this point we've already pivoted the months. It could be as issue, but the final AssertionError would catch any instances where there were duplicate gen records with only a part of one getting dropped. if that wasn't there the PK check in `core_eia923__monthly_generation` would catch it

src/pudl/transform/eia923.py
cmgosnell: i did basically this and added a unit test for this step

src/pudl/transform/eia923.py
cmgosnell: i ended up checking if the bad years exist in the gen_df because it'd be great to learn one day this is no longer a problem

cmgosnell: i converted this to a set instead i liked that suggestion

src/pudl/transform/eia923.py
```diff
@@ -999,15 +999,63 @@ def _core_eia923__generation(raw_eia923__generator: pd.DataFrame) -> pd.DataFram
...
+    # they are duplicate except for having different prime movers (which we
+    # very much don’t expect to be the primary key for this table)
+    # we are going to find them... make sure they are the plant we expect... then
+    # aggregate them and effectively drop their prime mover code bc that is
+    # only column that differs in these records. We don't expect this table
+    # to vary by pm code.... also generators definitionally shouldn't have
+    # two prime_mover_codes
+    still_dupe_mask = gen_df.duplicated(subset=unique_subset, keep=False)
+    still_dupes = gen_df[still_dupe_mask]
+    if set(gen_df.report_date.dt.year.unique()) >= set({2012, 2013}):
```
zaneselvans, line 1036: I did not know you could use >= with sets like that. Cool.

zaneselvans: This will behave strangely if someone runs an ETL with just one of 2012 or 2013 data in it, but why would anybody do that.

zaneselvans (approved):
There ended up being a handful of row count discrepancies between the output of the branch build, and my local full ETL, which is a little disturbing. Hopefully it's just some discrepancy in how I ran it locally and not a persistent difference between the local and remote ETL outputs. :-/ 

I used the outputs from your branch deployment. Here are the differences:

```patch
diff --git a/dbt/seeds/etl_full_row_counts.csv b/dbt/seeds/etl_full_row_counts.csv
index 7b97511bb..1cf7980c2 100644
--- a/dbt/seeds/etl_full_row_counts.csv
+++ b/dbt/seeds/etl_full_row_counts.csv
@@ -2134,7 +2134,7 @@ out_eia923__monthly_generation_fuel_by_generator_energy_source,2008,337823
 out_eia923__monthly_generation_fuel_by_generator_energy_source,2009,339049
 out_eia923__monthly_generation_fuel_by_generator_energy_source,2010,346442
 out_eia923__monthly_generation_fuel_by_generator_energy_source,2011,352678
-out_eia923__monthly_generation_fuel_by_generator_energy_source,2012,359479
+out_eia923__monthly_generation_fuel_by_generator_energy_source,2012,359467
 out_eia923__monthly_generation_fuel_by_generator_energy_source,2013,362388
 out_eia923__monthly_generation_fuel_by_generator_energy_source,2014,368919
 out_eia923__monthly_generation_fuel_by_generator_energy_source,2015,369424
@@ -2258,7 +2258,7 @@ out_eia923__yearly_generation_fuel_by_generator_energy_source,2008,28559
 out_eia923__yearly_generation_fuel_by_generator_energy_source,2009,28455
 out_eia923__yearly_generation_fuel_by_generator_energy_source,2010,29013
 out_eia923__yearly_generation_fuel_by_generator_energy_source,2011,29577
-out_eia923__yearly_generation_fuel_by_generator_energy_source,2012,30127
+out_eia923__yearly_generation_fuel_by_generator_energy_source,2012,30126
 out_eia923__yearly_generation_fuel_by_generator_energy_source,2013,30347
 out_eia923__yearly_generation_fuel_by_generator_energy_source,2014,31031
 out_eia923__yearly_generation_fuel_by_generator_energy_source,2015,31039
@@ -2282,7 +2282,7 @@ out_eia923__yearly_generation_fuel_by_generator_energy_source_owner,2008,62274
 out_eia923__yearly_generation_fuel_by_generator_energy_source_owner,2009,62092
 out_eia923__yearly_generation_fuel_by_generator_energy_source_owner,2010,62842
 out_eia923__yearly_generation_fuel_by_generator_energy_source_owner,2011,64356
-out_eia923__yearly_generation_fuel_by_generator_energy_source_owner,2012,65246
+out_eia923__yearly_generation_fuel_by_generator_energy_source_owner,2012,65244
 out_eia923__yearly_generation_fuel_by_generator_energy_source_owner,2013,65742
 out_eia923__yearly_generation_fuel_by_generator_energy_source_owner,2014,67258
 out_eia923__yearly_generation_fuel_by_generator_energy_source_owner,2015,67040
@@ -2476,7 +2476,7 @@ out_eia__yearly_plant_parts,2007,155277
 out_eia__yearly_plant_parts,2008,161998
 out_eia__yearly_plant_parts,2009,169358
 out_eia__yearly_plant_parts,2010,163994
-out_eia__yearly_plant_parts,2011,170566
+out_eia__yearly_plant_parts,2011,170572
 out_eia__yearly_plant_parts,2012,178269
 out_eia__yearly_plant_parts,2013,190680
 out_eia__yearly_plant_parts,2014,198298
```

cmgosnell:
huh i got all the same row counts as the build locally except for I have the same out_eia__yearly_plant_parts for 2011.

zaneselvans:
* Hmm. I re-ran the full ETL locally overnight and I'm still getting the same diff as above.
* It's weird that 3 of them are from 2012, and then one from 2011.
* Given the tables and 2012 year, it seems like it could be flowing from the bug fix you made. But then it's surprising that it doesn't show up in any other downstream generation tables.

PR #4540, opened 2025-08-13
Fix pytest logs

### LINKED ISSUE #4482: No Dagster logs but many uninformative logs in pytest
# Overview
 
* I recently noticed that the integration tests seemed to be hanging hanging when I run them locally.
* They have tons of output from both alembic and Arelle, but nothing from the FERC data extractions or ETL.
* However, what really seems to have happened is we're no long outputting the logs generated by Dagster.
* We've also accumulated a lot of uninformative logging output through bitrot and some debug level logs.
* There are also many deprecation/future warnings coming from libraries that we use which aren't our problem.
* We have `filterwarnings` set in `pyproject.toml` that help with some things.
*  But the overall signal-to-noise ratio of the logs has gotten very bad, which makes debugging real issues or seeing warnings that we should take seriously harder.
* Example of the logging output in the comments.
* This problem happens in the CI on GitHub as well. See [these logs from a recent merge queue](https://github.com/catalyst-cooperative/pudl/actions/runs/16531319598/job/46757155012).

## Specific Issues

* [ ] Ensure that Dagster logs come through in pytest output.
* [x] Every error appears to be coming from the `pudl_service_territories` script, simply because it's the first test collected by pytest, which triggers the ETL. Can we force `test/unit/etl_test.py::test_pudl_engine` to be first instead? Yes we can. See [pytest-order](https://github.com/pytest-dev/pytest-order) -- fixed in #4474 
* [x] #4419 
* [ ] Arelle logs like: `[DEBUG] arelle:23 Try #0: taxonomy_source=<arelle.FileSource.FileSource object`
* [ ] DEBUG output from dbt phoning home using `urllib3.connectionpool`
* [ ] Thousands of lines of unhelpful DEBUG output from `matplotlib.font_manager`
* [ ] Alembic INFO/DEBUG logs from `alembic.runtime.migration`
* [ ] `catalystcoop.pudl.output.ferc1:2138` Calculations in Exploded Metadata can not be represented as a forest! (it's fine. it's a DAG)
* [ ] Chained assignment / inplace warning from `pudl/etl/glue_assets.py:689`
* [ ] `dagster/_core/definitions/definitions_class.py:476: UserWarning: Found asset job named etl_full of type <class 'dagster._core.definitions.unresolved_asset_job_definition.UnresolvedAssetJobDefinition'> passed to `jobs` parameter. Starting in dagster 1.11, you must now use Definitions.resolve_job_def to correctly retrieve this job definition.`
* [ ] `/Users/zane/code/catalyst/pudl/test/unit/workspace/datastore_test.py:310: PytestMockWarning: Mocks returned by pytest-mock do not need to be used as context managers.`

zschira:
# Overview

Closes #4482.

This PR fixes an issue with PUDL logs not showing up when running tests. This is done by appending the [`-s` option](https://docs.pytest.org/en/latest/how-to/capture-stdout-stderr.html#setting-capturing-methods-or-disabling-capturing) to the `pytest` command. I've added this to `pytest_args` in the `Makefile`, so if running pytest manually, a user would still need to add this argument to the command.

This PR also adds a new argument to the `configure_root_logger` function that allows for specifying per dependency log-levels, and increases the log-level for several dependencies that produce excessive logs when running tests.


test/conftest.py
```diff
@@ -42,6 +42,15 @@
 from pudl.workspace.datastore import Datastore
 from pudl.workspace.setup import PudlPaths
 
+pudl.logging_helpers.configure_root_logger(
+    dependency_loglevels={
+        "numba": logging.DEBUG,
+        "fsspec": logging.INFO,
+        "asyncio": logging.INFO,
+        "google": logging.INFO,
+        "alembic": logging.WARNING,
```
jdangerx: Nice to be able to set these independently :)

I'm non-blockingly curious about a couple of these:
* we were setting `numba` to `WARNING` before, why the downgrade to `DEBUG`?
* for `fsspec`, `asyncio`, and `google`, were they somehow being set to some non-`INFO` level in some other place?

Makefile
```diff
@@ -1,6 +1,6 @@
 gcs_cache_path := --gcs-cache-path=gs://zenodo-cache.catalyst.coop
 covargs := --append
-pytest_args := --durations 20 ${gcs_cache_path} --cov-fail-under=0
+pytest_args := --durations 20 ${gcs_cache_path} --cov-fail-under=0 -s
```
jdangerx: non-blocking: `-s` has been around and a stable part of `pytest` forever... do you know _why_ the `-s` is now needed while the logs were working before? Some sort of dagster change?

zaneselvans: Do we understand why this was working before, and suddenly stopped? Did pytest change?

We also have pytest configurations in `pyproject.toml` which apply universally. Would it make more sense to put this configuration in there? Is there anything in there that will conflict with this? Like the `log_cli*` options? The current configs include:

```toml
addopts = """\
    --verbose \
    --cov \
    --cov-append \
    --cov-config=pyproject.toml \
    --cov-report= \
    """
log_format = "%(asctime)s [%(levelname)8s] %(name)s:%(lineno)s %(message)s"
log_date_format = "%Y-%m-%d %H:%M:%S"
log_cli = true
log_cli_level = "DEBUG"
```

Will using `-s` (aka `--capture=no`) affect our ability to capture and write the logs out to a file in the nightly builds?

Hmm no I guess we're calling pytest directly in the nightly builds.

test/conftest.py
```diff
@@ -42,6 +42,15 @@
 from pudl.workspace.datastore import Datastore
 from pudl.workspace.setup import PudlPaths
 
+pudl.logging_helpers.configure_root_logger(
+    dependency_loglevels={
+        "numba": logging.DEBUG,
```
zaneselvans: Do we really want DEBUG logs for numba? It looks like we had it set to WARNING before.

src/pudl/logging_helpers.py
```diff
@@ -10,16 +10,24 @@ def get_logger(name: str):
     return logging.getLogger(f"catalystcoop.{name}")
 
 
-def configure_root_logger(logfile: str | None = None, loglevel: str = "INFO"):
+def configure_root_logger(
+    logfile: str | None = None,
+    loglevel: str = "INFO",
+    dependency_loglevels: dict[str, int] = {"numba": logging.DEBUG},
```
zaneselvans: Why the switch to DEBUG for numba here?

jdangerx (approved):
Glad this works, curious about why it does but that's never stopped me before.

I do note that I don't seem to have to add `-s` to get logs out when I'm running `pytest test/integration` on my machine but :shrug:

Makefile
zaneselvans: Maybe `capture=no` should be the default (set in `pyproject.toml`) and then we override that to allow the logs to be captured in the nightly builds?

test/conftest.py
zschira: Ah yes `numba` should be `WARNING`, fixed

Makefile
zschira: TBH I'm not really sure *why* this has changed. I also assumed it might dagster related, but what I can tell from the [dagster logging docs](https://docs.dagster.io/guides/monitor/logging/python-logging), dagster shouldn't be capturing the logs, and the fact that `-s` works to get the logs outputting again makes me think it doesn't have anything to do with dagster.

zaneselvans: Making this `--capture=no` rather than the shorthand would make it a little clearer what is being done, in case someone has to debug this behavior in the future.

zschira:
@zaneselvans I think I figured out the actual root of the problem here. I was trying to look a little bit deeper into a workaround to avoid using `-s` since it felt somewhat brute force, and I eventually realized that the logs were only being captured for the `catalystcoop` logger. If I created a logger with a different root the logs would show up fine. I now believe you and I are seeing this issue, but @jdangerx is not because in the `eqr-scoping` repo we used a dagster managed logger with the `catalystcoop` root, and now dagster is deciding to always capture the logs. I've found if I modify the name returned by `get_logger` in `python.logging_helpers` to anything other than `catalystcoop`, logs start showing up again. The thing is, I can't seem to figure out how to stop dagster from capturing the logs without changing the name.... There's probably something we can delete from `DAGSTER_HOME` to get this working, but I'm not sure what that would be.

zaneselvans:
Well THAT is mysterious. Hilarious.

zaneselvans:
I like using `catalystcoop` as the root because it lets us capture all and only the stuff that we're responsible for, since all of our packages use that ownership prefix for the package.

zaneselvans:
Wait @zschira I'm skeptical of this explanation because we're also losing the logs in CI on GitHub.

src/pudl/helpers.py
```diff
@@ -2045,7 +2045,7 @@ def get_dagster_execution_config(
 
     start_method_config = {}
     if "forkserver" in multiprocessing.get_all_start_methods():
-        start_method_config = {"forkserver": {"preload_modules": ["pudl"]}}
+        start_method_config = {"forkserver": {"preload_modules": ["pudl.init_logging"]}}
```
zschira, line 2048: I've added this `init_logging` module, which never gets imported except by the dagster multiprocess executor. This still accomplishes the goal of pre-loading PUDL, but also allows us to call `configure_root_logger` whithout it getting called at import time when running `pytest`.

test/conftest.py
```diff
@@ -400,7 +400,21 @@ def dataset_settings_config(request, etl_settings: EtlSettings):
 
 
 @pytest.fixture(scope="session")
-def pudl_datastore_config(request) -> dict[str, Any]:
+def logger_config():
```
zschira, line 403: Create a new fixture to configure the root logger and have `pudl_datastore_config` depend on this fixture to ensure it gets run early in the tests.

zschira:
@zaneselvans if you could test my latest round of changes here to see if it fixes the issue for you, that'd be great, because it feels like a fairly brittle fix to me, but it does avoid having to use the `-s` option.

Altogether my investigations into this logging stuff has left me feeling like the way `pytest` handles logging is very janky and confusing, so maybe no workaround that's not also kinda janky.

zaneselvans:
It seems like the PUDL logs are showing up in the unit tests now at least. Let me check out the branch and run the integration tests locally..

Seems to get the logging back.  I made some additional suggestions for dependencies that need to STFU.

src/pudl/__init__.py
```diff
@@ -18,8 +18,6 @@
     workspace,
 )
 
-logging_helpers.configure_root_logger()
```
zaneselvans, line 21: If this isn't configured here, and we're messing around in a notebook or otherwise running things that aren't in the context of a Dagster ETL run will we still get logs? Or do we need to set up our own logger? Maybe we already have to do that.

test/conftest.py
```diff
@@ -399,6 +399,20 @@ def dataset_settings_config(request, etl_settings: EtlSettings):
...
+    """Configure root logger to filter out excessive logs from certain dependencies."""
+    pudl.logging_helpers.configure_root_logger(
+        dependency_loglevels={
+            "numba": logging.WARNING,
+            "fsspec": logging.INFO,
+            "asyncio": logging.INFO,
+            "google": logging.INFO,
+            "alembic": logging.WARNING,
+        }
```
zaneselvans: I get a lot of DEBUG output from `arelle` as well. Should we set it to INFO or WARNING? omg also I get tens of thousands of lines of garbage from Matplotlib.

```suggestion
        dependency_loglevels={
            "numba": logging.WARNING,
            "fsspec": logging.INFO,
            "asyncio": logging.INFO,
            "google": logging.INFO,
            "alembic": logging.WARNING,
            "arelle": logging.INFO,
            "matplotlib": logging.WARNING,
        }
```

test/unit/dbt_helper_test.py
```diff
@@ -1199,6 +1200,7 @@ def test_complex_schema_diff_output():
     ],
 )
 def test_update_table_row_counts_update(
+    caplog,
```
zaneselvans, line 1203: Is this some magical built-in pytest fixture?

src/pudl/__init__.py
zschira: Yeah that's a good point. We should still get logs, but we wouldn't have the configured colored logs. I tried some work arounds where I was trying to check if `pytest` is running, then call `configure_root_logger` here if it's not, but the methods I could find for detecting `pytest` didn't end up working for me, which is why I went with the alternate `forkserver` based approach.

test/unit/dbt_helper_test.py
zschira: Yeah [here's](https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=caplog#std-fixture-caplog) documentation for it. Also added a comment with some explanation where I actually use the fixture.

I changed this because we were directly checking `stdout` for expected logs, and even with this logging fix I was still having trouble getting this test working, so I decided to use this to check the logger directly.

zaneselvans:
@zschira this seems like a significant improvement over the current situation, and now we have a place to shush logs from our dependencies if they're making a mess, so I think it's good to merge if you do.

PR #4544, opened 2025-08-14
Add documentation about using Postgres for Dagster logging.

zaneselvans:
# Overview

Adds some basic docs about how to set up Postgres for Dagster event logging, which should help with running the full ETL locally.

### DIFF
--- a/docs/dev/run_the_etl.rst
+++ b/docs/dev/run_the_etl.rst
@@ -152,8 +152,9 @@ There are two main Definitions in the PUDL processing pipeline:
    is partitioned by state and year (for the EPA CEMS).
 
 Both definitions have two preconfigured jobs:
-  - ``etl_fast`` processes one year of data
-  - ``etl_full`` processes all years of data
+
+  * ``etl_fast`` processes 1-2 years of data depending on the dataset.
+  * ``etl_full`` processes all available data.
 
 .. _run-dagster-ui:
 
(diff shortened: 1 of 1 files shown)


docs/dev/run_the_etl.rst
cmgosnell, line 1: my first inclination is that this whole thing should be over in the `dev_dagster` page or the `dev_setup` page.

docs/dev/run_the_etl.rst
```diff
@@ -294,7 +295,55 @@ want to rerun the entire ETL.
...
+writes its logs to a database. By default it uses SQLite, which has support built-in to
+Python. However, SQLite can't deal with concurrent writes, which means it can't handle
+the logging from our full ETL when it's highly parallelized.
+
+In the nightly builds we use a Postgres database to store this logging and events output
+instead of SQLite.  If you want to run the full ETL locally and use all available CPU
+cores to parallelize the process, you will also probably need to set up Postgres on your
+computer. This means installing Postgres, and creating a database for Dagster, and then
+telling Dagster how to access it. With Postgres set up, your
+``$DAGSTER_HOME/dagster.yaml`` configuration might look like this:
```
cmgosnell, line 315:
```suggestion
Dagster generates a large volume of its own logging and outputs to a database, which by
default gets written to a SQLite database. SQLite can't deal with concurrent writes, which
means it can't handle the logging from our full ETL when it's highly parallelized.

In the nightly builds we use a Postgres database to store this logging and events output
instead of SQLite and you can do this locally if you have any problems with the logging.
Setting up Postgres can be different on different operating systems but here is a
quickstart:

1. `Install Postgres  <https://www.postgresql.org/download/>`__ or `Postgres App for macs <https://postgresapp.com/downloads.html>`__.
2. Create a database for Dagster called `something?!?!?` with a username (?!?) and password (!?!?).
3. Add the following code block to  ``$DAGSTER_HOME/dagster.yaml``
```

docs/dev/run_the_etl.rst
zaneselvans: I looked at both of those places first too. The `dev_dagster` page iss more about how to create new assets & work with Dagster in a software development context, and the `dev_setup` page doesn't contain anything about Dagster at all right now -- it's all the environment setup that you do before getting to the Dagster part, so that didn't seem right either. This was the only place that really focused on the mechanics of running the ETL and working with the Dagster UI, so it seemed the most related.  Maybe it should be called something other than "Running the ETL?"

docs/dev/run_the_etl.rst
zaneselvans: I believe the magic incantations within Postgres are:

```sql
-- Create the database
CREATE DATABASE dagster;

-- Create the user with a password
CREATE USER dagster WITH PASSWORD 'dagster_password';

-- Grant all privileges on the database to the user
GRANT ALL PRIVILEGES ON DATABASE dagster TO dagster;
```

zaneselvans:
It turns out even with the friendly blue elephant of Postgres.app this was too finnicky for general consumption, and also Christina's issue wasn't the event logging volume anyway. If we want to make this work, we should probably do it with Docker. So, closing.

Issue #4381, opened 2025-07-02
Improve usability of Kaggle notebooks / datasets


zaneselvans:
# Overview

* Every time you fire up a Kaggle notebook, it has to copy all the input datasets that it depends on into a local workspace.
* [The PUDL Kaggle dataset](https://www.kaggle.com/datasets/catalystcooperative/pudl-project) is large. It contains both the SQLite and Parquet versions of the outputs, plus all the raw FERC SQLite DBs, which adds up to ~30GB when decompressed.
* This means that right now it takes ~6 minutes just to get where you can execute a cell of one of our notebooks.
* However, we also have the data available in our free S3 bucket, and while there's longer lag for each query when you get the data over the network, there's no initial startup lag.
* Pulling the data directly from S3 would also mean that the same notebook could work either on Kaggle or locally, or on some other hosted service, in browser, or with a GitHub action (assuming there's enough memory / CPU).
* **Unfortunately**, the raw FERC data is only available as SQLite DBs, which can't be read remotely.
* Apparently [DuckDB supports remote querying over HTTPS / S3](https://duckdb.org/docs/stable/guides/network_cloud_storage/duckdb_over_https_or_s3)
* It's now possible to create a collection of notebooks that's independent of any dataset!
* Unfortunately it's impossible to share a collection publicly.
* All of the Catalyst notebooks [are listed on our profile](https://www.kaggle.com/catalystcooperative/code) but if they don't depend on the PUDL Kaggle dataset, they won't show up on the dataset page.
* We could split out the FERC SQLite databases into a much smaller dataset for now, with an eye toward publishing them in DuckDB for remote access later.

# Success Criteria

How will we know that we're done?

* A user can open/fork one of our example notebooks on Kaggle and be executing cells within 30 seconds.
* The same notebooks can be run locally or on Kaggle without the user needing to mess around with paths / the filesystem.
* All of our example notebooks are tested on a regular basis so we know if they are still working or if changes to the nightly build outputs have broken them.
* All of our build outputs are available on Kaggle (including FERC SQLite outputs) (?)

zaneselvans:
I've updated the following Kaggle notebooks to:

* Pull data directly from the nightly build outputs in our free S3 bucket
* Not depend on the PUDL Kaggle dataset
* Run daily to check if they still work with the most recent nightly build output

###  Notebooks Updated
- [x] [01-pudl-data-access](https://www.kaggle.com/code/catalystcooperative/01-pudl-data-access)
[...]

PR #4546, opened 2025-08-16
Add GeoParquet support and integrate Census DP1 tables

zaneselvans:
# Overview

- We depend on ephemeral Census DP1 geospatial outputs in the PUDL ETL, but have not been saving them and distributing them. This was historically because neither SQLite  nor Parquet would handle spatial data easily, and without native support it could mean  very large output files.
- Instead we have distributed the relatively unprocessed Census DP1 dataset as SQLite, and we rely on that geospatial data downstream in our example notebooks. It also turned out we were relying on it directly in a few downstream FERC-714 / EIA-861 service territory assets and reading from the SQLite DB was real slow.
- Switching to pulling PUDL data from S3 in the notebooks makes them more ergonomic and portable, but so long as they also depend on the CensusDP1 SQLite DB, they still need to have local data access.
- In the last year GeoParquet has gained traction, and is now easy to use with  GeoPandas and DuckDB, which means it's easy to output this spatial data in an  analogous format to our other Parquet data.
- Rather than trying to hack around the Census DP1 SQLite DB on Kaggle I added support for GeoParquet / GeoDataFrames in our Parquet IO Manager and dtypes.
- I also added definitions for the Census DP1 fields and resources and `out_ferc714__georeferenced_respondents` which we've been creating but not distributing.
- In the course of debugging this I found that the metadata I was getting from the Pandera schema check failures wasn't super informative so I added some additional information.

In support of #4381 

# Testing

- [x] Materialized the new assets and read them in with both DuckDB and GeoPandas to verify they had real geoemetry columns.
- [x] Check that CRS is present in outputs & available via GeoPandas and DuckDB
- [x] Run `make pytest-coverage` locally
- [x] Run CI on GitHub
- [x] Run full ETL locally
- [x] Run dbt tests on full ETL outputs locally
- [x] Attempt to make a map from our Example Notebooks using the GeoParquet outputs
- [ ] Run a successful branch deployment

# Remaining Tasks

- [x] Do a self-review
- [x] Create a basic `censusdp1tract` data source page.
- [x] Split out a separate `GeoParquetIOManager` class.
- [x] Move asset check factory stuff into its own module.
- [x] Simplify renaming of Census DP1 columns
- [x] Simplify creation of `geo_metadata`
- [x] Simplify asset check dtype metadata
- [x] Improve census DP1 table descriptions.
- [x] Improve FERC-714 georeferenced respondents table description
- [x] Update the release notes.


src/pudl/metadata/enums.py
```diff
@@ -696,3 +696,22 @@
     "utility_operating_expenses",
 ]
 """Income types for FERC Form 1 data."""
+
+FUNCTIONAL_STATUS_CODES_CENSUS: list[str] = [
```
zaneselvans, line 700: Ideally should bring the short descriptions over too, but I'm just messing around right now.

src/pudl/metadata/constants.py
```diff
@@ -12,17 +13,19 @@
...
     "datetime": "datetime64[s]",
+    "geometry": "geometry",
     "integer": "Int64",
     "number": "float64",
     "string": "string",
     "year": "datetime64[s]",
 }
 """Pandas data type by PUDL field type (Data Package `field.type`)."""
 
-FIELD_DTYPES_PYARROW: dict[str, pa.lib.DataType] = {
```
zaneselvans, line 22: `pa.lib` was not actually part of the public API.

src/pudl/metadata/classes.py
```diff
@@ -647,7 +649,7 @@ def to_sql_dtype(self) -> type:  # noqa: A003
             return sa.Enum(*self.constraints.enum)
         return FIELD_DTYPES_SQL[self.type]
 
-    def to_pyarrow_dtype(self) -> pa.lib.DataType:
```
zaneselvans, line 650: `pa.lib` wasn't actually part of the public API.

src/pudl/analysis/service_territory.py
```diff
@@ -686,39 +679,22 @@ def pudl_service_territories(
...
-            "core_eia861__yearly_service_territory", pudl_engine
+        out_eia__yearly_utilities=get_parquet_table("out_eia__yearly_utilities"),
+        core_eia861__yearly_service_territory=get_parquet_table(
+            "core_eia861__yearly_service_territory"
         ),
-        core_eia861__assn_utility=pd.read_sql("core_eia861__assn_utility", pudl_engine),
+        core_eia861__assn_utility=get_parquet_table("core_eia861__assn_utility"),
         census_counties=county_gdf,
         dissolve=dissolve,
         save_format="geoparquet",
```
zaneselvans: This whole shebang was still reading everything from SQLite and was hella slow and memory intensive. Somehow we missed it in the transition to Parquet.

src/pudl/io_managers.py
```diff
@@ -309,27 +309,84 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
-        schema = res.to_pyarrow()
-        with pq.ParquetWriter(
-            where=parquet_path,
-            schema=schema,
-            compression="snappy",
-            version="2.6",
-        ) as writer:
-            writer.write_table(
-                pa.Table.from_pandas(df, schema=schema, preserve_index=False)
-            )
```
zaneselvans, line 330: It turns out this was more involved than it needed to be. We can just use `pq.write_table()` instead.

The only really annoying part about adding GeoParquet outputs support, which prevented me from just using `gdf.to_parquet()` rather than `df.to_parquet()` was the need to use an explicit schema. I couldn't pass the schema in to `gdf.to_parquet()` because it generates its own schema to handle the geometry metadata / types and they end up conflicting. And I couldn't seem to get it to handle the dictionary encoding of our categorical columns without the explicit schema, which we construct in `classes.py` -- so they kept turning into "string" columns breaking data type consistency in the round-trip to disk.

I feel like there should have been a cleaner way to do this.

src/pudl/analysis/service_territory.py
```diff
@@ -686,40 +678,22 @@ def pudl_service_territories(
...
+        core_eia861__assn_utility=get_parquet_table("core_eia861__assn_utility"),
+        census_counties=get_parquet_table(
+            "out_censusdp1tract__counties",
+            columns=["county_id_fips", "geometry", "county", "dp0010001"],
         ),
-        core_eia861__assn_utility=pd.read_sql("core_eia861__assn_utility", pudl_engine),
-        census_counties=county_gdf,
         dissolve=dissolve,
         save_format="geoparquet",
         output_dir=output_dir,
```
zaneselvans, line 699: Rather than reading all the data from SQLite, pull it from the parquet files, both for PUDL and Census DP1 data.

src/pudl/analysis/service_territory.py
zaneselvans, line 1: Minor naming tweaks in here but the main change is that the CLI pulls from Parquet rather than SQLite now, see the bottom.

src/pudl/analysis/state_demand.py
zaneselvans, line 1: Nothing major in here, just using the new table & column names.

src/pudl/etl/__init__.py
```diff
@@ -142,26 +267,32 @@ def asset_check_from_schema(
...
+            metadata.update(_process_schema_errors(schema_errors))
+            return AssetCheckResult(passed=False, metadata=metadata)
+
+        except Exception as e:
+            metadata["unexpected_error"] = {
+                "error_type": type(e).__name__,
+                "error_message": str(e),
+                "error_args": str(e.args) if hasattr(e, "args") else "No args",
+            }
+            return AssetCheckResult(passed=False, metadata=metadata)
```
zaneselvans: While debugging failures in the GeoParquet schema checks, the metadata I was getting in Dagster was extremely terse, so I added some additional information here and it was very helpful. Initial version was done by an LLM and I edited it to be a bit better organized. See helper functions above.

src/pudl/helpers.py
```diff
@@ -2278,15 +2279,24 @@ def get_parquet_table(
...
+        )
+    except (ValueError, TypeError):  # Fall back to normal Pandas DataFrame
+        df = pq.read_table(
+            source=parquet_path,
+            schema=pyarrow_schema,
+            columns=columns,
+            filters=filters,
+            use_threads=True,
+            memory_map=True,
+        ).to_pandas()
```
zaneselvans: This try-except feels like a hack, but we don't have metadata anywhere that indicates which tables are geospatial right now, so we can't know until we've tried to read the data. But it does work for now.

src/pudl/metadata/constants.py
```diff
@@ -12,17 +14,19 @@
     "boolean": "boolean",
     "date": "datetime64[s]",
     "datetime": "datetime64[s]",
+    "geometry": "geometry",
```
zaneselvans, line 17: This string shortcut works as expected so long as the dataframe it's being used in is a GeoDataFrame. So, we shouldn't name any columns anywhere else `geometry`.

src/pudl/metadata/constants.py
```diff
@@ -12,17 +14,19 @@
...
     "year": "datetime64[s]",
 }
 """Pandas data type by PUDL field type (Data Package `field.type`)."""
 
-FIELD_DTYPES_PYARROW: dict[str, pa.lib.DataType] = {
+FIELD_DTYPES_PYARROW: dict[str, pa.DataType] = {
     "boolean": pa.bool_(),
     "date": pa.date32(),
     "datetime": pa.timestamp("ms"),
+    "geometry": ga.wkb(),
```
zaneselvans, line 29: wkb = "well known binary" a generic geospatial storage type, only recently made available through GeoArrow.

src/pudl/metadata/fields.py
zaneselvans, line 1: A handful of human readable name changes in the non-demographic DP1 columns.  I compiled the descriptions & types for the demographic columns with an LLM, consuming public documentation and feeding it into our field metadata template and it worked well.

src/pudl/metadata/resources/censusdp1tract.py
```diff
@@ -0,0 +1,263 @@
...
+    "dp0210003",
+    "dp0220001",
+    "dp0220002",
+    "dp0230001",
+    "dp0230002",
+]
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "out_censusdp1tract__states": {
+        "description": "",
```
zaneselvans: Need to flesh out the descriptions of these tables.

src/pudl/metadata/resources/ferc714.py
```diff
@@ -142,9 +142,39 @@
...
+                "population",
+                "area_km2",
+                "demand_annual_mwh",
+            ],
+            "primary_key": ["respondent_id_ferc714", "report_date"],
+        },
+        "sources": ["ferc714", "censusdp1tract", "eia861"],
         "field_namespace": "ferc714",
         "etl_group": "outputs",
+        "create_database_schema": False,
```
zaneselvans: This asset already existed -- all I'm doing in this PR is writing it out to GeoParquet. It overlaps with the summarized demand asset substantially. We have some other pseudo-spatial tables that we should think about how to organize better. Generally the EIA-861 / FERC-714 service territory table organization is pretty messy. It didn't really get standardized when we moved to Dagster.

src/pudl/metadata/resources/ferc714.py
```diff
@@ -270,8 +300,6 @@
 }
 """FERC Form 714 resource attributes by PUDL identifier (``resource.name``).
 
-Keys are in alphabetical order.
-
```
zaneselvans, line 274: Lies!

src/pudl/metadata/classes.py
```diff
@@ -566,7 +567,9 @@ class Field(PudlMeta):
 
     name: SnakeCase
     # Shadows built-in type.
-    type: Literal["string", "number", "integer", "boolean", "date", "datetime", "year"]
+    type: Literal[
+        "string", "number", "integer", "boolean", "date", "datetime", "year", "geometry"
+    ]
```
zaneselvans, line 572: Just adding `geometry` and the appropriate types where needed.

zaneselvans:
I ran a branch deployment over the weekend and it failed, but not for reasons related to this PR. Just stochastic FERC-714 imputation data validation test failure (which seems to happen once a month or so...)

src/pudl/etl/__init__.py
```diff
@@ -132,6 +135,128 @@
 )
 
 
+def _collect_asset_metadata(asset_value, resource: Resource) -> dict[str, Any]:
+    """Collect basic metadata about the asset."""
+    return {
+        "asset_type": str(type(asset_value)),
+        "asset_shape": list(getattr(asset_value, "shape", "No shape attribute")),
```
jdangerx: Oooh helpful.

src/pudl/etl/__init__.py
jdangerx: This definitely seems helpful! I think it might make sense to pull the helper functions out of the `__init__.py` just to keep it from growing *more* into the Weird Grab Bag it already is.

src/pudl/output/censusdp1tract.py
```diff
@@ -70,10 +113,22 @@ def get_layer(layer, dp1_engine):
...
 
-            gdf = gpd.read_postgis(
-                table_name, dp1_engine, geom_col=geom_col, crs=crs_auth_str
+            gdf: gpd.GeoDataFrame = (
+                gpd.read_postgis(
+                    sql=table_name, con=dp1_engine, geom_col=geom_col, crs=crs_auth_str
+                )
+                .rename_geometry("geometry")
+                .drop(columns=["objectid"])
+                .rename(columns=layer_params[layer]["rename"] | common_rename)
```
jdangerx: nit: in the unlikely case we want to rename a column in the layer-params that shares a name with a column that's renamed in the `common_rename`, `layer_params[layer]["rename"]` will be overwritten, which is probably not what we want.

Quickest change is to swap to `columns=common_rename | layer_params[layer]["rename"]`.

I think a slightly more readable version of this is actually to define `common_rename` before `layer_params`, and when defining `layer_params` have something like:

```python
        "tract": {
            "plural": "tracts",
            "rename": common_rename | {
                "geoid10": "tract_id_fips",
                "namelsad10": "tract_name",
            },
        },
```

Then we can just have `.rename(columns=layer_params[layer]["rename"])` and there's just a little less state that we have to keep floating around in our heads.

src/pudl/output/censusdp1tract.py
```diff
@@ -9,34 +11,75 @@
...
     def census_layer(raw_censusdp1tract__all_tables, **kwargs) -> gpd.GeoDataFrame:
         """Select one layer from the Census DP1 database.
 
         Uses information within the Census DP1 database to set the coordinate reference
         system and to identify the column containing the geometry. The geometry column
-        is renamed to "geom" as that's the default within Geopandas. No other column
-        names or types are altered.
+        is renamed to "geometry" as that's the default within GeoPandas. Non-demographic
+        columns are renamed for readability, but the ~175 demographic data column names
+        are left untouched.
```
jdangerx, line 78: non-blocking, just for my clarification: What's the rationale for not pulling those names in?

src/pudl/metadata/constants.py
jdangerx: By this you mean, if it's not a GeoDataFrame we shouldn't have any columns with `Field.type == "geometry"`?

src/pudl/metadata/fields.py
```diff
@@ -1282,6 +1284,771 @@
         "description": "Dollar value of reported income, expense, asset, or liability.",
         "unit": "USD",
     },
+    # Census DP1 specific field definitions
+    "geometry": {
```
jdangerx, line 1288: non-blocking musing: it might be nice to one day break this out into something like `pudl.metadata.fields.censusdp1.FIELDS` so that we can introduce a bit more structure to this writhing mass of field names.

src/pudl/metadata/resources/censusdp1tract.py
```diff
@@ -0,0 +1,263 @@
+"""Resources derived from the US Census Demographic Profile 1 dataset."""
+
+from typing import Any
+
+CENSUS_DP1_DEMOGRAPHIC_COLUMNS = [
```
jdangerx, line 5: non-blocking: seems sort of funny to have this list of strings twice - I guess one advantage of pulling the fields into a separate variable as suggested above is that you could say `CENSUS_DP1_DEMOGRAPHIC_COLUMNS = [f for f in pudl.metadata.fields.census_dp1.FIELDS if f.startswith("dp")]` or something. Or even have `from pudl.metadata.fields.census_dp1 import DEMOGRAPHIC_COLUMNS; DEMOGRAPHIC_FIELD_NAMES = DEMOGRAPHIC_COLUMNS.keys()`.

jdangerx: continued musing: If, instead of "look up the field name in a big dictionary", we have "reference the field with a variable name", then `pudl.metadata.fields.census_dp1.DEMOGRAPHIC_COLUMNS` could have a list of field objects, and we would be able to reference it directly below...

obviously not in scope for this PR. Just noodling on this idea still.

src/pudl/output/censusdp1tract.py
```diff
@@ -9,34 +11,75 @@
 
 logger = pudl.logging_helpers.get_logger(__name__)
 
-LAYER_NAMES: dict[str, str] = {
-    "state": "states",
-    "county": "counties",
-    "tract": "tracts",
-}
+
+class LayerParams(TypedDict):
```
jdangerx, line 15: Huh! Never knew about `TypedDict`, always used `NamedTuple` for this sort of thing. Doesn't seem to make a real difference either way here but I'll keep this in my back pocket.

src/pudl/helpers.py
jdangerx: Agree that this, as well as the "well if it's GeoParquet do X and if it's not then do Y" in the IO manager below, feels a little hacky.

Personally I would split this into `get_parquet_table` and `get_geoparquet_table`:

Pros:
* makes each function have cleaner types, which helps your IDE
* the caller probably knows if it expects a GeoDataFrame or just a normal DataFrame
* you don't have this weird try-except

Cons:
* some code duplication in schema-finding etc.
* now the caller has to articulate if it expects GDF or just DF, but this might already happen because you'll be calling all the geopandas methods on the data anyways.

src/pudl/io_managers.py
```diff
@@ -309,27 +309,82 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
-
-    def load_input(self, context: InputContext) -> pd.DataFrame:
+        pa_schema = res.to_pyarrow()
+
+        if isinstance(df, gpd.GeoDataFrame):
```
jdangerx: Similarly to the try-except above, this feels like a sign that we should actually split the GeoDataFrame handling from the DataFrame handling - making a new `PudlGeoParquetIOManager` (what a mouthful!) and having the asset definition decide whether the output is GeoParquet or Pandas would be reasonable. The asset def knows whether it wants to make a geospatial resource or not, so we an avoid all this type-checking, branching, etc in here.

src/pudl/io_managers.py
jdangerx: Ugh, frustrating and weird. But the schema munging code looks totally readable to this newcomer. I think an explanatory comment like what you just wrote out would be helpful for future readers though!

src/pudl/io_managers.py
```diff
@@ -309,27 +309,82 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
+            pa_table = pa_table.replace_schema_metadata(metadata)
+        else:
+            # Regular DataFrame - the built in from_pandas() method is fine.
+            pa_table = pa.Table.from_pandas(df, schema=pa_schema, preserve_index=False)
+
+        pq.write_table(pa_table, parquet_path)
+
+    def _create_geoparquet_metadata(self, gdf: gpd.GeoDataFrame, res: Resource) -> str:
+        """Create GeoParquet metadata JSON string."""
+        import json
```
jdangerx: nit: why import json here instead of at the top level?

src/pudl/io_managers.py
```diff
@@ -309,27 +309,82 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
-            writer.write_table(
-                pa.Table.from_pandas(df, schema=schema, preserve_index=False)
-            )
-
-    def load_input(self, context: InputContext) -> pd.DataFrame:
+        pa_schema = res.to_pyarrow()
+
+        if isinstance(df, gpd.GeoDataFrame):
+            # Extract metadata before modifying geometry columns
+            geo_metadata = self._create_geoparquet_metadata(df, res)
```
jdangerx: nit: we have no use for this string, except for to encode it into `bytes`.

- could we just `geo_metadata = self._create_geoparquet_metadata(df, res).encode()`?
- could we have `self._create_geoparquet_metadata` return `bytes` directly?

Alternatively, we might want to actually manipulate the `geo_metadata` as a structured object. (Maybe? I can't really imagine a need...) In that case we could have `self._create_geoparquet_metadata` return a `dict` or a `TypedDict` or something. And then here we could have `geo_metadata = json.dumps(self._create_geoparquet_metadata(df, res)).encode()`. :shrug: This seems less likely to be useful, to me.

src/pudl/analysis/service_territory.py
jdangerx: Curious what the CLI is used for, and also whether we still need all these different output formats for the `compile_geoms` method. Though I'm totally fine leaving `compile_geoms` alone for this PR.

src/pudl/etl/__init__.py
```diff
@@ -142,26 +267,32 @@ def asset_check_from_schema(
...
+
     pandera_schema = resource.schema.to_pandera()
 
     @asset_check(asset=asset_key, blocking=True)
     def pandera_schema_check(asset_value) -> AssetCheckResult:
+        # Collect all metadata
+        metadata = {}
+        metadata.update(_collect_asset_metadata(asset_value, resource))
+        metadata.update(_collect_dtype_metadata(asset_value, resource, pandera_schema))
+        metadata.update(_collect_geometry_metadata(asset_value))
```
jdangerx: nit: this looks nicer to me but maybe not to you, I have a 1/10 attachment to this.

```suggestion
        metadata = (
            _collect_asset_metadata(asset_value, resource)
            | _collect_dtype_metadata(asset_value, resource, pandera_schema)
            | _collect_geometry_metadata(asset_value)
        )
```

src/pudl/etl/__init__.py
```diff
@@ -132,6 +135,128 @@
...
 
+def _collect_asset_metadata(asset_value, resource: Resource) -> dict[str, Any]:
+    """Collect basic metadata about the asset."""
+    return {
+        "asset_type": str(type(asset_value)),
+        "asset_shape": list(getattr(asset_value, "shape", "No shape attribute")),
+    }
+
+
+def _collect_dtype_metadata(asset_value, resource, pandera_schema) -> dict[str, Any]:
```
jdangerx: This function could probably use a little more editing post-LLM-output.

All we're doing here is comparing the `asset_value.dtypes` dictionary (if it exists) with the pandera schema. There's some slight munging to get some string values to compare but we don't need all this machinery...

* there's no real reason to have both `expected_dtypes` *and* `pandera_dtypes` - they're effectively two copies of the same data
* do we need `resource` and `pandera_schema` as params? I don't really think so, they both define the expected schema and `pandera_schema` is just `resource.schema.to_pandera()` anyways
* in fact, we don't need to go manually reconstructing a mapping from column names to column types, that's literally what the pandera schema is
* "column_present" seems like a silly thing to have if we're also going to output data in the shape of "here's what's missing from observed data & here's what's extra in the observed data" :shrug: - overall that output formatting block doesn't seem particularly considered.

jdangerx (commented):
Looks like it's going in a good direction! My main blocking concerns would be:

* I think it's *probably* better to split the general parquet and geoparquet machinery (the IO manager and the `get_parquet_table` function). At the call site it should be clear which one we want, and so there's no need to add complexity within the function, mess with our types, etc. What do you think?
* There's one function for the pandera stuff that needs a little more post-LLM cleanup.
* We might want to move all those helper funcs out into a different module to avoid cluttering `etl/__init__.py` even more

Other than that I have some non-blocking musings about future code organization and slight cleanups.

src/pudl/analysis/service_territory.py
zaneselvans: - This CLI is a vestigial, and could probably be replaced with some number of new GeoParquet output tables, but we didn't have the infrastructure to do geospatial outputs before. We should think about what spatial outputs make sense now, given the new tooling. But I think that's a separate PR.
- I think GeoParquet as the output is clearly what we want now.
- We need to be able to do both Utility & BA outputs, since they're different territories.
- We don't really understand whether limiting the included counties by state is appropriate / better. It was a heuristic.
- We should talk to potential users to understand what the use cases are for dissolved (one record with all the counties mushed together into a single geometry) vs. undissolved (all the counties separate, with one per row, linked by an ID) outputs.

src/pudl/etl/__init__.py
zaneselvans: I was just adding to where this kind of thing already was. Should we create a generic asset check / asset check factory / pandera-specific module?

src/pudl/io_managers.py
zaneselvans: No reason. LLM just seems to like local imports.

src/pudl/output/censusdp1tract.py
zaneselvans: I'm not 100% sure what you mean but, the Census has its own huge namespace of illegible data series identifiers, that look like `dp00100123` and if you're familiar with Census data, you probably know what they look like. They don't really have canonical legible names as far as I saw, but they do map to a hierarchical set of groups of field descriptions, which is what I used to construct the metadata.  Because we hardly use the demographic columns, and there are almost 200 of them, and their IDs have meaning in the Census column it didn't seem worthwhile to try and come up with our own fake names for them.  Whereas many of the non-demographic columns already had analogues in our metadata, and we are actually using them.

src/pudl/metadata/constants.py
zaneselvans: This may have been a brain fart.  What I meant was "If we name a column `geometry` it is going to be assumed (somewhere) to be of type geometry." but looking at the code now it seems like that's not the case.

A real thing here though is that in the context of a GeoDataFrame a column type of "geometry" will be correctly interpreted (in the same way as type "string" automagically becomes `pd.StringDType()`)  While in a normal DataFrame, pandas will have no idea what it means. Whereas I think a the dtype class / an instance of the class might be interpreted correctly in both cases.

src/pudl/io_managers.py
zaneselvans: Moved it up top.

src/pudl/helpers.py
zaneselvans: It's easy to split out the Parquet vs. GeoParquet in the IO manager since every asset has an IO manager associated with it, and we can set them appropriately to use the right IO Manager.

With this helper function that's not the case. I don't think it's reasonable to expect the caller to know whether it expects a DataFrame or GeoDataFrame. The function is used in a much more ad-hoc way, in notebooks, in iterations through lists of tables, in scripts, etc. For ergonomics I think it really needs to be able to do the right thing without anything except the table name to work from, and without forcing users to think about what kind of parquet it's grabbing.

In most cases, the user will probably just be using the DataFrame methods, and GeoDatFrames provide a superset of pandas functionality.

src/pudl/output/censusdp1tract.py
zaneselvans: Implemented. Definitely neater.

src/pudl/etl/__init__.py
```diff
@@ -132,40 +131,6 @@
 )
 
 
-def asset_check_from_schema(
```
zaneselvans, line 135: Moved into `asset_checks.py` along with all the other asset check factory stuff

src/pudl/metadata/fields.py
zaneselvans: The whole universal fields namespace thing definitely needs a re-think.

src/pudl/etl/__init__.py
zaneselvans: Okay, I pared this down a bit. But there's still some construction of the dtype dicts, mostly because we want to get informative outputs even when something is messed up, and trying to inline that when reading out of the pandera schema ended up seeming less readable overall.

src/pudl/io_managers.py
zaneselvans: Moved the encoding into the helper function itself so it returns bytes.

zaneselvans:
I think I've addressed the codey things. Still need to flesh out the descriptions and try running everything locally to make sure it still works, and do a branch deployment.

src/pudl/output/censusdp1tract.py
jdangerx: That's exactly what I was asking, thanks!

Makefile
```diff
@@ -69,6 +69,7 @@ docs-clean:
 	rm -f docs/data_dictionaries/pudl_db.rst
 	rm -f docs/data_dictionaries/codes_and_labels.rst
 	rm -rf docs/data_dictionaries/code_csvs
+	rm -f docs/data_sources/censusdp1tract.rst
```
jdangerx, line 72: ha. good catch!

docs/release_notes.rst
```diff
@@ -6,6 +6,48 @@ PUDL Release Notes
 v202X.XX.x (YYYY-MM-DD)
 ---------------------------------------------------------------------------------------
 
+Enhancements
+^^^^^^^^^^^^
+
+* We've started producing `GeoParquet <https://geoparquet.org/>`__ outputs that include
+  explicit geometries for use with `GeoPandas <https://geopandas.org/>`__ and other
+  mapping and geospatial analysis packages. See :func:`geopandas.read_parquet` for
+  dcoumentation on how to read We've also tested it with the `DuckDB
```
jdangerx:
```suggestion
  dcoumentation on how to read them. We've also tested it with the `DuckDB
```

docs/release_notes.rst
```diff
@@ -6,6 +6,48 @@ PUDL Release Notes
 v202X.XX.x (YYYY-MM-DD)
 ---------------------------------------------------------------------------------------
 
+Enhancements
+^^^^^^^^^^^^
+
+* We've started producing `GeoParquet <https://geoparquet.org/>`__ outputs that include
```
jdangerx: <sub>nit</sub> should we list which outputs have been geographized?

src/pudl/etl/__init__.py
jdangerx: Sweet, thanks for cleaning up my mess :)

src/pudl/io_managers.py
```diff
@@ -309,52 +310,39 @@ def load_input(self, context: InputContext) -> pd.DataFrame:
...
+            raise TypeError(f"Only pandas dataframes are supported, got {type(obj)}.")
         table_name = get_table_name_from_context(context)
         parquet_path = PudlPaths().parquet_path(table_name)
         parquet_path.parent.mkdir(parents=True, exist_ok=True)
 
         res = Resource.from_id(table_name)
-        df = res.enforce_schema(df)
+        df = res.enforce_schema(obj)
         pa_schema = res.to_pyarrow()
+        # Test whether these are actually necessary, or if we can use df.to_parquet()
```
jdangerx: Looks like we *can* use `df.to_parquet()`? Or is that waiting on branch builds for verification?

src/pudl/io_managers.py
```diff
@@ -381,13 +369,45 @@ def _create_geoparquet_metadata(self, gdf: gpd.GeoDataFrame, res: Resource) -> s
...
+            # warning from Geopandas about the geometry column not being a geometry.
+            logger.info(f"Convert geometry column {table_name}.{field_name} to WKB.")
+            gdf[field_name] = gdf[field_name].to_wkb()
+
+        # Convert to PyArrow table with explicit schema
+        pa_table = pa.Table.from_pandas(
+            gdf, schema=res.to_pyarrow(), preserve_index=False
+        )
+        # Add GeoParquet metadata
+        metadata = pa_table.schema.metadata or {}
```
jdangerx, line 402: non-blocking Q: does `pa_table.schema.metadata` have to have `bytes` instead of `str` everywhere?

src/pudl/metadata/classes.py
jdangerx, line 1: Huh. A bunch of funny reflowing, not sure if this was particularly intended but seems fine.

src/pudl/helpers.py
```diff
@@ -2281,22 +2280,22 @@ def get_parquet_table(
...
             filters=filters,
             schema=pyarrow_schema,
             use_threads=True,
             memory_map=True,
         )
     except (ValueError, TypeError):  # Fall back to normal Pandas DataFrame
-        df = pq.read_table(
-            source=parquet_path,
-            schema=pyarrow_schema,
+        df = pd.read_parquet(
```
jdangerx, line 2294: non-blocking: curious why you decided to split the IO managers into geo/normal but not this `get_parquet_table` function? This is fine, it works, etc. just wondering about your intent.

jdangerx (commented):
Looks nicer! I think there's a small amount of cleanup remaining & I have a few non-blocking questions. Happy to take one more crack at this review once everything is sorted :)

src/pudl/helpers.py
zaneselvans: Huh, I left a long comment explaining this.

src/pudl/helpers.py
zaneselvans: @jdangerx explanation of not splitting here.  Not sure why the comments on this PR are such a mess.

src/pudl/metadata/classes.py
zaneselvans: I had my editor reflow so I could read it while trying to figure out how the resource description stuff works.

src/pudl/io_managers.py
zaneselvans: Whoops yes! Seems to work fine. Will remove.

src/pudl/io_managers.py
zaneselvans: Actually it seems like this byte encoding is entirely unnecessary. [Just leaving it a JSON string](https://geoparquet.org/releases/v1.0.0/).

> A GeoParquet file MUST include a geo key in the Parquet metadata (see [FileMetaData::key_value_metadata](https://github.com/apache/parquet-format#metadata)). The value of this key MUST be a JSON-encoded UTF-8 string representing the file and column metadata that validates against the [GeoParquet metadata schema](https://geoparquet.org/releases/v1.0.0/schema.json). The file and column metadata fields are described below.

docs/release_notes.rst
zaneselvans: Sure, I'll add it.

src/pudl/helpers.py
jdangerx: Yeah, I can see it being easier for the caller to just always call `get_parquet_table('name')` for sure.

I'm down to keep it in one function, *and* I still think it's pretty reasonable for the caller to decide how to read the table. What they need to know is not "is this a geo table?", it's "do I want to *treat* this as a geo table?"

I guess the tradeoff is between:
* "I just want the most capable dataframe possible from this table name with the minimum amount of fuss" - keep it as one function, you might run into a situation where you thought something was a `GeoDataFrame` but it's actually a `DataFrame` :shrug: annoying but not the end of the world.
* "I don't want to have to check which kind of dataframe came back from the function, before figuring out what I can safely do with it" - split it into two functions, you'll never run into that kind of bug above but also you have to decide when reading a table whether you care about the geo functionality or not.

zaneselvans: I guess I'm not seeing the case for wanting to be able to try and read a table into a container that can't fully acommodate the data it contains.

- You *can* read a geoparquet file into a normal pandas `DataFrame`. But you'll just have a big binary `geometry` column of type `object` that you can't do anything with directly.
- You *can't* read data without a valid `geometry` column into a `GeoDataFrame` -- without that geospatial data and associated metadata it will fail.

I guess another way to go about this is to check the schema for the table and see if there's a `geometry` column first, and if there is use `GeoDataFrame` and if there isn't, use `DataFrame`.  This would also be better since it's possible to read only a subset of the columns in a table with the helper function, and if they pull only non-geometry columns from a geoparquet file, they won't be able to make a GeoDataFrame anyway...

src/pudl/helpers.py
```diff
@@ -2278,7 +2278,10 @@ def get_parquet_table(
     resource = Resource.from_id(table_name)
     pyarrow_schema = resource.to_pyarrow()
 
-    try:  # Attempt to read as a GeoDataFrame
+    is_geospatial = (columns is not None and "geometry" in columns) or (
```
jdangerx: We should probably key on the type in the schema, not the column name - it looks like we can set the active geometry column to an arbitrary name in GeoDataFrames. At least that's how I'm reading [the docs](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.html).

> geometry: str or array-like (optional)
>
> Value to use as the active geometry column. If str, treated as column name to use. If array-like, it will be added as new column named ‘geometry’ on the GeoDataFrame and set as the active geometry column.
>
> Note that if geometry is a (Geo)Series with a name, the name will not be used, a column named “geometry” will still be added. To preserve the name, you can use [rename_geometry()](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.rename_geometry.html#geopandas.GeoDataFrame.rename_geometry) to update the geometry column name.

src/pudl/helpers.py
jdangerx: > You can read a geoparquet file into a normal pandas DataFrame. But you'll just have a big binary geometry column of type object that you can't do anything with directly.

Yeah, but maybe you don't care about the geometry column :shrug: 

> You can't read data without a valid geometry column into a GeoDataFrame -- without that geospatial data and associated metadata it will fail.

Right, and if you *think* there's a valid geometry column but there isn't, the "try GDF, then try DF if it fails" approach will silently spit out a DataFrame. Then when you try to do something Geo-ey to the DataFrame it will blow up. Which should happen pretty quick and not be too confusing, but it's clearer what's going on when you explicitly say "read this as gdf" vs "read this as normal df".

src/pudl/helpers.py
zaneselvans: * In theory you can have any number of geometry columns, and change which one is currently the active "primary" geometry column in the GeoParquet metadata.
* Currently the only column that is of type `geometry` in PUDL is the column named `geometry`.
* But that could change. So now I iterate through all of the columns that we're actually going to read, and check whether in the schema of the table we are going to read any of them are of type `geometry`. And if so, then we're reading a GeoParquet file.
* I think there are other things that would have to change elsewhere to enable multiple geometry columns with other names. But let's cross that bridge later.

src/pudl/helpers.py
```diff
@@ -2301,7 +2301,7 @@ def get_parquet_table(
         )
 
     # Only enforce schema if we're reading all columns
-    if columns is None:
+    if set(columns) == set(resource.get_field_names()):
```
jdangerx, line 2304: slick!

jdangerx (approved):
Since the ongoing back and forth over `get_[geo]parquet_table` is more "academic" than "actually useful" I think we can try to merge this, pending the branch builds working with the latest tweak to the calculation of `is_geospatial`. :shipit:

src/pudl/helpers.py
zaneselvans: Seemed closer to what we *actually* mean.

zaneselvans:
Okay, I'm merging #4557 and once it lands on `main` I will kick off a branch build for these changes, and if it passes I'll merge this in tonight.

PR #4548, opened 2025-08-17
Stop quietly nulling values that conflict with ENUM constraints

### LINKED ISSUE #4482: No Dagster logs but many uninformative logs in pytest

zaneselvans:
# Overview

- Stop quietly nulling out all values that don't conform to stated ENUM constraints.
- Raise an error instead of logging a warning in call to `enforce_schema`
- Add the rare `energy_source_code` 9-12 to ENUM constraints. (previously nulled)
- Fix Chugach Electric Co-op timezone: `Anchorage` -> `America/Anchorage` (previously nulled)
- Add Alembic migration to accommodate new ENUM values

## Other tweaks
- Switch to warning on non-DAG structure, rather than non-forest structure of calculation graph in exploded FERC 1 tables.
- Shush the vapid DEBUG logs from `urllib3`
- Remove unnecessary context manager for mocker in tests.
- Add FERC-714 XBRL to SQLite integration test, and run it before main PUDL ETL.

Closes #4482 

# Testing

- Ran full ETL locally and it failed because of the bad ENUM values.
- Fixed ENUM issues and re-ran full testes locally and they passed.

## Tasks

- [x] Address the ENUM values that were previously getting quietly nulled.


src/pudl/extract/dbf.py
```diff
@@ -585,7 +585,6 @@ def load_table_data(self):
                 continue
             new_df = self.transform_table(table, new_df)
 
-            logger.debug(f"    {table}: N = {len(new_df)}")
```
zaneselvans, line 588: Duplicative with logging below.

src/pudl/metadata/classes.py
```diff
@@ -1814,7 +1814,7 @@ def format_df(
                     if value not in dtypes[field.name].categories
                 ]
                 if uncategorized:
-                    logger.warning(
+                    raise ValueError(
```
zaneselvans, line 1817: This is the big change -- previously any value that didn't conform to the ENUM constraint would get logged as a warning and then... quietly nulled. And we didn't really watch the warnings. Now it'll fail the run.

src/pudl/metadata/codes.py
```diff
@@ -1823,7 +1823,7 @@
 BPAT,bonneville_power,Bonneville Power Administration,America/Los_Angeles,Northwest,,NW,western,False
 CHPD,public_utility_district_of_chelan_county,Public Utility District No. 1 of Chelan County,America/Los_Angeles,Northwest,,NW,western,False
 CISO,california_iso,California Independent System Operator,America/Los_Angeles,California,,CAL,western,False
-CEA,chugach_electric,Chugach Electric Assn Inc,Anchorage,,,,,
+CEA,chugach_electric,Chugach Electric Assn Inc,America/Anchorage,,,,,
```
zaneselvans, line 1826: Swap in the real value.

src/pudl/metadata/fields.py
```diff
@@ -2262,7 +2262,7 @@
             "reported in for the generator referenced in the same record."
         ),
         "constraints": {
-            "enum": sorted({f"energy_source_code_{n}" for n in range(1, 9)})
+            "enum": sorted({f"energy_source_code_{n}" for n in range(1, 13)})
```
zaneselvans, line 2265: Dunno if we ever actually see one of these values with an actual fuel, but they do show up in name.

src/pudl/output/ferc1.py
```diff
@@ -2190,9 +2190,9 @@ def forest(self: Self) -> nx.DiGraph:
         forest.remove_nodes_from(almost_pure_stepparents)
 
         forest = self.prune_unrooted(forest)
-        if not nx.is_forest(forest):
+        if not nx.is_directed_acyclic_graph(forest):
```
zaneselvans, line 2193: The Forest expectation is, I think, a vestige of the original design / expectations. And eventually we realized that the calculation network could be a DAG in reality.

test/conftest.py
```diff
@@ -410,6 +410,7 @@ def logger_config():
             "google": logging.INFO,
             "alembic": logging.WARNING,
             "arelle": logging.INFO,
+            "urllib3": logging.INFO,
```
zaneselvans, line 413: 🤫

test/integration/ferc_xbrl_extract_test.py
```diff
@@ -54,6 +55,23 @@ def test_ferc1_xbrl2sqlite(ferc1_engine_xbrl: sa.Engine, ferc1_xbrl_taxonomy_met
...
+        "identification_and_certification_01_1_duration"
+        in sa.inspect(ferc714_engine_xbrl).get_table_names()
+    )
+
+    assert isinstance(ferc714_xbrl_taxonomy_metadata, dict)
+    assert "core_ferc714__hourly_planning_area_demand" in ferc714_xbrl_taxonomy_metadata
+    assert len(ferc714_xbrl_taxonomy_metadata) > 1
+    assert len(ferc714_xbrl_taxonomy_metadata) < 10
+
+
```
zaneselvans: Nominal test, as with the FERC 1 -- the real point here is to make it clear that this is a separate pre-ETL step that needs to be run (via the fixture dependency) so that the real PUDL ETL can kick off, rather than having the fixtures mysteriously kick off activities for some not obviously related downstream dependency.

test/unit/workspace/datastore_test.py
```diff
@@ -306,11 +306,9 @@ def test_get_zipfile_resource_failure(mocker):
     ds = datastore.Datastore()
     ds.get_unique_resource = mocker.MagicMock(return_value=b"")
     sleep_mock = mocker.MagicMock()
-    with (
-        mocker.patch("time.sleep", sleep_mock),
-        mocker.patch("zipfile.ZipFile", side_effect=zipfile.BadZipFile),
-        pytest.raises(zipfile.BadZipFile),
-    ):
```
zaneselvans, line 313: These context managers were generating a bunch of logging messages because they aren't necessary -- the mock object cleans up after itself.

src/pudl/metadata/classes.py
jdangerx: Oof, yeah, we should raise an error here!

src/pudl/output/ferc1.py
jdangerx: Sweet, this is unrelated to the ENUM stuff, just quieting the logs yes?

test/conftest.py
jdangerx: I have often wondered to myself, "why do we output all these urllib3 debug logs?" and now I will wonder no longer.

test/integration/ferc_xbrl_extract_test.py
jdangerx: nit: why not put this useful comment into the docstring of the test?

test/unit/workspace/datastore_test.py
jdangerx: Eep, thanks for cleaning up after me hah! Not sure what I was thinking back then...

jdangerx (approved):
Wow, really glad this didn't bite us any harder than it did. Thanks for the incidental logging cleanup too!

src/pudl/output/ferc1.py
zaneselvans: Late in the protracted process of building these detailed asset/liability tables we finally accommodated the fact that the calculation graph was actually a DAG and not (as we'd originally imagined / required) a collection of trees (a forest). These BIG SCARY RED logging outputs were a holdover from the prior expectations (as are all of the now outdated names) and I'm changing the logging here to be aware of the thing we actually belatedly expect.

Not related to the ENUM constraints at all.

test/integration/ferc_xbrl_extract_test.py
zaneselvans: I went ahead and added notes like this to all 3 of the FERC extraction tests (FERC 1 DBF & XBRL, FERC 714 XBRL)

PR #4556, opened 2025-08-21
Update FERC-714 DOI and special handling for salt river 2024 filings

krivard:
# Overview

In release v2025.08.0, we held out the FERC-714 filings from Salt River Project, because they had text in the forecastYear field.

FERC had them submit a corrected filing. The new filing is good, but `filter_for_freshest_data` isn't quite able to match everything up to override the bad old values with the new good values, causing some replication. So, we still need to manually hold out the bad old Salt River Project entries, just do it in such a way that we don't accidentally drop the good new Salt River Project entries.

## What did you change?

This PR updates the DOI to pull the new good data, and updates the special SRP handling to drop the bad old data more precisely.

## Documentation

# Testing

## To-do list


zaneselvans:
This is kind of disturbing. Like there are thousands of revised filings from across the years. Why is this one being particularly ornery? Or are there lots of others which are having problems that we're just not noticing? Or did something change about how revisions are integrated?

krivard:
@zaneselvans My bet is that this is the first time we've had an error in a filing that (1) placed a string in an int field that's (2) part of the primary key. Otherwise, filter_for_freshest_data could've handled matching to the corrected filing.

zaneselvans:
Well, cool that the ticket with FERC resulted in a new filing at least. I wonder if we could set up some kind of automated system for identifying errors worth telling them about and send a regular summary report or something...

src/pudl/transform/ferc714.py
```diff
@@ -1203,14 +1203,17 @@ def spot_fix_forecast_years_xbrl(df):
...
+        # there was some mixup and C004245 (Salt River Project) put text in their
+        # forecast year fields in the 2024 report. they submitted a new filing with
+        # correct data, but filter_for_freshest_data isn't able to detect a match
+        # between the two sets of entries to be able to override the bad data, and
+        # keeps both sets of entries. we drop the bad entries manually here.
         text_in_year_mask = (
             (df.respondent_id_ferc714_xbrl == "C004245")
             & (df.report_year == 2024)
-            & (df.net_demand_forecast_mwh.isna())
+            & ~(df.forecast_year.isna() | df.forecast_year.str.startswith("2"))
```
cmgosnell, line 1217: I think i understand the gist of all of this but can you add an explanation for why the forecast_year startswith does the trick??

cmgosnell: oh i'm now understanding i didn't see the ~ and i thought you were removing everything w/ a forecast_year starting w/ 2

cmgosnell (approved):
it a bummer that the filter_freshest data process gives us some mixed results with this ... fresher data. but knowing a bit about how it chooses records I'm not entirely surprised. this all makes sense to me and looks reasonable.

src/pudl/transform/ferc714.py
```diff
@@ -1203,14 +1203,17 @@ def spot_fix_forecast_years_xbrl(df):
         expected range.
         """
         # [2025-aug kmm]
-        # there was some mixup and C004245 put text in their forecast year fields;
-        # excluding those records for now
+        # there was some mixup and C004245 (Salt River Project) put text in their
+        # forecast year fields in the 2024 report. they submitted a new filing with
+        # correct data, but filter_for_freshest_data isn't able to detect a match
+        # between the two sets of entries to be able to override the bad data, and
+        # keeps both sets of entries. we drop the bad entries manually here.
```
krivard:
```suggestion
        # there was some mixup and C004245 (Salt River Project) put text in their
        # forecast year fields in the 2024 report. they submitted a new filing with
        # correct data, but because the correction makes a change to a primary
        # key column, filter_for_freshest_data isn't able to detect a match
        # between the two sets of entries to be able to override the bad data, and
        # keeps both sets of entries. we drop the bad entries (which match this
        # respondent for this report year, and start with something other than "2")
        # manually here.
```

PR #4557, opened 2025-08-21
Fix remaining logging issues

zschira:
# Overview

This PR fixes issues with logs not being configured when running from dagster, and a build failure due to weird pytest/click interactions.

### DIFF
--- a/docs/dev/pudl_id_mapping.rst
+++ b/docs/dev/pudl_id_mapping.rst
@@ -168,16 +168,16 @@ to have the same PUDL ID.
     plants csv for the first plant under 5 MW and everything below that can remain
     unlinked.
 
-For each new record, search the entire plants_combined tab for a piece of the
+For each new record, search the ``pudl_id_mapping`` spreadsheet for a piece of the
 plant name string (e.g. for ``chenango solar``, you could search for ``chen``,
-or ``chenan``). Searching the entire plant tab helps find other records within
-both FERC and EIA that may be the same or part of the same facility. Searching
+or ``chenan``). If the new records you're mapping come from FERC, look for matches
+in the ``plant_name_eia`` column (and vice versa for EIA). Searching
 for a piece can help catch misspellings in the plant name, which are more common
-in the FERC records. Use the ``devtools/pudl_id_mapping_help.ipynb`` notebook to speed
-up this process.
+in the FERC records. You can also use the ``devtools/pudl_id_mapping_help.ipynb``
+notebook to speed up this process.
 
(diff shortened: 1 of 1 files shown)


src/pudl/init_logging.py
```diff
@@ -8,9 +8,6 @@
 ``log_cli`` option to loggers which are created at import time.
 """
 
-import os
-
 from pudl.logging_helpers import configure_root_logger
 
-if "PYTEST_CURRENT_TEST" in os.environ:
```
zschira, line 15: This was a typo where the I had the conditional flipped here. The fact that this didn't cause issues with the tests has made me realize that this check doesn't actually seem to be necessary in the first place.

I had included this `if` statement because `pytest` will run a test where it imports all of the modules within the `pudl` package, but I think because this is running as a test, and not running at import time, it doesn't seem to cause the same issues with creating the `logger` at import time.

test/unit/dbt_helper_test.py
```diff
@@ -1248,17 +1246,20 @@ def test_update_table_row_counts_update(
...
-            update_tables,
-            [
-                "test_source__table_name",
-                "--update",
-                "--row-counts",
-            ],
-        )
-    assert "Successfully updated row counts for test_source__table_name" in caplog.text
+    # Mock logger.info to check for expected logging outputs
+    logger_mock = mocker.patch("pudl.scripts.dbt_helper.logger.info")
```
zschira, line 1250: It seems like there's some really weird interactions between `click` and `logging`, which are documented in [this issue](https://github.com/pallets/click/issues/824). The issue was recently closed, but I'm not sure the changes are actually in a release that we can use yet, so I just switched to fully `mocking` the `logger.info` method and checking that rather than trying to deal with capturing the logger/stdout. This works for me with either method of running `pytest`.

zaneselvans:
# Stuff to check:

- [x] Logs w/ formatting and INFO level are visible in Dagster UI
- [x] `pytest test/unit/dbt_helper.py` doesn't crash
- [x] Logs are visible when running pytest directly
- [x] Logs are not polluted with junk from our dependencies when running pytest directly
- [x] Logs are visible and appropriately formatted and unpolluted when running stuff via a CLI
- [x] Logs are visible and appropriately formatted and unpolluted when running stuff via Make

zaneselvans (approved):
Well, all the things that I checked seemed to behave as desired.

zaneselvans:
I'm going to go ahead and merge this in the hopes that we can get a good build tonight and be able to run some usable branch deployments this evening...

PR #4561, opened 2025-08-22
Update docs/release notes to reflect new Kaggle setup.

### LINKED ISSUE #4381: Improve usability of Kaggle notebooks / datasets

zaneselvans:
# Overview

- Update our docs & release notes to reflect the new direct-to-s3 setup on Kaggle.

Closes #4381


docs/data_access.rst
```diff
@@ -198,12 +198,18 @@ Kaggle
...
-outputs automatically update `the PUDL Project Dataset on Kaggle
-<https://www.kaggle.com/datasets/catalystcooperative/pudl-project>`__ once a week. There
-are `several notebooks
-<https://www.kaggle.com/datasets/catalystcooperative/pudl-project/code>`__ associated
-with the dataset, both curated by Catalyst and contributed by other Kaggle users.
+available PUDL data without needing to do any environment setup?  We provide several
+`example notebooks on Kaggle <https://www.kaggle.com/catalystcooperative/code>`__.
+(they are also pushed to our `PUDL Examples Repo on GitHub
+<https://github.com/catalyst-cooperative/pudl-examples>`__). These notebooks pull data
+directly from the Parquet outputs in S3 (see below) and so can also be used locally if
```
jdangerx, line 205: Nice, hadn't thought about that but we are reducing the Kaggle vendor lock-in. :muscle: :boom: :chains:

zaneselvans: Yeah, after messing around with Marimo & Polars a little this weekend I'm excited to try and set up some [WASM notebooks](https://molab.marimo.io/blog/wasm-github-pages) on GitHub Pages that also pull data directly from S3.

zaneselvans:
Oooh it did the fast merge thing correctly!

PR #4562, opened 2025-08-22
Sec10k docs

zaneselvans:
# Overview

Closes #XXXX.

## Documentation

# Testing

## To-do list


src/pudl/metadata/sources.py
```diff
@@ -851,19 +851,9 @@
...
-unstructured Exhibit 21 attachment, which describes the ownership relationships between
-the parent company and its subsidiaries. This data is used to create a linkage between
-EIA utilities and SEC reporting companies, to better understand the relationships
-between utilities and their affiliates, and the resulting economic and political impacts.
-
-This data was originally downloaded from the SEC and processed using a machine learning
-pipeline found here: https://github.com/catalyst-cooperative/mozilla-sec-eia"""
+            "The SEC Form 10-K is an annual report required by the U.S. Securities and "
+            "Exchange Commission (SEC), that gives a comprehensive summary of a "
+            "company's financial performance."
```
zaneselvans, line 856: The newlines in here were causing formatting issues in the RST Jinja template so I shortened it, since there's more information elsewhere now.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+unstructured PDFs or HTML.
+
+Our main interest in the SEC Form 10-K is providing better access to some of that
+unstructured data, primarily an attachment called Exhibit 21 which reports "Subsidiaries
+of the Registrant" as described in `17 CFR § 229.601 <https://www.law.cornell.edu/cfr/text/17/229.601>`__.
+
+List all subsidiaries of the registrant, the state or other jurisdiction of
+incorporation or organization of each, and the names under which such subsidiaries do
+business. This list may be incorporated by reference from a document which includes a
+complete and accurate list.
```
krivard: Is this meant to be a block quote?

```suggestion
  List all subsidiaries of the registrant, the state or other jurisdiction of
  incorporation or organization of each, and the names under which such subsidiaries do
  business. This list may be incorporated by reference from a document which includes a
  complete and accurate list.
```

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+individual utilities.
+
+In 2024 we captured a snapshot of the raw HTML of the SEC 10-K filings, and any
+associated Exhibit 21 attachments and processed it using the machine learning pipeline
+found in `this repository <https://github.com/catalyst-cooperative/mozilla-sec-eia>`__.
+
+Metadata related to each filing was parsed out of the plaintext headers of the HTML
+documents. Some of these headers pertain to the filing as a whole and others describe
+attributes of the individual companies associated with the filing. Each filing has a
+single company that is the primary filer and the filing is associated with their
```
krivard:
```suggestion
single company that is the primary filer, and the filing is associated with their
```

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+`Central Index Key <https://www.sec.gov/search-filings/cik-lookup>`__ (CIK) -- a
+persistent company identifier assigned by the SEC that is more durable and standardized
+than the company name.
+
+These plaintext headers are not necessarily intended to be machine readable, but they
+are highly structured and allow us to compile a database of all SEC 10-K filings and the
+companies involved. They are the source material for most of the ``core_sec10k`` tables
+found in PUDL, with the exception of the data extracted from the unstructured Exhibit 21
+itself.
```
krivard: > They are the source material for most of the ``core_sec10k`` tables
> found in PUDL, with the exception of the data extracted from the unstructured Exhibit 21
> itself.

It looks like you're using a "most Q have X property but some Q have Y property" pattern but switching to a different Q half way through, so it's not clear what you mean.

So probably not:

* Q = `core_sec10k` tables, some of which were extracted from Exhibit 21,
 X = sourced from the database we compiled,
 Y = sourced from somewhere else unnamed

But maybe:

* Q = `core_sec10k` tables, 
 X = sourced from the database we compiled,
 Y = sourced from Exhibit 21
* Q = data in `core_sec10k` tables
 X = sourced from the database we compiled,
 Y = sourced from Exhibit 21
* Q = source material for the `core_sec10k` tables,
 X = extracted from structured plaintext headers,
 Y = extracted from unstructured Exhibit 21 attachments

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+  Subsidiaries to SEC 10-K Filers. In cases where the SEC 10-K filer has been linked
+  to an EIA utility, this also results in a link between the subsidiary and the EIA
+  utility.
+
+For more details on how these record linkages are done, and how the Exhibit 21 data
+was extracted see :doc:`/methodology/sec10k_modeling`
+
+Linkage Completeness & Quality
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
krivard: Maybe an introductory paragraph here before jumping into an example?

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+          "parent_company_name",
+          "subsidiary_company_id_sec10k",
+          "subsidiary_company_central_index_key",
+          "subsidiary_company_utility_id_eia",
+          "subsidiary_company_name",
+          "fraction_owned",
+      ],
+      dtype_backend="pyarrow"
+      ).query("parent_company_central_index_key == '0000072903'")
+      ).query("report_date.dt.year.isin([2001, 2023]")
```
krivard:
```suggestion
      ).query("report_date.dt.year.isin([2001, 2023])")
```

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+* Based on company names, addresses, and other information we use
+  `Splink <https://github.com/moj-analytical-services/splink>`__ to do a statistical
+  record linkage between the universe of SEC 10-K Filers and EIA Utilities.
+* Based just on standardized company names and locations, we try to link Exhibit 21
+  Subsidiaries to SEC 10-K Filers. In cases where the SEC 10-K filer has been linked
+  to an EIA utility, this also results in a link between the subsidiary and the EIA
+  utility.
+
+For more details on how these record linkages are done, and how the Exhibit 21 data
+was extracted see :doc:`/methodology/sec10k_modeling`
```
krivard:
```suggestion
For more details on how these record linkages are done and how the Exhibit 21 data
was extracted, see :doc:`/methodology/sec10k_modeling`.
```

krivard: or if the record linkages are complete at this time / no longer being updated,

```suggestion
For more details on how these record linkages were done and how the Exhibit 21 data
was extracted, see :doc:`/methodology/sec10k_modeling`.
```

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+          ]
+      ).filter(
+          pl.col("report_date").dt.year().is_in([2001, 2023]),
+          pl.col("parent_company_central_index_key").eq("0000072903")
+      )
+
+   .. csv-table::
+      :file: sec10k/xcel_sec10k_example.csv
+      :header-rows: 1
+      :widths: auto
```
krivard: Three pages of tabular data is too much at once -- I recommend splitting this into smaller chunks relevant to the different effects you want to describe, and showing each chunk near the text that discusses that effect.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+
+* The parent company we've selected, Xcel Energy, is linked to a corresponding EIA
+  Utility ID.
+* In 2023, 4 of 18 listed Xcel subsidiaries are associated with their own SEC 10-K
+  filer CIK. In 2001, it was 3 out of 23.
+* Across both years, only 1 subsidiary company was linked to an EIA Utility ID.
+* 2 subsidiaries that were linked to CIKs in 2023 also appear in 2001, but they were
+  not linked to EIA Utilities, apparently due differences in how their names are
+  spelled.
```
krivard:
```suggestion
* 2 subsidiaries that were linked to CIKs in 2023 also appear in 2001, but they were
  not linked to EIA Utilities, apparently due to differences in how their names are
  spelled.
```

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+Looking at just companies with industry codes that suggest involvement in electricity
+generation ("electric services" and "electric & other services combined") we find that
+50-60% of companies have been consistently associated with EIA Utility IDs across the
+full range of years. "cogeneration services & small power producers" and "paperboard
+mills" (which frequently engage in cogeneration) also average more than 50% linkage to
+EIA. Inspecting the electric services companies that have not been linked to an EIA
+Utility ID, we find many familiar utilities that ideally would be matched, so be aware
+that the current linkage is incomplete.
+
+* What fraction of all EIA Utility IDs in EIA-923 get linked to an SEC 10-K filer?
```
krivard: (assuming this was the todo for the immediately preceding paragraph)
```suggestion
```

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+* :download:`{{ filename.stem.replace("_", " ").title() }} ({{ filename.suffix.replace('.', '').upper() }}) <{{ filename }}>`
+{% endfor %}
+{% endblock %}
+
+{% block availability %}
+
+The SEC filing database is called EDGAR. The agency provides an `online search interface
+<https://www.sec.gov/search-filings>`__ for all filings as well as `a REST API
+<https://www.sec.gov/search-filings/edgar-application-programming-interfaces>`__ and
+`several RSS feeds <https://www.sec.gov/about/rss-feeds>`__.
```
krivard: It's not clear how EDGAR is related to what data is available through PUDL. Does this text belong in a different block?

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+than the company name.
+
+These plaintext headers are not necessarily intended to be machine readable, but they
+are highly structured and allow us to compile a database of all SEC 10-K filings and the
+companies involved. They are the source material for most of the ``core_sec10k`` tables
+found in PUDL, with the exception of the data extracted from the unstructured Exhibit 21
+itself.
+
+Three Sets of Companies and Two Record Linkages
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
krivard: Using `~` here has caused later subsections of Background to appear as subsubsections of "What can't this data be used for?":

> image

Two options:

* Switch to "---" to place "Three Sets of Companies" & friends on an equal level with "Download Additional Documentation"
* Add some kind of grouping subsection ("Data Complexity"? "Challenges"? "Additional Context"?) with ---  so that these ~~~ sections can be subsubsections without disturbing the level of "Download Additional Documentation" & friends.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+  of matching, first between the subsidiary name/location and the database of SEC 10-K
+  filers, and second between the SEC 10-K filers and the EIA Utilities.
+
+{% endblock %}
+
+{% block download_docs %}
+{% for filename in download_paths -%}
+* :download:`{{ filename.stem.replace("_", " ").title() }} ({{ filename.suffix.replace('.', '').upper() }}) <{{ filename }}>`
+{% endfor %}
+{% endblock %}
```
krivard, line 136: Looks like there aren't any documents yet -- if you want to suppress that section entirely, you can do

```suggestion
{% block download_title %}
{% endblock %}
{% block download_docs %}
{% endblock %}
```

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,284 @@
...
+This data is relevant to the US energy system and the transition away from fossil fuels
+because electricity and natural gas utility holding companies often have complex
+ownership relationships that tie together the economic and political interests of many
+nested parent and subsidiary companies. Without understanding that network of
+relationships, it is hard to understand the incentives driving the behavior of
+individual utilities.
+
+In 2024 we captured a snapshot of the raw HTML of the SEC 10-K filings, and any
+associated Exhibit 21 attachments and processed it using the machine learning pipeline
+found in `this repository <https://github.com/catalyst-cooperative/mozilla-sec-eia>`__.
```
krivard: i have a small allergy to link anchor text of the "click here"/"this page" variety

```suggestion
In 2024 we captured a snapshot of the raw HTML of the SEC 10-K filings and any
associated Exhibit 21 attachments. We wrote a machine learning pipeline to process
the data to X, Y, and Z.
To make the code available publicly, we have published a `GitHub repository with
the machine learning pipeline and other processing materials 
<https://github.com/catalyst-cooperative/mozilla-sec-eia>`__.
```

docs/templates/sec10k_child.rst.jinja
zaneselvans: They're definitely not complete and there's room for improvement with more training data or enrichment and I think we want it to seem (potentially) alive so I'll make it present tense.

docs/templates/sec10k_child.rst.jinja
zaneselvans: Yeah, I know this is way too much. I kind of want to just restrict it to the companies that match to something somewhere in the timeseries and exclude all the other apparently irrelevant subsidiaries.  I also couldn't find a way to make the table more compact and dataframe like unfortunately.  I really want a better system for mixing notebook like sections into our documentation.

docs/templates/sec10k_child.rst.jinja
zaneselvans: This is a reminder to look at the data from the EIA side as well as from the SEC side. We have a bunch of utilities and other kinds of power plant owning companies reporting in the EIA-860. What fraction of *them* are being matched to an entity in the SEC 10-K? (As opposed to what fraction of the SEC 10-K "electricity" companies are getting matched to an EIA utility).

Is that a useful thing to look at? What fraction of EIA respondents would we expect to show up in the SEC 10-K, either as a filer or as a subsidiary? My intuition is that if we had perfect information it would be a large proportion, since a lot of companies that own generation facilities are either themselves large enough to need to file the SEC 10-K or are owned by a parent that is large enough (more than $10M in assets, more than 2000 shareholders).

docs/templates/sec10k_child.rst.jinja
zaneselvans: EDGAR is the original source of all the SEC 10-K data we have. I'll add a sentence indicating that.

docs/templates/sec10k_child.rst.jinja
zaneselvans: Yes! This is exactly what I want to do. I've been plagued by all these empty sections.

docs/templates/sec10k_child.rst.jinja
zaneselvans: I'm not sure I understand the ambiguity or confusion you're describing here.

What I'm trying to communicate is that there are 2 different kinds of source data that feed from the SEC 10-K into the `core_sec10k` tables in PUDL:
* Plaintext headers which we parse into tables in a pretty straightforward way using python.
* Unstructured Exhibit 21 attachments that need to be turned into tables using the ML pipeline.

I'll re-word to try and make this clearer.

docs/templates/sec10k_child.rst.jinja
zaneselvans: Blargh, struggling with the template here.  I want to have subsections within Background. Maybe I'm just using the wrong character for the header underlines? Okay I think "---" does what we want.

docs/templates/sec10k_child.rst.jinja
zaneselvans: Reworded.

docs/templates/sec10k_child.rst.jinja
zaneselvans: Maybe I should just leave the table out entirely and link to the notebook, or leave the code snippets in if someone wants to follow along on their own.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,352 @@
...
+
+A bug in our extraction of company information from the plaintext headers of the SEC
+10-K filings meant that we were unable to associate about 5% of the reported company
+mailing addresses with the other reported company information, which may have slightly
+impacted the quality of the record linkage to EIA Utilities and leads to more null
+values in those fields than would be observed in the source data. See issue
+:issue:`4165` and PR :pr:`4134`. This bug has been fixed, and the lost addresses can
+be recovered by re-running the upstream extraction.
+
+Subisidiary company IDs lack continuity between filings
```
krivard:
```suggestion
Subsidiary company IDs lack continuity between filings
```

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,352 @@
...
+guaranteed to be assigned a different ID in each of them. Unless a subsidiary has also
+been associated with its own EIA CIK or EIA utility ID, we can't see how the same
+company is connected to multiple parents. Given how little information we have about
+each subsidiary, it will be difficult to generate authoritiative IDs that can be applied
+across filings from different parent companies.
+
+Because subsidiary names and locations may be reported differently from year to year
+even within the same parent company's filings, we can't authoritatively track changes
+in those parent-subsidiary relationships through time. By analyzing the set of
+subisidiaries that appear under a given filer each year we can probably better
```
krivard:
```suggestion
subsidiaries that appear under a given filer each year we can probably better
```

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,352 @@
...
+standardize their names and improve the temporal continuity of these IDs.
+
+The fraction of a subisidiary owned by its parent is rarely reported
```
krivard:
```suggestion
The fraction of a subsidiary owned by its parent is rarely reported
```

krivard:
Meeting notes:

generally trying to do too many things at once:

- blog post
- data source page
- methodology page
- demo notebook
- planning research pitches

Strengths of the dataset are:

+ high-precision match between eia and sec
+ if you have the CIK, we can give you info
+ if you know the parent CIK, we can give you lists of subsidiaries over time


Next steps:

* KL - support on figuring out what is methodology and what is data source
* KL - accuracy pass
* align strengths to text
* KM - sort irregularities into appropriate sections

src/pudl/output/sec10k.py
```diff
@@ -21,6 +22,111 @@ def _filename_sec10k_to_source_url(
...
+    # 2820 has one entry with garbled spelling.
+    # 1040 we just munged up above.
+    # The industry names associated with them will be overwritten by the official
+    # SEC list.
+    known_nonunique_sics = {"1040", "2820"}
+    non_unique_sics = sic_nunique_names[sic_nunique_names["industry_name_sic"] > 1][
+        "industry_id_sic"
+    ].to_list()
+    assert set(non_unique_sics) == known_nonunique_sics, (
+        f"Found non-unique mapping of SIC ID to industry name! {non_unique_sics}"
```
krivard: If this assert fails, the first thing we'll need to do is figure out which non-unique SICs were found that were not expected, so might as well do that in the error message itself

```suggestion
        f"Found non-unique mapping of SIC ID to industry name! {non_unique_sics-known_nonunique_sics}"
```

src/pudl/metadata/dfs.py
```diff
@@ -1016,3 +1016,468 @@ class ImputationReasonCodes(Enum):
...
+"""
+    ),
+    dtype="string",
+)
+"""
+A table of Standard Industrial Classification codes and descriptions used by SEC.
+
+These codes have mostly been supplanted by the NAICS codes, which provide more detail
+and modernized industrial categories, but are still in use be SEC. Crosswalks exist that
+can convert between SIC and NAICS codes, but the correspondents isn't perfect.
```
krivard:
```suggestion
and modernized industrial categories, but are still in use by SEC. Crosswalks exist that
can convert between SIC and NAICS codes, but the correspondence isn't perfect.
```

krivard (commented):
nits

docs/templates/sec10k_child.rst.jinja
zaneselvans: I remain confused as to why the typo checker is not catching these. 😢 In the config I don't see why these files would be getting excluded.

src/pudl/metadata/sources.py
```diff
@@ -851,19 +851,9 @@
         "title": "U.S. Securities and Exchange Commission Form 10-K",
```
katie-lamb: Nit but it would be nice to have "SEC" in the title to make it more searchable and have it match the other data source titles which have the dataset abbreviation. "U.S. Securities and Exchange Commission (SEC) Form 10-K" is my suggestion.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+EDGAR database, including the Exhibit 21 attachments. We extract two kinds of data from
+this raw data source using different methods.
+
+First, metadata related to each filing was parsed out of the plaintext headers of the
+HTML documents. Some of these headers pertain to the filing as a whole and others
+describe attributes of the individual companies associated with the filing. Each filing
+has a single company that is the primary filer, and the filing is associated with their
+`Central Index Key <https://www.sec.gov/search-filings/cik-lookup>`__ (CIK) -- a
+persistent company identifier assigned by the SEC that is more durable and standardized
+than the company name.
```
katie-lamb: To me this Background section feels a little too long and the level of detail makes it a bit confusing. I left some suggestions for the next three paragraphs:

"First, metadata related to each filing _and the companies associated with the filing_ was parsed out of the plaintext headers of the HTML documents. ~~Some of these headers pertain to the filing as a whole and others
describe attributes of the individual companies associated with the filing.~~ Each filing
has a single company that is the primary filer, and the filing is associated with their
`Central Index Key <https://www.sec.gov/search-filings/cik-lookup>`__ (CIK) -- a
persistent company identifier assigned by the SEC ~~that is more durable and standardized
than the company name.~~"

The sentence about the headers pertaining to the filing as a whole vs attributes of the individual filers doesn't feel necessary and was a bit confusing. It could belong in the table level doc string of the filing info table and/or the company info table?

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+relationships, it is hard to understand the incentives driving the behavior of
+individual utilities.
+
+In order to link this subsidiary company information to other datasets we also need
+contextual information about the SEC 10-K filings they are part of and the companies
+submitting those filings.
+
+In 2024 we captured a snapshot of the raw HTML of the SEC 10-K filings from the SEC's
+EDGAR database, including the Exhibit 21 attachments. We extract two kinds of data from
+this raw data source using different methods.
```
katie-lamb: Suggestion:

> We extract two kinds of data from this raw data source using different methods.

-->
"We extract two kinds of data from this raw data source: metadata about the filing companies and filing itself, and ownership data about the company's subsidiaries."

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+HTML documents. Some of these headers pertain to the filing as a whole and others
+describe attributes of the individual companies associated with the filing. Each filing
+has a single company that is the primary filer, and the filing is associated with their
+`Central Index Key <https://www.sec.gov/search-filings/cik-lookup>`__ (CIK) -- a
+persistent company identifier assigned by the SEC that is more durable and standardized
+than the company name.
+
+The plaintext headers are not necessarily intended to be machine readable, but they
+are highly structured and allow us to compile a database of all SEC 10-K filings and the
+companies involved. The ``core_sec10k`` tables derived from the headers provide the
```
katie-lamb: I think this first sentence "The plaintext headers..." could be cut and put into a table level docstring for the company info or filing info tables.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+context that's necessary to link the subsidiary company information extracted from
+Exhibit 21 to other sets of companies, including the SEC 10-K filers themselves, as well
+as companies that file the EIA Form 860.
```
katie-lamb: I think this paragraph strays away from the main focus of this section which is the two kinds of data that we extract from the raw data source. To be more focused, I think we could move this sentence on using info pulled from headers to link to EIA into the linkage section below.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+Second, we wrote `a machine learning pipeline
+<https://github.com/catalyst-cooperative/mozilla-sec-eia>`__ to extract structured
+tabular data from the unstructured Exhibit 21 attachments. This includes an ID
+indicating what filing the attachment was part of, the subisdiary company name and
```
katie-lamb: typo: subisidiary -> subsidiary

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+The same company can file its own SEC 10-K, appear as a subsidiary company in another
+company's SEC 10-K Exhibit 21, and be a utility that reports its generation to EIA. The
+challenge of integrating the SEC 10-K into PUDL is linking these three types of entities
+to each other when they refer to the same company. We attempt to make all three possible
+linkages among these entities, prioritizing precision over recall: we only include
+high-confidence links.
+
+* Based on company names, addresses, and other information we use
+  `Splink <https://github.com/moj-analytical-services/splink>`__ to do a statistical
+  record linkage between the universe of SEC 10-K Filers and EIA Utilities. This is
```
katie-lamb: nit but "probabilistic record linkage" may be more accurate than "statistical record linkage", not sure

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+  the most complete linkage, because both the SEC 10-K and EIA datasets are
+  relatively rich in details about the companies.
```
katie-lamb: I'm not sure it's right to say this is the most complete linkage. It's the linkage that has the most shared attributes to link with, but in reality it might be more complete / accurate to link on company name alone (as is done in the subsidiary to filer and EIA utility link). If the shared attributes are reported differently to each agency, then this may well be a less complete linkage. I know you have some stats to back up the relative completeness of the linkages, so maybe you feel like this is a safe enough statement.

I think we should say something like "We have the most information when doing this linkage, because both the SEC 10-K and EIA datasets have shared attributes about the companies."

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+* Based on standardized company names and incorporation locations, we try to link
+  each Exhibit 21 Subsidiary to a corresponding SEC 10-K Filer. This linkage is sparse
+  because we have so little information about the subsidiary. It is likely that additional
+  links exist, but we do not have enough evidence to confirm them.
+* Similarly, we try to link individual subsidiaries to corresponding EIA utilities
```
katie-lamb: Instead of "individual subsidiaries" I would say "Exhibit 21 subsidiaries" to use the same language that we use in the previous bullet point and avoid potential confusion.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+  based on company name. In cases where the SEC 10-K filer has been linked
+  to both an EIA utility and a subsidiary, we can also indirectly infer a link between
+  the subsidiary and the EIA utility.
```
katie-lamb: The "In cases where..." sentence feels like too much detail for this bullet and should be kept to the table level doc string for the subsidiaries to EIA association table.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+What can this data be used for?
+-------------------------------
+
+* Linking companies that file SEC 10-K to EIA Utilities. 50-60% of SEC 10-K filers
+  that identify primarily as providing electric services are matched to an EIA
+  Utility ID.
+* Tracking SEC 10-K company name and address changes and other attributes such as
+  taxpayer ID and industrial classification over time to aid in entity matching in other
+  datasets.
```
katie-lamb: Suggested rewrite:

* Tracking _changes in_ SEC 10-K company name, ~~and~~ address ~~changes~~, and other attributes such as
  taxpayer ID and industrial classification over time ~~to aid in entity matching in other
  datasets.~~

Would the changes in these attributes actually aid in entity matching to other datasets or do the attributes themselves make it possible to link to other datasets? I feel like it's implied that this is useful for entity matching based on the previous section about record linkages.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+* In cases where subsidiary companies have been successfully associated with SEC 10-K
+  filers, those subsidiaries may also be linked to EIA Utilities.
```
katie-lamb: I think in many cases it's pretty accurate to link from a subsidiary straight to an EIA utility based solely on company name. It's difficult to measure how accurate this is, and obviously we'd like to link a subsidiary to a filer CIK first, but I don't think it's accurate to say that that link to filer CIK is totally necessary in order to link to EIA.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+  links exist, but we do not have enough evidence to confirm them.
+* Similarly, we try to link individual subsidiaries to corresponding EIA utilities
+  based on company name. In cases where the SEC 10-K filer has been linked
+  to both an EIA utility and a subsidiary, we can also indirectly infer a link between
+  the subsidiary and the EIA utility.
+
+For more details on how these record linkages are done and how the Exhibit 21 data is
+extracted, see :doc:`/methodology/sec10k_modeling`.
+
+What can this data be used for?
```
katie-lamb, line 92: I would add:

* Tracking changes in an owner company's set of subsidiary companies over time
* Looking up a company name and tracking changes in that company's ownership over time and across multiple owner companies. We don't have unique IDs to track subsidiaries to over time and across ownership, so look up is only possible by the company name.

For example, for Josh Basseches' book fact checking work I used the extracted exhibit 21 data to look at what year a subsidiary was no longer under the ownership of a parent company. I also searched by subsidiary company name and could see all the owner companies that have listed that company name in its exhibit 21. This is difficult to do without bulk access to Ex. 21 data. Obviously it would be better to look up a subsidiary company by ID, but it worked well enough to look up by company name.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+---------------------------------
+
+We encountered substantial challenges in working with this dataset, and many use cases
+we had hoped to support are not possible or severely limited. See the Irregularities
+section below for a full listing, but notably, this dataset is not appropriate for:
+
+* Quantitatively tracking how ownership of subsidiary companies has changed over time.
+  This is not possible because subsidiary company IDs are specific to each parent
+  company, and also change over time. In addition, ownership percentages are only rarely
+  reported.
```
katie-lamb: See note above in "What can this data be used for".

Maybe this could be changed to "Deterministically tracking how ownership of subsidiary companies has changed over time using an ID to uniquely identify a subsidiary over time". Also, I think the sentence about ownership percentages should be broken out into a different bullet for clarity.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+* Confidently linking subsidiary companies to EIA Utilities. This linkage requires
+  either a perfect name match between the subsidiary and the EIA Utility, or two layers
+  of matching, first between the subsidiary name/location and the database of SEC 10-K
+  filers, and second between the SEC 10-K filers and the EIA Utilities.
```
katie-lamb, line 129: Note that there is some string cleaning/standardization done to both EIA and SEC subsidiary names before performing the merge, so it's not actually a perfect name match that's necessary.

I guess if we're going to call out this match then I would argue for just saying you can't treat any of the matches as deterministic. Do we have metrics to back up this particular match being worse than say subsidiaries to filers?

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+inherent in the dataset.
+
+Our SEC 10-K dataset is not currently being updated
+---------------------------------------------------
+
+Unlike almost all of the the other datasets that feed into PUDL, we are not currently
+updating the underlying SEC 10-K data and generating new outputs on a regular basis. We
+do not currently have any dedicated funding for its ongoing maintenance, and the ML
+pipeline that extracts the Exhibit 21 attachment data is outside the scope of the main
+PUDL data processing pipeline.
```
katie-lamb: I would simplify this sentence to:

We do not currently have any dedicated funding for its ongoing maintenance and update.

I think talking about the ML pipeline being out of scope makes it sound like we won't ever touch it again, which we would be able to if we had additional resources.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+The SEC 10-K filer to EIA utility linkage is based only on 2023 data
+--------------------------------------------------------------------
+
+Due to memory constraints, the statistical record linkage used to associate the SEC's
+``central_index_key`` with the EIA's ``utility_id_eia`` was only run on 2023 data.
+This means companies that do not appear in the 2023 data have not been linked. With
+improved blocking by year or the use of deduplicated entity tables, we should be able
+to re-run the linkage across all years of data.
```
katie-lamb: I would simplify this sentence to:

With additional funding, we should be able to re-run the linkage across all years of data.

It seems like too much info to introduce  "improved blocking" and "entity tables" without context. Also I believe it was run on the EIA entity table.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+Only 14% of EIA utilities are linked to an SEC filer or subsidiary
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Ideally, we would be able to link most companies with EIA ids to either a parent or a
+subsidiary company listed in the SEC 10-K.
+
+As of PUDL release v2025.9.1, only 2,308 out of 16,636 utilities with EIA ids have been
+identified in the SEC data; just under 14%. While these are all high-confidence matches,
+the link rate may not be enough for many desired use cases.
```
katie-lamb: I would take this part of the sentence out and say instead:

"While these are all high-confidence matches, there is not high coverage in this connection between SEC and EIA."

"link rate" is confusing to me.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+   * - EIA utility IDs
+     - Count
+   * - All
+     - 16,636
+   * - Mapped to an SEC 10-K filer central index key
+     - 529
+   * - Mapped to an SEC 10-K subsidiary
+     - 1,940
+
+Maps to an SEC 10-K filer central index key are superior, since the data we have about filers
```
katie-lamb: Suggested rewrite:

"Connections to an SEC 10-K filer are more trustworthy, since the data we have about filers is significantly richer than the data we have about subsidiaries who don't file."

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: auto
+
+   * - EIA utility IDs
+     - Count
+   * - All
+     - 16,636
+   * - Mapped to an SEC 10-K filer central index key
```
katie-lamb: --> "Mapped to an SEC 10-K filer (with CIK)"

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+Maps to an SEC 10-K filer central index key are superior, since the data we have about filers
+is significantly richer than the data we have about subsidiaries.
+
+
+Only half of SEC filers providing electric services are linked to EIA utilities
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Ideally, we would be able to link most SEC companies that are clearly utilities to their
+associated EIA utility id.
```
katie-lamb: --> "Inversely, we would also like to be able to link most SEC companies that are clearly utilities to their associated EIA utility ID."

I think we want to draw a parallel between this section and the one above, as I found I had to reread a few times to understand the difference.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+utilities.
+
+If we want to know how many utility-like SEC companies have a match, we can approach it from
+two angles:
+
+* Do most matches occur in utility-like industries? --> What are the most common SIC codes among matches?
+* Do most utility-like industries have good match coverage? --> What SIC codes have the highest match rates?
+
+As of PUDL release v2025.9.1, only 8,993 out of the 15,178 SEC filers we can match to EIA
+utility ids list themselves as electric services under SIC 4911 or 4931; just under 60%.
```
katie-lamb: Found this sentence confusing, suggested rewrite:

"As of PUDL release v2025.9.1, only 8,993 out of the 15,178 SEC filers ~~we can~~ _matched_ to EIA
utility ids list themselves as electric services ~~under~~ _(SIC 4911 or 4931)_; just under 60%."

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+.. list-table:: Top 10 SICs by frequency within EIA-matched SEC filers
+   :header-rows: 1
+   :widths: auto
+
+   * - Percent of EIA-matched SEC filers
+     - Standard Industrial Code (SIC)
+     - Percent of SIC filers with an EIA match
+   * - 46%
+     - 4911
```
katie-lamb: I would add "(electrical services)" after 4911

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+     - 58%
+   * - 14%
+     - 4931
```
katie-lamb: I would add "(electrical services)" after 4931

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+A very long tail of low-incidence SIC codes covers the remaining 40% of the matches; see
+some of the more common ones in the table below.
+
+.. list-table:: Top 10 SICs by frequency within EIA-matched SEC filers
+   :header-rows: 1
+   :widths: auto
+
+   * - Percent of EIA-matched SEC filers
+     - Standard Industrial Code (SIC)
+     - Percent of SIC filers with an EIA match
```
katie-lamb: I'm having a hard time conceptualizing the difference between % of EIA-matched SEC filers and % of SIC filers with an EIA match in this table. Maybe it would be more clear if SIC code was the furthest left column?

It might be that "SIC filers" looks too much like a typo for "SEC filers". Is an SIC filer a filing company that reports that SIC code? What about "SIC reporters"? Change to "% of SIC-code reporting companies with an EIA match" or something? 

edit: oh i see that there's an explanation of this column below the table. maybe that explanation should be moved up?

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+   * - 0.82%
+     - 4961
+     - 7.9%
+
+As of PUDL release v2025.9.1, only 8,993 out of the 15,845 SEC filers who file as
+electric services under SIC 4911 ("electric services") or 4931 ("electric & other
+services combined"; often joint electricity & natural gas utilities) have matches to EIA
+utilities; just under 57%. Inspection of the other SEC 10-K filers with these SIC codes
+suggests that the vast majority of them ought to be EIA 860/923 respondents, but the
+system was not able to find enough evidence to make a link. With further refinement the
```
katie-lamb: > but the system was not able to find enough evidence to make a link

I'm finding this nebulous / confusing. What about "but the record linkage model didn't assign a high enough probability for these matches to make a link."

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+match rate in these SIC codes can likely be improved, given the richeness of both the
+EIA and SEC filer datasets. Manually labeled training data, use of year-specific company
+names, and enrichment of both datasets with additional company addresses or other
+attributes may all help.
```
katie-lamb: I would cut this last sentence or move it to the future improvements section of the methodology page.

We do use a small set of manually labeled data for training and validation of this match, and the company names often show no change between years once they've been standardized/normalized. Really the change in address attributes over in time may be what changes the link across report years. I think most likely improvements would come from model parameter tweaking or moving the match threshold for higher recall vs precision. I guess it would be nice if we could bring in additional addresses, but I'm not sure what data that would come from. There is a attribute selection process that happens before the actual link, and it could likely be improved, but it's not as low hanging fruit as the model parameter tweaking process.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+Other SIC categories where we have relatively high match rates are in the table below.
+
+.. list-table:: Top 10 SICs by match rate
+   :header-rows: 1
```
katie-lamb, line 378: I'm having a hard time understanding what the difference is between this table and the one above it. I guess the denominator is different? 

I think I'd suggest cutting this table entirely or adding further clarification on what "match rates" means here.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+mailing addresses with the other reported company information, which may have slightly
+impacted the quality of the record linkage to EIA Utilities and leads to more null
+values in those fields than would be observed in the source data. See issue
+:issue:`4165` and PR :pr:`4134`. This bug has been fixed, and the lost addresses can
+be recovered by re-running the upstream extraction.
+
+Identification of subsidiary companies is extremely sparse
+----------------------------------------------------------
+
+As of PUDL release v2025.9.1, less than 2% of all subsidiary companies have a match to
```
katie-lamb: Note that this 2% stat is using the total pool of subsidiary company names extracted from Ex. 21 docs as the denominator, which doesn't actually represent the total number of subsidiary companies. If we could use a process to unify subsidiaries across time with an ID, then more would be linked to CIKs.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+an SEC 10-K filer. The match rate varies by year, and stabilizes after 2012 at around
```
katie-lamb: I think adding a time component to this section makes it more confusing. The first sentence is looking at total match rate with all subsidiaries and second is match rate per year? Also it's not entirely accurate to look at post 2012 because we don't have 2018-2022 data. I would suggest taking this sentence about per year match rate out.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+1%. This is not necessarily bad -- many subsidiary companies may not be large enough to
+be required to file a 10-K -- but spot checks have shown many instances where a match
+missed by the system can easily be found by hand. Again, the system retains only
+high-confidence matches, so minor variations can cause a match to be excluded from the
```
katie-lamb: Instead of "the system" I would say "matches missed in _our automated connection_ can easily be found by hand. _Our connection only keeps_ high-confidence matches, so minor variations _in company name_ can cause a match ..."

"The system" is really just string standardization + a merge

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+The match rate is significantly better, but still low, within the electricity services
+Standard Industrial Classification (SIC) codes 4911 and 4931: 12% overall, and steadily
+increasing from under 4% in 1994 to nearly 20% in 2023.
+
+Subsidiary company IDs lack continuity between filings
+-------------------------------------------------------
+
+No meaningful company ID is reported for subsidiaries that appear in the Exhibit 21
+attachments. We construct an ad-hoc ID by concatenating the Central Index Key (CIK) of
+the main filer with the name and location of the subsidiary company as observed in the
```
katie-lamb: I would say:

"of the parent company" or maybe "of the parent company who is filing" instead of "main filer". I'm finding "main filer" to be confusing since we've only mentioned one or two times previously in this data source page that there is more than one company in a filing. Also, we've made the assumption (and it seems backed up by SEC filing language) that the main filer is the owner company whose subsidiaries are reported in the attached Ex. 21

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+The match rate is significantly better, but still low, within the electricity services
+Standard Industrial Classification (SIC) codes 4911 and 4931: 12% overall, and steadily
+increasing from under 4% in 1994 to nearly 20% in 2023.
+
+Subsidiary company IDs lack continuity between filings
+-------------------------------------------------------
+
+No meaningful company ID is reported for subsidiaries that appear in the Exhibit 21
+attachments. We construct an ad-hoc ID by concatenating the Central Index Key (CIK) of
```
katie-lamb: Maybe:

"We construct an ad-hoc ID for each subsidiary company by..."

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+the set of subsidiaries that appear under a given filer each year.
+
+The fraction of a subsidiary owned by its parent is rarely reported
+--------------------------------------------------------------------
+
+In most cases, no ownership fraction is reported in the Exhibit 21, so we cannot state
+confidently how much of each subsidiary the parent company owns. Anecdotally, based on
+outside information, many of the subsidiaries that do not report an ownership fraction
+seem to be entirely owned by a single parent company, but we don't know how common that
+really is.
```
katie-lamb: It's extremely common for an Ex. 21 to say at the bottom "all subsidiaries are wholly owned" or something like that. So I wouldn't say this is "outside information". Maybe simplify this to say:

"When ownership fractions are not included, it's very common for an Ex. 21 to include that all subsidiaries are wholly owned by the parent, but our existing model does not detect when this information is included."

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+
+Industry classifications applied to companies are not fully standardized
+------------------------------------------------------------------------
+
+A couple of the most useful fields in the SEC 10-K company information are
+``industry_id_sic`` and ``industry_name_sic`` which indicates what sector the company
+is part of. However, both columns contain significant numbers of null values and the
+name field is not entirely standardized. Further cleaning of these columns is needed to
```
katie-lamb: This high null percentage is really because of a lack of reporting in the raw data right? Or are we not correctly grabbing this from headers?

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,420 @@
...
+farther. EDGAR is the original source of all of the SEC 10-K data in PUDL.
+
+The original data is a collection of HTML documents with custom plaintext SEC headers.
+The PUDL output tables contain a field ``source_url`` which points to these original
+documents as plain text on the SEC website. `See this filing as an example
+<https://www.sec.gov/Archives/edgar/data/1058624/0001104659-09-021814.txt>`__.
+
+{% endblock %}
+
+{% block notable_irregularities %}
```
katie-lamb, line 175: I'm finding this section a little overwhelming and hard to parse through. Can the irregularities that pertain to record linkage be put in a different section specific to record linkage? That would be:

- The SEC 10-K filer to EIA utility linkage is based only on 2023 data
- Matches between EIA utility ids and SEC 10-K energy companies are sparse (and everything under this)

katie-lamb (changes requested):
I left some edits and suggestions on the data source docs page. I also will add a section to the methodology page and irregularities section of the data source page on how we don't parse through nested layers of subsidiaries.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,431 @@
...
+We don't attempt to capture nested subsidiary relationships
+-----------------------------------------------------------
+
+Exhibit 21 attachments often present subsidiary relationships in a nested format,
+using indentation to indicate ownership hierarchies. Our dataset does not attempt to
+capture this nested structure directly. Instead, if an Exhibit 21 lists multiple
+layers of subsidiaries, we extract all subsidiaries as direct children of the parent
+company that filed the Exhibit 21. In practice, many of the intermediate subsidiaries
+also file their own 10-K reports, where they list their own subsidiaries, and which
+can be used to infer the full ownership hierarchy.
```
katie-lamb, line 208: I added this section to Irregularties. Does it make sense?

docs/templates/sec10k_child.rst.jinja
zaneselvans: I felt like it was important to communicate that much of what we are doing with this data source is compiling a database of SEC 10-K filers & filings, and doing it in a kind of opportunistic way -- by parsing these headers, rather than pulling from some canonical API or existing SEC database that provides the company information. I think this becomes important later on when trying to interpret the company information table, because it helps explain why there are multiple not necessarily identical entries for the same companies in association with many different filings.

docs/templates/sec10k_child.rst.jinja
zaneselvans: Does anybody understand why our typo linter isn't catching these spelling errors?

docs/templates/sec10k_child.rst.jinja
zaneselvans: I think I was trying to communication 2 things here:

* This linkage is the only one where there's really rich information to work with, such that there's the opportunity to tweak and improve it, or enrich it by merging in other associated information from other sources.
* As a result of that, it also happens to result in a higher proportion of the entities being linked (around half of the "electricity" companies which we'd expect to have connections to the EIA utilities). Though I agree that didn't necessarily have to end up being the case.

docs/templates/sec10k_child.rst.jinja
zaneselvans: On both of those potential use cases I think we need to make sure we add a caveat that the names can change over time and also even when the names don't change the exact string that's being reported can vary, so if you try and do this programmatically, you will miss some parent-subsidary relationships, and so doing a visual inspection of the subsidiary names is important.  Most of the utility examples I looked at exhibited this behavior (e.g. Xcel Energy, Berkshire Hathaway Energy)

docs/templates/sec10k_child.rst.jinja
zaneselvans: What I was trying to communicate here is that one might naively treat these company attributes as static, and compile a canonical set of company attributes based on the most recent filing, but over decades many of them actually change, so if you try and do a record linkage with any static set of company attributes, against a larger database that includes the more diverse historical set of attributes, you'll probably miss a lot of matches. Having the time varying attributes means you can match more valid entities, and potentially block on year/date to reduce the number of entities that need to be considered in the matching, while still ensuring that the most likely to match attributes are present within the block.

docs/templates/sec10k_child.rst.jinja
zaneselvans: What I was trying to convey here is that there's the potential for a transitive linkage to EIA.  If the subsidiary didn't match to EIA based on name, but we did match the subsidiary to a CIK, and we matched the company with the CIK to EIA based on the larger collection of attributes available (even if the names aren't exactly the same) then we have also effectively associated an EIA utility with the subsidary indirectly.

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,431 @@
...
+
+Ideally, we would be able to link most SEC companies that are clearly utilities to their
+associated EIA utility id.
+
+This is a little trickier to evaluate. SEC filers use a four-digit Standard Industrial
+Classification (SIC) code to identify the company's primary industry, but not all utilities
+file under the SICs for electric services (4911 and 4931), and not all companies that
+file as electric services are utilities. We see SICs as unexpected as computer storage
+devices, nursing facilities, and real estate among our -- high-confidence -- matches to EIA
+utilities.
```
zaneselvans: An intuitive expectation that I have based on what I know about the SEC 10-K filers and browsing through the companies listed under SICs 4911 and 4931 is that conceptually, a very high proportion of the filers in those industries should really show up in the set of EIA utilities.  I think to a much higher degree than I would expect that the EIA utilities would necessarily show up as SEC 10-K filers, since there are so many smaller special-purpose companies that exist only to own a single generator or slices of some plant, or to act as an operator without owning the underlying asset that they're operating.

Meaning, if a company is explicitly in the electricity services business, AND it is big enough to be filing the SEC 10-K, then I would be pretty surprised if it doesn't also show up in the list of EIA utilities.  Maybe there's some name or other data mismatch that's keeping us from identifying it correctly, but off the top of my head I'm not sure what the common case would be where the electricity company is an SEC 10-K filer and also doesn't report to EIA.

docs/templates/sec10k_child.rst.jinja
zaneselvans: If there's horizontal space, having the names as well as the codes in the table would help make it more conceptually legible. Agree with Katie that putting the code on the left hand side like an index column would also help.

docs/templates/sec10k_child.rst.jinja
zaneselvans: Is it not just a typo?

docs/templates/sec10k_child.rst.jinja
zaneselvans: I haven't gone back to look at the raw headers to see if we're somehow not extracting data that's there. But I wouldn't be surprised at all if there are just a lot of truly missing values.

In the code changes in this PR I did go ahead and standardize the names (based on canonical names from SEC) and filled in missing names when there was a code, and foreward/backward filled gaps where the values before and after the gap were the same.  But there are still a lot of missing values because there were significant numbers of companies that had never reported a SIC code.

So the bit in this paragraph about needing to do further cleaning / standardization can probably be removed.

docs/templates/sec10k_child.rst.jinja
katie-lamb: I thought that it was a typo at first, but then was very confused about the difference between the columns: "Percent of EIA-matched SEC filers" and "Percent of SEC filers with an EIA match". Did you make this table?

docs/templates/sec10k_child.rst.jinja
krivard: Suggestive evidence:

* 2% of parent company CIKs have a non-null EIA utility ID
* 0.4% of subsidiary company SEC10k IDs have a non-null EIA utility ID

But not conclusive -- for that we'd need to know ground truth of how many parent CIKs were EIA utility companies, and how many subsidiary SEC10k IDs were utility companies.

What is our prior here -- do we expect the proportion of 10k parent companies that are utilities to be about the same as the proportion of subsidiary companies that are utilities? if we expect the proportions to be different, is it plausible for it to be 5x more likely that a parent company is a utility than for a subsidiary company to be a utility?

<details><summary>computing the above figures</summary>
```
import pandas as pd

parsubs = pd.read_parquet("https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/out_sec10k__parents_and_subsidiaries.parquet")

(
    parsubs
    .loc[:,["parent_company_central_index_key","parent_company_utility_id_eia",]]
    .drop_duplicates()
    .parent_company_utility_id_eia.notna().mean()
)

(
    parsubs
    .loc[:, ["subsidiary_company_id_sec10k","subsidiary_company_utility_id_eia",]]
    .drop_duplicates()
    .subsidiary_company_utility_id_eia.notna().mean()
)
```
</details>

docs/templates/sec10k_child.rst.jinja
krivard: I definitely struggled to establish clear but concise vocabulary here.

What I mean is to highlight the differences between and independent usefulness of:

* How many SEC filers could we match to an EIA utility id? of those, how many used this Standard Industrial Code?
* How many SEC filers used this Standard Industrial Code? of those, how many could we match to an EIA utility id?

docs/templates/sec10k_child.rst.jinja
krivard: Subsidiary companies to EIA utilities: 0.4%

Subsidiary companies to CIKs: 1.8%

<details><summary>supporting query</summary>
```
import pandas as pd

parsubs = pd.read_parquet("https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/out_sec10k__parents_and_subsidiaries.parquet")

(
    parsubs
    .loc[:, ["subsidiary_company_id_sec10k","subsidiary_company_utility_id_eia",]]
    .drop_duplicates()
    .subsidiary_company_utility_id_eia.notna().mean()
)

(
    parsubs
    .loc[:, ["subsidiary_company_id_sec10k","subsidiary_company_central_index_key",]]
    .drop_duplicates()
    .subsidiary_company_central_index_key.notna().mean()
)
```
</details>

docs/templates/sec10k_child.rst.jinja
krivard: This is the plot I'm trying to summarize here:

image

There's a polars version in the data review notebook, or a pandas equivalent below.

The transition is abrupt and feels important to disclose in some way. Oddly, the 2018-2022 gap doesn't show up at all.

Proposed new text:

> As of PUDL release v2025.9.1, less than 2% of all subsidiary records have a match to
an SEC 10-K filer. The percentage of subsidiary records that have matches varies by year. From 1994 to 2011, this percentage ranged from 1.2% to 4.3%. From 2012 on, this percentage was much more stable, ranging only from 1.0% to 1.3%. 
>
> That the percentage is so low is not necessarily bad [...]

<details><summary>pandas source</summary>
```
import pandas as pd

parsubs = pd.read_parquet("https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/out_sec10k__parents_and_subsidiaries.parquet")

(
    parsubs
    .assign(report_year=pd.to_datetime(parsubs.report_date).dt.year)
    .groupby("report_year")
    .subsidiary_company_central_index_key
    .agg(lambda x: x.notna().mean())
).plot.bar()
```
</details>

docs/templates/sec10k_child.rst.jinja
krivard: I lean toward using vocabulary consistent with the SEC -- it would be weird for someone familiar with the SEC 10-k to be excited to use our data only to find we were using different words for central concepts

docs/templates/sec10k_child.rst.jinja
krivard: I've made an attempt at getting at both of these ideas (yes, you can look up individual companies and use your human brain to interpret name lookup results; no, you cannot expect that to work in bulk) by describing them as "case studies" in the Useful For section and "programmatic tracking" in the Not Useful For section below

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,490 @@
...
+we had hoped to support are not possible or severely limited. See the Irregularities
+section below for a full listing, but notably, this dataset is not appropriate for:
+
+* Programmatic or bulk tracking of subsidiary ownership over time. Because subsidiary
+  company IDs are specific to each parent company filing, it is not possible to link
+  subsidiary records across name changes or variations in spelling between filings,
+  in aggregate.
+* Tracking ownership percentages over time. An explicit percentage is only rarely
+  reported, and our extraction pipeline did not track unstructured statements like
+  "wholly owned TODO".
```
krivard: haha whoops
```suggestion
  "all subsidiaries are wholly owned."
```

docs/templates/sec10k_child.rst.jinja
krivard: Reorganized as suggested (with some minor exceptions for things below Matches that didn't seem to be about linkages)

docs/templates/sec10k_child.rst.jinja
krivard: Reworked descriptions and table layout; see if that is better?

docs/templates/sec10k_child.rst.jinja
```diff
@@ -0,0 +1,490 @@
...
+
+A bug in our extraction of company information from the plaintext headers of the SEC
+10-K filings meant that we were unable to associate about 5% of the reported company
+mailing addresses with the other reported company information, which may have slightly
+impacted the quality of the record linkage to EIA Utilities and leads to more null
+values in those fields than would be observed in the source data. See issue
+:issue:`4165` and PR :pr:`4134`. This bug has been fixed, and the lost addresses can
+be recovered by re-running the upstream extraction.
+
+Industry classifications applied to companies have poor coverage
```
krivard, line 221: I want to collect some more numbers on this, and add a sentence or two on the actual coverage

docs/templates/sec10k_child.rst.jinja
katie-lamb: looks good

src/pudl/output/sec10k.py
```diff
@@ -21,6 +22,111 @@ def _filename_sec10k_to_source_url(
...
+    # 2820 has one entry with garbled spelling.
+    # 1040 we just munged up above.
+    # The industry names associated with them will be overwritten by the official
+    # SEC list.
+    known_nonunique_sics = {"1040", "2820"}
+    non_unique_sics = sic_nunique_names[sic_nunique_names["industry_name_sic"] > 1][
+        "industry_id_sic"
+    ].to_list()
+    assert set(non_unique_sics) == known_nonunique_sics, (
+        f"Found non-unique mapping of SIC ID to industry name! {non_unique_sics - known_nonunique_sics}"
```
krivard:
```suggestion
        f"Found non-unique mapping of SIC ID to industry name! {set(non_unique_sics) - known_nonunique_sics}"
```

src/pudl/output/sec10k.py
```diff
@@ -21,6 +22,111 @@ def _filename_sec10k_to_source_url(
...
+    )
+    # 2820 has one entry with garbled spelling.
+    # 1040 we just munged up above.
+    # The industry names associated with them will be overwritten by the official
+    # SEC list.
+    known_nonunique_sics = {"1040", "2820"}
+    non_unique_sics = sic_nunique_names[sic_nunique_names["industry_name_sic"] > 1][
+        "industry_id_sic"
+    ].to_list()
+    assert set(non_unique_sics) == known_nonunique_sics, (
```
krivard:
```suggestion
    assert set(non_unique_sics) <= known_nonunique_sics, (
```

docs/templates/sec10k_child.rst.jinja
krivard: Future work! https://github.com/catalyst-cooperative/pudl/issues/4643

PR #881, opened 2021-01-10
Adjust FERC 714 service territories by using modified versions of EIA 861 tables

ezwelty:
Implements the fixes enumerated in #618 by either modifying the utilities associated with a balancing authority for a given year ("Alter geometries") or by removing mention of a balancing authority, optionally reassigning its utilities to a balancing authority that lists it as a utility ("Treat as Utilities not Balancing Authorities").

PR #4568, opened 2025-08-26
Limit EIA-861 years in the fast ETL

### LINKED ISSUE #2628: Enable filtering by year for EIA 861 and FERC 714 ETLs
Right now the EIA 861 and FERC 714 interim ETLs process data for all years for both the full and the fast ETL. As we add more of this data into dagster, and notably as we incorporate the service territory analyses (#2437), the inability to subset this data by year is causing memory issues for the CI.

To make this change, we would need to do the following:
- [ ] For FERC 714 data: Change the extractor to read in the CSV files and subset data by years
- [ ] For EIA 861 data: Currently, discontinued tables break the single year ETL. To enable subsetting by year, we'll need to address this issue first. See #1671, #2265. We should decide at what stage we want to do the subsetting.

zaneselvans:
# Overview

For a long time we have processed *all* years of the EIA-861 data in both the full and fast ETL. Initially this was a hack to get around the fact that some of the tables / columns are discontinued and so result in entirely null columns if you just process the last year or two of data, and our old data validation tests couldn't accommodate that expectation.  But with #4105 /  #4382 that's no longer a problem.

This PR switches to processing just a couple of years of data for the EIA-861 in the Fast ETL like we do with all of the other datasets. This will speed up the fast ETL a bit, and reduce the amount of data required for the CI.

Closes #2628

## Complications

* Initially I thought we'd just need to be more specific about the data validation tests so that null columns in years where those columns weren't reported in the EIA-861 didn't cause a failure.
* However, it turned out that downstream in some of the service territory related FERC-714 outputs, we were doing some BA / Utility association filling, one year of data as a template for other years that were missing or had bad data, and the way we'd specified those fixes in the `ASSOCIATIONS` constant functionally assumed all years of data were always available.
* The way those fixes were specified was very compact, but that made the fixes and their application a little hard to read, and also hard to only conditionally apply based on the years of data being processed.
* This PR expands that data structure and reshapes it to have a set of fixes specified for each target year (year to be fixed).  It encodes exactly the same spot fixes.
* We haven't actually updated the fixes since these outputs were first written (in 2020) so ironically there are no fixes that involve the years currently being processed in the fast ETL (2020 & 2024) so the fixes are a no-op in the fast ETL at the moment.
* The new logic skips if the target year isn't one of the years being processed, and in the event that the target year IS being processed but the source year IS NOT being processed, it also skips, but logs a warning.

## Testing

- Added unit tests for the `output/ferc714.py` functions that are no longer going to be exercised in the fast ETL / CI both to maintain coverage, and so I could see that the behaved the same way before and after refactoring them.
- Added the ability to compare geospatial outputs against nightly builds (uses geopandas since polars doesn't yet support geospatial data natively)

Verified that the fast ETL / CI tests pass locally:

```console
pixi run pytest-ci
```

Ran a full ETL locally and checked outputs vs. nightly build outputs with:

```console
pixi run python devtools/check_against_nightly.py \
  "$PUDL_OUTPUT/parquet/out_ferc714__respondents_with_fips.parquet" \
  "$PUDL_OUTPUT/parquet/out_ferc714__summarized_demand.parquet" \
  "$PUDL_OUTPUT/parquet/out_ferc714__georeferenced_respondents.parquet"
```

The differences are in imputed float columns, and look minor, which is expected, since the imputation isn't 100% deterministic.

Checked that the modified dbt schemas still round-trip:

```console
pixi run pytest --no-cov tests/integration/dbt_schema_round_trip_test.py
```

## To-do list

- [x] Update the EIA-861 transforms to allow only a few years to be processed.
- [x] Update the dbt `expect_columns_not_all_null` test parameters so they don't fail on just a few years of data.
- [x] Refactor BA association table repair in the FERC-714 outputs to not rely on always having all years of data.
- [x] Add unit tests for functions from `output/ferc714.py` which will no longer run in CI.
- [x] Add GeoParquet / geopandas support to `devtools/check_against_nightly.py`
- [x] Check that refactored repair has no impact on output tables vs. nightly builds.
- [x] Move manually compiled test specifications from `schema.yml` to `schema.human.yml` for EIA-861
- [x] Update release notes
- [ ] Create follow up issues

### Follow up issues
- [ ] Persist both original (core_) and repaired (out_) service territories for EIA-861 utilities and BAs
- [ ] Build a dashboard that facilitates browsing and comparison of service territories: original vs. repaired, year X vs. year Y, BA vs. utility, this entity vs. that entity. 
- [ ] Audit original and repaired EIA-861 service territories for BAs and utilities.
- [ ] Audit reconstructed FERC-714 planning area service territories (mostly: choice of which EIA BA/Util to use)
- [ ] Update `BA_FIXES` to cover 2020-present
- [ ] Audit our service territory repair algorithm, given all of the above.


zaneselvans:
@e-belfer The part of this I was already familiar with was a quick fix, but now in the FERC-714 Outputs I see that the BA association fixes depend on always having all years of EIA-861 data available, and they feel a little inscrutable to me. I don't know if any of that context is still rattling around inside your brain but if so, do you have any thoughts on how this might be refactored to work cleanly when there's only a subset of the EIA-861 years available?

e-belfer:
> @e-belfer The part of this I was already familiar with was a quick fix, but now in the FERC-714 Outputs I see that the BA association fixes depend on always having all years of EIA-861 data available, and they feel a little inscrutable to me. I don't know if any of that context is still rattling around inside your brain but if so, do you have any thoughts on how this might be refactored to work cleanly when there's only a subset of the EIA-861 years available?

I've never looked at this part of the code before, so I am equally as lost as you are! We could probably add more conditionals in there but it'd make it even jankier, so maybe that section of the code is wanting a more thorough refactor? I don't however have an immediate insight into the best way forward.

zaneselvans:
@e-belfer Oh weird. Git was saying that all of this code came from you. But looking at the actual PR #2550 it looks like it was a mass cut-and-paste in the general Dagster re-organization.

The `ASSOCIATIONS` defining the fixes were created by Ethan, so maybe this is actually his code. So I guess we just have to figure it out. But it does seem pretty brittle / complex as it is.

src/pudl/transform/eia861.py
```diff
@@ -552,6 +552,10 @@ def _pre_process(df: pd.DataFrame) -> pd.DataFrame:
 
 def _post_process(df: pd.DataFrame) -> pd.DataFrame:
     """Post-processing applied to all EIA-861 dataframes."""
+    # If we're only processing some years of data, we may have entirely empty dataframes
+    # in the extraction phase, in which case the data_maturity field doesn't get added.
+    if "data_maturity" not in df.columns and len(df) == 0:
+        df["data_maturity"] = pd.NA
```
krivard: nit, because this function is already here, so we might as well use it: putting an extraction step in the transform module is a weird architectural choice

src/pudl/transform/eia861.py
```diff
@@ -727,10 +731,10 @@ def _tidy_class_dfs(
     # of tables.
     data_dupe_mask = data_cols.duplicated(subset=idx_cols + [class_type], keep=False)
     data_dupes = data_cols[data_dupe_mask]
-    fraction_data_dupes = len(data_dupes) / len(data_cols)
+    fraction_data_dupes = len(data_dupes) / len(data_cols) if len(data_cols) else 0
     denorm_dupe_mask = denorm_cols.duplicated(subset=idx_cols, keep=False)
     denorm_dupes = denorm_cols[denorm_dupe_mask]
-    fraction_denorm_dupes = len(denorm_dupes) / len(data_cols)
+    fraction_denorm_dupes = len(denorm_dupes) / len(data_cols) if len(data_cols) else 0
```
krivard, line 745: probably want to add the same change to line 739

src/pudl/transform/eia861.py
```diff
@@ -1476,6 +1484,9 @@ def core_demand_side_management_eia861(
         .query("utility_id_eia not in [88888]")
     )
 
+    # This happens if the extracted dataframe was empty, as is the case in the fast ETL.
+    if "data_maturity" not in transformed_dsm1.columns:
+        transformed_dsm1["data_maturity"] = pd.NA
```
krivard: Doesn't `_post_process` handle this?

src/pudl/output/ferc714.py
```diff
@@ -186,28 +186,35 @@ def filled_core_eia861__yearly_balancing_authority(
...
+    cleaned :ref:`core_eia861__yearly_balancing_authority` table, using a dictionary
+    of manual fixes. It uses the reference year as a template. The function also removes
+    balancing authorities that are manually categorized as utilities.
     """
     df = core_eia861__yearly_balancing_authority
     index = ["balancing_authority_id_eia", "report_date"]
     dfi = df.set_index(index)
     # Prepare reference rows
-    keys = [(fix["id"], pd.Timestamp(fix["from"], 1, 1)) for fix in ASSOCIATIONS]
+    eia861_years = df["report_date"].dt.year.unique()
```
krivard: Consider adding an early exit here if `eia861_years` is empty

src/pudl/output/ferc714.py
```diff
@@ -186,28 +186,35 @@ def filled_core_eia861__yearly_balancing_authority(
...
+    keys = [
+        (fix["id"], pd.Timestamp(fix["from"], 1, 1))
+        for fix in ASSOCIATIONS
+        if fix["from"] in eia861_years
+    ]
```
krivard: Consider defining a local version of ASSOCIATIONS to use both here and at line 209

src/pudl/output/ferc714.py
```diff
@@ -186,28 +186,35 @@ def filled_core_eia861__yearly_balancing_authority(
...
     # Build table of new rows
     # Insert row for each target balancing authority-year pair
     # missing from the original table, using the reference year as a template.
     rows: list[dict[str, Any]] = []
-    for ref, fix in zip(refs, ASSOCIATIONS, strict=True):
+    for ref, fix in zip(
+        refs, [fx for fx in ASSOCIATIONS if fx["from"] in eia861_years], strict=True
+    ):
         for year in range(fix["to"][0], fix["to"][1] + 1):
             key = (fix["id"], pd.Timestamp(year, 1, 1))
```
krivard: would it make sense to skip iterations here when year is not in eia861_years, instead of filtering at 216?

krivard (commented):
Mostly minor nudges. The ferc714 bits are a real trip.

It's not clear to me where `data_maturity` should get added to empty data frames, but we probably shouldn't try to do it twice?

src/pudl/transform/eia861.py
zaneselvans: Unfortunately, there's an explicit reference to the `data_maturity` field within the transform function, so it needs to happen before we get to `_post_process`.

I suspect the "right" place to ensure that data maturity is added (even in the case of empty dataframes) is in the EIA Excel extractor... Maybe I should look into that more deeply.

zaneselvans:
Anybody else looking at this -- the FERC-714 stuff doesn't actually work right now. I tried to make some little tweaks, but eventually came to understand that a bigger refactor of the spot-fix application (which deeply depends on having all years of data available) is necessary to make this work (and probably desirable just in general since it's so complex as it is)

src/pudl/transform/eia861.py
zaneselvans: I think it's okay there because it's just printing out the strings of the two numbers, not actually doing the division. And it could be nice to see the actual numbers even in this case.

src/pudl/transform/eia861.py
zaneselvans: I poked at trying to add `data_maturity` in the extract where it's being set for non-empty dataframes, and didn't see an ideal place -- since the code that does that work doesn't get run at all if the dataframe is going to end up empty. Instead I moved it into the `_pre_process()` function in this module, which means it doesn't need to be anywhere else in here. The only other changes now are to accommodate the possibility of having zero rows in division.

src/pudl/transform/eia861.py
```diff
@@ -542,7 +542,12 @@ def _pre_process(df: pd.DataFrame) -> pd.DataFrame:
       data is an early release, and we extract this information from the filenames, as
       it's uniform across the whole dataset.
     * Convert report_year column to report_date.
+    * If we've gotten an empty dataframe, make sure it has a data_maturity column.
     """
+    # If we're only processing some years of data, we may have entirely empty dataframes
+    # in the extraction phase, in which case the data_maturity field doesn't get added.
+    if df.empty:
+        df["data_maturity"] = pd.NA
```
krivard, line 552: 🎉

krivard (commented):
Nice!

e-belfer:
The `ASSOCIATIONS` method originates from https://github.com/catalyst-cooperative/pudl/issues/618, and was created in #881. I'm planning on working backwards and converting these to manual spot-fixes targeting the values we want to include, rather than back or forward filling based on different years of the tables.

tests/integration/scripts/console_scripts_test.py
```diff
@@ -32,7 +32,7 @@ def test_pudl_datastore(script_runner, command: str):
     "command,filename,expected_cols",
     [
         (
-            "pudl_service_territories --entity-type balancing_authority -y 2022 --limit-by-state --no-dissolve -o ",
+            "pudl_service_territories --entity-type balancing_authority -y 2024 --limit-by-state --no-dissolve -o ",
```
zaneselvans, line 35: Switched the arguments in here to only refer to years that are part of the fast ETL.

src/pudl/transform/eia861.py
```diff
@@ -1095,14 +1101,15 @@ def sum_numeric_values_when_strings_match(group):
...
+    # across all years (as of Aug 2025) are: BA: 1, OD: 16, Sales: 32, UD: 15, DP: 8.
+    # A threshold of 100 (~3× the historical max) catches genuine runaway cases
+    # without being sensitive to dataset size (full ETL vs. fast ETL subsets).
     len_diff = len(recombined_df) - len(df)
-    if (len_diff < -1) and (len_diff / len(df) < -0.0015):
+    if -len_diff > 100:
         raise AssertionError(
-            f"Expected reduction in 88888 rows less than ~{len(df) * 0.001} but found: {-len_diff}!"
+            f"Expected at most ~100 dropped 88888 rows but found: {-len_diff}! "
+            f"Historical maximums by table (Aug 2025): BA: 1, OD: 16, Sales: 32, UD: 15, DP: 8."
```
zaneselvans, line 1112: I relaxed this defensive check a bit to allow both the fast and full ETL to pass.

devtools/check_against_nightly.py
```diff
@@ -21,12 +21,87 @@
...
+# read GeoParquet's union-typed geometry column, so these are compared using
+# geopandas + pyarrow instead of polars.
+GEO_TABLES: frozenset[str] = frozenset(
+    {
+        "out_ferc714__georeferenced_respondents",
+        "out_censusdp1tract__states",
+        "out_censusdp1tract__counties",
+        "out_censusdp1tract__tracts",
+    }
+)
```
zaneselvans, line 44: With `geoparquet_io_manager` being retired in #5347 we might want to maintain a centrally located list of geoparquet outputs, or a helper that can identify them programmatically (probably any table with a `geometry` column or any column that has a `geometry` dtype in the metadata?)

devtools/check_against_nightly.py
```diff
@@ -21,11 +21,65 @@
...
 
+import io
 from pathlib import Path
 
 import click
+import geopandas
+import pandas as pd
 import polars as pl
 from polars.testing import assert_frame_equal
+from upath import UPath
```
zaneselvans, line 32: Since we're potentially looking at all of: S3, GCS, and local filesystem paths, UPath seemed like a simpler way to do it. This does require `pl.read_parquet()` instead of `scan_parquet()` but we were materializing the dataframe immediately with `assert_frame_equal` already.

src/pudl/output/ferc714.py
```diff
@@ -152,19 +272,19 @@ def categorize_eia_code(
...
+    util_series = (
         pd.Series(util_ids, name="utility_id_eia").drop_duplicates().convert_dtypes()
     )
 
-    df = eia_codes.merge(
-        ba_ids, left_on="eia_code", right_on="balancing_authority_id_eia", how="left"
-    ).merge(util_ids, left_on="eia_code", right_on="utility_id_eia", how="left")
+    df = codes_df.merge(
+        ba_series, left_on="eia_code", right_on="balancing_authority_id_eia", how="left"
+    ).merge(util_series, left_on="eia_code", right_on="utility_id_eia", how="left")
```
zaneselvans, line 338: Just syntax and naming changes to clean up some types and avoid clobbering pre-existing variables.

src/pudl/output/ferc714.py
```diff
@@ -219,49 +359,65 @@ def filled_core_eia861__assn_balancing_authority(
...
-        ref = df[mask]
-        if "exclude" in fix:
-            # Exclude utilities by state
-            mask = ~ref["state"].isin(fix["exclude"])
-            ref = ref[mask]
-        refs.append(ref)
-    # Build a table of new rows
-    # Insert (or overwrite) rows for each target balancing authority-year pair,
-    # using the reference year as a template.
+    eia861_years = set(df["report_date"].dt.year)
```
zaneselvans: @krivard You'd suggested bailing early in these functions if we discover that there's an empty list of `eia861_years` since in that case they should all become no-ops. However, if we get to this point in the ETL and there are no `eia861_years` something is very wrong, since we're way downstream of a bunch of EIA-861 data processing. So I'd almost rather raise an exception. Does that seem reasonable?

src/pudl/output/ferc714.py
```diff
@@ -219,49 +359,65 @@ def filled_core_eia861__assn_balancing_authority(
...
-    # Append to original table with matching rows removed
-    df = pd.concat([df[~replaced], apply_pudl_dtypes(pd.concat(tables), group="eia")])
-    # Remove balancing authorities treated as utilities
-    mask = np.zeros(df.shape[0], dtype=bool)
-    tables = []
+    new_tables: list[pd.DataFrame] = []
+
+    for target_year, fixes in ASSOCIATIONS.items():
+        if target_year not in eia861_years:
+            continue
```
zaneselvans, line 475: Skip if the target year isn't one of the years we're processing.

src/pudl/output/ferc714.py
```diff
@@ -219,49 +359,65 @@ def filled_core_eia861__assn_balancing_authority(
...
+        for fix in fixes:
+            source_year = fix["source_year"]
+            if source_year not in eia861_years:
+                logger.warning(
+                    f"Skipping BA association repair for id={fix['id']} "
+                    f"target_year={target_year}: source_year={source_year} not in "
+                    "available EIA-861 data. If running the full ETL, verify the "
+                    "fix spec is correct."
+                )
+                continue
```
zaneselvans, line 486: Skip if the source year isn't in one of the years we're processing -- but with a warning, since this means we're NOT going to apply a fix that we (in theory) have available to the target year which we ARE processing, which would lead to different outputs in that year than we would see in the full ETL with all years available.

src/pudl/output/ferc714.py
```diff
@@ -284,64 +440,76 @@ def filled_core_eia861__assn_balancing_authority(
...
-        pd.concat([df[~mask]] + tables)
+        pd.concat([df[~removal_mask]] + reassignment_tables)
         .drop_duplicates()
         .pipe(apply_pudl_dtypes, group="eia")
     )
 
 
 def filled_service_territory_eia861(
-    core_eia861__assn_balancing_authority: pd.DataFrame,
+    filled_assn: pd.DataFrame,
```
zaneselvans, line 567: Re-use the already filled association table, rather than re-computing it in here as we did before.

src/pudl/output/ferc714.py
```diff
@@ -617,7 +786,7 @@ def out_ferc714__georeferenced_respondents(
     you can see which respondent-years have both reported demand and decent geometries,
     calculate their areas to see if something changed from year to year, etc.
     """
-    respondents_gdf = (
+    return gpd.GeoDataFrame(
```
zaneselvans, line 923: The `GeoDataFrame()` is just  to ensure the type matches the function return type, since `apply_pudl_dtypes` can produce either DataFrame of GeoDataFrame outputs (and in this case it will always be GeoDataFrame)

src/pudl/output/ferc714.py
```diff
@@ -22,43 +23,200 @@
...
+                        f"source_year ({fix.source_year}) must not equal target_year."
+                    )
+        return self
+
+    def items(self):
+        """Iterate over (target_year, fixes) pairs, delegating to the root dict."""
+        return self.root.items()
+
+
+BA_FIXES = BaFixMap.model_validate(
```
zaneselvans, line 76: The old structure was very compact, but also felt a bit opaque, and needed to be expanded anyway in order to filter per-year for inputs and outputs, which was necessary to apply the fixes per-year. So I kind of turned it inside out. It's a little duplicative, but not gigantic, and easy to see what is where and filter based on source or target year.

src/pudl/output/ferc714.py
```diff
@@ -22,43 +23,200 @@
...
-  Rows in `core_eia861__yearly_balancing_authority` are added (if missing) for every target year
-  with the attributes from the reference year.
-  Rows in `core_eia861__assn_balancing_authority` are added (or replaced, if existing)
-  for every target year with the utility associations from the reference year.
-  Rows in `core_eia861__yearly_service_territory` are added (if missing) for every target year
-  with the nearest year's associated utilities' counties.
-* `exclude` (Optional[List[str]]): Utilities to exclude, by state (two-letter code).
-  Rows are excluded from `core_eia861__assn_balancing_authority` with target year and state.
+
+class BaFixMap(RootModel[dict[int, list[BaFix]]]):
```
zaneselvans, line 43: Because these fixes are hand compiled and there are some combinations of values that wouldn't really be valid, I went ahead and made a Pydantic class so they can be validated to avoid cut-and-paste errors and typos silently causing data loss or weirdness.

It's a `RootModel` rather than a `BaseModel` because instead of having named fields, the top-level structure is a mapping of years (which can't be known ahead of time) to lists of `BaFix` objects.

src/pudl/output/ferc714.py
```diff
@@ -181,35 +339,77 @@ def categorize_eia_code(
 ################################################################################
 
 
+def _apply_exclude_states(
```
zaneselvans, line 355: This guards against an edge case that could come up for smaller BAs, if we inadvertantly excluded all of the states it was operating in (and some only exist in a single state). Might also indicate a typo or cut and paste error where we're using the wrong BA ID.

devtools/check_against_nightly.py
```diff
@@ -36,29 +90,37 @@
...
+        ref_upath = ref_base / local_path.name
+        if local_path.stem in GEO_TABLES:
+            ok = _compare_geo(local_path, ref_upath, pdb)
+            click.echo("OK" if ok else "")
         else:
-            click.echo("OK")
+            local = pl.read_parquet(local_path)
+            reference = pl.read_parquet(io.BytesIO(ref_upath.read_bytes()))
+            try:
+                assert_frame_equal(local, reference, check_column_order=False)
```
krivard, line 113: for a second I was worried about whether switching from lazy to eager changes the memory footprint or this particular part of the script but after poking around I'm convinced it doesn't, entirely because the old version never explicitly requested the streaming engine.

devtools/check_against_nightly.py
```diff
@@ -21,11 +21,65 @@
...
+        "out_censusdp1tract__counties",
+        "out_censusdp1tract__tracts",
+    }
+)
+
+
+def _to_upath(reference_dir: str) -> UPath:
+    """Convert reference_dir string to a UPath with appropriate storage options."""
+    if reference_dir.startswith("s3://"):
+        return UPath(reference_dir, anon=True)
```
krivard, line 50: nice

devtools/check_against_nightly.py
```diff
@@ -21,11 +21,65 @@
...
+    return UPath(reference_dir)
+
+
+def _compare_geo(local_path: Path, ref_upath: UPath, pdb: bool) -> bool:
+    """Compare a GeoParquet file against a reference. Returns True on match."""
+    local = geopandas.read_parquet(local_path)
+    reference = geopandas.read_parquet(io.BytesIO(ref_upath.read_bytes()))
+
+    geo_col = local.geometry.name
```
krivard, line 59: TIL there is only ever one active geometry column in a geodataframe

devtools/check_against_nightly.py
```diff
@@ -21,11 +21,65 @@
...
+    non_geo = [c for c in local.columns if c != geo_col]
+    sort_cols = [c for c in non_geo if not pd.api.types.is_float_dtype(local[c])]
```
krivard, line 61: > pd.api.types.is_float_dtype(

that's intense

does this line say "every column that isn't a geometry column and that isn't storing a ~continuous value is a column we'd like to use for sorting"?

src/pudl/output/ferc714.py
```diff
@@ -22,43 +22,163 @@
...
-  Rows in `core_eia861__yearly_service_territory` are added (if missing) for every target year
-  with the nearest year's associated utilities' counties.
-* `exclude` (Optional[List[str]]): Utilities to exclude, by state (two-letter code).
-  Rows are excluded from `core_eia861__assn_balancing_authority` with target year and state.
+    id: int
+    source_year: int
+    exclude_states: NotRequired[list[str]]
+
+
+ASSOCIATIONS: dict[int, list[BaFix]] = {
```
krivard: I am not going to rigorously check these. lmk if you are looking for an extra pair of eyes on whether this version correctly re-encodes the old associations (has everything that used to be there and nothing that wasn't) and i can do a second pass.

src/pudl/output/ferc714.py
```diff
@@ -22,43 +22,163 @@
...
+    2018: [
+        # Nevada Power Company (now NV Energy)
+        {"id": 13407, "source_year": 2013},
+    ],
+    2019: [
+        # Nevada Power Company (now NV Energy)
+        {"id": 13407, "source_year": 2013},
+    ],
+}
+"""Spot-fixes for known EIA-861 balancing authority ID errors, keyed by target year.
```
krivard: this is near the upper edge of what seems reasonable to maintain in a Python file; if we think this will grow over time it might be good to write an icebox issue to convert it to CSV or YAML next time we have to touch it

src/pudl/output/ferc714.py
```diff
@@ -77,14 +197,14 @@
...
-* `replace` (Optional[bool]): Whether to remove rows where `id` is a utility in
-  `core_eia861__assn_balancing_authority`. Applies only if `reassign=True`.
+* ``id`` (int): EIA balancing authority (BA) identifier (``balancing_authority_id_eia``).
+  Rows for ``id`` are removed from ``core_eia861__yearly_balancing_authority``.
+* ``reassign`` (Optional[bool]): Whether to reassign utilities to parent BAs.
+  Rows for ``id`` as BA in ``core_eia861__assn_balancing_authority`` are removed.
+  Utilities assigned to ``id`` for a given year are reassigned
+  to the BAs for which ``id`` is an associated utility.
+* ``replace`` (Optional[bool]): Whether to remove rows where ``id`` is a utility in
+  ``core_eia861__assn_balancing_authority``. Applies only if ``reassign=True``.
```
krivard, line 258: RST syntax repair only

src/pudl/output/ferc714.py
krivard: were we relying on clobbering to improve memory efficiency in this area? this area of the code doesn't seem directly familiar but I remember 714 having a couple of hotspots.

src/pudl/output/ferc714.py
```diff
@@ -186,30 +306,50 @@ def filled_core_eia861__yearly_balancing_authority(
...
+            if source_year not in eia861_years:
+                logger.warning(
+                    f"Skipping BA repair for id={fix['id']} target_year={target_year}: "
+                    f"source_year={source_year} not in available EIA-861 data. "
+                    "If running the full ETL, verify the fix spec is correct."
+                )
+                continue
+            target_key = (fix["id"], target_date)
+            if target_key in dfi.index:
+                continue
```
krivard, line 417: it looks like this is skipping when the target key is present. From the docstring,

> Each key is a year whose reported BA-utility associations are known to be incorrect.

I expected that, if the target key existed, it would mean the data there was wrong and needed to be fixed, so we *shouldn't* skip it.

Is the logic here flipped or did I misunderstand the docs?

src/pudl/output/ferc714.py
krivard: Yeah, an assert or other exception-generating check feels right here.

src/pudl/output/ferc714.py
krivard: checking understanding: this warning appears when we know the data is wrong, but we don't have the year of data we would (in the full ETL) use to repair it?

src/pudl/output/ferc714.py
```diff
@@ -219,49 +359,65 @@ def filled_core_eia861__assn_balancing_authority(
...
+                "report_date"
+            ].eq(target_date)
+            replaced |= target_mask.to_numpy(bool)
+            new_tables.append(ref.assign(report_date=target_date))
+
+    if new_tables:
+        new_rows = apply_pudl_dtypes(pd.concat(new_tables), group="eia")
+        df = pd.concat([df[~replaced], new_rows], axis="index")
+    else:
+        df = df[~replaced]
```
krivard: how do we get into a state where there are rows to replace but nothing in `new_tables`?

krivard: note to self: here this set of comments stops; will resume next review

krivard (commented):
Partial review; will resume next week

devtools/check_against_nightly.py
zaneselvans: One one that's ever active, but IIRC recent versions of GeoParquet / GeoArrow / GeoPandas do allow multiple geometry columns to be stored and managed w/ independent CRS and activated depending on what you're doing.

src/pudl/output/ferc714.py
zaneselvans: Yes, that's right. The `target_year` is the year we're trying to fix which IS being processed in this scenario, and the `source_year` is the source of the fix, which is NOT being processed here, so we know we have a year we'd like to fix, but we don't have the data required to fix it.

src/pudl/output/ferc714.py
```diff
@@ -284,64 +499,76 @@ def filled_core_eia861__assn_balancing_authority(
...
+
+    # Collect (ba_id, date) pairs relevant to any BA_FIXES fix so we can
+    # select the matching rows from the filled association table.
+    relevant: set[tuple] = set()
+    for target_year, fixes in BA_FIXES.items():
+        for fix in fixes:
+            if target_year in eia861_years:
+                relevant.add((fix.id, pd.Timestamp(target_year, 1, 1)))
+            if fix.source_year in eia861_years:
+                relevant.add((fix.id, pd.Timestamp(fix.source_year, 1, 1)))
```
krivard, line 605: such that `relevant` contains entries for all the known-bad data _and_ entries for all the known-good data we want to patch it with

src/pudl/package_data/settings/etl_fast.yml
```diff
@@ -53,35 +53,9 @@ pudl:
...
-          2017,
-          2018,
-          2019,
-          2020,
-          2021,
-          2022,
-          2023,
-          2024,
-        ]
+      # These years must overlap with FERC-714 to allow assets that depend on both to
```
krivard, line 56: do we only need one year in both eia861 and ferc714, or do the lists need to match exactly? they currently match exactly but for me "overlap" suggests a more lenient constraint

tests/integration/scripts/console_scripts_test.py
krivard: will this need to be re-edited when the fast ETL config is updated? if so we should leave a note in the config

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+
+Selected BA IDs used in tests:
+  - 56669 (MISO): source_year=2011, target_years=[2009, 2010]
+  - 59504 (SWPP): source_year=2014, target_years=[2006..2013];
+      exclude_states=["NE"] for 2006-2009
+  - 13407 (Nevada Power): source_year=2013, target_years=[2014..2019]
+
+Selected UTILITIES IDs used in tests:
+  - 4254  (Consumers Energy): no reassign — just removed
+  - 14328 (Pacific G&E):      reassign=True — children move to parent BA
```
krivard: nit: probably should use the same notation for both.

either:


```suggestion
  - 4254  (Consumers Energy): no reassign — just removed
  - 14328 (Pacific G&E):      must reassign — children move to parent BA
```

krivard: or:

```suggestion
  - 4254  (Consumers Energy): reassign=False — just removed
  - 14328 (Pacific G&E):      reassign=True — children move to parent BA
```

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+                [r.get("name", "Dummy") for r in rows], dtype=pd.StringDtype()
+            ),
+        }
+    )
+
+
+def _ba_assn(*rows: dict) -> pd.DataFrame:
+    """Build a minimal core_eia861__assn_balancing_authority DataFrame.
+
+    Each dict may have keys: ba_id, util_id, state, year.
```
krivard:
```suggestion
    Each dict must have keys: ba_id, util_id, state, year.
```

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+        result = filled_core_eia861__yearly_balancing_authority(df)
+        added = result[
+            (result["balancing_authority_id_eia"] == 56669)
+            & (result["report_date"].dt.year == 2009)
+        ]
+        assert len(added) == 1
+        assert added.iloc[0]["balancing_authority_code_eia"] == "MISO"
+        assert added.iloc[0]["balancing_authority_name_eia"] == "Midwest ISO"
+
+    def test_does_not_overwrite_existing_target_year_row(self):
```
krivard, line 159: whether this test is good depends on whether the fixlist targets identify bad data or just missing data. if bad data, then we really do want it to overwrite extant targets, no?

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+            )
+        )
+        result_ids = set(
+            zip(
+                result["balancing_authority_id_eia"],
+                result["report_date"].dt.year,
+                strict=False,
+            )
+        )
+        # Any rows present in result but not original must be from UTILITIES reassignment,
```
krivard: do we expect UTILITIES removal here or reassignment? for removal I would expect empty `added`

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+        The 2009 assn rows are replaced by the 2011 rows (with date changed to 2009).
+        """
+        df = _ba_assn(
+            # Source rows (2011)
+            {"ba_id": 56669, "util_id": 100, "state": "IL", "year": 2011},
+            {"ba_id": 56669, "util_id": 200, "state": "MN", "year": 2011},
+            # Existing wrong rows for 2009 (to be replaced)
+            {"ba_id": 56669, "util_id": 999, "state": "WI", "year": 2009},
+            # Dummy row to put 2009 in eia861_years
+            {"ba_id": 99999, "util_id": 1, "state": "TX", "year": 2009},
```
krivard: why is this needed? does the WI row not ensure 2009 is in eia861_years?

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+        # Old wrong row should be gone
+        assert 999 not in miso_2009["utility_id_eia"].to_numpy()
+        # Source year rows should be present with target date
+        assert set(miso_2009["utility_id_eia"].to_numpy()) == {100, 200}
+
+    def test_adds_target_year_assn_rows_when_absent(self):
+        """Source year assn rows are copied to a target year that has none."""
+        df = _ba_assn(
+            {"ba_id": 56669, "util_id": 100, "state": "IL", "year": 2011},
+            {"ba_id": 99999, "util_id": 1, "state": "TX", "year": 2009},
```
krivard, line 266: oh wait maybe you meant the comment above to go here
```suggestion
            # Dummy row to put 2009 in eia861_years
            {"ba_id": 99999, "util_id": 1, "state": "TX", "year": 2009},
```

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+        swpp_2009 = result[
+            (result["balancing_authority_id_eia"] == 59504)
+            & (result["report_date"].dt.year == 2009)
+        ]
+        assert 100 in swpp_2009["utility_id_eia"].to_numpy(), (
+            "KS utility should be included"
+        )
+        assert 200 not in swpp_2009["utility_id_eia"].to_numpy(), (
+            "NE utility should be excluded"
+        )
```
krivard: yes good

tests/unit/output/ferc714_test.py
krivard, line 1: any reason you're checking individual cells here, and not using pd.testing.assert_frame_equal? Checking (a wide variety of different) individual cells is harder to skim, but if we need it we need it

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+        filled_assn = _ba_assn(
+            {"ba_id": 56669, "util_id": 100, "state": "CA", "year": 2009},
+            {"ba_id": 56669, "util_id": 100, "state": "CA", "year": 2011},
+            # Utility 200 in NV: never has county data
+            {"ba_id": 56669, "util_id": 200, "state": "NV", "year": 2009},
+            {"ba_id": 56669, "util_id": 200, "state": "NV", "year": 2011},
+        )
+        st = _service_territory(
+            {"util_id": 100, "state": "CA", "year": 2009, "county_id_fips": pd.NA},
+            {"util_id": 100, "state": "CA", "year": 2011, "county_id_fips": "06001"},
```
krivard: do we need these CA rows for the fill to work? if not drop them for clarity

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Tests for BaFixMap validation
+# ---------------------------------------------------------------------------
+
+
+class TestBaFixMap:
+    def test_rejects_duplicate_ba_in_same_target_year(self):
+        """Two entries with the same BA id in one target year raise ValueError."""
```
krivard:
```suggestion
        """Sourcing a single target from multiple years is ambiguous and not permitted."""
```

tests/unit/output/ferc714_test.py
```diff
@@ -0,0 +1,516 @@
...
+                {
+                    2009: [
+                        {"id": 56669, "source_year": 2011},
+                        {"id": 56669, "source_year": 2012},
+                    ]
+                }
+            )
+
+    def test_rejects_self_referential_source_year(self):
+        """source_year == target_year raises ValueError."""
```
krivard:
```suggestion
        """A source matching the target is a no-op and not permitted."""
```

krivard (changes requested):
Blocking question: my understanding of `BA_FIXES` is that it encodes (target, BA, source) triples where the data for (target, BA) has utility mappings that are _actively incorrect_. & that what we want is if (target, BA) data exists, and (source, BA) data is available, we should ~substitute (source, BA) data instead. yes/no?

 filled_core_eia861__yearly_balancing_authority is *skipping* entries where (target, BA) data exists, which seems like it would leave actively incorrect data in place even when (source, BA) data is available. see: https://github.com/catalyst-cooperative/pudl/pull/4568/changes#r3484025672

TestFilledYearlyBalancingAuthority.test_does_not_overwrite_existing_target_year_row verifies that (target, BA) data is not overwritten when it exists, which seems wrong. see: https://github.com/catalyst-cooperative/pudl/pull/4568/changes#r3493850519

everything else is minor

src/pudl/output/ferc714.py
zaneselvans: Yeah, it seems like a good candidate for a JSON or YAML file.  I would expect that there should be a handful of updates to it each year, but I think that will only happen if we can come up with a more automated way of identifying bad data that needs fixing.

tests/unit/output/ferc714_test.py
zaneselvans: No, this was an LLM artifact. I've switched over to using whole dataframes with a little helper that sets the indices using PKs from the PUDL metadata and checks without caring about the row or column ordering.

src/pudl/output/ferc714.py
zaneselvans: I think you're right. I'll replace this assignment with an assertion to check that this expectation is actually true.

src/pudl/output/ferc714.py
zaneselvans: This table in particular is a tiny one, aimed at getting the right associations between the EIA and FERC-714 utility / BA IDs, so I don't think this is a concern.

src/pudl/output/ferc714.py
zaneselvans: Okay added an assertion.

devtools/check_against_nightly.py
zaneselvans: Yeah, this is kind of a brute-force way of attempting to enforce identical sorting between the dataframes.

src/pudl/output/ferc714.py
zaneselvans: I also did not go over it with a fine toothed comb, but it produces exactly the same set of output geometries, as the previous set of fixes.

src/pudl/package_data/settings/etl_fast.yml
zaneselvans: I'm not sure. I made them the same to maximize the chances that things would work, since we're not trying to test a variety of different subsets of data processing -- just one that will run end to end.

tests/unit/output/ferc714_test.py
zaneselvans: I was able to simplify the tests in here a bit -- switched to comparing entire dataframes, got rid of some extraneous records, standardized `expected` and `actual` naming.

tests/unit/output/ferc714_test.py
zaneselvans: Nope. Nixed.

src/pudl/output/ferc714.py
zaneselvans: I think the docstrings were not reflecting what the system was actually doing. The three `filled_core_eia861__` functions don't all take the same approach -- clobbering vs. filling.

zaneselvans:
@krivard Okay, I cleaned up the test module, and looking at the three `filled_core_eia861__` functions, I think you're right that they don't all take the same approach to clobbering vs. filling in data in the target years.

I've updated the docstrings and some comments to hopefully more accurately describe what each of them does, and the code exactly reproduces the outputs in nightly, so in terms of refactoring to allow us to use a subset of years in the fast ETL, I think the current state is good.

However, I'm less sure that what the code is currently doing (and has been doing for years) is correct. I'm not sure why we would want to fill sometimes and clobber other times in these different kinds of tables, and I think to really understand that I would need to be comparing maps of the utility and balancing authority service territories that correspond to the filled vs. unfilled tabular data, and comparing the source & target years for individual service territories, which I think is beyond the scope of this PR.

But! That same infrastructure would make for a really nice historical service territory browser / interactive dashboard. And now that we have real geoparquet outputs it should be easy to get it working. And this would help us update the fixes to include more recent years of data too. I think we should probably also persist geoparquet outputs of both the originally reported EIA-861 service territories and the "filled" EIA-861 service territories, for both utilities and balancing authorities, totally independent of the FERC-714 planning area outputs. 

One thing geoparquet does is use dictionary encoding for geometries, so you can have a bajillion rows that contain the same shape in a parquet file, and it's only stored once. So I think we can more economically publish a bunch of this data as real spatial outputs now.

src/pudl/output/ferc714.py
```diff
@@ -22,43 +23,213 @@
...
+Each key is a target year that requires a manual repair for one or more balancing
+authorities. Each value is a list of :class:`BaFix` entries describing which BA to
+repair and which single ``source_year`` should be treated as the authoritative template
+for that specific target year.
+
+The fixes are consumed by three functions:
+:func:`filled_core_eia861__yearly_balancing_authority`,
+:func:`filled_core_eia861__assn_balancing_authority`, and
+:func:`filled_core_eia861__yearly_service_territory`. They all use the same explicit
+``source_year -> target_year`` mappings, but apply them differently:
```
krivard, line 213: ⭐ excellent, this and what follows is superior text

src/pudl/output/ferc714.py
```diff
@@ -181,104 +352,182 @@ def categorize_eia_code(
...
-    uses the reference year as a template. The function also removes balancing
-    authorities that are manually categorized as utilities.
+    Applies the explicit year-to-year repairs described in :data:`BA_FIXES` to the
+    BA identity table.
+
+    For each ``source_year -> target_year`` mapping, adds a row for the target
+    year if one is missing, copying all attributes from the source year.
+    Existing target-year rows are left unchanged. This is therefore an
+    explicit year-to-year copy for missing BA identity rows, not a general
+    forward-fill or backward-fill across a range of years.
```
krivard: this is clear enough but if we wanted to use the word "imputation" here I wouldn't be mad

src/pudl/output/ferc714.py
krivard: new docs make it clear! ty

src/pudl/package_data/settings/etl_fast.yml
```diff
@@ -53,35 +53,9 @@ pudl:
...
+      # be materialized.
```
krivard:
```suggestion
      # be materialized. If you modify this list, be sure to update the matching
      # pudl_service_territories commands in
      # tests/integration/scripts/console_scripts_test.py
```

tests/unit/output/ferc714_test.py
krivard: resolved with docs update!

krivard (approved):
Good fixes! Left a suggest to breadcrumb the updates we'll need to make in console_scripts_test whenever we update the fast etl years (strongly recommended), and an excuse to use the word "imputation" in a sentence (totally optional) but this is good to merge.

src/pudl/output/ferc714.py
zaneselvans: Maybe I'm conflating it with interpolation, but imputation doesn't feel quite right -- we can be filling in the middle, or before/after the ranges in which we have data. It's a fill-from-other-source. A copy. Whereas I generally associate imputation with a more dynamic process that's using a procedure to estimate a value. I've tried to simplify the language and make it a more literal description of what happens.

PR #4554, opened 2025-08-21
Run dbt-autofix deprecations on our dbt project.

zaneselvans:
# Overview

* We were getting a bunch of deprecation warnings because of a change in how arguments are supposed to be specified for generic tests.  I ran the [`dbt-autofix`](https://pypi.org/project/dbt-autofix/) script to update our project to the new syntax automatically.
* I didn't look at every single change. The only type of change I saw when I scanned through was the addition of the explicit `arguments:` section under `data_tests`, though the script can also fix other types of deprecations.

## Documentation

@jdangerx @krivard are there any examples in the dbt dev docs update that need to be updated?

# Testing

- I ran the dbt tests locally on a fresh ETL and didn't get any deprecation warnings or errors.

PR #4532, opened 2025-08-08
dbt docs: debugging failures

krivard:
# Overview

Closes #4126 .

## What did you change?

* Initial draft for debugging data validation failures.

## Documentation

Make sure to update relevant aspects of the documentation:

- [ ] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html): reference the PR and related issues.
- [ ] Update relevant Data Source jinja templates (see `docs/data_sources/templates`).
- [ ] Update relevant table or source description metadata (see `src/metadata`).
- [ ] Review and update any other aspects of the documentation that might be affected by this PR.

# Testing

How did you make sure this worked? How can a reviewer verify this?

## To-do list

- [ ] This was committed `no-verify` so it needs. y'know. verification.
- [ ] dbt wrapper function missing a docstring but I abandoned that so it can probably just go away
- [ ] Line lengths in data_validation.rst are probably horrific
- [ ] Resolve remaining `.. todo::` blocks in debugging section
  - [ ] When to use `--store-failures`
  - [ ] Dedicated guide for rowcounts failures
  - [ ] Refresh dedicated guide for constraints failures

PR #4571, opened 2025-08-26
test: update for new 'arguments' dbt test definition syntax

jdangerx:

`dbt_helper update-tables --row-counts` doesn't work anymore because it can't find the various arguments passed into the `check_row_counts_per_partition` tests.

Follow-up to #4554 .

## What did you change?

Grab those damn arguments.

Update the tests.

## Documentation

:shrug: 

# Testing

1. Run `dbt_helper update-tables <table> --row-counts --clobber` on `main`. See that we don't get partitions anymore
2. Run same command on this branch. See that the partitions are preserved
3. See that unit tests pass.

## To-do list

### DIFF
--- a/docs/dev/data_validation_quickstart.rst
+++ b/docs/dev/data_validation_quickstart.rst
@@ -14,7 +14,7 @@ The ``dbt/`` directory contains the PUDL dbt project which manages our `data tes
 The data validation tests run on the Parquet outputs that are in your
 ``$PUDL_OUTPUT/parquet/`` directory. It's important that you ensure the outputs you're
 testing are actually the result of the code on your current branch, otherwise you may
-be surprised when the data test fails in CI or the nightly builds.
+be surprised when the data test passes locally but fails in CI or the nightly builds.
 
 We have a script, :mod:`pudl.scripts.dbt_helper`, to help with some common workflows.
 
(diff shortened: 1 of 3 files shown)


src/pudl/scripts/dbt_helper.py
```diff
@@ -551,21 +551,11 @@ def _extract_row_count_partitions(table: DbtTable) -> list[str | None]:
                 raise ValueError(f"Row counts test expected to be a dictionary: {test}")
             test_def = test.get("check_row_counts_per_partition")
             if isinstance(test_def, dict):
-                partitions.append(test_def.get("partition_expr"))
+                partitions.append(test_def.get("arguments", {}).get("partition_expr"))
 
     return partitions
 
 
-def get_row_count_test_dict(table_name: str, partition_expr: str):
```
jdangerx: This function wasn't used anywhere except for a test. So I deleted it.

test/unit/dbt_helper_test.py
jdangerx, line 1: Just updating all the expected values to include the "arguments" key, mostly.

krivard (approved):
🔥 

In a followup we should add an integration test so we don't get bit by something like this again

zaneselvans:
@jdangerx  Did all of these changes make it into #4532 instead it looks like?

zaneselvans:
I'm gonna close since there's no diff with `main` and it looks like all the changes got merged. Let me know if I'm confused!

jdangerx:
Ah yes. To get #4532 to both (a) be accurate and (b) pass CI I had to include the changes in that PR too. I had originally planned to do some sort of stacked PR situation but that clearly didn't happen, oops!

Issue #4481, opened 2025-07-25
Shut down Datasette


zaneselvans:
# Overview

- Our [Datasette](https://data.catalyst.coop/) instance is no longer functional, and all the links we have to it are just dead, and are probably confusing / frustrating for anyone who tries to follow them.
- We haven't reached [datasette parity](https://github.com/catalyst-cooperative/eel-hole/issues/36) with the Eel Hole yet, but the FERC SQLite DBs on Datasette aren't doing anybody any good as it is.

## Success Criteria

How will we know that we're done?

* [ ] We no longer have public references anywhere to Datasette.
* [ ] The PUDL Viewer is at at https://data.catalyst.coop
* [ ] Anybody going to one of the Datasette index page URLs gets redirected to the PUDL Viewer landing page
* [ ] Anybody going to the old `viewer.catalyst.coop` URL gets redirected to the new PUDL Viewer landing page
* [ ] Code specific to Datasette has been removed from the PUDL Repo
* [ ] We've shut down our Fly.io app

## Tasks

* [ ] Shut down our Datasette app on Fly.io
* [ ] Change the domain mapping for the PUDL Viewer to use `data.catalyst.coop`
* [ ] Set up a redirect from `viewer.catalyst.coop` to `data.catalyst.coop`
* [ ] Swap Datasette references for PUDL Viewer in:
    * [ ] Sphinx documentation
    * [ ] Kaggle Notebooks / the PUDL Examples repository
    * [ ] Our GitHub organization page
    * [ ] Any social profiles where we shared links.
* [ ] Remove everything under `devtools/datasette`
* [ ] Remove `datasette` and any related dependencies from `pyproject.toml`
* [ ] Remove `pudl.metadata.DatasetteMetadata` class
* [ ] Remove PUDL datasette Jinja metadata template
* [ ] Remove integration tests related to datasette metadata
* [ ] Update release notes & publicize the change

jdangerx:
We have a lot of links that go to `data.catalyst.coop/<db_name>/<table_name>` - I think we would rather *not* go and change all of those now and then change them again once catalyst-cooperative/eel-hole#10 works again.

But right now of course datasette is just down all the time. Some options, in order of effort:

* ignore this and just have a bunch of dead links that we don't intend to go anywhere
* add custom routes to PUDL viewer that say "sorry, we don't have this table as a live preview yet, download the parquet file here and use `vd` or Tad to look at it"
[...]

PR #4573, opened 2025-08-26
Fixes to appease make pre-commit

krivard:
# Overview

`make pre-commit` doesn't currently pass in main. Two precommits fail:

* shellcheck, which fails if you don't have docker running, and if you do have docker running, finds minor problems in `devtools`
* nb-output-clear, which takes issue with pretty much everything in notebooks/work-in-progress

We currently recommend running `make ci` as part of the default PR to-do list (as below). Since the `ci` target depends on `pre-commit`, fixing the above failures is necessary to ensure we don't send folks off on a tangent debugging problems they didn't create.

## What did you change?

* switched the ShellCheck pre-commit to the python version, so it no longer requires Docker
* fixed the shellcheck lints
  * use double quotes around variables that might contain paths, so that whitespace in those variables doesn't cause a problem
  * use `$()` instead of backticks
* fixed the nb-output-clear, uh, lints? or whatever you wanna call them.
  * dropped all the output cells in notebooks/work-in-progress

# Testing

`make pre-commit` passes

## To-do list

- [x] Run `make ci` locally to ensure that the merge queue will accept your PR.


devtools/datasette/fly/run.sh
```diff
@@ -8,5 +8,5 @@ ls
 mv all_dbs.tar.zst /data
 zstd -f -d /data/all_dbs.tar.zst -o /data/all_dbs.tar
 tar -xf /data/all_dbs.tar --directory /data
-datasette serve --host 0.0.0.0 ${DATABASES} --cors --inspect-file inspect-data.json --metadata metadata.yml --setting sql_time_limit_ms 5000 --port $DATASETTE_PORT > /dev/null &
+datasette serve --host 0.0.0.0 "${DATABASES}" --cors --inspect-file inspect-data.json --metadata metadata.yml --setting sql_time_limit_ms 5000 --port "$DATASETTE_PORT" > /dev/null &
```
krivard, line 11: maybe: drop this file entirely, we're done with datasette

devtools/pyspy_pudl_etl.sh
```diff
@@ -3,8 +3,8 @@
...
 
-PROFILE_DIR=`git rev-parse --show-toplevel`/devtools/profiles
-OUTFILE=pyspy_pudl_etl_`date +%Y%m%d%H%M%S`.json
-mkdir -p $PROFILE_DIR
+PROFILE_DIR=$(git rev-parse --show-toplevel)/devtools/profiles
+OUTFILE=pyspy_pudl_etl_$(date +%Y%m%d%H%M%S).json
+mkdir -p "$PROFILE_DIR"
 
-py-spy record -r 10 -f speedscope -o "$PROFILE_DIR/$OUTFILE" -- pudl_etl --clobber $1
+py-spy record -r 10 -f speedscope -o "$PROFILE_DIR/$OUTFILE" -- pudl_etl --clobber "$1"
```
krivard, line 10: this file is clearly unfinished ("this script takes two arguments ... we don't talk about the second argument")

but at least now shellcheck believes in it

notebooks/work-in-progress/replace_eia_api.ipynb
```diff
@@ -1 +1,281 @@
...
+  },
+  "vscode": {
+   "interpreter": {
+    "hash": "adf7f0b18919f80aa48732f306897016c591ffeed72c8be7c2b511e7f25e6b58"
+   }
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 2
+}
```
krivard, line 281: this file has so many new lines because the old version was minified to only have one line and the new version is prettyprinted

krivard (commented):
It's possible that work-in-progress is intended to retain all the notebook outputs. if so we should update the filter so nb-output-clear ignores them.

devtools/datasette/fly/run.sh
zaneselvans: We have #4481 (Shut Down Datasette) as an issue in the current sprint so this should all go away soon.

devtools/pyspy_pudl_etl.sh
zaneselvans: Lol yes, this is a very old somewhat throwaway script but maybe about to become useful again as we tackle performance issues.

notebooks/work-in-progress/replace_eia_api.ipynb
zaneselvans: We should probably have a joint notebook culling session. None of these are tested, and most of this work is not in progress, so I suspect many of them are obsolete and can be removed.

zaneselvans (approved):
Thank you for doing this cleanup!

We should probably make a "Cull obsolete WIP notebooks" issue. And also test notebooks that we do expect to stick around and want to keep working.

PR #4576, opened 2025-08-27
Update links in data dictionary. Add code snippets.

zaneselvans:
# Overview

* Datasette is dead 💀 Long live Datasette!
* This means the links in our data dictionary go to a graveyard. 🧟 
* We should update those links to point at the [PUDL Viewer](https://viewer.catalyst.coop) page for those tables instead.
* We've also long discussed offering code-snippets here that people can cut-and-paste to pull the table, and it turns out this is very easy to do so why not roll it in here.?

## What did you change?

* Updated the Jinja template we use to output resources to create link to the PUDL Viewer for that table.
* Added tabbed interface with pandas, polars, and SQL snippets to pull the table.

## Issues / Questions

* Some of the name-based links don't seem to do what we want...
  * ~[out_eia923__fuel_receipts_costs](https://viewer.catalyst.coop/search?&q=name:out_eia923__fuel_receipts_costs) shows multiple tables, not just the one I’m trying to get.~ Currently this is the expected behavior. See https://github.com/catalyst-cooperative/eel-hole/issues/56 
  * ~[out_censusdp1tract__counties](https://viewer.catalyst.coop/search?&q=name:out_censusdp1tract__counties) shows nothing~ See #4577 
* ~The [official Sphinx Tabs plug-in](https://sphinx-tabs.readthedocs.io/en/latest/) is not responding to the light/dark setting of the theme as a whole -- it's always light, which makes reading the appropriately highlighted code snippets difficult.~ (seems to be my local setup, [looks fine on RTD](https://catalystcoop-pudl--4576.org.readthedocs.build/en/4576/data_dictionaries/pudl_db.html))
* ~It would probably be good to add a tab for `data.frame` (R) in here too if @zschira or @krivard know what that would look like.~ Done!
* Are there any other data-wrangling crowd favorites we should include?
* Some of the tables are larger than memory and if someone tries to pull them directly, it'll take a while and then their computer will maybe crash... How can we warn about this in a context-appropriate way?

# Testing

- Built the docs locally and inspected the data dictionary page.

## To-do list

- [x] Add `data.frame` R-based code snippet to the template.
- [x] Make code snippets use light/dark theme settings appropriately.
- [x] Fix direct-to-table links.
  - [x] See #4577 
  - [x] See https://github.com/catalyst-cooperative/eel-hole/issues/56


zaneselvans:
Huh, the light/dark thing seems like maybe it's just my local setup, since it looks fine on RTD.

docs/templates/resource.rst.jinja
```diff
@@ -7,12 +7,36 @@
 Access methods
 ^^^^^^^^^^^^^^^
 
-{% if resource.create_database_schema -%}
-* `Browse or query this table in Datasette. <https://data.catalyst.coop/pudl/{{ resource.name }}>`__
-{% else -%}
-* This table is not published to Datasette.
-{%- endif %}
-* `Download this table as a Parquet file. <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/{{ resource.name }}.parquet>`__
+* `Browse and query this table online <https: //viewer.catalyst.coop/search?&q=name:{{ resource.name }}>`__
```
zaneselvans, line 10: This table link isn't ideal since it may point at nothing, or multiple tables, until we address #4577 and https://github.com/catalyst-cooperative/eel-hole/issues/56 but, uh, it's better than the guaranteed to be dead links that we've got now!

docs/templates/resource.rst.jinja
```diff
@@ -7,12 +7,36 @@
...
+* `Download this table as a Parquet file <https: //s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/{{ resource.name }}.parquet>`__
+
+.. tabs::
```
zaneselvans, line 13: Are there any other popular `read_parquet()` methods out there that we should include?

zaneselvans: I looked at the Rust version of polars and Julia and they were both much more involved. I think this is good for now.

zaneselvans:
Oh no! It doesn't behave like that for me using Firefox on MacOS, but if it's broken on Chrome then it'll be broken for a lot of people. 😞 

Have you also observed this behavior [on Read The Docs](https://catalystcoop-pudl--4576.org.readthedocs.build/en/4576/data_dictionaries/pudl_db.html)? Or is it just on the local build?

Maybe we should also create [an issue in their repo](https://github.com/executablebooks/sphinx-tabs/issues/204)

It also seems to have the scrolling behavior on Safari.

zaneselvans:
@krivard I switched from using the built-in `.. code-tab::` shortcut, which also groups tabs, to just manually including a `.. code-block::` inside each tab in the default ungrouped tabs. It still looks right on Firefox, and seems to have addressed the crazy scrolling on RTD, at least in Safari.

docs/templates/resource.rst.jinja
```diff
@@ -7,12 +7,44 @@
...
+
+   .. tab:: pandas
+
+      .. code-block:: python
+
+         import pandas as pd
+         df = pd.read_parquet(
+             "s3://pudl.catalyst.coop/nightly/{{ resource.name }}.parquet",
+             dtype_backend="pyarrow",
+         )
```
zaneselvans, line 23: Manually inserting (ungrouped) code-blocks rather than using the (automatically grouped) `code-tabs` shortcut seems to have addressed the crazy scrolling.

PR #4578, opened 2025-08-28
Add new PUDL citations for August 2025

e-belfer:
# Overview

Add more PUDL cites found in writing the newsletter.

## What did you change?
Updated `catalyst_cites.bib`

## Documentation


docs/catalyst_cites.bib
```diff
@@ -1,30 +1,22 @@
...
-}
```
zaneselvans: It looks like you might be working from an old version of this file? These shouldn't be removed should they? (I merged the existing citation update PR the other day)

docs/catalyst_cites.bib
```diff
@@ -1,30 +1,22 @@
...
+@misc{Buchsbaum_Kahn-Lang_2025,
```
zaneselvans: Should probably have a different ID for this entry vs. the next one.  Looks like a cut and paste error.

docs/catalyst_cites.bib
```diff
@@ -346,15 +338,6 @@ @techreport{FlatPowerDemandOver2023
...
 
-@techreport{StrategicIndustriesSurging2024,
-    title={Strategic Industries Surging: Driving US Power Demand},
-    author={Wilson, John D. and Zimmerman, Zach and Gramlich, Rob},
-    institution ={Grid Strategies LLC},
-    year={2024},
-    month=dec,
-    url={https://gridstrategiesllc.com/wp-content/uploads/National-Load-Growth-Report-2024.pdf},
-}
-
```
zaneselvans: Probably didn't mean to nix this one?

zaneselvans (changes requested):
- I think you might need to pull in the changes from `main`.
- Need to update an ID for the testimony entry I think?

zaneselvans (approved):

PR #4580, opened 2025-08-29
Fix

cmgosnell:
# Overview

Attempts to fix the bug discussed in [this discussion](https://github.com/orgs/catalyst-cooperative/discussions/4558).

## What did you change?
.... I can successfully identify and remove the records that are rarely a non-total utility type during the transformation process. But this a) breaks the `reconcile_table_calculations` at the end of the transform step... and b) i'm now not sure if this solves the problem.

there are two places in our transform step where we add in dimensions into the calculation component process....
1) during the transform step: in `reconcile_table_calculations` w/ `_add_intra_table_calculation_dimensions`
2) while making `_core_ferc1_xbrl__calculation_components`

## Documentation

# Testing

## To-do list

### DIFF
--- a/src/pudl/transform/ferc1.py
+++ b/src/pudl/transform/ferc1.py
@@ -5132,6 +5132,60 @@ class IncomeStatementsTableTransformer(Ferc1AbstractTableTransformer):
         raw_dbf = super().process_dbf(raw_dbf)
         return raw_dbf
 
+    def remove_rare_utility_type_subdimensions_rows(
+        self: Self, df: pd.DataFrame
+    ) -> pd.DataFrame:
+        """Remove the rare, non-total utility types when all values are duplicated.
+
+        This data isn't incorrect, it just interferes with how we process the calculations
+        embedded within these tables.
+
+        """
+        xbrl_factoid = "income_type"
+        dimension_col = "utility_type"
+        idx = ["utility_id_ferc1", "report_year", xbrl_factoid]
+
+        # first we need to find the utility_types where there are mostly
+        # only totals. Which is to say there are a few rare non-total
+        # utility_type's.
+        tot_mask = df[dimension_col] == "total"
+        util_type_count = pd.merge(
+            pd.DataFrame(df.loc[tot_mask, xbrl_factoid].value_counts()),
+            pd.DataFrame(df.loc[~tot_mask, xbrl_factoid].value_counts()),
+            on=xbrl_factoid,
+            how="outer",
+            suffixes=("_total", "_other"),
+        )
+        mostly_totals = util_type_count[
+            util_type_count.count_other.notnull()
+            & ((util_type_count.count_other / util_type_count.count_total) < 0.80)
+        ].index
+
+        # add a utility type count column. Because we only care about this when there
+        # are more than one util type
+        df.loc[:, "util_type_count"] = df.groupby(idx)[["dollar_value"]].transform(
+            "count"
+        )
+        mostly_totals_mask = df[xbrl_factoid].isin(
+            [fact for fact in mostly_totals if fact != "net_utility_operating_income"]
+        ) & (df.util_type_count > 1)
+
+        elec_utils = df[mostly_totals_mask][idx].drop_duplicates().set_index(idx)
+        mixed_typed_income = df.set_index(idx).loc[elec_utils.index]
+        if not (
+            actual_dupes := mixed_typed_income[
+                ~mixed_typed_income.duplicated(keep=False, subset=["dollar_value"])
+            ]
+        ).empty:
+            raise AssertionError(f"Ah we found dupes:\n{actual_dupes}")
+
+        return pd.concat(
+            [
+                df[~mostly_totals_mask],
+                mixed_typed_income[mixed_typed_income[dimension_col] == "total"],
+            ]
+        )
+
     def transform_main(self: Self, df: pd.DataFrame) -> pd.DataFrame:
         """Drop duplicate records from f1_income_stmnt.
 
@@ -5147,7 +5201,9 @@ class IncomeStatementsTableTransformer(Ferc1AbstractTableTransformer):
                 & (df.income_type == "net_utility_operating_income")
             )
         ]
-        return apply_pudl_dtypes(df, group="ferc1")
+        return apply_pudl_dtypes(df, group="ferc1").pipe(
+            self.remove_rare_utility_type_subdimensions_rows
+        )
 
 
 class RetainedEarningsTableTransformer(Ferc1AbstractTableTransformer):


src/pudl/transform/ferc1.py
```diff
@@ -5132,6 +5132,60 @@ def process_dbf(self: Self, raw_dbf: pd.DataFrame) -> pd.DataFrame:
...
+            pd.DataFrame(df.loc[tot_mask, xbrl_factoid].value_counts()),
+            pd.DataFrame(df.loc[~tot_mask, xbrl_factoid].value_counts()),
+            on=xbrl_factoid,
+            how="outer",
+            suffixes=("_total", "_other"),
+        )
+        mostly_totals = util_type_count[
+            util_type_count.count_other.notnull()
+            & ((util_type_count.count_other / util_type_count.count_total) < 0.80)
+        ].index
```
zaneselvans: less than 80% doesn't feel particularly rare.

src/pudl/transform/ferc1.py
```diff
@@ -5132,6 +5132,60 @@ def process_dbf(self: Self, raw_dbf: pd.DataFrame) -> pd.DataFrame:
...
+        # only totals. Which is to say there are a few rare non-total
+        # utility_type's.
+        tot_mask = df[dimension_col] == "total"
+        util_type_count = pd.merge(
+            pd.DataFrame(df.loc[tot_mask, xbrl_factoid].value_counts()),
+            pd.DataFrame(df.loc[~tot_mask, xbrl_factoid].value_counts()),
+            on=xbrl_factoid,
+            how="outer",
+            suffixes=("_total", "_other"),
+        )
```
zaneselvans: Having trouble visualizing what this is.

src/pudl/transform/ferc1.py
```diff
@@ -5132,6 +5132,60 @@ def process_dbf(self: Self, raw_dbf: pd.DataFrame) -> pd.DataFrame:
...
+        mostly_totals = util_type_count[
+            util_type_count.count_other.notnull()
+            & ((util_type_count.count_other / util_type_count.count_total) < 0.80)
+        ].index
+
+        # add a utility type count column. Because we only care about this when there
+        # are more than one util type
+        df.loc[:, "util_type_count"] = df.groupby(idx)[["dollar_value"]].transform(
+            "count"
+        )
```
zaneselvans: Is it impossible for there to be only one utility type and for it **not** to be `total`?

src/pudl/transform/ferc1.py
```diff
@@ -5132,6 +5132,60 @@ def process_dbf(self: Self, raw_dbf: pd.DataFrame) -> pd.DataFrame:
...
+        mostly_totals_mask = df[xbrl_factoid].isin(
+            [fact for fact in mostly_totals if fact != "net_utility_operating_income"]
```
zaneselvans: Why is this one fact excluded?

src/pudl/transform/ferc1.py
```diff
@@ -5132,6 +5132,60 @@ def process_dbf(self: Self, raw_dbf: pd.DataFrame) -> pd.DataFrame:
...
+        ) & (df.util_type_count > 1)
+
+        elec_utils = df[mostly_totals_mask][idx].drop_duplicates().set_index(idx)
+        mixed_typed_income = df.set_index(idx).loc[elec_utils.index]
+        if not (
+            actual_dupes := mixed_typed_income[
+                ~mixed_typed_income.duplicated(keep=False, subset=["dollar_value"])
+            ]
+        ).empty:
+            raise AssertionError(f"Ah we found dupes:\n{actual_dupes}")
```
zaneselvans: More context in the logging message will be helpful later. What kind of dupes did we find? Where did we find them?

zaneselvans (commented):
I'm having a hard time following the intention in the new function. I think I need a walkthrough + more comments.

src/pudl/transform/ferc1.py
krivard: the full count of that row would be count_other + count_total, so this is checking for  non-total as <40% of the full count

so looking for majority-total more than rarely-other

cmgosnell: I decided to a) adopt @krivard 's more readable count_other + count_total as the denominator & b) change the threshold to 20%. There are only a small number (~2) of xbrl_factoids that are up in the 40% zone. All of those are dollar_value duplicates but ya that doesn't really fit the spirit of this being rare and a first pass of making this work hence the change to a 20% threshold.

src/pudl/transform/ferc1.py
cmgosnell: i added some inline comments explaining

src/pudl/transform/ferc1.py
cmgosnell: not impossible. not the case in this table but not impossible at all.

src/pudl/transform/ferc1.py
cmgosnell: this has been changed. there was an instance of this one having true non-duplicative values but with the current "rarity" threshold and other configurations this isn't an issue anymore

cmgosnell:
i marked this as ready for review but I still need to update the row counts off a branch build. I'd like to get a re-review before I kick off builds -> new row counts so I don't have to do it twice. I updated from a local run using: 
`dbt_helper update-tables out_ferc1__yearly_detailed_income_statements --row-counts --clobber`

to get a sense but i'd like to be sure with a branch build

src/pudl/transform/ferc1.py
```diff
@@ -6322,6 +6266,100 @@ def table_to_column_to_check() -> dict[str, list[str]]:
...
+    # that's a little bit arbitrary... This was tested up to 40% but that seems
+    # too aggressive.
+    mostly_total_xbrl_factoids = util_type_count[
+        util_type_count.count_other.notnull()
+        & (
+            (
+                util_type_count.count_other
+                / (util_type_count.count_other + util_type_count.count_total)
+            )
+            < 0.40
```
zaneselvans: With the new (A / (A+B)) calculation should the threshold here be `< 0.20` now?

src/pudl/transform/ferc1.py
```diff
@@ -6322,6 +6266,100 @@ def table_to_column_to_check() -> dict[str, list[str]]:
...
+    idx = ["utility_id_ferc1", "report_year", xbrl_factoid]
+
+    # first we need to find the xbrl_factoid's where there are mostly only totals
+    # within the dimension_col. Which is to say there are a few rare non-total
+    # utility_type's.
+    tot_mask = df[dimension_col] == "total"
+    # build a dataframe of xbrl_factoid's with a column for the count of the records
+    # with utility_type total and another column with the count for all the other
+    # utility_type's
+    util_type_count = pd.merge(
```
zaneselvans: If we do generalize to other dimensions in other tables, we'll probably want to make `util_type_count` more generic to go along with `dimension_col`.

src/pudl/transform/ferc1.py
```diff
@@ -6322,6 +6266,100 @@ def table_to_column_to_check() -> dict[str, list[str]]:
...
+        [fact for fact in mostly_total_xbrl_factoids if "correction" not in fact]
+    ) & (df.util_type_count > 1)
+
+    mostly_totals_idx = (
+        df.loc[mostly_totals_mask, idx].drop_duplicates().set_index(idx).index
+    )
+    mixed_typed_income = df.set_index(idx).loc[mostly_totals_idx].reset_index()
+    # first remove the records with null non-total records
+    maybe_unique = mixed_typed_income[
+        ~(mixed_typed_income[dimension_col] != "total")
```
zaneselvans: Is there a reason for the double negative here?  Why `~(x != "total")` rather than `(x == "total")`?  Or were you trying to negate the combination of the two conditions?  If we want to drop all the records where dollar value is null, and we have a non-total dimension, would it be:

```py
maybe_unique = mixed_type_income[
    ~(
        (mixed_typed_income[dimension_col] != "total")
        & (mixed_typed_income["dollar_value"].isna())
    )
]
```

zaneselvans (changes requested):
I could be misunderstanding the calculations but a couple of codey things seemed like maybe they weren't aligned with the comments around them.

src/pudl/transform/ferc1.py
cmgosnell: good catch yes. ( i kept testing different thresholds and forgot to turn it back down)

src/pudl/transform/ferc1.py
cmgosnell: yea i didn't bc the more i preemptively generalize the more unreadable it seems to me... but i can

src/pudl/transform/ferc1.py
cmgosnell: oh good catch. yea i meant to put it all in () bc yea its a combo of these two.... which ended up with two factoids that actually have unique values.. there are three util types & these don't add up.
image

Based on the rest of the table I think the best thing to do it still to remove these non-total values so imma changed the assertion threshold.

src/pudl/transform/ferc1.py
zaneselvans: I don't think it needs to be swapped now necessarily just noticed that you'd generalized some parts and not others.

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -2823,37 +2823,37 @@ out_ferc1__yearly_detailed_balance_sheet_liabilities,2021,15018
...
+out_ferc1__yearly_detailed_income_statements,2015,22663
+out_ferc1__yearly_detailed_income_statements,2016,22391
+out_ferc1__yearly_detailed_income_statements,2017,22605
+out_ferc1__yearly_detailed_income_statements,2018,22513
+out_ferc1__yearly_detailed_income_statements,2019,22920
+out_ferc1__yearly_detailed_income_statements,2020,23105
+out_ferc1__yearly_detailed_income_statements,2021,43609
+out_ferc1__yearly_detailed_income_statements,2022,43096
+out_ferc1__yearly_detailed_income_statements,2023,43216
+out_ferc1__yearly_detailed_income_statements,2024,43148
```
zaneselvans, line 2856: Wow that's a lot more records! Like 10% more!

src/pudl/transform/ferc1.py
```diff
@@ -6333,21 +6336,26 @@ def remove_rare_utility_type_subdimensions_rows(df: pd.DataFrame) -> pd.DataFram
...
+                # total's here
+                & (maybe_unique[dimension_col] != "total")
+            ]
+        )
+        > 4
+    ):
         raise AssertionError(
             "Ah we found actually unique values within the records with xbrl_factoid's "
-            f"that have rare non-total {dimension_col} when we expected none"
+            f"that have rare non-total {dimension_col} when we expected only 4 records"
```
zaneselvans, line 6360: Do we know what's up with these 4 weirdos?

src/pudl/transform/ferc1.py
```diff
@@ -6333,21 +6336,26 @@ def remove_rare_utility_type_subdimensions_rows(df: pd.DataFrame) -> pd.DataFram
...
-            & (maybe_unique[dimension_col] != "total")
-        ]
-    ).empty:
+    if (
+        len(
+            actually_unique := maybe_unique[
+                ~maybe_unique.duplicated(keep=False, subset=idx + ["dollar_value"])
+                # bc we removed some of the null non-totals we've gotta leave out the
+                # total's here
+                & (maybe_unique[dimension_col] != "total")
```
zaneselvans: I'm definitely not fully understanding what's going on here.

krivard (commented):
okay so this is gnarly as heck but you know that

here's what it looks like you're doing:

1. start with a df indexed by utility id, report year, and xbrl factoid
2. find the ratio of total to nontotal entries for each xbrl factoid
3. make a list of xbrl factoids with <20% nontotal entries
4. throw out any xbrl factoids with "correction"
5. make a list of all the index buckets whose xbrl factoid is still on our list
6. keep the index buckets that include more than one utility type
6. select only those index buckets from df
7. drop any entries with null nontotal dollar values
8. drop any entries which duplicate utility id, report year, xbrl factoid, and dollar value
9. drop any total entries
10. if anything is left, scream

and it looks like what we want to know is

1. are there any nontotal utility type entries that
  A. come from a factoid that's at least 80% total utility type across all utilities+years, and
  B. have non-null dollar value, and
  C. do not show up as a total utility type entry with the same dollar value, utility id, and year?

is that close?

cmgosnell:
@krivard yea you totally got it. it is.... a little gnarly for sure.

src/pudl/transform/ferc1.py
cmgosnell: i screenshot them above...  its these two xbrl_factoids
image

This is the only instance in which a util/year/fact from these rarely non-total factoids have actually unique information in there non-total records. I am personally okay with effectively stranding these guys for the detailed/explosion tables considering how many more records we are getting in exchange.

zaneselvans: Ah sorry I missed that. Looks good.

PR #4611, opened 2025-09-10
Redeploy the Eel Hole in nightly builds

### LINKED ISSUE #4457: Migrate resource description metadata for ferc714 (7 tables, 460 words)

zaneselvans:
# Overview

- Redeploy the Eel Hole / PUDL Viewer service in Cloud Run as soon as the new S3 data has been distributed in a nightly build.
- Update error codes / slack message to include the outcome of the command.
- Exclude anything under `docker/*` from triggering the integration tests in CI, since none of that gets touched in the integration tests -- we have to do a nightly build or (depending) a workflow dispatch / versioned release to see what will happen. So better not to spend the GHA resources and need to wait 1+ hours.

Closes #4457

## Questions

Does this need to be contingent on the S3 data distribution having been successful?  If it fails somehow, restarting the Eel Hole won't be harmful -- it'll still point at the last successful nightly outputs.  And if the nightly outputs are hosed somehow, that will break the Eel Hole even if we don't redeploy.  So it seems like it doesn't need to be contingent.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -32,6 +32,7 @@ jobs:
           predicate-quantifier: "every"
           filters: |
             code:
+              - '!docker/*'
               - '!docs/**.rst'
               - '!docs/**.md'
               - '!docs/**.pdf'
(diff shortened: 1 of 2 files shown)


.github/workflows/pytest.yml
```diff
@@ -32,6 +32,7 @@ jobs:
           predicate-quantifier: "every"
           filters: |
             code:
+              - '!docker/*'
```
jdangerx, line 35: I suppose that, while changes in `docker/*` can certainly break the builds, mostly CI won't catch that

zaneselvans: We attempt to build the Docker container, and we have shellcheck linting the scripts, so I think we will catch the real dumdum mistakes (which are checked outside of the integration tests) but yeah right now regardless of what we do here the only real test is attempting to do a deployment.

PR #4613, opened 2025-09-11
feat: add pudl viewer *deployer* service account + WIF so we can use it from GHA

jdangerx:
# Overview

Companion to https://github.com/catalyst-cooperative/eel-hole/pull/72

We need to set up a service account so that the eel-hole deployment GHA workflow can actually do the deploy.

## What did you change?

* add service account
* reformat some terraform stuff
* add a permission that was necessary for the nightly build service account as well ("permission to be a different service account" - i.e., for the deployment service account to be different from the deployed)

## Documentation

# Testing

I *ran* `terraform apply` already since the plan looked safe (just adding new service accounts and permissions) and all the things that depend on these permissions/service accounts existing seem to work.

## To-do list

### DIFF
--- a/terraform/nightly_builds.tf
+++ b/terraform/nightly_builds.tf
@@ -6,16 +6,12 @@ resource "google_service_account" "nightly_build" {
   description = "This service account is used by the nightly and branch PUDL builds."
 }
 
-// allow nightly builds to deploy cloud run service (for pudl viewer)
-resource "google_project_iam_member" "nightly_build_cloud_run" {
+resource "google_project_iam_member" "nightly_build" {
+  for_each = toset([
+    "roles/run.developer", // update cloud run services
+    "roles/iam.serviceAccountUser", // the cloud run service can use a service account that is different from the nightly build one
+  ])
   project = var.project_id
-  role    = "roles/run.developer"
-  member  = google_service_account.nightly_build.member
-}
-
-// need to be able to *use* service accounts too, since the pudl viewer service SA is different from the nightly build SA
-resource "google_project_iam_member" "nightly_build_use_cloud_run_sa" {
-  project = var.project_id
-  role    = "roles/iam.serviceAccountUser"
+  role    = each.key
   member  = google_service_account.nightly_build.member
 }
--- a/terraform/pudl-viewer.tf
+++ b/terraform/pudl-viewer.tf
@@ -24,21 +24,16 @@ resource "google_service_account" "pudl_viewer_gha" {
   account_id   = "pudl-viewer-gha"
   display_name = "PUDL Viewer GitHub Actions Service Account"
 }
-resource "google_project_iam_member" "pudl_viewer_gha_artifact_registry" {
-  project = var.project_id
-  role    = "roles/artifactregistry.writer"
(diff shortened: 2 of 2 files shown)


terraform/nightly_builds.tf
```diff
@@ -0,0 +1,21 @@
+// nightly build service account
+resource "google_service_account" "nightly_build" {
+  account_id   = "deploy-pudl-vm-service-account"
+  display_name = "Batch Build Service Account"
+  // this was once used for the builds on the deploy-pudl VM. but now we just use it for Batch.
+  description = "This service account is used by the nightly and branch PUDL builds."
+}
+
+// allow nightly builds to deploy cloud run service (for pudl viewer)
+resource "google_project_iam_member" "nightly_build_cloud_run" {
```
jdangerx: May be nicer to do something more like


```terraform
resource "google_project_iam_member" "nightly_build_sa_permissions" {
  for_each = toset([
    "roles/run.developer",
    "roles/iam.serviceAccountUser",
  ])

  role   = each.key
  member = google_service_account.nightly_build.member
}
```

Let me poke at this a little bit more and see if that generates any new terraform plan situation.

We could even port all the other permissions we have on the nightly build machine into here someday :sparkling_heart:

jdangerx:
> Would establishing a naming convention make it easier to debug these in the future?
> 
> Right now we have
> 
>     * cloud_run vs deploy
> 
>     * use_cloud_run_sa vs use_cloud_run
> 
> 
> resource 	role
> nightly_build_cloud_run 	roles/run.developer
> nightly_build_use_cloud_run_sa 	roles/iam.serviceAccountUser
> pudl_viewer_gha_artifact_registry 	roles/artifactregistry.writer
> pudl_viewer_gha_deploy 	roles/run.developer
> pudl_viewer_gha_use_cloud_run 	roles/iam.serviceAccountUser

Yes, I think so - I consolidated all the different permissions under one name for each service account (so each will show up as `google_project_iam_member.<service_account_name>[<role name>]`. I also took the `_sa` suffix out since project IAM membership either applies to a service account or a user, and that should be pretty straightforward to disambiguate.

terraform/main.tf
```diff
@@ -71,17 +71,21 @@ module "gh_oidc" {
       attribute = "attribute.repository/catalyst-cooperative/mozilla-sec-eia"
     }
     "nrel-finito-inputs-gha" = {
-      sa_name   = "projects/${var.project_id}/serviceAccounts/${google_service_account.nrel_finito_inputs_gha.email}"
+      sa_name   = google_service_account.nrel_finito_inputs_gha.id
       attribute = "attribute.repository/catalyst-cooperative/nrel-fuel-and-industry-inputs"
     }
     "pudl-usage-metrics-dashboard-deploy-gha" = {
-      sa_name   = "projects/${var.project_id}/serviceAccounts/${google_service_account.pudl_usage_metrics_dashboard_deploy_gha.email}"
+      sa_name   = google_service_account.pudl_usage_metrics_dashboard_deploy_gha.id
```
zschira, line 78: Was the whole `projects/${var.project_id}/serviceAccounts/` just unnecessary?

PR #4625, opened 2025-09-18
Update gdal and fix row counts / dbt test that failed nightly build

zaneselvans:
# Overview

* GDAL 3.11.4 was released, and this updates to the new version.
* Also fixes some belated row-count changes from the final EIA-860 updates that failed the nightly builds on 2025-09-19.
* Re-lock dependencies & update pre-commit hooks so they are consistent with the conda environment.


krivard (approved):
Needs relocked but the gdal fixes look good.

dbt/models/sec10k/core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities/schema.yml
```diff
@@ -10,7 +10,7 @@ sources:
               config:
                 enabled: "{{ target.name == 'etl-full'}}"
               arguments:
-                min_value: 5544
+                min_value: 5543
```
zaneselvans, line 13: I guess this changed as a result of the EIA860 updates? It is an association table that links to EIA utilities.

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -1832,22 +1832,22 @@ core_nrelatb__yearly_technology_status,2021,131
...
-core_pudl__assn_utilities_plants,,19448
+core_pudl__assn_utilities_plants,,19484
 core_pudl__codes_data_maturities,,4
 core_pudl__codes_datasources,,13
 core_pudl__codes_imputation_reasons,,12
 core_pudl__codes_subdivisions,,69
-core_pudl__entity_plants_pudl,,20470
-core_pudl__entity_utilities_pudl,,17121
+core_pudl__entity_plants_pudl,,20506
+core_pudl__entity_utilities_pudl,,17145
```
zaneselvans, line 1850: I pulled these updated row counts form the failed nightly build output. I think what happened is @e-belfer belatedly updated the plant & utility mappings, but that it happened after the row counts had been captured, so they were a little out of sync.

.pre-commit-config.yaml
```diff
@@ -30,7 +30,7 @@ repos:
   ####################################################################################
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.13.0
+    rev: v0.13.1
```
zaneselvans, line 33: Just to keep it synced with the conda environment.

pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "catalystcoop.pudl"
 description = "An open data processing pipeline for US energy data"
 readme = { file = "README.rst", content-type = "text/x-rst" }
 authors = [{ name = "Catalyst Cooperative", email = "pudl@catalyst.coop" }]
-requires-python = ">=3.13,<3.14"
+requires-python = ">=3.13,<3.14.0a0"
```
zaneselvans, line 10: I've occasionally gotten some weird solves where it fails because it can't install Python 3.14.0rc2, which is spurious and obscures the real dependency issue, so I thought I would make this explicit.

src/pudl/metadata/classes.py
```diff
@@ -1957,7 +1957,7 @@ def _build_aggregation_report(self, df: pd.DataFrame, errors: dict) -> dict:
             freports[field.name] = {
                 "valid": stats["actual"] <= stats["tolerance"],
                 "stats": stats,
-                "errors": errors.get(field.name, None),
+                "errors": errors.get(field.name),
```
zaneselvans, line 1960: I think this is a fix from ruff via the pre-commit hooks.

Issue #2819, opened 2023-09-01
Improve performance of integration tests on GitHub


zaneselvans:
Our integration tests are painfully slow in CI, and are also currently running out of memory after running for about an hour, but we're on the verge of being able to change all that pretty dramatically.  In the past we have been unable to effectively split up the tests and run them in parallel because:
- they all depend on a resource intensive fixture that runs for a long time -- the one that builds the PUDL DB.
- there were many additional resource intensive fixtures that did computations using the `PudlTabl` output object.

Now we've got all but one of the output tables pulled into the DB, and will soon be able to deprecate `PudlTabl` and remove it from our tests entirely. Because all of the ETL steps are now being done within Dagster, that computation can be run in parallel very effectively -- the fast tests saturate the 10 cores on my laptop for the full duration of the tests, so we might even be able to use a 16 core GitHub runner effectively.

On a 10 core MacBookPro with 32GB of memory:
- FERC Extract: **2min 22sec** (DBF and XBRL extractions run serially, but XBRL uses multiple cores)
- Fast ETL: **16min 29sec** (fully parallelized, saturating all cores)
- `pytest --live-dbs test/integration`: **3min 21sec** (2min 10sec of which is reading from the DB)
- Total runtime: **22min 12sec**

```[tasklist]
- [ ] Switch to a [4CPU/16GB GitHub runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners) for integration test workflow to avoid running out of memory.
- [ ] Make Dagster run the ETL using all CPUs rather than in-process to speed up the ETL.
- [ ] Make building the PUDL DB its own pre-pytest task, rather than an expensive fixture.
- [ ] Figure out how to collect test coverage data when building the PUDL DB outside of pytest.
- [ ] Make the current `pudl_engine` fixture just return a connection engine pointing at the fresh DB.
- [ ] (optional) Use `pytest-xdist` or something similar to run the post-ETL integration tests in parallel.
- [ ] (optional) Restrict the years of data processed by `imputed_hourly_demand_ferc714` (by far the longest running asset, but also parallelized internally, so it may not end up controlling runtime)
```

zaneselvans:
Maybe converting the FERC data and running the PUDL ETL outside of pytest while collecting coverage is as easy as doing what we already do inside `tox -e nuke`?

```bash
coverage run --append src/pudl/convert/ferc_to_sqlite.py src/pudl/package_data/settings/etl_fast.yml
[...]

PR #4627, opened 2025-09-20
Experiment with 3rd-party GitHub Runners

zaneselvans:
# Overview

This PR switches our `ci-integration` job to use a 3rd party runner from blacksmith.  I took notes on the process in the comments on #2819.

- The cost per minute is half that of the GitHub runners.
- For some reason, our CI runs successfully on one of their runners with 16GB rather than 32GB of RAM.
- The integration tests run in about 55 minutes rather than 80.
- They offer 3000 free minutes per month. We use about 3000 minutes per month.
- So this should save us around $125/mo in GitHub fees.
- Even without the free minutes, we'd probably cut our CI costs from ~$75-100/mo to ~$20/mo and it also reduces individual runtime by about 25%

Other changes

- Add the `workflow-telemetry-action` to the `ci-integration` job, so we can see what its CPU and RAM usage are like.
- Switch to using relative file paths when generating coverage, to avoid discrepancies between the GitHub and Blacksmith coverage reports.
- Add some `du -sh` and `df` and `find` commands to the `ci-integration` workflow so we can see how much disk space we are using. Probably too much of this right now. [Re-running](https://github.com/catalyst-cooperative/pudl/actions/runs/17918836931/job/50948139451) `ci-integration` now to see what makes sense to keep.

# Testing

- Ran `pytest` workflow manually with a `workflow_dispatch`

## To-do list

### DIFF
--- a/.github/actionlint.yml
+++ b/.github/actionlint.yml
@@ -1,6 +1,7 @@
 # these are GitHub-hosted but for some reason actionlint couldn't find them
 self-hosted-runner:
   labels:
+    - ubuntu-latest-4core
     - ubuntu-latest-8core
     - blacksmith-4vcpu-ubuntu-2404
     - blacksmith-8vcpu-ubuntu-2404
--- a/.github/actionlint.yml
+++ b/.github/actionlint.yml
@@ -2,3 +2,5 @@
 self-hosted-runner:
   labels:
     - ubuntu-latest-8core
+    - blacksmith-4vcpu-ubuntu-2404
+    - blacksmith-8vcpu-ubuntu-2404
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -177,9 +177,21 @@ jobs:
       - name: Collect Workflow Telemetry
         uses: catchpoint/workflow-telemetry-action@v2
 
+      - name: Show disk space usage 1
+        run: |
+          df -h
+          du -sh /*
+          du -sh /home/runner/*
+
       - uses: actions/checkout@v5
         with:
           fetch-depth: 2
+
+      - name: Show disk space usage 2
+        run: |
+          df -h
+          du -sh /*
+          du -sh /home/runner/*
 
       - name: Install Conda environment using mamba
         uses: mamba-org/setup-micromamba@v2
@@ -187,6 +199,12 @@ jobs:
           environment-file: environments/conda-lock.yml
           environment-name: pudl-dev
           cache-environment: true
+
+      - name: Show disk space usage 3
+        run: |
+          df -h
+          du -sh /*
+          du -sh /home/runner/*
 
       - name: Log environment details
         run: |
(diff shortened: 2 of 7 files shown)


.github/actionlint.yml
```diff
@@ -1,4 +1,7 @@
 # these are GitHub-hosted but for some reason actionlint couldn't find them
 self-hosted-runner:
   labels:
+    - ubuntu-latest-4core
     - ubuntu-latest-8core
+    - blacksmith-4vcpu-ubuntu-2404
+    - blacksmith-8vcpu-ubuntu-2404
```
zaneselvans, line 7: To satisfy the action linter.

.github/workflows/pytest.yml
```diff
@@ -161,7 +161,8 @@ jobs:
           path: .coverage
 
   ci-integration:
-    runs-on: ubuntu-latest-8core
+    # runs-on: ubuntu-latest-8core
+    runs-on: blacksmith-4vcpu-ubuntu-2404
```
zaneselvans, line 170: It wasn't *quite* a one-line change, but close.

pyproject.toml
```diff
@@ -362,6 +362,7 @@ zip = ">=3.0"
 
 [tool.coverage.run]
 source = ["src/pudl/", "test/integration/", "test/unit/"]
+relative_files = true
```
zaneselvans, line 365: Needed to avoid incompatibilities between the coverage reports generated on GitHub and non-GitHub runners, which put the code in different directories in the runner's home directory.

.github/workflows/pytest.yml
```diff
@@ -172,19 +173,36 @@ jobs:
     defaults:
       run:
         shell: bash -l {0}
-
     steps:
+      - name: Collect Workflow Telemetry
+        uses: catchpoint/workflow-telemetry-action@v2
```
zaneselvans, line 183: This generates plots of CPU usage, RAM usage, Disk I/O, Network I/O. Unfortunately, if the runner gets summarily shut down they aren't generated. Also unfortunately, it doesn't show disk usage.

.github/workflows/pytest.yml
```diff
@@ -172,19 +173,36 @@ jobs:
...
+
+      - name: Show disk space usage 1
+        run: |
+          df -h
+          du -sh /home/runner/*
```
zaneselvans: I probably went overboard with these.  Also, disk isn't our limiting factor right now, so we can probably cut most of these.

jdangerx: I think these are useful! I would keep 3, 5, and 6. 1, 2, and 4 don't really seem actionable to me...

.github/workflows/pytest.yml
jdangerx: Let's chuck a comment in here explaining the difference between ubuntu and blacksmith runners.

jdangerx (approved):
Glad this was a small change in the end, after all that investigation! Couple small things if you want to clean things up. Otherwise good to go!

.github/workflows/pytest.yml
zaneselvans: I found the `find` output a bit too much. I've cut it down to the 3 that you recommended and limited it to the `du -sh` calls.  Looks like the pytest output is ~6.6 GB, so we're actually very close to the 14 GB max that we could deal with on a free runner, unless we shuffle something around.  Cutting the SQLite output would probably save a lot of the output size.

.github/workflows/pytest.yml
zaneselvans: Comment added, with a link to the dashboard: https://app.blacksmith.sh/catalyst-cooperative/

zaneselvans:
Hopefully it still works in the merge queue!

PR #4507, opened 2025-07-31
Ferc1 2024 2nd pass

aesharpe:
# Overview

Closes #4493
Note that #4189 is still an issue and OOS for this PR

## What problem does this address?

Adds late respondents to the FERC 1 2024 data.

## What did you change?

- Add newest FERC 1 archive DOI
- Map new FERC 1 utilities
- Update row counts for FERC 1 tables

## Documentation

Make sure to update relevant aspects of the documentation:

- [x] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html): reference the PR and related issues.
- [ ] Update relevant Data Source jinja templates (see `docs/data_sources/templates`).
- [ ] Update relevant table or source description metadata (see `src/metadata`).
- [ ] Review and update any other aspects of the documentation that might be affected by this PR.

# Testing

**How did you make sure this worked? How can a reviewer verify this?**

- Pull the most recent nightly build, update alembic, and run `group:raw_ferc1+` in dagster to materialize all assets upstream of FERC 1.
- Run `dbt_helper validate --asset-select "group:raw_ferc1+"`

## To-do list

- [x] If updating analyses or data processing functions: make sure to update row count expectations in `dbt` tests.
- [ ] Run `make ci` locally to ensure that the merge queue will accept your PR. **--> I tried this and it failed and I have no idea how to interpret the failures...**
- [x] Review the PR yourself and call out any questions or issues you have.
- [x] For PRs that change the PUDL outputs significantly, run the full ETL locally and then [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/data_validation.html) using dbt. If you can't run the ETL locally then run the `build-deploy-pudl` GitHub Action manually and ensure that it succeeds.

Issue #4375, opened 2025-06-30
Add data source documentation for Census PEP


zaneselvans:

e-belfer:
@cmgosnell was the one who integrated this dataset, not sure if she wants to write these docs up?

zaneselvans:
Ah okay! I thought it was your data baby. Probs she's the right one then.

PR #4630, opened 2025-09-23
Update FERC DOIs

e-belfer:
# Overview

We wanted to update all FERC DOIs since our last update in #4531, #4507 and #4530.

In doing so, I learned that we changed the FERC Form 2 archiver in https://github.com/catalyst-cooperative/pudl-archiver/pull/840, which made the partition for DBT data change from None to "all". This broke extraction for FERC Form 2.

## What did you change?
- Update the DOIs for all new FERC archives
- Update the definition of a valid partition for DBF data to look for `"all"` partitions rather than `None` partitions.
- Update the row counts
- Also updated `make docs-build` to remove both sets of census data docs before building, now that we have #4375 

## Documentation

# Testing

## To-do list


e-belfer:
Error observed in logs for the [attempted build](https://console.cloud.google.com/storage/browser/builds.catalyst.coop/2025-09-22-1948-a04a6b165-update-ferc-dois):

`The above exception was caused by the following exception:
KeyError: "Multiple resources found for ferc2: {'year': 1996, 'data_format': 'dbf'}"`

@zschira @krivard Does this ring a bell from any of the changes made recently to the FERC XBRL extraction? As far as I can see nothing about the 1996 data itself has changed from the run where they were updated last (https://github.com/catalyst-cooperative/pudl-archiver/issues/783).

zschira:
@e-belfer I believe this is an unexpected side effect from my recent archiver updates to avoid falsely detecting changes with FERC archives. One change that was included in that work was removing the `part` partition key from FERC 2 monolithic DBF files, but for a couple years we have both monolithic and split DBF files, so now the `Datastore` can't distinguish between these. I created a pr [here](https://github.com/catalyst-cooperative/pudl-archiver/pull/840) that should hopefully fix this.

src/pudl/extract/ferc2.py
```diff
@@ -48,7 +48,7 @@ def finalize_schema(self: Self, meta: sa.MetaData) -> sa.MetaData:
     @staticmethod
     def is_valid_partition(fl: dict[str, Any]):
         """Drops partition with non-empty `part` fields."""
-        return fl.get("part") is None
+        return fl.get("part") == "all"
```
aesharpe, line 51: Just out of curiosity what is this change doing?

e-belfer: Ah sorry, I completely failed to document this in the PR. I'll note what's changing above.

PR #4631, opened 2025-09-23
EIA AEO multi year capacity (plus 2025)

### LINKED ISSUE #4591: Enable multiple years of AEO data
### Overview

The current AEO data pipeline only accommodates one report year's worth of data. We'd like to change that so we can have multiple report years and so compare projections over time. We'll use integrating the 2025 projections as a way to test this capability. 

### Success Criteria
 - [ ] We're able to integrate future years of AEO data without replacing the old data
 - [ ] We've integrated 2025 projections into PUDL.

### Out of Scope
- We are _not_ integrating historic data, just future data.

aesharpe:
# Overview

Closes #4591.

AEO extractor currently only has the capacity to process 1 year of data. This PR makes it so we can integrate multiple years of data. 

This PR also adds the newest available year of AEO data.

## What did you change?

- Update release notes
- Add new alembic migration
- Add new enums from 2025 data
- Fix erroneous duplicates that EIA said they wouldn't fix until the next AEO publication...
- Enable multiple years of AEO data
- Update expected number of categorical values

## Documentation

# Testing

## To-do list

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,7 @@ repos:
       - id: check-toml
       - id: check-case-conflict # Avoid case sensitivity in file names.
       - id: debug-statements # Watch for lingering debugger calls.
+        exclude: "^devtools/check_against_nightly.py$"
       - id: mixed-line-ending # Use Unix line-endings to avoid big no-op CSV diffs.
         args: ["--fix=lf"]
       - id: trailing-whitespace
--- a/devtools/check_against_nightly.py
+++ b/devtools/check_against_nightly.py
@@ -1,0 +1,65 @@
+#! /usr/bin/env python3
+
+"""Check a local parquet file or parquet files against the nightly build.
+
+Usage:
+
+    ./check_against_nightly.py <local_targets>
+
+Example usage:
+
+Check against S3 nightly:
+
+    ./check_against_nightly.py core_eia930__*.parquet
+
+Check against local copy:
+
+    ./check_against_nightly.py --reference-dir $PUDL_OUTPUT/../nightly/parquet core_eia930__*.parquet
+
+Check against GCS build outputs (requires gcloud application default credentials):
+
+    ./check_against_nightly.py --reference-dir gs://builds.catalyst.coop/2025-10-11-0605-a531a9a7f-main/parquet core_eia930__*.parquet
+"""
+
+from pathlib import Path
+
+import click
+import polars as pl
+from polars.testing import assert_frame_equal
+
+
+@click.command(help=__doc__)
+@click.argument("local-targets", nargs=-1, type=click.Path())
+@click.option(
+    "--pdb", is_flag=True, help="Drop into a breakpoint if any dataframes don't match."
+)
+@click.option(
+    "--reference-dir",
+    type=str,
+    default="https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly",
+    help="The directory containing the parquet files to check against. Defaults to S3 nightly.",
+)
+def cli(local_targets: tuple[Path], pdb: bool, reference_dir):
+    """Compare local Parquet files against nightly."""
+    local_paths = [Path(target) for target in local_targets]
+    click.echo(
+        f"Comparing {', '.join(p.stem for p in local_paths)} against those in {reference_dir}..."
+    )
+    for local_path in local_paths:
+        click.echo(f"Comparing {local_path.stem}...", nl=False)
+        local = pl.scan_parquet(local_path)
+        reference = pl.scan_parquet(f"{reference_dir}/{local_path.name}")
+        try:
+            assert_frame_equal(local, reference)
+        except AssertionError as e:
+            if pdb:
+                breakpoint()
+            click.echo(
+                f"MISMATCH\n{e}".replace("left", "local").replace("right", "reference")
+            )
+        else:
+            click.echo("OK")
+
+
+if __name__ == "__main__":
+    cli()
--- a/devtools/memory_profile.py
+++ b/devtools/memory_profile.py
@@ -42,6 +42,9 @@ def materialize_assets(asset_selection: str) -> None:
     asset_selection_with_multi_asset_siblings = dg.AssetSelection.from_string(
         asset_selection
     ).required_multi_asset_neighbors()
+    click.echo(
+        f"Found {asset_selection_with_multi_asset_siblings.resolve(defs.resolve_asset_graph())}"
+    )
 
     full_etl_job = defs.get_job_def("etl_full")
     dg.materialize(
@@ -72,7 +75,12 @@ def cli(asset_selection, aggregate, directory):
     file_format = (
         FileFormat.AGGREGATED_ALLOCATIONS if aggregate else FileFormat.ALL_ALLOCATIONS
     )
-    with Tracker(file_name=profile_location, follow_fork=True, file_format=file_format):
+    with Tracker(
+        file_name=profile_location,
+        follow_fork=True,
+        file_format=file_format,
+        native_traces=True,
+    ):
         materialize_assets(asset_selection)
     run(["/usr/bin/env", "memray", "flamegraph", str(profile_location)])  # noqa: S603
 
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -82,6 +82,8 @@ Dev tooling
   dataset, but we anticipate using these tools in other contexts to address performance
   bottlenecks. See issue :issue:`4663` and PR :pr:`4676` for the conversion of EPA
   CEMS from Dask to Polars.
+* We also added ``devtools/check_against_nightly.py`` to quickly compare local
+  Parquet outputs with those from the nightly builds.
 
 .. _release-v2025.10.0:
 


src/pudl/extract/eiaaeo.py
```diff
@@ -438,21 +444,28 @@ def raw_eiaaeo(context: AssetExecutionContext):
...
-        yield Output(
-            value=taxonomy.get_table(name_to_number[asset_name]), output_name=asset_name
-        )
+        all_years_data = []
+        for year in years:
+            filename = f"AEO{year}.txt"
+            with ds.get_zipfile_resource("eiaaeo", year=year).open(
+                filename, mode="r"
+            ) as aeo_raw:
+                taxonomy = AEOTaxonomy(io.TextIOWrapper(aeo_raw))
```
krivard, line 466: tricky -- it looks like this will read each AEO file multiple times, once for each selected asset. We have more assets than years right now, so that's not great.

But if we flip the order of the loops, so that we go through years in the outer loop and assets in the inner loop, we'll need to hold a bunch of stuff in memory until all the years have been processed before we can yield the outputs. Also not great.

What's your sense of how resource-intensive aeo extraction is? if it's slow, or a memory hog, we can look at that tradeoff more closely, but if it's speedy and tiny then maybe it doesn't matter and we can stick with what's easiest to read

src/pudl/metadata/enums.py
```diff
@@ -396,14 +402,17 @@
...
     "pipeline_natural_gas",
     "propane",
+    "purchased_electricity",
     "renewable_energy",
     "residual_fuel_oil",
     "steam_coal",
     "total",
+    "total_energy",  # this could be the same as total?
 ]
 """Energy use types from Table 2 of EIA AEO 2023.
```
krivard:
```suggestion
"""Energy use types from Table 2 of EIA AEO 2023-2025.
```

src/pudl/metadata/enums.py
```diff
@@ -337,6 +337,9 @@
 
 MODEL_CASES_EIAAEO: list[str] = [
     "aeo2022",
+    "aeo2023",
+    "alternative_electricity",
+    "alternative_transportation",
```
krivard: would there be utility to labeling these with which year added them, or is that unlikely to come up in debugging? or maybe if it does come up we just use `git annotate`

src/pudl/transform/eiaaeo.py
```diff
@@ -150,18 +157,13 @@ def filter_enrich_sanitize(
...
         **{
             colname: __sanitize_string(enriched[colname])
             for colname in enriched.columns
-            if colname not in {"projection_year", "value"}
-        }
-    ).pipe(
-        lambda df: df.assign(
-            report_year_series=df.report_year_series.str.replace("aeo", ""),
-            value=pd.to_numeric(df.value.str.replace("- -", "")),
-        )
```
krivard, line 164: are these no longer needed?

src/pudl/transform/eiaaeo.py
```diff
@@ -171,12 +173,11 @@ def filter_enrich_sanitize(
...
     )
 
-    # TODO 2024-04-20: one day we'll include report year in the extract step -
-    # at that point we'll have to compare that with the report year extracted
-    # from the series name.
-    return sanitized.drop(columns=["region_series", "case_series"]).rename(
-        columns={"region_category": "region", "report_year_series": "report_year"}
-    )
+    # Dropping report_year_series here because we don't need it but we MIGHT want it later
+    # So not fully removing it from get_series_info()
```
krivard, line 177: do we have a thing that checks whether report_year from extraction matches report_year_series?

src/pudl/extract/eiaaeo.py
```diff
@@ -311,7 +314,7 @@ def __sanitize(self, s: str) -> str:
         return re.sub(self.__sanitize_re, "_", s.lower().strip().replace(" : ", "__"))
 
     def __series_to_records(
-        self, series_id: str, potential_parents: set[int]
+        self, series_id: str, potential_parents: set[int], year: int
```
krivard: Slight preference for using `report_year` instead of `year` for this name, if only because I've been tripped up by the report vs model case vs prediction year values before

krivard (commented):
early look!
the extraction asset may need some more thought but it depends on how annoying it is to materialize. if we decide it's worth it, there are a bunch of options, not just flipping the loop order, so lmk

src/pudl/extract/eiaaeo.py
aesharpe: Hmm this is a good point. Memory stuff is not my forte -- do you know what I should be looking at to get a sense of whether it's resource intensive or not? I can time how long both versions of the loop take?

src/pudl/metadata/enums.py
aesharpe: More info is never a bad thing! I can add a comment next to them

src/pudl/extract/eiaaeo.py
aesharpe: great point! fixing

src/pudl/transform/eiaaeo.py
aesharpe: yeah, now the `report_year` is a column of it's own and we don't need to make it from the aeo2025 field.

src/pudl/transform/eiaaeo.py
```diff
@@ -554,6 +552,18 @@ def core_eiaaeo__yearly_projected_energy_use_by_sector_and_type(
...
+    # labeled as Electricity Imports, causing a duplicate primary key with the actual
+    # Electricity Imports data. Get the correct series name from the series_id for now.
+    if not sanitized.drop(columns=["value", "series_id"]).duplicated().any():
+        raise AssertionError(
+            "No more duplicates! Yay! AEO fixed the problem and you can remove this code."
+        )
+
+    dupe_mask = sanitized.drop(columns=["value", "series_id"]).duplicated(keep=False)
+    byprdh2_mask = sanitized.series_id.str.contains("byprdh2")
+    sanitized.loc[dupe_mask & byprdh2_mask, "variable_name"] = "byproduct_hydrogen"
```
aesharpe, line 567: I thought about doing something more generic and embedding it into the `filter_enrich_sanitize()` function applied to all tables, but it felt like overkill given that it is something EIA says they will fix next year. 

If we add more tables (others that might have other types of duplicates) we'll have to address that, but I didn't want to do anything too fancy just yet.

src/pudl/transform/eiaaeo.py
```diff
@@ -580,6 +590,7 @@ def core_eiaaeo__yearly_projected_energy_use_by_sector_and_type(
             columns=[
                 "topic",
                 "units",
+                "series_id",
```
aesharpe, line 595: I added this so we could use the `series_id` to fix the Production Hydrogen dupes

src/pudl/transform/eiaaeo.py
```diff
@@ -676,24 +690,24 @@ class AeoCheckSpec:
...
+    "model_case_eiaaeo": 20,
     "projection_year": 30,
     "electricity_market_module_region_eiaaeo": 26,
 }
 check_specs = [
     AeoCheckSpec(
         name="gen_in_electric_sector_by_tech",
         asset="core_eiaaeo__yearly_projected_generation_in_electric_sector_by_technology",
-        num_rows_by_report_year={2023: 166972},
+        num_rows_by_report_year={2023: 166972, 2025: 126750},
```
zaneselvans: We should probably remove all of these row count checks rather than updating them, since this is being checked in a standardized way with dbt now.

src/pudl/transform/eiaaeo.py
aesharpe: correct! We get report_year from somewhere else now

src/pudl/transform/eiaaeo.py
aesharpe: no, but I'm not too worried about it because there are only two years and it's pretty obvious if it's wrong.

src/pudl/transform/eiaaeo.py
aesharpe: That's a fair point

src/pudl/extract/eiaaeo.py
```diff
@@ -362,15 +370,17 @@ def __series_to_records(
...
+                "report_year": report_year,  # Add the report year here
             }
             for d in series["data"]
         )
         return records
 
-    def get_table(self, table_number: int) -> pd.DataFrame:
-        """Get a specific table number as a DataFrame."""
+    def get_table(self, table_number: int, report_year: int) -> pd.DataFrame:
+        """Get a specific table number as a DataFrame, including the report year."""
```
krivard:
```suggestion
        """Get a specific table number and report year as a DataFrame."""
```

src/pudl/transform/eiaaeo.py
krivard: the way you have it here is very tidy 👍

src/pudl/transform/eiaaeo.py
```diff
@@ -726,11 +734,6 @@ def make_check(spec: AeoCheckSpec) -> AssetChecksDefinition:
     @asset_check(asset=spec.asset, blocking=True)
     def _check(df):
         errors = []
-        for year, expected_rows in spec.num_rows_by_report_year.items():
-            if (num_rows := len(df.loc[df.report_year == year])) != expected_rows:
-                errors.append(
-                    f"Expected {expected_rows} for report year {year}, found {num_rows}"
-                )
```
krivard, line 733: yessssss get gone, row count asset checks

src/pudl/metadata/enums.py
```diff
@@ -396,16 +406,19 @@
...
     "petrochemical_feedstocks",
     "pipeline_fuel_natural_gas",
     "pipeline_natural_gas",
     "propane",
+    "purchased_electricity",
     "renewable_energy",
     "residual_fuel_oil",
     "steam_coal",
     "total",
+    "total_energy",  # this could be the same as total?
```
krivard, line 419: is this a question for future-us or does it need to be resolved before merging?

aesharpe: Question for future us

Issue #4586, opened 2025-09-01
Show or summarize available years in table descriptions


krivard:
Add a new component to table descriptions called `Availability`.

For this first round, success criteria is:

- [ ] The availability component shows the most recent year where data is available
- [x] Default availability is computed from the data source partitions
- [x] Manual override is available for discontinued tables

The next round (issue TBD) will figure out what to do with earliest-year-available, since that's trickier to figure out.

<details><summary>Former definition:</summary>

Add a new component to table descriptions called `Availability`.

The availability component must:

- [ ] always be present; falling back to a default if unknown
- [ ] only be present when known

The availability of a table is known when:

- [ ] manually specified
- [ ] computed from the materialized asset
- [ ] computed from the source metadata

Authors should expect to manually specify the availability component:

- [ ] never
- [ ] when a table is retired and won't be updated again
- [ ] at every release, when a table's availability differs from the default

The default availability for a table should be based on:

- [ ] the source metadata (use union when multiple sources are listed)
- [ ] the source metadata (use intersection when multiple sources are listed)
- [ ] the materialized asset (tricky; would need more discussion)

</details>

krivard:
Challenge: EIA API doesn't have year partitions. In this case should we:

- [ ] Show most-recent data as "Unknown"
- [ ] Not show most-recent data at all

Example:

> <img width="668" height="196" alt="Image" src="https://github.com/user-attachments/assets/ac123809-f38a-4d82-a9ea-5d24d7c02c5f" />

vs

> <img width="668" height="182" alt="Image" src="https://github.com/user-attachments/assets/d40c1b68-5c49-41ad-b064-a5bf81ce0a55" />

krivard:
Challenge:

In normal operation, table description text is resolved and built during [`Resource.dict_from_resource_descriptor()`](https://github.com/catalyst-cooperative/pudl/blob/b9a869f88ba23159d05dd8f671611c45cc08266a/src/pudl/metadata/classes.py#L1532), after most entity mentions from the raw metadata dict have been dereferenced from strings to their proper objects.

[...]

Issue #4327, opened 2025-06-17
EIA 860M data provenance inaccurately reflected in current DAG


e-belfer:
### Describe the bug

Right now, we have `raw_eia860m` data, and the only asset downstream of this is `_core_eia860m__changelog_generators`. However, we're also integrating the latest month of EIA 860M data into the `raw_eia860` assets in `extract_eia860()`:

```
raw_eia860__all_dfs = pudl.extract.eia860m.append_eia860m(
            eia860_raw_dfs=raw_eia860__all_dfs, eia860m_raw_dfs=eia860m_raw_dfs
        )
```

What does this mean? Our Dagster asset graph (DAG) doesn't accurately represent EIA 860M data as being upstream of the EIA 860 assets, or the many assets downstream of them. It's also not particularly clear where the EIA 860M data is getting merged into the 860 data, and the way we're handling the input data here is non-standard with the rest of our system.

### Bug Severity
How badly is this bug affecting you?
 - **Low:** The bug isn't causing me problems, but something's still wrong here.

### Expected behavior
A clear and concise description of what you expected to happen, or what you expected the data to look like.

- We should remove the `eia_settings.eia860.eia860m` setting from the ETL, as we never set it to be False and it's not clear what the use case is here.
- We should add `raw_eia860m__all_dfs()` as an input asset to the `extract_eia860()` multi_asset, with the step filtering the last year of data from each dataframe rather than reextracting the 860M data from scratch.

PR #4632, opened 2025-09-24
Show most recent data available as a component in table descriptions

krivard:
# Overview

Addresses #4586.

## What did you change?

- Add a new description component called "availability"
  - configured using key "availability_text" xor "availability_offset" (only one is allowed)
  - defaults to the most-recent partition of row counts for the table. if row counts are not partitioned, grabs the most-recent partition of the table's source. if availability_offset is specified, offsets the partition we got from the source by that many partitions (forward or back). if more than one source is listed, or if the only source partitions are non-temporal, return unknown.
  - if availability is unknown, hides the availability component
- Add a new method to DataSource that lists all temporal partitions
- Add some new functions in a lookup table in pudl.metadata.descriptions for doing partition arithmetic
- Refactor Resource.dict_from_resource_descriptor to pull some useful stuff into a separate function that can be called by both Resource.dict_from_resource_descriptor and the unit tests
- Refactor to move DBT_DIR to somewhere that doesn't import `pudl.metadata.classes`; `pudle.workspace.setup` seemed fine

Sample rendered output:

> image

Previously reviewed in this PR, but already merged to main via a rebase:
- ~~Refactor description machinery [according to plan](https://docs.google.com/document/d/1E9wSR6J9nxP-NbYtJdyjs_QjUHRa61ME5rsJ4Ktn2P0/edit?tab=t.z4l3jis729gk#heading=h.8xhgps3swnhi)~~
  - [x] Extract a data class from ResourceDescriptionBuilder
  - [x] Use decorators to define components

## Documentation

# Testing

* Used the metadata wizard to examine normal tables and some weird ones
  * normals: out_eia__monthly_generators, core_epacems__hourly_emissions, out_sec10k__parents_and_subsidiaries
  * weirdos: out_censusdp1tract__states, out_gridpathratoolkit__hourly_available_capacity_factor, core_eia861__yearly_green_pricing

## checking tables with year-based availability

Ran some code to check for gross mismatches. It is pretty quick:

<details><summary>code behind the cut</summary>

```python
import duckdb

from pudl.metadata.classes import PudlResourceDescriptor, Resource
from pudl.metadata.descriptions import ResourceDescriptionBuilder
from pudl.metadata.resources import RESOURCE_METADATA

pp = PudlPaths()

def prefer(candidates, fn):
    preferred_candidates = [c for c in candidates if fn(c)]
    if len(preferred_candidates) > 0:
        return preferred_candidates
    return candidates

def find_date_column(fields):
    candidates = [c for c in fields if (
        c.endswith("date") or c.endswith("year") or c.endswith("datetime_utc")
    )]
    if len(candidates)>1:
        candidates = prefer(candidates, lambda c: c.startswith("report"))
    if len(candidates)>1:
        candidates = prefer(candidates, lambda c: c.endswith("year"))
    if len(candidates)==1:
        return candidates[0]
    return None

for resource_id, resource_dict in sorted(RESOURCE_METADATA.items()):
    desc = ResourceDescriptionBuilder(
        resource_id=resource_id,
        settings=Resource._resolve_references_from_resource_descriptor(
            resource_id, PudlResourceDescriptor.model_validate(resource_dict)
        ),
    ).build()
    date_column = find_date_column(resource_dict["schema"]["fields"])
    if not date_column:
        print(f"no_date {resource_id}")
        continue
    try:
        resource_path = str(pp.parquet_path(resource_id))
    except FileNotFoundError as e:
        print(f"no_file {resource_id}")
        continue
    try:
        query = f"""SELECT MAX({date_column}) as max_date FROM '{resource_path}'"""
        df = duckdb.sql(query).df()
        max_year = df.max_date[0]
        if not date_column.endswith("year"):
            max_year = df.max_date.dt.year[0]
        if desc.availability.type != "True":
            print(f"no_avail {resource_id} {date_column} {max_year}")
            continue
        predicted_max_year = int(desc.availability.description[:4])
        if max_year != predicted_max_year:
            print(f"mismatch {resource_id} {date_column} {max_year} {desc.source.type} {predicted_max_year}")
    except:
        print(f"wtf {resource_id} {date_column}")
        raise
```
</details>

The first pass found the following, which have all been addressed using row counts:

<details><summary>[RESOLVED] 8 lag a year behind;</summary>

* _core_eia923__monthly_cooling_system_information (2024 from report_date vs 2025 from eia923)
* _core_eia923__yearly_byproduct_disposition (2024 from report_year vs 2025 from eia923)
* _core_eia923__yearly_byproduct_expenses_and_revenues (2024 from report_year vs 2025 from eia923)
* _core_eia923__yearly_fgd_operation_maintenance (2024 from report_date vs 2025 from eia923)
* out_eia923__yearly_boiler_fuel (2024 from report_date vs 2025 from eia923)
* out_eia923__yearly_fuel_receipts_costs (2024 from report_date vs 2025 from eia923)
* out_eia923__yearly_generation (2024 from report_date vs 2025 from eia923)
* out_eia923__yearly_generation_fuel_combined (2024 from report_date vs 2025 from eia923)

</details>

<details><summary>[RESOLVED] 15 are a year ahead;</summary>

* core_eia860__scd_boilers (2025 from report_date vs 2024 from eia860)
* core_eia860__scd_generators (2025 from report_date vs 2024 from eia860)
* core_eia860__scd_plants (2025 from report_date vs 2024 from eia860)
* core_eia860__scd_utilities (2025 from report_date vs 2024 from eia860)
* core_eia860m__changelog_generators (2025 from report_date vs 2024 from eia860m)
* out_eia__monthly_generators (2025 from report_date vs 2024 from eia)
* out_eia__yearly_assn_plant_parts_plant_gen (2025 from report_date vs 2024 from eia)
* out_eia__yearly_boilers (2025 from report_date vs 2024 from eia)
* out_eia__yearly_generators (2025 from report_date vs 2024 from eia)
* out_eia__yearly_generators_by_ownership (2025 from report_date vs 2024 from eia)
* out_eia__yearly_plant_parts (2025 from report_year vs 2024 from eia)
* out_eia__yearly_plants (2025 from report_date vs 2024 from eia)
* out_eia__yearly_utilities (2025 from report_date vs 2024 from eia)
* out_ferc714__georeferenced_respondents (2024 from report_date vs 2023 from ferc714)
* out_ferc714__respondents_with_fips (2024 from report_date vs 2023 from ferc714)

</details>

## tables with subannual availability

[RESOLVED] All the tables below have temporal partitions in row counts, and no longer show up with subannual availability.

* CEMS is the only source with quarters. The only CEMS table we publish also draws from eia860, which only tracks annual availability. Because we base availability off of the least recent among the available sources, availability for core_epacems__hourly_emissions will only ever report the annual figure from eia860. This means it was checked adequately in the previous section ✅ .
* ~7 eia930 tables and 1 eia code table use half-years. here are the predicted availability and max date for each table:~
  * ✅ core_eia930__hourly_interchange 2025half2 2025-10-31 07:00:00
  * ✅ core_eia930__hourly_net_generation_by_energy_source 2025half2 2025-11-02 07:00:00
  * ✅ core_eia930__hourly_operations 2025half2 2025-11-02 07:00:00
  * ✅ core_eia930__hourly_subregion_demand 2025half2 2025-11-01 07:00:00
  * ✅ no date columns in core_eia__codes_balancing_authority_subregions (though we still display availability for it)
  * ✅ out_eia930__hourly_aggregated_demand 2025half2 2025-11-02 07:00:00
  * ✅ out_eia930__hourly_operations 2025half2 2025-11-02 07:00:00
  * ✅ out_eia930__hourly_subregion_demand 2025half2 2025-11-01 07:00:00

## To-do list

- [x] Check CEMS and other tables with subannual releases, since the above code probably doesn't catch them


src/pudl/metadata/descriptions.py
```diff
@@ -136,6 +137,66 @@ def first_non_none(*args):
...
+    def jinja_render(self, jinja_environment):
+        """Render all description components into the full static description text block using the resource_description template."""
+        return (
+            jinja_environment.get_template("resource_description.rst.jinja").render(
+                descriptions=self
+            )
+        ).strip()
+
+
+def component(fn):
```
jdangerx: Might be nice to put a type hint on this guy so that we know what a component is supposed to do:

```diff
diff --git a/src/pudl/metadata/descriptions.py b/src/pudl/metadata/descriptions.py
index 93d708b98..ed3368e55 100644
--- a/src/pudl/metadata/descriptions.py
+++ b/src/pudl/metadata/descriptions.py
@@ -2,6 +2,7 @@
 
 import re
 from collections import namedtuple
+from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Self
 
@@ -191,7 +192,9 @@ class ResolvedResourceDescription:
         ).strip()
 
 
-def component(fn):
+def component(
+    fn: Callable[..., ResourceTrait | list[ResourceTrait]],
+) -> Callable[..., ResourceTrait | list[ResourceTrait]]:
     """Decorator for functions which resolve a description component."""
     fn.__is_component = True
     return fn
```

And then your typechecking tooling will also yell at you for putting `@component` on a bad function.

jdangerx (dismissed):
Looks good to me, this feels easier to read and maintain :) Just one nit about type hints.

src/pudl/metadata/classes.py
```diff
@@ -983,6 +983,14 @@ def get_resource_ids(self) -> list[str]:
             if value.get("etl_group") == self.name
         )
 
+    def get_temporal_partitions(self) -> list:
+        """Return a list of temporal partitions encoding the time span covered by the data source."""
+        partitions = self.working_partitions
+        for key in ["years", "half_years", "year_quarters", "year_months"]:
+            if key in partitions:
+                return partitions[key]
```
krivard, line 1024: ...could replace this with 

```suggestion
        for key in partitions:
            if "year" in key:
                return partitions[key]
```

it would be a little more flexible and require a little less maintenance, but would be a little more likely to generate spooky behavior if we were ever to add a particularly weird key

src/pudl/metadata/classes.py
```diff
@@ -1238,6 +1246,22 @@ class PudlDescriptionComponents(PudlMeta):
         If None or otherwise left unset, will be filled in with auto warnings only. If
         no auto warnings apply, hides the Usage Warnings section entirely."""
 
+        availability_offset: int = 0
+        """Partition offset of most recent data available from that claimed by the data
+        source.
+
+        Only consulted when ``availability_text`` is None or otherwise unset.
```
krivard: perhaps

```suggestion
        Only permitted when ``availability_text`` is None or otherwise unset.
```

src/pudl/metadata/classes.py
```diff
@@ -1536,11 +1560,11 @@ def dict_from_id(resource_id: str) -> dict:
         return Resource.dict_from_resource_descriptor(resource_id, descriptor)
 
     @staticmethod
-    def dict_from_resource_descriptor(  # noqa: C901
+    def _resolve_references_from_resource_descriptor(  # noqa: C901
         resource_id: str,
         descriptor: PudlResourceDescriptor,
     ) -> dict:
-        """Get a Resource-shaped dict from a PudlResourceDescriptor.
+        """Partially constructs a Resource-shaped dict from a PudlResourceDescriptor.
```
krivard, line 1610: This refactor is needed so that we can adequately test components of the description that now rely on other parts of the metadata being fully-resolved (to DataSource or Schema or w/e)

src/pudl/metadata/descriptions.py
```diff
@@ -93,6 +94,34 @@
...
+    "half_years": half_year_offset,
+    "year_months": lambda partition, offset: str(
+        pd.Period(pd.Timestamp(partition) + pd.DateOffset(months=offset), freq="M")
+    ).lower(),
+}
+"""Lookup table for computing offsets of temporal partitions.
+
+Maps each partition key to a function that, given a partition (2026, 2026q3,
+2026half1, 2026-07) and an integer offset, returns the partition ``offset``
+partition-units away from the input partition."""
```
krivard: These will need to be kept in sync with the temporal partition keys used by DataSource.

I'm pretty sure the description machinery is the only stuff that needs to do partition arithmetic, but if these would be useful elsewhere in the codebase lmk and I can file a followup issue to pull this into utils instead.

src/pudl/metadata/descriptions.py
```diff
@@ -147,6 +176,7 @@ class ResolvedResourceDescription:
     There are six description components:
 
     * summary
+    * availability [under development]
```
krivard:
```suggestion
    * availability
```

src/pudl/metadata/descriptions.py
```diff
@@ -175,12 +206,13 @@ def summarize(self) -> str:
...
-        PK [{self.primary_key.type}]: {self.primary_key.description}
-  Warnings [{len(self.usage_warnings)}]:{"\n\t".join([""] + [f"{uw.type} - {uw.description}" for uw in self.usage_warnings])}
-   Details [{self.details.type}]:
+     Summary [{self.summary.type}]: {self.summary.description}
+Availability [{self.availability.type}]: {self.availability.description}
+       Layer [{self.layer.type}]: {self.layer.description}
+      Source [{self.source.type}]: {self.source.description}
+          PK [{self.primary_key.type}]: {self.primary_key.description}
+    Warnings [{len(self.usage_warnings)}]:{"\n\t".join([""] + [f"{uw.type} - {uw.description}" for uw in self.usage_warnings])}
+     Details [{self.details.type}]:
```
krivard, line 224: The non-Availability parts of this are just a lateral whitespace change

src/pudl/metadata/descriptions.py
```diff
@@ -251,6 +283,25 @@ def build(self):
...
+    @staticmethod
+    def offset_source_availability(source, offset: int) -> str | None:
+        """Compute an availability date from the most recent available partition for a DataSource, offset by some number of partition-length units."""
+        availability = max(source.get_temporal_partitions() or [None])
+        if availability is None:
+            # source availability isn't temporal so we don't know how to offset it
+            return availability
+        availability = str(availability)
+        if offset == 0:
+            # we're not offsettting the source availability
```
krivard:
```suggestion
            # we're not offsetting the source availability
```

src/pudl/metadata/descriptions.py
```diff
@@ -251,6 +283,25 @@ def build(self):
...
+            return availability
+        for partition_key in source.working_partitions:
+            if partition_key in partition_offsets:
+                return partition_offsets[partition_key](availability, offset)
+        # no partition key had "year" in it
```
krivard:
```suggestion
        # we don't have an offset function configured for this key
```

src/pudl/metadata/descriptions.py
```diff
@@ -251,6 +283,25 @@ def build(self):
             },
         )
 
+    @staticmethod
+    def offset_source_availability(source, offset: int) -> str | None:
+        """Compute an availability date from the most recent available partition for a DataSource, offset by some number of partition-length units."""
+        availability = max(source.get_temporal_partitions() or [None])
+        if availability is None:
+            # source availability isn't temporal so we don't know how to offset it
```
krivard:
```suggestion
            # source availability isn't temporal so we don't care about it for offset purposes
```

src/pudl/metadata/classes.py
```diff
@@ -1238,6 +1246,22 @@ class PudlDescriptionComponents(PudlMeta):
         If None or otherwise left unset, will be filled in with auto warnings only. If
         no auto warnings apply, hides the Usage Warnings section entirely."""
 
+        availability_offset: int = 0
+        """Partition offset of most recent data available from that claimed by the data
+        source.
+
+        Only consulted when ``availability_text`` is None or otherwise unset.
+        Useful in cases where a particular resource relies on a subset of source data that
+        is updated on a slower cadence than most other data in the source."""
```
e-belfer:
```suggestion
        is updated on a slower cadence than most other data in the source (e.g., most EIA 923 tables are published monthly, but some are only published annually)."""
```

src/pudl/metadata/descriptions.py
```diff
@@ -93,6 +94,34 @@
...
+def half_year_offset(partition: str, offset: int) -> str:
+    """Offset a half_year partition by the specified number of half_years."""
+    yr, _, half = partition.partition("half")
+    offset = pd.Timestamp(
+        year=int(yr), month=(int(half) - 1) * 6 + 1, day=1
+    ) + pd.DateOffset(months=offset * 6)
+    return f"{offset.year}half{offset.month / 6 + 1:.0f}"
+
+
+partition_offsets: dict[str, Callable] = {
```
e-belfer: We typically upper case these kinds of dictionaries, e.g., `PARTITION_OFFSETS`.

src/pudl/metadata/descriptions.py
```diff
@@ -147,6 +176,7 @@ class ResolvedResourceDescription:
     There are six description components:
```
e-belfer: seven?

e-belfer (commented):
I'm fully aware that I'm reviewing this fairly late into the process and missed the design conversations, so I'm not going to fully block this PR from going ahead with this version of the implementation. I do want to flag some hesitations about the current setup and one concern that is actually blocking at this point.

I think I'm personally highly averse to showing metadata that's incorrect, even if it's only incorrect for a small percentage of cases. If I saw that a table that was supposed to have 2025 data in it didn't in the description, I might not explore it. If I saw availability said through 2024 but there was 2025 data in the table, I'd be even more confused.

The tables that are currently somewhat wrong are some of our most important and most used tables, which makes me even less inclined to report their availability if it isn't actually correct.

> core_eia860__scd_boilers (2025 from report_date vs 2024 from eia860)
> core_eia860__scd_generators (2025 from report_date vs 2024 from eia860)
> core_eia860__scd_plants (2025 from report_date vs 2024 from eia860)
> core_eia860__scd_utilities (2025 from report_date vs 2024 from eia860)
> core_eia860m__changelog_generators (2025 from report_date vs 2024 from eia860m)
> out_eia__monthly_generators (2025 from report_date vs 2024 from eia)
> out_eia__yearly_assn_plant_parts_plant_gen (2025 from report_date vs 2024 from eia)
> out_eia__yearly_boilers (2025 from report_date vs 2024 from eia)
> out_eia__yearly_generators (2025 from report_date vs 2024 from eia)
> out_eia__yearly_generators_by_ownership (2025 from report_date vs 2024 from eia)
> out_eia__yearly_plant_parts (2025 from report_year vs 2024 from eia)
> out_eia__yearly_plants (2025 from report_date vs 2024 from eia)
> out_eia__yearly_utilities (2025 from report_date vs 2024 from eia)

Why? These are all incorporating 860M data (snuck in through EIA 860 for the eia860 source tables, see #4327, which means that the 860 data availability is an unreliable source of truth) and the eia_yearly tables also have EIA 923 data in them.

I think fixing the dates for our core tables is absolutely a blocker. I understand we could do this by adding more offsets into the individual tables and handling these edge cases explicitly, though I think there could be times of year or other even edgier edge cases where this offsetting might not work as expected. We could also not show availability for these tables until we can fix this issue, but again I think that'd be somewhat confusing for the user if this data appeared in other tables.

From an external contribution perspective, conceptually going forward, I think it's going to be tricky for a user integrating a table to think, "Ah yes, this table is -1 year available from the most recent data in the same data source", though of course this is something we could rectify as internal developers.

Big picture, I have a question about whether there are other ways of making this happen that are less complex and less fragile. Some of these avenues you might've already explored and ruled out, some might be more/less complicated and probably more/less out of scope for this round:

- Least complicated: Just pulling the min and max years from the DBT row counts CSV by table, and using them as the source of truth for the date range seems pretty lightweight and reliable. Are there any concrete implementation barriers to doing so that I'm missing?
- More complex: Using a DBT test to check for observed min/max dates by table and updating a CSV similar to our row counts, which the metadata build then reads in.
- More more complex: Adding more useful metadata into the creation of the datapackage.json at the archiving step, which could me more/less possible by data source.

src/pudl/metadata/descriptions.py
krivard: None of our current tables need this right now. I could rip it out, but it's written and tested. wdyt?

src/pudl/scripts/dbt_helper.py
```diff
@@ -14,10 +14,10 @@
 from deepdiff import DeepDiff
 from pydantic import BaseModel
 
-from pudl.dbt_wrapper import DBT_DIR, build_with_context, dagster_to_dbt_selection
+from pudl.dbt_wrapper import build_with_context, dagster_to_dbt_selection
 from pudl.logging_helpers import configure_root_logger, get_logger
 from pudl.metadata.classes import PUDL_PACKAGE
-from pudl.workspace.setup import PudlPaths
+from pudl.workspace.setup import DBT_DIR, PudlPaths
```
krivard, line 20: This refactor avoids a circular import

src/pudl/workspace/setup.py
```diff
@@ -14,6 +14,7 @@
 PotentialDirectoryPath = DirectoryPath | NewPath
 
 PUDL_ROOT_PATH = Path(__file__).parent.parent.parent.parent
+DBT_DIR: Path = PUDL_ROOT_PATH / "dbt"
```
krivard, line 17: This was moved from dbt_wrapper

src/pudl/dbt_wrapper.py
```diff
@@ -15,12 +15,10 @@
 
 import pudl.etl
 from pudl.logging_helpers import get_logger
-from pudl.workspace.setup import PUDL_ROOT_PATH, PudlPaths
+from pudl.workspace.setup import DBT_DIR, PUDL_ROOT_PATH, PudlPaths
 
 logger = get_logger(__name__)
 
-DBT_DIR: Path = PUDL_ROOT_PATH / "dbt"
-
```
krivard, line 24: This refactor avoids a circular import

test/unit/metadata_test.py
```diff
@@ -249,11 +355,15 @@ def test_frictionless_data_package_resources_populated():
 )
 def test_description_compliance(resource_id):
     resource_dict = RESOURCE_METADATA[resource_id]
-    assert isinstance(resource_dict["description"], dict), (
-        f"""Table {resource_id} is listed as description-compliant in metadata_test.py, but the "description" key is not a dictionary. (In theory pydantic should have screamed about that before you got this far)"""
+    description_dict = resource_dict["description"]
+    assert isinstance(description_dict, dict), (
+        f"""Table {resource_id} must have a dictionary under the "description" key, but instead I found a {type(description_dict)}"""
```
krivard, line 591: Small refactor so we can reuse `description_dict` a bunch of times later

test/unit/metadata_test.py
```diff
@@ -276,6 +386,25 @@ def test_description_compliance(resource_id):
...
+        "core_pudl__assn_utilities_plants",
+        "core_pudl__codes_datasources",
+        "core_pudl__codes_imputation_reasons",
+        "core_pudl__codes_subdivisions",
+        "core_pudl__entity_plants_pudl",
+        "core_pudl__entity_utilities_pudl",
+    }:
+        assert resolved.availability.type == "True", (
+            f"Missing availability for {resource_id}"
+        )
```
krivard, line 638: This will scream if we get a new table that defies our availability heuristic

test/unit/metadata_test.py
```diff
@@ -232,6 +236,108 @@ def test_frictionless_data_package_resources_populated():
...
+):
+    """Verify fallback flow is correct for resolving most recently available data."""
+    mocker.patch(
+        "pudl.metadata.descriptions.ResourceDescriptionBuilder.compute_rowcounts_availability",
+        mocker.Mock(side_effect=given_rowcounts, __is_component=False),
+    )
+    mocker.patch(
+        "pudl.metadata.descriptions.ResourceDescriptionBuilder.offset_source_availability",
+        mocker.Mock(side_effect=given_source, __is_component=False),
+    )
```
krivard, line 531: Don't actually read from row counts file or source metadata -- we're just exercising edge cases here

src/pudl/metadata/classes.py
```diff
@@ -1262,6 +1270,29 @@ class PudlDescriptionComponents(PudlMeta):
...
+        availability_text: str | None = None
+        """Most recent data available. If None or otherwise left unset, will be
+        filled in with the most recent partition listed in the row counts file
+        for this resource. If this is None/unset **and** the row counts
+        partition is null, then will be filled in with the most recent partition
+        listed for the data source, optionally offset by ``availability_offset``
+        partitions.
+
+        Generally only set when one table from a data source has been discontinued, but
+        the remaining tables continue being updated."""
```
krivard:
```suggestion
        Generally only set when a discontinued table does not use temporal partitioning
        in the row counts file, but we still know when its freshest data was from."""
```

src/pudl/metadata/descriptions.py
```diff
@@ -259,6 +292,44 @@ def build(self):
...
+        )
+
+    @staticmethod
+    def compute_rowcounts_availability(resource_id) -> str | None:
+        """Compute an availability date from the most recent available partition for a resource, according to the row counts file."""
+        rowcount_partition_lookups = (
+            pd.read_csv(
+                DBT_DIR / "seeds" / "etl_full_row_counts.csv",
+                # [kmm dec 2025] currently all non-null row count partitions are integer years.
+                # if we add other kinds of partitions to the row counts file, we'll need to revise this code.
```
krivard, line 321: (confirmed this is still true in apr 2026)

src/pudl/metadata/descriptions.py
```diff
@@ -304,6 +375,34 @@ def summary(self, settings, defaults: "ResourceNameComponents") -> ResourceTrait
             description=" ".join(c for c in components if c is not None),
         )
 
+    @component
+    def availability(self, settings, defaults) -> ResourceTrait:
+        """Compute the availability component of the resource description."""
```
e-belfer: Can you add more about the intended behavior of this function? It'd be helpful to spell out the override behavior in the docstring.

src/pudl/metadata/descriptions.py
```diff
@@ -304,6 +375,34 @@ def summary(self, settings, defaults: "ResourceNameComponents") -> ResourceTrait
...
+                    )
+                    # skip sources with weird partitions (inner)
+                    if avail is not None
+                ]
+                or [None]  # skip sources with weird partitions (outer)
+            ),
+            "Unknown",
+        )
+        return ResourceTrait(
+            type=str(most_recent_data != "Unknown"), description=most_recent_data
```
e-belfer, line 419: I find the syntax here somewhat confusing - can you clarify what the behavior around "Unknown" is expected to return in these cases?

e-belfer (commented):
The docs build as expected and 98% of these look correct! Don't forget to add the new override field into the metadata editing developer documentation.

I have one small thought about a user warning change that might clarify what we mean, as we're now using this label more broadly:

> The data is no longer being collected or reported in this way. --> "The **original** data is no longer being collected or reported in this way."

The questions for me come in the 3rd series of overrides originating from the complex infrastructure of inferring data availability by source. These are now predominantly applying to the entity and association tables, and I still find them to produce somewhat confusing results.

For example:

- `core_eia923__entity_coalmine` has most recent availability as 2025, even though we're integrating 2026 EIA 923 data and this table does not have a year column.
- `core_eia__entity_boilers` has the most recent year as 2024 (coming from 860), but also includes later data from 923 and 860M. I remember the intended behavior here is to heuristically take the "oldest new" data source as the referent year, but I'm wondering whether this is actually more confusing for users.
- `core_pudl__assn_eia_pudl_utilities` similarly definitely contains data past 2024, which is the last year of most recent 860M data.

I think there are a few options:

- Prioritize completion over accuracy - Change the heuristic from "least recent" to "most recent" data source, but keep the infrastructure as is - this will probably produce some different inaccuracies.
- Prioritize accuracy over completion - Drop the third backstop and let association and entity tables have no dates for now, and handle these edge cases in a follow-up issue.

krivard:
Decision from breakout:

- Publish availability only where unambiguous: where a data partition is available from row counts or where a table has only one source
- Suppress availability where ambiguous: where we can't use row counts and the table has multiple sources
- Write up multi-source ideas/desiderata for an icebox issue

docs/data_dictionaries/usage_warnings.rst
```diff
@@ -14,15 +14,15 @@ Derived values:
...
+  The original data is no longer being collected or reported in this way.
 
 Discontinued by us:
   PUDL does not currently update its copy of this data.
 
   If you would be interested in funding additional updates to this data, get in touch!
 
 Early release:
-  May contain early release data. EIA releases some of it's annual data early for
+  May contain early release data. EIA releases some of its annual data early for
```
krivard, line 25: (unrelated opportunistic copy edit)

docs/dev/metadata.rst
```diff
@@ -254,6 +277,7 @@ Metadata for each field is primarily stored in
 the same name share the same general definition, no matter where you find them.
 
 Field names should:
+
```
krivard, line 280: (unrelated opportunistic RST fix)

src/pudl/metadata/descriptions.py
krivard: Added plain-language for full logic cascade below

src/pudl/metadata/descriptions.py
krivard: It's the same logic we use for additional-details -- added clarifying comment for both locations

src/pudl/metadata/warnings.py
```diff
@@ -30,7 +30,7 @@
         "`Notable Irregularities <https://catalystcoop-pudl.readthedocs.io/en/latest/data_sources/ferc1.html#notable-irregularities>`_ "
         "for details."
     ),
-    "discontinued_data": "The data is no longer being collected or reported in this way.",
+    "discontinued_data": "The original data is no longer being collected or reported in this way.",
```
krivard, line 33: :increments counter for number of edits we've had to copy-paste between this file and usage_warnings.rst:

some day we'll add helper code but that day is not today

test/unit/metadata_test.py
krivard: hmm, git rebased this comment badly. it belongs with L589:

>    description_dict = resource_dict["description"]

test/unit/metadata_test.py
```diff
@@ -425,12 +429,150 @@ def test_frictionless_data_package_filter_resources():
...
+    "core_pudl__assn_utilities_plants",
+    "core_pudl__codes_data_maturities",
+    "core_pudl__codes_datasources",
+    "core_pudl__codes_imputation_reasons",
+    "core_pudl__codes_subdivisions",
+    "core_pudl__entity_plants_pudl",
+    "core_pudl__entity_utilities_pudl",
+    "core_rus__codes_fuel_types",
+    "core_rus__codes_investment_types",
+}
```
krivard, line 574: it turns out there are more like ~2 dozen of these but they're still all codes, entity, and association tables so it feels okay

docs/dev/metadata.rst
```diff
@@ -183,8 +183,15 @@ clarification:
...
-
-For a full reference on available description fields, including fields that
+* ``availability_text``: Only set for tables that we no longer update **and**
+  which do not have a temporal partition in the row counts file. In such cases,
+  this field is optional for tables discontinued at the same time as their
+  source, or which come from a monolithic source (having no source partitions).
+  This field **must** be set for tables discontinued before their source, since
+  otherwise they will be listed with the most recent source partition, and that
+  will not match the availability of the table.
+
```
e-belfer:
```suggestion
* ``availability_text``: An optional field to override the default availability
   of a source table. Only set for tables which do not have a temporal partition
   in the row counts file **and** which lag the availability of their corresponding
   source data (e.g., discontinued tables). This field is optional for tables discontinued at the 
   same time as their source, or which come from a monolithic source (having no source 
   partitions, e.g., the EIA bulk electricity dataset). This field **must** be set for tables 
   discontinued before their source, since otherwise they will be listed with the most recent 
   source partition, and that will not match the availability of the table.
```

Tried to clarify wording a bit - we are technically still updating these tables, there's just no new data at the source.

test/unit/metadata_test.py
e-belfer: Great, these all seem reasonable. And testing this list here is verify useful!

e-belfer (approved):
Excited to build on this to include date ranges and eventually months of availability.

Built the docs, and spot checked that dates were as expected for tables from 860, 923, 860M, FERC Form 1, RUS, entity, code and association tables.

docs/dev/metadata.rst
krivard: hmm I'm going to massage this a bit further to discourage uses of this field that require ongoing updates

PR #4636, opened 2025-09-30
feat/clean_eia923_schedule8b

alexclippinger:
# Overview
Closes https://github.com/catalyst-cooperative/pudl/issues/4099

Transforms `raw_eia923__byproduct_expenses_and_revenues` into `_core_eia923__byproduct_expenses_and_revenues`, a new clean table ready to be combined with other EIA 923 and 860 data in PUDL.

# To-do
- [x] Ran unit-tests
- [x] Materialized the new asset in Dagster (with `@asset` decorator)
- [ ] Write the table to the PUDL database and [update alembic schema](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_the_etl.html#database-schema-migration). Waiting to do this after initial review (if this is within scope? Not sure if it would be done at the same time as adding dbt tests, which I think may be out of scope?)

I think this is ready for an initial review. Let me know what the next steps are!

### DIFF
--- a/src/pudl/metadata/fields.py
+++ b/src/pudl/metadata/fields.py
@@ -592,6 +592,14 @@ FIELD_METADATA: dict[str, dict[str, Any]] = {
         "description": "Total installed (nameplate) capacity, in megawatts.",
         "unit": "MW",
     },
+    "capex_air_abatement": {
+        "type": "number",
+        "description": (
+            "Cost of new structures and/or equipment purchased to reduce, monitor, or "
+            "eliminate airborne pollutants"
+        ),
+        "unit": "USD",
+    },
     "capex_annual_addition": {
         "type": "number",
         "description": "Annual capital addition into `capex_total`.",
@@ -652,6 +660,14 @@ FIELD_METADATA: dict[str, dict[str, Any]] = {
         "description": "Cost of plant: land and land rights (USD).",
         "unit": "USD",
     },
+    "capex_other_abatement": {
+        "type": "number",
+        "description": (
+            "Other amortizable expenses and purchases of new structures and or equipment "
+            "when such purchases are not allocated to a particular unit or item."
+        ),
+        "unit": "USD",
+    },
     "capex_other": {
         "type": "number",
         "description": "Other costs associated with the plant (USD).",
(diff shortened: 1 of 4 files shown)


src/pudl/transform/eia923.py
```diff
@@ -1587,3 +1587,39 @@ def disposition_continuity_check(bpd):
...
+    df = df.drop(["early_release"], axis=1)
+
+    df = pudl.helpers.fix_eia_na(df)
+
+    # One dupe for plant_id_eia=6504 and report_year=2010 with no differences in byproduct dollars reported
+    df = pudl.helpers.dedupe_and_drop_nas(
+        df, primary_key_cols=["plant_id_eia", "report_year"]
+    )
+
+    # Convert thousands of dollars to dollars and remove suffix from column name
```
e-belfer, line 1625: I'm seeing that the columns here end with `_1000_dollars` rather than `_thousand_dollars`, so this isn't working as expected.

src/pudl/metadata/fields.py
```diff
@@ -592,6 +592,14 @@
         "description": "Total installed (nameplate) capacity, in megawatts.",
         "unit": "MW",
     },
+    "capex_air_abatement": {
+        "type": "number",
+        "description": (
+            "Cost of new structures and/or equipment purchased to reduce, monitor, or "
+            "eliminate airborne pollutants"
```
```suggestion
            "eliminate airborne pollutants."
```

src/pudl/transform/eia923.py
alexclippinger: oops fixed! Had this in my inspect notebook then copied from another eia923 transform without seeing the difference

alexclippinger:
> This is looking really good! It seems like the *1000 isn't working as intended currently, but should be a quick fix. The column descriptions all look good, with a tiny nit re: a missing period.
> 
> One additional request:
> 
> * Since `revenues_total_byproduct_1000_dollars`, `opex_total_collection_abatement_1000_dollars`, `opex_total_disposal_abatement_1000_dollars`, and `opex_total_other_abatement_1000_dollars` are all totals of other columns, it would be great to characterize how accurately they represent totals of their components by using the dbt test `expect_sum_close_to_total_column`. See `dbt/models/eia923/_core_eia923__yearly_fgd_operation_maintenance/schema.yml` for an example of how to implement this, and let me know if you have any questions here.

Addressed all of your feedback, thanks for reviewing! Just FYI I did try to run the dbt tests locally (with `dbt build --select "source:pudl._core_eia923__byproduct_expenses_and_revenues"`) but got a runtime error: "No files found that match the pattern". I'll get my local environment set up correctly on Monday to confirm again that these work (guessing I need to materialize/re-materialize something, or update a config?). I did test via pandas that the sum columns do add up to the total column as expected.

alexclippinger:
Ok, got the parquet file downloaded and dbt tests pass.

PR #4655, opened 2025-10-14
Memory profiling tools

### LINKED ISSUE #4619: Easy tools to get detailed CPU and memory usage for an asset on-demand
We're collecting a list of slow or big assets in #4565 - but we need tools to help us make them faster!

## Success criteria

* [ ] documented process for identifying what parts of an asset are allocating lots of memory or using lots of CPU

## Next steps
* [ ] figure out how to hook up memray to a single asset materialization
* [ ] figure out how to hook up py-spy to an asset materialization
* [ ] timebox: 1h - figure out how to instrument these in a way that lets you trigger things *from dagster dev*

jdangerx:
# Overview

Closes #4619.

It's hard to see the memory usage of asset materializations!

## What did you change?

* Added a new devtools script - `memory_profile.py` - that engages a memory profiler while materializing assets.

It took a long time because I went down the rabbit hole of "wouldn't it be nice if we could just run things from the Dagster UI and get memory profiles?" which almost worked except for it relies on the `forkserver` multiprocessing start method which seems to cause intermittent failures on MacOS.

## Documentation

# Testing

Tried running this against a couple different assets & asset selections.

## To-do list

### DIFF
--- a/devtools/memory_profile.py
+++ b/devtools/memory_profile.py
@@ -46,6 +46,7 @@ def materialize_assets(asset_selection: str) -> None:
     full_etl_job = defs.get_job_def("etl_full")
     dg.materialize(
         assets=defs.assets,
+        instance=dg.DagsterInstance.get(),
         resources=full_etl_job.resource_defs,
         selection=asset_selection_with_multi_asset_siblings,
     )


src/pudl/profiling.py
jdangerx, line 1: Should probably rename this to "_forkserver_setup" or something more explicitly user-hostile.

devtools/materialize_to_feather.py
jdangerx, line 1: Should remove this file from the PR completely.

.gitignore
```diff
@@ -49,3 +49,7 @@ dbt/dbt_packages/
 dbt/target/
 dbt/logs/
 dbt/.user.yml
+
+# memray profiles
```
jdangerx, line 53: The `-d` parameter lets you put the profiles anywhere, but I assume people will end up with a few lying around in the repo that don't need to get tracked.

devtools/memory_profile.py
```diff
@@ -0,0 +1,74 @@
...
+        asset_selection: a string that we can turn into an asset key.
+    """
+    # NOTE (2025-10-21): Putting the PUDL import here means we don't have to wait for the lengthy import.
+    from pudl.etl import defs
+
+    asset_selection_with_multi_asset_siblings = dg.AssetSelection.from_string(
+        asset_selection
+    ).required_multi_asset_neighbors()
+
+    full_etl_job = defs.get_job_def("etl_full")
```
zaneselvans, line 46: I wonder if there are cases where it would be faster and also still informative to be able to run the fast ETL version of an asset?

More generally, this makes me wonder about the context in which the asset is being materialized -- do all of its upstream dependencies need to already be materialized and up to date? If they aren't will the script fail (my guess) or attempt to materialize them?

If the script depends on the asset being materialized having all of its upstream dependencies available and ready to go (e.g. you need to have run the full ETL locally) then we should probably document that in the module level doc. And if it's going to materialize everything upstream that's required to profile your fresh new asset, that should probably be documented too.  Maybe choosing between these behaviors should be an option, actually?

zaneselvans (changes requested):
Request for documentation about what Dagster state the script requires / expects to be able to do its work.

devtools/memory_profile.py
jdangerx: I would expect running `asset_name` to only attempt to run the asset, and if the upstream dependencies are not available it will fail.

I would expect running `+asset_name` to run the upstream dependencies in addition to `asset_name`, just like in the Dagster UI.

Basically, I would expect asset selection to work the same way here as it does elsewhere. But I should probably check and document :)

zaneselvans: Okay that makes sense. I think a reminder about needing to have all the dependencies ready to go or materializing upstream stuff with the right asset selection would be helpful in the docs.

jdangerx: Tested, confirmed, and updated the docs!

zaneselvans (approved):
Great, thanks for the docs!

zaneselvans:
Kicked off RTD again b/c sad Zenodo timeout.

zaneselvans:
Re-added to the queue because internet made failures.

PR #4659, opened 2025-10-16
Data bug template

zschira:
# Overview

This PR adds a new dedicated "data bug" issue template, and updates docs so their should be a link for each table in the data dictionary to open a data bug issue. This link should also auto-fill the `table` section in the issue template.

This gives users a clear/straightforward path to report any issues they've encountered with our data.

## What did you change?
- Added a new issue template
  - This template uses github's new [yaml templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) because we can auto-fill specific sections (I think, this will need testing once the template is merged in) based on the url, and it creates a nice form that should be very easy to fill out
- Update the data dictionary jinja template to add links to open new data bugs


docs/templates/resource.rst.jinja
```diff
@@ -1,6 +1,7 @@
 -------------------------------------------------------------------------------
 {{ resource.name }}
 -------------------------------------------------------------------------------
+Noticed any irregularities with this table? `File a report. <https://github.com/catalyst-cooperative/pudl/issues/new?template=data_bug_report.yml&table={{ resource.name }}>`__
```
zschira: I put this at the top because it felt like it would be awkward if it broke up any of the other sections, and would be hard to find at the bottom. Open to suggestions though.

.github/ISSUE_TEMPLATE/data_bug_report.yml
```diff
@@ -0,0 +1,44 @@
+name: Data bug report.
```
jdangerx:
```suggestion
name: Data bug report
```

jdangerx (changes requested):
Yay, this is something we should have :) Tiny copy suggestion.

Is there a good way to test that the URL works? The easiest thing I could come up with was to split this into two PRs and get the template in `main` first but open to other ideas :)

.github/ISSUE_TEMPLATE/data_bug_report.yml
```diff
@@ -0,0 +1,44 @@
+name: Data bug report
+description: Report irregularities with PUDL data.
+title: ""
+labels: ["bug", "data-validation"]
```
e-belfer, line 4: Non-blocking drive-by comment: I'm not sure we need the `data-validation` tag here?

jdangerx (approved):
Let's merge this so we can test the fancy URL thing!

PR #4661, opened 2025-10-17
Grab new row counts from failed nightly build

zaneselvans:
# Overview

Last night's build failed, but the only issues seem to have been row counts. I grabbed the new ones from the build outputs and added them here.


zaneselvans:
🤷🏼‍♂️

Issue #4081, opened 2025-02-24
Clean EIA-923 Schedule 8C Air Emissions Control Info


cmgosnell:
## **Overview**

The EIA 923 Schedule 8C Air Emissions Control Information table is the first part of the Schedule 8 Annual Environmental Information reported by power plants with a capacity over 100 megawatts. The table contains important information about how plants are storing and disposing of byproducts from power generation.

Schedule 8C is one of the last 923 tables that has not yet been integrated into PUDL, and we need your help to get it in! These tables have already been extracted into PUDL via #3372, but have not yet been transformed.

## **Goal**
Transform `raw_eia923__emissions_control` into `_core_eia923__emissions_control`, a new clean table ready to be combined with other EIA 923 and 860 data in PUDL.

There is no one size fits all in table transforms, though the transforms of other EIA 923 tables will be a helpful starting point. We document our approach to cleaning data [here](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_guidelines.html#examples-of-acceptable-changes).

The new table should meet the following criteria:
* a well defined and unique primary key: one or more columns that naturally explain what a row in the table corresponds to. For this table, this _should_ be one row per year, plant ID and control ID (report_date, plant_id_eia and some field that combines the various equipment control IDs (e.g., `pollution_control_id_eia`)). You'll need to check that the combination of these fields is unique, and in the case of duplications, propose a strategy for creating a unique primary key.
* all categorical columns should be constrained (e.g., a Y/N column should only contain Y/N).
* all columns containing codes (in this case `operating_status`) should be encoded.
* nulls should be standardized
* units should be standardized
* data types should be defined and enforced
* metadata should be added to the database schema for each new field and for the table itself

*Out of scope*: 
* Combining the data with other EIA datasets - we'll handle this in #3365.

## **Tasks**

**Get set up**
- [ ] Fork the PUDL repository and follow the steps to set up the [PUDL development environment](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/dev_setup.html)
- [ ] Activate the pudl-dev environment: `mamba activate pudl-dev`
- [ ] Grab the latest EIA 923 raw data: `pudl_datastore --dataset eia923`
[...]

PR #4426, opened 2025-07-15
Standardize 1000 tons to tons in fgd_sorbent_consumption column

msa856:
# Overview

[Partially] Closes #4301 for 1 column: fgd_sorbent_consumption_1000_tons

## What problem does this address?

Updates units from "thousands of tons" to "tons" for eia923 `_core_eia923__yearly_fgd_operation_maintenance.fgd_sorbent_consumption_1000_tons` to match consistency across PUDL datasets.

## What did you change?
I changed the column name `fgd_sorbent_consumption_1000_tons` to `fgd_sorbent_consumption_tons` at the following locations:

- dbt/models/eia923/_core_eia923__yearly_fgd_operation_maintenance/schema.yml
- src/pudl/metadata/fields.py (units and description were updated)
- src/pudl/metadata/resources/eia923.py
- src/pudl/transform/eia923.py (`fgd_sorbent_consumption_1000_tons` is multiplied by 1000 and rounded the to nearest 100)

I updated the database to reflect the new column name: migrations/versions/a9b37124d6c8_standardize_fgd_sorbent_1000_tons_column.py

## Documentation

Make sure to update relevant aspects of the documentation:

- [ ] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html): reference the PR and related issues.
- [x] Update relevant Data Source jinja templates (see `docs/data_sources/templates`).
- [x] Update relevant table or source description metadata (see `src/metadata`).
- [ ] Review and update any other aspects of the documentation that might be affected by this PR.

# Testing

How did you make sure this worked? How can a reviewer verify this?

## To-do list

- [x] If updating analyses or data processing functions: make sure to update row count expectations in `dbt` tests.
- [ ] Run `make pytest-coverage` locally to ensure that the merge queue will accept your PR.
- [x] Review the PR yourself and call out any questions or issues you have.
- [ ] For minor ETL changes or data additions, once `make pytest-coverage` passes, make sure you have a fresh full PUDL DB downloaded locally, materialize new/changed assets and all their downstream assets and [run relevant data validation tests](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `pytest` and `--live-dbs`.
- [ ] For bigger ETL or data changes run the full ETL locally and then [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/testing.html#data-validation) using `make pytest-validate`.
[...]

Issue #4301, opened 2025-05-28
Standardize remaining handful of "thousands of" units in PUDL


e-belfer:
### Overview

What is the problem we're solving? For very simple items, this can be encapsulated in the success criteria.

Elsewhere we convert our units from thousand of dollars to dollars, and thousands of lbs to lbs. We haven't done consistently with all fields however. Notably, these fields are still in "thousands of" units:

- `fgd_sorbent_consumption_1000_tons` (to be transformed in the `_core_eia923__fgd_operation_maintenance` function in `pudl.transform.eia923.py`)
- `max_steam_flow_1000_lbs_per_hour` (to be transformed in the `_core_eia860__boilers` function in `pudl.transform.eia860.py`)
- `steam_load_1000_lbs` (should be converted in the `pudl.extract.epacems.py` module)

### Success Criteria

How will we know that we're done?

* [x] `fgd_sorbent_consumption_1000_tons` -> `fgd_sorbent_consumption_tons`
* [ ] `max_steam_flow_1000_lbs_per_hour` -> `max_steam_flow_lbs_per_hour`
* [ ] `steam_load_1000_lbs` -> `steam_load_lbs`

### Get set up
- [ ] Fork the PUDL repository and follow the steps to set up the [PUDL development environment](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/dev_setup.html)
- [ ] Activate the pudl-dev environment: mamba activate pudl-dev
- [ ] Grab the latest raw data for any relevant tables - in this case, EIA 860, EIA 923 and EPACEMS. To do so, run `pudl_datastore --dataset eia923`, ` pudl_datastore --dataset eia860`, and `pudl_datastore --dataset epacems` (the last one pulls a large chunk of data, and might take a while).
- [ ] For the first two variables: First, generate the raw assets upstream of the transform: follow the instructions to [open Dagster](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_the_etl.html#running-the-etl-via-the-dagster-ui), and in the left-hand menu find the "raw_eia923" asset group, e.g.. Click "Materialize all".

### Next steps
* [ ] For each field, identify the first `_core` or `core` table in which it is transformed (see pudl.metadata.resources`).
* [ ] In the corresponding transform function, divide this field by one thousand.
* [ ] Update the field description, units and name
[...]

PR #4668, opened 2025-10-21
Initial integration of EIA-923 Schedule 8C

alexclippinger:
# Overview
Closes https://github.com/catalyst-cooperative/pudl/issues/4081

Transforms `raw_eia923__emissions_control` into `_core_eia923__emissions_control`, a new clean table ready to be combined with other EIA 923 and 860 data in PUDL.

# To-do
- [X] Encode `operational_status`
- [x] Ran unit-tests
- [x] Materialized the new asset in Dagster (with `@asset` decorator)
- [ ] Figure out primary keys
- [ ] Ran dbt tests
- [ ] Write the table to the PUDL database and [update alembic schema](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_the_etl.html#database-schema-migration). 

# Questions / Comments
1. There is already a `so2_removal_efficiency_tested` with the same definition as `so2_removal_efficiency_100pct_load` (which is in `raw_eia923__emissions_control`), so I re-named this column and `particulate_removal_efficiency_tested`.
2. The two test date columns are mostly NULL so I'm not sure how useful they are, but I figured they needed to be cleaned. I didn't see a cleaning function that could be used for this use case? Should I move that `clean_date` function somewhere?
3. Not sure what the primary keys are for this table yet. `report_year` and `plant_id_eia` alone do not represent a unique row. It seems it could be some combo of the other ID columns, but there are some rows where all of these are NULL, leaving only report_year, plant_id_eia, and numeric columns.
    - Just copy/pasting this help from the original issue description:
    > For this table, this should be one row per year, plant ID and control ID (report_date, plant_id_eia and some field that combines the various equipment control IDs (e.g., pollution_control_id_eia)). You'll need to check that the combination of these fields is unique, and in the case of duplications, propose a strategy for creating a unique primary key.
5. Left `fgd_sorbent_consumption_1000_tons` in 1000 tons because this column exists already.

### DIFF
--- a/src/pudl/transform/eia923.py
+++ b/src/pudl/transform/eia923.py
@@ -1385,6 +1385,170 @@ def cooling_system_information_continuity(csi):
     )
 
 
+def _normalize_emissions_control_year(
+    raw_year: pd.Series, current_year: int
+) -> pd.Series:
+    """Normalize 2- and 4-digit year strings to integer years."""
+    year = pd.to_numeric(raw_year, errors="raise")
+    two_digit = raw_year.str.len().eq(2)
+    year = year.where(~two_digit, 2000 + year)
+    year = year.where(~(two_digit & year.gt(current_year)), year - 100)
+    return year
+
+
+def _build_emissions_control_dates(
+    raw_value: pd.Series,
+    year: pd.Series,
+    month: pd.Series,
+    day: pd.Series,
+    min_valid_year: int,
+    current_year: int,
+) -> pd.Series:
+    """Validate date parts and convert into pandas timestamps."""
+    year = year.astype("Int64")
+    month = month.astype("Int64")
+    day = day.astype("Int64")
+
+    month = month.mask(month.eq(0), 1)
+
+    invalid_year = year.ne(0) & (year.lt(min_valid_year) | year.gt(current_year))
+    if invalid_year.any():
+        bad = raw_value[invalid_year].dropna().drop_duplicates().tolist()
+        raise ValueError(
+            f"Year out of expected range ({min_valid_year}-{current_year}) in values: {bad}"
+        )
+
+    invalid_month = ~month.between(1, 12, inclusive="both")
+    if invalid_month.any():
+        bad = raw_value[invalid_month].dropna().drop_duplicates().tolist()
+        raise ValueError(f"Month out of range in values: {bad}")
+
+    rendered = (
+        year.astype(str).str.zfill(4)
+        + "-"
+        + month.astype(str).str.zfill(2)
+        + "-"
+        + day.astype(str).str.zfill(2)
+    )
+    parsed = pd.to_datetime(rendered, errors="coerce")
+    parsed = parsed.where(~year.eq(0))
+
+    invalid_day = year.ne(0) & parsed.isna()
+    if invalid_day.any():
+        bad = raw_value[invalid_day].dropna().drop_duplicates().tolist()
+        raise ValueError(f"Invalid date value in values: {bad}")
+
+    return parsed
+
+
+def _clean_emissions_control_dates(
+    col: pd.Series, spot_fixes: dict[str, pd.Timestamp] | None = None
+) -> pd.Series:
+    """Parse raw EIA-923 emissions-control date strings into datetimes.
+
+    Supported formats include compact month-year (``012024``), delimited month-year
+    (``01-2024``, ``1/20``), full date (``1/11/07``), and year-only values
+    (``1990`` -> January 1st). Known non-date tokens (e.g. ``na``, ``CEM``) are
+    converted to ``NaT``.
+
+    Parsing is strict: unknown structures or impossible date parts raise
+    ``ValueError`` so new upstream format changes are caught quickly.
+
+    Args:
+        col: Raw date-like values to clean.
+        spot_fixes: Optional table-specific mapping of raw string values to corrected
+            timestamps. If provided, these mappings are applied before pattern parsing.
+    """
+    current_year = pd.Timestamp.now().year
+    min_valid_year = 1950
+    null_tokens = {"", "na", "n.a.", "n/a", "cem", "en"}
+    spot_fixes = spot_fixes or {}
+
+    raw = col.astype("string").str.strip()
+    out = pd.Series(pd.NaT, index=raw.index, dtype="datetime64[ns]")
+
+    lowered = raw.str.lower()
+    is_nullish = raw.isna() | lowered.isin(null_tokens)
+
+    is_spot_fix = raw.isin(spot_fixes)
+    out.loc[is_spot_fix] = raw.loc[is_spot_fix].map(spot_fixes)
+
+    to_parse = ~(is_nullish | is_spot_fix)
+
+    bad_chars = to_parse & ~raw.str.fullmatch(r"[0-9\-/]+")
+    if bad_chars.any():
+        bad = raw[bad_chars].drop_duplicates().tolist()
+        raise ValueError(f"Column contains non-digit/non-dash characters: {bad}")
+
+    digits_only = raw.str.replace(r"[-/]", "", regex=True)
+    all_zero = to_parse & digits_only.str.fullmatch(r"0+")
+    to_parse = to_parse & ~all_zero
+
+    full_date = to_parse & raw.str.fullmatch(r"\d{1,2}[/-]\d{1,2}[/-]\d{2,4}")
+    month_year_delimited = to_parse & raw.str.fullmatch(r"\d{1,2}[/-]\d{2,4}")
+    month_year_compact = to_parse & raw.str.fullmatch(r"\d{1,2}\d{4}")
+    year_only = to_parse & raw.str.fullmatch(r"\d{4}")
+
+    unrecognized = to_parse & ~(
+        full_date | month_year_delimited | month_year_compact | year_only
+    )
+    if unrecognized.any():
+        bad = raw[unrecognized].drop_duplicates().tolist()
+        raise ValueError(f"Unexpected emissions-control date format: {bad}")
+
+    if full_date.any():
+        parts = raw[full_date].str.extract(
+            r"(?P<month>\d{1,2})[/-](?P<day>\d{1,2})[/-](?P<year>\d{2,4})"
+        )
+        out.loc[full_date] = _build_emissions_control_dates(
+            raw_value=raw[full_date],
+            year=_normalize_emissions_control_year(parts["year"], current_year),
+            month=pd.to_numeric(parts["month"], errors="raise"),
+            day=pd.to_numeric(parts["day"], errors="raise"),
+            min_valid_year=min_valid_year,
+            current_year=current_year,
+        )
+
+    if month_year_delimited.any():
+        parts = raw[month_year_delimited].str.extract(
+            r"(?P<month>\d{1,2})[/-](?P<year>\d{2,4})"
+        )
+        out.loc[month_year_delimited] = _build_emissions_control_dates(
+            raw_value=raw[month_year_delimited],
+            year=_normalize_emissions_control_year(parts["year"], current_year),
+            month=pd.to_numeric(parts["month"], errors="raise"),
+            day=pd.Series(1, index=parts.index),
+            min_valid_year=min_valid_year,
+            current_year=current_year,
+        )
+
+    if month_year_compact.any():
+        parts = raw[month_year_compact].str.extract(
+            r"(?P<month>\d{1,2})(?P<year>\d{4})"
+        )
+        out.loc[month_year_compact] = _build_emissions_control_dates(
+            raw_value=raw[month_year_compact],
+            year=_normalize_emissions_control_year(parts["year"], current_year),
+            month=pd.to_numeric(parts["month"], errors="raise"),
+            day=pd.Series(1, index=parts.index),
+            min_valid_year=min_valid_year,
+            current_year=current_year,
+        )
+
+    if year_only.any():
+        year = pd.to_numeric(raw[year_only], errors="raise")
+        out.loc[year_only] = _build_emissions_control_dates(
+            raw_value=raw[year_only],
+            year=year,
+            month=pd.Series(1, index=year.index),
+            day=pd.Series(1, index=year.index),
+            min_valid_year=min_valid_year,
+            current_year=current_year,
+        )
+
+    return pd.to_datetime(out)
+
+
 @asset(io_manager_key="pudl_io_manager")
 def _core_eia923__yearly_fgd_operation_maintenance(
     raw_eia923__fgd_operation_maintenance: pd.DataFrame,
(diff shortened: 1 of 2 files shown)


alexclippinger:
Sorry for the delay getting a PR open for this. Left some initial comments / questions on the data!

src/pudl/transform/eia923.py
```diff
@@ -1640,3 +1640,65 @@ def _core_eia923__yearly_byproduct_expenses_and_revenues(
...
+        df[col] = df[col].apply(clean_date)
+
+    # Rename to be consistent with existing column names
+    df = df.rename(
+        {
+            "so2_removal_efficiency_100pct_load": "so2_removal_efficiency_tested",
+            "particulate_removal_efficiency_100pct_load": "particulate_removal_efficiency_tested",
+        },
+        axis=1,
+    )
```
zaneselvans: We should definitely re-use the existing columns. However, this rename should happen in the CSV that maps the spreadsheet columns to their PUDL database names so they have the preferred name from the beginning. I'll update these two as an example -- there are some other names I think we should update that I've highlighted in `fields.py`.

We recently transposed these mapping CSVs so that the annual updates only affect a single row and give easier to read diffs. I merged `main` into this branch so we don't have to worry about big conflicts when updating the mapping.

src/pudl/metadata/fields.py
```diff
@@ -61,6 +61,10 @@
         "type": "boolean",
         "description": "Indicates whether the emissions control equipment controls acid (HCl) gas.",
     },
+    "acid_gas_removal_efficiency": {
+        "type": "number",
+        "description": "Percent removal efficiency for acid gas emissions",
+    },
```
zaneselvans, line 240: For columns that are actually percentages (numbers between 0 and 100, rather than 0.0 and 1.0) we generally end them with `_pct`. When we aren't changing the meaning of the column through transforms, we also like to do the renames in the extract step, by changing the header in the column mapping CSV, in this case:

```
src/pudl/package_data/eia923/column_maps/emissions_control.csv
```

This naming change should also probably apply to the other "efficiency" columns, if they're stated as percentages.

src/pudl/metadata/fields.py
```diff
@@ -4789,6 +4827,26 @@
...
         "description": "If boiler is not in compliance with particulate matter regulations, strategy for compliance.",
     },
+    "particulate_emission_rate_lb_per_mmbtu": {
+        "type": "number",
+        "description": "Average annual emission removal rate for particulate matter.",
+    },
+    "particulate_removal_efficiency_tested": {
+        "type": "number",
+        "description": "The tested efficiency for the removal of particulate matter of an equipment or strategy.",
+    },
```
zaneselvans, line 5792: Is this one also a percentage? If so see comment above about naming columns that report percentages.

src/pudl/metadata/fields.py
```diff
@@ -3766,6 +3789,17 @@
...
+        "type": "number",
+        "description": (
+            "Actual controlled (or uncontrolled) mercury emission rate, based on "
+            "data from CEMS, where possible."
+        ),
+    },
+    "mercury_removal_efficiency": {
+        "type": "number",
+        "description": "Percent removal efficiency for mercury emissions rounded to nearest 0.1 percent of weight",
+    },
```
zaneselvans, line 4546: See comment above about naming columns that report percentages.

src/pudl/transform/eia923.py
```diff
@@ -1640,3 +1640,65 @@ def _core_eia923__yearly_byproduct_expenses_and_revenues(
...
+    """
+    df = raw_eia923__emissions_control.copy()
+
+    # This column is dropped from all EIA 923 tables
+    df = df.drop(["early_release"], axis=1)
+
+    df = pudl.helpers.standardize_na_values(df)
+
+    # Fix malformed report-year string columns (i.e., should be MM-YYYY)
+    def clean_date(val):
```
zaneselvans: * We try and avoid using `df.apply()` whenever possible and use vectorized operations instead because they're much faster and use less memory, so this function should probably take a `pd.Series` and be used to set the value of a whole column in the dataframe being cleaned.
* We follow ISO-8601 for date formats, so the output should probably be `YYYY-MM` which will allow them to sort chronologically and be converted into dates unambiguously.
* A brief docstring that describes what kinds of malformed dates the function is intended to deal with would be helpful. It should also throw an error if it encounters anything outside of those expectations (so if a new kind of badness shows up in future years, we don't silently mangle the dates in unexpected ways)
* We should check that all of the post-conversion values are conceivably valid dates, and decide what to do with invalid ones.
* I see that there are a number of 0000 years and 00 months, which cause `pd.to_datetime()` to fail.
* For cases where both month and year are nothing but 0 values, we should set it to NA.
* For cases where we have a year but no month, I think filling in Jan 1st makes sense so we at least preserve coarse grained temporal information.
* If we have a month but zero year, it should be set to NA.
* We should also check that the dates are within the expected range. Looking at the data, reasonable bounds look like a year between 1950 and the current year. This could either be a mid-transform check here, or a data validation in dbt.
* For development / testing at least, it's probably useful to try and convert these date columns in the asset definition using `pd.to_datetime()` to make sure they're date compatible.  With `YYYY-MM` strings, that will mean we get a date for the first of the month, which is our standard for "monthly" dates.
* I think we have a number of bespoke table or module specific date cleanup functions scattered around PUDL because different data comes in different states of badness. I think pulling this one out into a local `_clean_emissions_control_dates()` function that we can make a little unit test for and give a real docstring makes sense.

src/pudl/transform/eia923.py
```diff
@@ -1640,3 +1640,65 @@ def _core_eia923__yearly_byproduct_expenses_and_revenues(
...
+    * Rename columns to align with other EIA tables
+
+    Args:
+        raw_eia923__emissions_control: The raw ``raw_eia923__emissions_control``
+        dataframe.
+
+    Returns:
+        Cleaned ``_core_eia923__emissions_control`` dataframe ready for harvesting.
+    """
+    df = raw_eia923__emissions_control.copy()
```
zaneselvans: Is this just precautionary, or was there a real issue? We're trying avoid doing blanket dataframe copies like this when possible because they can unnecessarily increase memory use, which ends up limiting our ability to parallelize the overall data processing pipeline.

src/pudl/metadata/resources/eia923.py
```diff
@@ -958,6 +958,51 @@
...
+        "schema": {
+            "fields": [
+                "plant_id_eia",
+                "report_year",
+                "acid_gas_removal_efficiency",
+                "annual_nox_emission_rate_lb_per_mmbtu",
+                "data_maturity",
+                "environmental_equipment_name",
+                "fgd_electricity_consumption_mwh",
+                "fgd_sorbent_consumption_1000_tons",
```
zaneselvans: We do already have this column with the `_1000_tons` suffix, but there's also #4301 / #4426 (a bit stale) which is about standardizing on tons across the board for any `_1000_tons` columns across the board. So we can probably leave the column as is here and deal with that rename / unit conversion separately (unless you're enthusiastic about updating it across the board)

src/pudl/transform/eia923.py
```diff
@@ -1640,3 +1640,65 @@ def _core_eia923__yearly_byproduct_expenses_and_revenues(
...
+
+    # Encode operational_status
+    df["operational_status"] = df.operational_status.str.upper().map(
+        pudl.helpers.label_map(
+            CODE_METADATA["core_eia__codes_operational_status"]["df"],
+            from_col="code",
+            to_col="operational_status",
+            null_value=pd.NA,
+        )
+    )
```
zaneselvans: We have a slightly arcane but automated way of encoding categorical values based on the foreign key relationships to the coding tables which should be used here instead -- it'll encode any columns with those FK relationships, and for this table in particular, provide more granular categories that are specific to emissions equipment (e.g. `OZ` means "operates only during ozone season"):

```suggestion
    df = PUDL_PACKAGE.encode(df)
```

However, I don't think this will work until the database schema / metadata is updated to reflect the existence of the new table (and all the automatically detected FK relationships linking it to the coding tables).

src/pudl/metadata/resources/eia923.py
```diff
@@ -958,6 +958,51 @@
...
+                "particulate_test_date",
+                "so2_control_id_eia",
+                "so2_removal_efficiency_tested",
+                "so2_removal_efficiency_annual",
+                "so2_test_date",
+            ],
+            "primary_key": [
+                "plant_id_eia",
+                "report_year",
+            ],
```
zaneselvans: At this point in the table's processing, I don't think we should expect it to have a natural primary key. My guess is that if we normalized it into several narrower tables, each of which pertained to just a single type of emissions control equipment with its own ID, each of them would have a natural primary key of `(report_date, plant_id_eia, equipment_id)` but lots of plants won't have lots of kinds of equipment, so looking at all those IDs together in the wide unnormalized table will include a bunch of NA values.

src/pudl/metadata/resources/eia923.py
```diff
@@ -958,6 +958,51 @@
...
+                "Actual rate and removal efficiency for air emissions reported by "
+                "thermoelectric or combined-cycle power plants with a total steam turbine capacity "
+                "greater than or equal to 10 megawatts."
+            ),
+            "additional_source_text": "(Schedule 8C)",
+        },
+        "schema": {
+            "fields": [
+                "plant_id_eia",
+                "report_year",
```
zaneselvans: While the reporting frequency for this table is annual, we use `report_date` with a datetime type and January 1st as the date so the column can be used in conjunction with other datetime columns. So this should be converted to a date in the transform and renamed to `report_date`.

I see we missed this in some recently integrated EIA tables so I create an issue here #4741

src/pudl/transform/eia923.py
```diff
@@ -1640,3 +1640,65 @@ def _core_eia923__yearly_byproduct_expenses_and_revenues(
     df.columns = df.columns.str.replace("_1000_dollars", "")
 
     return df
+
+
+@asset(io_manager_key="pudl_io_manager")
+def _core_eia923__emissions_control(
```
zaneselvans: This should be renamed to include the temporal resolution, which I think is yearly, so `_core_eia923__yearly_emissions_control`, and the resource definition will need to be updated as well.

zaneselvans (commented):
* I left inline comments on the data cleaning and naming questions.
* I went ahead and added a preliminary alembic migration to get the new table and its foreign keys into the schema.
* Having that information available allows us to use the generic `PUDL_PACKAGE.encode(df)` which preserves all of the codes and links to the operational status coding table.
* I moved the column renames upstream into the spreadsheet column mapping file to provide an example of how that works, since I think we'll probably want to standardize some of the other column names in here too.
* I also merged `main` into the branch since a lot of changes have accumulated include the transposition of all those spreadsheet column mapping files.

## Percentages vs. Fractions

* One kind of data entry error that we've come across frequently is that some respondents mix up fractions (like 0.75) and percentages (like 75).
* We should plot a histogram of the values in the percentage efficiency columns, and see if there's an abnormal proportion of values in the 0.0-1.0 range. If so then we might want to convert those from fractions to percentages (I'm guessing that most of the real efficiencies here should be much larger than 1%).
* Looking at these histograms I see that virtually all of the values are between 0.0 and 1.0, so these aren't percentages, and that's fine. We should change the column descriptions to reflect that fact and do not need to add `_pct` suffixes to the column names (since "efficiency" is typically between 0 and 1).
* The one exception looks like `particulate_removal_annual_efficiency` which has a handful of values between 1 and 10 that we should look at more closely.

## Primary Key

* At this point in the table's processing, I don't think we should expect it to have a natural primary key.
* My guess is that if we normalized it into several narrower tables, each of which pertained to just a single type of emissions control equipment with its own ID each of them would have a natural primary key of `(report_date, plant_id_eia, equipment_id)`.
* Skimming the table, I suspect that the equipment IDs are often the same as the plant component that they're attached to, like the boiler or smokestack (which is often distinct from the boiler, since in some cases multiple boilers or gas turbines will share the same smokestack).
* But because there are many different kinds of emissions control equipment, and many plants will not have one or more of them installed, while other plants may have more than one of a given type installed, so long as all of the different equipment types are being stored in the same table and we can't have NA values as part of the primary key (a limitation of SQLite) I think we should leave the primary key undefined for now (SQLite will assign an integer `row_id`).
* In a database that allowed NULL primary key values I would expect that the PK here would be:
  * `report_year`
  * `plant_id_eia`
  * `mercury_control_id_eia`
  * `nox_control_id_eia`
  * `particulate_control_id_eia`
  * `so2_control_id_eia`
* But poking at the data I see there still a lot of duplicates even with all of those values included in the index / groupby.
* It looks like the distribution of duplicates is also pretty even across all the years, so it's not just an "old data is messy" issue.
* A useful plant for exploring this kind of thing is Barry (`plant_id_eia == 3`) because it's big and complicated. It also comes up at the top of every table sorted by plant ID ;-)
* Looking at Barry it has no useful IDs in 2012, but in subsequent years there are a bunch of IDs, and it seems to be organized such that in any row there is only one ID set, so while each row has columns for all the different kinds of equipment, in practice only a single piece of equipment is being identified. If this structure is common across lots of plants it would suggest normalizing into separate core tables organized each different type of emissions control equipment.
* But in some years Barry does describe more than one type of equipment on a single line, but only when the IDs are the same. Weird, but that wouldn't keep us from normalizing the table by type of equipment.

## What to do now?

* For the moment let's focus on getting all the column names standardized, and clean up the data, then we can look at whether there's a simple normalization we can apply to create equipment-type specific tables organized by equipment ID.

zaneselvans:
@alexclippinger sorry for the delay in reviewing this! Let me know if you'd like to talk through anything. Thanks for your help.

zaneselvans:
## Efficiency Fractions vs. Percent

```python
ec_core = get_asset("_core_eia923__emissions_control")
ec_core.filter(like="efficiency").hist(bins=50)
```

image

## Malformed dates

```python
ec_core.particulate_test_date.str.split("-").str[1].value_counts().sort_index().plot.bar()
```
image

```python
ec_core.particulate_test_date.str.split("-").str[0].value_counts().sort_index().plot.bar()
```
image

```python
ec_core.so2_test_date.str.split("-").str[1].value_counts().sort_index().plot.bar()
```
image

```python
ec_core.so2_test_date.str.split("-").str[0].value_counts().sort_index().plot.bar()
```
image

```

src/pudl/transform/eia923.py
alexclippinger: Copied from another transform function used as a starting point - removed

src/pudl/metadata/resources/eia923.py
alexclippinger: I can do that! Would you prefer in this PR or separately/follow-up?

zaneselvans: I think hold off for the moment since @msa856 has been working on it for a while and has just jumped back in.

aesharpe: I added it quickly!

src/pudl/transform/eia923.py
zaneselvans: Addressed in #5048

zaneselvans:
Work continued and merged in #5048

PR #4669, opened 2025-10-21
Transpose the extraction CSV's

### LINKED ISSUE #4397: Transpose package data column map csvs
### Overview

The column maps for each data source in package data have new records stored as columns instead of rows. This makes the csv harder to edit and view. It would be easier if we could transpose it to have each new month or year of data on a single row.

### Success Criteria

* [ ] Column map csvs are all transposed such that a new year or month of data is added on a new row
* [ ] Any reference to the column maps is modified to ensure it eads correctly
* [ ] Any documentation that refers to the column maps is updated to reflect the new structure

cmgosnell:
# Overview

Closes #4397.

All the lines of diffs all the time for every new partition of data in the mapped extracted files.

## What did you change?
Transposed the package_data extraction CSV's and updated the corresponding extractor classes.

This is going to end up being one gazillion lines changed... but the only code lines that will need changing are those in `pudl.extract` (almost entirely in `extractor.py`) and are pretty minimal and should stay that way.

this is my somewhat messy code snippet im using to actually do the transposing just for reference (this will not be checked in):

```python
dataset = "test"
pudl_paths = pudl.workspace.setup.PudlPaths()
map_path = os.path.join(Path(pudl.package_data.__file__).parents[0], dataset) # Get path to mapping CSVs
column_map_path = os.path.join(map_path , "column_maps")
csvs = [
    os.path.join(map_path, f) for f in os.listdir(map_path) if f.endswith(".csv")] + [
    os.path.join(column_map_path, f) for f in os.listdir(column_map_path) if f.endswith(".csv")
    ]
for og_csv_name in csvs:
    og_csv_df = pd.read_csv(og_csv_name, index_col=0)
    transposed_df = np.transpose(og_csv_df)
    transposed_df.to_csv(og_csv_name,index_label=og_csv_df.index.name)
```
^^ i hand fed it the datasets based on whether the package_data dir contains info for the standard extractors or not
& i had to revert some csv's that shouldn't had been mapped this way like `phmsagas/part_page_map.csv` and `censuspep/territories.csv`

## Documentation

# Testing

- [ ] **[in progress]** running fast elt locally
- [ ] will probably kick off a gha full build to make sure nothing downstream changed 

## To-do list


cmgosnell:
@e-belfer  i still have a lot of fast an full testing to do but i've spot tested basically all of the raw steps. once i get through fast and a local ci run & a review from you i'd like to kick off a gha full build to ensure nothing downstream changed unexpectedly

src/pudl/extract/extractor.py
```diff
@@ -181,7 +187,7 @@ def process_renamed(
 
     def get_page_cols(self, page: str, partition_selection: str) -> pd.RangeIndex:
         """Get the columns for a particular page and partition key."""
-        col_map = self._metadata._column_map[page]
+        col_map = self._metadata._column_map[page].T
```
krivard: Looks like this is the only place we're using `.T` now -- maybe swap the `.loc` order instead? a la

```
        return col_map.loc[
            [partition_selection],
            (col_map[partition_selection].notnull())
            & (col_map[partition_selection] != -1),
        ].columns
```

src/pudl/extract/extractor.py
```diff
@@ -58,9 +58,15 @@ def get_dataset_name(self) -> str:
...
         """Load metadata from a filename that is found in a package."""
-        return pd.read_csv(
+        df = pd.read_csv(
             importlib.resources.files(package) / filename, index_col=0, comment="#"
         )
+        # we are assigning the index which contains the partitions as a sting dtype.
+        # many of the partitions are years which are reasonably interpreted as ints,
+        # but some are dates or other actual strings. We force the partitions to be
+        # strings here and via _get_partition_selection.
+        df.index = df.index.astype(pd.StringDtype())
```
e-belfer, line 68: Are all of our partitions cast to strings somewhere, even the years? It's a bit surprising to me we have to do this, maybe it would be helpful to reference where in the code this happens in this comment here (which otherwise offers helpful clarification about what's going on).

src/pudl/extract/extractor.py
e-belfer: I find the `.T` clearer than the suggestion above, personally! :shrug:

cmgosnell: @krivard do you find the shorthandness of the `.T` opaque? I could swap it for `.transpose()`. I agree with @e-belfer  here that I personally like the simplicity of the transpose option.

cmgosnell: oh @krivard now i see wat you mean... we no longer need this bc the locs in the the loc below are just the things we could swap. i misunderstood you before! will swap and remove the `.T`

src/pudl/extract/extractor.py
cmgosnell: It was always converting into a string because the partitions used to be the column names and thus were always strings no matter what.

the cast to strings things via _get_partition_selection (which is references at the end of this comment). and also in a gazillion places all over the extractor code. I am tempted to try to rip it all up and see if it can just let the pandas read the dtypes but it seems more foolproof this way tbh.

src/pudl/extract/extractor.py
cmgosnell: it was slightly less straightforward than i wanted it to be without doing a `.T` but it reads similarly/slightly better imo

cmgosnell:
THEY PASSSSED

cmgosnell:
the docs builds failed again due to the scipy timeout thing again... which seems to just keep happening occasionally :-/ imma just try it again

zaneselvans:
It seems like the scipy docs thing started around the AWS disaster times, but if it keeps being an issue we can pretty harmlessly just comment that line out for now.

PR #4677, opened 2025-10-24
Drop PD901 ignore

krivard:
# Overview

Ruff linter rule PD901 (df as variable name) has been removed from ruff and no longer has any effect.

This PR drops it from our ignorelist, which should eliminate one of the nuisance messages in our precommit checks.


zaneselvans:
@krivard I added this change into #4689

krivard:
Fixed in #4869

Issue #4526, opened 2025-08-07
Hourly output tables claim they are denormalized when they are not


krivard:
### Overview

The hourly tables in the `out` layer are not denormalized, since they are prohibitively large (millions+).

Currently, the standardized text for all `out` tables says they are denormalized. This is misleading in the case of the hourly tables.

We don't want to demote the hourly tables to `core`, since they really are fully-processed and ready for use in analyses.

We should add standardized layer/processing-level text for hourly output tables to more accurately describe their level of processing as distinct from all other output tables.

### Success Criteria

How will we know that we're done?

* [ ] Hourly output tables no longer claim to be denormalized

We may want to include other information in the standardized text for hourly output tables, such as:

> explain that if you need additional attributes for filtering or calculation, they need to be merged in, and that if you’re going to aggregate e.g. hourly to monthly you should do that before the merge.

### Next steps

* [ ] Draft standard text
* [ ] Add logic to table description flow to consider timeseries resolution when selecting layer/processing-level text

krivard:
There are 76 tables in the output layer.

There are 7 hourly tables affected by this issue:

- out_eia930__hourly_aggregated_demand
- out_eia930__hourly_operations
- out_eia930__hourly_subregion_demand
- out_ferc714__hourly_estimated_state_demand
- out_ferc714__hourly_planning_area_demand
- out_gridpathratoolkit__hourly_available_capacity_factor
- out_vcerare__hourly_available_capacity_factor

Zane reported finding other tables which were in the output layer and not hourly but also not denormalized either, so I did a quick survey pass with the following heuristics:

- an output table counts as denormalized if there is a core table with the same slug, and the output table has all the core columns plus some plausibly-denormalization columns (with exceptions for two weirdos: out_eia923__monthly_fuel_receipts_costs appears denormalized but drops a bunch of code columns from core; out_sec10k__quarterly_filings had core already pretty denormalized and only added a url)
- an output table counts as denormalized if there is no core table with the same slug, but it has at least two columns with "_name" in the name that are plausibly denormalization columns. Will also accept only one "_name"-like column if each row in the table plausibly refers to a single entity, such as out_eia__yearly_utilities

Using these heuristics there are 6-7 other output tables which are not hourly but also do not appear to be denormalized:

- out_eia__yearly_assn_plant_parts_plant_gen - has codes and ids with no names (okay it has the ferc account name but that's it)
[...]

PR #4700, opened 2025-10-27
Add description support for narrow output tables

krivard:
# Overview

Addresses #4526.

7 hourly output tables and 7 additional output tables ([see post for how non-hourly tables were identified](https://github.com/catalyst-cooperative/pudl/issues/4526#issuecomment-3234754931)) currently claim to be in wide/denormalized format when they in fact have been left narrow.

Add variant descriptor text for narrow output tables. Use it by default for hourly output tables, and specify it manually to cover the remaining 7 tables.

## What did you change?

* Add an "out_narrow" layer code. This code will never be automatically detected from a table name, but can be applied through other means.
* Add corresponding descriptive text
* Apply "out_narrow" to hourly output tables, unless a manual override exists
* Install manual overrides on non-hourly narrow output tables

## Documentation

# Testing

Viewed hourly, non-hourly narrow, and regular output tables in the wizard to confirm the correct text appears in the Processing section.

## To-do list

### DIFF
--- a/src/pudl/metadata/descriptions.py
+++ b/src/pudl/metadata/descriptions.py
@@ -29,7 +29,8 @@ layer_descriptions: dict = {
         "Data has been expanded into a wide/denormalized format, with IDs and codes "
         "accompanied by human-readable names and descriptions."
     ),
-    "out_narrow": (  # this is a pseudo layer that does not appear in table names, but
+    "out_narrow": (
+        # this is a pseudo layer that does not appear in table names, but
         # allows us to accurately describe the few out tables that remain in
         # normalized format.
         "Data is ready for use in analyses, but for practical reasons has not been "


src/pudl/metadata/descriptions.py
```diff
@@ -27,6 +29,10 @@
         "Data has been expanded into a wide/denormalized format, with IDs and codes "
         "accompanied by human-readable names and descriptions."
     ),
+    "out_narrow": (
+        "Data is ready for use in analyses, but has not been denormalized and remains "
+        "in narrow format."
```
krivard: This can be extended on a per-table basis by setting "additional_layer_text"

krivard: add "for practical purposes"

krivard:
```suggestion
        "Data is ready for use in analyses, but for practical reasons has not been "
        "denormalized and remains in narrow format."
```

src/pudl/metadata/resources/sec10k.py
```diff
@@ -251,11 +251,12 @@
...
+            "layer_code": "out_narrow",
             "additional_details_text": """
-We use the company name reported in association each name change block in the company
-information table to fill in the most recent value of ``company_name_new``. Roughly
-1,000 reported "name changes" in which the old and new names were identical have been
-dropped.""",
+We use the company name reported in association with each name change block in the
+company information table to fill in the most recent value of ``company_name_new``.
+Roughly 1,000 reported "name changes" in which the old and new names were identical
+have been dropped.""",
```
krivard, line 259: (unrelated copy edit to resolve missing word)

src/pudl/metadata/descriptions.py
```diff
@@ -29,6 +29,10 @@
         "Data has been expanded into a wide/denormalized format, with IDs and codes "
         "accompanied by human-readable names and descriptions."
     ),
+    "out_narrow": (
```
krivard, line 32:
```suggestion
    "out_narrow": (  # this is a pseudo layer that does not appear in table names, but
                     # allows us to accurately describe the few out tables that remain in
                     # normalized format.
```

src/pudl/metadata/descriptions.py
```diff
@@ -29,6 +29,12 @@
         "Data has been expanded into a wide/denormalized format, with IDs and codes "
         "accompanied by human-readable names and descriptions."
     ),
+    "out_narrow": (  # this is a pseudo layer that does not appear in table names, but
+        # allows us to accurately describe the few out tables that remain in
```
krivard:
```suggestion
    "out_narrow": (
        # this is a pseudo layer that does not appear in table names, but
        # allows us to accurately describe the few out tables that remain in
```

src/pudl/metadata/descriptions.py
```diff
@@ -334,6 +341,14 @@ def _generic_component(
     @component
     def layer(self, settings, defaults):
         """Compute the processing layer component of the resource description."""
+        if (  # use narrow variant for hourly output tables
```
e-belfer, line 344: Non-blocking:
Given that there are only seven of them, I'm slightly inclined to manually override them rather than add more complexity that will require digging into the code to trace the origins of. But in practice, I can't imagine denormalizing any hourly output tables, so this is entirely non-blocking.

e-belfer (approved):
Non-blocking question: The original issue noted wanting to add a warning about usage. Is that still desired here, or out of scope? (Either way, this infrastructure makes it trivial to update this text in the future as desired).

Otherwise, one other non-blocking comment, but looks good to merge to me.

Verified using `resource_description` that all looks as expected for the following tables:

- `out_eia861__yearly_utility_service_territory`
- `out_eia923__yearly_generation_fuel_by_generator_energy_source_owner`
- `out_eia930__hourly_aggregated_demand`

Verified no changes for a handful of plain old output tables:
- `out_eia860__yearly_emissions_control_equipment`
- `out_ferc1__yearly_all_plants`

src/pudl/metadata/descriptions.py
krivard: we can always back it off to manual if we find a use case for hourly denormalized!

krivard:
We played with some language about usage, but couldn't find something we liked that would work as part of the standardized text.

Issue #4587, opened 2025-09-01
Show whether the data is currently being collected or not (code support)


krivard:

krivard:
Consider doing this as a usage warning instead

cmgosnell:
I think this one is donzo now so imma close but @krivard  if im reading this wrong feel free to repoen!

PR #4702, opened 2025-10-27
Add usage warning for tables not receiving ongoing updates in PUDL

krivard:
# Overview

Addresses #4587.

Show whether the data is currently being collected or not.

## What did you change?

* Add a usage warning called `discontinued_pudl` with text "PUDL does not currently update its copy of this data." This is distinct from the warning called `discontinued_data`, which refers to data which is no longer available from the source ("The data is no longer being collected or reported in this way.")
* Label sec10k tables as `discontinued_pudl`
* Fill in missing `discontinued_data` labels for tables in 861 and gridpathratoolkit where data has been discontinued at the source, and layer in some cross-references for the 861 tables for which newer data may be available in a new format

## Documentation

# Testing

Since most of our data sets have 0-1 years of lag, I identified candidate tables for this warning by running the following code to list tables whose most recent data is earlier than 2024:

```
import re
import pandas as pd

from pudl.metadata.resources import RESOURCE_METADATA
from pudl.metadata.classes import PudlResourceDescriptor, Resource, DataSource
from pudl.metadata.descriptions import ResourceDescriptionBuilder

def has_discontinued(resolved):
    for warning in resolved.usage_warnings:
        if warning.type.startswith("discontinued"):
            return True
    return False

YEAR = re.compile(r"\d{4}")

availability = pd.read_csv(
    "../../dbt/seeds/etl_full_row_counts.csv",dtype={"partition":"string"}).groupby("table_name").partition.max()
by_availability = defaultdict(list)

for resource_id, resource_dict in sorted(RESOURCE_METADATA.items()):
    desc = ResourceDescriptionBuilder(
        resource_id=resource_id,
        settings=resource_dict,
    ).build()
    resource_availability = pd.NA
    if resource_id in availability.index:
        resource_availability = availability.loc[resource_id]
    if pd.isna(resource_availability):
        candidates = []
        for src in resource_dict["sources"]:
            coverage = DataSource.from_id(src).get_temporal_coverage()
            if coverage:
                start, _, end = coverage.replace("through ","-").partition("-")
                m = YEAR.search(end)
                if m:
                    candidates.append(m.group(0))
        if candidates:
            resource_availability = min(candidates)
    if pd.isna(resource_availability):
        resource_availability = "UNKN"
    by_availability[(resource_availability, has_discontinued(desc))].append(resource_id)
for yr, discontinued in sorted(by_availability.keys(), reverse=True):
    if ((yr == "UNKN") or (yr < "2024")):
        print(f"{yr} {discontinued}\n\t{'\n\t'.join(by_availability[(yr, discontinued)])}")
```

with output:

```
UNKN False
	core_gridpathratoolkit__assn_generator_aggregation_group
	core_pudl__codes_datasources
	core_pudl__codes_imputation_reasons
	core_pudl__codes_subdivisions
	core_pudl__entity_plants_pudl
	core_pudl__entity_utilities_pudl
	out_censusdp1tract__counties
	out_censusdp1tract__states
	out_censusdp1tract__tracts
2023 False
	core_eia176__yearly_gas_disposition_by_consumer
	core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities
	core_sec10k__assn_exhibit_21_subsidiaries_and_filers
	core_sec10k__assn_sec10k_filers_and_eia_utilities
	core_sec10k__changelog_company_name
	core_sec10k__quarterly_company_information
	core_sec10k__quarterly_exhibit_21_company_ownership
	core_sec10k__quarterly_filings
	out_sec10k__changelog_company_name
	out_sec10k__parents_and_subsidiaries
	out_sec10k__quarterly_company_information
	out_sec10k__quarterly_filings
	out_vcerare__hourly_available_capacity_factor
2021 False
	out_gridpathratoolkit__hourly_available_capacity_factor
2015 False
	core_eia861__yearly_distributed_generation_fuel
	core_eia861__yearly_distributed_generation_misc
	core_eia861__yearly_distributed_generation_tech
2012 True
	core_eia861__yearly_demand_side_management_ee_dr
	core_eia861__yearly_demand_side_management_misc
	core_eia861__yearly_demand_side_management_sales
	core_eia861__yearly_green_pricing
```

I then investigated and labeled as follows (going from least to most recent):

* EIA 861 Distributed Generation tables: discontinued by EIA in 2016 when they switched to reporting these figures in Non Net Metering instead. `discontinued_data`
* GridPath RA Toolkit tables: discontinued by GridPath; no new data is anticipated. `discontinued_data`
* VCE RARE tables: 2024 data isn't available yet, so this is indistinguishable from a normal update schedule. No warning needed.
* SEC 10-k tables: discontinued by us due to lack of funding. `discontinued_pudl`
* EIA 176 tables: new table, driven by external contributors. not enough information yet; no warning needed.
* 🚨 Census DP1 tables: didn't label these yet; see Slack thread for discussion
* PUDL: on a normal update schedule, no warning needed.

## To-do list

- [x] label SEC 10-K tables
- [x] survey for other tables where more recent data is available but not part of quarterly or annual updates

### DIFF
--- a/src/pudl/metadata/resources/eia861.py
+++ b/src/pudl/metadata/resources/eia861.py
@@ -2,6 +2,11 @@
 
 from typing import Any
 
+DISTRIBUTED_GENERATION_NON_NET_METERING_TRANSITION = """The raw EIA861 distributed generation (DG) table (split into three normalized tables in
+PUDL) was renamed in 2016 to Non-Net Metering to prevent double counting. The data in
+the Non-Net Metering table (2016+) are split by sector, contain fuel cell information,
+and convert capacity reported in DC units to AC units."""
+
 RESOURCE_METADATA: dict[str, dict[str, Any]] = {
     "core_eia861__yearly_advanced_metering_infrastructure": {
         "description": {
@@ -256,10 +261,15 @@ representative from EIA.""",
     "core_eia861__yearly_distributed_generation_fuel": {
         "description": {
             "additional_summary_text": "the energy sources used for utility or customer-owned distributed generation capacity.",
-            "additional_details_text": """The raw EIA861 distributed generation (DG) table (split into three normalized tables in
-PUDL) was renamed in 2016 to Non-Net Metering to prevent double counting. The data in
-the Non-Net Metering table (2016+) are split by sector, contain fuel cell information,
-and convert capacity reported in DC units to AC units.""",
+            "usage_warnings": [
+                "discontinued_data",
+            ],
+            "additional_details_text": f"""From 2016 on, there's no great match for the data found in this table.
+You may be able to compute an analog for fuel percents using the capacity and tech class
+figures in :ref:`core_eia861__yearly_non_net_metering_customer_fuel_class`, but no
+corresponding consumption or generation information is available in the new format.
+
+{DISTRIBUTED_GENERATION_NON_NET_METERING_TRANSITION}""",
         },
         "schema": {
             "fields": [
@@ -280,10 +290,13 @@ and convert capacity reported in DC units to AC units.""",
         # TODO: might want to rename this table to be _capacity
         "description": {
             "additional_summary_text": "the capacity and quantity of utility or customer-owned distributed generation.",
-            "additional_details_text": """The raw EIA861 distributed generation (DG) table (split into three normalized tables in
-PUDL) was renamed in 2016 to Non-Net Metering to prevent double counting. The data in
-the Non-Net Metering table (2016+) are split by sector, contain fuel cell information,
-and convert capacity reported in DC units to AC units.""",
+            "usage_warnings": [
+                "discontinued_data",
+            ],
+            "additional_details_text": f"""From 2016 on, data similar to that found in this table can be found in
+:ref:`core_eia861__yearly_non_net_metering_misc`.
+
+{DISTRIBUTED_GENERATION_NON_NET_METERING_TRANSITION}""",
         },
         "schema": {
             "fields": [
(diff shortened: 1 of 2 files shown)


src/pudl/metadata/warnings.py
```diff
@@ -27,5 +27,6 @@
         "for details."
     ),
     "discontinued_data": "The data is no longer being collected or reported in this way.",
+    "discontinued_pudl": "PUDL does not currently update its copy of this data.",
```
zaneselvans, line 30: Might we want to differentiate between data that we'd like to be updating (if we had the resources) and data that's we've decided to deprecate for other reasons?

zaneselvans (commented):
* The `gridpathratoolkit` dataset should probably get one of these too, since it was a one-time compilation that's not being updated by the original creators.
* Some kind of note about `nrelatb` being silently not updated on the normal schedule (without an announcement about it) might also be good.

src/pudl/metadata/warnings.py
krivard: We can deal with nuance like that in Additional Details -- that will also cover cases where the original source changed the data so significantly that we're no longer updating our copy

src/pudl/metadata/resources/eia861.py
```diff
@@ -2,6 +2,11 @@
 
 from typing import Any
 
+DISTRIBUTED_GENERATION_NON_NET_METERING_TRANSITION = """The raw EIA861 distributed generation (DG) table (split into three normalized tables in
+PUDL) was renamed in 2016 to Non-Net Metering to prevent double counting. The data in
+the Non-Net Metering table (2016+) are split by sector, contain fuel cell information,
+and convert capacity reported in DC units to AC units."""
+
```
krivard, line 9: This is a refactor to specify a single copy of this text which can be used by everybody who needs it, instead of copy-pasting

Issue #4704, opened 2025-10-27
Add usage warning to distinguish harvested tables from other multi-source tables (code support + 8 tables)


krivard:
- [ ] Add new warning about harvesting
- [ ] Label affected tables: core harvest 8 + downstreams with 3 or more columns from the harvest list
- [ ] Add Usage Warnings explainer
- [ ] Add test to scream if a downstream table has 3 or more columns from the harvest list and doesn't have the warning

krivard:
There are 8 tables which are harvesting assets, that is, resulting from `pudl.transform.eia.harvest_entity_tables()`:

* core_eia__entity_plants
* core_eia__entity_utilities
* core_eia__entity_boilers
* core_eia__entity_generators
* core_eia860__scd_plants
* core_eia860__scd_utilities
* core_eia860__scd_boilers
* core_eia860__scd_generators

Each of these tables has a list of harvested columns.

There are a further shitton of tables (=110) downstream from these assets which include one or more columns with the same name as one on the harvested list. This does not _strictly speaking_ guarantee that those columns came from the harvested table (the asset could be a multi-source asset and drawing that column from a non-harvested source), but it is a strong indicator.

Do we want to list the harvest warning for all tables that may include harvested values, or only the tables that are the core source of those harvested values?

Do we want to list the harvest warning for downstream tables only if they have a significant number of potentially-harvested columns? Like, more than 3? more than 5?

<details><summary>Here is the exhaustive list of every downstream table that might include harvested values, and the columns they include that might have come from the harvesting process.</summary>

The format is: 

`[harvested table] [downstream table] [pipe-delimited list of columns in the downstream table that share a name with a harvested column in the harvested table]`

```
core_eia__entity_plants out_eia__yearly_boilers latitude|city|longitude|street_address|plant_name_eia|zip_code|state|county
core_eia__entity_plants out_eia923__fuel_receipts_costs state|plant_name_eia
core_eia__entity_plants out_eia__monthly_generators latitude|city|longitude|street_address|plant_name_eia|zip_code|state|county
core_eia__entity_plants out_eia923__yearly_fuel_receipts_costs state|plant_name_eia
core_eia__entity_plants out_eia__yearly_plants latitude|city|longitude|street_address|plant_name_eia|zip_code|state|county
core_eia__entity_plants core_epacems__hourly_emissions state
core_eia__entity_plants out_eia923__yearly_generation_fuel_by_generator plant_name_eia
[...]

PR #4705, opened 2025-10-27
Add usage warning to distinguish harvested tables from other multi-source tables

krivard:
# Overview

Addresses #4704.

**This PR is based on #4806 and should merge after it (after a rebase).**

~We automatically apply a multi-source warning to tables whose metadata claim more than one data source.
A subset of these tables are multi-source because they are _harvested_, which requires some additional considerations by users.~ <-- that was the original aim, but the new one is: Users frequently ask about discrepancies between harvested tables and the raw data, let's see if actually labeling tables affected by harvesting helps! Lots more discussion + analysis in the issue thread for the curious.

## What did you change?

* Add a usage warning for harvested tables.
* Install the harvested warning on tables flagged as harvested in [the metadata tracking sheet](https://docs.google.com/spreadsheets/d/1Hv82QN5mVKieRC8B34LxQIsWw3WMjfkarRLgIfvYLX8/edit?gid=120132036#gid=120132036)
* Do a downstream search from those 8 tables to see if columns with the same names show up. If a table has three or more columns that may have come from a harvest table, flag it as well.
* Add appropriate explanatory text to Additional Details for the initial 8 harvest tables
* Add complimentary explanatory text to Additional Details for the downstream tables
* Add a data dictionary page for usage warnings, and explain more of the harvesting details there.
* Centralize utilities for harvest-releated additional details text and description dict wrangling in `pudl.metadata.resource_helpers`

## Documentation

- [x] Update relevant data dictionary docs (see `docs/data_dictionary`).

# Testing

Previewed results in wizard.

Built documentation and browsed results:
* Data Dictionaries > PUDL Usage Warnings
* Data Dictionaries > PUDL Data Dictionary (& ^F "canonical" and "consistent")

Ran `pytest test/unit/metadata_test.py::test_merge_descriptions` and it passed.

## To-do list

- [x] Finalize text of harvesting warning
- [x] Organize new functions somewhere more obvious

### DIFF
--- a/src/pudl/metadata/resource_helpers.py
+++ b/src/pudl/metadata/resource_helpers.py
@@ -1,0 +1,80 @@
+"""Functions for resource metadata.
+
+These live in pudl.metadata and not in pudl.metadata.resources because we have
+machinery that iterates over the contents of pudl.metadata.resources and needs
+each module there to actually store resource metadata.
+"""
+
+from typing import Any
+
+HARVESTING_DETAIL_TEXT = """EIA reports many attributes in many different tables across
+EIA-860 and EIA-923. In order to compile tidy, well-normalized database tables, PUDL
+collects all instances of these values and and chooses a canonical value. By default,
+PUDL chooses the most consistently reported value of a given attribute as long as it
+is at least 70% of the given instances reported. If an attribute was reported
+inconsistently across the original EIA tables, then it will show up as a
+null value."""
+
+
+def canonical_harvested_details(entities: str, is_static: bool) -> str:
+    """Generate additional details text for one of the eight core harvested tables.
+
+    We have one core harvested table for each combination of (plants, utilities,
+    boilers, generators) X (static cols, annual cols):
+
+      * ``core_eia__entity_{plants|utilities|boilers|generators}`` - static cols
+      * ``core_eia860__scd_{plants|utilities|boilers|generators}`` - annual cols
+
+    This text helps users cross reference where the canonical values for each
+    type of entity come from, and why they may differ from a value they find in a
+    raw source.
+
+    Args:
+        entities: string containing the plural of an entity type; e.g., "plants"
+        is_static: True if the table this text is destined for contains the static cols
+            for the entity, False otherwise. Static cols are stored in tables with a name
+            like "core_eia__entity_X", and annual cols are stored in tables with a name
+            like "core_eia860__scd_X".
+    """
+    return f"""This is one of two tables where canonical
+values for {entities} are set. It contains values which are expected to {"remain fixed" if is_static else "vary slowly"}, while
+:ref:`core_eia{"860" if is_static else ""}__{"scd" if is_static else "entity"}_{entities}` contains those {"which may vary from year to year" if is_static else "expected to remain fixed"}.
+{HARVESTING_DETAIL_TEXT}
+All tables downstream of this one inherit the canonical values established here."""
+
+
+def inherits_harvested_values_details(entities: str) -> str:
+    """Generate additional details text for a table which inherits harvested values from one of the eight core harvested tables.
+
+    A table inherits harvested values from one of the eight core harvested tables
+    if it is downstream of one or more tables ``core_eia__entity_{plants|utilities|boilers|generators}``
+    or ``core_eia860__scd_{plants|utilities|boilers|generators}`` and includes one or
+    more columns from the static or annual column lists in :data:`pudl.metadata.resources.ENTITIES`.
+
+    We have chosen to only add this warning to tables that inherit 3 or more columns from harvested
+    tables.
+
+    Args:
+        entities: a prose string listing which harvested entities contributed
+            columns to this table; e.g., "generators and plants" for a table with
+            ``core_eia860__scd_generators`` and ``core_eia860__scd_plants`` upstream.
+    """
+    return f"""This table inherits canonicalized values for {entities}.
+{HARVESTING_DETAIL_TEXT}"""
+
+
+def merge_descriptions(left: dict[str, Any], right: dict[str, Any]) -> dict[str, Any]:
+    """Merge two description dictionaries."""
+    result = {}
+    result.update(left)
+    for key in right:
+        if key in result:
+            if key == "usage_warnings":
+                result[key] = result[key] + right[key]
+            elif key == "additional_details_text":
+                result[key] = f"{result[key]}\n\n{right[key]}"
+            else:
+                result[key] = f"{result[key]} {right[key]}"
+        else:
+            result[key] = right[key]
+    return result


docs/data_dictionaries/usage_warnings.rst
```diff
@@ -0,0 +1,103 @@
+=====================================================================================================
+PUDL Usage Warnings
+=====================================================================================================
```
krivard, line 3: Most text in this file is the same as the usage warning text in metadata/warnings.py. I'll flag the two places I added more info.

docs/data_dictionaries/usage_warnings.rst
```diff
@@ -0,0 +1,103 @@
...
+  The 70% threshold is the default, and we use different rules for columns with
+  additional requirements:
+
+  * Latitude and longitude are particularly noisy, and 70% consistency is not
+    attainable very often. We use the 70% threshold when possible, but for
+    records that don't meet the threshold, we do a second pass after rounding
+    latitude and longitude to the nearest tenth of a degree.
+  * Generator operating date has an unusual pattern of missingness that permits
+    the most recently reported operating date to be reliable when 70%
+    consistency cannot otherwise be reached.
```
krivard, line 64: This stuff is new

docs/data_dictionaries/usage_warnings.rst
```diff
@@ -0,0 +1,103 @@
...
+Derived values:
+  Contains columns derived from inputs and not originally present in sources.
+
+Discontinued by the source:
+  The data is no longer being collected or reported in this way.
+
+Discontinued by us:
+  PUDL does not currently update its copy of this data.
+
+  If you would be interested in funding additional updates to this data, get in touch!
```
krivard, line 22: This line is new

src/pudl/metadata/resources/eia923.py
```diff
@@ -2,6 +2,26 @@
 
 from typing import Any
 
+from pudl.metadata.resources.eia import inherits_harvested_values_details
+
+
+def merge_descriptions(left: dict[str, Any], right: dict[str, Any]) -> dict[str, Any]:
+    """Merge two description dictionaries."""
```
krivard: we need this because 923 has a buuuuuunch of description dictionaries that are shared between core and out, but only the out layer tables need the harvest warnings

krivard: I guess this could live somewhere more central but 923 is the only source that had this particular problem

test/unit/metadata_test.py
```diff
@@ -232,6 +233,28 @@ def test_frictionless_data_package_resources_populated():
...
+        "usage_warnings": ["blue"],
+    }
+    result = {
+        "additional_summary_text": "one fish",
+        "additional_details_text": "two\n\npoisson",
+        "usage_warnings": ["red", "blue"],
+    }
+    assert merge_descriptions(left, right) == result
+    # make sure we didn't accidentally modify the source list during the merge
+    assert left["usage_warnings"] == ["red"]
```
krivard, line 255: this line is a regression test; my first draft just kept appending harvest warnings onto the shared 923 dicts until there were like 8 copies

src/pudl/metadata/warnings.py
```diff
@@ -29,4 +29,10 @@
     "discontinued_data": "The data is no longer being collected or reported in this way.",
     "discontinued_pudl": "PUDL does not currently update its copy of this data.",
     "experimental_wip": "This table is experimental and/or a work in progress and may change in the future.",
+    "harvested": (
+        "Data has been drawn from several EIA sources which are not always consistent with each other, and PUDL has "
+        "overridden unusual entries in this table with a more consistent value. See "
+        "`Harvesting <https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/usage_warnings.html#harvested>`_ "
+        "for details."
```
krivard: had to make this shorter to still be skimworthy, but I tried to make up for it by fleshing out the warnings dictionary page

docs/data_dictionaries/usage_warnings.rst
krivard: We could absolutely layer in detail to more of these, but let's do that in a separate pass.

I thought about implementing this as a generated file like `codes_and_labels` is so that we don't have to worry about keeping the warnings texts in sync, but that seemed like a big lift? I can look at whether there's a middle way if having a static copy like this feels like too much of a maintenance nightmare, lmk

src/pudl/metadata/resources/eia.py
```diff
@@ -6,6 +6,54 @@
...
+        is_static: True if the table this text is destined for contains the static cols
+            for the entity, False otherwise. Static cols are stored in tables with a name
+            like "core_eia__entity_X", and annual cols are stored in tables with a name
+            like "core_eia860__scd_X".
+    """
+    return f"""This is one of two tables where canonical
+values for {entities} are set. It contains values which are expected to {"remain fixed" if is_static else "vary slowly"}, while
+:ref:`core_eia{"860" if is_static else ""}__{"scd" if is_static else "entity"}_{entities}` contains those {"which may vary from year to year" if is_static else "expected to remain fixed"}.
+{HARVESTING_DETAIL_TEXT}
+All tables downstream of this one inherit the canonical values established here."""
```
krivard: Sample output for core_eia860__scd_boilers:

> This is one of two tables where canonical values for boilers are set. It contains values which are expected to vary slowly, while [core_eia__entity_boilers](https://github.com/catalyst-cooperative/pudl/pull/4705/files#core-eia-entity-boilers) contains those expected to remain fixed. EIA reports many attributes in many different tables across EIA-860 and EIA-923. In order to compile tidy, well-normalized database tables, PUDL collects all instances of these values and and chooses a canonical value. By default, PUDL chooses the most consistently reported value of a given attribute as long as it is at least 70% of the given instances reported. If an attribute was reported inconsistently across the original EIA tables, then it will show up as a null value. All tables downstream of this one inherit the canonical values established here.

src/pudl/metadata/resources/eia.py
```diff
@@ -6,6 +6,54 @@
...
+    or ``core_eia860__scd_{plants|utilities|boilers|generators}`` and includes one or
+    more columns from the static or annual column lists in :data:`pudl.metadata.resources.ENTITIES`.
+
+    Args:
+        entities: a prose string listing which harvested entities contributed
+            columns to this table; e.g., "generators and plants" for a table with
+            ``core_eia860__scd_generators`` and ``core_eia860__scd_plants`` upstream.
+    """
+    return f"""This table inherits canonicalized values for {entities}.
+{HARVESTING_DETAIL_TEXT}"""
```
krivard: Sample output for out_pudl__yearly_assn_eia_ferc1_plant_parts:

> This table inherits canonicalized values for generators and plants. EIA reports many attributes in many different tables across EIA-860 and EIA-923. In order to compile tidy, well-normalized database tables, PUDL collects all instances of these values and and chooses a canonical value. By default, PUDL chooses the most consistently reported value of a given attribute as long as it is at least 70% of the given instances reported. If an attribute was reported inconsistently across the original EIA tables, then it will show up as a null value.

src/pudl/metadata/resources/eia.py
```diff
@@ -6,6 +6,54 @@
...
+All tables downstream of this one inherit the canonical values established here."""
+
+
+def inherits_harvested_values_details(entities: str) -> str:
+    """Generate additional details text for a table which inherits harvested values from one of the eight core harvested tables.
+
+    A table inherits harvested values from one of the eight core harvested tables
+    if it is downstream of one or more tables ``core_eia__entity_{plants|utilities|boilers|generators}``
+    or ``core_eia860__scd_{plants|utilities|boilers|generators}`` and includes one or
+    more columns from the static or annual column lists in :data:`pudl.metadata.resources.ENTITIES`.
```
krivard: We're only labeling tables that inherit 3 or more columns, but technically tables that inherit 1 or 2 columns count for this paragraph. is that confusing?

src/pudl/metadata/resources/eia.py
cmgosnell: this is great!

src/pudl/metadata/resources/eia923.py
cmgosnell: hm ya. this is a good solution to this setup. but i personally think it would be more reasonable to live somewhere more central. we don't really have a place for generic-ish resource helper functions but i wonder if making a new helpers module. I think moving this and all the new lil harvesting functions there would make sense and feel a lil more tidy to me. mostly i think it would help someone find it in the future (i would _never_ look in here for this.

src/pudl/metadata/warnings.py
cmgosnell: hm ya its difficult to be succinct but informative here.... The first part of the first sentence feels real good to me! But the second part i guess i don't know if this is the most relevant thing to highlight... I would suggest focusing more on choosing a single source of truth/canonical values/normalized tables.

(i am maybe being persnickety but the word `overridden` doesn't feel right even though we are dropping/loosing the inconsistently reported values by choosing **a value**)

```suggestion
        "Data has been drawn from several EIA sources which are not always consistent with each other, and PUDL chooses "
        "the most consistent or relevant value to have canonical values. See "
        "`Harvesting <https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/usage_warnings.html#harvested>`_ "
        "for details."
```

cmgosnell:
```suggestion
        "Data has been drawn from several EIA sources which are not always consistent with each other, and PUDL chooses "
        "the most consistent or relevant value to have a single source of truth. See "
        "`Harvesting <https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/usage_warnings.html#harvested>`_ "
        "for details."
```

cmgosnell:
```suggestion
        "Data has been drawn from several EIA sources which are not always consistent with each other, and PUDL chooses "
        "the most consistent or relevant value to have a single source of truth for tidy/well-normalized tables. See "
        "`Harvesting <https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/usage_warnings.html#harvested>`_ "
        "for details."
```

docs/data_dictionaries/usage_warnings.rst
cmgosnell: hm ya it would definitely feel nicer to pull this from `metadata.warnings.py` and add more color when we want to (like 🌾 )... at the very least I think we should add some internal comments here and/or in`warnings.py` nothing that this content is duplicated.

src/pudl/metadata/resources/eia.py
cmgosnell: my gut says this is okay!.... but maybe adding a note for our future selves will be helpful:

```suggestion
    A table inherits harvested values from one of the eight core harvested tables
    if it is downstream of one or more tables ``core_eia__entity_{plants|utilities|boilers|generators}``
    or ``core_eia860__scd_{plants|utilities|boilers|generators}`` and includes one or
    more columns from the static or annual column lists in :data:`pudl.metadata.resources.ENTITIES`.
    
    We have chosen to only add this warning to tables that inherit 3 or more columns from harvested
    tables.
```

src/pudl/metadata/resources/eia.py
cmgosnell: canonicalized is quite a word! but it is the thing we are doing.

docs/data_dictionaries/usage_warnings.rst
cmgosnell: this all looks real good!! just one small addition of the few columns we always make sure to get values for

```suggestion
  When there are multiple values reported for the same entity, in most cases,
  PUDL chooses the most consistent value reported which is found in at least 70%
  of available entries, and if no value occurs more than 70% of the time, PUDL
  fills in a null value. Internally, we refer to this process as **harvesting**.

  The 70% threshold is the default, and we use different rules for columns with
  additional requirements:

  * Latitude and longitude are particularly noisy, and 70% consistency is not
    attainable very often. We use the 70% threshold when possible, but for
    records that don't meet the threshold, we do a second pass after rounding
    latitude and longitude to the nearest tenth of a degree.
  * Generator operating date has an unusual pattern of missingness that permits
    the most recently reported operating date to be reliable when 70%
    consistency cannot otherwise be reached.
  * We set the consistency threshold to 0% for a few columns so that we always
    get a value: Plant name, utility name and prime mover code.
```

cmgosnell (changes requested):
this looks heckin great! made some light addition/clarification suggestion + a request to move the resource helper functions out into a central place (i will not die on that hill but i think it'd be nice and should be easy peasy)

src/pudl/metadata/resources/eia923.py
krivard: cool will move and then ping

src/pudl/metadata/warnings.py
krivard: Hmm what I like about `override` is that I think the most critical thing for users to know is that data in this table may not match what they get directly from EIA. how about:

```suggestion
        "Data has been drawn from several EIA sources which are not always consistent with each other, and PUDL chooses "
        "the most consistent or relevant value to facilitate cross-referencing even if that means some values"
        " will differ from the raw sources. See "
        "`Harvesting <https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/usage_warnings.html#harvested>`_ "
        "for details."
```

docs/data_dictionaries/usage_warnings.rst
krivard: I added a comment in warnings.py

src/pudl/metadata/warnings.py
```diff
@@ -29,4 +33,11 @@
     "discontinued_data": "The data is no longer being collected or reported in this way.",
     "discontinued_pudl": "PUDL does not currently update its copy of this data.",
     "experimental_wip": "This table is experimental and/or a work in progress and may change in the future.",
+    "harvested": (
+        "Data has been drawn from several EIA sources which are not always consistent with each other, and PUDL chooses "
+        "the most consistent or relevant value to facilitate cross-referencing even if that means some values"
+        " will differ from the raw sources. See "
+        "`Harvesting <https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/usage_warnings.html#harvested>`_ "
+        "for details."
```
krivard, line 41: does this establish the perspective we want or is it still off?

cmgosnell: ya i think this is relatively succinct (relatively lol) and has a great level of detail. Plus a link to more! great

cmgosnell (approved):
🌾 🎉

PR #4712, opened 2025-10-30
Update non-standard SEC 10-K row count check in dbt

zaneselvans:
# Overview

* With the integration of the recent final EIA 860/923 data, the row counts for `core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities` changed.
* Previously `core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities` had row counts that were different depending on whether the ETL was run locally or in the nightly builds. This was weird and disturbing, but we worked around it by defining a special test for the table that would accept row counts within a given range, which we missed when updating the row counts. This resulted in a nightly build failure.
* However, now it looks like the nightly & local ETLs result in the same number of rows. This could be a fluke, but on the off chance that whatever weirdness was resulting in different results on different platforms, I've removed the special case test, and brought the row counts for this table into the normal row-count check. We'll see what happens...

# Testing

* Checked the row counts for `core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities` in the nightly builds (5560)
* Ran the ETL locally and checked my row counts for `core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities` (5560)
* Updated the `dbt` test specifications for the table.
* Re-ran the `dbt` tests locally (and fixed my cut-and-paste error... 🤦🏼 )

## To-do list

### DIFF
--- a/dbt/models/sec10k/core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities/schema.yml
+++ b/dbt/models/sec10k/core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities/schema.yml
@@ -7,7 +7,7 @@ sources:
           - expect_columns_not_all_null
           - check_row_counts_per_partition:
               arguments:
-                table_name: out_sec10k__changelog_company_name
+                table_name: core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities
                 partition_expr:
         columns:
           - name: subsidiary_company_id_sec10k


dbt/models/sec10k/core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities/schema.yml
```diff
@@ -5,13 +5,10 @@ sources:
...
-          - dbt_expectations.expect_table_row_count_to_be_between:
-              description: Nightly builds and local ETL yield slightly different output
-              config:
-                enabled: "{{ target.name == 'etl-full'}}"
+          - check_row_counts_per_partition:
               arguments:
-                min_value: 5543
-                max_value: 5546
+                table_name: core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities
+                partition_expr:
```
zaneselvans, line 11: Maybe I'm being overly optimistic, but I'm getting the same row-counts locally and in the nightly builds now, so let's try getting rid of this weird special case. Maybe some bugfix in pandas or another library we're using made something platform-dependent platform-independent... 🤞🏼

PR #4714, opened 2025-10-31
Retry on BadZipFile with fresh BytesIO

zaneselvans:
# Overview

* We occasionally get `zipfile.BadZipFile` errors when reading out of the PUDL Datastore.
* We currently retry on this error, but we don't actually recreate the `BytesIO` object that we're trying to read from for the retry, so if that's where the problem is, it does not get fixed. Based on the md5 sums being reported and comparisons with the (good) files on disk, it seems like this may be the real source of the problems.
* This PR switches to recreating the `BytesIO` for the retries.

# Testing

* We don't really know what is causing the ephemeral badness... not sure how to test that this gets around it.

## To-do list


src/pudl/workspace/datastore.py
```diff
@@ -432,7 +432,11 @@ def get_zipfile_resource(self, dataset: str, **filters: Any) -> zipfile.ZipFile:
             f"Got resource {dataset=}, {filters=}, {md5sum=}, "
             f"{len(resource_bytes)} bytes; turning into ZipFile"
         )
-        return retry(zipfile.ZipFile, retry_on=(zipfile.BadZipFile), file=resource)
+        return retry(
+            zipfile.ZipFile,
+            retry_on=(zipfile.BadZipFile),
+            file=io.BytesIO(self.get_unique_resource(dataset, **filters)),
+        )
```
zaneselvans: I don't love that this only reports the md5 sum the first time around -- it would be nice to see that it has changed between retries to verify that this is actually attempting to operate on something different... but it does seem like retrying with exactly the same inputs every time is likely to not do anything useful.

krivard: Agree having a retry on `zipfile.ZipFile` without any updated arguments is futile.

With this edit though I'm pretty sure self.get_unique_resource will only get called once (when retry is called), and not on every attempted call of ZipFile -- so it doesn't really address the "no updated arguments" problem.

For something that actually rebuilt the arguments to ZipFile on every retry, we'd either want to wrap the call to get_zipfile_resource in `retry()` instead of calling it here, or define a local function that called both self.get_unique_resource and zipfile.ZipFile and retry that.

zaneselvans: Ahhhh, you're right.  We call `get_zipfile_resource()` in a lot of places -- basically all the dataset specific extractors that are pulling from zipfiles. So rather than scattering `retry()` all over those places maybe doing the local function here, and having it log the md5 sums every time is the better approach.

zaneselvans: Would that look like this? Or do we have to pass in a bunch of arguments because otherwise they only get instantiated on import / when the inner function is first defined?

```py
    def get_zipfile_resource(self, dataset: str, **filters: Any) -> zipfile.ZipFile:
        """Retrieves unique resource and opens it as a ZipFile."""

        def retryable() -> zipfile.ZipFile:
            resource_bytes = self.get_unique_resource(dataset, **filters)
            resource = io.BytesIO(resource_bytes)
            md5sum = hashlib.file_digest(resource, "md5").hexdigest()
            logger.info(
                f"Got resource {dataset=}, {filters=}, {md5sum=}, "
                f"{len(resource_bytes)} bytes; turning into ZipFile"
            )
            return zipfile.ZipFile(resource)

        return retry(retryable, retry_on=(zipfile.BadZipFile))
```

krivard: your sample looks exactly right -- self. get_unique_resource() will execute every time retryable() is run, which is what we want.

zaneselvans: I re-ran some assets that use this and they still seemed to work just fine (though I haven't had a BadZipFile in a while) so maybe this is ready to go?

PR #4715, opened 2025-10-31
Reduce memory for out_ferc714__hourly_estimated_state_demand by ~7GB

### LINKED ISSUE #4683: `out_ferc714__hourly_estimated_state_demand` high memory usage
### Overview

This asset is marked as `"memory-usage": "high"`.

### Success Criteria

* [x] measured initial peak memory usage:
* [x] identified memory usage hotspots
* [ ] resolved or decided to give up on memory usage hotspots

### Next steps
* [x] profile this asset

krivard:
# Overview

Closes #4683.

* out_ferc714__hourly_estimated_state_demand peaks at 12.8GB.
* With this change, it peaks at 5.7GB instead.

## What did you change?

* Converted `state_demand()` to use polars for most of the computations
* Converted one of the inputs to read as a LazyFrame
* Modified the IO Manager to typecast after reading instead of using `schema=` at scan time

## Documentation

# Testing

```
$ devtools/memory_profile.py -a out_ferc714__hourly_estimated_state_demand -d profiles/
```

## To-do list

## Discussion

I also tried converting the remaining inputs to LazyFrames, but:

* It only only squeezed another gigabyte off the peak
* I couldn't convince Polars to ignore the `geometry` column of `out_censusdp1tract__counties` at read time, and wound up creating an entire new asset `core_censusdp1tract__counties` to save as a pandas dataframe, which came with a whole lot of baggage that probably isn't worth it

### DIFF
--- a/src/pudl/analysis/state_demand.py
+++ b/src/pudl/analysis/state_demand.py
@@ -230,7 +230,7 @@ def out_ferc714__hourly_estimated_state_demand(
             return df
         return (
             df.join(
-                df.groupby(["county_id_fips", "year"]).agg(count=pl.len()),
+                df.group_by(["county_id_fips", "year"]).agg(count=pl.len()),
                 on=["county_id_fips", "year"],
             )
             .with_columns(weight=pl.col("weight") / pl.col("count"))
@@ -258,17 +258,19 @@ def out_ferc714__hourly_estimated_state_demand(
 
     def rescale_using_sales(
         df: pl.LazyFrame,
-        state_sales: pd.DataFrame | None,
+        core_eia861__yearly_sales: pd.DataFrame | None,
     ) -> pl.LazyFrame:
         """Optionally scale estimates using state sales data."""
-        if state_sales is None:
+        if core_eia861__yearly_sales is None:
             return df
         return df.join(
             # compute scale factor between current and target state totals
             df.group_by(["state_id_fips", "year"])
             .agg(pl.col("demand_mwh").sum())
             .join(
-                pl.from_pandas(state_sales).lazy(),
+                pl.from_pandas(
+                    total_state_sales_eia861(core_eia861__yearly_sales)
+                ).lazy(),
                 on=["state_id_fips", "year"],
                 suffix="_sales",
             )
@@ -298,7 +300,7 @@ def out_ferc714__hourly_estimated_state_demand(
         .pipe(distribute_demand_to_states, hourly_demand=hourly_demand)
         .pipe(
             rescale_using_sales,
-            state_sales=total_state_sales_eia861(core_eia861__yearly_sales),
+            core_eia861__yearly_sales,
         )
         # sum by state-hour to yield hourly estimates
         .group_by(["state_id_fips", "datetime_utc"])
--- a/test/unit/analysis/state_demand_test.py
+++ b/test/unit/analysis/state_demand_test.py
@@ -1,10 +1,15 @@
 """Tests for timeseries anomalies detection and imputation."""
 
+import geopandas as gpd
 import numpy as np
 import pandas as pd
+import polars as pl
 import pytest
 
-from pudl.analysis.state_demand import lookup_state
+from pudl.analysis.state_demand import (
+    lookup_state,
+    out_ferc714__hourly_estimated_state_demand,
+)
 
 AK_FIPS = {"name": "Alaska", "code": "AK", "fips": "02"}
 
@@ -34,3 +39,126 @@ AK_FIPS = {"name": "Alaska", "code": "AK", "fips": "02"}
 def test_lookup_state(state: str | int, expected: dict[str, str | int]) -> None:
     """Check that various kinds of state lookups work."""
     assert lookup_state(state) == expected
+
+
+@pytest.fixture
+def out_ferc714__hourly_estimated_state_demand_args():
+    return {
+        # two respondents with total demand 2 and average 1
+        "out_ferc714__hourly_planning_area_demand": pl.DataFrame(
+            {
+                "respondent_id_ferc714": [1, 2],
+                "demand_imputed_pudl_mwh": [1, 1],
+                # datetime doesn't matter; using 1900 for its obvious placeholder vibes
+                "datetime_utc": pd.to_datetime(["1900-01-01T12:00:00"] * 2),
+            }
+        ).lazy(),
+        # toy county, but the first two chars have to line up with a state listed in POLITICAL_SUBDIVISIONS
+        # so let's go with Alabama
+        "out_censusdp1tract__counties": gpd.GeoDataFrame(
+            {
+                "county_id_fips": ["01000"],
+                "dp0010001": [1],
+            }
+        ),
+        # assign both respondents to the same county
+        "out_ferc714__respondents_with_fips": pd.DataFrame(
+            {
+                "respondent_id_ferc714": [1, 2],
+                "county_id_fips": ["01000"] * 2,
+                "report_date": pd.to_datetime(["1900-01-01"] * 2),
+            }
+        ),
+    }
+
+
+def mock_out_ferc714__hourly_estimated_state_demand_context(
+    mocker, mean_overlaps: bool
+):
+    """Generate a mock context suitable for use in out_ferc714__hourly_estimated_state_demand."""
+    # this is as gross as it is because Dagster messes with the context between when you
+    # call out_ferc714__hourly_estimated_state_demand and when
+    # out_ferc714__hourly_estimated_state_demand actually executes.
+    # the thing out_ferc714__hourly_estimated_state_demand actually receives is
+    # context.bind(), so we make a thing such that
+    # context.bind().op_config["mean_overlaps"] has the value we want.
+    return mocker.Mock(
+        bind=mocker.Mock(
+            return_value=mocker.Mock(op_config={"mean_overlaps": mean_overlaps})
+        )
+    )
+
+
+def test_sales(mocker, out_ferc714__hourly_estimated_state_demand_args):
+    pd.testing.assert_frame_equal(
+        out_ferc714__hourly_estimated_state_demand(
+            # we need this context because we're calling the asset def directly instead of using dg.materialize().
+            # arbitrarily we choose to run with mean_overlaps off, since that's the default.
+            context=mock_out_ferc714__hourly_estimated_state_demand_context(
+                mocker, mean_overlaps=False
+            ),
+            # pass in non-None core_eia861__yearly_sales
+            core_eia861__yearly_sales=pd.DataFrame(
+                {
+                    "state": ["AL"],
+                    "report_date": pd.to_datetime(["1900-01-01"]),
+                    "sales_mwh": [10.0],
+                }
+            ),
+            **out_ferc714__hourly_estimated_state_demand_args,
+        )
+        .collect()
+        .to_pandas(),
+        pd.DataFrame(
+            {
+                "state_id_fips": ["01"],
+                "datetime_utc": pd.to_datetime(["1900-01-01T12:00:00"]),
+                "demand_mwh": [2.0],
+                # this column must be present when core_eia861__yearly_sales is non-None
+                "scaled_demand_mwh": [10.0],
+            }
+        ),
+    )
+
+
+@pytest.mark.parametrize(
+    "mean_overlaps,expected",
+    [
+        (
+            True,
+            pd.DataFrame(
+                {
+                    "state_id_fips": ["01"],
+                    "datetime_utc": pd.to_datetime(["1900-01-01T12:00:00"]),
+                    "demand_mwh": [1.0],  # use the average
+                }
+            ),
+        ),
+        (
+            False,
+            pd.DataFrame(
+                {
+                    "state_id_fips": ["01"],
+                    "datetime_utc": pd.to_datetime(["1900-01-01T12:00:00"]),
+                    "demand_mwh": [2.0],  # use the sum
+                }
+            ),
+        ),
+    ],
+)
+def test_overlaps(
+    mocker, out_ferc714__hourly_estimated_state_demand_args, mean_overlaps, expected
+):
+    """Make sure we average county overlaps when mean_overlaps is true, and sum them when mean_overlaps is false."""
+
+    pd.testing.assert_frame_equal(
+        expected,
+        out_ferc714__hourly_estimated_state_demand(
+            context=mock_out_ferc714__hourly_estimated_state_demand_context(
+                mocker, mean_overlaps=mean_overlaps
+            ),
+            **out_ferc714__hourly_estimated_state_demand_args,
+        )
+        .collect()
+        .to_pandas(),
+    )


src/pudl/helpers.py
```diff
@@ -2186,7 +2186,9 @@ def get_parquet_table_polars(table_name: str) -> pl.LazyFrame:
     paths = PudlPaths()
     parquet_path = paths.parquet_path(table_name)
 
-    return pl.scan_parquet(parquet_path, schema=schema)
+    return pl.scan_parquet(parquet_path).with_columns(
+        [pl.col(name).cast(dtype, strict=False) for name, dtype in schema.items()]
+    )
```
krivard: This resolves a weird error I was getting on inputs:

```
polars.exceptions.SchemaError: data type mismatch for column respondent_id_ferc714: incoming: Int32 != target: Int64, hint: pass cast_options=pl.ScanCastOptions(integer_cast='upcast')
```

The hint didn't fully resolve the problem, it just progressed to:

```
polars.exceptions.SchemaError: data type mismatch for column datetime_utc: incoming: Datetime('ms') != target: Datetime('μs')
```

The explicit cast works but it feels like we should be able to control the datatype in the file when we write it

jdangerx:
Compared between this version & what's on nightly builds. A few things popped up, notes below. I'll run whatever's on `main` locally and then recreate to see if there's anything strange going on there, before going farther into this rabbit hole.

### Two extra rows

Where did these come from....?

```pdb
(Pdb) local_32 = local.cast(reference.schema)
(Pdb) local_32.join(reference, on=["state_id_fips", "datetime_utc"], how="anti")
shape: (2, 4)
┌───────────────┬─────────────────────┬────────────┬───────────────────┐
│ state_id_fips ┆ datetime_utc        ┆ demand_mwh ┆ scaled_demand_mwh │
│ ---           ┆ ---                 ┆ ---        ┆ ---               │
│ str           ┆ datetime[ms]        ┆ f32        ┆ f32               │
╞═══════════════╪═════════════════════╪════════════╪═══════════════════╡
│ 17            ┆ 2021-01-01 05:00:00 ┆ 24363.0    ┆ 13744.976562      │
│ 32            ┆ 2006-01-01 07:00:00 ┆ 11.160131  ┆ 13.273124         │
└───────────────┴─────────────────────┴────────────┴───────────────────┘
```

### Slight mismatches

get the diffs:

```python
local_32 = local.cast(reference.schema)
diffs = local_32.join(
    reference,
    on=["state_id_fips", "datetime_utc"],
    how="inner").filter(pl.col("demand_mwh_right") != 0
).select(
    [
        "state_id_fips",
        "datetime_utc",
        "demand_mwh",
        "demand_mwh_right",
        "scaled_demand_mwh",
        "scaled_demand_mwh_right",
        (pl.col("demand_mwh") / pl.col("demand_mwh_right") - 1).abs().alias("demand_diff"),
        (pl.col("scaled_demand_mwh") / pl.col("scaled_demand_mwh_right") - 1).abs().alias("scaled_demand_diff")
    ]
)

diffs.filter((pl.col("scaled_demand_diff") > 1e-5) | (pl.col("demand_diff") > 1e-5)).describe()
shape: (9, 9)
┌────────────┬───────────────┬────────────────────────────┬───────────────┬───┬───────────────────┬─────────────────────────┬─────────────┬────────────────────┐
│ statistic  ┆ state_id_fips ┆ datetime_utc               ┆ demand_mwh    ┆ … ┆ scaled_demand_mwh ┆ scaled_demand_mwh_right ┆ demand_diff ┆ scaled_demand_diff │
│ ---        ┆ ---           ┆ ---                        ┆ ---           ┆   ┆ ---               ┆ ---                     ┆ ---         ┆ ---                │
│ str        ┆ str           ┆ str                        ┆ f64           ┆   ┆ f64               ┆ f64                     ┆ f64         ┆ f64                │
╞════════════╪═══════════════╪════════════════════════════╪═══════════════╪═══╪═══════════════════╪═════════════════════════╪═════════════╪════════════════════╡
│ count      ┆ 1345338       ┆ 1345338                    ┆ 1.345338e6    ┆ … ┆ 1.345338e6        ┆ 1.345338e6              ┆ 1.345338e6  ┆ 1.345338e6         │
│ null_count ┆ 0             ┆ 0                          ┆ 0.0           ┆ … ┆ 0.0               ┆ 0.0                     ┆ 0.0         ┆ 0.0                │
│ mean       ┆ null          ┆ 2014-12-19 06:51:05.902000 ┆ 16855.914062  ┆ … ┆ 15857.198242      ┆ 15857.208008            ┆ 0.302034    ┆ 0.016978           │
│ std        ┆ null          ┆ null                       ┆ 19367.777344  ┆ … ┆ 15512.105469      ┆ 15500.69043             ┆ 0.47826     ┆ 0.037156           │
│ min        ┆ 01            ┆ 2006-01-01 05:00:00        ┆ 9.342072      ┆ … ┆ 102.624901        ┆ 107.133331              ┆ 0.0         ┆ 0.0                │
│ 25%        ┆ null          ┆ 2010-05-25 04:00:00        ┆ 3111.312744   ┆ … ┆ 4612.055176       ┆ 4616.615723             ┆ 0.000197    ┆ 0.000026           │
│ 50%        ┆ null          ┆ 2015-01-07 20:00:00        ┆ 8154.200684   ┆ … ┆ 7427.603516       ┆ 7424.520508             ┆ 0.02189     ┆ 0.001862           │
│ 75%        ┆ null          ┆ 2019-06-16 01:00:00        ┆ 22089.132812  ┆ … ┆ 26694.082031      ┆ 27058.599609            ┆ 0.436358    ┆ 0.017556           │
│ max        ┆ 56            ┆ 2023-12-31 23:00:00        ┆ 106900.523438 ┆ … ┆ 92477.492188      ┆ 92475.046875            ┆ 4.689918    ┆ 5.74891            │
└────────────┴───────────────┴────────────────────────────┴───────────────┴───┴───────────────────┴─────────────────────────┴─────────────┴────────────────────┘
```

I set the closeness relative tolerance to 1e-5 to hopefully filter out any floating point conversion stuff. We can see there's ~1.3M (so ~15% of the data) where there's a surprisingly large diff between the two sides. Particularly of note is that the demand seems to have more *particularly large* diffs (see the 44% diff size at 75%ile for demand vs. 2% for scaled demand). And it's not because demand is much smaller than scaled demand either :shrug:

zaneselvans:
The demand imputation process is non-deterministic, and I think the estimated state-level demand is downstream of the imputation, so I'm not surprised to see that the two are different by more than one part in 1e5, since they'll be the product of two different imputation runs. @zschira might be able to say more about what level of variability would be reasonable.  We had to allow large proportional differences when the demand numbers were small (comparing imputed to input data).

If the scaled demand is different by a much smaller amount than the unscaled demand, that makes me wonder if the scaled demand is scaling the unimputed demand inputs.

jdangerx:
Ahh. I compared the values from `main` vs. this branch, off the same upstream values, and both of these differences went away, which supports your theory!

src/pudl/analysis/state_demand.py
```diff
@@ -189,56 +187,96 @@ def out_ferc714__hourly_estimated_state_demand(
...
+        .select(["respondent_id_ferc714", "year"])
+    )
+
     df = (
-        count_assign_ferc714
-        # Drop respondent-years with no demand
-        .merge(with_demand, on=["respondent_id_ferc714", "year"])
-        # Merge with counties and state-county assignments
-        .merge(counties, on=["county_id_fips"])
+        # Merge counties with respondent- and state-county assignments,
```
jdangerx: Kind of a nit - instead of threading `df` throughout I think it would be clearer to name these guys, though that makes the conditional flows for `mean_overlaps` and `total_sales_eia861` a little funny.

Another option would be to pull each step (assigning county populations to respondent-years, calculating population weights, maybe normalizing the overlaps, distributing the state demand, maybe scaling based on sales, aggregating up to the state level) into a separate function and punching it through polars.pipe(). That *could* be nice, here's an LLM-generated, me-proofread version of that. I'm not quite sure about how nice it really is (hard to tease out my anti-LLM bias...) but see what you think. The docstrings might be a little much, though I do find the descriptions of what each step *does* to be useful. Maybe we pull that into the top-level docstring or something.

```python
def out_ferc714__hourly_estimated_state_demand(
    context,
    out_ferc714__hourly_planning_area_demand: pl.LazyFrame,
    out_censusdp1tract__counties: gpd.GeoDataFrame,
    out_ferc714__respondents_with_fips: pd.DataFrame,
    core_eia861__yearly_sales: pd.DataFrame | None = None,
) -> pl.LazyFrame:
    """Estimate hourly electricity demand by state.

    Args:
        out_ferc714__hourly_planning_area_demand: Hourly demand timeseries, with imputed demand.
        out_censusdp1tract__counties: The county layer of the Census DP1 shapefile.
        out_ferc714__respondents_with_fips: Annual respondents with the county FIPS IDs
            for their service territories.
        core_eia861__yearly_sales: EIA 861 sales data. If provided, the predicted hourly
            demand is scaled to match these totals.

    Returns:
        LazyFrame with columns ``state_id_fips``, ``datetime_utc``, ``demand_mwh``, and
        (if ``state_totals`` was provided) ``scaled_demand_mwh``.
    """

    def prepare_county_respondent_assignments(
        county_assign_ferc714: pd.DataFrame,
        counties: pd.DataFrame,
        with_demand: pl.LazyFrame,
    ) -> pl.LazyFrame:
        """Merge counties with respondent assignments, keeping only respondent-years with demand.

        Args:
            county_assign_ferc714: County assignments for FERC 714 respondents.
            counties: County attributes including population and state.
            with_demand: Respondent-years that have nonzero demand.

        Returns:
            LazyFrame with merged county-respondent-state assignments.
        """
        return (
            pl.from_pandas(county_assign_ferc714)
            .lazy()
            .join(with_demand, on=["respondent_id_ferc714", "year"])
            .join(pl.from_pandas(counties).lazy(), on=["county_id_fips"])
        )

    def calculate_population_weights(df: pl.LazyFrame) -> pl.LazyFrame:
        """Calculate county weights based on population share within each respondent-year.

        Args:
            df: LazyFrame with county-respondent assignments and population.

        Returns:
            LazyFrame with added weight column.
        """
        respondent_population = (
            df.group_by(["respondent_id_ferc714", "year"])
            .agg(respondent_population=pl.col("population").sum())
            .select(["respondent_id_ferc714", "year", "respondent_population"])
        )
        return (
            df.join(
                respondent_population,
                on=["respondent_id_ferc714", "year"],
                how="left",
            )
            .with_columns(
                # weight counties by population fraction within each respondent-year
                # TODO: Use more county attributes in the calculation of their weights
                weight=pl.col("population") / pl.col("respondent_population"),
            )
            .drop("respondent_population")
        )

    def normalize_overlaps(df: pl.LazyFrame, mean_overlaps: bool) -> pl.LazyFrame:
        """Normalize county weights by county occurrences to handle overlaps.

        When a county is assigned to multiple respondents, this prevents double-counting
        by dividing weights by the number of respondents serving that county.

        Args:
            df: LazyFrame with weighted county-respondent assignments.
            mean_overlaps: Whether to normalize weights for overlapping counties.

        Returns:
            LazyFrame with potentially normalized weights.
        """
        if not mean_overlaps:
            return df
        return (
            df.join(
                df.groupby(["county_id_fips", "year"]).agg(count=pl.len()),
                on=["county_id_fips", "year"],
            )
            .with_columns(weight=pl.col("weight") / pl.col("count"))
            .drop("count")
        )

    def distribute_demand_to_states(
        df: pl.LazyFrame,
        hourly_demand: pl.LazyFrame,
    ) -> pl.LazyFrame:
        """Distribute respondent demand among states using computed weights.

        Args:
            df: LazyFrame with state-level weights for each respondent-year.
            hourly_demand: Hourly planning area demand by respondent.

        Returns:
            LazyFrame with hourly demand distributed to states.
        """
        return (
            df.group_by(["respondent_id_ferc714", "year", "state_id_fips"])
            .agg(pl.col("weight").sum())
            .select(["respondent_id_ferc714", "year", "state_id_fips", "weight"])
            .join(
                hourly_demand,
                on=["respondent_id_ferc714", "year"],
            )
            .with_columns(
                demand_mwh=pl.col("demand_imputed_pudl_mwh") * pl.col("weight"),
            )
            .drop("demand_imputed_pudl_mwh")
        )

    def scale_to_sales_totals(
        df: pl.LazyFrame,
        total_sales: pd.DataFrame | None,
    ) -> pl.LazyFrame:
        """Scale demand estimates to match known state sales totals from EIA 861.

        Args:
            df: LazyFrame with distributed demand by state.
            total_sales: Annual sales totals by state from EIA 861, or None.

        Returns:
            LazyFrame with scaled_demand_mwh column added if total_sales provided.
        """
        if total_sales is None:
            return df

        return df.join(
            # compute scale factor between current and target state totals
            df.group_by(["state_id_fips", "year"])
            .agg(pl.col("demand_mwh").sum())
            .join(
                pl.from_pandas(total_sales).lazy(),
                on=["state_id_fips", "year"],
                suffix="_sales",
            )
            .with_columns(scale=pl.col("demand_mwh_sales") / pl.col("demand_mwh"))
            .select(["state_id_fips", "year", "scale"]),
            on=["state_id_fips", "year"],
        ).with_columns(scaled_demand_mwh=pl.col("demand_mwh") * pl.col("scale"))

    def aggregate_to_state_hour(df: pl.LazyFrame) -> pl.LazyFrame:
        """Aggregate demand to state-hour level by summing all respondent contributions.

        Args:
            df: LazyFrame with distributed demand.

        Returns:
            LazyFrame with hourly estimates by state.
        """
        # Determine which demand fields to aggregate based on what columns exist
        schema = df.collect_schema()
        demand_fields = ["demand_mwh"]
        if "scaled_demand_mwh" in schema:
            demand_fields.append("scaled_demand_mwh")

        return (
            df.group_by(["state_id_fips", "datetime_utc"])
            .agg(*[pl.col(x).sum() for x in demand_fields])
            .select(["state_id_fips", "datetime_utc"] + demand_fields)
        )

    # Get config
    mean_overlaps = context.op_config["mean_overlaps"]

    # Prepare input data
    county_assign_ferc714 = county_assignments_ferc714(
        out_ferc714__respondents_with_fips
    )
    counties = census_counties(out_censusdp1tract__counties)
    counties["state_id_fips"] = counties["county_id_fips"].str[:2]
    total_sales_eia861 = total_state_sales_eia861(core_eia861__yearly_sales)

    # Add year column to hourly demand
    hourly_demand = out_ferc714__hourly_planning_area_demand.with_columns(
        year=pl.col("datetime_utc").dt.year()
    )

    # Identify respondent-years with nonzero demand
    with_demand = (
        hourly_demand.group_by(["respondent_id_ferc714", "year"])
        .agg(pl.col("demand_imputed_pudl_mwh").sum())
        .filter(pl.col("demand_imputed_pudl_mwh") > 0)
        .select(["respondent_id_ferc714", "year"])
    )

    # Build pipeline
    return (
        prepare_county_respondent_assignments(
            county_assign_ferc714, counties, with_demand
        )
        .pipe(calculate_population_weights)
        .pipe(normalize_overlaps, mean_overlaps=mean_overlaps)
        .pipe(distribute_demand_to_states, hourly_demand=hourly_demand)
        .pipe(scale_to_sales_totals, total_sales=total_sales_eia861)
        .pipe(aggregate_to_state_hour)
    )
```

src/pudl/analysis/state_demand.py
```diff
@@ -189,56 +187,96 @@ def out_ferc714__hourly_estimated_state_demand(
...
+        # keeping only respondent-years with nonzero demand
+        pl.from_pandas(count_assign_ferc714)
```
jdangerx: This typo from the original confused me for a sec. We'd need to update the variable assignment on line 184 too.

```suggestion
        pl.from_pandas(county_assign_ferc714)
```

src/pudl/helpers.py
jdangerx:
```diff
diff --git a/src/pudl/helpers.py b/src/pudl/helpers.py
index 847350620..2f6c2a003 100644
--- a/src/pudl/helpers.py
+++ b/src/pudl/helpers.py
@@ -2176,9 +2176,7 @@ def get_parquet_table_polars(table_name: str) -> pl.LazyFrame:
     paths = PudlPaths()
     parquet_path = paths.parquet_path(table_name)
 
-    return pl.scan_parquet(parquet_path).with_columns(
-        [pl.col(name).cast(dtype, strict=False) for name, dtype in schema.items()]
-    )
+    return pl.scan_parquet(parquet_path, schema=schema)
 
 
 def get_parquet_table(
diff --git a/src/pudl/io_managers.py b/src/pudl/io_managers.py
index d305b6cac..a9f9d32bd 100644
--- a/src/pudl/io_managers.py
+++ b/src/pudl/io_managers.py
@@ -326,8 +326,10 @@ class PudlParquetIOManager(IOManager):
                 schema=pa_schema,
             )
         elif isinstance(obj, pl.LazyFrame):
-            logger.warning("PudlParquetIOManager does not do any schema enforcement.")
-            obj.sink_parquet(
+            res = Resource.from_id(table_name)
+            schema = res.to_polars_dtypes()
+
+            obj.cast(schema).sink_parquet(
                 parquet_path,
                 engine="streaming",
                 row_group_size=100_000,
```

Should to do the trick for updating the types at write time, I tested by:

* replacing my `$PUDL_OUTPUT/parquet/out_ferc714__hourly_planning_area_demand.parquet` with the one from nightly build (with bad types)
* materializing `out_ferc714__hourly_estimated_state_demand` and seeing it fail
* stubbing the `hourly_planning_area_demand` with an asset that reads the reference nightly output and spits that immediately back out (with schema enforcement)
* materializing `hourly_planniing_area_demand` and seeing it succeed
* materializing `hourly_estimated_state_demand` and seeing it succeed

krivard:
Stats:

Command line: devtools/memory_profile.py -a out_ferc714__hourly_estimated_state_demand -d profiles/
Start time: Thu Nov 13 2025 15:14:04 GMT-0500 (Eastern Standard Time)
End time: Thu Nov 13 2025 15:14:22 GMT-0500 (Eastern Standard Time)
Duration: 0:00:18.484000
Total number of allocations: 2648015
Total number of frames seen: 0
Peak memory usage: 5.6 GB
Python allocator: pymalloc

& I double-checked CEMS doesn't blow up with the added `cast`:

Command line: devtools/memory_profile.py -a core_epacems__hourly_emissions -d profiles/
Start time: Thu Nov 13 2025 13:40:30 GMT-0500 (Eastern Standard Time)
End time: Thu Nov 13 2025 13:50:09 GMT-0500 (Eastern Standard Time)
Duration: 0:09:38.370000
Total number of allocations: 5446904
Total number of frames seen: 0
Peak memory usage: 16.0 GB
Python allocator: pymalloc

src/pudl/helpers.py
krivard: The cast-on-write approach (& your test of it) worked for me but the integration tests in ci weren't convinced:

https://github.com/catalyst-cooperative/pudl/actions/runs/19344643582/job/55341762324

I found this bit of `Field` that uses 32-bit ints/floats when in some sort of "compact" mode; is it possible we write a compact parquet to disk for the 714 hourly data? We're definitely not telling polars how to cope with that; how do we do it in pandas? /does pandas just figure it out on its own?

https://github.com/catalyst-cooperative/pudl/blob/6e2516fcbac9ad16b69c4e17cb1b371c9f97057a/src/pudl/metadata/classes.py#L647-L651

krivard: I can't find anywhere we're using `compact`, but I do think I've narrowed down why this works in testing but not in the ETL: `out_ferc714__hourly_estimated_state_demand` is materialized in pandas, not polars, so it gets written using the pyarrow schema, not the polars one.

Polars uses 64-bit ints and floats:

https://github.com/catalyst-cooperative/pudl/blob/6e2516fcbac9ad16b69c4e17cb1b371c9f97057a/src/pudl/metadata/constants.py#L14-L23

PyArrow uses 32-bit ints, floats, and dates:

https://github.com/catalyst-cooperative/pudl/blob/6e2516fcbac9ad16b69c4e17cb1b371c9f97057a/src/pudl/metadata/constants.py#L37-L46

The PyArrow schema is used when writing pandas data frames:

https://github.com/catalyst-cooperative/pudl/blob/6e2516fcbac9ad16b69c4e17cb1b371c9f97057a/src/pudl/io_managers.py#L319-L327

The Polars schema is used when scanning input assets, irrespective of how they were written:

https://github.com/catalyst-cooperative/pudl/blob/6e2516fcbac9ad16b69c4e17cb1b371c9f97057a/src/pudl/helpers.py#L2167-L2179

This means any time we're scanning into Polars an asset that was written from pandas, and that asset contains ints or numbers, Polars really should be screaming at us, because `.scan` isn't permitted to upcast unless you explicitly tell it to do so.

This is probably not what we want?

krivard:
Pytest is failing quantile constraints on unrelated tables; merging main and trying again

Issue #4696, opened 2025-10-27
Create documentation page for EIA 176


e-belfer:
### Overview

What is the problem we're solving? For very simple items, this can be encapsulated in the success criteria.
In order to make this data usable, we'll need to 

### Success Criteria

How will we know that we're done?

* [x] Write a data source page for EIA 176
* [x] Add the PDF of the survey to a `docs/data_sources/eia176` folder
* [x] Add any other high-priority instructions/documentations to the same folder

### Next steps
* [x] Download any relevant documentation / instructions relating to the 176 survey, and add them to a `docs/data_sources/eia176` folder
* [x] In `docs/templates`, create a new `eia176_child.rst.jinja` template by copying the `generic_child_template.rst.jinja`. Fill in each section as much as possible based on what you've learned transforming the data. See our other [Data Sources](https://catalystcoop-pudl.readthedocs.io/en/latest/data_sources/index.html) for guidelines.

PR #4721, opened 2025-11-03
Update extraction and transformation for `raw_eia176__data` and `_core_eia176__data` to improve legibility of data inputs

### LINKED ISSUE #4694: Update the existing transform method for `_core_eia176` assets
### Overview

Right now, the `_core_eia176` assets are pivoting a combination of ID and data unit to put each variable into one column. This produces a list of columns that is hard to decipher, and also drops crucial company data we probably want to keep alongside our tables. Before we split out our data into separate tables, we will need to fix this pivoting step. To do so, we will need to make some updates to the existing transformations in `_core_eia176__data()`.

### Success Criteria

- [x] The `_core_eia176` assets can be successfully generated, with column names that correspond to understandable variables, rather than strings and numbers
- [x] The asset check `validate_totals()` passes successfully

### Next steps
- [x] Update the `get_wide_table()` method to use the `item` column instead of the `variable_name` column created in `_core_eia176__data()`. This will pivot the field names instead of creating a composite of ID and unit that we later have to decode.
- [x] Run `pudl.helpers.simplify_columns()` on the `wide_company()` and `wide_aggregate()` tables to clean up column names
- [x] Right now, the transformation in `transform/eia176.py` drops several fields during the transformation process, losing the company name from the processed databases. Rejoin the `company` field onto the `wide_company()` table to keep the company name together with the company data.
- [x] Rename `company` to `company_name`, `id` to `report_id` and `area` to `state` to clarify their content.
- [x] Update the asset check `validate_totals()` to refer to the alternative fuel fleet boolean column instead of `3014_ct`. This asset check verifies that the total data exactly matches the company-level reports, so once this asset check passes we can drop this data and report out only the granular company data.

e-belfer:
# Overview

Closes #4694 

Right now, the _core_eia176 assets are pivoting a combination of ID and data unit to put each variable into one column. This produces a list of columns that is hard to decipher, and also drops crucial company data we probably want to keep alongside our tables. Before we split out our data into separate tables, we will need to fix this pivoting step. To do so, we will need to make some updates to the existing transformations in _core_eia176__data().

While we're at it, let's also make our column names more closely reflect our existing data standards to prevent multiple identical renames later on.

## What did you change?

-  Update the get_wide_table() method to use the item column instead of the variable_name column created in _core_eia176__data(). This will pivot the field names instead of creating a composite of ID and unit that we later have to decode.
- Run pudl.helpers.simplify_columns() on the wide_company() and wide_aggregate() tables to clean up column names.
- Right now, the transformation in transform/eia176.py drops several fields during the transformation process, losing the company name from the processed databases. Rejoin the company field onto the wide_company() table to keep the company name together with the company data.
- Update the asset check validate_totals() to refer to the alternative fuel fleet boolean column instead of 3014_ct. This asset check verifies that the total data exactly matches the company-level reports, so once this asset check passes we can drop this data and report out only the granular company data.
- In the process of renaming the column on extraction, I noticed that the current extraction code is not actually renaming the columns at all as expected, and cannot be configured to rename columns on read-in. I reverted the extraction code to map each year of data to a set of columns to more closely mirror our existing extraction code. This method is mirrored in EIA 191 and 757A extraction and needs to be addressed there (see #4722). I needed to do this in order to better name the columns on read-in, which I have also done in this PR.
- I also moved some text-based clean-up out of the process_renamed() extraction step and into `_core_eia176__data`, which is more consistent with how we handle transformations elsewhere.

# Testing

Generate the raw and core EIA 176 asset groups.

## To-do list

### DIFF
--- a/src/pudl/package_data/eia176/column_maps/data.csv
+++ b/src/pudl/package_data/eia176/column_maps/data.csv
@@ -1,4 +1,4 @@
-year_index,operator_name,operating_area,value,variable_name,form_line_numbers,operator_id_eia,line,unit_type
+year_index,operator_name,operating_state,value,variable_name,form_line_numbers,operator_id_eia,line,unit_type
 1997,company,area,value,item,itemsort,id,line,atype
 1998,company,area,value,item,itemsort,id,line,atype
 1999,company,area,value,item,itemsort,id,line,atype
--- a/src/pudl/transform/eia176.py
+++ b/src/pudl/transform/eia176.py
@@ -31,7 +31,7 @@ def _core_eia176__data(
     """
     raw_eia176__data = raw_eia176__data.astype({"report_year": int, "value": float})
 
-    aggregate_primary_key = ["report_year", "operating_area"]
+    aggregate_primary_key = ["report_year", "operating_state"]
     company_primary_key = aggregate_primary_key + ["operator_id_eia", "operator_name"]
     company_drop_columns = ["form_line_numbers", "unit_type", "line"]
     # We must drop 'id' here and cannot use as primary key because its arbitrary/duplicate in aggregate records
@@ -39,7 +39,7 @@ def _core_eia176__data(
     aggregate_drop_columns = company_drop_columns + ["operator_id_eia", "operator_name"]
 
     # Clean up text columns to ensure string matching is precise
-    for col in ["operator_name", "operating_area"]:
+    for col in ["operator_name", "operating_state"]:
         raw_eia176__data[col] = raw_eia176__data[col].str.strip().str.lower()
 
     long_company = raw_eia176__data.loc[
@@ -94,7 +94,7 @@ def validate_totals(
     """
     # First make it so we can directly compare reported aggregates to groupings of granular data
     comparable_aggregates = _core_eia176__yearly_aggregate_data.sort_values(
-        ["report_year", "operating_area"]
+        ["report_year", "operating_state"]
     ).fillna(0)
 
     # Group company data into state-level data and compare to reported totals
@@ -102,28 +102,28 @@ def validate_totals(
         _core_eia176__yearly_company_data.drop(
             columns=["operator_id_eia", "operator_name"]
         )
-        .groupby(["report_year", "operating_area"])
+        .groupby(["report_year", "operating_state"])
         .sum()
         .round(4)
         .reset_index()
     )
     aggregate_state = comparable_aggregates[
-        comparable_aggregates.operating_area != "u.s. total"
+        comparable_aggregates.operating_state != "u.s. total"
     ].reset_index(drop=True)
     # Compare using the same columns
     state_diff = aggregate_state[state_data.columns].compare(state_data)
 
     # Group calculated state-level data into US-level data and compare to reported totals
     us_data = (
-        state_data.drop(columns="operating_area")
+        state_data.drop(columns="operating_state")
         .groupby("report_year")
         .sum()
         .sort_values("report_year")
         .reset_index()
     )
     aggregate_us = (
-        comparable_aggregates[comparable_aggregates.operating_area == "u.s. total"]
-        .drop(columns="operating_area")
+        comparable_aggregates[comparable_aggregates.operating_state == "u.s. total"]
+        .drop(columns="operating_state")
         .sort_values("report_year")
         .reset_index(drop=True)
     )
--- a/test/unit/transform/eia176_test.py
+++ b/test/unit/transform/eia176_test.py
@@ -4,7 +4,7 @@ from pytest import fixture
 from pudl.transform.eia176 import _core_eia176__data, get_wide_table, validate_totals
 
 COLUMN_NAMES = [
-    "operating_area",
+    "operating_state",
     "unit_type",
     "operator_name",
     "operator_id_eia",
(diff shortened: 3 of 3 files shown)


src/pudl/package_data/eia176/column_maps/data.csv
```diff
@@ -1,2 +1,28 @@
...
+2014,company,area,value,item,itemsort,id,line,atype
+2015,company,area,value,item,itemsort,id,line,atype
+2016,company,area,value,item,itemsort,id,line,atype
+2017,company,area,value,item,itemsort,id,line,atype
+2018,company,area,value,item,itemsort,id,line,atype
+2019,company,area,value,item,itemsort,id,line,atype
+2020,company,area,value,item,itemsort,id,line,atype
+2021,company,area,value,item,itemsort,id,line,atype
+2022,company,area,value,item,itemsort,id,line,atype
+2023,company,area,value,item,itemsort,id,line,atype
```
zaneselvans, line 28: I guess adding the 2024 data is part of the archive update?

zaneselvans (approved):
Well these tables are much more legible.

If I understand correctly from looking at them now, the two tables contain the same kind of data, but one is aggregated, and the other is company level, so the aggregated data could (in theory) be reproduced based on the company level data if we needed to, and  so we won't keep the aggregated table around once we've verified that they line up in the asset check. Previously I'd thought that there was information *about* the companies in the company level data, not that it was a finer granularity, but it looks like Part 3 in #4697 has the company details.

src/pudl/package_data/eia176/column_maps/data.csv
e-belfer: 2024 data won't be published until November: https://www.eia.gov/naturalgas/ngqs/

e-belfer:
> Well these tables are much more legible.
> 
> If I understand correctly from looking at them now, the two tables contain the same kind of data, but one is aggregated, and the other is company level, so the aggregated data could (in theory) be reproduced based on the company level data if we needed to, and so we won't keep the aggregated table around once we've verified that they line up in the asset check. Previously I'd thought that there was information _about_ the companies in the company level data, not that it was a finer granularity, but it looks like Part 3 in #4697 has the company details.

Spot-on - we're basically just checking the company + adjustment fields sum to the state/U.S. level totals of the same exact data fields.

src/pudl/transform/eia176.py
```diff
@@ -29,27 +30,28 @@ def _core_eia176__data(
...
-    raw_eia176__data["variable_name"] = (
-        raw_eia176__data["line"] + "_" + raw_eia176__data["atype"]
-    )
 
-    aggregate_primary_key = ["report_year", "area"]
-    company_primary_key = aggregate_primary_key + ["id"]
-    company_drop_columns = ["itemsort", "item", "atype", "line", "company"]
+    aggregate_primary_key = ["report_year", "operating_area"]
+    company_primary_key = aggregate_primary_key + ["operator_id_eia", "operator_name"]
+    company_drop_columns = ["form_line_numbers", "unit_type", "line"]
```
zaneselvans: Any color on why we're choosing "operator" as the noun here?  In the EIA-860 there's a differentiation between "owner" and "operator" with "operator" being the assumed default, and "owner" only coming up when we need to specifically indicate that the ID being referred to is the owner (and may or may not be the operator).  But in the electricity based forms we use "utility" to identify the entity.

zaneselvans: For `operating_area` I see that there's a mix of place types. I think we have some relatively generic column names for this kind of thing elsewhere already that we probably want to conform to if possible.  Thinking of the jurisdictions table and the VCE not-just-counties place name columns.

zaneselvans: For data provenance purposes, if there's some way we can document which line numbers map to what variables that could be helpful in deciphering where the values came from on the form down the line.

I imagine we'll need to refer to the `unit_type` values when we're defining the columns in the schema (and might want to add real unit suffixes if that makes sense).

e-belfer: > Any color on why we're choosing "operator" as the noun here? In the EIA-860 there's a differentiation between "owner" and "operator" with "operator" being the assumed default, and "owner" only coming up when we need to specifically indicate that the ID being referred to is the owner (and may or may not be the operator). But in the electricity based forms we use "utility" to identify the entity.

Based on discussions and reviewing the [instructions](https://www.eia.gov/survey/form/eia_176/instructions.pdf), it seems like operator is the correct entity here.

e-belfer: > For `operating_area` I see that there's a mix of place types. I think we have some relatively generic column names for this kind of thing elsewhere already that we probably want to conform to if possible. Thinking of the jurisdictions table and the VCE not-just-counties place name columns.

Renamed to `operating_state` - only fields that aren't states are for adjustment companies, we'll need to investigate this further but in practice companies are reporting for U.S. states.

e-belfer: > For data provenance purposes, if there's some way we can document which line numbers map to what variables that could be helpful in deciphering where the values came from on the form down the line.
> 
> I imagine we'll need to refer to the `unit_type` values when we're defining the columns in the schema (and might want to add real unit suffixes if that makes sense).

As discussed elsewhere - the `unit_type` doesn't add any information not contained in the variable name, and the `line` number is not unique without the unit type - we should definitely think about how to refer to the line numbers of the original data in the #4696 issue however.

PR #4726, opened 2025-11-03
Use default spawn rather than forkserver to avoid ChildProcess errors on MacOS

### LINKED ISSUE #4662: Dagster assets failing in local ETL with `ChildProcessCrashException`
# Overview

When running the ETL locally, we frequently encounter `ChildProcessCrashException` on random assets. For example:

```
dagster._core.executor.child_process_executor.ChildProcessCrashException

  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.13/site-packages/dagster/_core/executor/multiprocess.py", line 272, in execute
    event_or_none = next(step_iter)
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.13/site-packages/dagster/_core/executor/multiprocess.py", line 408, in execute_step_out_of_process
    for ret in execute_child_process_command(multiproc_ctx, command):
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-dev/lib/python3.13/site-packages/dagster/_core/executor/child_process_executor.py", line 174, in execute_child_process_command
    raise ChildProcessCrashException(pid=process.pid, exit_code=process.exitcode)
```

You can work around it by letting the ETL finish, and then re-executing failed assets, but it's a hassle and it's not obvious why the child processes should be crashing -- this seems to happen even when we're nowhere close to running out of memory. It also means that you can't start the ETL and walk away to make lunch, and come back to it being ready to use, prolonging the whole "gimme fresh data" process unnecessarily.

zaneselvans:
# Overview

* Stop using the forkserver in Dagster because it causes `ChildProcessKilled` errors on MacOS.
* Switch to calling `configure_root_logger()` in the top-level `__init__.py` so we get logs in the Dagster UI again.
* Use [`get_dagster_logger()`](https://docs.dagster.io/api/dagster/utilities#dagster.get_dagster_logger) in our logging helpers to integrate our output with Dagster more smoothly.
* Use `logger.propagate=True` when configuring the root logger in pytest to ensure that pytest also gets the logs, even after Dagster.

Closes #4662 

# Testing

* I have been running Dagster with the forkserver config commented out locally for weeks and have run the full ETL repeatedly and had zero ChildProcess errors.
* I ran pytest at the command line & with Make and got logging output!
* I materialized some assets in the Dagster UI and also got logging output!

## To-do list

### DIFF
--- a/environments/conda-linux-64.lock.yml
+++ b/environments/conda-linux-64.lock.yml
@@ -23,7 +23,7 @@ dependencies:
   - antlr-python-runtime=4.13.2=pyhd8ed1ab_1
   - anyio=4.11.0=pyhcf101f3_0
   - appdirs=1.4.4=pyhd8ed1ab_1
-  - arelle-release=2.37.68=pyhd8ed1ab_0
+  - arelle-release=2.37.69=pyhd8ed1ab_0
   - argon2-cffi=25.1.0=pyhd8ed1ab_0
   - argon2-cffi-bindings=25.1.0=py313h07c4f96_1
   - arpack=3.9.1=nompi_hf03ea27_102
@@ -92,7 +92,7 @@ dependencies:
   - charset-normalizer=3.4.4=pyhd8ed1ab_0
   - click=8.1.8=pyh707e725_0
   - click-default-group=1.2.4=pyhd8ed1ab_1
-  - cloudpickle=3.1.1=pyhd8ed1ab_0
+  - cloudpickle=3.1.2=pyhd8ed1ab_0
   - colorama=0.4.6=pyhd8ed1ab_1
   - coloredlogs=14.0=pyhd8ed1ab_3
   - comm=0.2.3=pyhe01879c_0
@@ -119,7 +119,7 @@ dependencies:
   - dask-core=2025.10.0=pyhcf101f3_0
   - dask-expr=2.0.0=pyhd8ed1ab_0
   - databricks-sdk=0.71.0=pyhd8ed1ab_0
-  - dbt-adapters=1.17.3=pyhd8ed1ab_0
+  - dbt-adapters=1.18.0=pyhd8ed1ab_0
   - dbt-common=1.35.0=pyhd8ed1ab_0
   - dbt-core=1.10.13=pyhd8ed1ab_0
   - dbt-duckdb=1.9.4=pyhd8ed1ab_0
@@ -149,9 +149,9 @@ dependencies:
   - exceptiongroup=1.3.0=pyhd8ed1ab_0
   - execnet=2.1.1=pyhd8ed1ab_1
   - executing=2.2.1=pyhd8ed1ab_0
-  - fastapi=0.120.4=h9378a72_0
+  - fastapi=0.121.0=h67827d6_0
   - fastapi-cli=0.0.13=pyhcf101f3_0
-  - fastapi-core=0.120.4=pyhcf101f3_0
+  - fastapi-core=0.121.0=pyhcf101f3_0
   - filelock=3.20.0=pyhd8ed1ab_0
   - flask=3.1.2=pyhd8ed1ab_0
   - flask-cors=6.0.1=pyhe01879c_0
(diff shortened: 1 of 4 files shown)


jdangerx:
Did you try making all the loggers in `logging_helpers` use [get_dagster_logger](https://docs.dagster.io/api/dagster/utilities#dagster.get_dagster_logger) instead? That seems like it would get around the context-passing issue...

zaneselvans:
Ooh, no I had not seen that. Let me give it a try.

src/pudl/analysis/record_linkage/name_cleaner.py
```diff
@@ -2,15 +2,16 @@
...
 from importlib.resources import files
 from typing import Self
 
 import pandas as pd
 from pydantic import BaseModel, Field, model_validator
 
-logger = logging.getLogger(__name__)
+from pudl.logging_helpers import get_logger
+
+logger = get_logger(__name__)
```
zaneselvans, line 14: A lingering non-PUDL configured logger.

src/pudl/__init__.py
```diff
@@ -18,6 +20,8 @@
     workspace,
 )
 
+configure_root_logger()
+
```
zaneselvans, line 24: Need to do this here now, since we don't have the special dagster config hook place to do it.

src/pudl/helpers.py
```diff
@@ -2048,17 +2043,12 @@ def get_dagster_execution_config(
...
-    if "forkserver" in multiprocessing.get_all_start_methods():
-        start_method_config = {"forkserver": {"preload_modules": ["pudl.init_logging"]}}
-
     return {
         "execution": {
             "config": {
                 "multiprocess": {
                     "max_concurrent": num_workers,
                     "tag_concurrency_limits": tag_concurrency_limits,
-                    "start_method": start_method_config,
```
zaneselvans, line 2061: Revert to using the default spawn rather than forkserver, since it's b0rken on MacOS.

src/pudl/logging_helpers.py
```diff
@@ -3,33 +3,39 @@
...
+    return get_dagster_logger(f"catalystcoop.{name}")
 
 
 def configure_root_logger(
     logfile: str | None = None,
     loglevel: str = "INFO",
-    dependency_loglevels: dict[str, int] = {"numba": logging.WARNING},
-):
+    dependency_loglevels: dict[str, int] | None = None,
+    propagate: bool = False,
```
zaneselvans, line 18: Add `propagate` as a new parameter so we can use it in the pytest configuration.

src/pudl/logging_helpers.py
```diff
@@ -3,33 +3,39 @@
...
 def get_logger(name: str):
     """Helper function to append 'catalystcoop' to logger name and return logger."""
-    return logging.getLogger(f"catalystcoop.{name}")
+    return get_dagster_logger(f"catalystcoop.{name}")
 
 
 def configure_root_logger(
     logfile: str | None = None,
     loglevel: str = "INFO",
-    dependency_loglevels: dict[str, int] = {"numba": logging.WARNING},
```
zaneselvans, line 16: Move mutable default value into the function body.

src/pudl/logging_helpers.py
```diff
@@ -3,33 +3,39 @@
...
             This allows us to filter excessive logs from dependencies.
+        propagate: Whether to propagate logs to ancestor loggers. Useful for ensuring
+            that pytest has access to PUDL logs during testing.
     """
+    if dependency_loglevels is None:
+        dependency_loglevels = {"numba": logging.WARNING}
     # Explicitly set log-level for dependency loggers
     [
-        logging.getLogger(dependency_name).setLevel(dependency_loglevel)
+        get_dagster_logger(dependency_name).setLevel(dependency_loglevel)
```
zaneselvans, line 34: Presumably we want to use the dagster logger here too, to shush these guys?

zaneselvans:
`get_dagster_logger()` + `logger.propagate = True` seems to have done the trick.

Not sure what RTD's deal is with the python environment. c'mon man

src/pudl/analysis/record_linkage/name_cleaner.py
jdangerx: Gaspo! Good catch.

src/pudl/logging_helpers.py
jdangerx: I think that makes sense!

Guessing that get_dagster_logger gets the existing logger configured for the dependency and registers it with Dagster, so either `logging.getLogger` or `get_dagster_logger` will both return something that we can set the loglevel of. I suspect `get_dagster_logger` will also allow us to pull logs from the dependencies into Dagster in the unlikely event we choose to do so.

jdangerx (approved):
:christmas_tree: :carpentry_saw:

PR #4728, opened 2025-11-04
Create core_eia176__yearly_gas_disposition_by_consumer

### LINKED ISSUE #4709: Create `core_eia176__yearly_gas_disposition_by_consumer` (Part 6A)
### Overview

Part 6A: Natural and Supplemental Gas Disposition for the Report State (Lines 10.0-11.6) of the EIA 176 form contains valuable information about an operator's gas deliveries to end-use consumers. Let's bring this data into PUDL!

### Success Criteria

There is no one size fits all in table transforms, though we document our approach to cleaning data [here](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_guidelines.html#examples-of-acceptable-changes).

The new table should meet the following criteria:
* a well defined and unique primary key: one or more columns that naturally explain what a row in the table corresponds to. For this table, this _should_ be one row per year, ID and consumer type. You'll need to check that the combination of these fields is unique, and in the case of duplications, propose a strategy for creating a unique primary key.
* all categorical columns should be constrained (e.g., the state column shouldn't contain "New York City" in it).
* nulls should be standardized
* units should be standardized
* data types should be defined and enforced
* metadata should be added to the database schema for each new field and for the table itself
* data should be validated in DBT

### Prerequisites
- #4694 

### Additional investigation required
- Right now, there's a free-text field specified for "other" kinds of consumers. I haven't been able to find this data in the [NGQs](https://www.eia.gov/naturalgas/ngqs/#?year1=2020&year2=2023&company=Name) reports. If possible, can you locate it anywhere?

### Transform the data
- [ ] Make a copy of `devtools/inspect-assets.ipynb`. Use the notebook to load the `_core_eia176__yearly_company_data` asset and inspect it. This asset contains all years of the table extracted and concatenated together.
- [ ] Isolate the fields relevant to this table (lines 10.0-11.6). You can use the [survey form](https://www.eia.gov/survey/form/eia_176/form.pdf) and the `form_line_numbers` field in the `raw_eia176__data` table to identify them.
- [ ] In the notebook, explore the table and transform it. Here are some docs re [standard pudl transforms](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_guidelines.html#examples-of-acceptable-changes).
    - [ ] Make sure there are no non-standard nulls (e.g., "-"). If there are, replace them with `pd.NA`.
    - [ ] Pull the consumer type out of the column names and into its own categorical column to create a narrow dataframe. We should wind up with one consumer type column (e.g., residential), and one column each per data value (e.g., number of customers, revenue).
    - [ ] Check for primary keys of the table and drop if there are **clear** errors (ex: duplicates on expected PKs but with clear bad records aka one decent looking record and one with mostly nulls).
    - [ ] Handle unexpected dtypes: e.g., no numbers in the state column
    - [ ] Convert the `state` column to well-behaved state subdivision codes (e.g., Alaska to AK). Do this by mapping the existing state names against `core_pudl__codes_subdivisions`, cleaning up strings as is needed. We'll need to decide how to handle non-state values (e.g., Mexico, other) reported for the correction rows.
    - [ ] Clean up column names as needed to create succinct but informative variable names.

#### Bring the tables into PUDL's ETL
- [ ] Once you're satisfied with the table, it's time to bring it into dagster.
   - [ ] In `pudl/src/pudl/transform/eia176.py` add a new `core` asset function based on [asset naming conventions](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/naming_conventions.html#asset-naming-conventions). In this case: `core_eia176__yearly_gas_disposition_by_consumer`
   - [ ] add the `@asset()` decorator to this function.
   - [ ] Migrate your transforms for the notebook into this function.
   - [ ] [Reload the code location in dagster](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/dev_dagster.html#reloading-code-locations), search for your new table in the "Asset" tab and materialize it. See [our docs on Dagster](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/dev_dagster.html) for support.
- [ ] Add metadata for this table to add this table to the database schema
   - [ ] define the table  in `pudl/src/metadata/resources/eia176.py`
       - [ ] define the fields (aka column names)
       - [ ] add primary keys if possible
       - [ ] add `"field_namespace": "eia",`
       - [ ] add `"sources": ["eia176"],`
       - [ ] add `"etl_group": "eia176",`
       - [ ] consult with a Catalyst developer about how to update table documentation
    - [ ] Go add any new fields into `pudl/src/pudl/metadata/fields.py`
       - [ ] check that each field doesn't already exist.
       - [ ] add a description for each field.
       - [ ] add the data type
       - [ ] add constraints (enums/categories or patterns) if relevant.
       - [ ] add units
- [ ] Write the table to the PUDL database
   - [ ] Add the io manager to the asset() decorator: `@asset(io_manager_key="pudl_io_manager")`
   - [ ] [update alembic schema](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_the_etl.html#database-schema-migration)
- [ ] Materialize your table via the [dagster ui](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_the_etl.html#running-the-etl-via-the-dagster-ui)
- [ ] Add a table schema in [dbt](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_validation_reference.html#adding-new-tables) and add [expected row counts](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_validation_quickstart.html#row-counts)
- [ ] Add any other DBT tests that seem sensible (e.g., volume can't be negative).

MeadBarrel:
# Overview

Closes #4709.

Create core_eia176__yearly_gas_disposition_by_consumer

## What did you change?

- [x] Make a copy of devtools/inspect-assets.ipynb. Use the notebook to load the _core_eia176__yearly_company_data asset and inspect it. This asset contains all years of the table extracted and concatenated together.
- [x] Isolate the fields relevant to this table (lines 10.0-11.6). You can use the [survey form](https://www.eia.gov/survey/form/eia_176/form.pdf) and the form_line_numbers field in the raw_eia176__data table to identify them.
- [x] In the notebook, explore the table and transform it. Here are some docs re [standard pudl transforms](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_guidelines.html#examples-of-acceptable-changes).
  - [x] Make sure there are no non-standard nulls (e.g., "-"). If there are, replace them with pd.NA.
  - [x] Pull the consumer type out of the column names and into its own categorical column to create a narrow dataframe. We should wind up with one consumer type column (e.g., residential), and one column each per data value (e.g., number of customers, revenue).
  - [x] Check for primary keys of the table and drop if there are clear errors (ex: duplicates on expected PKs but with clear bad records aka one decent looking record and one with mostly nulls). 
  
    *Primary key is `report_year`, `operating_state`, and `customer_type`*
  - [x] Handle unexpected dtypes: e.g., no numbers in the state column
  - [x] Convert the state column to well-behaved state subdivision codes (e.g., Alaska to AK). Do this by mapping the existing state names against core_pudl__codes_subdivisions, cleaning up strings as is needed. We'll need to decide how to handle non-state values (e.g., Mexico, other) reported for the correction rows.  
  
    *There are three operating states in `_core_eia176__yearly_company_data` that aren't covered by `core_pudl__codes_subdivisions`: "fed. gulf of mexico", "other", and "mexico". For now i left them as is.*
  - [x] Clean up column names as needed to create succinct but informative variable names.
- [x] Once you're satisfied with the table, it's time to bring it into dagster.
  - [x] In pudl/src/pudl/transform/eia176.py add a new core asset function based on [asset naming conventions](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/naming_conventions.html#asset-naming-conventions). In this case: core_eia176__yearly_gas_disposition_by_consumer
  - [x] add the @asset() decorator to this function.
  - [x] Migrate your transforms for the notebook into this function.
  - [x] [Reload the code location in dagster](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/dev_dagster.html#reloading-code-locations), search for your new table in the "Asset" tab and materialize it. See [our docs on Dagster](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/dev_dagster.html) for support.
- [x] Add metadata for this table to add this table to the database schema
  - [x] define the table in pudl/src/metadata/resources/eia176.py
    - [x] define the fields (aka column names)
    - [x] add primary keys if possible
    - [x] add "field_namespace": "eia",
    - [x] add "sources": ["eia176"],
    - [x] add "etl_group": "eia176",
    - [x] consult with a Catalyst developer about how to update table documentation    
  - [ ] Go add any new fields into pudl/src/pudl/metadata/fields.py
    - [x] check that each field doesn't already exist.
    - [x] add a description for each field.
    - [x] add the data type
    - [x] add constraints (enums/categories or patterns) if relevant.
    - [x] add units
- [x] Write the table to the PUDL database
  - [x] Add the io manager to the asset() decorator: @asset(io_manager_key="pudl_io_manager")
  - [x] [update alembic schema](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_the_etl.html#database-schema-migration)
- [x] Materialize your table via the [dagster ui](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_the_etl.html#running-the-etl-via-the-dagster-ui)
- [x] Add a table schema in [dbt](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_validation_reference.html#adding-new-tables) and add [expected row counts](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_validation_quickstart.html#row-counts)
- [x] Add any other DBT tests that seem sensible (e.g., volume can't be negative).

## Documentation

# Testing

Reviewed the data and ran `make ci` locally

## To-do list

- [x] Run `make ci` locally to ensure that the merge queue will accept your PR.

### DIFF
--- a/src/pudl/transform/eia176.py
+++ b/src/pudl/transform/eia176.py
@@ -313,8 +313,8 @@ def core_eia176__yearly_gas_disposition_by_consumer(
         )
         for customer_class in customer_classes
     )
-    assert mismatched == 28, (
-        f"{mismatched} mismatched volume totals found, expected 28."
+    assert mismatched <= 28, (
+        f"{mismatched} mismatched volume totals found, expected no more than 28."
     )
 
     df = _core_eia176__yearly_company_data.filter(primary_key + keep)


src/pudl/metadata/resources/eia176.py
```diff
@@ -0,0 +1,32 @@
...
+
+from typing import Any
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "core_eia176__yearly_gas_disposition_by_consumer": {
+        "description": {
+            "additional_summary_text": (
+                "Information about a company's gas deliveries to end-use consumers from eia176"
+            )
+        },
```
e-belfer, line 12:
```suggestion
        "description": {
            "additional_summary_text": (
                "a company's natural gas deliveries to end-use consumers within the report state."
            ),
            "additional_source_text": "(Part 6, Lines 10.0-11.6)",
        },
```

We recently changed how we construct our metadata and don't have up to date documentation on this, sorry! If you want to preview the metadata for this table as you're compiling it, you can run:
`resource_description -n core_eia176__yearly_gas_disposition_by_consumer`.

If you have any notes on this table's unusual behaviors, or anything else you want to add, it can get noted under the `additional_details_text` key.

src/pudl/metadata/resources/eia176.py
```diff
@@ -0,0 +1,32 @@
...
+                "transport_consumers",
+                "transport_revenue",
+                "transport_volume",
+                "volume",
+            ],
+            "primary_key": ["report_year", "operator_id_eia", "consumer_type"],
+        },
+        "field_namespace": "eia",
+        "sources": ["eia176"],
+        "etl_group": "eia176",
```
e-belfer, line 33: In `pudl.metadata.classes.py`, you'll need to add `eia176` to the `Literal` constraint for the `etl_group` field.

MeadBarrel:
> * I think validating that the `volume` column is the sum of the other two fields and then dropping it seems completely reasonable, especially since it's not actually in the survey itself.

There are 28 records for 6 operators where the sum doesn't match. But only if null values are assumed to be 0, otherwise if we filter out instances where one or the other value is null (say `residential_sales_volume`=5000, `residential_transport_volume`=null, `residential_volume`=5765), the rest of the sums match.

> * The `CUSTOMER_CLASSES` enum in `metadata/enums.py` originates from EIA 861 data, and we likely should keep these two separate. I'd suggest creating a new `CUSTOMER_CLASSES_EIA176` enum with the valid values and using that to constrain the field.

Ok, so i create an override in FIELD_METADATA_BY_RESOURCE right?

What about `operator_id_eia`? Right now they look like this "17600032KS" - it's a number with a state abbreviation added. I take it EIA id should be an integer. They seem to all follow this pattern.

For "fed. gulf of mexico" the suffix is FX.
For "other" it's OO
For "mexico" it's MX

I take it i need to extract the numerical part and use `operator_utility_id_eia" as field name, defined in fields.py as
```
    "operator_utility_id_eia": {
        "type": "integer",
        "description": (
            "The EIA utility Identification number for the operator utility."
        ),
    },
```

MeadBarrel:
> * I _believe_ that all the non-normative states in this table (mexico, fed. gulf of mexico and other) are reporting all zero values. If this is the case, we should assert this and then drop these records, which aren't reporting real data and are in reality adjustment fields for other records.

Yeah you're correct they're all zeroes

src/pudl/transform/eia176.py
```diff
@@ -152,3 +155,173 @@ def validate_totals(
...
+        temp_df, id_vars=primary_key, var_name="metric", value_name="value"
+    ).reset_index()
+    temp_df[["customer_class", "metric_type"]] = temp_df["metric"].str.extract(
+        r"(residential|commercial|industrial|electric_power|vehicle_fuel|other)_(.+)$"
+    )
+    temp_df = temp_df.pivot(
+        index=primary_key + ["customer_class"], columns="metric_type", values="value"
+    ).reset_index()
+
+    temp_df = temp_df[
```
e-belfer: Can you add a docstring here to explain that you're checking the volume totals against the sub-components?

src/pudl/metadata/resources/eia176.py
```diff
@@ -6,24 +6,28 @@
...
-            )
+                "a company's natural gas deliveries to end-use consumers within the report state."
+            ),
+            "additional_source_text": "(Part 6, Lines 10.0-11.6)",
         },
         "schema": {
             "fields": [
                 "report_year",
-                "operator_id_eia",
+                "operator_utility_id_eia",
```
e-belfer: This should be reverted to `operator_id_eia` and defined as a separate field - this means something different and more generic than an operator utility.

src/pudl/metadata/resources/eia176.py
```diff
@@ -6,24 +6,28 @@
...
+                "customer_class",
+                "revenue_class",
+                "consumers",
+                "revenue",
+                "volume_mfc",
+            ],
+            "primary_key": [
+                "report_year",
+                "operator_utility_id_eia",
+                "operating_state",
```
e-belfer: See my comments below - I don't think we want to split state out of the ID, and so the operating state shouldn't be a part of this table's PK.

src/pudl/transform/eia176.py
```diff
@@ -241,20 +274,54 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+    primary_key.extend(["customer_class", "revenue_class"])
     df = df.pivot(
         index=primary_key, columns="metric_type", values="value"
     ).reset_index()
     df.columns.name = None
 
+    # Exctract numeric part from operator_id_eia and rename the column
+    df["operator_id_eia"] = df["operator_id_eia"].str[:-2].astype(int)
+
+    # Rename columns as needed
```
e-belfer, line 348: See note above, these are actually two distinctive things and I don't think we want to perform this rename.

src/pudl/transform/eia176.py
```diff
@@ -241,20 +274,54 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
     df = df.drop(columns="metric").reset_index()
-    primary_key.append("customer_type")
+    primary_key.extend(["customer_class", "revenue_class"])
     df = df.pivot(
         index=primary_key, columns="metric_type", values="value"
     ).reset_index()
     df.columns.name = None
 
+    # Exctract numeric part from operator_id_eia and rename the column
+    df["operator_id_eia"] = df["operator_id_eia"].str[:-2].astype(int)
```
e-belfer: This was my instinct as well initially, but the ID as assigned by EIA 176 has the state in it, and this is consistent over time.

Looking at the [company list](https://www.eia.gov/naturalgas/ngqs/#?report=RP6&year1=2023&year2=2023&company=Name), companies with the same name don't have one consistent ID with multiple states, but rather operations in each state appear to have a unique numeric ID with a state prefix.

Given that this terrible ID system exists, we shouldn't strip the state from the ID, as the numeric part by itself has no explanatory power in the EIA ecosystem.

(If you have different information though or use the numeric portion of the ID in a way I'm missing, would love to hear it!)

src/pudl/transform/eia176.py
```diff
@@ -241,20 +274,54 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+    df = df.dropna(subset=["operating_state"])
+    df = df.dropna(subset=["consumers", "revenue", "volume_mfc"], how="all")
+
+    # Ensure data types are homogenous
+
+    # Convert consumers to integers - can't have half a consumer
+    assert cast(pd.Series, df["consumers"].dropna() % 1 == 0).all(), (
+        "Consumer values are all expected to be valid integers"
+    )
+    df["consumers"] = pd.to_numeric(df["consumers"], errors="coerce").astype("Int64")
```
e-belfer: We'll cast these fields to their appropriate datatypes using the dtypes defined in the `fields.py` file when we add the IO manager in dagster, so there's no need to do this manually here.

e-belfer (changes requested):
A few requests re: field changes, and a note that there's no need to individually cast or assert data types - this all gets handled by our IO managers.

Once these changes are integrated, you should be clear to move onto the last few steps!

zaneselvans:
@MeadBarrel for the documentation:

* We'll need to add a bullet under the New Data section of the release notes highlighting the new tables and (finally!) deeper integration of EIA natural gas data into PUDL. You can find examples of what this looks like further down in the release notes New Data sections for previous releases.
* We'll need to create a new data source documentation page for the `eia176`, the first step of which is making a copy of `docs/templates/generic_child_template.rst.jinja` at `docs/templates/eia176_child.rst.jinja`, but it's probably easier if we do this part since the way the data source docs are built out of the metadata is currently under-documented. If you have any notes on weirdness in the data we might want to bring them into the errata section.
* We may want to add some more table-level information to the description in the resource metadata, but we can probably handle that too.

zaneselvans:
The RTD build failure should go away if you merge `main` into your branch -- there's a package version in our conda lockfile that got yanked.

zaneselvans:
@MeadBarrel I made a [`eia176-data-docs` branch](https://github.com/catalyst-cooperative/pudl/tree/eia176-data-docs) of of your PR branch and created an initial EIA-176 data source documentation page using our template. Feel free to merge it into your PR, or I can figure out how to make a PR into your fork if you want to review it that way.

zaneselvans:
I'm going to rebase the eia176 docs off of main and we can merge it in right after your PR goes in.

src/pudl/metadata/fields.py
```diff
@@ -59,14 +59,15 @@
 # )
 
 FIELD_METADATA: dict[str, dict[str, Any]] = {
-    "consumers": {"type": "integer", "description": "number of consumers"},
-    "revenue": {
-        "type": "number",
-        "description": "revenue",
+    "consumers": {"type": "integer", "description": "Number of consumers"},
```
```suggestion
    "consumers": {"type": "integer", "description": "Number of end-use consumers within the report state."},
```

src/pudl/metadata/fields.py
```diff
@@ -59,14 +59,15 @@
...
+    "operator_id_eia": {
+        "type": "string",
+        "description": "Operator EIA id",
```
```suggestion
        "description": "The unique EIA identifier for an operator in a given state. The last two letters of the ID indicate the state.",
```

src/pudl/metadata/fields.py
```diff
@@ -59,14 +59,15 @@
...
     },
-    "volume_mfc": {
+    "volume_mcf": {
         "type": "number",
-        "description": "volume",
+        "description": "Total volume of the deliveries",
+        "unit": "McF",
```
```suggestion
        "description": "Total volume of natural gas deliveries in the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit.",
        "unit": "Mcf",
```

src/pudl/metadata/fields.py
```diff
@@ -8522,7 +8523,16 @@
 
 FIELD_METADATA_BY_RESOURCE: dict[str, dict[str, Any]] = {
     "core_eia176__yearly_gas_disposition_by_consumer": {
+        "operating_state": {
+            "description": "State that the distribution utility is reporting for.",
```
```suggestion
            "description": "State that the operator is reporting for.",
```

src/pudl/metadata/fields.py
```diff
@@ -8522,7 +8523,16 @@
...
+        },
+        "revenue": {
+            "description": "Revenue including taxes",
```
```suggestion
            "description": "Revenue including taxes, rounded to the nearest whole dollar.",
```

src/pudl/metadata/fields.py
```diff
@@ -8522,7 +8523,16 @@
...
+        },
         "customer_class": {
+            "description": (
+                f"High level categorization of customer type: {CUSTOMER_CLASSES_EIA176}"
```
```suggestion
                f"High level categorization of customer type (e.g., commercial, residential)."
```

I know this is how the other field is currently defined, but we don't typically do this for fields and I think it'll be more straightforward this way.

src/pudl/transform/eia176.py
```diff
@@ -291,37 +324,22 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
         .sum()
         == 0
     )
     df = df.dropna(subset=["operating_state"])
-    df = df.dropna(subset=["consumers", "revenue", "volume_mfc"], how="all")
-
-    # Ensure data types are homogenous
+    df = df.dropna(subset=["consumers", "revenue", "volume_mcf"], how="all")
 
     # Convert consumers to integers - can't have half a consumer
```
e-belfer: Is this comment left over from something you are no longer doing?

migrations/versions/0f8abb125d4b_add_core_eia176__yearly_gas_disposition_.py
```diff
@@ -0,0 +1,38 @@
+"""Add core_eia176__yearly_gas_disposition_by_consumer
+
+Revision ID: ad781ad93cc8
+Revises: d629674ba279
```
e-belfer: A new migration has been merged, so you'll need to revise this to get the ordering correct.

```suggestion
Revises: e91f15227f60
```

Because I'm asking you to update the field definitions, you'll want to either 1) make the revisions I suggest here, delete your local database and run `alembic upgrade head`, then handle any changes you make in a second new migration or 2) run `alembic downgrade head-1`, delete this migration entirely, and create a new migration with updated fields. Either should work fine!

migrations/versions/0f8abb125d4b_add_core_eia176__yearly_gas_disposition_.py
```diff
@@ -0,0 +1,38 @@
...
+Create Date: 2025-11-07 09:12:26.552827
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'ad781ad93cc8'
+down_revision = 'd629674ba279'
```
e-belfer:
```suggestion
down_revision = 'e91f15227f60'
```
Same here!

src/pudl/transform/eia176.py
```diff
@@ -152,3 +155,191 @@ def validate_totals(
...
+        .sum()
+        == 0
+    )
+    df = df.dropna(subset=["operating_state"])
+    df = df.dropna(subset=["consumers", "revenue", "volume_mcf"], how="all")
+
+    # Convert consumers to integers - can't have half a consumer
+
+    df["customer_class"] = df["customer_class"].astype("category")
+    df["revenue_class"] = df["revenue_class"].astype("category")
```
e-belfer: The ENUM constraint in `fields.py` will handle this type conversion automatically, so there's no need to call `astype()` here.

src/pudl/transform/eia176.py
```diff
@@ -152,3 +155,191 @@ def validate_totals(
...
+    temp_df = _core_eia176__yearly_company_data.filter(primary_key + keep + volumes)
+    temp_df = pd.melt(
+        temp_df, id_vars=primary_key, var_name="metric", value_name="value"
+    ).reset_index()
+    temp_df[["customer_class", "metric_type"]] = temp_df["metric"].str.extract(
+        r"(residential|commercial|industrial|electric_power|vehicle_fuel|other)_(.+)$"
+    )
+    temp_df = temp_df.pivot(
+        index=primary_key + ["customer_class"], columns="metric_type", values="value"
+    ).reset_index()
```
e-belfer: Non-blocking: Right now you're performing the same pivot/extraction step twice - once on the subsetted DF and once on the entire dataframe. I'd propose either writing a small function to do this and calling it twice, or moving the volume check until after the dataframe has been pivoted.

src/pudl/transform/eia176.py
```diff
@@ -152,3 +155,191 @@ def validate_totals(
...
+    # Assert that volume totals match the sum of sub-components
+    temp_df = temp_df[
+        temp_df["sales_volume"].notna()
+        & temp_df["transport_volume"].notna()
+        & temp_df["volume"].notna()
+    ]
+    assert cast(
+        pd.Series,
+        (temp_df["sales_volume"] + temp_df["transport_volume"]) == temp_df["volume"],
+    ).all(), "Volumes don't match"
```
e-belfer:
```suggestion
    temp_df = temp_df[
        (temp_df["sales_volume"].notna() | temp_df["transport_volume"].notna())
        & (temp_df["volume"].notna())
    ]
    
    (mismatched_series := pd.Series((df["revenue"] + df["consumers"]) != df["volume_mcf"]).sum()) <= 28, f"{mismatched_series} mismatched volume totals found, expected 28."
```

> There are 28 records for 6 operators where the sum doesn't match. But only if null values are assumed to be 0, otherwise if we filter out instances where one or the other value is null (say residential_sales_volume=5000, residential_transport_volume=null, residential_volume=5765), the rest of the sums match.

I think I would want to know whether this number was drastically changing over time, because 28 seems reasonable but a big growth would cause me to reconsider dropping this column. I've proposed a slightly more specific assertion above, which looks for changes in the counts of mismatched columns.

dbt/models/eia176/core_eia176__yearly_gas_disposition_by_consumer/schema.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: core_eia176__yearly_gas_disposition_by_consumer
+
+        data_tests:
+          - expect_columns_not_all_null
+          - check_row_counts_per_partition:
+              arguments:
+                table_name: core_eia176__yearly_gas_disposition_by_consumer
+                partition_expr: "report_year"
+          - dbt_utils.expression_is_true:
+              expression: "(volume_mcf is null or volume_mcf >= 0) and (consumers is null or consumers >=0)"
```
e-belfer: These are all good things to test, but it might be helpful to split these into two separate expressions so we can more easily tell whether the failure is in the volume or the consumer column if something goes wrong.

dbt/models/eia176/core_eia176__yearly_gas_disposition_by_consumer/schema.yml
```diff
@@ -0,0 +1,24 @@
...
+
+        columns:
+          - name: report_year
+          - name: operator_id_eia
+          - name: operating_state
+          - name: customer_class
+          - name: revenue_class
+          - name: consumers
+          - name: revenue
+          - name: volume_mcf
```
e-belfer, line 30:
```suggestion
        columns:
          - name: report_year
          - name: operator_id_eia
          - name: operating_state
          - name: customer_class
          - name: revenue_class
          - name: consumers
            data_tests:
            - dbt_expectations.expect_column_min_to_be_between:
                arguments:
                  min_value: 0.0
          - name: revenue
            data_tests:
            - dbt_expectations.expect_column_min_to_be_between:
                arguments:
                  min_value: 0.0
          - name: volume_mcf
            data_tests:
            - dbt_expectations.expect_column_min_to_be_between:
                arguments:
                  min_value: 0.0
```

This could replace the dbt expression_is_true call that you've written above, and achieve the same check.

src/pudl/transform/eia176.py
MeadBarrel: > or moving the volume check until after the dataframe has been pivoted.

Can't do that because the total volume information is lost after we categorize by revenue_class. And i personally think it *needs* to be done and it makes sense, though maybe add a separate asset with just the total volume?

e-belfer: Got it! In that case, I think this seems fine, and there's no need to add another asset into the mix.

e-belfer:
@MeadBarrel There's now #4696 tracking the documentation changes separately, so I think we can leave these out of this PR. Sorry about the confusion there!

src/pudl/metadata/fields.py
```diff
@@ -1082,11 +1215,15 @@
...
-        "description": "The most recently updated effective date on which the generator is scheduled to start operation",
+        "description": (
+            "The most recently updated effective date on which the generator is scheduled to start operation"
+        ),
     },
     "customer_class": {
         "type": "string",
-        "description": f"High level categorization of customer type: {CUSTOMER_CLASSES}.",
+        "description": (
+            f"High level categorization of customer type: {CUSTOMER_CLASSES}."
```
e-belfer: As noted in my last review, we don't typically list valid categories in our field descriptions, so I think we'd want to update this to be something like 
```suggestion
            f"High level categorization of customer type (e.g., commercial, residential)."
```

src/pudl/metadata/fields.py
```diff
@@ -7411,6 +8530,30 @@
...
+                f"High level categorization of customer type: {CUSTOMER_CLASSES_EIA176}"
+            ),
+            "type": "string",
+            "constraints": {
+                "enum": CUSTOMER_CLASSES_EIA176,
+            },
+        },
+        "revenue_class": {
+            "type": "string",
+            "description": f"Source of revenue: {REVENUE_CLASSES_EIA176}",
```
e-belfer: As noted in my last review, we don't typically list valid categories in our field descriptions, so I think we'd want to update this to be something like
```suggestion
            "description": f"Source of revenue: whether revenue originates from gas owned directly by the operator (sales) or gas transported by the operator (transport).",
```

PR #4730, opened 2025-11-04
Small devtools updates: memory_profile + compare_against_nightly

jdangerx:
# Overview

* `memory_profile` wasn't tracking native allocations, which leads to big empty spaces on the flamegraph when e.g. Polars decides to allocate 20GB of memory outside of pyalloc.
* I wanted to know which assets were actually getting materialized, which gives me something to do while waiting for the run to start.
* I wanted to do some quick regression tests of EIA 930 core tables that are downstream of the raw work I was doing.

## What did you change?
* Fixed those problems.

## Documentation

# Testing

Used the tools and found them satisfactory.

### DIFF
--- a/devtools/check_against_nightly.py
+++ b/devtools/check_against_nightly.py
@@ -8,7 +8,17 @@ Usage:
 
 Example usage:
 
+Check against S3 nightly:
+
     ./check_against_nightly.py core_eia930__*.parquet
+
+Check against local copy:
+
+    ./check_against_nightly.py --reference-dir $PUDL_OUTPUT/../nightly/parquet core_eia930__*.parquet
+
+Check against GCS build outputs (requires gcloud application default credentials):
+
+    ./check_against_nightly.py --reference-dir gs://builds.catalyst.coop/2025-10-11-0605-a531a9a7f-main/parquet core_eia930__*.parquet
 """
 
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


devtools/check_against_nightly.py
```diff
@@ -0,0 +1,46 @@
...
+    """Compare local Parquet files against nightly."""
+    local_paths = [Path(target) for target in local_targets]
+    click.echo(f"Comparing {', '.join(p.stem for p in local_paths)} against nightly...")
+    for local_path in local_paths:
+        click.echo(f"Comparing {local_path.stem}...", nl=False)
+        local_lf = pl.scan_parquet(local_path)
+        nightly_base = "https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly"
+        remote_lf = pl.scan_parquet(f"{nightly_base}/{local_path.name}")
+        try:
+            assert_frame_equal(local_lf, remote_lf)
```
zschira: With CEMS I had to filter to one year at a time, then call `assert_frame_equal` to avoid blowing up memory, so it may be nice to add the ability to do this check in chunks. We could potentially add a CLI arg to set a groupby column, then do the check on individual groups. I'd also be fine with merging this as is, then adding that functionality when it becomes necessary, so I'll make this non-blocking.

jdangerx: Let's merge as is, and when we run into memory issues we can either work around it ad-hoc or add functionality to the tool!

Issue #3571, opened 2024-04-15
`raw_*__all_dfs` dictionary output is memory-inefficient


zaneselvans:
### Overview

For many of our CSV and Excel based datasets, we read many partitions of many "pages", concatenate the page fragments together, and produce a dictionary of dataframes as an interim asset that gets pickled and written to disk.

For the following datasets, we then read in the whole dictionary in downstream assets to pick the different keys of the dictionary apart into separate assets:

- `eia930`
- `vcerare`
- `nrelatb`
- `eia757a`
- `eia191`
- `eia176`
- `censuspep`

We believe this is memory-inefficient, though we should, of course, profile the assets first to check.

If this is, indeed, a large memory issue (particularly for `raw_eia930__balance`, `raw_eia930__interchange`, and `raw_eia930__subregion` assets), we should fix it. The fix for eia930 should incidentally decrease memory usage elsewhere.

Since the downstream assets appear to literally just read the value from the dictionary and return it, this can hopefully be ameliorated by eliminating that extra layer of serialization/deserialization and turning the `all_dfs` into a multi-asset instead.

This will also likely benefit from any IO Manager work done for #4666 .

### Success criteria
* `raw_eia930__(balance|interchange|subregion)` assets have low memory usage

krivard:
Initial profile for `raw_eia930__balance`:

```
Command line: devtools/memory_profile.py -a raw_eia930__balance -d profiles/
Start time: Wed Oct 29 2025 13:47:59 GMT-0400 (Eastern Daylight Time)
End time: Wed Oct 29 2025 13:48:25 GMT-0400 (Eastern Daylight Time)
Duration: 0:00:26.559000
Total number of allocations: 3032360
Total number of frames seen: 0
Peak memory usage: 10.3 GB
Python allocator: pymalloc
```

6.5GB of that is loading the input pickle, so yeah, let's try a multi-asset

krivard:
hmm, due to the use of DynamicOut and the general hairy-ness of computing the partition selection, I couldn't figure out how to do this without infringing on #4666 -- we should consider merging the two.

Things I tried:

- Use a dictionary return from raw_dfs to encode a graph multi asset -- can't unpack the concat_pages result to do this inside the graph node
- Use a dictionary return from concat_pages to encode a graph multi asset -- dict ends up in the wrong place and dagster can't detect it at graph build time
[...]

PR #4731, opened 2025-11-04
Reduce EIA 930 extraction memory usage

### LINKED ISSUE #4666: `pudl.extract.extractor.raw_df_factory` high memory usage
# Overview

Currently these assets all use the `raw_df_factory`, which uses the `concat_pages` op which is marked as high memory usage.

- raw_censuspep__all_dfs
- raw_eia861__all_dfs
- raw_nrelatb__all_dfs
- raw_phmsagas__all_dfs
- raw_eia757a__all_dfs
- raw_eia191__all_dfs
- raw_eia860__all_dfs
- raw_eia923__all_dfs
- raw_vcerare__all_dfs
- raw_eia930__all_dfs
- raw_eia176__all_dfs

We hypothesize that there are two memory issues here:
* spitting out Pandas DataFrames in `partition_extractor`, which are then eagerly loaded into memory when we pass the partitions into `concat_pages`
* `concat_pages` eagerly creating a new DataFrame in memory from all the partitions before immediately dumping to disk

We should profile these assets and verify where the memory hotspots are. Ideally we could fix things once in `raw_df_factory` and deal with all of these at once.

To fix things, we will likely need to create a new IO manager that: 
* serializes LazyFrames to disk (probably in the `ipc` format, not `parquet`, since many of our raw tables don't have a stable enough schema to be serialized as Parquet)
* reads LazyFrames from disk

It would be best to not mess with our existing IO managers until this new one has proved stable, to avoid destabilizing too much of the ETL at once.

## Success Criteria

* we've measured the initial peak memory usage of above assets
* the peak memory usage of the above assets is reduced to < 2 GB each

jdangerx:
# Overview

Closes #4666 #3571.

30GB peak memory usage -> ~12GB
8 min runtime -> <1 min

The raw EIA 930 extraction had two major memory issues:

1) Due to the `raw_df_factory` structure, we were first making a giant dictionary of dataframes called `raw_eia930__all_dfs`, then loading the whole dictionary each time we wanted to extract a value from it (`raw_eia930__balance|interchange|subregion`). This lead to 3 6.5GB allocations unpickle the giant dictionary.
2) We read all the individual CSVs into memory, and then concatenated them together in-memory. This lead to `raw_eia930__all_dfs` using 30GB of memory on my machine.

## What did you change?

* stopped using the `raw_df_factory` and the `GenericExtractor` for EIA 930 extraction; used a duckdb-based extraction pattern instead. If we want to apply the pattern in other extractors we can, but if they don't have huge memory pressure maybe we don't need to. In general, decoupling the easy/simple CSV extraction in EIA 930 from all the complication in the Excel extractors will make both easier to work with, and that's a hill I am happy to die on.

* **made the raw EIA930 assets write out to Parquet.** They're not in the datapackage so they won't show up in the viewer, but they *will* show up on S3 unless we decide to clean them up... should we clean them up or are we OK to put raw data on S3?

## Documentation

# Testing

See #4730 - compared the `core_eia930__*` assets against nightly. Got one mismatch, which I looked into:
```
Comparing core_eia930__hourly_interchange, core_eia930__hourly_net_generation_by_energy_source, core_eia930__hourly_operations, core_eia930__hourly_subregion_demand against nightly...
Comparing core_eia930__hourly_interchange...OK
Comparing core_eia930__hourly_net_generation_by_energy_source...OK
Comparing core_eia930__hourly_operations...OK
Comparing core_eia930__hourly_subregion_demand...MISMATCH
DataFrames are different (height (row count) mismatch)
[left]: 5159280
[right]: 4792684
```

Also compared the `raw_eia930__balance|interchange|subregion` dataframes generated locally before these changes against the post-change ones and saw the same row-count difference in the subregion table.

**This turns out to be because of a bug in the original EIA 930 extraction where we were skipping the 2018h2 subregion data** - investigation below. **If I filter to only 2019+ data the *raw* data frames are equal; there's a little messiness with the *core* datasets on 2019-01-01 due to us stripping out the timezone information, but if we filter to 2019-01-02+ data the *core* datasets are also equal.**

```python
>>> import polars as pl                                                                                                                                                                                   
>>> from polars.testing import assert_frame_equal                                                                                                                                                         
>>> obs = pl.scan_parquet("/home/daz/pudl_output/nightly/parquet/raw_eia930__subregion.parquet.parquet")                                                                                                           # ignore the funny path... some typos I had to iron out
>>> exp = pl.scan_parquet("/home/daz/scratch/raw_eia930__all_dfs/subregion.parquet")                               
>>> assert_frame_equal(obs, exp, check_column_order=False, check_dtypes=False) # columns out of order and new version has ints instead of floats
>>> exp.describe()
shape: (9, 8)
┌────────────┬─────────────────────────────┬────────────────────────────┬────────────────────────────┬────────────────────────────┬─────────────────────┬────────────────────────────┬───────────────────┐
│ statistic  ┆ balancing_authority_code_ei ┆ balancing_authority_subreg ┆ datetime_local             ┆ datetime_utc               ┆ demand_reported_mwh ┆ report_date                ┆ report_hour_local │
│ ---        ┆ a                           ┆ ion_…                      ┆ ---                        ┆ ---                        ┆ ---                 ┆ ---                        ┆ ---               │
│ str        ┆ ---                         ┆ ---                        ┆ str                        ┆ str                        ┆ f64                 ┆ str                        ┆ f64               │
│            ┆ str                         ┆ str                        ┆                            ┆                            ┆                     ┆                            ┆                   │
╞════════════╪═════════════════════════════╪════════════════════════════╪════════════════════════════╪════════════════════════════╪═════════════════════╪════════════════════════════╪═══════════════════╡
│ count      ┆ 4792684                     ┆ 4792684                    ┆ 4792684                    ┆ 4792684                    ┆ 4.699296e6          ┆ 4792684                    ┆ 4.792684e6        │
│ null_count ┆ 0                           ┆ 0                          ┆ 0                          ┆ 0                          ┆ 93388.0             ┆ 0                          ┆ 0.0               │
│ mean       ┆ null                        ┆ null                       ┆ 2022-04-18 00:37:35.639971 ┆ 2022-04-18 05:45:10.854043 ┆ 4203.934726         ┆ 2022-04-17 12:07:35.410120 ┆ 12.499922         │
│ std        ┆ null                        ┆ null                       ┆ null                       ┆ null                       ┆ 851733.956932       ┆ null                       ┆ 6.922248          │
│ min        ┆ CISO                        ┆ 0001                       ┆ 2019-01-01 01:00:00        ┆ 2019-01-01 06:00:00        ┆ -1144.0             ┆ 2019-01-01 00:00:00        ┆ 1.0               │
│ 25%        ┆ null                        ┆ null                       ┆ 2020-08-24 13:00:00        ┆ 2020-08-24 18:00:00        ┆ 797.0               ┆ 2020-08-24 00:00:00        ┆ 6.0               │
│ 50%        ┆ null                        ┆ null                       ┆ 2022-04-18 01:00:00        ┆ 2022-04-18 06:00:00        ┆ 1777.0              ┆ 2022-04-18 00:00:00        ┆ 12.0              │
│ 75%        ┆ null                        ┆ null                       ┆ 2023-12-10 12:00:00        ┆ 2023-12-10 18:00:00        ┆ 4467.0              ┆ 2023-12-10 00:00:00        ┆ 18.0              │
│ max        ┆ SWPP                        ┆ ZONK                       ┆ 2025-08-03 00:00:00        ┆ 2025-08-03 07:00:00        ┆ 1.5266e9            ┆ 2025-08-02 00:00:00        ┆ 25.0              │
└────────────┴─────────────────────────────┴────────────────────────────┴────────────────────────────┴────────────────────────────┴─────────────────────┴────────────────────────────┴───────────────────┘
>>> obs.describe()
shape: (9, 8)
┌────────────┬─────────────────────────────┬────────────────────────────┬───────────────────┬────────────────────────────┬─────────────────────┬────────────────────────────┬────────────────────────────┐
│ statistic  ┆ balancing_authority_code_ei ┆ report_date                ┆ report_hour_local ┆ balancing_authority_subreg ┆ demand_reported_mwh ┆ datetime_local             ┆ datetime_utc               │
│ ---        ┆ a                           ┆ ---                        ┆ ---               ┆ ion_…                      ┆ ---                 ┆ ---                        ┆ ---                        │
│ str        ┆ ---                         ┆ str                        ┆ f64               ┆ ---                        ┆ f64                 ┆ str                        ┆ str                        │
│            ┆ str                         ┆                            ┆                   ┆ str                        ┆                     ┆                            ┆                            │
╞════════════╪═════════════════════════════╪════════════════════════════╪═══════════════════╪════════════════════════════╪═════════════════════╪════════════════════════════╪════════════════════════════╡
│ count      ┆ 5159280                     ┆ 5159280                    ┆ 5.15928e6         ┆ 5159280                    ┆ 4.96665e6           ┆ 5159280                    ┆ 5159280                    │
│ null_count ┆ 0                           ┆ 0                          ┆ 0.0               ┆ 0                          ┆ 192630.0            ┆ 0                          ┆ 0                          │
│ mean       ┆ null                        ┆ 2022-01-15 11:43:07.639205 ┆ 12.500092         ┆ null                       ┆ 4167.397151         ┆ 2022-01-16 00:13:07.307066 ┆ 2022-01-16 05:20:36.144578 │
│ std        ┆ null                        ┆ null                       ┆ 6.922347          ┆ null                       ┆ 828493.240544       ┆ null                       ┆ null                       │
│ min        ┆ CISO                        ┆ 2018-07-01                 ┆ 1.0               ┆ 0001                       ┆ -1144.0             ┆ 2018-07-01 01:00:00        ┆ 2018-07-01 05:00:00        │
│ 25%        ┆ null                        ┆ 2020-04-08                 ┆ 7.0               ┆ null                       ┆ 787.0               ┆ 2020-04-08 13:00:00        ┆ 2020-04-08 17:00:00        │
│ 50%        ┆ null                        ┆ 2022-01-15                 ┆ 13.0              ┆ null                       ┆ 1770.0              ┆ 2022-01-16 00:00:00        ┆ 2022-01-16 05:00:00        │
│ 75%        ┆ null                        ┆ 2023-10-25                 ┆ 18.0              ┆ null                       ┆ 4448.0              ┆ 2023-10-25 12:00:00        ┆ 2023-10-25 17:00:00        │
│ max        ┆ SWPP                        ┆ 2025-08-02                 ┆ 25.0              ┆ ZONK                       ┆ 1.5266e9            ┆ 2025-08-03 00:00:00        ┆ 2025-08-03 07:00:00        │
└────────────┴─────────────────────────────┴────────────────────────────┴───────────────────┴────────────────────────────┴─────────────────────┴────────────────────────────┴────────────────────────────┘
>>> h2_2018 = obs.filter(pl.col("report_date") < datetime(2019, 1, 1))
>>> h2_2018.collect().height
366596
>>> exp.collect().height
4792684
>>> exp.collect().height + h2_2018.collect().height
5159280
>>> obs.collect().height
5159280
>>> obs_post_2019 = obs.filter(pl.col("report_date") >= datetime(2019, 1, 1))
>>> assert_frame_equal(obs_post_2019.select(exp.columns), exp, check_dtypes=False)
                                                                
```

## To-do list

### DIFF
--- a/src/pudl/extract/eia930.py
+++ b/src/pudl/extract/eia930.py
@@ -60,7 +60,11 @@ def raw_eia930__subregion(context) -> Path:
     )
 
 
-def extract_page(datastore: Datastore, page: str, half_years: list[str]) -> Path:
+def extract_page(
+    datastore: Datastore,
+    page: str,
+    half_years: list[str],
+) -> Path:
     """Pull data for a page across many half-years into a Parquet file.
 
     This involves reading each half-year, of course, but also concatenating them
@@ -96,7 +100,8 @@ def extract_page(datastore: Datastore, page: str, half_years: list[str]) -> Path
         for view_name in individual_views
     )
     output_path = PudlPaths().parquet_path(f"raw_eia930__{page}")
-    con.query(union_query).to_parquet(str(output_path))
+    all_partitions = con.query(union_query)
+    all_partitions.to_parquet(str(output_path))
     return output_path
 
 
@@ -132,10 +137,19 @@ def extract_half_year_page(
     def clean_name(name):
         return re.sub(r"\W+", "_", name.lower()).strip("_")
 
-    selects = ", ".join(
+    cleaned_column_names = {clean_name(col_name) for col_name in csv_rel.columns}
+    assert cleaned_column_names == set(column_map.keys())
+
+    select_existing_columns = [
         f'"{col_name}" as "{column_map[clean_name(col_name)]}"'
         for col_name in csv_rel.columns
-    )
+    ]
+
+    expected_null_cols = set(metadata.get_all_columns(page)) - set(column_map.values())
+    select_null_columns = [f"NULL as {col_name}" for col_name in expected_null_cols]
+
     view_name = f"v_{clean_name(filename)}"
-    csv_rel.select(selects).to_view(view_name)
+    csv_rel.select(", ".join(select_existing_columns + select_null_columns)).to_view(
+        view_name
+    )
     return view_name
--- a/devtools/check_against_nightly.py
+++ b/devtools/check_against_nightly.py
@@ -50,7 +50,7 @@ def cli(local_targets: tuple[Path], pdb: bool, reference_dir):
         local = pl.scan_parquet(local_path)
         reference = pl.scan_parquet(f"{reference_dir}/{local_path.name}")
         try:
-            assert_frame_equal(local, reference)
+            assert_frame_equal(local, reference, check_column_order=False)
         except AssertionError as e:
             if pdb:
                 breakpoint()
--- a/src/pudl/extract/extractor.py
+++ b/src/pudl/extract/extractor.py
@@ -107,12 +107,16 @@ class GenericMetadata:
         """Returns list of all pudl columns for a given page across all partitions."""
         return sorted(self._column_map[page].columns)
 
-    def get_column_map(self, page, **partition) -> dict:
-        """Return dictionary of original columns to renamed columns for renaming in a given partition and page."""
+    def get_column_map(self, page, **partition) -> dict[str, str]:
+        """Return dictionary of original columns to renamed columns for renaming in a given partition and page.
+
+        Columns that don't exist in this partition/page will show up as pd.nan, so we need to filter those out.
+        """
         return {
             v: k
             for k, v in self._column_map[page]
             .loc[str(self._get_partition_selection(partition))]
+            .dropna()
             .to_dict()
             .items()
             if v != -1


src/pudl/extract/eia930.py
```diff
@@ -1,97 +1,145 @@
...
-        name=f"raw_eia930__{page}",
-        op_tags={"memory-use": "high"},
-        compute_kind="pandas",
+@asset(required_resource_keys={"datastore", "dataset_settings"})
+def raw_eia930__balance(context) -> Path:
+    """Raw balance page."""
+    return extract_page(
+        datastore=context.resources.datastore,
+        page="balance",
+        half_years=context.resources.dataset_settings.eia.eia930.half_years,
```
krivard, line 35: Ohhhhhh

yes it would make sense that the partition selection for a single source is significantly more straightforward than PartitionSelection makes it look, since it has to handle all sources.

⭐

src/pudl/extract/eia930.py
```diff
@@ -1,97 +1,145 @@
...
+    half_years_parsed = ((int(year), int(half)) for year, half in half_years_split)
+    valid_half_years = [
+        f"{year}half{half}"
+        for year, half in half_years_parsed
+        if (year > 2018) or (year == 2018 and half == 2)
+    ]
+    return extract_page(
+        datastore=context.resources.datastore,
+        page="subregion",
+        half_years=valid_half_years,
```
krivard: if you prefer not to rely on string-sorting hocus pocus then keep the parsed version, but this totally works:

```suggestion
    return extract_page(
        datastore=context.resources.datastore,
        page="subregion",
        half_years=[h for h in context.resources.dataset_settings.eia.eia930.half_years if h>="2018half2"],
```

src/pudl/extract/eia930.py
```diff
@@ -1,97 +1,145 @@
...
+
+def extract_page(datastore: Datastore, page: str, half_years: list[str]) -> Path:
+    """Pull data for a page across many half-years into a Parquet file.
+
+    This involves reading each half-year, of course, but also concatenating them
+    together and expanding the schema to fit all the columns we see.
+
+    If we were to return the `con.query()` and use an IOManager to manage
+    the Parquet IO, we would have to manage the DuckDB connection lifetime
+    so we don't try to write out from a closed DuckDB connection. So, we just
```
krivard: (less garden path)
```suggestion
    to avoid trying to write out from a closed DuckDB connection. So, we just
```

src/pudl/extract/eia930.py
```diff
@@ -1,97 +1,145 @@
...
+        extract_half_year_page(
+            con,
+            datastore=datastore,
+            half_year=half_year,
+            page=page,
+        )
+        for half_year in half_years
+    ]
+    union_query = " UNION ALL BY NAME ".join(
+        f"SELECT * FROM {view_name}"  # noqa: S608 (we trust this view name)
```
krivard, line 99: ⭐ 
that's gorgeous

src/pudl/extract/eia930.py
```diff
@@ -1,97 +1,145 @@
...
+    filename = f"eia930-{half_year}-{page}.csv"
+    metadata = GenericMetadata(dataset_name)
+    column_map = metadata.get_column_map(page=page, half_year=half_year)
+
+    with datastore.get_zipfile_resource(dataset_name, half_year=half_year) as zf:
+        unzipped = zf.extract(filename, datastore.temporary_extraction_dir.name)
+    csv_rel = con.read_csv(unzipped)
+
+    def clean_name(name):
+        return re.sub(r"\W+", "_", name.lower()).strip("_")
```
krivard, line 138: I found https://github.com/catalyst-cooperative/pudl/blob/39b4c71925ad2febb8ffb25181edaccd168683b2/src/pudl/helpers.py#L1100

is that useful here?

src/pudl/transform/eia930.py
```diff
@@ -19,7 +21,7 @@
     compute_kind="pandas",
 )
 def core_eia930__hourly_operations_assets(
-    raw_eia930__balance: pd.DataFrame,
+    raw_eia930__balance: Path,
```
krivard, line 24: This will do in a pinch, but I think in an ideal world an IOManager would handle reading from that path, so that `raw_eia930__*` assets are data frames just like the other `raw_*__*` assets.

krivard: It doesn't show up very many times, and the whole file is <500 lines, so it won't be a big maintainability threat if it stays. But I still squint my eyes at it menacingly.

src/pudl/workspace/datastore.py
```diff
@@ -329,6 +330,10 @@ def __init__(
         """
         self._cache = resource_cache.LayeredCache()
         self._datapackage_descriptors: dict[str, DatapackageDescriptor] = {}
+        # If you want to extract a file to *disk* instead of memory, it helps
+        # to have a temporary directory that sticks around until the Datastore
+        # object is deleted
+        self.temporary_extraction_dir = TemporaryDirectory()
```
krivard, line 336: this is good

krivard (approved):
Optional nits and enhancements.

This change is so small and has such a good payoff that I hesitate to suggest bloating it by adding a new IOManager to handle duckdb Path outputs, but it may be good to track as a followup, especially if we use that pattern again.

src/pudl/extract/eia930.py
jdangerx: Yeah, I thought about that, and it felt a little brittle but it is enticingly short.... I guess the parsing is brittle too. It's clear and documented, let's go for it.

src/pudl/transform/eia930.py
jdangerx: I wanted to use `ParquetIOManager` directly but it assumes the table corresponds to a resource for which we've defined a schema. Which means non-raw assets, unfortunately.

Re: adding a DuckDBIOManager, maybe we do this the next time we duckdb-ify something. We'd need to pass back the `DuckDBPyConnection` as well as the `DuckDBPyRelation`, so we probably want to bundle those together in some sort of `DuckDBRelationOutput` class or something. Or maybe `con` gets closed on function return regardless of whether we pass the connection back, which would necessitate a `duckdb` resource that holds onto a connection for the duration of an asset materialization, sort of like how we shoved `temporary_extraction_directory` into the `Datastore`. Might be cleaner to do it that way regardless tbh.

Anyways. Agree that it seems sus and also like a good follow-up!

src/pudl/extract/eia930.py
jdangerx: It's specialized for pandas dataframes unfortunately. `DuckDBPyRelation.columns` is just a normal dict and doesn't have the `.str` methods on it.

zaneselvans:
@jdangerx  I think we should not distribute the raw parquet files. That will break a bunch of patterns we've established with the other data we distribute. They won't have any documentation, and we don't really want anybody using them and will take up space and potentially create confusion if someone lists the bucket contents. They'll also get distributed to Zenodo and Kaggle. It would be pretty easy to just avoid anything that starts with `raw_` or we could create a `raw_parqet` directory in `$PUDL_OUTPUT` that holds these intermediary files that's ignored for distribution (as is the case with the per-quarter parquet outputs that are concatenated together for EPA CEMS in their own subdirectory now). Or keep them alongside all the other undistributed assets with a raw Parquet IO Manager that doesn't need the database schema infrastructure.

jdangerx:
OK, how does this sound for raw file behavior?

* write raw files to `$PUDL_OUTPUT/parquet/`
* *don't* zip the `raw_*` files up into the `pudl_parquet.zip` or distribute them in `/parquet`
* *do* keep them in the `builds.catalyst.coop` buckets

Separately, tweak the `check_against_nightly.py` introduced in #4730 to accept different locations for the reference table: at least a local directory, but if it's not too hard to do something like `--gcs branch_name` that might be a nice ergonomic improvement.

Should be easy enough to do by removing the `raw_*` files in `clean_up_outputs_for_distribution` since that happens after we save to GCS...

zaneselvans:
* I'm agnostic as to where the raw files get written, so long as we are careful not to distribute them.
* I think the more (files) the merrier in `build.catalyst.coop` since they can help us debug build weirdness and isn't publicly visible.
* Allowing a `UPath` style URL for the directory in which to find the reference file, with `s3://pudl.catalyst.coop/nightly/` as the default seems like it would be easy and potentially very useful, since it could be used to point at local outputs with `file://...` and also to specific build outputs with `gs://builds.catalyst.coop/...`

PR #4737, opened 2025-11-05
Extract remaining report data for EIA 176

### LINKED ISSUE #4710: Extract the free-form and categorical data from EIA 176 raw tables
Once https://github.com/catalyst-cooperative/pudl-archiver/issues/909 merges, we should extract the tables containing the data we're currently missing in order to get the required free-text and categorical fields.

### Extract the data
Follow the [instructions](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/existing_data_updates.html#map-the-structure-of-the-new-data) to extract the following datasets into new tables:

- [ ] [RP4](https://www.eia.gov/naturalgas/ngqs/#?report=RP4&year1=2023&year2=2023&company=Name) (Types of Operation and Sector Items) data should get extracted into a `raw_eia176__operation_types_and_sector_items` table
- [ ] [RP5](https://www.eia.gov/naturalgas/ngqs/#?report=RP5&year1=2023&year2=2023&company=Name) (Continuation Text Lines) data should get extracted into a `raw_eia176__continuation_text_lines` table

We should also rename `raw_eia176__data` to `raw_eia176__numeric_data` now that it isn't the only one.

As we dig deeper into each table of data, we may find that other reports need to be extracted because they are more reliable, contain missing fields, etc. For now, these are the tables that are definitely in scope.

e-belfer:
# Overview

Closes #4710 

- Add missing variables for Form EIA 176 by adding in multiple other forms
- Rename the `raw_eia176__data` asset to `raw_eia176__custom` to better distinguish from other data sources

## What did you change?
- Add column maps for other datasets
- Update csv extractor to search for particular filename
- Extract these datasets
- Update the extracted dataset names in the transform and test modules
- Update the `csv_test.py` module to use EIA 191 for now, as EIA 176 has more than one page and a new more complex format that isn't handled by `GenericMetadata()` anymore

# Testing

Generate all raw_eia176 and core_eia176 assets

## To-do list

- [x] Once #https://github.com/catalyst-cooperative/pudl-archiver/pull/926 merges, switch to production DOI

### DIFF
--- a/test/integration/etl_test.py
+++ b/test/integration/etl_test.py
@@ -45,7 +45,7 @@ class TestCsvExtractor:
     def test_extract_eia176(self, pudl_datastore_fixture):
         """Spot check extraction of eia176 csv files."""
         extractor = pudl.extract.eia176.Extractor(pudl_datastore_fixture)
-        page = "data"
+        page = "custom"
         year = 2018
         if "company" not in extractor.load_source(page=page, year=year).columns:
             raise AssertionError(f"page {page} not found in datastore for {year}")


src/pudl/package_data/eia176/column_maps/company_list.csv
```diff
@@ -0,0 +1,28 @@
+year_index,operating_state,operator_id_eia,operator_name,operator_status,report_year
+1997,report_state,id,company_name,-1,last_update
+1998,report_state,id,company_name,-1,last_update
+1999,report_state,id,company_name,-1,last_update
+2000,-1,-1,-1,-1,-1
+2001,-1,-1,-1,-1,-1
```
zaneselvans, line 6: What's up with 2000, 2001, and 2019?

src/pudl/extract/eia176.py
```diff
@@ -19,24 +19,79 @@ def __init__(self, *args, **kwargs):
...
 
 @asset
-def raw_eia176__data(raw_eia176__all_dfs):
+def raw_eia176__custom(raw_eia176__all_dfs):
+    """Extract raw EIA company data from CSV sheets into dataframes.
+
+    Returns:
+        An extracted EIA 176 dataframe.
+    """
+    return Output(value=raw_eia176__all_dfs["custom"])
```
zaneselvans: Is there a more specific and informative name that we could use for this asset in lieu of `custom`? That makes sense in the context of the archiving process and needing to create a custom report, but it's kind of mysterious over here, in contrast to the other asset names which are more descriptive.

Do we need to use the `Output()` class for the returns in any of these? They're all simple one table outputs, with names that correspond to the asset definition, so I think we can just add a `pd.DataFrame` type hint for the return value (so Dagster knows what type to expect) and return the `DataFrame`.

```suggestion
def raw_eia176__custom(raw_eia176__all_dfs) -> pd.DataFrame:
    """Extract raw EIA company data from CSV sheets into dataframes.

    Returns:
        An extracted EIA 176 dataframe.
    """
    return raw_eia176__all_dfs["custom"]
```

src/pudl/extract/eia176.py
```diff
@@ -19,24 +19,79 @@ def __init__(self, *args, **kwargs):
...
+        Returns:
+            string name of the CSV file
+        """
+        partition_selection = self._metadata._get_partition_selection(partition)
+        # Company list doesn't exist in certain years
+        if page == "company_list" and (
+            int(partition_selection[0:4]) in [2000, 2001, 2019]
+        ):
+            return "-1"
+        return f"{self._dataset_name}_{partition_selection}_{page}.csv"
```
zaneselvans, line 47: Can you explain this special handling a little in the docstring? Is the lack of these files in some years the thing that necessitates the override for `source_filename()`?  Normally this would produce the value that's stored in the `file_map.csv` wouldn't it? But we don't have a `file_map.csv` for this data source.  Or is that only in the Excel extractors?

zaneselvans (approved):
Some non-blocking requests for clarification / suggestions.

src/pudl/extract/eia176.py
e-belfer: It's to spare us from having to make a file_map csv since we've programmatically created the CSV file names and expect them to behave nicely.

src/pudl/extract/eia176.py
e-belfer: Does `numeric_data` seem more informative?

src/pudl/extract/eia176.py
zaneselvans: Added notes explaining this special casel.

src/pudl/extract/eia176.py
zaneselvans: Renamed this and replaced the individual asset definitions with a factory that will create assets for each raw data frame.

src/pudl/package_data/eia176/column_maps/company_list.csv
zaneselvans: Missing. See below.

zaneselvans:
The extract and transform are working, but I'm getting an error on the asset check that verifies the totals. When I check out the commit from before I made any changes, it is also failing in the same way, so I think this is a change in the expectations around the data, not something I did. So I will investigate and update the asset check.

zaneselvans:
* There appear to be real cases in which the aggregated company level data and the reported state-level aggregations are not equal.
* For each column it's just a few dozen records (except for the known `alternative_fuel_fleet_1_yes_0_no_reported` case), but they are not always the same few dozen records.
* All the discrepancies are in 2013.
* It looks like all of the discrepancies are due to the reported values being zero, but the aggregated values being non-zero.

I'm using this little function to grab the records that don't match

```python
def state_discrepancies(
    state_data: pd.DataFrame,
    aggregate_state: pd.DataFrame,
    state_diff: pd.DataFrame,
    col: str,
) -> pd.DataFrame:
    merge_cols = ["report_year", "operating_state"]
    unmatched_idx = state_diff.loc[state_diff[(col, "self")].notna()].index

    return pd.merge(
        state_data.loc[unmatched_idx, merge_cols + [col]],
        aggregate_state.loc[unmatched_idx, merge_cols + [col]],
        on=["report_year", "operating_state"],
        suffixes=("_reported", "_aggregated"),
    )

for col in state_diff.columns.levels[0]:
    display(
        state_discrepancies(
            state_data=state_data,
            aggregate_state=aggregate_state,
            state_diff=state_diff,
            col=col,
        )
    )
```

I'm going to switch the asset check to being non-blocking for the moment.

@MeadBarrel Would you rather we merge the new extraction in, or hold off until we can debug this discrepancy on Monday?

e-belfer:
> * There appear to be real cases in which the aggregated company level data and the reported state-level aggregations are not equal.
> 
>     * For each column it's just a few dozen records (except for the known `alternative_fuel_fleet_1_yes_0_no_reported` case), but they are not always the same few dozen records.
> 
>     * All the discrepancies are in 2013.
> 
>     * It looks like all of the discrepancies are due to the reported values being zero, but the aggregated values being non-zero.
> 
> 
> I'm using this little function to grab the records that don't match
> 
> ```python
> def state_discrepancies(
>     state_data: pd.DataFrame,
>     aggregate_state: pd.DataFrame,
>     state_diff: pd.DataFrame,
>     col: str,
> ) -> pd.DataFrame:
>     merge_cols = ["report_year", "operating_state"]
>     unmatched_idx = state_diff.loc[state_diff[(col, "self")].notna()].index
> 
>     return pd.merge(
>         state_data.loc[unmatched_idx, merge_cols + [col]],
>         aggregate_state.loc[unmatched_idx, merge_cols + [col]],
>         on=["report_year", "operating_state"],
>         suffixes=("_reported", "_aggregated"),
>     )
> 
> for col in state_diff.columns.levels[0]:
>     display(
>         state_discrepancies(
>             state_data=state_data,
>             aggregate_state=aggregate_state,
>             state_diff=state_diff,
>             col=col,
>         )
>     )
> ```
> 
> I'm going to switch the asset check to being non-blocking for the moment.
> 
> @MeadBarrel Would you rather we merge the new extraction in, or hold off until we can debug this discrepancy on Monday?

This wound up being an artifact of the last data archive, where some 2013 appeared missing. This seems to have been rectified by the latest archiver run.

zaneselvans:
* @e-belfer it was a one line fix to pull the `custom` rather than `data` page in the integration test.
* I also set the order of the integration tests to put these datastore checks before the long slow ETL, so we'd get a quick failure in the future.

PR #4733, opened 2025-11-04
Add CEMS Q4 2025 DOI

aesharpe:
# Overview

Closes #4723

## What problem does this address?

Adds new quarter of CEMS data

## What did you change?

- Update CEMS DOI
- Add release notes
- Update row counts
- Update working partitions

## Documentation

Make sure to update relevant aspects of the documentation:

- [x] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html): reference the PR and related issues.
- [ ] Update relevant Data Source jinja templates (see `docs/data_sources/templates`).
- [ ] Update relevant table or source description metadata (see `src/metadata`).
- [ ] Review and update any other aspects of the documentation that might be affected by this PR.

# Testing

How did you make sure this worked? How can a reviewer verify this?

## To-do list

- [ ] If updating analyses or data processing functions: make sure to update row count expectations in `dbt` tests.
- [ ] Run `make ci` locally to ensure that the merge queue will accept your PR.
- [ ] Review the PR yourself and call out any questions or issues you have.
- [ ] For PRs that change the PUDL outputs significantly, run the full ETL locally and then [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/data_validation.html) using dbt. If you can't run the ETL locally then run the `build-deploy-pudl` GitHub Action manually and ensure that it succeeds.

PR #4742, opened 2025-11-07
Update `core_epa__assn_eia_epacamd_subplant_ids` row counts for 2025Q3 EPA CEMS data

zaneselvans:
# Overview

* In #4733 we added new data from 2025Q3 for EPA CEMS, but did not update the row counts of the assets downstream of EPA CEMS, so the nightly builds failed. This PR updates those row counts.
* It also adds `dbt/seeds/etl_full_row_counts.csv` to the list of files which do not trigger the integration tests in CI, because we only run the row count checks with the full ETL, and there are (so far as I could tell) no other tests which depend on the row counts. `dbt_helper_test` mocks the file internally.


krivard (approved):
🚀 

filtering out rowcount updates from integration testing feels sound.

jdangerx:
Belatedly agreeing that we should skip `ci-integration` for updates to the `etl_full_row_counts.csv` file - I know I was tagged for my opinion here :)

PR #4743, opened 2025-11-07
Integrate 2025 Q4 update for EIA-930

krivard:
# Overview

Closes #4719 

## What did you change?

* Updated DOIs
* Updated rowcounts

## Documentation

Make sure to update relevant aspects of the documentation:

- [x] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html): reference the PR and related issues.

# Testing

How did you make sure this worked? How can a reviewer verify this?

Ran full ETL, and verified most-recent data in EIA-930 output tables advanced from August to November.

## To-do list

- [x] If updating analyses or data processing functions: make sure to update row count expectations in `dbt` tests.
- [x] Run `make ci` locally to ensure that the merge queue will accept your PR.
- [x] Review the PR yourself and call out any questions or issues you have.
- [x] For PRs that change the PUDL outputs significantly, run the full ETL locally and then [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/data_validation.html) using dbt. If you can't run the ETL locally then run the `build-deploy-pudl` GitHub Action manually and ensure that it succeeds.

PR #4744, opened 2025-11-08
Update row-count and nullness expectations for eia930 subregion demand

zaneselvans:
# Overview

* After merging in [updated EIA-930 data for 2025Q3](https://github.com/catalyst-cooperative/pudl/pull/4743) there were some data validation test failures.
* They all seem to be in `{core,out}_eia930__hourly_subregion_demand`
* Slight increase in missingness of `demand_reported_mwh` and `demand_imputed_pudl_mwh_imputation_code`
* Slight increase in row counts for 2019.
* Appearance of half a year's worth of data for 2018 that previously wasn't in the row counts at all?
* Maybe this is a result of the combination of the new DuckDB extractor and the new 930 data?
* Seems odd that these old years would be changing so much.
* Tagging @krivard and @jdangerx to take a look since those EIA-930 PRs were from y'all.

## Testing

- Ran the full ETL locally and ran the dbt tests on the outputs with the updated data validation params, verifying that they pass now.

## To-do list


dbt/models/eia930/out_eia930__hourly_subregion_demand/schema.yml
```diff
@@ -38,5 +38,5 @@ sources:
                     value_set: ["simulated"]
               - expect_missingness_between:
                   arguments:
-                    lower_bound: 0.89
+                    lower_bound: 0.85
```
zaneselvans, line 41: The changes in this file don't seem particularly worrying, though I didn't try to track down which chunk of the data they were coming from. I imagine it's the new data? Maybe the half-year of 2018 data that's suddenly appeared out of nowhere?

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -1279,7 +1279,8 @@ core_eia930__hourly_operations,2022,548958
 core_eia930__hourly_operations,2023,541664
 core_eia930__hourly_operations,2024,535824
 core_eia930__hourly_operations,2025,445994
-core_eia930__hourly_subregion_demand,2019,726527
+core_eia930__hourly_subregion_demand,2018,366043
+core_eia930__hourly_subregion_demand,2019,727080
```
zaneselvans, line 1283: * Weird that the 2019 data would change by like 500 records.
* Even weirder that ~half a year of 2018 data would appear!

krivard: DX noted in the [duckdb pr](https://github.com/catalyst-cooperative/pudl/pull/4731) that we had a bug that caused us to skip 2018half2, and that some timezone wonkiness created some bleed into 2019-01-01, so this makes sense to me.

Shame on me for reading the bug, going "that makes sense" but not also looking for corresponding rowcounts updates before approving the PR. I merged q4 before the duckdb stuff went in, but only *just* before, so I think we each thought the other would cover any additional changes.

PR #4748, opened 2025-11-11
Remove 'make ci' target since pre-commit --all-files is too invasive

zaneselvans:
# Overview

* Our PR template has been instructing folks to run `make ci` which also calls `pre-commit --all-files`
* This is actually more invasive than what `pre-commit.ci` runs in our CI, since some of the local pre-commit hooks are disabled in the `pre-commit.ci` configuration.
* This was resulting in undesirable behavior like unnecessarily regenerating the conda lockfiles, even when no dependencies had been updated.
* Reverting to telling contributors to run `make pytest-coverage` will avoid those unnecessary actions.
* However, we **do** want everybody to be running the local pre-commit hooks during development, since they don't get run by `pre-commit.ci` and without this step, that won't happen if the contributor does not install the pre-commit hooks in their development environment.
* This means we need to keep an eye out for things like unsorted row-counts, or conda lockfiles that haven't been updated when a new dependency has been added.
* If we see `pre-commit.ci` consistently modifying the PR when new pushes are made we should ask whether the pre-commit hooks are installed.

## Documentation

- Updated the dev docs page.
- Updated the PR template.

## To-do list

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,15 +162,3 @@ unmapped-ids:
 		--ignore-foreign-key-constraints \
 		--etl-settings ${etl_full_yml} \
 		test/integration/glue_test.py
-
-########################################################################################
-# Continuous Integration Tests
-########################################################################################
-.PHONY: pre-commit
-pre-commit:
-	pre-commit run --all-files
-
-# This target will run all the tests that typically take place in our continuous
-# integration tests on GitHub (with the exception building our docker container).
-.PHONY: ci
-ci: pre-commit pytest-coverage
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -162,15 +162,3 @@ unmapped-ids:
...
-# Continuous Integration Tests
-########################################################################################
-.PHONY: pre-commit
-pre-commit:
-	pre-commit run --all-files
-
-# This target will run all the tests that typically take place in our continuous
-# integration tests on GitHub (with the exception building our docker container).
-.PHONY: ci
-ci: pre-commit pytest-coverage
```
krivard, line 176: I double-checked and neither of these appears in any of our GHA workflows

krivard (approved):
thanks for this! i've been diligently working around the weird make ci effects for ages, and this is much better

Makefile
zaneselvans: I grepped, but thank you for the second set of 👀 !

PR #4750, opened 2025-11-11
Reduce memory usage of flag timeseries assets by ~half

### LINKED ISSUE #4678: `pudl.analysis.timeseries_cleaning.impute_timeseries_asset_factory` high memory usage
### Overview

These two assets share usage of the `_flag_timeseries_matrix` asset within the `impute_timeseries_asset_factory`:

* `_out_eia930__combined_imputed_demand`
* `out_ferc714__hourly_planning_area_demand`

We should profile and figure out if there are possible memory usage wins here.

### Success Criteria

* [x] identified memory hotspots
* [ ] resolved or determined that memory hotspots are too hard to resolve

### Next steps
* [x] profile those two assets

zschira:
# Overview

Closes #4678.

This PR improves the memory performance of the `_flag_timeseries` assets that make up one portion of our demand imputation pipeline that operates on FERC 714 and EIA 930 data. The specific assets impacted by this change are:

- `_out_ferc714__hourly_planning_area_demand_timeseries_matrix_flags`
- `_out_eia930__combined_demand_timeseries_matrix_flags`

## What did you change?
The peak memory usage comes from a specific call to `numpy.nanmedian` that operates on a very large 3D array. To decrease memory usage I've brought `Dask` back in to our dependency stack (after a short lived absence) due to their [array api](https://docs.dask.org/en/stable/array.html), which implements much of the Numpy api, but allows out-of-memory processing.

For more details on other options explored see #4678.

### Results
This change reduces peak memory usage by roughly half. This still leads to quite high memory usage for the 714 asset (11.9GB), but it is still a significant improvement. I spent a good amount of time attempting to tweak Dask configuration and [array chunksize](https://docs.dask.org/en/stable/array-chunks.html#chunks) to try to further decrease emory usage, but this is as low as I could get it. There may be further room for improvement here, but median computation is inherently expensive, so I'm not sure there's a lot more to squeeze out of this.

#### `_out_ferc714__hourly_planning_area_demand_timeseries_matrix_flags`
[714-baseline-profile.html](https://github.com/user-attachments/files/23528501/memray-flamegraph-_out_ferc714__hourly_planning_area_demand_timeseries_matrix_flags-2025-11-11T19.07.21.html) -> [714-optimized-profile.html](https://github.com/user-attachments/files/23528810/memray-flamegraph-_out_ferc714__hourly_planning_area_demand_timeseries_matrix_flags-2025-11-11T18.00.36.html)

#### `_out_eia930__combined_demand_timeseries_matrix_flags`
[930-basline-profile.html](https://github.com/user-attachments/files/23528825/memray-flamegraph-_out_eia930__combined_imputed_demand_timeseries_matrix_flags-2025-11-10T19.03.21.html) -> [930-optimized-profile.html](https://github.com/user-attachments/files/23528846/memray-flamegraph-_out_eia930__combined_imputed_demand_timeseries_matrix_flags-2025-11-11T14.49.28.html)

# Testing
To validate that the results of this change I ran the `check_against_nightly` script on the following assets:

- `out_ferc714__hourly_planning_area_demand`
- `out_eia930__hourly_subregion_demand`
- `out_eia930__hourly_operations`

These assets are all downstream of the asset actually being modified, but they contain the results of the imputation. Because the actual imputation is non-deterministic, this ends up failing, but for all of the tables I only see differences in the `demand_imputed_pudl_mwh` column. Since this change applies to the flagging step, we would expect to see changes in the `demand_imputed_pudl_mwh_imputation_code` column if this change had impacted flagging.

## To-do list

- [x] Figure out why 714 asset has different row counts
    - I had an outdated upstream asset causing the mismatch.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -33,13 +33,8 @@ jobs:
           filters: |
             code:
               - '!dbt/seeds/etl_full_row_counts.csv'
-              - '!docker/*'
-              - '!docs/**.rst'
-              - '!docs/**.md'
-              - '!docs/**.pdf'
-              - '!docs/**.html'
-              - '!docs/**.rst.jinja'
-              - '!docs/**.bib'
+              - '!docker/**'
+              - '!docs/**'
               - '!.github/**.md'
               - '!.github/FUNDING.yml'
               - '!.github/actionlint.yml'
@@ -52,6 +47,12 @@ jobs:
               - '!LICENSE.txt'
               - '!CITATION.cff'
               - '!.zenodo.json'
+              - '!**.gitignore'
+              - '!.git-blame-ignore-revs'
+              - '!.pre-commit-config.yaml'
+              - '!.readthedocs.yaml'
+              - '!.mypy.ini'
+              - '!test/unit/**'
       - name: print out components
         run: |
           echo "event name (${{ github.event_name }}) is workflow dispatch: ${{ github.event_name == 'workflow_dispatch' }}"
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -23,6 +23,6 @@ authors:
     orcid: https://orcid.org/0009-0009-9063-080X
 
 title: "The Public Utility Data Liberation (PUDL) Project"
-version: 2025.10.0
+version: 2025.11.0
 doi: 10.5281/zenodo.3404014
-date-released: 2025-10-14
+date-released: 2025-11-13
--- a/README.rst
+++ b/README.rst
@@ -116,6 +116,9 @@ What data is available?
 
 PUDL currently integrates data from:
 
+* **EIA Form 176** (a few tables -- work in progress):
+  - `Source Docs <https://www.eia.gov/dnav/ng/TblDefs/NG_DataSources.html#s176>`__
+  - `PUDL Docs <https://catalystcoop-pudl.readthedocs.io/en/nightly/data_sources/eia176.html>`__
 * **EIA Form 860**:
   - `Source Docs <https://www.eia.gov/electricity/data/eia860/>`__
   - `PUDL Docs <https://catalystcoop-pudl.readthedocs.io/en/nightly/data_sources/eia860.html>`__
@@ -132,6 +135,7 @@ PUDL currently integrates data from:
   - `PUDL Docs <https://catalystcoop-pudl.readthedocs.io/en/nightly/data_sources/eia930.html>`__
 * **EIA Annual Energy Outlook (AEO)** (a few tables):
   - `Source Docs <https://www.eia.gov/outlooks/aeo/>`__
+  - `PUDL Docs <https://catalystcoop-pudl.readthedocs.io/en/nightly/data_sources/eiaaeo.html>`__
 * **EPA Continuous Emissions Monitoring System (CEMS)**:
   - `Source Docs <https://campd.epa.gov/>`__
   - `PUDL Docs <https://catalystcoop-pudl.readthedocs.io/en/nightly/data_sources/epacems.html>`__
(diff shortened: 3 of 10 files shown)


src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -878,11 +881,20 @@ def median_of_rolling_median_offset(
...
+        array_file = Path(tmp_dir) / "window.npy"
+        np.save(array_file, shifted)
+        del shifted
+        with da.config.set(scheduler="threads", num_workers=2):
+            return da.array.nanmedian(
+                da.array.from_array(
+                    np.load(array_file, mmap_mode="r"), chunks=(21, 10791, 19)
+                ),
+                axis=0,
+            ).compute()
```
zaneselvans: I wonder if this could also be done in a memory efficient way with DuckDB, which is apparently happy to [query numpy arrays](https://duckdb.org/docs/stable/guides/python/import_numpy)?

zschira: Wow duckdb can do everything. Unfortunately, I spent some time investigating this option this afternoon and ran into the same issues as polars where you need to offload the array onto disk to get any efficiency gains, and that ends up being a massive bottleneck. Also duckdb can't do `SELECT median(*) FROM table`, so you need to auto-generate a query to compute the median on each individual column and there's 10s of thousands of columns.

zaneselvans: Ah okay! Was just hoping to avoid Yet Another Big Data Library but so be it.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -878,11 +881,20 @@ def median_of_rolling_median_offset(
         else:
             shifted[i, :] = offset
     # Ignore warning for rows with all null values
-    with warnings.catch_warnings():
+    with warnings.catch_warnings(), tempfile.TemporaryDirectory() as tmp_dir:
         warnings.filterwarnings(
             "ignore", category=RuntimeWarning, message="All-NaN slice encountered"
         )
-        return np.nanmedian(shifted, axis=0)
+        array_file = Path(tmp_dir) / "window.npy"
```
zschira: Offload the array to disk as an `npy` file, then use dask to compute `nanmedian` out of memory.

src/pudl/analysis/timeseries_cleaning.py
zschira: Yeah it's a bit of a shame to bring Dask back in to only be used in one place.

krivard:
We might not actually need dask for this -- on a hunch I tried a drop-in replacement for nanmedian from `bottleneck` and got a 2/3 reduction in memory (24->8) and 1/3 reduction in running time (18->12)

* profile from main: 
[714-baseline-profile.html](https://github.com/user-attachments/files/23550404/memray-flamegraph-_out_ferc714__hourly_planning_area_demand_timeseries_matrix_flags-2025-11-14T14.46.53.html)

* profile with the below diff: 
[714-bottleneck-profile.html](https://github.com/user-attachments/files/23550406/memray-flamegraph-_out_ferc714__hourly_planning_area_demand_timeseries_matrix_flags-2025-11-14T15.09.25.html)

```
diff --git a/src/pudl/analysis/timeseries_cleaning.py b/src/pudl/analysis/timeseries_cleaning.py
index 7e262b2c5..d300401b7 100644
--- a/src/pudl/analysis/timeseries_cleaning.py
+++ b/src/pudl/analysis/timeseries_cleaning.py
@@ -37,6 +37,7 @@ from collections.abc import Callable, Sequence
 from dataclasses import dataclass, field
 from typing import Any, Literal
 
+import bottleneck as bn
 import numpy as np
 import pandas as pd
 import pandera.pandas as pa
@@ -882,7 +883,7 @@ def median_of_rolling_median_offset(
         warnings.filterwarnings(
             "ignore", category=RuntimeWarning, message="All-NaN slice encountered"
         )
-        return np.nanmedian(shifted, axis=0)
+        return bn.nanmedian(shifted, axis=0)
 
 
 def rolling_iqr_of_rolling_median_offset(
```

zschira:
@krivard nice! I wasn't familiar with `bottleneck`, but it's simpler and more efficient than `dask`, so I've switched over to this implementation.

Also removed the dask dependency.

src/pudl/analysis/timeseries_cleaning.py
```diff
@@ -882,7 +883,7 @@ def median_of_rolling_median_offset(
         warnings.filterwarnings(
             "ignore", category=RuntimeWarning, message="All-NaN slice encountered"
         )
-        return np.nanmedian(shifted, axis=0)
+        return bn.nanmedian(shifted, axis=0)
```
zaneselvans, line 886: It will probably be hard to beat this for "memory reduction per characters changed"

zaneselvans:
@zschira Zenodo is having a rough time today for some reason, and causing the RTD builds to fail. I wouldn't worry about it. Do we want to go ahead and merge? Is there additional testing to do?

zaneselvans:
Imma go ahead and merge this guy in since it's such a tiny change for such a big improvement.

PR #4758, opened 2025-11-12
feat: add labels to cloud resources

jdangerx:
# Overview

Our cloud resources didn't have labels, so it's hard to figure out how much $ is being spent on what part of our system.

## What did you change?

Added some labels to what we have defined in this repo.

# Testing

* ran terraform plan, looked sensible
* ran terraform apply, looked for labels in GCP console
* ran branch build to see if we can see the label on the corresponding batch job

## To-do list

- [x] go through and label things in the console that are not managed by terraform

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -55,6 +55,7 @@ jobs:
               - '!.pre-commit-config.yaml'
               - '!.readthedocs.yaml'
               - '!.mypy.ini'
+              - '!terraform/**'
               - '!test/unit/**'
       - name: print out components
         run: |


jdangerx:
This should take a few days before it starts to show up in the billing console but otherwise is done!

zschira:
I'm co-opting this PR to also increase the memory for the `pudl_usage_metrics_dashboard`. This is because running `terraform apply` with just the memory increase would end up destroying all of these labels that have already been applied.

zaneselvans:
@zschira if we've already applied these changes, should we just go ahead and merge the PR to avoid this kind of potential conflict?

zaneselvans:
@jdangerx the new metrics dashboard resources that I set manually in the Console seem to have worked well (modulo the 10m timeout?) in our check-in today. Do we want to integrate them in here and get this merged before any additional changes accumulate?

zaneselvans:
@jdangerx is there any reason that we shouldn't exclude everything under `terraform/` from triggering the integration tests in CI?

PR #4746, opened 2025-11-10
Create initial EIA-176 data source documentation page

zaneselvans:
# Overview

- Create a data source documentation page for the EIA 176 (natural gas sources & dispositions)
- Add historical Form 176 instructions and blank forms to our docs for reference.
- Remove EIA 176 from "Future Data" -- THE FUTURE IS NOW
- Update PUDL Data Viewer URLs in the data source documentation pages to use https://data.catalyst.coop
- Fix a couple or RST bugs in the generic child page Jinja template

## Documentation

- [x] Update relevant Data Source jinja templates (see `docs/data_sources/templates`).

# Testing

- Built the docs locally and inspected them

## To-do list

- [x] Review the PR yourself and call out any questions or issues you have.

PR #4759, opened 2025-11-12
Add a bunch of additional files that should not trigger pytest-integration in CI

zaneselvans:
# Overview

* Inspired by #4746 taking a long time to merge, I went digging for additional files which should not trigger our slow-ass integration tests in CI on GitHub.
* Things I'm not 100% sure about that would be nice if they work:
  * I **think** that I've now actually excluded **everything** under `docs/`.  In which case we should just exclude `docs/*`. Is there any scenario where a change in our docs would break the integration tests, but not break the docs build?
  * Is there any scenario where a change to `test/unit` would break the integration tests but not the unit tests? Also do I have the file globbing syntax right there?
  * Do I have the syntax on `.gitignore` right? Or should it be `!./**.gitignore`? And does that seem like a bad thing to exclude?
  * Any other configuration files that could be safely excluded from integration testing? Most of it is in `pyproject.toml` which we shouldn't exclude.
* Also randomly noticed our Docker container image was out of date.

## To-do list

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -34,7 +34,7 @@ jobs:
             code:
               - '!dbt/seeds/etl_full_row_counts.csv'
               - '!docker/**'
-              - '!docs/**
+              - '!docs/**'
               - '!.github/**.md'
               - '!.github/FUNDING.yml'
               - '!.github/actionlint.yml'


.github/workflows/pytest.yml
```diff
@@ -52,6 +58,12 @@ jobs:
               - '!LICENSE.txt'
               - '!CITATION.cff'
               - '!.zenodo.json'
+              - '!**.gitignore'
```
zaneselvans, line 50: @copilot will this '!**.gitignore' entry ignore *all* .gitignore files in the repository?

.github/workflows/pytest.yml
zaneselvans: I think this is now literally every file that's checked-in under `docs/` -- can we safely exclude everything under `docs/` from triggering the integration tests?  If it's going to break anything it **seems** like it should break the docs build.

cmgosnell: i think it makes sense to not kick off an integration run for just docs changes given the docs build. I agree with your suggestion in the pr comment that this would be more straightforward imo to just include the whole directory not list every single possible file type.

cmgosnell (commented):
this looks overall good to me but i'm not sure if you want me to asses the working-ness of some of the things in your pr comment or not.

- ignoring all docs sounds reasonable to me.
- i can't think of why changing unit would break integration but not break the unit tests. i could just not be imaginative enough but seems like a fine bet - we'd certainly learn.
- on gitignore here you are trying to all of those files anywhere? i think that makes sense but as you know i learned about these files yesterday. it looks okay to me given my just now research but idk.
- not that i can think of right now - we can always go back and add more.

Issue #4727, opened 2025-11-04
Resolve loose threads and unanswered questions about EIA 176


e-belfer:
### Overview

What is the problem we're solving? For very simple items, this can be encapsulated in the success criteria.

This issue keeps track of a number of missing data fields, unexpected behaviors and topics that require further investigation. We should email the EIA and ask for explanations, seek alternative data sources for these fields, or at the very least document their missingness in our documentation once other avenues have been exhausted.

#### Missing other transport revenue

In line 11.6, we see `other transport revenue` as a field in our [custom report](https://www.eia.gov/naturalgas/ngqs/#?report=RPC&year1=1997&year2=2023&company=Name&items=1160CS,1160CT,1160VL), but there is never any data reported. Does this data exist anywhere, and is there a reason it isn't currently being published by the EIA NGQS?

PR #4765, opened 2025-11-14
Create `core_eia176__yearly_gas_disposition`

### LINKED ISSUE #4708: Create `core_eia176__yearly_gas_disposition` (Part 6B)
### Overview

Part 6B: Natural and Supplemental Gas Disposition for the Report State (Lines 9.0 and 12.0-20.0) of the EIA 176 form contains valuable information about an operator's gas consumption, transfers, disposition and losses. Let's bring this data into PUDL!

### Success Criteria

There is no one size fits all in table transforms, though we document our approach to cleaning data [here](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_guidelines.html#examples-of-acceptable-changes).

The new table should meet the following criteria:
* a well defined and unique primary key: one or more columns that naturally explain what a row in the table corresponds to. For this table, this _should_ be one row per year and ID. You'll need to check that the combination of these fields is unique, and in the case of duplications, propose a strategy for creating a unique primary key.
* all categorical columns should be constrained (e.g., the state column shouldn't contain "New York City" in it).
* nulls should be standardized
* units should be standardized
* data types should be defined and enforced
* metadata should be added to the database schema for each new field and for the table itself
* data should be validated in DBT

### Prerequisites
- #4694 

### Transform the data
- [ ] Make a copy of `devtools/inspect-assets.ipynb`. Use the notebook to load the `_core_eia176__yearly_company_data` asset and inspect it. This asset contains all years of the table extracted and concatenated together.
- [ ] Isolate the fields relevant to this table (lines 9.0 and 12.0-20.0). You can use the [survey form](https://www.eia.gov/survey/form/eia_176/form.pdf) and the `form_line_numbers` field in the `raw_eia176__data` table to identify them.
- [ ] In the notebook, explore the table and transform it. Here are some docs re [standard pudl transforms](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_guidelines.html#examples-of-acceptable-changes).
    - [ ] Make sure there are no non-standard nulls (e.g., "-"). If there are, replace them with `pd.NA`.
    - [ ] Check for primary keys of the table and drop if there are **clear** errors (ex: duplicates on expected PKs but with clear bad records aka one decent looking record and one with mostly nulls).
    - [ ] Handle unexpected dtypes: e.g., no numbers in the state column
    - [ ] Convert the `operating_state` column to well-behaved state subdivision codes (e.g., Alaska to AK). Do this by mapping the existing state names against `core_pudl__codes_subdivisions`, cleaning up strings as is needed. We'll need to decide how to handle the non-state values (e.g., Mexico, other) reported for correction records.
    - [ ] Clean up column names as needed to create succinct but informative variable names.

#### Bring the tables into PUDL's ETL
- [ ] Once you're satisfied with the table, it's time to bring it into dagster.
   - [ ] In `pudl/src/pudl/transform/eia176.py` add a new `core` asset function based on [asset naming conventions](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/naming_conventions.html#asset-naming-conventions). In this case: `core_eia176__yearly_gas_disposition`
   - [ ] add the `@asset()` decorator to this function.
   - [ ] Migrate your transforms for the notebook into this function.
   - [ ] [Reload the code location in dagster](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/dev_dagster.html#reloading-code-locations), search for your new table in the "Asset" tab and materialize it. See [our docs on Dagster](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/dev_dagster.html) for support.
- [ ] Add metadata for this table to add this table to the database schema
   - [ ] define the table  in `pudl/src/metadata/resources/eia176.py`
       - [ ] define the fields (aka column names)
       - [ ] add primary keys if possible
       - [ ] add `"field_namespace": "eia",`
       - [ ] add `"sources": ["eia176"],`
       - [ ] add `"etl_group": "eia176",`
       - [ ] consult with a Catalyst developer about how to update table documentation
    - [ ] Go add any new fields into `pudl/src/pudl/metadata/fields.py`
       - [ ] check that each field doesn't already exist.
       - [ ] add a description for each field.
       - [ ] add the data type
       - [ ] add constraints (enums/categories or patterns) if relevant.
       - [ ] add units
- [ ] Write the table to the PUDL database
   - [ ] Add the io manager to the asset() decorator: `@asset(io_manager_key="pudl_io_manager")`
   - [ ] [update alembic schema](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_the_etl.html#database-schema-migration)
- [ ] Materialize your table via the [dagster ui](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/run_the_etl.html#running-the-etl-via-the-dagster-ui)
- [ ] Add a table schema in [dbt](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_validation_reference.html#adding-new-tables) and add [expected row counts](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/data_validation_quickstart.html#row-counts)
- [ ] Add any other DBT tests that seem sensible (e.g., volume can't be negative).

MeadBarrel:
# Overview

Closes #4708.

## Documentation

# Testing

## To-do list

### DIFF
--- a/migrations/versions/94f669451682_add_core_eia176__yearly_gas_disposition.py
+++ b/migrations/versions/94f669451682_add_core_eia176__yearly_gas_disposition.py
@@ -1,7 +1,7 @@
 """Add core_eia176__yearly_gas_disposition
 
 Revision ID: 94f669451682
-Revises: 5ba8ad3ad527
+Revises: aeb66b0b4481
 Create Date: 2025-12-03 15:27:46.520134
 
 """
@@ -11,7 +11,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = '94f669451682'
-down_revision = '5ba8ad3ad527'
+down_revision = 'aeb66b0b4481'
 branch_labels = None
 depends_on = None
 


src/pudl/metadata/fields.py
```diff
@@ -8530,6 +8674,7 @@
 """
 
 FIELD_METADATA_BY_RESOURCE: dict[str, dict[str, Any]] = {
+    "core_eia176__yearly_gas_disposition": {},
```
zaneselvans: If there's no field metadata to override do we need this here? Or is it just a placeholder for now?

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
 # )
 
 FIELD_METADATA: dict[str, dict[str, Any]] = {
+    "heat_content_of_delivered_gas": {
+        "description": "Heat content of gas delivered to consumers.",
+        "unit": "Btu/cf",
+        "type": "number",
+    },
```
zaneselvans: ## Units

Everywhere else I think the units we're using for natural gas are MMBtu/Mcf which results in numbers in the range of 0.8 to 1.2, which is also nice for human readability (a barrel of oil is ~6 MMBtu, a ton of coal is ~17-25 MMBtu, so those are also readable, though ridiculous, units). All the other volumes in this table appear to be in Mcf, so I think converting this column to MMbtu/Mcf (dividing the number by 1000) would be preferable.

## Naming

Elsewhere we've used fields named `heat_content` to mean the total energy content (e.g. MMBtu) rather than the energy density, so I could see this being a little confusing.

In the `out_eia923__fuel_receipts_costs` table we have a field with analogous units of energy contained per unit of a fuel but it's generic (literally `fuel_mmbtu_per_unit`) because the table contains many different types of fuel, which are measured in different units.

Something like `volumetric_energy_density_mmbtu_per_mcf` would be very precise, but in the industry it seems like people use "heat content" to mean both the total energy and the energy density, so the precise name might not actually be the most readable. So I think I would go with `delivered_gas_heat_content_mmbtu_per_mcf`

## Description

Maybe include more detail in the description to avoid any ambiguity "The volumetric energy density (heat content per unit) of gas delivered to consumers."  I assume this is an average over all deliveries for that state-operator-year? If so maybe "The average volumetric energy..."

Do we know if this number really only applies to the gas delivered to consumers? Or would it apply to the other dispositions listed in the table?

src/pudl/metadata/resources/eia176.py
```diff
@@ -31,5 +31,46 @@
         "field_namespace": "eia",
         "sources": ["eia176"],
         "etl_group": "eia176",
-    }
+    },
+    "core_eia176__yearly_gas_disposition": {
```
zaneselvans, line 35: The name of this table suggests that it is all-inclusive, but I'm not sure if it reports the gas that's in the `core_eia176__yearly_gas_disposition_by_consumer` table, or if the two tables report entirely separate gas dispositions. Some clarification in the description would be helpful, and if the two tables are reporting separate gas quantities, maybe we can come up with a more informative table name for this one that indicates how it's different.

src/pudl/metadata/resources/eia176.py
```diff
@@ -31,5 +31,46 @@
...
+                "disposition_to_storage_operators_volume",
+                "total_disposition_volume",
+                "unaccounted_for",
+                "out_of_state_deliveries_volume",
+                "disposition_to_other_volume",
+            ],
+            "primary_key": [
+                "operator_id_eia",
+                "report_year",
+            ],
```
zaneselvans, line 84: I'm surprised that `operating_state` is not also part of the primary key here.  Is that just because the state abbreviation is part of `operator_id_eia` so the same `operator_id_eia` never appears in more than one state? Conceptually it seems like it should be part of the PK, and I think I'm inclined to include it even if it's not literally required to make the PKs unique, since it's kind of incidental that the state is part of the operator ID.

src/pudl/metadata/fields.py
zaneselvans, line 1: ## Naming

Other than the one heat content field, it looks like all of the data fields in this table are volumes of gas reported in Mcf, so they should probably all reflect that in the same way in their names.  Mcf is a unit of volume, so if we explicitly state the unit with an `_mcf` suffix as we have for some other gas volume fields, then we don't need to also say volume in the field name.

Given that this is a natural gas data source and a table about natural gas, would it make sense to say "gas" in every column name, or can we just leave that implied by the context?

If I understand correctly, all of these columns are "dispositions" -- somewhere the gas ends up -- but it seems like there are several sub-categories. What are they, and can we make them clear in the naming so that those categories can be selected with a naming filter?

### Operational Consumption
  * `operational_consumption_facility_space_heat_mcf`
  * `operational_consumption_new_pipeline_fill_mcf`
  * `operational_consumption_compressors_mcf`
  * `operational_consumption_lng_vaporization_liquefaction_mcf`
  * `operational_consumption_vehicles_mcf`
  * `operational_consumption_other_mcf`

### Operational Storage
  * `operational_storage_underground_mcf`
  * `operational_storage_lng_mcf`

### Producer
  * `producer_lease_use_mcf`
  * `producer_reservoir_repressurization_mcf`

### Disposition

It feels a little weird to say disposition only here, since all of the fields are dispositions of some kind.

  * `disposition_distribution_companies_mcf`
  * `disposition_other_pipelines_mcf`
  * `disposition_storage_operators_mcf`
  * `disposition_out_of_state_mcf`
  * `disposition_other_mcf`

### Losses?

Maybe these should get a `disposition_` prefix for structural uniformity / pivot-ability?

* `losses_mcf` (it's all types of losses not just leaks)
* `unaccounted_for_mcf`

## A possible alternative "tidy" normalization

* If there's effectively a hierarchical index for the gas dispositions, with higher level categories like operational consumption, we could also pivot the table so that there's a single `gas_volume_mcf` data column (holding out the PK columns, and broadcasting state and heat content across all applicable records), and a couple of categorical columns which indicate the category and sub-category of gas disposition.
* This arrangement would allow easy aggregation by explicit higher level categories rather than storing that information implicitly in the column names, and would result in "tidier" data (one variable per column, one measurement per row)

## Total and consumers?

* Do the volumes of gas delivered to consumers show up in any of the categories in this table?
* Does `total_disposition_volume` represent the total of just the values listed in this table? Or does it also include gas delivered to consumers?
* Ideally we would verify that the total is consistent with the other data in the table in the transform step, and then drop it from the `core_` table since it is a derived value.

src/pudl/metadata/fields.py
zaneselvans: * "psia" should be capitalized PSIA.
* There are a few places that have a "Dispositiont" typo that was cut-and-pasted.
* Some places where the "(14.73..." needs a space before it to separate it from previous words in the descriptions.
* Note that multi-line description strings get concatenated and need to have a space explicitly added, e.g. at the end of each line.

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+    "out_of_state_deliveries_volume": {
+        "description": (
+            "Total volume of operator's out of state deliveries"
+            "(14.73 psia and 60° Fahrenheit)"
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
```
zaneselvans: Are the out of state deliveries greater than or equal to zero by definition? Would a negative number indicate receipt of gas from another state?  Or would that be accounted for separately in the supply table? If we know, can we clarify in the description?

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+    "op_gas_consumption_by_other": {
+        "description": (
+            "Volume of natural gas used in other company's operations"
+            "(14.73 psia and 60° Fahrenheit)"
+        ),
```
zaneselvans, line 106: Is this "another company's operations" or "other operations" -- miscellaneous uncategorized operations of the reporting company?

src/pudl/transform/eia176.py
```diff
@@ -359,3 +349,134 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+@asset(io_manager_key="pudl_io_manager")
+def core_eia176__yearly_gas_disposition(
+    _core_eia176__yearly_company_data: pd.DataFrame,
+    core_pudl__codes_subdivisions: pd.DataFrame,
+    raw_eia176__continuation_text_lines: pd.DataFrame,
+) -> pd.DataFrame:
+    """Produce company-level gas disposition (EIA176, Part B)"""
+    extras = ["operating_state"]
+
+    keep = [
```
zaneselvans, line 368: When available, it would be helpful to include the appropriate line number in the field description, so that a user can conclusively trace the reported value back to where it would have appeared in a filing.

zaneselvans (commented):
Focused on naming & description issues with a little bit of units conversion.

We might also want to consider a skinnier "tidy" normalization of this table, or at lease setting up the naming of the columns to make a programmatic reshaping of the table into such that it has a hierarchical categorical index as easy as possible if someone wants that arrangement (based on the data column names).

@krivard do you have any naming thoughts you'd like to add?

krivard:
@zaneselvans I reformatted what I think are your recommendations as a table in `src/pudl/metadata/resources/eia176.py` order to make sure I caught them all, can you check this is what you meant?

| name in e93d380 | proposed change
| - | -
| operator_id_eia | ✅ 
| report_year | ✅
| operating_state | ✅
|heat_content_of_delivered_gas | heat_content_mmbtu_per_mcf
|op_gas_consumption_by_facility_space_heat | operational_consumption_facility_space_heat_mcf
|op_gas_consumption_by_new_pipeline_fill | operational_consumption_new_pipeline_fill_mcf
|op_gas_consumption_by_pipeline_dist_storage_compressor_use | operational_consumption_compressors_mcf
|op_gas_consumption_by_vaporization_liquefaction_lng_fuel | operational_consumption_lng_vaporization_liquefaction_mcf
|op_gas_consumption_by_vehicle_fuel | operational_consumption_vehicles_mcf
|op_gas_consumption_by_other | operational_consumption_other_mcf
|operations_volume_underground_storage_injections | operational_storage_underground_mcf
|operations_volume_lng_storage_injections | operational_storage_lng_mcf
|lease_use_volume | producer_lease_use_mcf
|returns_for_repress_reinjection_volume | producer_reservoir_repressurization_mcf
|losses_from_leaks_volume | losses_mcf
|disposition_to_distribution_companies_volume | disposition_distribution_companies_mcf
|disposition_to_other_pipelines_volume | disposition_other_pipelines_mcf
|disposition_to_storage_operators_volume | disposition_storage_operators_mcf
|total_disposition_volume | ❌ [drop if derivative]
|unaccounted_for | unaccounted_for_mcf
|out_of_state_deliveries_volume | disposition_out_of_state_mcf
|disposition_to_other_volume | disposition_other_mcf

My thoughts:

(& just in case nobody pointed you at this yet: [PUDL naming conventions for columns & fields](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/naming_conventions.html#columns-and-field-names))

* for the prefixes, consider operator_, producer_, and consumer_
  * pro: Makes all the prefixes the same kind of word (the agent of disposition), instead of a mix
  * con: "consumer" isn't great, since (1) we'd normally identify "consumer" with "end user", but these dispositions aren't themselves end users, and (2) it's quite close to "consumption", but not all these dispositions end up actually burning the gas. If we can find an agent word that more clearly indicates "somebody who isn't the operator and didn't produce it" we should use that instead.
* is `injection` an industry keyword we should try to retain in column names? if so consider `*_injection_mcf` as a pattern; would affect these three:
  * operational_storage_underground_mcf (/operations_volume_underground_storage_injections)
  * operational_storage_lng_mcf (/operations_volume_lng_storage_injections)
  * producer_reservoir_repressurization_mcf (/returns_for_repress_reinjection_volume)
* is a pipeline operator fundamentally different from a utility? for electricity we try to use "utility" everywhere (including places where eia uses "operator"), but it looks like the existing PHMSA tables use "operator" instead. I'm okay if we want to use operator for all things gas, but we should make that choice intentionally and then update the naming conventions. would affect:
  * operator_id_eia

src/pudl/metadata/fields.py
MeadBarrel: > Does total_disposition_volume represent the total of just the values listed in this table? Or does it also include gas delivered to consumers?

Yeah this includes gas delivered to consumers, everything from 10.1 to 18.4. This table doesn't include 10.1 - 11.6, they're handled in `core_eia176__yearly_gas_disposition_by_consumer`. I'm not sure what to do with it. Maybe add a new column here like "disposition_consumer_mcf" which would be a sum of 10.1 - 11.6 volumes? It would makes sense if this one is all-inclusive. Then `core_eia176__yearly_gas_disposition_by_consumer` would be a detailed breakdown.

That would result in two sources of truth for consumer disposition but well nothing is perfect.

MeadBarrel: > If there's effectively a hierarchical index for the gas dispositions, with higher level categories like operational consumption, we could also pivot the table so that there's a single gas_volume_mcf data column (holding out the PK columns, and broadcasting state and heat content across all applicable records), and a couple of categorical columns which indicate the category and sub-category of gas disposition.

I thought of that, though personally i hate it when data is duplicated, which would have to be the case with `heat_content_of_delivered_gas_btu_cf` as it'll need to be broadcasted. Maybe there's some other table this column could belong to?

src/pudl/metadata/fields.py
MeadBarrel: Yeah it's "other operations". The survey form has a text field to specify the type of the operations, maybe it should be included.

src/pudl/metadata/fields.py
MeadBarrel: Just a placeholder.

src/pudl/metadata/resources/eia176.py
MeadBarrel: Check https://github.com/catalyst-cooperative/pudl/pull/4765#discussion_r2545580606

> but I'm not sure if it reports the gas that's in the core_eia176__yearly_gas_disposition_by_consumer table, or if the two tables report entirely separate gas dispositions. 

19.0 (`total_disposition_volume` currently) is the sum of everything from 10.1 to 18.4, i.e. every volume value in this table plus `core_eia176__yearly_gas_disposition_by_consumer`.

src/pudl/metadata/fields.py
MeadBarrel: Well at least in the data they're all >=0, so there's that.

src/pudl/metadata/resources/eia176.py
e-belfer: The operator ID (both number and state) are assigned uniquely to each operator in a state (e.g., the same operator might be 100AK in Alaska and 107ID in Idaho). In that sense, state is a conceptual part of the primary key but not at all required to make the operator ID uniquely identifying.

src/pudl/transform/eia176.py
```diff
@@ -359,3 +349,134 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+        }
+    )
+
+    return df
+
+
+def _normalize_operating_states(
+    core_pudl__codes_subdivisions: pd.DataFrame, df: pd.DataFrame
+):
+    """Normalize operating states in-place"""
```
e-belfer:
```suggestion
    """Map full state names to their postal abbreviations.
    
    This uses the latest year of Census PEP data as the reference. If a full state name is not included in this data, it is set to NA."""
```

src/pudl/transform/eia176.py
```diff
@@ -359,3 +349,134 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+    codes = (
```
e-belfer, line 532: We are quietly assuming here that we aren't nulling any good state names that are typo'd or otherwise don't show up in the census data. It would be helpful to add an assertion that checks against known exceptions we expect to drop (e.g., Mexico, gulf of Mexico) so we know we're only dropping expected bad cases.

src/pudl/transform/eia176.py
e-belfer: We prefer to avoid in-place calls and return the `df` to avoid any confusion around missing assignments.

src/pudl/transform/eia176.py
```diff
@@ -359,3 +349,134 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+        columns=["deliveries_out_of_state_volume", "disposition_to_other_volume"]
+    )
+    df = df.rename(
+        columns={
+            1400: "deliveries_out_of_state_volume",
+            1840: "disposition_to_other_volume",
+        }
+    )
+
+    _normalize_operating_states(core_pudl__codes_subdivisions, df)
```
e-belfer: See my comment below re: avoiding in-place calls in the codebase.

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
 # )
 
 FIELD_METADATA: dict[str, dict[str, Any]] = {
+    "heat_content_of_delivered_gas": {
```
e-belfer: See my meta comment re renaming these fields.

dbt/models/eia176/core_eia176__yearly_gas_disposition/schema.yml
```diff
@@ -0,0 +1,108 @@
...
+
+        columns:
+          - name: operator_id_eia
+          - name: report_year
+          - name: operating_state
+          - name: heat_content_of_delivered_gas
+            data_tests:
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
```
e-belfer: From the [instructions](https://www.eia.gov/survey/form/eia_176/instructions.pdf), we actually see:

> The value for heat content is expected to be in the range of 900 to 1,200 Btu/cf

Can we try and test this range instead to see whether it's met?

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        "description": "Heat content of gas delivered to consumers.",
+        "unit": "Btu/cf",
```
```suggestion
        "description": "The average annual heat content of gas delivered directly to consumers.",
        "unit": "MMBtu_per_Mcf",
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        "type": "number",
+    },
+    "op_gas_consumption_by_facility_space_heat": {
+        "description": (
+            "Volume of natural gas consumed for space heat of operator's facilities "
+            "(14.73 psia and 60° Fahrenheit)"
+        ),
+        "unit": "Mcf",
```
```suggestion
            "Volume of natural gas consumed as fuel for space heat of the operator's own facilities within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
        ),
        "unit": "Mcf",
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        "type": "number",
+    },
+    "op_gas_consumption_by_new_pipeline_fill": {
+        "description": (
+            "Volume of natural gas consumed for operator's new pipeline fill "
+            "(14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "Volume of natural gas consumed as fuel for the operator's new pipeline fill within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
e-belfer: > * "psia" should be capitalized PSIA.

It isn't capitalized in the form and isn't capitalized in most of the natural gas documentation I see, so I've suggested to leave it uncapitalized as well.

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "op_gas_consumption_by_pipeline_dist_storage_compressor_use": {
+        "description": (
+            "Volume of natural gas consumed by operator's pipeline distribution or storage "
+            "compressor use (14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "Volume of natural gas consumed as fuel for the operator's own pipeline distribution or storage compressor use within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "op_gas_consumption_by_vaporization_liquefaction_lng_fuel": {
+        "description": (
+            "Volume of natural gas consumed by operator for vaporization/liquefaction/LNG fuel"
+            "(14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "Volume of natural gas consumed by the operator for vaporization, liquefaction, and LNG fuel within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "op_gas_consumption_by_vehicle_fuel": {
+        "description": (
+            "Volume of natural gas used as company's fleet vehicle fuel"
+            "(14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "Volume of natural gas used in the operator's company-owned fleet within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "op_gas_consumption_by_other": {
+        "description": (
+            "Volume of natural gas used in other company's operations"
+            "(14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "Volume of natural gas consumed as fuel by the operator for other purposes within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
e-belfer: I agree, we should do this. See my main comment about adding the detail field in from the continuation text lines table and not groupby/summing this field.

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "operations_volume_underground_storage_injections": {
+        "description": (
+            "Underground storage injections volume (14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "The total volume added to underground storage operated by the company within the report state, regardless of ownership of the gas. This includes new fields. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "operations_volume_lng_storage_injections": {
+        "description": (
+            "Liquified natural gas (LNG) storage injections volume (14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "The total volume added to LNG storage operated by the company within the report state, regardless of ownership of the gas. For LNG import and export marine terminals, this should exclude injections of LNG during the course of routine operations for handling imports, as well as LNG held in storage for future use. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "lease_use_volume": {
+        "description": (
+            "Lease use volume (producers only) (14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "The total volume of gas used in the company's well, field and lease operations. This field should only be reported by producers. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "returns_for_repress_reinjection_volume": {
+        "description": (
+            "Volume of gas returned to oil or gas reservoirs, used for repressuring, reinjection "
+            "(producers only) (14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "The volume of gas delivered directly from the operator's system to oil or gas fields located within the report state for repressuring, pressure maintenance, and cycling operations. This field should only be reported by producers. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "losses_from_leaks_volume": {
+        "description": (
+            "Losses from leaks, damage, accidents, migration and/or blow down"
+            "(14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "Known loss volumes as a result of leaks, damage, accidents, migration and blow down within the report state where the events took place, as well as estimated losses from leaks occurring during distribution activities. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "disposition_to_distribution_companies_volume": {
+        "description": (
+            "Disposition to distribution companies within the report state"
+            "(14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "Disposition to distribution companies within the report state that does not fall into one of the other reported categories in lines 10.1-17.0. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "disposition_to_other_pipelines_volume": {
+        "description": (
+            "Dispositiont to other pipelines within the report state"
```
```suggestion
           "Disposition to other pipelines within the report state that does not fall into one of the other reported categories in lines 10.1-17.0. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+            "(14.73 psia and 60° Fahrenheit)"
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "disposition_to_storage_operators_volume": {
+        "description": (
+            "Dispositiont to storage operators within the report state"
+            "(14.73 psia and 60° Fahrenheit)"
```
```suggestion
           "Disposition to storage operators within the report state that does not fall into one of the other reported categories in lines 10.1-17.0. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "total_disposition_volume": {
+        "description": (
+            "Total disposition within the report state(14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "Total disposition within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "unaccounted_for": {
+        "description": (
+            "Difference between gas supply (+) and disposition (-)"
+            "(14.73 psia and 60° Fahrenheit)"
```
```suggestion
        "description": (
            "The difference between gas supply and disposition. A positive entry indicates supply in excess of accounted-for disposition. A negative entry indicates accounted-for disposition exceeds reported supply. This is calculated as the difference between Part 4 Line 7.0 and Part 6 Line 19.0, and is reported as Line 20.0 in the original form."
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+        "description": (
+            "Total volume of operator's out of state deliveries"
+            "(14.73 psia and 60° Fahrenheit)"
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "disposition_to_other_volume": {
+        "description": (
+            "Disposition not included in other fields(14.73 psia and 60° Fahrenheit)"
```
```suggestion
            ""Other disposition within the report state that does not fall into one of the other reported categories in lines 10.1-17.0. This has been summed from the detailed data reported by each company on Line 18.4 of the original form in order to preserve the primary key of the table. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

See my comments in the main block - I think leaving this as an aggregated field makes sense, and we could consider adding a second detail table just for the "other" records in this field if time allows.

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+            "Difference between gas supply (+) and disposition (-)"
+            "(14.73 psia and 60° Fahrenheit)"
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "out_of_state_deliveries_volume": {
+        "description": (
+            "Total volume of operator's out of state deliveries"
+            "(14.73 psia and 60° Fahrenheit)"
```
```suggestion
            "Total volume of the operator's deliveries across or to state lines or U.S. borders. This has been summed from the detailed data reported by each company on Line 14.0 of the original form in order to preserve the primary key of the table. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```

See my main comments re: what to do with the set of fields reported in 14.0. When we add the detailed table, we should reference it explicitly in this description.

e-belfer (commented):
Digesting the other comments and my own reflections into a series of tasks:

**Overarching comments**
- [x] There's a new alembic migration in main, so you'll need to update the latest migration to point at it or merge the two divergent migrations
- [x] There is one negative `losses_from_leaks_volume` currently, for 17617106KS in 2008. If the total field is in fact a sum of the other components minus the absolute value of this record, we should treat this as a sign error and flip it. If not, we'll need to ask the EIA what's up with this record.
- [x] At this stage, the big totals field (Line 19.0) contains original data including values not in this table, so I think we should leave it as is and not delete it.

**Heat content of delivered gas**

Image

- [x] There are a large number of exact `9999` values prior to 2012. These should be treated as `NA` values and converted to nulls.
- [x] Once we've done this, we can convert the `heat_content_of_delivered_gas` to MMbtu/Mcf (dividing the number by 1000) to be consistent with our other units

**Line 12.6**
- There's only ever one "other" type per operator ID/year, so I think we should *not* groupby sum and aggregate this field. Instead, we should preserve `reference_company_or_line_description` as a new field `operational_consumption_other_detail`
- There are two records in here that have "Vaporization/LNG" as the other field. One of these reports other and different values in the dedicated vaporization/LNG field however, so we should just leave these as is and not reallocate them.

**Line 14.0**
- **The detailed data should live in a separate table that captures the vessel type, company name, state/county, and volume.**
- Leaving the aggregated field in the main table with a note about its processing and a reference to the additional table seems fine to me, however. I've added a suggestion in `fields.py` about how to do so.

**Line 18.4**
- ~.01% of gas is getting reported as other, and each operator can report multiple other fields. Sometimes these fields have distinct "other" type descriptions, and sometimes they don't. Given that this would not produce an additional well-behaved PK, I think the approach of groupby and sum that is currently used to produce a single field is well-justified.
- **As a next step, we should consider making an additional table that contains the unaggregated records and the detail field from `reference_company_or_line_description`  as a separate detail table.**

**Column names and metadata**
In terms of columns, distilling the feedback from above, the following renames make sense to me:
- [ ] `heat_content_of_delivered_gas` to `delivered_gas_heat_content_mmbtu_per_mcf`
- [ ] `op_gas_consumption_by_facility_space_heat` to `operational_consumption_facility_space_heat_mcf`
- [ ] `op_gas_consumption_by_new_pipeline_fill` to `operational_consumption_new_pipeline_fill_mcf`
- [ ] `op_gas_consumption_by_pipeline_dist_storage_compressor_use`  to `operational_consumption_compressors_mcf`
- [ ] `op_gas_consumption_by_vaporization_liquefaction_lng_fuel` to `operational_consumption_lng_vaporization_liquefaction_mcf`
- [ ] `op_gas_consumption_by_vehicle_fuel` to `operational_consumption_vehicle_fuel_mcf`
- [ ] `op_gas_consumption_by_other` to `operational_consumption_other_mcf`
- [ ] `operations_volume_underground_storage_injections` to `operational_storage_underground_mcf`
- [ ] `operations_volume_lng_storage_injections to `operational_lng_storage_injections_mcf`
- [ ] `out_of_state_deliveries_volume` to `disposition_out_of_state_mcf`
- [ ] `lease_use_volume` to `producer_lease_use_mcf`
- [ ] `returns_for_repress_reinjection_volume` to `producer_returned_for_repressuring_reinjection_mcf`
- [ ] `losses_from_leaks_volume` to losses_mcf`
- [ ] `disposition_to_distribution_companies_volume` to `other_disposition_distribution_companies_mcf`
- [ ] `disposition_to_other_pipelines_volume` to `other_disposition_pipelines_mcf`
- [ ] `disposition_to_other_volume` to `other_disposition_all_other_mcf`
- [ ] `disposition_to_storage_operators_volume` to `disposition_storage_operators_mcf`
- [ ] `unaccounted_for` to `unaccounted_for_mcf`
- [ ] `total_disposition_volume` to `total_disposition_mcf`
- [ ] Where the data corresponds to a line in the original form, we should add "This field is reported as Line X.X in EIA Form 176" to the field description.

src/pudl/metadata/resources/eia176.py
```diff
@@ -31,5 +31,46 @@
...
         "etl_group": "eia176",
-    }
+    },
+    "core_eia176__yearly_gas_disposition": {
+        "description": {
+            "additional_summary_text": (
+                "a company's natural and supplemental gas disposition withn the report state."
+            ),
+            "additional_source_text": "(Part6, Lines 9, 12-20)",
+        },
```
e-belfer, line 54: Flag for myself - I am happy to rewrite this description to conform to our standards, and will get back to this in my next round of reviews.

src/pudl/transform/eia176.py
MeadBarrel: You're right. I did the check manually and there are just exception cases, but yeah need to add an assertion.

src/pudl/metadata/fields.py
```diff
@@ -62,7 +62,8 @@
 FIELD_METADATA: dict[str, dict[str, Any]] = {
     "delivered_gas_heat_content_mmbtu_per_mcf": {
         "description": (
-            "The average annual heat content of gas delivered directly to consumers."
+            "The average annual heat content of gas delivered directly to consumers. "
+            "Note: operator 17617651KY has several values that are slightly higher than 1.2"
```
e-belfer: Ah apologies for not being clearer, I meant the `description` field of the `dbt` test: see `dbt/models/eia/out_eia__yearly_generators/schema.yml` for an example of what this looks like.

MeadBarrel:
> * There is one negative `losses_from_leaks_volume` currently, for 17617106KS in 2008. If the total field is in fact a sum of the other components minus the absolute value of this record, we should treat this as a sign error and flip it. 

This is indeed the case. Unfortunately `unaccounted_for_mcf` is missing for a lot of rows so we can't really assert that this is always the case.

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -410,6 +410,14 @@ _core_phmsagas__yearly_distribution_misc,2021,1444
 _core_phmsagas__yearly_distribution_misc,2022,1449
 _core_phmsagas__yearly_distribution_misc,2023,1440
 _core_phmsagas__yearly_distribution_misc,2024,1428
+core_eia176__yearly_gas_disposition,2016,133
```
e-belfer: I'm seeing closer to 2k rows per year partition when I materialize this table, and the row counts test is failing. This should get updated.

docs/release_notes.rst
```diff
@@ -46,6 +46,9 @@ run the full ETL locally. See below for all the details.
 New Data
 ^^^^^^^^
 
+ * contains cleaned company-wide natural gas disposition data from Part 6B
+   of the EIA 176 survey. See :pr:`4765`
```
e-belfer: Looks like something got clipped here. You'll want to add the release notes under the `v2025.12.x` header, rather than the already-released `v2025.11.0`

dbt/models/eia176/core_eia176__yearly_gas_disposition/schema.yml
```diff
@@ -0,0 +1,111 @@
...
+          - name: disposition_other_pipelines_mcf
+            data_tests:
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
+          - name: disposition_storage_operators_mcf
+            data_tests:
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
```
e-belfer, line 97: This is currently failing for one record, ID 17690211AL in 2014. I added a note in #4727, but we can adapt this test as follows for now.


```suggestion
          - name: disposition_storage_operators_mcf
            data_tests:
              - dbt_expectations.expect_column_min_to_be_between:
                  arguments:
                    min_value: 0.0
                  config:
                    error_if: ">1"
                  description: "This is currently failing for one record, ID 17690211AL in 2014."
```

dbt/models/eia176/core_eia176__yearly_gas_disposition/schema.yml
```diff
@@ -0,0 +1,111 @@
...
+          - name: operator_id_eia
+          - name: report_year
+          - name: operating_state
+          - name: delivered_gas_heat_content_mmbtu_per_mcf
+            data_tests:
+              - dbt_expectations.expect_column_values_to_be_between:
+                  row_condition: operator_id_eia != '17617651KY'
+                  arguments:
+                    min_value: 0.8
+                    max_value: 1.2
```
e-belfer, line 25: As discussed elsewhere, a lot of records fail this test. Let's mark this as a warning rather than a failure for now. I'll add a note to investigate this further in #4727, and I've suggested below some changes to further reduce the number of failing records to <300 (~0.5% of records)


```suggestion
          - name: delivered_gas_heat_content_mmbtu_per_mcf
            data_tests:
              - dbt_expectations.expect_column_values_to_be_between:
                  config:
                    error_if: ">300"
                  arguments:
                    min_value: 0.8
                    max_value: 1.2
                   
```

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
 # )
 
 FIELD_METADATA: dict[str, dict[str, Any]] = {
+    "delivered_gas_heat_content_mmbtu_per_mcf": {
+        "description": (
+            "The average annual heat content of gas delivered directly to consumers. "
+            "Note: operator 17617651KY has several values that are slightly higher than 1.2"
```
```suggestion
            "The average annual heat content of gas delivered directly to consumers. "
```

A lot of values don't meet this criteria now, so we can drop the caveat.

src/pudl/metadata/fields.py
e-belfer: Seems like we don't need this in the end, in which case it can get dropped,

src/pudl/transform/eia176.py
```diff
@@ -350,12 +346,216 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+    )
+    df = df.rename(
+        columns={
+            1400: "deliveries_out_of_state_volume",
+            1840: "disposition_to_other_volume",
+        }
+    )
+
+    df = _normalize_operating_states(core_pudl__codes_subdivisions, df)
+    df.dropna(subset=["operating_state"])
```
e-belfer: You aren't assigning this to anything currently, so the final dataframe has NA values in the state column.

src/pudl/transform/eia176.py
```diff
@@ -350,12 +346,216 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+        "lease_use_volume",
+        # 16
+        "returns_for_repress_reinjection_volume",
+        # 17
+        "losses_from_leaks_volume",
+        # 18
+        "disposition_to_distribution_companies_volume",  # 1
+        "disposition_to_other_pipelines_volume",  # 2
+        "disposition_to_storage_operators_volume",  # 3
+        "disposition_to_other_volume",  # 4 (check)
```
e-belfer: Does this still need to be checked?

src/pudl/transform/eia176.py
```diff
@@ -350,12 +346,216 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+            "disposition_other_pipelines_mcf",
+            "disposition_out_of_state_mcf",
+            "other_disposition_all_other_mcf",
+            "total_disposition_mcf",
+            "losses_mcf",
+            "unaccounted_for_mcf",
+        ]
+    ]
+
+    df["losses_mcf"] = df["losses_mcf"].abs()
```
e-belfer: I think we want to be much more targeted here to avoid accidentally bulk-converting large amounts of negative data in the future. Something like:


```suggestion
    df.loc[(df.operator_id_eia == "17617106KS")&(df.report_year==2008), "losses_mcf"] = df.loc[(df.operator_id_eia == "17617106KS")&(df.report_year==2008), "losses_mcf"].abs()
```

src/pudl/transform/eia176.py
```diff
@@ -350,12 +346,216 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+    )
+
+    df = _normalize_operating_states(core_pudl__codes_subdivisions, df)
+    df.dropna(subset=["operating_state"])
+    df = df.dropna(subset=keep, how="all")
+
+    # Replace 9999 values prior to 2012 with NA
+    df.loc[
+        (df["heat_content_of_delivered_gas_btu_cf"] == 9999)
+        & (df["report_year"] < 2012),
```
e-belfer:
```suggestion
        & (df["report_year"] <= 2012),
```

There are actually still ~500 values in 2012 that are exactly 9999.

e-belfer: We should also set 0 values in this column equal to NA (see discussion above).

dbt/models/eia176/core_eia176__yearly_gas_disposition/schema.yml
e-belfer: When I look at the values prior to processing, this condition is met following 2012, and seems more or less met prior to 2005, with 9999 values skewing the intermediate years: 

Image

I noted below that 2012 also seems to have ~580 9999 values that should get dropped.

When we look at all time the majority of data does seem to generally fall within this range: 

Image

Of the ~800 records that don't meet this condition, about 600 are exactly equal to zero. We should set these records to be equal to NA as well.

|   report_year |   count |
|--------------:|--------:|
|          2004 |     107 |
|          2005 |      65 |
|          2006 |      71 |
|          2007 |      58 |
|          2008 |      43 |
|          2009 |      35 |
|          2010 |      37 |
|          2011 |      27 |
|          2012 |      24 |
|          2013 |      34 |
|          2015 |      21 |
|          2016 |      20 |
|          2017 |      17 |
|          2018 |      10 |
|          2020 |       2 |
|          2023 |       2 |

The remaining ~300 values are genuinely baddies, so we can reconfigure the DBT test to warn us about them rather than error for now.

e-belfer: So in summary, action items for this column are:

- [ ] Drop 9999 records for all time stamps, not just the current range
- [ ] Also convert 0 records to pd.NA
- [ ] Config the test as is suggested above

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,150 @@
...
+    "operational_consumption_other_mcf": {
+        "description": (
+            "Volume of natural gas consumed as fuel by the operator for other purposes within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "operational_consumption_other_detail": {
+        "description": (
+            "Free-text detail describing the operator’s specified “other purposes” for operational natural-gas consumption, corresponding to the volume reported in operational_consumption_other_mcf"
```
e-belfer: Great description, thanks!

```suggestion
            "Free-text detail describing the operator’s specified “other purposes” for operational natural-gas consumption, corresponding to the volume reported in operational_consumption_other_mcf."
```

src/pudl/transform/eia176.py
```diff
@@ -350,12 +346,216 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+        "total_disposition_volume",
+        # 20
+        "unaccounted_for",
+    ]
+
+    primary_key = ["operator_id_eia", "report_year"]
+
+    df = _core_eia176__yearly_company_data.filter([*primary_key, *extras, *keep])
+
+    tl = raw_eia176__continuation_text_lines
```
e-belfer: Some short docstrings would be helpful here to explain what you're doing with these two lines as compared to the line below.

src/pudl/transform/eia176.py
```diff
@@ -350,12 +346,216 @@ def core_eia176__yearly_gas_disposition_by_consumer(
...
+    tl = tl.groupby([*primary_key, "line"]).agg("sum").reset_index()
+    tl = tl.pivot(index=primary_key, columns="line", values="volume_mcf").reset_index()
+    tl = tl.filter([*primary_key, 1400, 1840])
+
+    df = df.merge(tl, how="left")
```
e-belfer:
```suggestion
    df = df.merge(tl, how="left", validate="1:1")
```

Typically a feature we like to add in to avoid unexpected duplication during merging.

e-belfer (changes requested):
I dug into the `delivered_gas_heat_content_mmbtu_per_mcf` and left some recommendations. Otherwise, my comments are mainly minor DBT and documentation fixes. Once these are integrated, I'll push some description/documentation changes and this should probably be ready to merge.

Excited to bring in all these new fields, thanks for your work on this!

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,147 @@
...
+    "operational_consumption_compressors_mcf": {
+        "description": (
+            "Volume of natural gas consumed as fuel for the operator's own pipeline distribution or storage compressor use within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "operational_consumption_lng_vaporization_liquefaction_mcf": {
+        "description": (
+            "Volume of natural gas consumed by the operator for vaporization, liquefaction, and LNG fuel within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```
zaneselvans, line 91: Can we offer any clarification on what "and LNG fuel" means here.  It seems like this is gas consumed to provide the energy required for operating an LNG terminal or storage facility, but "LNG fuel" as a standalone description seems weird.

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,147 @@
...
+    "operational_consumption_other_mcf": {
+        "description": (
+            "Volume of natural gas consumed as fuel by the operator for other purposes within the report state. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "operational_consumption_other_detail": {
+        "description": (
+            "Free-text detail describing the operator’s specified “other purposes” for operational natural-gas consumption, corresponding to the volume reported in operational_consumption_other_mcf."
```
zaneselvans: How are the embedded double-quotes in this string not messing things up?

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,147 @@
...
+        "description": (
+            "Disposition to distribution companies within the report state that does not fall into one of the other reported categories in lines 10.1-17.0. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "disposition_other_pipelines_mcf": {
+        "description": (
+            "Disposition to other pipelines within the report state that does not fall into one of the other reported categories in lines 10.1-17.0. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
+            "(14.73 psia and 60° Fahrenheit)"
```
zaneselvans: Extra line here with duplicate measurement conditions.

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,147 @@
...
+            "Disposition to storage operators within the report state that does not fall into one of the other reported categories in lines 10.1-17.0. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "total_disposition_mcf": {
+        "description": (
+            "Total disposition within the report state, as reported by the operator. "
+            "This includes disposition to consumers which is reported in core_eia176__yearly_gas_disposition_by_consumer. "
+            "Note that the reported total disposition and the sum of values in this table and core_eia176__yearly_gas_disposition_by_consumer often don't match as it would be expected. "
```
zaneselvans, line 176: Given that this is an inter-table irregularity it might be helpful to highlight it in the errata section for the EIA-176 data source.

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,147 @@
...
+    "unaccounted_for_mcf": {
+        "description": (
+            "The difference between gas supply and disposition. A positive entry indicates supply in excess of accounted-for disposition. A negative entry indicates accounted-for disposition exceeds reported supply. This is calculated as the difference between Part 4 Line 7.0 and Part 6 Line 19.0, and is reported as Line 20.0 in the original form."
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "disposition_out_of_state_mcf": {
+        "description": (
+            "Total volume of the operator's deliveries across or to state lines or U.S. borders. This has been summed from the detailed data reported by each company on Line 14.0 of the original form in order to preserve the primary key of the table. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```
zaneselvans, line 192: Do we anticipate that detailed breakdown also being available in its own separate table as well?

src/pudl/metadata/fields.py
```diff
@@ -60,6 +60,147 @@
...
+        ),
+        "unit": "Mcf",
+        "type": "number",
+    },
+    "other_disposition_all_other_mcf": {
+        "description": (
+            "Other disposition within the report state that does not fall into one of the other reported categories in lines 10.1-17.0. This has been summed from the detailed data reported by each company on Line 18.4 of the original form in order to preserve the primary key of the table. Reference conditions for measurement are 14.73 psia and 60° Fahrenheit."
```
zaneselvans, line 200: As above, are we going to publish this as a separate detailed breakdown?

src/pudl/metadata/fields.py
e-belfer: Yes, although I don't want to reference it until it exists. I can add a TODO here though to make it easier to spot.

src/pudl/metadata/fields.py
e-belfer: Yes, although I don't want to reference it until it exists. I can add a TODO here though to make it easier to spot.

src/pudl/metadata/fields.py
e-belfer: Good catch, removed!

src/pudl/metadata/fields.py
e-belfer: I have no idea, but I've updated to single strings.

src/pudl/metadata/fields.py
e-belfer: The [EIA instructions](https://www.eia.gov/survey/form/eia_176/instructions.pdf) note:

> Report the volume of gas consumed as fuel in your company's operations:
> Space heat of your own facilities (12.1), new pipeline fill (12.2), pipeline distribution or storage compressor use (12.3), gas used for vaporization, liquefaction, and LNG fuel (12.4).

That's been the case since 2021. It'd take more research for me to better expand on this, so I'm going to leave it as is for now.

PR #4771, opened 2025-11-17
Citations from July-November 2025

krivard:
# Overview

## What did you change?

new cites!

* Master's thesis
* Two preprints from CO School of Mines
* One preprint from UWisconsin-Madison
* One preprint from Technische Universität München

## Documentation


zaneselvans:
Huh. This should bypass the integration tests and go straight in after unit / docs pass.  I wonder if it bounced because of cloudflare's issues somehow?

PR #4778, opened 2025-11-19
Add zenodo-data-release.yml workflow.

### LINKED ISSUE #4290: Catch Zenodo Bad Gateway 502 errors and retry uploads
# Overview

While publishing PUDL v2025.5.0 we found that [some uploads were failing](https://github.com/catalyst-cooperative/pudl/issues/4278#issuecomment-2895913407) due to a Bad Gateway 502 error, but that failure wasn't detected, leading to an incomplete data release, which was only caught through manual inspection.

A similar problem has come up [in the PUDL Archiver](https://github.com/catalyst-cooperative/pudl-archiver/issues/458).

# Success Criteria

- Bad Gateway responses are caught when uploading files to a data release on Zenodo
- We retry the upload after detecting a 502 error
- If retries are not successful, the error is raised and the release publication is marked as failing

### LINKED ISSUE #4775: Separate data distribution from nightly builds
# Overview

* Currently our nightly builds run the full ETL and a bunch of software and data validation tests, which generally make pretty good use of the beefy VM's hardware.
* However, once a build completes successfully, we use the same VM for data distribution. Especially in the case of uploading data to Zenodo, this is typically limited by the speed of the upload.
* This can result in the VM running for 1-2 hours while hardly using any resources at all, which is inefficient.
* Zenodo is also one of the flakiest services we rely on, so the upload often fails, or appears to have failed even when it's succeeded, so the information we get from doing that distribution as part of the nightly build is limited, and it would also be useful to be able to re-run the distribution of data from a successful nightly / release build to Zenodo without needing to run a whole new build.
* We also want to automate the updating of Zenodo data release metadata, which is currently a slightly tedious manual process.
* For all of these reasons, it makes sense to try and separate the process of creating a new PUDL Zenodo data release from the process of running the nightly builds, and allow it to be run either locally as a script, or from a GitHub action.
* If we wanted to take this one step further, we could also decouple the distribution of data to GCS and S3 from the nightly builds, and instead have a data distribution script that reads the build outputs from `gs://builds.catalyst.coop/` and distributes them to all the different places they need to go (GCS, S3, & Zenodo right now).
* This separation would let us create a data release based on nightly build outputs that have already been generated, assuming that we've already closed out the release notes and exactly what we want to release is already in a nightly build output.

## Success Criteria

* Zenodo data release is no longer generated as part of the nightly or release build process.
* Zenodo data release can be run as a local script, based on already available S3 outputs.
* Zenodo data release can be run as a GitHub action, based on already available S3 outputs.
* Zenodo data release can be run in metadata-only mode, to make development of automated metadata updates easier.
* Zenodo data release can be sent either to Sandbox (for nightly testing) or Production (for versioned releases) Zenodo servers.
* Sandbox auto-publishes. Production awaits human approval.
* A new `zenodo-data-release` GitHub action can be triggered automatically when the nightly builds pass.
* A new `zenodo-data-release` GitHub action can be triggered automatically when a release build passes.
* A new `zenodo-data-release` GitHub action can be triggered with `workflow_dispatch` if we need to re-run manually, and allows us to set the necessary parameters (sandbox/production, `git_ref`, metadata-only, etc.)

## Tasks

* [x] Remove Zenodo data release related calls from the nightly build script.
* [x] Test that I can run `zenodo_data_release.py` successfully locally with some trivial files to upload.
* [x] Figure out arguments required to upload correct outputs from `nightly` build outputs.
* [x] If necessary, update the script so it can work with nightly build outputs.
* [x] Create GitHub workflow that can run `zenodo_data_release.py` with hard-coded arguments.
* [x] Give GitHub workflow ability to take external parameters and pass them in to `zenodo_data_release.py` appropriately.
* [ ] Test running a Zenodo sandbox data release from `nightly` outputs using `workflow_dispatch`.
* [ ] Test dispatching a trivial `zenodo-data-release` locally  using `curl`.
* [ ] Test dispatching a trivial `zenodo-data-release` from the build script using `curl`

zaneselvans:
# Overview

- Add a new `zenodo-data-release.yml` GitHub workflow that calls the `zenodo_data_release.py` script.
- Stop running the `zenodo_data_release.py` script in the nightly builds and instead dispatch the new workflow using `curl` and the GitHub API after the build outputs have been successfully distributed to S3.
- Move the `zenodo_data_release.py` script out of `devtools/zenodo` and into `src/pudl/scripts/` so it can be imported and tested and is available on the user's `$PATH`.
- Refactor `retry_request`, `create_bucket_file`, and `sync_directory` to restart the upload of a file from the beginning after encountering an error, and to do retries on common transient error statuses like 502 Bad Gateway.
- Added tests for the retry behavior.
- Disabled uploading of build outputs to S3 and GCS for `workflow_dispatch` builds, since we immediately delete those outputs anyway, the uploads have been quite robust, and there are egress fees associated with sending the data to S3.
- Update to `actions/checkout@v6` while we're in here.
- Cut the nightly build VM back down to 8 vCPUs and 64 GB of RAM since it's working again, and cheaper per run than the bigger machine (and not that much slower).

Closes #4775
Closes #4290 

## Documentation

## Testing

- I'm not really sure how to test how the code responds to this kind of intermittent network flakiness. There's an LLM generated test module but I can't evaluate whether it's testing things meaningfully so we should maybe discard it if someone else really knows how to test this kind of thing.
- I've tested behavior locally by running the script and then turning my wifi on and off to simulate a flaky network, and it has been able to do catch errors and do retries successfully.

## Tasks / Testing

- [x] Run the new workflow using `workflow_dispatch` to make sure it works with trivial data.
- [x] Quiet the logging about ignored files (set to `DEBUG`)
- [x] Run the new workflow using `workflow_dispatch` with a full data release's worth of data.
- [x] Try dispatching the workflow locally using `curl` in the same way as we would in the build script.
- [x] Add appropriate `nightly` dispatch of `zenodo-data-release` workflow to build script.
- [x] Add appropriate `stable` dispatch of `zenodo-data-release` workflow to build script.
- [x] Ensure that if the build is NOT successful that we DO NOT do a Zenodo release.
- [x] Re-open or rewind the "file" (`IOStream`) during retry, otherwise we upload zero bytes.
- [x] #4290 
- [x] Run a `workflow_dispatch` build with a real ETL and see what happens.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -12,6 +12,7 @@ env:
   PUDL_OUTPUT: /home/runner/pudl-work/output/
   PUDL_INPUT: /home/runner/pudl-work/input/
   DAGSTER_HOME: /home/runner/pudl-work/dagster_home/
+  ZENODO_SANDBOX_TOKEN_PUBLISH: ${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }}
 
 jobs:
   change_filter:
--- a/test/integration/console_scripts_test.py
+++ b/test/integration/console_scripts_test.py
@@ -1,5 +1,6 @@
 """Test the PUDL console scripts from within PyTest."""
 
+import os
 from pathlib import Path
 
 import geopandas as gpd
@@ -120,3 +121,37 @@ def test_resource_description(script_runner, resource_id: str):
         ["resource_description", "-n", resource_id], print_result=True
     )
     assert ret.success
+
+
+@pytest.mark.script_launch_mode("inprocess")
+@pytest.mark.skipif(
+    "ZENODO_SANDBOX_TOKEN_PUBLISH" not in os.environ,
+    reason="Zenodo sandbox token required to exercise uploads",
+)
+def test_zenodo_data_release(script_runner, tmp_path: Path):
+    """Round-trip upload a tiny release directory to the Zenodo sandbox.
+
+    The CLI accepts any fsspec-compatible path. Using ``Path.as_uri()`` yields a
+    ``file://`` prefix so the integration test exercises that codepath while
+    keeping all artifacts confined to pytest's temporary directory management.
+    """
+
+    source_dir = tmp_path / "release"
+    source_dir.mkdir()
+    (source_dir / "README.txt").write_text("Test release contents", encoding="utf-8")
+    (source_dir / "data.csv").write_text("value\n1\n", encoding="utf-8")
+    (source_dir / "metadata.json").write_text("{}\n", encoding="utf-8")
+
+    result = script_runner.run(
+        [
+            "zenodo_data_release",
+            "--env",
+            "sandbox",
+            "--source-dir",
+            source_dir.as_uri(),
+            "--publish",
+        ],
+        print_result=True,
+    )
+
+    assert result.success


docker/gcp_pudl_etl.sh
```diff
@@ -318,12 +291,22 @@ elif [[ "$BUILD_TYPE" == "workflow_dispatch" ]]; then
     DISTRIBUTION_BUCKET_SUCCESS=${PIPESTATUS[0]}
     # Remove those uploads since they were just for testing.
     remove_dist_path "$BUILD_ID" | tee -a "$LOGFILE"
-    # Remove individual parquet outputs and distribute just the zipped parquet
-    # archives on Zenodo, due to their number of files limit
-    rm -f "$PUDL_OUTPUT"/*.parquet
-    # push a data release to Zenodo sandbox
-    zenodo_data_release "$ZENODO_TARGET_ENV" 2>&1 | tee -a "$LOGFILE"
-    ZENODO_SUCCESS=${PIPESTATUS[0]}
+    curl -sS -X POST \
```
krivard:
```suggestion
    # launch zenodo release GHA
    curl -sS -X POST \
```

.github/workflows/zenodo-data-release.yml
```diff
@@ -5,26 +5,31 @@ on:
...
+        description: "Path to PUDL data to publish:"
         required: true
         default: "s3://pudl.catalyst.coop/nightly/"
       ignore_regex:
-        description: "A regular expression matching filenames not to upload"
+        description: "Ignore files matching this regex:"
         required: false
         default: '(^.*\.parquet$|^.*\.zip$)'
+        # The real data release pattern
+        # default: '(^.*\.parquet$|pudl_parquet_datapackage.json$)'
```
krivard: what's this for?

zaneselvans: I just want to test that things work with a few small files before telling it to upload 20 GB of stuff, which will take a couple of hours. But I wanted to make a note of what the real pattern should be.

docker/gcp_pudl_etl.sh
zaneselvans: I need to add `curl` calls to both of the other scenarios in here, but wanted to test that it actually does what I want first.

.github/ISSUE_TEMPLATE/nightly_build_failure.md
```diff
@@ -23,7 +23,7 @@ Once you've applied any necessary fixes, make sure that the nightly build output
 - [ ] [S3 distribution bucket](https://s3.console.aws.amazon.com/s3/buckets/pudl.catalyst.coop?region=us-west-2&bucketType=general&prefix=nightly/&showversions=false) was updated at the expected time
 - [ ] [GCP distribution bucket](https://console.cloud.google.com/storage/browser/pudl.catalyst.coop/nightly;tab=objects?project=catalyst-cooperative-pudl) was updated at the expected time
 - [ ] [GCP internal bucket](https://console.cloud.google.com/storage/browser/builds.catalyst.coop) was updated at the expected time
-- [ ] [Zenodo sandbox record](https://sandbox.zenodo.org/doi/10.5072/zenodo.5563) was updated to the record number in the logs (search for `zenodo_data_release.py` and `Draft` in the logs, to see what the new record number should be!)
+- [ ] [Zenodo sandbox record](https://sandbox.zenodo.org/doi/10.5072/zenodo.5563) was updated to the record number reported near the end of the [associated zenodo-data-release GitHub Action logs](https://github.com/catalyst-cooperative/pudl/actions/workflows/zenodo-data-release.yml)
```
zaneselvans, line 26: This was no longer accurate, since the zenodo data release call won't show up in the build logs.

.github/workflows/zenodo-data-release.yml
```diff
@@ -57,16 +62,20 @@ jobs:
           environment-name: pudl-dev
           cache-environment: true
 
+      - name: Install PUDL
+        run: |
+          pip install --no-deps --editable .
```
zaneselvans, line 67: We need to install the PUDL package now, since the script is installed by it.

.github/workflows/zenodo-data-release.yml
```diff
@@ -57,16 +62,20 @@ jobs:
...
+
       - name: Do a Zenodo data release
         env:
           ZENODO_SANDBOX_TOKEN_PUBLISH: ${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }}
           ZENODO_TOKEN_UPLOAD: ${{ secrets.ZENODO_TOKEN_UPLOAD }}
         run: |
-          python devtools/zenodo/zenodo_data_release.py \
+          zenodo_data_release \
```
zaneselvans, line 74: Switch to using the installed version of the script.

docker/gcp_pudl_etl.sh
```diff
@@ -106,19 +106,6 @@ function upload_to_dist_path() {
...
-    if [[ "$1" == "production" ]]; then
-        ~/pudl/devtools/zenodo/zenodo_data_release.py --no-publish --env "$1" --source-dir "$PUDL_OUTPUT" --ignore "$PUDL_OUTPUT"/pudl_parquet_datapackage.json
-    elif [[ "$1" == "sandbox" ]]; then
-        ~/pudl/devtools/zenodo/zenodo_data_release.py --publish --env "$1" --source-dir "$PUDL_OUTPUT" --ignore "$PUDL_OUTPUT"/pudl_parquet_datapackage.json
-    else
-        echo "Invalid Zenodo environment"
-        exit 1
-    fi
-}
-
```
zaneselvans, line 121: byeeeee!

docker/gcp_pudl_etl.sh
```diff
@@ -202,10 +188,36 @@ function clean_up_outputs_for_distribution() {
...
+}
+JSON
+)
+}
+
 ########################################################################################
 # MAIN SCRIPT
 ########################################################################################
 LOGFILE="${PUDL_OUTPUT}/${BUILD_ID}.log"
+ZENODO_IGNORE_REGEX="(^.*\\\\.parquet$|^.*\\\\.zip$)"
```
zaneselvans: The quadruple backslash seems kind of nuts, but we have two layers of escaping: once for the shell, and once for the JSON.

docker/gcp_pudl_etl.sh
```diff
@@ -216,7 +228,6 @@ WRITE_DATAPACKAGE_SUCCESS=0
 CLEAN_UP_OUTPUTS_SUCCESS=0
 DISTRIBUTION_BUCKET_SUCCESS=0
 DEPLOY_EEL_HOLE_SUCCESS=0
-ZENODO_SUCCESS=0
```
zaneselvans, line 219: Note that we will no longer get a Zenodo status directly from the builds. But we will get it from the workflow, and it'll also be sent to the `pudl-deployments` channel.

docker/gcp_pudl_etl.sh
```diff
@@ -282,12 +293,13 @@ if [[ "$BUILD_TYPE" == "nightly" ]]; then
...
-    # push a data release to Zenodo sandbox
-    zenodo_data_release "$ZENODO_TARGET_ENV" 2>&1 | tee -a "$LOGFILE"
-    ZENODO_SUCCESS=${PIPESTATUS[0]}
+    if [[ $DISTRIBUTION_BUCKET_SUCCESS == 0 ]]; then
+        zenodo_data_release \
+            "sandbox" \
+            "s3://pudl.catalyst.coop/nightly/" \
+            "${ZENODO_IGNORE_REGEX}" \
+            "publish" 2>&1 | tee -a "$LOGFILE"
+    fi
```
zaneselvans, line 299: If distribution was successful, pull from nightly builds outputs, upload to sandbox, and publish automatically.

docker/gcp_pudl_etl.sh
```diff
@@ -299,16 +311,17 @@ elif [[ "$BUILD_TYPE" == "stable" ]]; then
...
     # We can only do this on the GCS bucket, not S3
     gcloud storage --billing-project="$GCP_BILLING_PROJECT" objects update "gs://pudl.catalyst.coop/$BUILD_REF/*" --temporary-hold 2>&1 | tee -a "$LOGFILE"
     GCS_TEMPORARY_HOLD_SUCCESS=${PIPESTATUS[0]}
+    if [[ $DISTRIBUTION_BUCKET_SUCCESS == 0 ]]; then
+        zenodo_data_release \
+            "production" \
+            "s3://pudl.catalyst.coop/stable/" \
+            "${ZENODO_IGNORE_REGEX}" \
+            "no-publish" 2>&1 | tee -a "$LOGFILE"
+    fi
```
zaneselvans, line 321: If distribution to S3 was successful, pull from the new stable outputs, upload to production, and leave it a draft.

I used `stable` here initially since it'll always be the same, but maybe that opens it up to a kind of error, where somehow the build ref hasn't actually been created or stable hasn't actually been updated. So I think I'll point it at `${BUILD_REF}` instead.

docker/gcp_pudl_etl.sh
```diff
@@ -319,12 +332,18 @@ elif [[ "$BUILD_TYPE" == "workflow_dispatch" ]]; then
...
+    # has no build-specific outputs to publish. It just attempts to republish the
+    # nightly outputs from s3://pudl.catalyst.coop/nightly/ to make sure that the
+    # process is working.
+    if [[ $DISTRIBUTION_BUCKET_SUCCESS == 0 ]]; then
+        zenodo_data_release \
+            "sandbox" \
+            "s3://pudl.catalyst.coop/nightly/" \
+            "${ZENODO_IGNORE_REGEX}" \
+            "publish" 2>&1 | tee -a "$LOGFILE"
+    fi
```
zaneselvans, line 347: This could result in unexpected behavior (someone tries to do a `workflow_dispatch` and then doesn't see data update on the sandbox), but I don't see a way around it, since there's no durable output available from the `workflow_dispatch` and we don't want it to keep the VM running for another 1-2 hours to upload a transient output to the sandbox.

Another failure mode I don't think we can really avoid is what happens if more than one `workflow_dispatch` is running at the same time? They could both end up trying to write to Zenodo at the same time.  But this is a problem we already have and it hasn't blown up yet. And should be much less likely with actual releases.

src/pudl/scripts/zenodo_data_release.py
```diff
@@ -1,29 +1,69 @@
...
+    408,  # Request Timeout
+    429,  # Too Many Requests / rate limited
+    500,  # Internal Server Error
+    502,  # Bad Gateway
+    503,  # Service Unavailable
+    504,  # Gateway Timeout
+    520,  # Web server returned an unknown error (proxy)
+    522,  # Connection timed out (proxy)
+    524,  # A timeout occurred (proxy)
+}
```
zaneselvans, line 56: I asked the internet for what status codes are worth retrying, and this is what I got. Do these seem reasonable?

pyproject.toml
```diff
@@ -144,6 +144,7 @@ pudl_service_territories = "pudl.analysis.service_territory:pudl_service_territo
 dbt_helper = "pudl.scripts.dbt_helper:dbt_helper"
 pudl_null_cols = "pudl.scripts.pudl_null_cols:main"
 resource_description = "pudl.scripts.resource_description:main"
+zenodo_data_release = "pudl.scripts.zenodo_data_release:pudl_zenodo_data_release"
```
zaneselvans, line 147: Add this script to the list of those deployed by the package so they show up in the user's `$PATH`

test/unit/scripts/test_zenodo_data_release.py
zaneselvans, line 1: I don't really know how to meaningfully test this kind of sporadic network weirdness and our response to it. This is an LLM suggestion, but if you have real ideas that would be great.  I've tested it a bit locally by turning my wifi on and off during the upload, and it has successfully caught an error and recovered.

zaneselvans:
Docs build failure is due to intersphinx / network issues.

docker/gcp_pudl_etl.sh
krivard: GHA has some concurrency controls; would they help here?

https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

src/pudl/scripts/zenodo_data_release.py
krivard: Retrying on a rate limit seems rude, but the rest look fine

docker/gcp_pudl_etl.sh
zaneselvans: Ahh, great. I'll create two concurrency groups, one for each of the Zenodo environments.

src/pudl/scripts/zenodo_data_release.py
zaneselvans: Removed 429 from the list.

docker/gcp_pudl_etl.sh
```diff
@@ -107,16 +107,29 @@ function upload_to_dist_path() {
...
-    fi
+    ZENODO_ENV=$1
+    SOURCE_DIR=$2
+    IGNORE_REGEX=$3
+    PUBLISH=$4
+
+    # Trigger the zenodo data release workflow using the GitHub API
+    curl -sS -X POST \
+        -H "Accept: application/vnd.github+json" \
+        -H "Authorization: Bearer ${PUDL_BOT_PAT}" \
```
krivard, line 118: 🫳🤖

docker/gcp_pudl_etl.sh
```diff
@@ -319,12 +333,18 @@ elif [[ "$BUILD_TYPE" == "workflow_dispatch" ]]; then
...
-    # archives on Zenodo, due to their number of files limit
-    rm -f "$PUDL_OUTPUT"/*.parquet
-    # push a data release to Zenodo sandbox
-    zenodo_data_release "$ZENODO_TARGET_ENV" 2>&1 | tee -a "$LOGFILE"
-    ZENODO_SUCCESS=${PIPESTATUS[0]}
+    # NOTE: because we remove the test uploads and the zenodo data release workflow
+    # runs independent of the nightly build (it is not blocking), the workflow
+    # has no build-specific outputs to publish. It just attempts to republish the
+    # nightly outputs from s3://pudl.catalyst.coop/nightly/ to make sure that the
+    # process is working.
```
krivard, line 340: I'm not sure how to interpret this, particularly in the context of workflow dispatch. I have questions like:

- why does a workflow dispatch etl run care that the zenodo workflow runs independently of the nightly build?
- which workflow is looking for build-specific outputs to publish, the zenodo workflow or the build workflow?
- who is attempting to republish the nightly outputs, the zenodo workflow or this script?
- what process are we ensuring is working? why is this a good place to test that? (is this still a good place to test that, or is this leftover from development?)

Do you mean that there is no way to tell the zenodo workflow to upload the specific outputs from a workflow dispatch ETL? bc if so then it seems like we just...shouldn't run the zenodo workflow in this case.

docs/release_notes.rst
```diff
@@ -21,6 +21,13 @@ New Data Tests & Data Validations
 Bug Fixes
 ^^^^^^^^^
 
+* Increase the robustness of retries when uploading a PUDL data release to Zenodo,
+  catching common transient error status codes, and restarting file uploads from the
+  beginning of the file when an error occurs rather than uploading an empty file.
```
krivard: This sentence is ambiguous. Did you mean:

1. we use retries on all three of (uploading, catching, and restarting)
2. increasing robustness of retries, catching, and restarting are 3 separate-but-related improvements
3. catching and restarting are two examples of how we're increasing robustness of retries

for 1, consider:

> Increase the robustness of retries in three areas of Zenodo interactions: uploading a PUDL data release to Zenodo, catching common transient error status codes, and restarting file uploads from the beginning of the file when an error occurs rather than uploading an empty file.

for 2, consider:

> Increase the robustness of retries when uploading a PUDL data release to Zenodo, catch common transient error status codes, and restart file uploads from the beginning of the file when an error occurs rather than uploading an empty file.

for 3, consider:

> Increase the robustness of retries when uploading a PUDL data release to Zenodo, in two ways: catching common transient error status codes, and restarting file uploads from the beginning of the file when an error occurs rather than uploading an empty file.

docs/release_notes.rst
```diff
@@ -29,6 +36,13 @@ Quality of Life Improvements
   :ref:`out_eia__monthly_generators` and :ref:`out_eia__yearly_generators` tables, as
   multiple users have requested them. Most recently :user:`sam-hostetter` in issue
   :issue:`4772`. See :pr:`4776`.
+* Decoupl the publication of Zenodo data releases from the nightly and release builds
```
```suggestion
* Decouple the publication of Zenodo data releases from the nightly and release builds
```

src/pudl/scripts/zenodo_data_release.py
```diff
@@ -1,29 +1,67 @@
...
-"""Script to sync a directory up to Zenodo."""
+"""Upload a prepared PUDL data release directory to Zenodo.
+
+The PUDL data release process produces a directory of artifacts (zipped Parquet files,
+SQLite databases, JSON metadata, logs, etc.) that are uploaded to CERN's Zenodo data
+repository for long-term archival access. Each new versioned release of PUDL is
+associated with the same original PUDL concept DOI.
+
+This module provides a CLI that handles the process of uploading a new version to
+Zenodo, given a prepared directory of artifacts, typically produced by the PUDL builds.
```
krivard:
```suggestion
Zenodo, given a prepared directory of artifacts typically produced by the PUDL builds.
```

src/pudl/scripts/zenodo_data_release.py
```diff
@@ -1,29 +1,67 @@
...
+
+Uses state objects to ensure that Zenodo API calls happen in a valid order.
+
+Reads the files to upload via ``fsspec`` and stages remote files locally one at a time
+so uploads can be retried, without using excessive local disk space.
+
+Implements resilient retries for all requests to recover from transient network issues
+and Zenodo server flakiness. Zero-byte uploads are prevented.
+
+NOTE: PUDL nightly build outputs are NOT suitable a Zenodo data release unless the
```
krivard:
```suggestion
NOTE: PUDL nightly build outputs are NOT suitable for a Zenodo data release unless the
```

test/unit/scripts/test_zenodo_data_release.py
krivard: The canonical way is to mock all network requests so that the test can supply a specific response code and verify the resulting behavior is what we want. This is what the LLM does below, using [the built-in `monkeypatch` fixture in pytest](https://docs.pytest.org/en/stable/how-to/monkeypatch.html) (the docs at that link include an example of mocking requests.get, which is nearly-exactly our use case). 

In the PUDL codebase we generally prefer `pytest-mock` but there's at least one other use of `monkeypatch`. [Here's an explainer about the differences between monkeypatch, pytest-mock, and unittest.mock](https://github.com/pytest-dev/pytest/issues/4576#issuecomment-449864333). If you want to verify that the zenodo code really does call requests.put multiple times, or verify that the data factory gets called fresh every time, you'll want* the snooping capabilities of pytest-mock. if "well it got to the 200 eventually and it didn't crash" is sufficient, then monkeypatch is fine.

* the LLM below manually tracks the calls to `requests` in the `calls` array, which ...works. pytest-mock will do all that for you though.

TL;DR if you want a learning opportunity, convert the below to pytest-mock. if you just want it done, leave it as monkeypatch, it's fine.

test/unit/scripts/test_zenodo_data_release.py
```diff
@@ -0,0 +1,90 @@
...
+
+    calls: list[bytes] = []
+
+    def fake_request(*, method, url, headers, data, stream, timeout):  # noqa: ARG001
+        assert method == "PUT"
+        payload = data.read()
+        calls.append(payload)
+        if len(calls) == 1:
+            raise requests.ConnectionError("boom")
+        return _fake_response(200)
```
krivard, line 144: this would be less laborious in pytest-mock, but it gets the job done

krivard (approved):
No blocking feedback; just a buncha documentation nits, some confusion on what kind of zenodo workflow we want to trigger on workflow dispatch etls, and an optional learning opportunity sidequest for comparative mocking studies

zaneselvans:
Experimenting with some other build configurations to see what works and how fast/slow.

### Reduced compute
* 8 vCPU
* 64GB RAM
* OMP_NUM_THREADS=4
* 4.83 hrs total
* 1.26 hrs of 100% CPU during imputations
* Subsequent `zenodo-data-release` took 24 min

### Extra beefy
* 16 vCPU
* 128GB RAM
* OMP_NUM_THREADS=8
* parallelized zipping of SQLite DBs
* 3.47 hrs
* ~1 hr of heavy imputation work, only 20 min at 100% CPU.
* Subsequent `zenodo-data-release` took 29 min

docker/gcp_pudl_etl.sh
zaneselvans: * With the arguments below we are telling to `zenodo-data-release` workflow to (re-)publish the current nightly build outputs to the Zenodo Sandbox.
* We cannot tell it to publish the build outputs associated with the current build, because they will not be persisted in S3. Only `nightly` and `stable` outputs get saved there.
* I guess we could point it at the `gs://builds.catalyst.coop` outputs, but those are different than what's availble in the `pudl.catalyst.coop` bucket (the files aren't zipped or cleaned up for distribution).
* We're testing that the build script can kick off the zenodo data release (to sandbox), which was useful in development. But I was thinking I would leave it in so a `workflow_dispatch` is still testing all the mechanics of a real build.
* However, given that there are actual data egress costs to send the data out to S3 and then immediately delete it, maybe that step (above) isn't really worthwhile. I put that in after to broke once. But it hasn't broken since, and it gets tested in every nightly build.
* We could still leave in the `zenodo-data-release` workflow kickoff test even if the data distribution step above was removed, since it's distributing the `nightly` data anyway.
* We could also change the `upload_to_dist_path` and `remove_dist_path` functions to take a "protocol" input and only write to `gs` in `workflow_dispatch` since there's no egress fees for transfers within GCS.

docs/release_notes.rst
zaneselvans: Okay, tried to clarify exactly what the improvement is.

zaneselvans:
* I updated the docstrings / notes to try and be clearer about what I changed and what the previous situation was.
* I rewrote the test module to use `pytest-mock` and to test both retryable and non-retryable status codes, as well as raising the error after `max_retries`

Issue #4779, opened 2025-11-20
Add data or integration test verifying that DuckDB & SQLite outputs are the same


zaneselvans:
# Overview

* In [this `ferc-xbrl-extractor` PR](https://github.com/catalyst-cooperative/ferc-xbrl-extractor/pull/388) @zschira added a simple integration test that checks to see if the SQLite and DuckDB outputs extracted fromt he FERC XBRL filings all contain the same tables and the same columns in each table.
* With the integration of `catalystcooperative.ferc_xbrl_extractor` v1.7.1 in PUDL, we're now going to be generating both SQLite and DuckDB outputs for all of the converted FERC XBRL data.
* For peace of mind before we switch over to using just the DuckDB outputs, we should add a data validation test to PUDL that verifies the same thing for all of the FERC XBRL databases that we generate.
* In addition we could check that the number of records in every table is the same, with very little overhead.
* We could do this pretty easily using DuckDB in an integration test. Pulling it into dbt seems like more complexity, and we eventually expect to stop outputting SQLite entirely so seems less worth it.
* We should make sure this check is run on the fast ETL outputs in CI and the full ETL outputs in the nightly builds.

PR #4782, opened 2025-11-20
Update ferc xbrl extractor

zaneselvans:
# Overview

* Update the [FERC XBRL Extractor](https://github.com/catalyst-cooperative/ferc-xbrl-extractor) version to produce both SQLite and DuckDB outputs from the XBRL sources.
* This will allow the [PUDL Data Viewer](https://data.catalyst.coop) to access these databases remotely, since DuckDB allows remote attachment.
* However, there are some changes in the SQLite outputs / downstream PUDL assets that result from the new version, which is unexpected an undesirable.
* In particular the row counts in `core_ferc1__yearly_sales_by_rate_schedules_sched304` and `out_ferc1__yearly_sales_by_rate_schedules_sched304` decreased by about 2% in 2021-2024 (the XBRL years).
* In this PR we need to understand what those changes are, why they're happening, and whether they're of concern.
* This means comparing the new SQLite outputs with the old ones from the nightly builds, once we get a fresh outpu.
* We should also probably address #4779 to ensure that the SQLite and DuckDB outputs are identical across all of the FERC XBRL DBs, so that we can have confidence when we deprecate the SQLite outputs we don't have any unexpected changes.

## What I Changed

- Added an integration test that compares the SQLite and DuckDB databases that get extracted from the FERC XBRL data sources and verifies that they are equivalent. It checks that they both have the same tables, that no tables are empty, and that the contents of the tables are identical (using `pd.testing.assert_frame_equal()`). However it doesn't check that dtypes are identical, since the dtypes that DuckDB + Pandas creates when reading from SQLite vs. DuckDB are not identical. On the full ETL outputs, running this takes 1 minute. It should only take ~10 seconds on the fast ETL outputs.

## To-do list

- [x] Check that the new FERC XBRL derived SQLite and DuckDB outputs are identical.
- [x] Add an integration test that will check this programmatically as part of CI / builds.
- [x] :x: Check that the new SQLite and old SQLite from the most recent nightly build are identical.
- [x] Characterize differences between old and new SQLite extractions.
- [x] Update release notes
- [x] Update FERC to SQLite extraction docs to note that we are experimenting with DuckDB outputs.
- [x] Release a new version of `catalystcoop.ferc_xbrl_extractor` based on `string-datetimes` branch.
- [x] Merge in EIA-860M changes from `main` to get fresh row-counts.
- [x] Run the full ETL locally.
- [x] Update row counts.
- [x] Kick off a `workflow_dispatch` build and make sure it passes / possibly update row counts again.
- [ ] Make sure all dbt tests pass locally.

### Followup

- [ ] Update Data Access docs to note the experimental availability of DuckDB outputs via S3 and provide an example of how to query the remote DB in python or DuckDB.

### DIFF
--- a/src/pudl/metadata/resources/ferc1_eia_record_linkage.py
+++ b/src/pudl/metadata/resources/ferc1_eia_record_linkage.py
@@ -95,6 +95,9 @@ meant for use as an input into the record linkage between FERC1 plants and EIA."
         "description": {
             "additional_summary_text": "all EIA generators with ownership integrated.",
             "usage_warnings": ["aggregation_hazard"],
+            "additional_primary_key_text": "There are a handful of NULL Utility IDs in this table. "
+            "The natural primary key would be: "
+            "['plant_id_eia, 'generator_id', 'utility_id_eia', 'report_date', 'ownership_record_type'].",
         },
         "schema": {
             "fields": [
@@ -131,15 +134,6 @@ meant for use as an input into the record linkage between FERC1 plants and EIA."
                 "fraction_owned",
                 "ownership_record_type",
             ],
-            # These columns identify unique records, but can't be used as a PK because
-            # some of the utility_id_eia values are NA.
-            # "primary_key": [
-            #    "plant_id_eia",
-            #    "generator_id",
-            #    "utility_id_eia",
-            #    "report_date",
-            #    "ownership_record_type",
-            # ],
         },
         "sources": ["eia860", "eia923"],
         "etl_group": "outputs",
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
   ####################################################################################
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.7
+    rev: v0.14.8
     hooks:
       - id: ruff-check
         args: [--fix]
--- a/environments/conda-linux-64.lock.yml
+++ b/environments/conda-linux-64.lock.yml
@@ -34,7 +34,7 @@ dependencies:
   - at-spi2-atk=2.38.0=h0630a04_3
   - at-spi2-core=2.40.3=h0630a04_0
   - atk-1.0=2.38.0=h04ea711_2
-  - attrs=25.4.0=pyh71513ae_0
+  - attrs=25.4.0=pyhcf101f3_1
   - aws-c-auth=0.9.1=h194c533_5
   - aws-c-cal=0.9.8=h346e085_0
   - aws-c-common=0.12.5=hb03c661_1
@@ -101,7 +101,7 @@ dependencies:
   - coverage=7.12.0=py313h3dea7bd_0
   - cpython=3.13.10=py313hd8ed1ab_100
   - crashtest=0.4.1=pyhd8ed1ab_1
-  - cryptography=46.0.3=py313hafb0bba_0
+  - cryptography=46.0.3=py313heb322e3_1
   - curl=8.17.0=h4e3cde8_0
   - cycler=0.12.1=pyhcf101f3_2
   - cyrus-sasl=2.1.28=hd9c7081_0
@@ -119,7 +119,7 @@ dependencies:
   - dbt-core=1.10.15=pyhd8ed1ab_0
   - dbt-duckdb=1.9.4=pyhd8ed1ab_0
   - dbt-extractor=0.6.0=py313h5c7d99a_1
-  - dbt-protos=1.0.397=pyhcf101f3_0
+  - dbt-protos=1.0.402=pyhcf101f3_0
   - dbt-semantic-interfaces=0.9.2=pyh3cfb1c2_0
   - dbus=1.16.2=h3c4dab8_0
   - debugpy=1.8.17=py313h5d5ffb9_0
@@ -143,9 +143,9 @@ dependencies:
   - exceptiongroup=1.3.1=pyhd8ed1ab_0
   - execnet=2.1.2=pyhd8ed1ab_0
   - executing=2.2.1=pyhd8ed1ab_0
-  - fastapi=0.123.5=h6c771bf_0
+  - fastapi=0.123.8=h0694acc_0
   - fastapi-cli=0.0.16=pyhcf101f3_1
-  - fastapi-core=0.123.5=pyhcf101f3_0
+  - fastapi-core=0.123.8=pyhcf101f3_0
   - filelock=3.20.0=pyhd8ed1ab_0
   - flask=3.1.2=pyhd8ed1ab_0
   - flask-cors=6.0.1=pyhe01879c_0
@@ -201,7 +201,7 @@ dependencies:
   - graphql-core=3.2.7=pyhcf101f3_0
   - graphql-relay=3.2.0=pyhd8ed1ab_1
   - graphviz=13.1.2=h87b6fe6_0
-  - greenlet=3.2.4=py313h7033f15_1
+  - greenlet=3.3.0=py313h7033f15_0
   - grpcio=1.73.1=py313h2b3948d_1
   - grpcio-health-checking=1.73.1=pyhe01879c_0
   - grpcio-status=1.73.1=pyhd8ed1ab_0
@@ -222,7 +222,7 @@ dependencies:
   - humanfriendly=10.0=pyh707e725_8
   - humanize=4.14.0=pyhd8ed1ab_0
   - hyperframe=6.1.0=pyhd8ed1ab_0
-  - hypothesis=6.148.5=pyha770c72_0
+  - hypothesis=6.148.6=pyha770c72_0
   - icu=75.1=he02047a_0
   - identify=2.6.15=pyhd8ed1ab_0
   - idna=3.11=pyhd8ed1ab_0
@@ -286,11 +286,11 @@ dependencies:
   - libarrow-acero=20.0.0=h635bf11_27_cpu
   - libarrow-dataset=20.0.0=h635bf11_27_cpu
   - libarrow-substrait=20.0.0=h3f74fd7_27_cpu
-  - libblas=3.11.0=3_h4a7cf45_openblas
+  - libblas=3.11.0=4_h4a7cf45_openblas
   - libbrotlicommon=1.1.0=hb03c661_4
   - libbrotlidec=1.1.0=hb03c661_4
   - libbrotlienc=1.1.0=hb03c661_4
-  - libcblas=3.11.0=3_h0358290_openblas
+  - libcblas=3.11.0=4_h0358290_openblas
   - libcrc32c=1.1.2=h9c3ff4c_0
   - libcups=2.3.3=hb8b1518_5
   - libcurl=8.17.0=h4e3cde8_0
@@ -307,19 +307,19 @@ dependencies:
   - libffi=3.4.6=h2dba641_1
   - libfreetype=2.14.1=ha770c72_0
   - libfreetype6=2.14.1=h73754d4_0
-  - libgcc=15.2.0=he0feb66_14
-  - libgcc-ng=15.2.0=h69a702a_14
+  - libgcc=15.2.0=he0feb66_15
+  - libgcc-ng=15.2.0=h69a702a_15
   - libgd=2.3.3=h6f5c62b_11
   - libgdal-core=3.12.0=habacd5f_0
-  - libgfortran=15.2.0=h69a702a_14
-  - libgfortran5=15.2.0=h68bc16d_14
+  - libgfortran=15.2.0=h69a702a_15
+  - libgfortran5=15.2.0=h68bc16d_15
   - libgl=1.7.0=ha4b6fd6_2
   - libgl-devel=1.7.0=ha4b6fd6_2
   - libglib=2.86.0=h1fed272_0
   - libglvnd=1.7.0=ha4b6fd6_2
   - libglx=1.7.0=ha4b6fd6_2
   - libglx-devel=1.7.0=ha4b6fd6_2
-  - libgomp=15.2.0=he0feb66_14
+  - libgomp=15.2.0=he0feb66_15
   - libgoogle-cloud=2.39.0=hdb79228_0
   - libgoogle-cloud-storage=2.39.0=hdbdcf42_0
   - libgrpc=1.73.1=h3288cfb_1
@@ -329,7 +329,7 @@ dependencies:
   - libjpeg-turbo=3.1.2=hb03c661_0
   - libjxl=0.11.1=h6cb5226_4
   - libkml=1.3.0=haa4a5bd_1022
-  - liblapack=3.11.0=3_h47877c9_openblas
+  - liblapack=3.11.0=4_h47877c9_openblas
   - liblzma=5.8.1=hb9d3cd8_2
   - libmicrohttpd=1.0.2=hc2fc477_0
   - libmpdec=4.0.0=hb9d3cd8_0
@@ -351,8 +351,8 @@ dependencies:
   - libspatialite=5.1.0=gpl_h2abfd87_119
   - libsqlite=3.51.1=h0c1763c_0
   - libssh2=1.11.1=hcf80075_0
-  - libstdcxx=15.2.0=h934c35e_14
-  - libstdcxx-ng=15.2.0=hdf11a46_14
+  - libstdcxx=15.2.0=h934c35e_15
+  - libstdcxx-ng=15.2.0=hdf11a46_15
   - libtasn1=4.20.0=hb03c661_1
   - libthrift=0.22.0=h454ac66_1
   - libtiff=4.7.1=h9d88235_1
@@ -546,7 +546,7 @@ dependencies:
   - rsa=4.9.1=pyhd8ed1ab_0
   - ruamel.yaml=0.18.16=py313h07c4f96_0
   - ruamel.yaml.clib=0.2.14=py313h07c4f96_0
-  - ruff=0.14.7=h813ae00_0
+  - ruff=0.14.8=h813ae00_0
   - ruff-lsp=0.0.62=pyhd8ed1ab_0
   - s2n=1.6.0=h8399546_1
   - s3fs=2025.10.0=pyhd8ed1ab_0
--- a/environments/conda-lock.yml
+++ b/environments/conda-lock.yml
@@ -1274,11 +1274,11 @@ package:
     manager: conda
     platform: linux-64
     dependencies:
-      python: ">=3.10"
-    url: https://conda.anaconda.org/conda-forge/noarch/attrs-25.4.0-pyh71513ae_0.conda
+      python: ""
+    url: https://conda.anaconda.org/conda-forge/noarch/attrs-25.4.0-pyhcf101f3_1.conda
     hash:
-      md5: c7944d55af26b6d2d7629e27e9a972c1
-      sha256: f6c3c19fa599a1a856a88db166c318b148cac3ee4851a9905ed8a04eeec79f45
+      md5: 537296d57ea995666c68c821b00e360b
+      sha256: c13d5e42d187b1d0255f591b7ce91201d4ed8a5370f0d986707a802c20c9d32f
     category: main
     optional: false
   - name: attrs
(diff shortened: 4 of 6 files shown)


zaneselvans:
Comparing the nightly SQLite outputs with those that are produced by this PR, I see that there are many cases where there the data types and contents (like string formatting of timestamps, int vs. float) of the columns are different. I suspect this is because of the switch to using DuckDB for both the SQLite and DuckDB outputs in `ferc-xbrl-extractor`. It's quite widespread across many different tables.

It seems like trying to debug it piecemeal would be complex / annoying and probably not valuable long term. We might want to either:
* Switch back to using SQLite for the SQLite outputs in the XBRL Extractor.
* Actually make the switch and integrate the DuckDB outputs into PUDL / discontinue the SQLite outputs.

## Examples

I wrote a script that looks at the old (db1) and new (db2) SQLite outputs, enumerating any columns with different dtypes in the database (from DuckDB's point of view) and grabs an example value to compare. The results:

* [ferc1_dtype_differences.csv](https://github.com/user-attachments/files/23731275/ferc1_dtype_differences.csv)
* [ferc2_dtype_differences.csv](https://github.com/user-attachments/files/23731286/ferc2_dtype_differences.csv)
* [ferc6_dtype_differences.csv](https://github.com/user-attachments/files/23731295/ferc6_dtype_differences.csv)
* [ferc60_dtype_differences.csv](https://github.com/user-attachments/files/23731305/ferc60_dtype_differences.csv)
* [ferc714_dtype_differences.csv](https://github.com/user-attachments/files/23731315/ferc714_dtype_differences.csv)

zschira:
@zaneselvans I did a spot check on some of the cases where columns change from `DOUBLE->BIGINT` by checking the type specified in the XBRL taxonomy. It seems like the new version is actually handling these cases correctly, while the old version was ending up with incorrect dtypes. As for the `Timestamp->Sting` changes, I think this is a pretty straightforward fix. I'll try to spend a bit of time this afternoon looking at this and see if I can get a fix together.

zaneselvans:
@zschira there were also a number of cases where I saw `VARCHAR->BIGINT` and the example values all did appear to be integers. I'm not sure why they were previously getting interpreted as strings. Maybe some pollution with non-numeric values? And most of the columns that had this particular dtype issue in the FERC 1 data are in schedule 304, which is the table that ended up breaking the data validations, so I'm suspicious that that's why somehow.

Though filtering the table/column combos with the `VARCHAR->BIGINT` change I'm not finding any values that shouldn't actually be cast as integers.

zaneselvans:
After updating to the new version of the FERC XBRL Extractor with clean data types, we've still got some row count discrepancies. Looking at them to see what the actual differences are, and whether we should be concerned.

### `core_ferc1__yearly_sales_by_rate_schedules_sched304`
| table_name                               |   expected_partition |   observed_partition |   expected_count |   observed_count |   diff_relative_to_expected |
|:-----------------------------------------|---------------------:|---------------------:|-----------------:|-----------------:|----------------------------:|
| core_ferc1__yearly_sales_by_rate_schedules_sched304 |                 2021 |                 2021 |            13348 |            13019 |                  -0.0246479 |
| core_ferc1__yearly_sales_by_rate_schedules_sched304 |                 2022 |                 2022 |            11726 |            11391 |                  -0.028569  |
| core_ferc1__yearly_sales_by_rate_schedules_sched304 |                 2023 |                 2023 |            11969 |            11618 |                  -0.0293258 |
| core_ferc1__yearly_sales_by_rate_schedules_sched304 |                 2024 |                 2024 |            12098 |            11761 |                  -0.0278558 |

### `core_ferc714__hourly_planning_area_demand`
| table_name                               |   expected_partition |   observed_partition |   expected_count |   observed_count |   diff_relative_to_expected |
|:-----------------------------------------|---------------------:|---------------------:|-----------------:|-----------------:|----------------------------:|
| core_ferc714__hourly_planning_area_demand |                 2021 |                 2021 |           796322 |           796321 |                -1.25577e-06 |
| core_ferc714__hourly_planning_area_demand |                 2024 |                 2024 |           781692 |           781691 |                -1.27928e-06 |

### `out_eia__yearly_plant_parts`
| table_name                  |   expected_partition |   observed_partition |   expected_count |   observed_count |   diff_relative_to_expected |
|:----------------------------|---------------------:|---------------------:|-----------------:|-----------------:|----------------------------:|
| out_eia__yearly_plant_parts |                 2011 |                 2011 |           170572 |           170566 |                -3.51758e-05 |

### `out_ferc1__yearly_sales_by_rate_schedules_sched304`
| table_name                               |   expected_partition |   observed_partition |   expected_count |   observed_count |   diff_relative_to_expected |
|:-----------------------------------------|---------------------:|---------------------:|-----------------:|-----------------:|----------------------------:|
| out_ferc1__yearly_sales_by_rate_schedules_sched304 |                 2021 |                 2021 |            13348 |            13019 |                  -0.0246479 |
| out_ferc1__yearly_sales_by_rate_schedules_sched304 |                 2022 |                 2022 |            11726 |            11391 |                  -0.028569  |
| out_ferc1__yearly_sales_by_rate_schedules_sched304 |                 2023 |                 2023 |            11969 |            11618 |                  -0.0293258 |
| out_ferc1__yearly_sales_by_rate_schedules_sched304 |                 2024 |                 2024 |            12098 |            11761 |                  -0.0278558 |

### `out_ferc714__hourly_planning_area_demand`
| table_name                               |   expected_partition |   observed_partition |   expected_count |   observed_count |   diff_relative_to_expected |
|:-----------------------------------------|---------------------:|---------------------:|-----------------:|-----------------:|----------------------------:|
| out_ferc714__hourly_planning_area_demand |                 2021 |                 2021 |           796322 |           796321 |                -1.25577e-06 |
| out_ferc714__hourly_planning_area_demand |                 2024 |                 2024 |           781692 |           781691 |                -1.27928e-06 |

zaneselvans:
## ✅ FERC 1 Sales by Rate Schedules (Schedule 304)

- All of the records that are missing from the new version of the data have only NA/NaN values in all of the numerical data columns.
- However, they have values in the `rate_schedule_description` column that look like they might be "section headers" for multi-row groups of related records like we often see in unstructured tables that are basically replicating a paper form / PDF meant for human consumption.
- This does not seem like it's related to the timestamp vs. string dtype changes.
- Maybe it's due to some `dropna()` behavior we have in the Schedule 304 ETL that only looks at the numerical columns?
- Yes, the numerical columns are the ones specified in `drop_invalid_rows` -- any time they have a non-NA value, the row should be retained. If they are all NA, then the row should be dropped.
- The new outputs seem to reflect the desired behavior.
- In the previous setup, the only specified invalid value was `np.nan` so I wonder if there were a handful of numeric values that ended up with `pd.NA` or some other null-ish value that resulted in them not being dropped?
- Adding the `rate_schedule_description` column to the set of columns which indicate a row should be retained when they contain a valid value increases the number of retained rows by more than a factor of 20x.
- If I expand the set of "invalid" values to include `pd.NA` and the empty string and `None` as we have for many of the other `drop_invalid_rows` parameters, with the `rate_schedule_description` column included there's still a huuuuge number of additional rows.
- Looking at the output, what's happening is that there are 10 different `rate_schedule_type` categories (residential, industrial, commercial, etc.) and 3 different `billing_status` values (billed, unbilled, and total), and every possible combination of them is represented within each homogeneous `rate_schedule_description` section. But each rate schedule at most corresponds to 3 rows (one for each `billing_status` within the apropriate `rate_schedule_type`)
- So rather than having 10,000 records (in the XBRL years) including all non-null `rate_schedule_description` values means that we get ~300,000 rows, ~290,000 of which have to end up being NA in the data columns.
- In the XBRL years at least the `rate_schedule_description` while each rate description is freeform, the table is pretty well structured. These aren't human-readable section headers.
- So I think the new behavior is expected and desirable.  It looks like the new dtypes mean that we're catching a handful of records that really should have been NA before, but weren't, because we were only using `np.nan` and no null values appropriate to non-float values, which some of the raw inputs weirdly were.
- To confirm that this dtype / NA value explanation is correct:
  - Re-run ETL for Schedule 304 but with more expansive NA values.
  - We should expect to see the same ~2% reduction in row counts for the XBRL years.
  - Using `invalid_values = `[np.nan, pd.NA, "", None]` and `ferc-xbrl-extractor` v1.6.0 I get exactly the same reduction in the number of rows in the XBRL years as with new version of `ferc-xbrl-extractor` and just `np.nan` so I'm confident this is the explanation.

## ✅ FERC 714 Hourly Planning Area Demand

- There are only 2 records that got lost. It looks like they're both last-hour-of-year (UTC).
- I suspect that these are due to some difference between the upstream vs. PUDL ETL methods of addresssing invalid hour-24 to next-day hour-00 conversions.
- After reverting the changes that were editing these timestamps, and going back to treating timestamps as strings (since some of them require cleaning to be valid) these changes have vanished.

## ✅ Plant Parts EIA

- All 6 rows that are no longer present with the new FERC XBRL extraction are related to `plant_id_eia==50973` and only in 2011.
- There are also a handful of records where the value of `ferc_acct_name` has changed between the nightly and the new outputs, affecting the `Other` and `Steam` categories.

```py
df_new.ferc_acct_name.value_counts()
```
```
ferc_acct_name
Other        3436213
Steam         831841
Hydraulic     741946
Nuclear        45097
Name: count, dtype: int64
```
```py
df_night.ferc_acct_name.value_counts()
```
```
ferc_acct_name
Other        3436197
Steam         831851
Hydraulic     741946
Nuclear        45097
Name: count, dtype: Int64
```

### Plant Parts Debugging Plan

* Materialize the `out_eia__yearly_generators_by_ownership` table with both the new and old FERC XBRL Extractor.
  * This is the table that the `ferc_acct_name` values get merged on to with `get_ferc_account_map` which reads from a CSV that we have checked into the repo.  Each combination of EIA technology description and prime mover code is associated with one of Other, Steam, Hydraulic, Nuclear.
  * Merge the old and new versions together on their shared PK and see what's different between them.
* Merging with a PK of `[report_date, plant_id_eia, generator_id, utility_id_eia, ownership_record_type]` it looks like they have the exact same set of records -- there are no instances that appear in one side and not the other.
* **However** there are a handful of different `ferc_acct_name` values. 2 additional `Other` and 2 fewer `Steam`
* Looking at the records that differ, they're all for a single generator:
  * `report_date`: 2011-01-01
  * `plant_id_eia`: 50973
  * `generator_id`: GN27
  * `utility_id_eia`: 12981
* The nightly vs. new outputs **also** have different prime movers! Which is why the `ferc_acct_name` got merged on differently. The PM switched from `ST` to `CA` for some reason.
* The heat rates, capacity factors, etc. are also different presumably because we're looking at outputs from different prime movers way upstream.

### Why different Prime Movers?

* Looking at `core_eia860__scd_generators` I see that in the new outputs, the PM is `CA` for every year except for 2010, which also happens to be the last year that the generator was in operation. In 2010 only, it is `ST`.
* Pulling the nightly version of the same table, **both** 2010 and 2011 have `ST` but all the other values are `CA`.
* Merging together the nightly and new versions of this table and checking for records where the PM code is not the same... this is *literally* the only one. WTF.
* Going back to the spreadsheets, I see that the file format transitions from XLS to XLSX between 2010 and 2011.
* Looking up the individual generator PM codes, it looks like 
  * 2010 50973 GN27 ST (operating)
  * 2011 50973 GN27 CA (retired) 
* So the NEW outputs appear to reflect the PM codes that are in the spreadsheet, and the OLD outputs seem to be inconsistent with the spreadsheet. Which is disturbing.
* Okay but why tf would this have anything to do with changes to the FERC XBRL extraction?!
* Under the hood XLSX is a zipped XML document, and upgrading to the new FERC XBRL extractor involved a major version bump to the XML libraries we have installed. But the Excel engine we're using is `calamine` which is written in rust and presumably doesn't depend on these Python & C XML libraries. And even if it did it would link its own version internally.
* Major dependency changes include:
  * `igraph` v0.10.16 --> v1.0.0 (high performance graph analysis)
  * `gdal` v3.11.4 --> v3.12.0 (geospatial stuff)
  * `harfbuzz` v11.5.1 --> v12.1.0 (text reshaping)
  * `lxml` v5.4.0 --> v6.0.2 (XML parsing)
  * `python-igraph` v0.11.9 --> v1.0.0 (python wrapper for igraph)
  * `rpds-py` v0.29 --> v0.30 (R to Python binding layer library)
* Maybe I've messed something up here and this is actually a consequence of the `calamine` update. Re-installing 0.5.3 locally and regenerating `core_eia860__scd_generators` to see if I get the same results.
* Nope. Same result regardless of the calamine version.
* Is this somehow a consequence of the harvesting process? If I look at `_core_eia860__generators` what does this generator look like?
  * Comparing new and old `_core_eia860__generators, **both** only have prime mover ST in 2010, not 2011. So the difference seems to be getting introduced in the harvesting process.
  * Looking at the only other place that `generator_id` and `prime_mover_code` show up together as inputs into harvesting, the `_core_eia923__generation` table, there is exactly 1 2011 record, and it has ST as the PM.  So this change is the result of the tie between these two inputs being resolved in a different way and is not of concern.

zaneselvans:
* @cmgosnell I didn't get to the root of the change in the plant parts, but I'm skeptical it has anything to do with the FERC XBRL Extractor directly. I suspect it's related to one of the dependencies that got a significant update in the process.
* The change seems to be happening somewhere in the harvesting process because the pre-harvested generators have the same prime movers on main and with the new XBRL extractor, but the `core_eia860__scd_generators` table differs between the two.
* This is the one and only example of a `prime_mover_code` changing between the old and new setups, and the new behavior reflects what's in the input spreadsheets, while the old behavior did not. So I'm inclined to go ahead and merge.
* If you think that's reasonable then I'll create a new version of the XBRL extractor and update the dependencies and row counts in here.

src/pudl/transform/params/ferc1.py
```diff
@@ -4941,7 +4941,7 @@
         },
         "drop_invalid_rows": [
             {
-                "invalid_values": [np.nan],
+                "invalid_values": [np.nan, pd.NA, "", None],
```
zaneselvans, line 4944: These NULL values result in the same behavior regardless of the version of `ferc_xbrl_extractor` so I thought I'd just go ahead and leave them in place.

zaneselvans:
After much sisyphean digging, it looks like this prime mover code flip is due to a tie between `ST` and `CA` for that generator in 2011 -- there's just one record for each, in the EIA-860 generators table and in the EIA-923 generation table, and previously it was being resolved to ST, but now it's being resolved to CA. We always take the most-common prime mover code -- there's no minimum consistency threshold -- so both of these are valid outcomes. And it seems like some mysterious change in the dependencies or cosmic rays flipped it, and it has nothing to do with the FERC XBRL (which makes sense).  So this is good to go.

dbt/models/sec10k/core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities/schema.yml
```diff
@@ -5,10 +5,13 @@ sources:
...
-          - check_row_counts_per_partition:
+          - dbt_expectations.expect_table_row_count_to_be_between:
+              description: Nightly builds and local ETL yield slightly different output
+              config:
+                enabled: "{{ target.name == 'etl-full'}}"
               arguments:
-                table_name: core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities
-                partition_expr:
+                min_value: 5543
+                max_value: 5564
```
zaneselvans, line 14: For a month or so the row counts on this table had stabilized, but now they're different local and remote again, so back to the range.

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -2315,7 +2315,6 @@ core_pudl__codes_imputation_reasons,,12
 core_pudl__codes_subdivisions,,69
 core_pudl__entity_plants_pudl,,20744
 core_pudl__entity_utilities_pudl,,17303
-core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities,,5564
```
zaneselvans, line 2318: Migrated back to checking a range of row counts on this table.

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -2028,10 +2028,10 @@ core_ferc1__yearly_sales_by_rate_schedules_sched304,2017,10105
...
 core_ferc1__yearly_sales_by_rate_schedules_sched304,2019,11102
 core_ferc1__yearly_sales_by_rate_schedules_sched304,2020,11020
-core_ferc1__yearly_sales_by_rate_schedules_sched304,2021,13348
-core_ferc1__yearly_sales_by_rate_schedules_sched304,2022,11726
-core_ferc1__yearly_sales_by_rate_schedules_sched304,2023,11969
-core_ferc1__yearly_sales_by_rate_schedules_sched304,2024,12098
+core_ferc1__yearly_sales_by_rate_schedules_sched304,2021,13019
+core_ferc1__yearly_sales_by_rate_schedules_sched304,2022,11391
+core_ferc1__yearly_sales_by_rate_schedules_sched304,2023,11618
+core_ferc1__yearly_sales_by_rate_schedules_sched304,2024,11761
```
zaneselvans, line 2034: These ~2% reductions are expected; see thread in main PR comments.

dbt/seeds/etl_full_row_counts.csv
```diff
@@ -2950,7 +2949,7 @@ out_eia__yearly_plant_parts,2007,155277
 out_eia__yearly_plant_parts,2008,161998
 out_eia__yearly_plant_parts,2009,169358
 out_eia__yearly_plant_parts,2010,163994
-out_eia__yearly_plant_parts,2011,170566
+out_eia__yearly_plant_parts,2011,170572
```
zaneselvans, line 2952: Go back to using the row count from the nightly build, which appears to be consistent.

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -131,6 +131,15 @@
...
             ],
+            # These columns identify unique records, but can't be used as a PK because
+            # some of the utility_id_eia values are NA.
+            # "primary_key": [
+            #    "plant_id_eia",
+            #    "generator_id",
+            #    "utility_id_eia",
+            #    "report_date",
+            #    "ownership_record_type",
+            # ],
```
cmgosnell: can you add this info into `additional_primary_key_text` in the `? (go look at `out_ferc714__respondents_with_fips` for an example of where to put this and we are actively writing up better docs about how to add to the description)

docs/dev/run_the_etl.rst
```diff
@@ -240,6 +240,13 @@ You can adjust the years to process for each dataset using the Launchpad tab::
           years:
           - 2021
 
+.. note::
+
+    We are experimenting with producing DuckDB outputs from the XBRL (and possibly DBF)
+    data that FERC publishes, since DuckDB databases can be queried remotely, unlike
+    SQLite, have richer data types, and are generally smaller and much more performant.
+    For the time being, ``ferc_to_sqlite`` will produce both SQLite and DuckDB outputs
+    by default.
```
cmgosnell: non-blocking nit but in these instructional docs it feel important to be succinct/ i don't think it is necessary to have any of the why here... 
```suggestion
    We are experimenting with producing DuckDB outputs from the XBRL (and possibly DBF)
    data that FERC publishes. For the time being, ``ferc_to_sqlite`` will produce both SQLite
    and DuckDB outputs by default.
```

src/pudl/extract/xbrl.py
```diff
@@ -120,8 +125,8 @@ def convert_form(
     # we already delete the existing base on `clobber=True` in `xbrl2sqlite`
     run_main(
         filings=filings_archives,
-        db_path=sql_path,
-        clobber=False,
+        sqlite_path=sqlite_path,
+        duckdb_path=duckdb_path,
```
cmgosnell, line 129: did you mean to remove `clobber=False` here?

zaneselvans: Yes, the `clobber` option has been removed. It always clobbers now, since we never want to layer multiple extractions on top of each other in the same DB.

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
zaneselvans: Ah okay I thought there was some place to put something like that.

zaneselvans: Updated with the appropriate description metadata.

src/pudl/extract/xbrl.py
cmgosnell: ah okay i got confused but i didn't realize this `run_main` is not in this repo and i didn't see that change.

Issue #4773, opened 2025-11-17
Develop generic `duckdb` based CSV extractor


zschira:
### Overview
During our push to improve PUDL performance, a common pattern we've identified involves using duckdb to extract hourly data from CSV files partitioned by year or quarter. Given that we have now done this in multiple places (and identified other candidates conforming to this pattern) like EIA930, VCERARE, and EPA CEMS, we should develop generic tooling for this.

Issue #4734, opened 2025-11-04
`raw_vcerare__*` high memory usage [13.8GB]


jdangerx:
### Overview

This asset is marked as `"memory-usage": "high"`.

We have the "return a big dictionary and then read the big dictionary in several places" pattern, plus the `__all_dfs` asset uses 13.8GB on its own. Should be able to adapt the patterns from #4731 here.

Initial profiling stats:

```
Command line: ./memory_profile.py --directory profiles/ --aggregate -a raw_vcerare__all_dfs
Start time: Tue Nov 04 2025 16:44:16 GMT-0500 (Eastern Standard Time)
End time: Tue Nov 04 2025 16:49:16 GMT-0500 (Eastern Standard Time)
Duration: 0:05:00.304000
Total number of allocations: 87503
Total number of frames seen: 0
Peak memory usage: 13.8 GB
Python allocator: pymalloc
```

Hotspots are:
* pickle.load [3.0GB]
* pickle.dump [3.3GB]
* pd.concat in concat_pages [3.3GB]
* pd.assign in process_raw [3.3GB]

Haven't profiled the downstream guys but since it looks like the outputs are ~3.3GB in memory I would expect that those reads are also hotspots downstream.

Taking a look at the extraction, it seems possibly duckdb-able which should cut down on the memory usage by a lot.

Full flamegraph:

[memray-flamegraph-raw_vcerare__all_dfs-2025-11-04T21:44:16.html](https://github.com/user-attachments/files/23346264/memray-flamegraph-raw_vcerare__all_dfs-2025-11-04T21.44.16.html)

### Success Criteria

How will we know that we're done?

* [x] measured initial peak memory usage:
* [x] identified memory usage hotspots
* [ ] resolved or decided to give up on memory usage hotspots

### Next steps
* [x] profile this asset

PR #4676, opened 2025-10-23
Polarize CEMS ETL

zschira:
# Overview

This PR uses Polars to improve performance of the EPA CEMS ETL. It uses Polars' [Lazy API](https://docs.pola.rs/user-guide/concepts/lazy-api/) and [streaming engine](https://docs.pola.rs/user-guide/concepts/streaming/) for fast and memory efficient processing of CEMS data.

Theoretically, Polars should be capable of reading the raw CSVs, performing transform steps, and writing results to parquet all in one step, but due to our archive structure (CSVs wrapped up in zipfiles) we require an intermediate step to get the data out of the zipfiles and into a more accessible format on disk. If we didn't perform this intermediate step, we would be forced to do all of our processing in-memory. This implementation keeps the existing `graph_asset` structure to extract and write year-quarter partitioned parquet files to a sub-directory in the `$PUDL_OUTPUT/parquet/` directory. It then streams from this parquet dataset to perform all transforms and writing to a monolithic parquet file in one step.

## What problem does this address?

This improves the memory usage and high runtime for CEMS processing. On my machine, these changes bring processing time down to <10 minutes, but peak memory usage is still pretty high. This may be a result of Polars intentionally using all available memory in its parallelized processing. We should perform further testing on a variety of computers to see if Polars is capable of automatically scaling to the available resources.

## What did you change?

* I updated the `pudl.{extract/transform}.epacems` modules to utilize the Polars Lazy API.
* Update the Parquet IO Manager to deal with both Pandas dataframes and Polars Lazyframes
* Substituted Polars for Dask in the few places where we were using Dask to deal with EPA CEMS
* Removed Dask as a dependency.
* Removed unused `pudl.analysis.epacamd_eia` and `pudl.output.epacems` modules.

Closes #4663

## Documentation

Make sure to update relevant aspects of the documentation:

- [x] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html): reference the PR and related issues.
- [x] Review and update any other aspects of the documentation that might be affected by this PR.

# Testing

Execute the `core_epacems__hourly_emissions` asset from the dagster UI and do some basic exploration / spot checking output table vs nightly version of table.

## To-do list

- [x] Test ETL on multiple devices
- [x] Do more data validation / regression testing
- [x] Update downstream asset to use Polars
- [x] Update all docstrings to reflect usage of Polars
- [x] Test simplified CEMS asset in GitHub CI
- [x] Test simplified CEMS asset in branch deployment
[...]

PR #4783, opened 2025-11-21
Improve `vcerare` memory performance

### LINKED ISSUE #4679: `out_vcerare__hourly_available_capacity_factor` high memory usage
### Overview

This asset is marked as `"memory-usage": "high"`.

### Success Criteria

* [ ] measured initial peak memory usage:
* [ ] identified memory usage hotspots
* [ ] resolved or decided to give up on memory usage hotspots

### Next steps
* [ ] profile this asset

zschira:
# Overview

Closes #4679, #4773, and #4734.

This PR attempts to reduce peak memory usage of `vcerare` extract and transform operations.

## Results
Getting an accurate measure of memory usage has proven incredibly difficult due to the way `duckdb`/`polars` allocate memory and what `memray` sees/reports. I often see peak memory usage as being higher than total available memory on my computer in the first place. I've tried to get a rough sense for usage by looking to a few different sources including, `memray` outputs, watching `htop` while running the ETL, and running on my lower powered laptop.

From these tests, my best guess is that the extract is now <5GB peak usage, while the transform is somewhere in the 10-12GB range. My estimate for the extract comes from the `memray` and `htop` outputs. `memray` shows the `heap` usage will continue rise throughout the run, but `resident` memory never goes above ~4GB, and this is roughly consistent to what I see in `htop`. As for the transform, I really don't have any quantitative metrics to look to, but I ran it on my laptop with 16GB of RAM, and it ran totally fine while also doing other things in the background.

I think both `polars` and `duckdb` will also use more memory than necessary if it's available, particularly when you have CPU capacity for them to do lots of internal parallelization, so I think results will vary greatly by machine and what's running at the time.

## What did you change?
This PR adds a new Dagster resource called `PudlParquetTransformer`, which can be used to offload raw/intermediate tables to disk as parquet files. This allows `duckdb`/`Polars` to operate on this data out-of-memory. It then converts the `extract`/`transform` steps of the `vcerare` ETL to use this new tooling.

The new resource also has a method called `extract_csv_to_parquet`, which uses `duckdb` to extract CSVs in a way that should be compatible with other CSV extractions, like that used for EIA930.

# Testing
I have a custom regression testing notebook to compare to nightly builds, because the current regression testing script will run out of memory.

## To-do list

- [ ] Get good profiling results

### DIFF
--- a/src/pudl/extract/vcerare.py
+++ b/src/pudl/extract/vcerare.py
@@ -75,7 +75,7 @@ def extract_vcerare(
             dataset="vcerare",
             partitions=partitions,
             pages=VCERARE_PAGES.keys(),
-            ds=context.resources.datastore,
+            datasore=context.resources.datastore,
             zip_path=Path(f"{year}/"),
         ):
             # Collect ParquetData objects for each year/page combo


src/pudl/resources.py
```diff
@@ -58,3 +64,135 @@ def datastore(init_context) -> Datastore:
         # it should be initialized to the right values.
         ds_kwargs["local_cache_path"] = PudlPaths().input_dir
     return Datastore(**ds_kwargs)
+
+
+class PudlParquetTransformer(ConfigurableResource):
```
zschira: This is a Dagster `Resource` that can be used anywhere, and provides utility functions for offloading data to parquet files and using those files as the basis for intensive transforms. It will write files to `${PUDL_WORKSPACE}/storage/parquet` where it creates a subdirectory for each table that contains parquet files.

src/pudl/transform/vcerare.py
```diff
@@ -445,50 +381,87 @@ def one_year_hourly_available_capacity_factor(
...
-        df_name: _spot_fix_great_lakes_columns(df)
+    [
+        _spot_fix_great_lakes_columns(df)
         .pipe(_check_for_valid_counties, fips_df_census, df_name)
         .pipe(_add_time_cols, df_name)
         .pipe(_drop_city_cols, df_name)
         .pipe(_make_cap_fac_frac, df_name)
         .pipe(_stack_cap_fac_df, df_name)
         .pipe(_clip_unexpected_2016_pv_capacity, df_name, year)
+        .pipe(transformer.offload_table, f"_core_vcerare__{df_name}", year=year)
```
zschira: Rather than returning a dict of large `DataFrame`s, this will offload these intermediate results to parquet where we can do the expensive process of combining these into a single table.

src/pudl/transform/vcerare.py
```diff
@@ -503,29 +476,39 @@ def _get_year(df, year):
...
+            raw_vcerare__onshore_wind_power_100m=transformer.get_df(
+                "raw_vcerare__onshore_wind_power_100m",
+                filters=[("report_year", "=", year)],
+            ),
+        )
+
+        # Merge table into final output table
+        merge_all_vce_tables(transformer, partitioned_output_table, year)
+
+    return transformer.get_lf(partitioned_output_table)
```
zschira: Let the parquet IO-manager deal turning the partitioned parquet files into one monolithic parquet file in the output directory.

src/pudl/resources.py
```diff
@@ -58,3 +64,135 @@ def datastore(init_context) -> Datastore:
...
+
+    def _get_parquet_name(self, table_name: str, **partitions: dict) -> str:
+        base_path = self._get_parquet_dir(table_name)
+        if partitions is not None:
+            filename = "_".join(map(str, partitions.values())) + ".parquet"
+        else:
+            filename = "monolithic.parquet"
+        return str(base_path / filename)
+
+    def extract_csv_to_parquet(
```
zschira: I designed this method with the EIA 930 CSV extraction in mind, and should work with other large hourly tables that are distributed as CSVs.

e-belfer: How does this relate to the methods implemented in #4676 for CEMS processing?

src/pudl/resources.py
```diff
@@ -58,3 +64,135 @@ def datastore(init_context) -> Datastore:
...
+        """Write data from DataFrame or LazyFrame to disk as a parquet file.
+
+        Offloading data to disk allows Polars and duckdb to perform highly efficient
+        transforms.
+        """
+        parquet_path = self._get_parquet_name(table_name, **partitions)
+        if isinstance(df, pd.DataFrame):
+            df.to_parquet(parquet_path)
+        else:
+            df.sink_parquet(parquet_path, engine="streaming")
```
e-belfer: This would be for Polars dataframes?

src/pudl/workspace/setup.py
```diff
@@ -44,6 +45,11 @@ def output_dir(self) -> Path:
         """Path to PUDL output directory."""
         return Path(self.pudl_output).absolute()
 
+    @property
+    def parquet_transform_dir(self) -> Path:
+        """Path to directory with parquet files used for intermediate transforms."""
+        return self.input_dir.parent / "storage" / "parquet"
```
e-belfer, line 51: Confirmed this is the storage folder inside `dagster_home`.

src/pudl/resources.py
e-belfer: From discussion, sounds like the plan is to eventually reform 930 and CEMS to use a similar process for extraction, ideally using a shared class or some common abstractions.

src/pudl/extract/vcerare.py
```diff
@@ -13,165 +13,62 @@
...
+    required_resource_keys={
+        "pudl_parquet_transformer",
+        "dataset_settings",
+    },
+)
+def extract_vcerare(context):
+    """Extract data from all vcerare pages and write to parquet files."""
+    dataset_settings = context.resources.dataset_settings
+    pudl_parquet_transformer = context.resources.pudl_parquet_transformer
+    pudl_parquet_transformer.extract_csv_to_parquet(
```
jdangerx: The extraction here is a bit different from that in EIA 930, so instead of making something that we think might work for both and stuffing it in a shared location, we should probably just inline the extraction.

Or, if you want to pull *something* into a shared function, I think you've correctly identified that there's a lot of similarity between the CSV gathering logic here and in EIA 930, CEMS, etc.

I think that the kernel of that similarity is "transposing from `{partition: {page: df}}` to `{page: {partition: df}`}" - and after that we get a bunch of divergence based on what we actually do with the dataframes (or lazyframes, etc.)

If we make this "partitions-to-pages" function:

```python
def get_pages_from_partitions(table_name, pages, partitions) -> {page_name: memory_efficient_pointer_to_data_on_disk}
```

Then we could call it in this function pretty cleanly:

```python
pages = get_pages_from_partitions(...)
processed = {page_name: handle_page(page_data) for page_name, page_data in pages.items()}
# return `processed` if you want PudlParquetIOManager to handle things, else write to disk manually
```

I think it's probably easiest to just inline the code though :wink:

src/pudl/resources.py
```diff
@@ -58,3 +64,135 @@ def datastore(init_context) -> Datastore:
...
+                before writing to parquet.
+        """
+        # Loop through zipfiles and translate data from CSV to parquet
+        for partitions, path in partition_paths:
+            with (
+                duckdb.connect() as conn,
+                self.ds.get_zipfile_resource(dataset=dataset, **partitions) as zf,
+            ):
+                for table_name, csv_name in table_name_map.items():
+                    with zf.open(str(path / csv_name)) as csv:
```
jdangerx: fwiw: `zf.open` puts everything in memory, `zf.extract` extracts to disk and then you can read into duckdb with memory-mapping the file instead of actually loading everything in - had to use `extract` in the 930 code.

src/pudl/resources.py
```diff
@@ -58,3 +64,135 @@ def datastore(init_context) -> Datastore:
...
+
+    def _get_parquet_dir(self, table_name: str) -> Path:
+        """Get path to directory for writing/reading parquet files."""
+        parquet_path = PudlPaths().parquet_transform_dir / table_name
+        parquet_path.mkdir(exist_ok=True, parents=True)
+        return parquet_path
+
+    def offload_table(
+        self, df: pd.DataFrame | pl.LazyFrame, table_name: str, **partitions: dict
+    ) -> Path:
```
jdangerx: note from call:

Have this spit out a `ParquetReader` or something that is a Path which knows how to be read as a DF, LF, or duckdb.

Overall this class seems to be doing too many things, we could split it into:
* thing that knows how to write stuff to parquet without you having to fuss with partition-specific paths and finding the dagster storage directory on your own
* thing that knows how to read a parquet file back into pandas/polars/duckdb
* thing that knows how to read CSVs for the VCERARE data (which, as above, we'll largely inline)

src/pudl/transform/vcerare.py
jdangerx, line 1: note from call: duckdb seemed to be choking on the "merge everything at once" which would have the nicest flow. Probably not worth re-opening the duckdb can of worms here, polars lazy frames are probably good enough.

Ideally we'd get a better idea of RAM usage for these two but seems like our profiling tools aren't super well set up to understand how tools like duckdb use memory in a way that's actually constraining our runs.

src/pudl/helpers.py
```diff
@@ -2327,3 +2332,152 @@ def standardize_phone_column(df: pd.DataFrame, columns: list[str]) -> pd.DataFra
         df[column] = df[column].mask(invalid_mask, pd.NA)
 
     return df
+
+
+class ParquetData(BaseModel):
```
zschira, line 2337: I now have this class as well as a few helper functions that either take it as a parameter or return one/more instance of `ParquetData`.

src/pudl/helpers.py
```diff
@@ -2327,3 +2332,152 @@ def standardize_phone_column(df: pd.DataFrame, columns: list[str]) -> pd.DataFra
...
+        """Get name of an individual parquet file corresponding to a single partition of data."""
+        if self.partitions != {}:
+            filename = "_".join(map(str, self.partitions.values())) + ".parquet"
+        else:
+            filename = f"{self.table_name}.parquet"
+        return self.parquet_directory / filename
+
+
+def offload_table(
+    table_data: pd.DataFrame | pl.LazyFrame | duckdb.DuckDBPyRelation,
```
zschira, line 2366: This is basically the same as the previous `offload_table` method, but I've also modified it to allowed passing a `duckdb` relation.

src/pudl/helpers.py
```diff
@@ -2327,3 +2332,152 @@ def standardize_phone_column(df: pd.DataFrame, columns: list[str]) -> pd.DataFra
...
+        use_all_partitions: If true read the entire directory of parquet files.
+            Otherwise only read data from the partition specified in parquet_data.
+    """
+    with duckdb.connect() as conn:
+        if use_all_partitions:
+            yield conn.read_parquet(f"{parquet_data.parquet_directory}/*.parquet")
+        yield conn.read_parquet(str(parquet_data.parquet_path))
+
+
+def duckdb_extract_zipped_csv(
```
zschira, line 2448: This is a new simplified helper function for extracting CSVs from zips. I think it has a constrained enough amount of functionality to be useful in other cases, but could also move this the `pudl.extract.vcerare` if it still feels too specific.

src/pudl/transform/vcerare.py
```diff
@@ -503,29 +481,37 @@ def _get_year(df, year):
...
+            raw_vcerare__offshore_wind_power_140m=df_from_parquet(
+                raw_vcerare__offshore_wind_power_140m[year],
+            ),
+            raw_vcerare__onshore_wind_power_100m=df_from_parquet(
+                raw_vcerare__onshore_wind_power_100m[year],
+            ),
+        )
+
+        # Merge table into final output table
+        merged_table = merge_all_vce_tables(
```
zschira: This overwrites `merged_table` every iteration because we call `lf_from_parquet` with `use_all_partitions=True` at the end. This works because each instance of `merged_table` is a `ParquetData` with `partitions` set to point to a single year, but when we set `use_all_partitions=True` it will create a `LazyFrame` reading from all years.

src/pudl/helpers.py
zaneselvans: There are also some extraction modules which aren't data source specific, e.g. `pudl/extract/csv.py` seems like it might be a good home.

zschira: This function is pretty tightly coupled to the other helper functions I added here (which aren't all related to CSV extraction), and I think it would be confusing to split them up.

src/pudl/transform/vcerare.py
jdangerx: Might be cleaner/more explicit to do `return lf_from_parquet(ParquetData(table_name=table_name), use_all_partitions=True)` and not have this overwriting.

src/pudl/helpers.py
jdangerx: Might be nice to mention `parquet` somewhere in the name.... `persist_table_as_parquet`?

src/pudl/helpers.py
```diff
@@ -2301,3 +2306,152 @@ def standardize_phone_column(df: pd.DataFrame, columns: list[str]) -> pd.DataFra
...
+
+
+class ParquetData(BaseModel):
+    """Wrap table data offloaded as parquet file(s) on disk.
+
+    Writing data to disk as parquet files enables the use of highly efficient
+    processing/transforms with tools like Polars or duckdb. This class provides
+    helpers for managing paths to parquet data on disk. This class and the helper
+    methods that operate on it are meant explicitly for intermediate data to use
+    for transforms, not for output data.
```
jdangerx: What makes this class only suitable for intermediate outputs & not for output data, beyond the hard-coded `.parquet_transform_dir`? Might be nice to flesh that out in the docstring.

src/pudl/helpers.py
zschira: Makes sense

src/pudl/helpers.py
zschira: I think I'll actually end up using this to write to output parquet files for the EQR work,  so I've removed this comment.

PR #4787, opened 2025-11-25
Update git auth in nightly build to work with actions/checkout@v6

zaneselvans:
# Overview

* Last night the builds failed because they couldn't update the `nightly` branch to point at the tag associated with the new nightly build tag.
* The branch update seems to have died after doing `git config --unset http.https://github.com/.extraheader`
* I'm not 100% sure but it seems like it might be related to [this change](https://github.com/actions/checkout/issues/2313) in how Git authentication is set up automatically in the new `actions/checkout@v6` release that hit the nightly builds for the first time last night.
* Locally I get an exit code of 5 (which is what the builds reported when I try to `git config unset` a configuration that isn't set at all. So I'm guessing that this config is no longer part of how the authentication is set up.
* It might be that there's some other auth we need to override... but it also sounds like now the authentication information is stored outside of the checked-out repo? So maybe we don't? This was always slightly hacky black magic to me.
* Unfortunately this fix will only be exercised when  the nightly build runs again.
* Also, use `set +x` to avoid printing out some tokens in the logs. Oops.

## Tasks 

- [x] Regenerate `PUDL_BOT_PAT` and update organization secret.
- [x] Update the `nightly` branch to point at the commit tagged with `nightly-2025-11-25`.
- [ ] Run a branch deployment to check if the other non-nightly branch update changes break anything?

### DIFF
--- a/docker/gcp_pudl_etl.sh
+++ b/docker/gcp_pudl_etl.sh
@@ -3,8 +3,10 @@
 # This script won't work locally because it needs adequate GCP permissions.
 
 function send_slack_msg() {
-    echo "sending Slack message"
-    curl -X POST -H "Content-type: application/json" -H "Authorization: Bearer ${SLACK_TOKEN}" https://slack.com/api/chat.postMessage --data "{\"channel\": \"C03FHB9N0PQ\", \"text\": \"$1\"}"
+    set +x &&
+        echo "sending Slack message" &&
+        curl -X POST -H "Content-type: application/json" -H "Authorization: Bearer ${SLACK_TOKEN}" https://slack.com/api/chat.postMessage --data "{\"channel\": \"C03FHB9N0PQ\", \"text\": \"$1\"}" &&
+        set -x
 }
 
 function authenticate_gcp() {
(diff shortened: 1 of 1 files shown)


docker/gcp_pudl_etl.sh
```diff
@@ -166,10 +170,7 @@ function notify_slack() {
 function merge_tag_into_branch() {
     TAG=$1
     BRANCH=$2
-    # When building the image, GHA adds an HTTP basic auth header in git
-    # config, which overrides the auth we set below. So we unset it.
-    git config --unset http.https://github.com/.extraheader &&
```
zaneselvans, line 171: This `git config --unset` fails with exit code 5 (which the builds got) if the value it's unsetting isn't set to begin with... so I'm hoping all we have to do is remove it and the auth will work.

docker/gcp_pudl_etl.sh
```diff
@@ -112,13 +114,14 @@ function zenodo_data_release() {
...
     PUBLISH=$4
 
-    # Trigger the zenodo data release workflow using the GitHub API
-    curl -sS -X POST \
-        -H "Accept: application/vnd.github+json" \
-        -H "Authorization: Bearer ${PUDL_BOT_PAT}" \
-        https://api.github.com/repos/catalyst-cooperative/pudl/actions/workflows/zenodo-data-release.yml/dispatches \
-        -d @<(
-            cat <<JSON
+    set +x &&
```
zaneselvans, line 117: I forgot to use `set +x` here so I probably need to go generate a new `PUDL_BOT_PAT` since it showed up in the logging output. (done)

PR #4788, opened 2025-11-26
Integrate EIA-860m 2025-oct

cmgosnell:
# Overview

Closes #XXXX.

## Documentation

Make sure to update relevant aspects of the documentation:

- [ ] Update the [release notes](https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html): reference the PR and related issues.
- [ ] Update relevant Data Source jinja templates (see `docs/data_sources/templates`).
- [ ] Update relevant table or source description metadata (see `src/metadata`).
- [ ] Review and update any other aspects of the documentation that might be affected by this PR.

# Testing

How did you make sure this worked? How can a reviewer verify this?

## To-do list

- [x] Run `make pytest-coverage` locally to ensure that the merge queue will accept your PR.
- [x] Review the PR yourself and call out any questions or issues you have.
- [ ] For PRs that change the PUDL outputs significantly, run the full ETL locally and then [run the data validations](https://catalystcoop-pudl.readthedocs.io/en/nightly/dev/data_validation.html) using dbt. If you can't run the ETL locally then run the `build-deploy-pudl` GitHub Action manually and ensure that it succeeds.
- [ ] If updating analyses or data processing functions: make sure to update row count expectations in `dbt` tests.

PR #4791, opened 2025-12-01
Update row counts that changed due to calamine upgrade

zaneselvans:
# Overview

The update to `python-calamine` v0.6.0 resulted in some material changes to the contents of the data we're extracting from Excel spreadsheets, including these row counts. But I wonder what other changes there might be.

| table_name                   | expected_partition   | observed_partition   |   expected_count |   observed_count |   diff_relative_to_expected |
|:-----------------------------|:---------------------|:---------------------|-----------------:|-----------------:|----------------------------:|
| core_eia923__entity_coalmine |                      |                      |             4501 |             4650 |                   0.0331038 |

| table_name                        | expected_partition   | observed_partition   |   expected_count |   observed_count |   diff_relative_to_expected |
|:----------------------------------|:---------------------|:---------------------|-----------------:|-----------------:|----------------------------:|
| core_pudl__assn_ferc1_pudl_plants |                      |                      |             8101 |             8100 |                -0.000123442 |

# Testing

- I switched back to `python-calamine==0.5.3` and rematerialized the above assets and everything upstream of them, and re-checked the row counts, and they were back to the expected values.

## To-do list

### DIFF
--- a/dbt/seeds/etl_full_row_counts.csv
+++ b/dbt/seeds/etl_full_row_counts.csv
@@ -1179,7 +1179,7 @@ core_eia861__yearly_utility_data_rto,2021,1564
 core_eia861__yearly_utility_data_rto,2022,1592
 core_eia861__yearly_utility_data_rto,2023,1621
 core_eia861__yearly_utility_data_rto,2024,1635
-core_eia923__entity_coalmine,,4501
+core_eia923__entity_coalmine,,4650
 core_eia923__monthly_boiler_fuel,2008,77537
 core_eia923__monthly_boiler_fuel,2009,79002
 core_eia923__monthly_boiler_fuel,2010,81197
@@ -2275,7 +2275,7 @@ core_phmsagas__yearly_distribution_operators,2024,1428
 core_pudl__assn_eia_pudl_plants,,18650
 core_pudl__assn_eia_pudl_utilities,,16760
 core_pudl__assn_ferc1_dbf_pudl_utilities,,411
-core_pudl__assn_ferc1_pudl_plants,,8101
+core_pudl__assn_ferc1_pudl_plants,,8100
 core_pudl__assn_ferc1_pudl_utilities,,432
 core_pudl__assn_ferc1_xbrl_pudl_utilities,,282
 core_pudl__assn_ferc714_csv_pudl_respondents,,213


zaneselvans:
## Further investigations

### `core_pudl__assn_ferc1_pudl_plants`

* There's hardly any processing that goes into `core_pudl__assn_ferc1_pudl_plants` and it only changed by one line. We compile this data. Identify that change.
* One of the plant names that we have mapped includes the unprintable ASCII control character`\x02` which means "start of text".
* The new version of `calamine` looks like it turns this into whitespace or strips it out entirely.
* This results in two mapped plant names becoming identical (`coit #2 peaking`) so one of them gets dropped.
* This seems like desirable / correct behavior?

### `core_eia923__entity_coalmine`

Really there is no good reason for this table to have been split out of FRC -- the coalmine information is messy and poorly processed, and should really come directly from MSHA and get merged in here based on the mine ID if we want it.

* The main difference here seems to be that with the new `calamine` there are many more records that contain valid `county_id_fips` values, but I'm not sure why.
* There don't seem to be any FIPS codes that contain weird unprintable characters.
* I notice that **all** and **only** the records with `data_maturity=="incremental_ytd"` have values for `mine_id_msha` -- all the `final` records have NA values.
* There are 726 records that only appear in the new data.
* There are 577 that only appear in the old data.
* Most of these differences are because an old record lacks a FIPS ID, and the new record has it.
* The remaining differences appear to result from the old data having only the version with an NA FIPS code, and the new data having both the NA version and the version with the FIPS code.
* Still not clear why this is happening... Is it because of strange non-printable characters in the county names? Seems like no -- for the coalmines in particular, we get a 3-digit county FIPS ID, and just use `add_fips_ids()` to convert the state abbreviation into a FIPS code, which we concatenate with the county FIPS code. So there's not a lot of potential for weird messy strings.
* However, overall the new data is better than the old data -- there are more coalmine records that have FIPS codes associated with them.
* Looking at the original spreadsheets for 2015 (the year the NA FIPS codes come from for `mine_id_msha==103150`) I see that it has numerical county FIPS code `73` while in 2010 it has a string FIPS code `073` (which is correct).
* In the transform for the table we are trying to fix the bad FIPS codes by zero padding to a width of 3, but the numeric 73 isn't even making it that far with the old `calamine`.
* ~If I install the new `calamine` and re-run the extraction, I get the same NA values in the same records.~ I was being a dumbass and now fully regenerating the raw asset. The raw extracted data **is** different depending on what version of `python-calamine` is installed, thank goodness.

zaneselvans:
## Calacoalmines Digging

### Raw Dataframes

* ✅ Produce `raw_eia923__fuel_receipts_costs` with both `python-calamine` v0.5.3 and v0.6.1
* ✅ Compare those raw inputs **THEY SHOULD BE DIFFERENT**
* ✅ Specifically look at the `littleton` coal mine, which we've observed different outputs for.
* ✅ How do the FIPS and MSHA Mine codes vary across years? Compare 2010 and 2015.
* ✅ Should have FIPS code `073` (2010) or `73` (2015) (based on spreadsheet)
  * All of them are correctly getting zero-padded to `073`
* ✅ Should have MSHA Mine ID `0103150` or `103050`
  * In the raw dataframe, these all show up as floats `103050.0`

### Processed Data

* ✅ Produce `core_eia923__entity_coalmine` with both `python-calamine` v0.5.3 and v0.6.1
* ✅ Select all records associated with `littleton` coal mine
* ✅ Expect to see an additional record from `calamine` v0.5.3, with `county_id_fips == pd.NA` What record would produce the NA value?
* Note that in the original spreadsheets, littleton has MSHA Mine IDs, but in the outputs they all seem to be NA! Why are they getting lost?

## Summarize

* Regenerating the coalmines asset from scratch including all upstream bits, we see that the difference is being introduced (as expected) during the extraction.
* The new version of calamine somehow seems to be capturing FIPS codes that are getting nulled by the old version.
* Looking at the spreadsheets:
  * 2010 FIPS codes for the `LITTLETON` mine are strings "073" and are retained by both versions.
  * 2015 FIPS codes for the `LITTLETON` mine are numeric "73" (somehow this ends up being set to NA)
  * There don't appear to be any non-standard / unprintable characters in the 2015 spreadsheet cells.
  * This kind of data type / int vs. string vs. float thing is pervasive in Excel spreadsheets. If it was a problem, I would have thought we'd have seen it many times before now.

zaneselvans:
I tried re-generating the `raw_eia860__plant` dataframes with both versions of calamine to look at whether the ZIP codes had a similar issue, since they also often need to have a leading zero and get mixed up as strings, integers, and floats.  However, both versions read exactly the same number of non-NA zip codes, and there appear to be no values that are read in as strings with leading zeroes.  There are a mix of ints and floats and zip codes that should have leading zeroes just have fewer digits.

zaneselvans:
The row-count updates in this PR are superseded by #4788 and we are somewhat apprehensively declare this outcome acceptable, even though we don't really know why `calamine` was not picking up all those FIPS codes before, and now it is, since it's better now than it was, and we haven't identified any other cases where this is happening.

PR #4809, opened 2025-12-04
Fix descriptions in code dictionary

krivard:
# Overview

Use new built descriptions properly in code dictionary pages.

We have [a code dictionary page in the "Data Dictionaries" section of the docs site](https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/codes_and_labels.html) that was pulling in the raw Python dictionary instead of rendering the description:

> image

## What did you change?

* When the code dictionary is built, look up the rendered description and use the first section (which is always the summary sentence, if it exists, or the empty string, if it does not)

## Documentation

# Testing

Ran `make docs-clean` locally and observed a successful result and proper table descriptions in the resulting HTML page.

## To-do list

### DIFF
--- a/src/pudl/metadata/resources/eia.py
+++ b/src/pudl/metadata/resources/eia.py
@@ -1218,9 +1218,22 @@ RESOURCE_METADATA: dict[str, dict[str, Any]] = {
     f"out_eia__{freq}_generators": {
         "description": {
             "additional_summary_text": "all generator attributes including calculated capacity factor, heat rate, fuel cost per MMBTU and fuel cost per MWh.",
-            "additional_details_text": """These calculations are based on
-the allocation of net generation reported on the basis of plant, prime mover and energy
-source to individual generators. Heat rates by generator-month are estimated by using
+            "additional_details_text": f"""This table includes all {freq} attributes
+for all generators reported to EIA-860 and EIA-923.
+
+To provide a complete picture of generator data, this table compiles data from many
+different EIA tables, including directly reported, imputed and calculated columns.
+We suggest using this table if you want to explore {freq} attributes about generators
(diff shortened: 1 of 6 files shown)


src/pudl/metadata/classes.py
```diff
@@ -529,7 +529,10 @@ def to_rst(
         )
         rendered = template.render(
             Encoder=self,
-            description=RESOURCE_METADATA[self.name]["description"],
+            # just get the resolved resource summary & drop all the other sections of the description
+            description=PUDL_PACKAGE.get_resource(self.name).description.partition(
+                "\n\n"
+            )[0],
```
krivard, line 535: We can use PUDL_PACKAGE here because this code does not run until well after PUDL_PACKAGE is guaranteed to exist.

e-belfer: Does the description ever not contain `\n\n` in it? If so, we'll expect this to return the entire string, so would just be helpful to confirm!

e-belfer (approved):
One peace of mind question about rendering in the description field, as splitting on characters always feels a bit brittle. Otherwise, this rendered as expected for me when running `make docs-build`.

src/pudl/metadata/classes.py
krivard: Yep, always always. The description renders the jinja template [resource_description.rst.jinja](https://github.com/catalyst-cooperative/pudl/blob/ff62c8323c64771736754cb9f16904e079b7da73/src/pudl/metadata/templates/resource_description.rst.jinja), which delimits every section with a double newline -- including the transition between the summary and the first dlistitem, which is the one we're operating on with this statement.

PR #4811, opened 2025-12-05
Check uniqueness of natural primary keys which contain NULL values using dbt

zaneselvans:
# Overview

- Added data validation checks that verify the uniqueness of natural primary keys which we've identified, but which contain NULL values and therefore cannot be used as PKs in SQLite.
- I searched for those resource definitions that has `additional_primary_key_text` and if they explicitly stated what the natural primary key ought to be, added a check using `dbt_expectations.expect_compound_columns_to_be_unique`
- The `out_ferc714__respondents_with_fips` table had some duplicate values, so I set it to `severity: warn` so we don't forget about it, but it doesn't blow anything up right now.

# Testing

I ran the following on my local full ETL outputs:

```console
dbt_helper validate --select "test_name:*expect_compound_columns_to_be_unique*"
```

# To-do list

- [x] update release notes


dbt/models/phmsagas/_core_phmsagas__yearly_distribution_misc/schema.yml
```diff
@@ -15,7 +15,6 @@ sources:
           - dbt_expectations.expect_compound_columns_to_be_unique:
               arguments:
                 column_list:
-                  - report_date
```
zaneselvans, line 18: `report_date` was not listed in the expected natural PK in the metadata annotations, doesn't show up in the other tables, and the PK was unique without it, so I removed it here.

dbt/models/phmsagas/_core_phmsagas__yearly_distribution_excavation_damages/schema.yml
```diff
@@ -18,3 +18,9 @@ sources:
               arguments:
                 table_name: _core_phmsagas__yearly_distribution_excavation_damages
                 partition_expr: "EXTRACT(YEAR FROM report_date)"
+          - dbt_expectations.expect_compound_columns_to_be_unique:
+              arguments:
+                column_list:
+                  - report_id
+                  - damage_type
+                  - damage_sub_type
```
zaneselvans, line 26: The `additional_primary_key_text` for this table says that this should be the natural PK, but that it has duplicates and the table isn't cleaned.  However, it appeared to be unique as-is.  The metadata specifies a PK with many more columns.

dbt/models/ferc714/out_ferc714__respondents_with_fips/schema.yml
```diff
@@ -9,6 +9,14 @@ sources:
               arguments:
                 table_name: out_ferc714__respondents_with_fips
                 partition_expr: "EXTRACT(YEAR FROM report_date)"
+          - dbt_expectations.expect_compound_columns_to_be_unique:
```
jdangerx, line 12: I saw this in the resource definition:

```
    "out_ferc714__respondents_with_fips": {
        "description": {
            "additional_summary_text": "Annual respondents with the county FIPS IDs for their service territories.",
            "additional_primary_key_text": (
                "The state and county FIPS columns "
                "which are part of the natural primary key can be null. "
                "The natural primary key would be: "
                "['respondent_id_ferc714', 'report_date', 'state_id_fips', 'county_id_fips']."
            ),
```
Is `state_id_fips` incorrectly missing in this file or incorrectly extraneous in `resources/ferc714`?

dbt/models/phmsagas/_core_phmsagas__yearly_distribution_excavation_damages/schema.yml
jdangerx: Should we update the docs then?

jdangerx (changes requested):
Mostly wondering if the omission of `state_id_fips` in `out_ferc714__respondents_with_fips` was intentional/the result of some investigation - beyond that nothing blocking!

dbt/models/ferc714/out_ferc714__respondents_with_fips/schema.yml
zaneselvans: The 2-digit `state_id_fips` is not necessary to produce a unique PK because the 5-digit `county_id_fips` includes the state FIPS code as the first two digits.

dbt/models/phmsagas/_core_phmsagas__yearly_distribution_excavation_damages/schema.yml
zaneselvans: I wasn't sure what to do here, since the desired future situation that the docs mention seems to be the current situation, but I didn't have enough context to know whether that additional cleaning had already happened or not. I think it was @e-belfer or @cmgosnell who integrated this table.

e-belfer: This looks like a copy-paste mistake to me that got missed in the integration of this table, based on looking at the original documentation and checking the index on the last build. @cmgosnell would be the one who integrated this though, so I defer to her.

If that's the case, we should drop the non-PK explanation and actually add the PK as part of this PR.

dbt/models/ferc714/out_ferc714__respondents_with_fips/schema.yml
e-belfer: It seems desirable that they should match - either by adding the state FIPs into the test or dropping it from the additional text.

zaneselvans: Updated the additional PK text to remove state FIPS code.

dbt/models/phmsagas/_core_phmsagas__yearly_distribution_excavation_damages/schema.yml
zaneselvans: Added the real PK and removed the metadata annotation.

cmgosnell: I agree w/ ella here. this table has a pk in the schema so we should delete this and the `additional_primary_key_text`

PR #4812, opened 2025-12-05
Increase VM size to avoid OOM

zaneselvans:
# Overview

- It's been a week since we got a successful build. There have been ~10 builds that made it to the end of the ETL since then, but last night we got what was probably an out-of-memory error. Not clear if this is a stochastic problem, and if so how common it would be.
- To make sure we've dealt with the other build issues (row count instability, etc.) I'm bumping the VM back up to 16 CPU and 128 GB RAM. Once we've gotten another good build we can revisit resource constraints.
- Also, exclude changes under `./devtools/**` from triggering the integration tests in CI.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -34,6 +34,7 @@ jobs:
           filters: |
             code:
               - '!dbt/seeds/etl_full_row_counts.csv'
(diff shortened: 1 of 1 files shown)


krivard (approved):
:shipit:

Issue #4797, opened 2025-12-02
Add information to table description and metadata for top priority tables


cmgosnell:
### Overview
Some of our table descriptions are pretty sparse (aka need more detail). Some don't have a description for why there is no primary key. Some tables need more descriptions about what we have done to the transform or modify the data. etc etc. We set up a prioritized list of our tables and flag for if it needs attention.

https://docs.google.com/spreadsheets/d/1Hv82QN5mVKieRC8B34LxQIsWw3WMjfkarRLgIfvYLX8/edit?gid=0#gid=0

### Success Criteria

How will we know that we're done?

* [x] we get through as many tables in this prioritized list as we can in ~8 hours:
   - out_eia__monthly_generators
   - core_epacems__hourly_emissions
   - core_nrelatb__yearly_projected_cost_performance
   - core_eia860__scd_generators
   - core_eia860__scd_plants
   - out_ferc714__hourly_estimated_state_demand
   - out_eia__yearly_generators
   - _core_eia860__cooling_equipment
   - core_eia861__yearly_operational_data_misc
   - out_eia__yearly_generators_by_ownership
   - out_eia923__monthly_generation_fuel_by_generator_energy_source
   - core_eia861__yearly_reliability
   - core_eia861__yearly_operational_data_revenue
   - out_eia923__yearly_generation_fuel_by_generator
   - out_eia923__yearly_generation_fuel_by_generator_energy_source_owner
   - out_eia923__monthly_generation_fuel_by_generator

### Next steps
* Add detail for the remaining tables! (See spreadsheet linked above for more information).

cmgosnell:
we are done this for now!

PR #4813, opened 2025-12-05
Add to the table description for `out_eia__yearly_generators_by_ownership`

cmgosnell:
# Overview

Working on #4797.

## What did you change?
- added a primary key to this table (i tested this locally via code snipped below...) just kidding i added a no pk note
- added info to this table.. please let me know if this makes sense its an odd table!

```python
own_gens = defs.load_asset_value("out_eia__yearly_generators_by_ownership")
pks= ["plant_id_eia", "generator_id", "report_date", "utility_id_eia", "ownership_record_type"]
own_gens = own_gens.set_index(pks)
assert own_gens[own_gens.index.duplicated()]
```

## Documentation

# Testing

## To-do list

### DIFF
--- a/migrations/versions/aeb66b0b4481_add_planned_repower_date_into_860m_.py
+++ b/migrations/versions/aeb66b0b4481_add_planned_repower_date_into_860m_.py
@@ -1,7 +1,7 @@
 """add planned_repower_date into 860m changelog
 
 Revision ID: aeb66b0b4481
-Revises: 5ba8ad3ad527
+Revises: 907de695bcc5
 Create Date: 2025-11-26 13:56:08.558987
 
 """
@@ -11,7 +11,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = 'aeb66b0b4481'
-down_revision = '5ba8ad3ad527'
+down_revision = '907de695bcc5'
 branch_labels = None
 depends_on = None
 
--- a/migrations/versions/d4f6362326cc_add_pk_to_861_yearly_operational_data.py
+++ b/migrations/versions/d4f6362326cc_add_pk_to_861_yearly_operational_data.py
@@ -1,7 +1,7 @@
 """Add pk to 861 yearly_operational_data
 
 Revision ID: d4f6362326cc
-Revises: 5ba8ad3ad527
+Revises: aeb66b0b4481
 Create Date: 2025-12-05 15:25:42.741782
 
 """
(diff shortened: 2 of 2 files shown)


zaneselvans:
I was messing with this table in the row-count investigation and noticed that there were some NA values in the `utility_id_eia` column, which I think means this PK won't work, since SQLite doesn't allow NULL PK values.

```python
pk = ["plant_id_eia", "generator_id", "report_date", "utility_id_eia", "ownership_record_type"]
dude: pd.DataFrame = defs.load_asset_value("out_eia__yearly_generators_by_ownership")
wtf = dude.set_index(pk)
assert wtf.index.is_unique
for col in pk:
    if (nulls := dude[col].isna().sum()) != 0:
        logger.warning(f"Column {col} has {nulls} missing values!")
```

```
2025-12-05 12:32:09 [ WARNING] dagster.builtin.catalystcoop.pudl:7 Column utility_id_eia has 200 missing values!
```

Though it also seems weird that there would be generators with ownership defined, but no utility ID.

cmgosnell:
ya i just noticed this from running tests. its almost certainly bc these gen-years are not it core_eia860__scd_ownership or are otherwise missing assocaited records with owner ids. that's okay i will convert this into a no-pk note

src/pudl/metadata/resources/ferc1_eia_record_linkage.py
```diff
@@ -98,13 +98,26 @@
...
             "additional_summary_text": "all EIA generators with ownership integrated.",
-            "additional_details_text": inherits_harvested_values_details(
-                "generators, plants, and utilities"
-            ),
+            "additional_details_text": f"""Generators that report to EIA can have
+multiple utility owners. EIA reports the ownership fraction for each of the owners
+of generators (see :ref:`core_eia860__scd_ownership`). This table includes two record
+for each generator's owner - one record represents a total ownership stake  and a
+second record with a proportional ownership stake (the ``ownership_record_type``
+column labeled as "total" and "owned" respectively). All of the values reported for
```
krivard: minor copy edit

```suggestion
of generators (see :ref:`core_eia860__scd_ownership`). This table includes two records
for each generator's owner: one record represents a total ownership stake and the
other record represents a proportional ownership stake (the ``ownership_record_type``
column labeled as "total" and "owned" respectively). All of the values reported for
```

krivard (approved):
This needs a rebase before merging so that it won't duplicate the harvest stuff but otherwise good!

PR #4814, opened 2025-12-05
Remove unused `field_namespace` from DataSource

krivard:
# Overview

Remove unused `field_namespace` from DataSource

While working on the metadata dev docs, we discovered that `DataSource.field_namespace` is never used.

## What did you change?

* Drop unused field `DataSource.field_namespace`
* Drop unused class method `DataSource.from_field_namespace()`
* Drop unused helper function `pudl.helpers.get_working_dates_by_datasource()`
* Drop `field_namespace` keys from `pudl.metadata.sources.SOURCES`

## Documentation

# Testing

Searched src, test, notebooks, devtools, and docs for all symbols being removed, with no results.
Ran `make pytest-coverage`.

## To-do list

### DIFF
--- a/src/pudl/helpers.py
+++ b/src/pudl/helpers.py
@@ -1506,26 +1506,6 @@ def iterate_multivalue_dict(**kwargs):
         yield result
 
 
-def get_working_dates_by_datasource(datasource: str) -> pd.DatetimeIndex:
-    """Get all working dates of a datasource as a DatetimeIndex."""
-    import pudl.metadata.classes
-
-    dates = pd.DatetimeIndex([])
-    for data_source in pudl.metadata.classes.DataSource.from_field_namespace(
-        datasource
-    ):
-        working_partitions = data_source.working_partitions
-        if "years" in working_partitions:
-            dates = dates.append(
-                pd.to_datetime(working_partitions["years"], format="%Y")
-            )
-        if "year_months" in working_partitions:
-            dates = dates.append(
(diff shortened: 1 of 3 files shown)


zaneselvans:
The intention with this was that we would use it to specify field-level metadata that would override the global metadata associated with a given "namespace" e.g. all of the EIA data sources could have a field namespace of "eia" if they had common columns that shared a meaning, but that was distinct from the global meaning of that field. IIRC at some point this particular override, but not the individual resource-level overrides (which still live at the bottom of `fields.py`) stopped working, and I spent an annoying amount of time trying to figure out why and failed, and ended up migrating the values that were being overridden in there to resource-level overrides (albeit with some duplication) since there were only a few of them. I probably neglected to remove this related infrastructure / hoped that it would get fixed.

src/pudl/helpers.py
```diff
@@ -1506,26 +1506,6 @@ def iterate_multivalue_dict(**kwargs):
         yield result
 
 
-def get_working_dates_by_datasource(datasource: str) -> pd.DatetimeIndex:
-    """Get all working dates of a datasource as a DatetimeIndex."""
```
aesharpe, line 1510: Is this something we _should_ be using?

krivard:
@aesharpe 
> Should we do a clean sweep of all field_namespace values or do some of them have different purposes?

The remaining `field_namespace` is for `Resource.field_namespace`, which still matters. This PR just removes `DataSource.field_namespace`.

PR #4815, opened 2025-12-05
Add to the table description for core_eia861__yearly_operational_data_*

krivard:
# Overview

Working on #4797.

_misc is the popular one but just about everything for _misc also applies to _revenue so I did both.

## What did you change?

* Added new primary keys for _misc and _revenue! these are possible now because @aesharpe fixed 88888
* Added the respective schedules as source annotations 
* Added a blurb about the 88888 redactions
* Added redaction warning
* Added imputation warning for _misc only
* Unforch not much more information seems to be available about these tables 😢 

## Documentation

# Testing

## To-do list

### DIFF
--- a/migrations/versions/f0b14d612162_.py
+++ b/migrations/versions/f0b14d612162_.py
@@ -1,0 +1,24 @@
+"""empty message
+
+Revision ID: f0b14d612162
+Revises: 7d9ec620e804, d4f6362326cc
+Create Date: 2025-12-11 13:08:01.120144
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'f0b14d612162'
+down_revision = ('7d9ec620e804', 'd4f6362326cc')
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    pass
+
+
+def downgrade() -> None:
+    pass


migrations/versions/fc84478c36f8_add_pk_to_861_yearly_operational_data.py
```diff
@@ -0,0 +1,108 @@
...
+               existing_type=sa.INTEGER(),
+               nullable=True)
+        batch_op.alter_column('report_date',
+               existing_type=sa.DATE(),
+               nullable=True)
+        batch_op.alter_column('nerc_region',
+               existing_type=sa.VARCHAR(length=9),
+               nullable=True)
+
+    op.create_table('_alembic_tmp_core_eia861__yearly_operational_data_',
```
krivard: What is this table and do I need to be concerned that alembic is messing with it?

src/pudl/metadata/resources/eia861.py
```diff
@@ -581,12 +586,19 @@
...
-                """Sources include: net generation
+                f"""Sources include: net generation
 purchases from electricity suppliers, exchanges received, exchanges delivered, exchanges
 net, wheeled received, wheeled delivered, wheeled net, transmission by others, and
-losses."""
+losses.
+
+{EIA861_88888}
+
+Rows where ``data_observed`` is False were labeled as imputed in the raw EIA data.
```
aesharpe, line 593: Do we know more about how they were imputed or what that means?

src/pudl/metadata/resources/eia861.py
```diff
@@ -581,12 +586,19 @@
...
         "description": {
             "additional_summary_text": "megawatt hours (MWH) for all a utility's sources of electricity and disposition of electricity listed.",
+            "additional_source_text": "(Schedule 2B)",
             "additional_details_text": (
-                """Sources include: net generation
+                f"""Sources include: net generation
 purchases from electricity suppliers, exchanges received, exchanges delivered, exchanges
 net, wheeled received, wheeled delivered, wheeled net, transmission by others, and
-losses."""
+losses.
```
aesharpe: I'm not 100% what "sources" means here

aesharpe (changes requested):
Just a couple clarifications

src/pudl/metadata/resources/eia861.py
krivard: I assumed the same as the "sources" in "megawatt hours (MWH) for all a utility's sources of electricity and disposition of electricity listed" from the summary. No?

src/pudl/metadata/resources/eia861.py
krivard: Alas, no.

The imputation column in this table is from the Operational Data file from the EIA, which contains a column 

> Data Type
> O = Observed
> I = Imputed

with no additional notes. In the file for 2024, all entries read "O" and none have "I". In `core_eia861__yearly_operational_data_misc` there are only 195 rows (of ~64k) where `data_observed` is false: 179 in 2004 and 16 in 2005; no utility in that list has more than 2 imputed values, and none more than 25% imputed values.

The EIA website mentions imputation in the context of [Net Metering](https://www.eia.gov/electricity/data/eia861/#:~:text=AC.%20For%20other%20energy%20sources%2C%20we%20have%20not%20established-,imputation,-procedures.) and [Non-Net Metering Distributed](https://www.eia.gov/electricity/data/eia861/#:~:text=schedule.%20For%20other%20energy%20sources%2C%20we%20have%20not%20established-,imputation,-procedures.), but there are no imputation columns in the Net Metering or Non-Net Metering Distributed files from the EIA.

Schedule 2B of the Form 861 ("Energy sources and disposition") makes no mention of imputation. Neither does Schedule 7A "Net metering" or Schedule 7B "Non net-metering distributed". The Form 861 instructions PDF makes no mention of imputation either. Would an addition like this help?


```suggestion
Rows where ``data_observed`` is False were labeled as imputed in the raw EIA data.
The EIA documentation does not specify what imputation applies to operational data,
only net metering and non net-metering distributed data. Less than 1% of rows are
labeled as imputed, all of which occur in 2004 or 2005. Imputed rows make up no
more than 25% of the rows for any utility.
```

aesharpe: That looks great!!

src/pudl/metadata/resources/eia861.py
aesharpe: Ahh ok, it makes more sense after reading it a few times / googling what disposition means from the prior sentence. But I wonder if it makes sense to include these sources in this section when they are essentially just the column names. The sales columns are left out of this list as well--was that intentional? Maybe we could just leave this section blank and let the columns do the talking.

Also, I think the "additional_summary_text" could probably benefit from a slight re-word. It was probably copied and pasted strait from EIA, but "megawatt hours (MWH) for all a utility's sources of electricity and disposition of electricity listed" is a little difficult to grasp. Maybe: "megawatt hours (MWH) for the sources and disposition of a utility's electricity." Just a little less clunky.

PR #4820, opened 2025-12-08
Improve performance of `out_eia__yearly_plant_parts`

### LINKED ISSUE #2920: Parallelize the creation of `plant_parts_eia`
Now that `plant_parts_eia` is dagsterized and part of the ETL, it's a bit of a bummer that it takes so long to generate. Currently, there's a loop in the creation of the table [here](https://github.com/catalyst-cooperative/pudl/blob/main/src/pudl/analysis/plant_parts_eia.py#L606) that loops through all the different parts and concatenates a bunch of dataframes together. This could be done concurrently instead of looping to speed things up.

One pattern that would make sense here would be to create an asset factory that generates one asset for each of the different elements of the `PLANT_PARTS` constant, and then have a separate asset that concatenates all of those outputs together into a single big table and add the other little finishing touches.

Currently the `out_eia__yearly_plant_parts` asset is marked as high memory usage, and takes about half an hour to run (on a Mac M1). It's the longest running bit of the plant parts / FERC-EIA association portion of the DAG by a significant margin.

katie-lamb:
# Overview

Closes #2920 .

`out_eia__yearly_plant_parts` is slow to generate. It's created by making a table for each plant part and then concatenating these part tables together. The creation of these plant parts can be run in parallel with Dagster. This PR creates an asset factory to generate an asset for each of the plant parts (named `out_eia__plant_part_{x}` and then concatenates it together into one.

 I can generate `out_eia__yearly_plant_parts` and its upstream `_out_eia__plant_part_x` assets in 6 minutes and 8 seconds. On `main`, it takes me 7 minutes and 13 seconds to generate `out_eia__yearly_plant_parts`, so overall it's not as big a speed up as I'd expect. However, I wonder if there is a bigger speed up on a slower computer. I have a Mac M3, but the issue says that this asset takes half an hour to generate on a Mac M1. It may be good to have someone with a slower computer run this as well. 

Is it worth taking on other performance improvement ideas for this table if this asset splitting didn't take much time off?

## What did you change?

Created an asset factory to generate an asset for each of the plant parts (named `out_eia__plant_part_{x}`) and then concatenate it together into one `out_eia__yearly_plant_parts`.

## Documentation

# Testing

- [x] Check to make sure the `out_eia__yearly_plant_parts` asset is identical when materialized on `main`.
- [x] Check the run time on `main`.

## To-do list

### DIFF
--- a/src/pudl/analysis/plant_parts_eia.py
+++ b/src/pudl/analysis/plant_parts_eia.py
@@ -367,6 +367,7 @@ plant_parts_assets = [
 @asset(
     io_manager_key="pudl_io_manager",
     compute_kind="Python",
+    op_tags={"memory-use": "high"},
     ins={
         part_name: AssetIn(key=AssetKey(f"_out_eia__plant_part_{part_name}"))
         for part_name in PLANT_PARTS


src/pudl/analysis/plant_parts_eia.py
```diff
@@ -338,23 +338,58 @@ def out_eia__yearly_generators_by_ownership(
...
+    if part_name != "plant_match_ferc1"
+]
+
+
 @asset(
     io_manager_key="pudl_io_manager",
     compute_kind="Python",
-    op_tags={"memory-use": "high"},
+    ins={
+        part_name: AssetIn(key=AssetKey(f"_out_eia__plant_part_{part_name}"))
```
katie-lamb, line 372: This asset name feels okay for the intermediate plant part tables?

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -338,23 +338,58 @@ def out_eia__yearly_generators_by_ownership(
...
+    plant_part_asset_factory(part_name)
+    for part_name in PLANT_PARTS
+    if part_name != "plant_match_ferc1"
+]
+
+
 @asset(
     io_manager_key="pudl_io_manager",
     compute_kind="Python",
-    op_tags={"memory-use": "high"},
```
zaneselvans, line 344: Have you profiled this asset to see how much memory it uses? I suspect it'll be pretty big with all the concat-ing.

src/pudl/analysis/plant_parts_eia.py
```diff
@@ -338,23 +338,58 @@ def out_eia__yearly_generators_by_ownership(
     )
 
 
+def plant_part_asset_factory(part_name: str) -> AssetsDefinition:
+    """Asset factory to create assets for each individual plant part."""
+    asset_name = f"_out_eia__plant_part_{part_name}"
+
+    @asset(
+        name=asset_name,
+        compute_kind="Python",
```
zaneselvans, line 347: We should probably profile each of these assets to identify which ones need to be marked high memory (larger than 4GB peak)

src/pudl/analysis/plant_parts_eia.py
zaneselvans: On my machine I get a peak of nearly 14GB:

```
Command line: devtools/memory_profile.py -d devtools/profiles -a out_eia__yearly_plant_parts
Start time: Tue Dec 09 2025 10:23:43 GMT-0600 (Central Standard Time)
End time: Tue Dec 09 2025 10:34:11 GMT-0600 (Central Standard Time)
Duration: 0:10:28.261000
Total number of allocations: 112909931
Total number of frames seen: 0
Peak memory usage: 13.8 GB
Python allocator: pymalloc
```

katie-lamb: Is it worth trying to reduce memory from concat-ing? I couldn't find a good structure for handling the concat-ing through anything other than `pd.concat`

src/pudl/analysis/plant_parts_eia.py
zaneselvans: I ran them all and they were between 2-3GB each so no highmem needed!

```bash
for part in plant_prime_fuel plant_gen plant_prime_mover plant_technology plant_operating_year plant plant_ferc_acct plant_unit;
do
    devtools/memory_profile.py -d devtools/profiles -a  "_out_eia__plant_part_$part" &
done;
```

src/pudl/analysis/plant_parts_eia.py
zaneselvans: Concatenation is a big memory suck all over the place for us, and I don't think we've come across a great solution. This is just a chonker of a table. And there are other high memory usage operations happening within the the main asset as well.

I think you've picked the low hanging fruit here and this should come close to cutting the time to run the plant parts in half, since now we have 8 low-memory assets that can run in parallel, each of which takes ~100 seconds, for the first half of the process, rather than needing to run for 800 seconds one after the other.

zaneselvans (approved):

Issue #4822, opened 2025-12-09
Migrate from ReadTheDocs to GitHub Pages


krivard:
### Overview

What is the problem we're solving? For very simple items, this can be encapsulated in the success criteria.

PUDL is no longer a standalone library, which is making builds on RTD more and more challenging to maintain.

We want our docs to be able to use things from the repo, not just the installed package. Sphinx itself doesn't care either way, so we're going to try GitHub Pages instead.

### Success Criteria

How will we know that we're done?

* [x] We have a GitHub Pages deployment with the most recent version of our documentation site
* [x] We have a GitHub Pages deployment with all versions of our documentation site
* [x] We have a GitHub Pages deployment with all versions of our documentation site, defaulting visitors to the most recent
* [x] We have any of the above deployed at https://docs.catalyst.coop
* [ ] We know why a GitHub Pages deployment is a bad choice for us

### Next steps

* [x] Brief research to see if we can avoid any pitfalls identified by others
* [x] Add a build-deploy-docs workflow
* [x] Configure Pages
* [x] Figure out how to handle versioning

We've decided to use the sphinx-pydata theme, which includes a version picker.

We've decided to use the sphinx-design plugin for tab displays, since sphinx-tabs has some weird difficult-to-replicate conflict with the sphinx-pydata version picker (and also sphinx-tabs generates deprecation warnings it will be satisfying not to see anymore).

Remaining work:

* [x] Reorganize top bar links
* [x] Fix CSS spacing and layout
* [x] #5176
* [ ] ⏩ https://github.com/catalyst-cooperative/pudl/issues/5146 Hack custom dropdowns into the top bar (maybe use [3-year-old pr](https://github.com/pydata/pydata-sphinx-theme/pull/1165) for inspiration)
* [ ] ⏩ https://github.com/catalyst-cooperative/pudl/issues/5146 Remove level-2 nav from secondary sidebar on data dictionary pages only

### Research

* https://pointlessramblings.com/posts/replacing-readthedocs-with-github/
* https://github.com/gammapy/gammapy/issues/1241
* https://tech.michaelaltfield.net/2020/07/18/sphinx-rtd-github-pages-1/
* https://github.com/fmtlib/fmt/issues/128
* https://coderefinery.github.io/documentation/gh_workflow/

krivard:
Versioning: I've done test install of pydata-sphinx-theme which has really nice version switcher support.

[...]

PR #4823, opened 2025-12-09
Add GH Pages workflow

krivard:
# Overview

Working on #4822.

If we merge this:

* A new workflow "build-deploy-docs" will show up in Actions
* That workflow may only run via dispatch
* If run via dispatch, that workflow will establish a branch `gh-pages` with a single commit in it, and whose file tree is rooted at `docs/_build/html` -- none of our other files will be in this branch. **This initial run of this workflow is expected to fail.**
* Once the `gh-pages` branch exists, we can activate Pages in repo settings with the following configuration:
  * Source: Deploy from a branch
  * Branch: `gh-pages` + `/ (root)`
* Once Pages is activated, we can run the `build-deploy-docs` action via dispatch once more, and the site will deploy at https://catalyst-cooperative.github.io/pudl
  * ...maybe. We might need to deactivate the CNAME deployment from zane's mkdocs prototype at https://github.com/catalyst-cooperative/catalyst-cooperative.github.io
* Once we are satisfied the site is building, we can uncomment the non-dispatch lines from the new workflow

For a demo, see this in action in my fork:

* https://krivard.github.io/pudl/release_notes.html
* https://github.com/krivard/pudl/tree/gh-pages
* https://github.com/krivard/pudl/commits/gh-pages/

## What did you change?

* New workflow to build docs and deploy to gh-pages

## Documentation

# Testing

## To-do list


.github/workflows/build-deploy-docs.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: mkdir -p ${{ env.PUDL_OUTPUT }} ${{ env.PUDL_INPUT}} ${{ env.DAGSTER_HOME }}
+
+      - name: Lint and build PUDL documentation with Sphinx
+        run: |
+          pip install --no-deps --editable .
+          make docs-build
+
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v4
+        # if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v20') }}
```
krivard, line 54: This is intended to be uncommented once we activate `push:` and `pull-request:` dispatch, so that the live docs only update automatically when we tag a new release. This is a more conservative choice than we have now, which continuously deploys documentation updates to RTD.

Since RTD permits you to browse backversions and this initial gh-pages version does not, the impact of clobbering the only live copy of the docs with an oops merge is higher, so continuously deploying docs to gh-pages feels more risky. thoughts?

.github/workflows/build-deploy-docs.yml
```diff
@@ -0,0 +1,61 @@
+name: build-deploy-docs
+
+on:
+  # push:
+  #   branches:
+  #     - '!gh-pages'
+  # pull_request:
```
krivard, line 7: We'll uncomment these after this PR, once we've kicked the tires on deploying to gh-pages

.github/workflows/build-deploy-docs.yml
zaneselvans: Oh no. There's no way to preserve access to stable vs. nightly vs. latest vs. v2025.11.0 etc versions of the documentation if it's hosted on GitHub Pages? That's a bummer.

By far the most commonly used version of our data is the nightly builds, so I think if there can only be one version of the docs, it needs to be the most recent version, otherwise the docs and the data will be up to a month (or a quarter) out of sync.  Maybe ideally it would be `nightly` since that's the deployed data. But if not `nightly` then whatever is on `main`.

krivard: We can do multiple versions, but it's not automatic -- we have to figure out how we want to organize different versions, how to build them, and how to arrange redirects on our own. For this pass, we only get one version at a time.

zaneselvans: While we're figuring out the new system it seems like there's also no problem with deploying to both GH Pages and RTD, and once we have the GH Pages setup how we like it we can deprecate RTD.

I guess there's also probably no facility for previewing the built docs in a PR?

zaneselvans:
I thought that the way the organization-wide pages configuration was set up, this would by default deploy to

https://docs.catalyst.coop/pudl

At least that's [what these docs](https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/about-custom-domains-and-github-pages#using-a-custom-domain-across-multiple-repositories) seem to suggest.

.github/workflows/build-deploy-docs.yml
krivard: > I guess there's also probably no facility for previewing the built docs in a PR?

Not built-in, but I bet we could set up something like Netlify to do that

zaneselvans (approved):
@krivard if we think this should work for the single-version deployment let's get it merged in so we can play with it live and iron out any custom domain deployment bugs while we figure out if/how to do historical versions.

zaneselvans:
I've activated github pages on the repo and voila: https://docs.catalyst.coop/pudl/

PR #4828, opened 2025-12-10
Apply polars dtypes to vcerare output table

zschira:
# Overview

This PR fixes an issue with a recent update to `vcerare` processing meant to improve memory performance. In that PR, we switched to using Polars/duckdb for processing, but this resulted in minor changes to dtypes. This PR makes sure that we apply the appropriate dtypes when writing to parquet.

### DIFF
--- a/src/pudl/transform/vcerare.py
+++ b/src/pudl/transform/vcerare.py
@@ -20,7 +20,6 @@ from pudl.helpers import (
     simplify_columns,
     zero_pad_numeric_string,
 )
-from pudl.metadata.classes import Resource
 from pudl.metadata.dfs import POLITICAL_SUBDIVISIONS
 from pudl.workspace.setup import PudlPaths
 
@@ -517,8 +516,4 @@ def out_vcerare__hourly_available_capacity_factor(
 
     return lf_from_parquet(
         ParquetData(table_name=partitioned_output_table), use_all_partitions=True
-    ).cast(
-        Resource.from_id(
-            "out_vcerare__hourly_available_capacity_factor"
-        ).to_polars_dtypes()
     )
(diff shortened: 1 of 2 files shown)


src/pudl/transform/vcerare.py
```diff
@@ -516,4 +517,8 @@ def out_vcerare__hourly_available_capacity_factor(
 
     return lf_from_parquet(
         ParquetData(table_name=partitioned_output_table), use_all_partitions=True
+    ).cast(
+        Resource.from_id(
+            "out_vcerare__hourly_available_capacity_factor"
+        ).to_polars_dtypes()
```
jdangerx: non-blocking: is there a way to enforce the schema at the IO manager level instead?

jdangerx (approved):
Looks good for a quick fix! Knowing you, you probably thought about solving this at the IO manager level also. Curious if you ran into some issues there that turned into a rabbit hole :rabbit2:

src/pudl/transform/vcerare.py
zschira: I think this exact same approach _should_ work, but I thought we had tried something earlier on and ran into issues with some assets. That being said, there's not that many assets using the io-manager with polars, so I'll see if I can test doing so before merging.

zaneselvans: I thought the issue was that it was blowing up memory usage? Or maybe we were trying to do the other parts of `enforce_schema` as well and something in there was responsible.

zschira: Just tested moving this to the io-manager with no problems! I think that the only assets using this right now are this one, `epacems`, and `out_ferc714__hourly_estimated_state_demand`, and all work well.

Issue #4834, opened 2025-12-11
Small number of duplicate record_id in core_ferc1__yearly_purchased_power_and_exchanges_sched326


krivard:
### Impacted table(s)

core_ferc1__yearly_purchased_power_and_exchanges_sched326

### How severe is this data bug?

Low

### Identify impacted data

```python
core_ferc1__yearly_purchased_power_and_exchanges_sched326 = pd.read_parquet(
    "https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/v2025.11.0/core_ferc1__yearly_purchased_power_and_exchanges_sched326.parquet"
)
sz = core_ferc1__yearly_purchased_power_and_exchanges_sched326.record_id.value_counts()
sz.loc[sz>1]
```

```
record_id
purchased_power_326_2022_c000618_other_net_metering_rq_0                2
purchased_power_326_2024_c000685_cald_bess_lu_ferc_vol_10               2
purchased_power_326_2024_c000913_pjm_interconnection_llc_os_na          2
purchased_power_326_2021_c001421_cheyenne_light_fuel__power_os          2
purchased_power_326_2024_c000537_vaisala_monthly_forecasting_f_os_na    2
Name: count, dtype: Int64
```

### Describe the bug

While hunting a primary key explanation for core_ferc1__yearly_purchased_power_and_exchanges_sched326, I found that there's only a small number of record ids that are duplicated, and some of them don't even have different data across the multiple uses. This seems suspicious.

```python
core_ferc1__yearly_purchased_power_and_exchanges_sched326.loc[
    core_ferc1__yearly_purchased_power_and_exchanges_sched326.record_id.isin(
        sz.loc[sz>1].index
    )
    ].sort_values("record_id")
```

* `purchased_power_326_2021_c001421_cheyenne_light_fuel__power_os` - all 3 non-null value columns differ
* `purchased_power_326_2022_c000618_other_net_metering_rq_0` - identical
* `purchased_power_326_2024_c000537_vaisala_monthly_forecasting_f_os_na` - identical
* `purchased_power_326_2024_c000685_cald_bess_lu_ferc_vol_10` - fills in a different and nonoverlapping subset of value columns
* `purchased_power_326_2024_c000913_pjm_interconnection_llc_os_na` - fills in a different subset of value columns with 1 overlapping column having a different value

We should:
* Figure out why those identical records are there, and get rid of them if we can
[...]

PR #4835, opened 2025-12-11
Add to the table description for *_ferc1__yearly_purchased_power_and_exchanges_sched326

krivard:
# Overview

Working on #4797.

## What did you change?

* Added a pk explanation for the shared yearly_purchased_power_and_exchanges_sched326 description.

There are only 5 duplicate record_ids preventing this table from having a primary key. For two of them, the records are identical, and probably shouldn't be duplicated. I have filed #4834 as a low-pri follow-up.

## Documentation

# Testing

Previewed in wizard.

### DIFF
--- a/src/pudl/metadata/resources/ferc1.py
+++ b/src/pudl/metadata/resources/ferc1.py
@@ -277,6 +277,7 @@ columns."""
             "and any settlements for imbalanced exchanges."
         ),
         "additional_source_text": "(Schedule 326)",
+        # Remove if bug is squashed: https://github.com/catalyst-cooperative/pudl/issues/4834
         "additional_primary_key_text": """The primary key for this table would be record_id,
   however a small number of these values are duplicated in 2021, 2022, and 2024.""",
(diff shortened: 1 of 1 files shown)


src/pudl/metadata/resources/ferc1.py
```diff
@@ -277,6 +277,8 @@
             "and any settlements for imbalanced exchanges."
         ),
         "additional_source_text": "(Schedule 326)",
+        "additional_primary_key_text": """The primary key for this table would be record_id,
+  however a small number of these values are duplicated in 2021, 2022, and 2024.""",
```
cmgosnell, line 282: very non-blocking but helpful for future us maybe
```suggestion
        # Remove if bug is squashed: https://github.com/catalyst-cooperative/pudl/issues/4834
        "additional_primary_key_text": """The primary key for this table would be record_id,
  however a small number of these values are duplicated in 2021, 2022, and 2024.""",
```

PR #4836, opened 2025-12-12
perf: reduce `out_eia923__monthly_generation_by_generator_energy_source_code` memory usage 38GB -> 22 GB

jdangerx:
# Overview

This asset was blowing up my local ETL runs by using 38GB of memory.

## What did you change?

* set `pd.options.mode.copy_on_write` which drastically reduces the amount of array copies that happen
* fix some parts of our energy source code munging where the energy source code ordering was not stable
* cleaned up a little bit of the "find the right MSN/MSB codes" logic

Looked at changing more, but chose not to:
* moving away from string manipulation as intermediate representation for the MSW -> MSN/MSB expansion - that just seems like a lot of extra logic for... no gain
* moving into duckdb for more out-of-memory processing - 22GB is small enough for this bad boy to run on our dev machines for now and we're out of time for further memory optimizations

## Documentation

I don't think there's anything a user would care about that I changed.

# Testing

Materialized with `memory_profile.py` and compared against nightly to make sure the output data was exactly the same.

## To-do list

### DIFF
--- a/src/pudl/analysis/allocate_gen_fuel.py
+++ b/src/pudl/analysis/allocate_gen_fuel.py
@@ -1664,7 +1664,9 @@ def adjust_msw_energy_source_codes(
         bf_by_gens.energy_source_code.unique()
     )
     # NOTE 2025-12-11: we sort this reverse-ly to match the indeterminate order that happens to be in nightly builds right now
-    msw_codes_used = sorted({"MSN", "MSB", "MSW"}.intersection(codes_used), reverse=True)
+    msw_codes_used = sorted(
+        {"MSN", "MSB", "MSW"}.intersection(codes_used), reverse=True
+    )
     # join these codes into a string that will be used to replace the MSW code
     replacement_codes = ",".join(msw_codes_used)
 


src/pudl/analysis/allocate_gen_fuel.py
```diff
@@ -247,6 +247,7 @@ def gen_fuel_by_gen_esc(
         gens: pd.DataFrame,
     ) -> pd.DataFrame:
         """Allocate net gen from gen_fuel to generator/energy_source_code level."""
+        pd.options.mode.copy_on_write = True
```
jdangerx, line 250: I put this in the asset level so it wouldn't mess with e.g. unit tests.

src/pudl/analysis/allocate_gen_fuel.py
```diff
@@ -1659,19 +1660,11 @@ def adjust_msw_energy_source_codes(
...
-        bf_by_gens.loc[
-            bf_by_gens["energy_source_code"].isin(["MSW", "MSB", "MSN"]),
-            "energy_source_code",
-        ].unique()
+    codes_used = set(gf.energy_source_code.unique()) | set(
+        bf_by_gens.energy_source_code.unique()
     )
-    msw_codes_used = list(msw_codes_in_gf | msw_codes_in_bf)
+    # NOTE 2025-12-11: we sort this reverse-ly to match the indeterminate order that happens to be in nightly builds right now
+    msw_codes_used = sorted({"MSN", "MSB", "MSW"} & codes_used, reverse=True)
```
jdangerx: This could probably be even shorter but seems clear enough.

cmgosnell: nit non-blocking preference for `intersection` instead of `&` bc it feels more explicit
```suggestion
    msw_codes_used = sorted({"MSN", "MSB", "MSW"}.intersection(codes_used), reverse=True)
```

cmgosnell (approved):
i haven't tried to run this locally but assuming you have materialized this asset and checked it against its nightly build counterpart, it looks like a straightforward change. I made a non-blocking suggestion just for clarity

PR #4839, opened 2025-12-13
Remove illegal ref label

zaneselvans:
# Overview

Removes the `build_ref` label from our Google Batch VM, since `$BUILD_REF` contains prohibited characters, which prevented the release build from kicking off.

Part of #4837

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -23,6 +23,6 @@ authors:
     orcid: https://orcid.org/0009-0009-9063-080X
 
 title: "The Public Utility Data Liberation (PUDL) Project"
-version: 2025.12.0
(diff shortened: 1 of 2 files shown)


devtools/generate_batch_config.py
```diff
@@ -72,7 +72,6 @@ def to_config(
         "logsPolicy": {"destination": "CLOUD_LOGGING"},
         "labels": {
             "component": "build",
-            "build_ref": env_dict.get("BUILD_REF", None),
```
pudlbot, line 75: It turns out labels cannot contain periods (`.`) and so our versions are not valid lables

PR #4848, opened 2025-12-15
Add new PUDL citation

zaneselvans:
# Overview

* Adds a new PUDL citation
* Removes a couple of ill-formed eprints that generated dead links.

# Testing

* Ran the docs build locally and inspected the bibliography page.

### DIFF
--- a/docs/catalyst_cites.bib
+++ b/docs/catalyst_cites.bib
@@ -1,3 +1,18 @@
+@article{Schivley_2025,
+doi = {10.1088/2753-3751/ade548},
+url = {https://doi.org/10.1088/2753-3751/ade548},
+year = {2025},
+month = {nov},
+publisher = {IOP Publishing},
+volume = {2},
+number = {4},
+pages = {045006},
+author = {Schivley, Greg and Barone, Aurora and Blackhurst, Michael and Hidalgo-Gonzalez, Patricia and Jenkins, Jesse and Lugovoy, Oleg and Luo, Qian and Roberts, Michael J and Zheng, Rangrang and Wade, Cameron and Fripp, Matthias},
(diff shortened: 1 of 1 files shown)


docs/catalyst_cites.bib
```diff
@@ -1,3 +1,18 @@
...
...nfigurations are harmonized, and (2) the degree to which varying model configurations affect outputs. We harmonize the inputs to each model using PowerGenome and use clearly defined scenarios (policy conditions) and configurations (model setup choices). This allows us to isolate how differences in model structure affect policy outcomes and investment decisions. Our framework allows each model to be tested on identical assumptions for policy, technology costs, and operational constraints, allowing us to focus on differences that arise from inherent model structures. Key findings highlight that, when harmonized, models produce very similar capacity portfolios under current policies and net-zero scenarios, with less than 1% difference in system costs for most configurations. This agreement among models allows us to focus on how configuration choices affect model results. For instance, configurations with unit commitment constraints or economic retirement yield different investments and system costs compared to simpler configurations. Our findings underscore the importance of aligning input data and transparently defining scenarios and configurations to provide robust policy insights.}
```
cmgosnell, line 13: i won't die on this hill but i don't love having this one have an abstract. it is not consistent with the rest of them and if they all had abstracts it'd be so long.

zaneselvans: Many of them do have abstracts, but the abstracts don't currently get displayed on the citation page. I like having the abstracts since it gives a lot more detail about what the heck the paper was, without needing to go off site. I'd like to have the option in the future of something like a "read more" or triangle to expand the abstract on each entry. I agree that having abstracts for all of them and displaying them all the time would be excessively long.

cmgosnell (approved):

PR #4857, opened 2025-12-20
Cache Zenodo archives on S3 instead of GCS

### LINKED ISSUE #4856: Zenodo fails with "429 too many requests" error
# Overview

* The week of 2025-12-15 we started getting unit test and documentation build failures due to a Zenodo timeout.
* These failures only seemed to happen in circumstances where we were unable to access our cache of Zenodo archives in GCS.
* The Zenodo API is generally slow and flaky, so we didn't think too much of it, and tried work around the problem by ensuring we could access the GCS cache with appropriate credentials in a larger number of places.
* The problem has persisted, and there's something unusual about it. When our Zenodo retries ultimately fail, they fail with [error code 429: too many requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429):

```
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='zenodo.org', port=443):
Max retries exceeded with url: /api/records/17223513/files/datapackage.json/content
(Caused by ResponseError('too many 429 error responses'))
```

* The error can sometimes be reproduced locally by running one of our integration tests, designed to verify that we can in fact pull data directly from Zenodo (rather than from the GCS cache):

```console
pytest --no-cov test/integration/console_scripts_test.py::test_pudl_datastore
```

Which attempts to download a single partition of data directly from Zenodo using one of our smallest datasets, FERC Form 60:

```console
pudl_datastore --dataset ferc60 --bypass-local-cache --partition year=2020
```

* If I run that command once, and I haven't interacted with Zenodo in a while, it seems like it will often pass.
* But if I run it a couple of times in quick succession, it seems to quickly yield the 429 error.
* This behavior might also help explain the recently reliable failure of any Dagster run that would need to download a new version of one of our input datasets -- if the user doesn't have their GCP credentials set up right, or the ETL is defaulting to pulling from Zenodo rather than GCS, it might quickly fail, with a ChildProcessKilled error.
* The Zenodo developer documentation [describes their rate limits](https://developers.zenodo.org/#rate-limiting):
  * 60/min or 2000/hr for unauthenticated users.
  * 100/min or 5000/hr for authenticated users.
* However they also state "For our content and services to be available to everyone we have to make sure that our resources are being distributed fairly. To achieve this, we have rate-limiting measures in place which limit the number of requests users can perform in a time window. Depending on the complexity and load of each request, different endpoints have different limits configured."

## Hypotheses

* Something about our Zenodo usage has changed recently, increasing the number of requests that we are making.  We are somehow identifiable as a single entity.  We are now coming up against a pre-existing rate limit.
* Something about how PUDL makes requests has changed, increasing the number of requests required to engage in the same usage pattern.
* Zenodo has recently imposed a stricter limit on the number of requests that can be made on everybody.
* Zenodo can identify our usage and has imposed a stricter limit on our individual use in particular.
* Our Zenodo API key has been leaked and somebody else is using it to make a ton of requests.

## Potential Solutions

* Make sure we are accessing Zenodo as an authenticated user whenever possible, in the hopes of increasing the number of requests we can make.
* Make sure that we are *not* accessing Zenodo as an authenticated user, so that our aggregate usage in many different contexts is not all counted against the same rate limit.
* Do a better job of limiting how much data we download from Zenodo, and the contexts in which that downloading takes place. Really treat Zenodo as cold storage and rely entirely on the cloud-based cache. This could be annoying because to access the GCS cache you need to be authenticated.
* Switch to publicizing the most recent Zenodo archives via the AWS Open Data Registry in S3. This would avoid the need for authentication, and make it possible for other users to easily access the archives, as well as making them easy to use in a variety of other contexts (CI runs initiated by open source contributors, docs builds on RTD, etc.).  The `zenodo-cache-sync` workflow could be updated to also refresh S3, and the files in there could be given an annual, quarterly, or monthly lifecycle deletion rule to ensure that any data not actively in use by PUDL is not taking up space. With a 400-day expiration, the data in `gs://zenodo-cache.catalyst.coop` takes up about 80GB, so in the context of our 1TB S3 quota this should be doable.

zaneselvans:
# Overview

- Pulling data directly from Zenodo is flaky and slow, and best avoided whenever possible.
- We have a GitHub workflow that automatically caches new Zenodo archives in cloud storage as soon as their DOIs appear in our datastore module.
- Ideally, that would be the only time that we need to pull data directly from Zenodo.
- Until now, we have cached the data in GCS. Unfortunately, this means to access the cache, you need to authenticate, and we don't have the ability to provide credentials in all circumstances, e.g. when an open source contributor is trying to run the CI, or RTD is trying to do our documentation build.
- However, we do have free, public storage available in S3 through their open data registry, and the Zenodo archives that are in active use at any given time are only a few 10s of GB.
- This PR adds an `S3Cache` to the collection of resource caching classes, so that we can cache raw input data on S3, and access it without needing to authenticate.
- Because this cache method should be more performant than Zenodo in all circumstances and doesn't require authentication for reading data, it is now set as the default, pulling data from `s3://pudl.catalyst.coop/zenodo`
- The `zenodo-cache-sync` workflow is also modified to cache new Zenodo data in S3 rather than GCS.
- The nightly build script is modified to rely on the S3 cache instead of GCS.
- I changed the `--gcs-cache-path` argument to be `--cloud-cache-path` and it can take either S3 or GCS URLs now.
- The default in all cases is now to read from the Zenodo cache in S3.

Closes #4856 (hopefully...)

## Review Questions / Considerations

- Did I go too hard on switching to using the S3 cache as the default? The ability to read from GCS still exists, but won't be exercised unless someone asks for it specifically, and it won't work without credentials.
- The `zenodo-cache-sync` workflow no longer updates the cache in GCS and I was planning to delete it once this is up and running.  Does that seem reasonable? If not I should re-enable updating that cache.
- At least on small archives like the EIA-176 or FERC-60 (which I was using for testing) it's still possible to get 429 "too many requests" errors. I suspect because it's making so many requests in quick succession? I'm not sure yet if this will be a problem in production, but it seems independent of the S3 vs. GCS caching, and using S3 means that the only time we should ever have to deal with downloading from Zenodo is in the `zenodo-cache-sync` workflow. Thankfully that workflow does make incremental progress in uploading data, so even if it does get the 429 error, the next time it tries, it'll get a little further.

## Testing

- Disabled my local AWS credentials and was able to download cached data from S3 using `pudl_datastore`
- Built the docs locally with AWS credentials disabled locally.
- The docs build in CI and RTD docs builds are passing with the new S3 cache default.
- With my AWS credentials re-enabled locally I deleted the eia176 archive from the S3 cache and was able to repopulate it using `pudl_datastore --bypass-local-cache`. (though when I did it a second time in quick succession, I got a 429 error)
- Verified that the `zenodo-cache-sync` workflow can actually sync data to S3 from GHA
- Ran the full integration tests in CI to ensure that test coverage was high enough, and that it worked w/ the S3 cache.
- Ran a branch deployment that worked (but also I've mucked with the CLI since then)

## To-do list

- [x] Remove xfail label from datastore CLI test
- [x] Set 400 day lifecycle deletion rule for objects under `s3://pudl.catalyst.coop/zenodo`
- [x] Add `--s3-cache-path` argument to the CLI tools which take `--gcs-cache-path` currently.
- [x] Update documentation to reflect S3 cache rather than GCS cache
- [x] Update the release notes / quality of life improvements
- [x] Combine S3 and GCS cache options into a single `--cloud-cache-path` option to simplify API.
- [x] Run a branch deployment to make sure this didn't mess up nightly builds.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -27,7 +27,7 @@ Performance Improvements
 Quality of Life Improvements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-* Switched from caching Zenodo archives in GCS to AWS S3, using our free an public AWS
+* Switched from caching Zenodo archives in GCS to AWS S3, using our free and public AWS
   Open Data Registry bucket at ``s3://pudl.catalyst.coop/zenodo``. This will make it
   easier for open source contributors to run continuous integration tests, since no
   cloud credentials are required to download the raw data from S3, and they will not be
--- a/src/pudl/workspace/resource_cache.py
+++ b/src/pudl/workspace/resource_cache.py
@@ -106,7 +106,7 @@ class LocalFileCache(AbstractCache):
         """Adds (or updates) resource to the cache with given value."""
         logger.debug(f"Adding {resource} to {self._resource_path}")
         if self.is_read_only():
-            logger.debug(f"Read only cache: ignoring set({resource})")
+            logger.warning(f"Read only cache: ignoring set({resource})")
             return
         path = self._resource_path(resource)
         path.parent.mkdir(parents=True, exist_ok=True)
@@ -116,7 +116,7 @@ class LocalFileCache(AbstractCache):
     def delete(self, resource: PudlResourceKey):
         """Deletes resource from the cache."""
         if self.is_read_only():
-            logger.debug(f"Read only cache: ignoring delete({resource})")
+            logger.warning(f"Read only cache: ignoring delete({resource})")
             return
         self._resource_path(resource).unlink(missing_ok=True)
 
@@ -253,7 +253,7 @@ class S3Cache(AbstractCache):
             ClientError: for S3 errors
         """
         if self.is_read_only():
-            logger.debug(f"Read only cache: ignoring add({resource})")
+            logger.warning(f"Read only cache: ignoring add({resource})")
             return
 
         if self._unsigned:
@@ -275,7 +275,7 @@ class S3Cache(AbstractCache):
             ClientError: for S3 errors
         """
         if self.is_read_only():
-            logger.debug(f"Read only cache: ignoring delete({resource})")
+            logger.warning(f"Read only cache: ignoring delete({resource})")
             return
 
         if self._unsigned:
@@ -363,7 +363,7 @@ class LayeredCache(AbstractCache):
     def add(self, resource: PudlResourceKey, value):
         """Adds (or replaces) resource into the cache with given value."""
         if self.is_read_only():
-            logger.debug(f"Read only cache: ignoring set({resource})")
+            logger.warning(f"Read only cache: ignoring set({resource})")
             return
         for cache_layer in self._caches:
             if cache_layer.is_read_only():
@@ -378,7 +378,7 @@ class LayeredCache(AbstractCache):
     def delete(self, resource: PudlResourceKey):
         """Removes resource from the cache if the cache is not in the read_only mode."""
         if self.is_read_only():
-            logger.debug(f"Readonly cache: not removing {resource}")
+            logger.warning(f"Read only cache: not removing {resource}")
             return
         for cache_layer in self._caches:
             if cache_layer.is_read_only():


.github/workflows/pytest.yml
```diff
@@ -103,14 +103,6 @@ jobs:
...
           printenv | sort
 
-      - name: Set default GCP credentials
-        id: gcloud-auth
-        continue-on-error: true
-        uses: "google-github-actions/auth@v3"
-        with:
-          workload_identity_provider: "projects/345950277072/locations/global/workloadIdentityPools/gh-actions-pool/providers/gh-actions-provider"
-          service_account: "tox-pytest-github-action@catalyst-cooperative-pudl.iam.gserviceaccount.com"
-
```
zaneselvans, line 113: No authentication is required now, so these steps can be removed.

.github/workflows/zenodo-cache-sync.yml
```diff
@@ -50,23 +49,16 @@ jobs:
...
-        id: gcloud-auth
-        uses: "google-github-actions/auth@v3"
+      - name: Authenticate to AWS
+        uses: aws-actions/configure-aws-credentials@v5
         with:
-          workload_identity_provider: "projects/345950277072/locations/global/workloadIdentityPools/gh-actions-pool/providers/gh-actions-provider"
-          service_account: "zenodo-cache-manager@catalyst-cooperative-pudl.iam.gserviceaccount.com"
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: "us-west-2"
```
zaneselvans, line 57: Here we do need to authenticate, since we're potentially going to write to S3, if there are new archives that need to be cached.

.github/workflows/zenodo-cache-sync.yml
```diff
@@ -50,23 +49,16 @@ jobs:
...
-
-      - name: Update GCS cache with any new Zenodo archives
-        run: |
-          pudl_datastore --gcs-cache-path ${{ env.INTERNAL_ZENODO_CACHE_BUCKET }} --bypass-local-cache --loglevel DEBUG
-
-      - name: Sync internal and public caches
+      - name: Update AWS S3 cache with any new Zenodo archives
         run: |
-          gcloud storage --billing-project=catalyst-cooperative-pudl rsync -r ${{ env.INTERNAL_ZENODO_CACHE_BUCKET }} ${{ env.PUBLIC_ZENODO_CACHE_BUCKET }}
+          pudl_datastore --cloud-cache-path ${{ env.CLOUD_CACHE_PATH }} --bypass-local-cache --loglevel DEBUG
```
zaneselvans, line 61: Including `CLOUD_CACHE_PATH` in the CLI isn't actually necessary since it's now the default. Should I remove it?

docker/gcp_pudl_etl.sh
```diff
@@ -40,16 +39,13 @@ function run_pudl_etl() {
         alembic upgrade head &&
         ferc_to_sqlite \
             --loglevel DEBUG \
-            --gcs-cache-path gs://internal-zenodo-cache.catalyst.coop \
```
zaneselvans, line 43: The default is now equivalent to `--cloud-cache-path s3://pudl.catalyst.coop/zenodo` so I left it out of the arguments entirely.

docs/dev/testing.rst
```diff
@@ -150,13 +150,15 @@ looking at the ``custom options`` section:
...
-                        If set, use this GCS path as a datastore cache layer.
+   Custom options:
+     --live-dbs            Use existing PUDL/FERC1 DBs instead of creating temporary ones.
+     --tmp-data            Download fresh input data for use with this test run only.
+     --etl-settings=ETL_SETTINGS
+                           Path to a non-standard ETL settings file to use.
+     --bypass-local-cache  If enabled, the local file cache for datastore will not be used.
+     --save-unmapped-ids   Write the unmapped IDs to disk.
+     --ignore-foreign-key-constraints
+                           If enabled, do not check the foreign keys.
```
zaneselvans, line 161: Updated to reflect the current options on the script.  Also, note that I've removed the option to set a custom cloud cache path.

docs/dev/testing.rst
```diff
@@ -183,14 +185,15 @@ We use the ``src/pudl/package_data/etl_full.yml`` settings file to specify an ex
...
-data to a temporary datastore that's later deleted automatically:
+Zenodo <https://zenodo.org/communities/catalyst-cooperative>`__. A copy of that data
+is cached locally so that it can be re-used later without needing to be downloaded
+every time. Because downloading data directly from Zenodo can be slow and unreliable,
+by default we download from a cached copy in Amazon's S3 storage, in a free bucket
+provided by the AWS Open Data Registry at ``s3://pudl.catalyst.coop/zenodo``.
+
+You can also force the tests to download of a fresh copy of the data to use just once,
+even if you already have a local copy, which is useful when you are testing the
+datastore functionality specifically.
```
zaneselvans, line 196: Update to reflect new default behavior.

src/pudl/etl/cli.py
```diff
@@ -68,16 +68,14 @@ def get_pudl_etl_job():
...
-        "Load cached inputs from Google Cloud Storage if possible. This is usually "
-        "much faster and more reliable than downloading from Zenodo directly. The "
-        "path should be a URL of the form gs://bucket[/path_prefix]. Internally we use "
-        "gs://internal-zenodo-cache.catalyst.coop. A public cache is available at "
-        "gs://zenodo-cache.catalyst.coop but requires GCS authentication and a billing "
-        "project to pay data egress costs."
+        "Load cached inputs from cloud object storage (S3 or GCS). This is typically "
+        "much faster and more reliable than downloading from Zenodo directly. By "
+        "default we read from the cache in PUDL's free, public AWS Open Data Registry "
+        "bucket."
```
zaneselvans, line 78: Should I explicitly include the default S3 path in the help message?

src/pudl/metadata/classes.py
```diff
@@ -1009,7 +1009,6 @@ def add_datastore_metadata(self) -> None:
         """Get source file metadata from the datastore."""
         dp_desc = Datastore(
             local_cache_path=PudlPaths().data_dir,
-            gcs_cache_path="gs://zenodo-cache.catalyst.coop",
```
zaneselvans, line 1012: Supplanted by the default `cloud_cache_path` value.

src/pudl/resources.py
```diff
@@ -36,10 +36,10 @@ def ferc_to_sqlite_settings(init_context) -> FercToSqliteSettings:
 
 @resource(
     config_schema={
-        "gcs_cache_path": Field(
+        "cloud_cache_path": Field(
             str,
-            description="Load datastore resources from Google Cloud Storage.",
-            default_value="",
+            description="Load datastore resources from this GCS or S3 path.",
+            default_value="s3://pudl.catalyst.coop/zenodo",
```
zaneselvans, line 42: Again, default to the thing we know works.

test/unit/settings_test.py
```diff
@@ -356,7 +356,6 @@ def test_partitions_with_json_normalize(pudl_etl_settings):
 
 
 @pytest.mark.slow
-@pytest.mark.xfail(reason="Zenodo is being particularly flaky. @zaneselvans 2025-12-19")
```
zaneselvans, line 359: This was just here to mitigate the Zenodo 429 errors causing the test to fail.

test/conftest.py
```diff
@@ -77,11 +77,6 @@ def pytest_addoption(parser):
         default=False,
         help="Path to a non-standard ETL settings file to use.",
     )
-    parser.addoption(
-        "--gcs-cache-path",
-        default=None,
-        help="If set, use this GCS path as a datastore cache layer.",
-    )
```
zaneselvans, line 84: Is there any scenario where we would actually want to run the tests without using the S3 cache? (Local cache will still always be preferred if it exists)

src/pudl/workspace/resource_cache.py
```diff
@@ -167,6 +171,152 @@ def contains(self, resource: PudlResourceKey) -> bool:
...
+    def add(self, resource: PudlResourceKey, content: bytes):
+        """Adds (or updates) resource to the cache with given content.
+
+        Raises:
+            RuntimeError: if cache is read-only or using unsigned requests
+            ClientError: for S3 errors
+        """
+        if self.is_read_only():
+            logger.debug(f"Read only cache: ignoring add({resource})")
+            return
```
zaneselvans, line 257: Will this be a confusing silent failure? Should trying to write to a read only cache produce an error instead?

test/unit/workspace/resource_cache_test.py
zaneselvans, line 1: These are mostly LLM generated unit tests that check lots of micro-behavior with mocking.

.github/workflows/zenodo-cache-sync.yml
zschira: I like making args like this explicit even if it's not necessary, so I'm good with leaving as is.

docs/release_notes.rst
```diff
@@ -2,6 +2,41 @@
...
+Bug Fixes
+^^^^^^^^^
+
+Performance Improvements
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Quality of Life Improvements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* Switched from caching Zenodo archives in GCS to AWS S3, using our free an public AWS
```
zschira:
```suggestion
* Switched from caching Zenodo archives in GCS to AWS S3, using our free and public AWS
```

src/pudl/etl/cli.py
zschira: I think it's fine to leave it out here since you mention that it defaults to the S3 cache.

src/pudl/workspace/datastore.py
```diff
@@ -335,19 +335,34 @@ def __init__(
...
-            except (DefaultCredentialsError, OSError) as e:
-                logger.info(
-                    f"Unable to obtain credentials for GCS Cache at {gcs_cache_path}. "
-                    f"Falling back to Zenodo if necessary. Error was: {e}"
+
+        if cloud_cache_path is not None:
+            parsed_url = urlparse(cloud_cache_path)
+            if parsed_url.scheme == "s3":
+                logger.info(f"Adding S3 cache layer at {cloud_cache_path}")
+                self._cache.add_cache_layer(resource_cache.S3Cache(cloud_cache_path))
```
zschira, line 349: Out of scope for this PR, but I think we could really simplify things if we created a generic `fsspec` chache class that could be used for GCS, S3, and local paths.

src/pudl/workspace/resource_cache.py
zschira: I could see this being confusing. Maybe a middle ground would be to just raise the log-level to `Warning` to guarantee that this isn't totally silent?

src/pudl/workspace/resource_cache.py
```diff
@@ -167,6 +171,152 @@ def contains(self, resource: PudlResourceKey) -> bool:
...
+
+    def delete(self, resource: PudlResourceKey):
+        """Deletes resource from the cache.
+
+        Raises:
+            RuntimeError: if cache is read-only or using unsigned requests
+            ClientError: for S3 errors
+        """
+        if self.is_read_only():
+            logger.debug(f"Read only cache: ignoring delete({resource})")
```
zschira: Again, maybe raise this log-level.

test/conftest.py
zschira: I don't really see any reason to use GCS over S3. Seems like it will be similar levels of performance / reliability, but with no credentials concerns

test/unit/workspace/resource_cache_test.py
```diff
@@ -204,3 +206,252 @@ def test_read_only_cache_ignores_modifications(self):
         assert not lc.contains(r_new)
         assert not self.cache_1.contains(r_new)
         assert not self.cache_2.contains(r_new)
+
+
+class TestS3Cache:
```
zschira, line 211: It seems like a lot of these tests could be generalized to all `AbstractCache`s, but I'm ok leaving that until we perhaps just create a generic `fsspec` cache in the future.

zaneselvans:
Agree that an `fsspec` based cache would simplify a lot of this. Maybe I should have done that instead! Ah well.  I'll create [an issue](https://github.com/catalyst-cooperative/pudl/issues/4860).

> How much total space will this require, and how much will that impact our free S3 storage?

I've set the Zenodo cache to expire and automatically get deleted with a lifecycle rule after 400 days, so that datasets which get updated on an annual basis will hang around until their next update. With that rule in place the cache takes up about 80GB of our 1TB quota. Each data release we're putting out right now takes up about 25GB, so if we keep doing monthly releases, we'll have ~300GB/yr of archived releases that we keep around. So we'll have to start flushing older ones at some point.  But even with the EQR (~100GB) and retaining 2 years of monthly release (~600GB), plus `/stable`, `/nightly` and `/eel-hole` (75GB) we would have a couple hundred GB to spare, so I think this is okay.  We can always go to retaining fewer old data releases and could be more aggressive about culling older versions of the Zenodo archives if we wanted to.

I will update the logging output for those quiet failures to be less quiet and fix that typo, and merge.

zaneselvans:
Failed on testing `zenodo_data_release` script with micro release to Zenodo sandbox... which has been passing just fine! Going to try again.

PR #4859, opened 2025-12-22
Standardize transmission structure types and materials in FERC Form 1 Schedule 422

### LINKED ISSUE #4658: Standardize FERC 1 transmission (Schedule 422) `supporting_structure_type` column
# Overview

* The `supporting_structure_type` column in the `out_ferc1__yearly_transmission_lines_sched422` table is a free form text field that is supposed to describe the type of support being used for wires on a given power line.
* It typically describes both the material (e.g. "steel" or "wood" or "cement") and the shape of the structure (e.g. "tower" or "pole").
* However, there are more than 1300 unique values in the column right now, which makes it useless for analysis purposes.
* Given that the data includes a mix of structure only and material only as well as structure-and-material values, it may be more useful to split this out into two separate fields, one which describes the material and another that describes the type of structure.

## Design Thoughts

* We already do this kind of free form categorical value cleanup on the messy fuel types that are reported in FERC Form 1 and can use the same pattern on this column. See the [`categorize_strings()`](https://github.com/catalyst-cooperative/pudl/blob/main/src/pudl/transform/classes.py#L487) method
* This would mean manually compiling a dictionary that maps a canonical value (e.g. "steel_tower") to the set of observed values which we think ought to really map to the same semantic category.
* However, given the split between structure type and structure material, we may want to have 2 dictionaries, one that maps observed values to a material (e.g. "steel") and another that maps to structure types (e.g. "tower").
* Inevitably, there are going to be some values that just don't make any sense and have to get set to NULL. The original data also already includes a large number of NULL values.  But at least it'll be more usable.
* If there are unique transmission line identifiers or consistent names, we could also forward / backward fill values across years, since the structure type seems likely to be pretty durable over time.  Assuming that some lines have reported values in some years and not others... and we don't see lots of examples of lines that seem to change their type.

## Data Thoughts

[The FERC Form 1 instructions](https://catalystcoop-pudl.readthedocs.io/en/nightly/_downloads/85f442fbf290270270ad3dd4a42b2a1f/ferc1_blank_2025-07-31.html) for filling out the supporting structure type in Schedule 422 say:

> Indicate whether the type of supporting structure reported in column (e) is: (1) single pole wood or steel; (2) H-frame wood, or steel poles; (3) tower; or (4) underground construction If a transmission line has more than one type of supporting structure, indicate the mileage of each type of construction by the use of brackets and extra lines. Minor portions of a transmission line of a different type of construction need not be distinguished from the remainder of the line.

### Data Sample

What are the most common strings that show up?

```python
(
    get_polars("out_ferc1__yearly_transmission_lines_sched422").select(
        pl.col("supporting_structure_type")
    ).to_series()
    .str.to_lowercase()
    .value_counts()
    .sort(by="count", descending=True)
    .head(50)
)
```

|    | supporting_structure_type   |   count |
|---:|:----------------------------|--------:|
|  0 |                             |   57743 |
|  1 | steel tower                 |   40140 |
|  2 | sp                          |   39757 |
|  3 | st                          |   32898 |
|  4 | steel pole                  |   31312 |
|  5 | wood pole                   |   23656 |
|  6 | 1                           |   19008 |
|  7 | wp                          |   15614 |
|  8 | tower                       |   14634 |
|  9 | 3                           |   13726 |
| 10 | h                           |   13333 |
| 11 | (3) steel                   |   12458 |
| 12 | s-t                         |   11836 |
| 13 | 2                           |    9415 |
| 14 | w-h-f                       |    8998 |
| 15 | wood h-frame                |    8597 |
| 16 | steel twr                   |    8021 |
| 17 | h-frame                     |    7420 |
| 18 | underground                 |    6870 |
| 19 | wood h frame                |    6758 |
| 20 | t                           |    6740 |
| 21 | various                     |    6459 |
| 22 | ug                          |    6232 |
| 23 | steel hfrm                  |    5717 |
| 24 | polesteel                   |    5402 |
| 25 | wood                        |    4519 |
| 26 | wood h                      |    4424 |
| 27 | s-steel                     |    4283 |
| 28 | wh                          |    4151 |
| 29 | w-h fr.                     |    4112 |
| 30 | single w.p.                 |    3891 |
| 31 | steel                       |    3870 |
| 32 | polewood                    |    3704 |
| 33 | single pole                 |    3271 |
| 34 | latticetower                |    3224 |
| 35 | steel towers                |    3090 |
| 36 | steel h-frame               |    3045 |
| 37 | wood - h                    |    3020 |
| 38 | (2) wood                    |    3000 |
| 39 | conc hfrm                   |    2864 |
| 40 | steel poles                 |    2801 |
| 41 | s-s-p                       |    2664 |
| 42 | s-p                         |    2648 |
| 43 | (1) steel                   |    2331 |
| 44 | s-w-p                       |    2275 |
| 45 | 0                           |    2255 |
| 46 | wd-h-frame                  |    2247 |
| 47 | ts                          |    2169 |
| 48 | u                           |    2137 |
| 49 | hfw                         |    2122 |

### Primary Categories

Most common materials:
- Steel --> `steel`
- Wood --> `wood`
- Concrete --> `concrete`

Most common structures:
- Pole (1) --> `pole`
- H-frame (2) --> `h_frame`
- Tower (3) --> `tower`
- Underground (4) --> `underground`

## Success Criteria

* The column has been replaced by 2 separate columns: `supporting_structure_type` and `supporting_structure_material`.
* Each of the new columns has small and meaningful number of categories that are intelligible and useful for analysis.
* All originally reported values are either categorized clearly into the above two columns, or set to `pd.NA`
* If new values appear in future years, they cause the ETL to fail until we categorize them.

dd921:
# Overview

Closes #4658.

The `supporting_structure_type` column in the `out_ferc1__yearly_transmission_lines_sched422` table contained over 1,300 unique free-form text values, making it unusable for analysis. The column mixes structure type information (pole, h-frame, tower, underground) with material information (steel, wood, concrete) in a single field. According to FERC Form 1 instructions, these should be separate concepts, and splitting them into two columns will make the data much more useful for analysis.

## What did you change?

1. **Created category dictionaries** (`src/pudl/transform/params/ferc1.py`):
   - Added `SUPPORTING_STRUCTURE_TYPE_CATEGORIES` mapping 50+ observed values to canonical structure types: `pole`, `h_frame`, `tower`, `underground`
   - Added `SUPPORTING_STRUCTURE_MATERIAL_CATEGORIES` mapping 50+ observed values to canonical materials: `steel`, `wood`, `concrete`
   - Both dictionaries follow the same pattern as `FUEL_CATEGORIES`, with comprehensive coverage of abbreviations, codes, and variations from the data sample

2. **Implemented transformation** (`src/pudl/transform/ferc1.py`):
   - Added `split_supporting_structure()` method to `TransmissionLinesTableTransformer` that:
     - Normalizes strings before categorization (handles mixed-case input from raw FERC data)
     - Splits the original `supporting_structure_type` column into two new columns: `supporting_structure_type` (structure) and `supporting_structure_material` (material)
     - Validates that all values are categorized, raising a `ValueError` if new uncategorized values appear (ensures ETL fails on new data until categories are updated)
   - Updated `transform_main()` to call the split method before parent transforms

3. **Updated metadata**:
   - Added `supporting_structure_material` field to both `core_ferc1__yearly_transmission_lines_sched422` and `out_ferc1__yearly_transmission_lines_sched422` table schemas in `src/pudl/metadata/resources/ferc1.py`
   - Added field descriptions in `src/pudl/metadata/fields.py` documenting the categorized values for both columns
   - Updated `supporting_structure_type` description to reflect its new categorical nature

4. **Updated dbt schemas**:
   - Added `supporting_structure_material` column to both dbt schema files:
     - `dbt/models/ferc1/core_ferc1__yearly_transmission_lines_sched422/schema.yml`
     - `dbt/models/ferc1/out_ferc1__yearly_transmission_lines_sched422/schema.yml`

5. **Added tests** (`test/unit/transform/ferc1_test.py`):
   - `test_split_supporting_structure()`: Verifies correct categorization of various structure type values including combined values like "steel tower" and "wood pole"
   - `test_split_supporting_structure_fails_on_uncategorized()`: Verifies that the ETL fails appropriately when uncategorized values are encountered

## Documentation

# Testing

- Added unit tests that verify:
  - Correct categorization of structure types (pole, h_frame, tower, underground)
  - Correct categorization of materials (steel, wood, concrete)
  - Proper handling of combined values (e.g., "steel tower" → structure: "tower", material: "steel")
  - Proper handling of values without material specification (e.g., "tower" → structure: "tower", material: NA)
  - Proper handling of NA values
  - ETL failure on uncategorized values

- All tests pass locally
- dbt schema drift test passes (schemas updated to match metadata)
- The implementation follows existing patterns used for fuel type categorization

A reviewer can verify this by:
1. Running the unit tests: `pytest test/unit/transform/ferc1_test.py::test_split_supporting_structure test/unit/transform/ferc1_test.py::test_split_supporting_structure_fails_on_uncategorized`
2. Running the dbt schema drift test: `pytest test/unit/dbt_schema_drift_test.py::test_dbt_schema_drift`
3. Reviewing the category dictionaries to ensure comprehensive coverage of common values
4. Checking that the transformation handles edge cases (NA values, mixed-case input, etc.)

## To-do list


src/pudl/metadata/fields.py
```diff
@@ -7359,7 +7359,17 @@
...
+            "Type of supporting structure for the transmission line. "
+            "Categorized values: pole, h_frame, tower, underground."
+        ),
+    },
+    "supporting_structure_material": {
+        "type": "string",
+        "description": (
+            "Material used in the supporting structure for the transmission line. "
+            "Categorized values: steel, wood, concrete."
+        ),
```
zaneselvans, line 7565: We should add enum constraints to both of these columns now that they are cleaned up, e.g.

```suggestion
    "supporting_structure_material": {
        "type": "string",
        "description": (
            "Material used in the supporting structure for the transmission line. "
            "Categorized values: steel, wood, concrete."
        ),
        "constraints": {
            "enum": ["steel", "wood", "concrete"]
        }
```

zaneselvans:
Hey @dd921! Thanks for taking on this issue. I'll try and get a review done tomorrow.

dd921:
> Hey @dd921! Thanks for taking on this issue. I'll try and get a review done tomorrow.

Thanks, @zaneselvans! I tried to follow the contribution guide as best I could - sorry for not connecting with you before beginning the task. Please let me know if there are any norms or processes I should adhere to moving forward that I may have missed!

zaneselvans:
Hey @dd921 sorry the holidays kind of got away from me. I'll take a look at this ASAP.

Note that there are some big changes to the development environment that just got merged in from `main` -- we've switched to using `pixi` rather than `mamba`, `conda-lock`, and `make` to manage the environment.  The [dev setup docs](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/dev_setup.html) have been updated to reflect the new arrangement.

src/pudl/transform/ferc1.py
```diff
@@ -4604,9 +4613,65 @@ class TransmissionLinesTableTransformer(Ferc1AbstractTableTransformer):
     table_id: TableIdFerc1 = TableIdFerc1.TRANSMISSION_LINES
     has_unique_record_ids: bool = False
 
+    def split_supporting_structure(self, df: pd.DataFrame) -> pd.DataFrame:
```
zaneselvans, line 4609: Structurally (ha) this all looks good.

src/pudl/transform/params/ferc1.py
```diff
@@ -772,6 +772,168 @@
 Case is ignored, as all fuel strings are converted to lower case in the data set.
 """
 
+SUPPORTING_STRUCTURE_TYPE_CATEGORIES: dict[str, set[str]] = {
```
zaneselvans: Both of these dictionaries look good structurally and thematically, but they seem suspiciously small.  IIRC there were ~1300 unique values in the column so I would expect there to be a similar number of values that get mapped here (though many of them might get mapped to NA if they're uninterpretable.)

I tried materializing the Schedule 422 table locally, and got a prolific `ValueError` complaining about uncategorized values (see below).

My guess is that you probably ran just the fast ETL, which only processes a couple of years of data, and so only encountered and categorized a small subset of the unique values.

```
ValueError: core_ferc1__yearly_transmission_lines_sched422: Found uncategorized supporting_structure_type values. Structure: {'', 'wood, lattice', 'wd h-fr; stl sp', '(1) (15.26)', 'wfs (30)', 'st-18.75', 'h-3.05', '1-0.92', 'tower h frame', 's pole&hframe', '1-0.8941', '2s', 'hf-w,sp-w', '1-8.26', '2-3.2', 'stl pl-0978', '1-16.19', '3-3.61', 'st (41)', 'sp-c', 'wh frame', 'c-u-g', '(2); (1); (3)', 'others t', '1-11.52', 'd; e; b; c; a', 'various oh/og', 'hf-w, sp-s', '1-0.451', '(3) steel (c)', '(steel pole)', '3-2.57', 'b-sa/dcsp', 'sp-hf', '1-.018', 'sp; sph; wp', 'steel - sp', 'p-3.12', '2 0.24', 'sttel pole', '2-4.8', '1-1.41', 'wd h-fr/sp,stl', 'steel monopole', 'a; d; c', 'z-6.56', 'st-1.26', 'included in 1', 'h. fr. wood', '(1); (4); (5)', '3,1s,w', '1-5.52', 'pp-3.76', '2-37.22', '1-2.98', 'ss-3.28', '2-1.494', 'twr on 0977', 'tu-1', 'sp-w & sp-s', '2-0.01', '(1) steel (a)', 'tower/concrete', '1 (3)an sub to sub', 'no structures', 'h-frame alumin', 'poles wood', '3 36.46', '1-2-3', '3-0.95', 'sp .08', 'st. lat. h.', 'guyed-v alumin', 'tower,stlh-frs', 'sttc-1.82', '1-4.79', '2-48.16', 'wd-stl-pole', 'h-.32;st-.07;', '3-13.3', 'e; d; a; c', 'st-1.65', 'b-3.49;h-9.54', 'wood s. p.', '1-8.48', '2-28.67', 'p6.46', '2 23.06', 'pp-.80', 'steel pole & wood h', '2-76.68', 'sgl. wood pole', 'hf,wp', '2-5.0983', 'steelpoletowe', 'underground 1', 'overhead woo', 'ref ln129', 'h-9.01', 'dc wood pole', 'wood/stl/concr', 'hfs', '(5); (1); (3)', 'steel towers"', '1-16.06', 'h-41.89', 'h-19.93', 'stl pole-0977', 'st,wp,sp', '1-16.43', '1-3.23', 'sgl steel pole', 'ts (35)', 's- tower', '2-22.93', '2-48.95', '2w, 1s, 2s, 3', 'st,whf,sp', 'p-7.10', '2-85.6', 'steel h fr', 'h-frm stl pole', 's14.13', 'steel-sp', 'wp-steel', 'ssp swp', 'wp; twr', 'h-wood and tos', '(s-p) (s-a)', 'ts (44)', 'sd-2.72', '3 10.73', 'a; c; d', '2-35.36', 'underbuild', 'underbuilt on', 'wood/steel', 'al-lpsg', 'st. tower/h-wood', 'r-spwd', '1-2.2031', '(1); (4); (3)', 'kfw 21.59', '(1); (2); (4); (3)', '2-11.84', 'tw on 0984', 'dc lattice', 'sp-wood/h-wood', '3 6.48', 'hf,sp & st', '2-17.78', '1-24.882', '1-3.82', '2 16.70', '1-2.03', '2-11.48', '0.00', 'single wp', '2 16.04', '(1) (0.06)', '1-2.8275', '2-6.61', 'st., sgle pole', 'c, a, b, e', 'switch', '2-1.7', 'cyl pole', '1-19.69', '2-115.47', 'kfw 115.64', '2 pole h', 'steel h_frame', 'p-9.15', 'st-.13', 'tower&stl h-fr', '1-39.75', '(1) (11.56)', 'swp ssp others', '(steel tower', '1-32.99', '2 25.11', '1-6.85', '1 ban sub to sub', 'steel pole/to', 'c; a; f', '1-25.9462', '1-0.78', '2-6.5517', '1-14.1567', 'bchp-b2sa', 'st-lpsg', 'wdpsc', 'stlh-fr/stl sp', 'wph; twr; wp;', 'sngl wood pole', '1-6.95', 't swp ssp', '2-37.7', 'steel guyed', 'stsc', 'sp: twr', 'hf,sp ,st & wp', 'h-22.10;st-.25', '1-0.2343', '1-0.63', 'wph; wp; sp', '1-2.96', '1-0.437', '& steel frame', '2-1.5899', 'twr on 0976', '1-1.51', 'c; d; a; f; b', '1-0.93', '(2) 9.98', '1-12.08', '1,2,3,', '1,2,3', 'hwp,st', 'cp - hf', 'pp-3.86', 'stl-wd-h-fr', '1-0.6', 'ts (26)', 'total', 'wooh h-frame', '1-14.94', '3-0.13', 'wood spsc', '1-1.28', '1-4.85', '1-3.06', '1-3.37', '& steel poles', '(1) (0.7)', '2-101.81', '1-5.5962', 's. stl. pole', '2-4.5', 'corten pole', '2-33.87', 'hh-3.50;', 'sgl.pole-w', 'h-3.03;p-2.58', 'scst, whf', 'h frame wp', 'hfs (14)', 'wp-hp', 'h-2.62', 'ts (32)', '3 20.34', 'tw on 0899', '1-2.1', 'a; d; e; b', '2 pole k', '1-2.39', 'steel pole-w', '2-28.75', 'steel pole and frame', 'h-45.97', 'kfw 38.48', 'b', 'alumin. towers', '1-32.85', '1-5.3', '1-18.72', 'hf,sp,wp,st', 'hfs (20)', 'towers (ab)', 'h.frame conrte', '2w,4', 'pp-.96;hh-6.52', '1-3.53', 'ss-2.43', 'sp 10.44', '1-2.93', 'stl pl/stl-h-f', 'concrete&steel', 'single concr', 'kfw 104.66', '(3) (0.79)', 'towers / wf. .', '-1', 'guyed', '2-10.1803', 'h-3.12', 'a; c; e', '2 15.84', 's-13.33', '1-8.73', '1w, 2w,1s, 2s', 'swp other', '1-5.88', 'lattice towe', 'stl.-lattice', '2 pole str.', 'h-14.26;p-.54', 'e; a; b; c', 'wfs (16)', '2-34.12', 's-lattice', 'rhes-mod,pset', 'guyed ste', '(w-h-f)(s-p)', 's tower', 'sp,st,whf,w', '1-4.0138', 'p-19.31', ...
```

test/unit/transform/ferc1_test.py
```diff
@@ -1315,3 +1316,95 @@ def test_filter_for_freshest_data_xbrl(df):
...
+        {
+            "supporting_structure_type": [
+                "steel tower",
+                "wood pole",
+                "steel h-frame",
+                "underground",
+                "tower",
+                "h-frame",
+                "pole",
+                pd.NA,
```
zaneselvans: Might want to throw it some examples with materials but no structure type to make sure that works. Like just "steel" and "wood" and "concrete" in isolation.

zaneselvans (changes requested):
Now that you've got a transform that seems to be doing all the right steps, you'll need to try and run it on all the years of data and categorize all the other values.  I suspect you just ran the fast ETL instead of the full ETL and so only hit a few of the possible values in this column.

You may need to re-run the FERC database extraction and make sure it extracts all years of available data: 1994-2020 for DBF, and 2021-2024 for XBRL.

Then in the Dagster UI you can just materialize the Schedule 422 assets, which will be very fast -- no need to run the whole ETL.

Looking at the unrecognized values that came up in the error message it also looks like there are a number of values that imply more than one structure or material type. There are so few potential combinations that it might be reasonable to have a categorical value for each observed combo of materials or combo of structures, if they can be clearly identified.

While I was looking at the string categorization dictionaries I saw there was a type hint error (pre-existing) and fixed it. I've pushed the change to your branch.

## To Do

* Capture and categorize the missing structure & material values across all years of data.
* Add appropriate ENUM constraints to the material and structure fields. (see code suggestion for one of them)
* Add some variety to the test cases (see note).

src/pudl/transform/params/ferc1.py
dd921: > My guess is that you probably ran just the fast ETL, which only processes a couple of years of data, and so only encountered and categorized a small subset of the unique values.

Yep thats correct. I wanted to make sure I was on the right track with this PR before doing all the extensive validation/documentation updates, etc. I will try to execute the full ETL and also update the dictionaries as I uncover new values. 

Also - if I have any issues with the new dev setup or running the full ETL I may reach out, but I'll try to get these updated soon. Thanks! @zaneselvans

dd921:
@zaneselvans please temporarily ignore [f3abc17](https://github.com/catalyst-cooperative/pudl/pull/4859/commits/f3abc17d70170675ceaae01d3d5651118d3da8d4). 

I was trying to figure out why I was seeing significantly more than 1300 unique values for `supporting_structure_type`. My eyes deceived me, and I'd been working with materialized values for `core_ferc1__yearly_transmission_lines_sched422` rather than  `out_ferc1__yearly_transmission_lines_sched422`.

I am working on revising this now using the correct dataset, but I am still seeing ~5300 unique values in the `supporting_structure_type` column (pretty much the same as what I saw before).

Is there perhaps some context I'm missing here or something I'm overlooking (assuming there should be an order of magnitude fewer unique values)?

zaneselvans:
The `core_ferc1` table is the right one to work with -- the `out` table is derived from it (and will just have some additional columns merged on for human readability).  The 1300 number could totally have been wrong. It was just a vague recollection. Mostly I just remembered it being much larger than the number of values that were categorized in your initial version.

Skimming over the enormous list in here it seems like there are many values that indicate multiple structures and/or materials, so a simple mapping of each category to a single material and a single structure type seems like it may not be appropriate. It seems like we may have to choose between listing all the values indicated, or none of them (setting ambiguous values to NA). It's not great from a database normalization point of view, but we might want to consider using lists rather than atomic values in the record. Though this isn't something that the mapping function supports currently.

For the patterns that recur we probably want to write up whatever the heuristics are that we're using for the categorization, e.g. based on the categorical numbers that show up.

It may also be the case that a large majority of data can be categorized with just the simple, straightforward values, so the marginal value of dealing with the messy ones may not be worth the effort.

I would start by categorizing only the straightforward values -- the ones that clearly map to one of the expected structures and/or materials, and put all the multi-category or ambiguous values in the `na_category` and see what fraction of the records are assigned a value for each of material and structure type. If it's like 90% of the values, then maybe that's good enough. If it's significantly lower, one non-destructive option would be to leave the original unstructured value in place so we don't prevent users from being able to interpret it themselves if they want to, and create two new clean columns, one for structure type and one for material, that are only filled in for the unambiguous values. That would be an acceptable initial outcom.

dd921:
@zaneselvans 
When I materialize `core_ferc1__yearly_transmission_lines_sched422` in dagster and evaluate the updated mappings (not yet committed here) I get the distributions shown below.

 I've also attached a `.txt` file [here representing the top 500 values,](https://github.com/user-attachments/files/24485360/top500supporting_struct_type.txt)  there is an extremely long tail given 5300 unique overall values.

Given this information, I'd lean towards the "obvious" mappings with a default to NaN for ambiguous or unintelligible values. I wonder though, if you still want to explciitly write every unknown value  in `transform/params/ferc1.py` or just leave them to default to `np.nan` rather than giving it a string category (currently "na_category")

The solution I have currently also retains an "original" `supporting_structure_type_original` column.

Screenshot 2026-01-07 at 7 53 39 PM
Screenshot 2026-01-07 at 7 53 51 PM

zaneselvans:
If there's a long tail of unintelligible values, or strings that seem to indicate multiple values rather than one, then I think retaining the original values in one column and mapping the clear / single-valued ones to the appropriate structure type and material seems fine.

It's not exactly aesthetically pleasing, but I think we do need to include all of the observed bad/weird/ambiguous values in the explicit mapping. The problem with defaulting to NA rather than enumerating all of the unmappable values is that next year when we get new filings with new values we've never seen before, some of them might be clearly mappable to the expected categories, and if we just default to NA for any unmapped value, then we'll never know it. The idea behind requiring every observed value to be explicitly mapped to *some* value (even if it's `na_category`) is that an error will be raised when we encounter never before seen values, and will be forced to categorize them appropriately.

dd921:
> It's not exactly aesthetically pleasing, but I think we do need to include all of the observed bad/weird/ambiguous values in the explicit mapping

Totally understand and agree, I'll play around with it a bit more an see if I can't pull a few more into defined categories.
Work has been ramping back up quickly in the new year so I might not be able to look at this again until tomorrow night, but I'll try to have another version for review by end of Friday.

On a slightly tangential note, I was having a bit of a hard time executing the full ETL pipeline because of my system specs (M1, 8GB RAM). I was able to limit the parallel workers to prevent it from crashing outright, but I killed the job after ~2 hours. I understand it could take close to 2 hours with nearly double the memory so I'm not sure how long it will take me to execute. Are there any recommendations for development on free cloud services with slightly larger instances than my local (thinking codespaces or something similar)?  @zaneselvans

zaneselvans:
@dd921 Oooh, yeah there are some individual assets that take more than 8GB of memory, so it will probably struggle on that machine. I wouldn't worry about running the whole ETL for these assets -- they're not deeply entangled with anything else outside of the FERC Form 1 tables. You can just run the ones that are up or down stream of the tables that you're changing.  In the Dagster UI you can use a selection like `+key:*sched422*+` to select anything that could be affected by the change. You'll still need the complete raw FERC Form 1 DBs. But if those give you trouble you could also download them and the associated metadata files directly from `s3://pudl.catalyst.coop/nightly/ferc1*` and drop them in your `$PUDL_OUTPUT` (and unzip the SQLite files) and that would probably be good enough. But it sounds like you've already managed to generate those DBs.

dd921:
@zaneselvans When it comes to original NA values, where  `supporting_structure_type_original` is NA from the start, should we leave it as such across the mapped columns as well? Here is the distribution adding "na_category", where the "`nan`" count is the proportion of actual original nulls compared to the screenshot in the last comment.

image

zaneselvans:
I'm not sure what you're asking. Are you wondering whether it should be possible to distinguish between a reported NA value, and a values that has been set to NA because it wasn't mappable?

If the original value is NA, then it should end up NA in the mapped column.
If the original value can't be mapped to one of the expected categories, then it should also end up NA in the mapped column.

I think any value that gets mapped to `na_category` should end up being set to NA after the mapping process, right?

Can you give me a concrete example of what you're imagining?

dd921:
> I think any value that gets mapped to na_category should end up being set to NA after the mapping process, right?

You're totally right - I think I confused myself when I was summarizing the data (in the first screenshot I shared) - the values of `nan` are all records that mapped to `na_category` (whether they were entered NA to begin with, or were unable to be categorized from the original entry).

For some reason I was expecting the values to contain the string `na_category` category, but that doesn't make sense as we're casting to NA. Disregard the message above. I will push my changes today, thanks for all the help here!

src/pudl/metadata/fields.py
```diff
@@ -7357,19 +7357,32 @@
...
     "supporting_structure_type": {
         "type": "string",
         "description": (
             "Type of supporting structure for the transmission line. "
-            "Categorized values: pole, h_frame, tower, underground."
+            "Categorized values: pole, h_frame, tower, underground. "
+            "Only filled for unambiguous values; set to NA for multi-category or ambiguous values."
         ),
+        "constraints": {
+            "enum": ["pole", "h_frame", "tower", "underground", "na_category"]
```
zaneselvans: `na_category` shouldn't be included in the enum constraint, since in the finished outputs it should have been converted to a true null value.

src/pudl/metadata/fields.py
zaneselvans: Here, as above, `na_category` should not be included, as it ought to get replaced with a true NULL value before being written out.

migrations/versions/a83ab61fa2e0_add_ferc1_transmission_materials_and_.py
zaneselvans, line 1: It seems like this is the only Alembic migration that actually has an effect. Ideally we would consolidate all of these migrations into a single one. Alembic can be kind of finicky so let me know if you would like me to do that.

src/pudl/transform/ferc1.py
```diff
@@ -4620,50 +4620,94 @@ def split_supporting_structure(self, df: pd.DataFrame) -> pd.DataFrame:
...
         # Categorize material (using normalized column)
-        df["supporting_structure_material"] = categorize_strings(
-            normalized_col, material_params
+        # categorize_strings will set uncategorized values to NA
+        categorized_material = categorize_strings(normalized_col, material_params)
+
+        # Log categorization statistics
+        total_rows = len(df)
+        structure_categorized = (categorized_structure != "na_category").sum()
+        material_categorized = (categorized_material != "na_category").sum()
```
zaneselvans: I think there should always be zero instances where either of these values equal na_category since at the end of categorize_strings all th na_category values are replaced with pd.NA.  To see what fraction of the reported values have been successfully categorized I think you'll probably want to count up the non-null values instead.

Materializing the asset locally I see that you're both getting non-100% values here, and that in the core table that's generated, there are no na_category values, which is confusing.

I see that the null values in the materialized table are also np.nan rather than pd.NA which is suspicious. Not sure exactly what's going on here. I have a hunch it might be due to the ENUM constraint.

src/pudl/transform/params/ferc1.py
```diff
@@ -772,78 +772,4850 @@
 Case is ignored, as all fuel strings are converted to lower case in the data set.
 """
 
-SUPPORTING_STRUCTURE_TYPE_CATEGORIES: dict[str, set[str]] = {
+SUPPORTING_STRUCTURE_TYPE_CATEGORIES: dict[str, dict[str, set[str]]] = {
```
zaneselvans: These parameter dictionaries should be embedded in the larger parameter data structure for the `core_ferc1__yearly_transmission_lines_sched422` table defined below.  Search for other examples that have a `categorize_strings` parameter as a guide.  It's fine to leave these monster dictionary definitions here and just reference them below (as is being done with e.g. `FUEL_CATEGORIES`) but to use the transformer class templates all the parameters need to be embedded in the parameter dictionaries, which are used to instantiate a big nested `TransformParams` class that guides the transformer behavior.

I'm going to hold off on commenting on the categories assigned to all of these strings for the moment and come back to that after all the code changes have been made.

src/pudl/transform/ferc1.py
zaneselvans: Sorry I was wrong about this! It's been a while since I worked on the FERC 1 transforms and realized when I was looking at the string categorization dictionary just now that it's not actually embedded in the larger transformer parameter definition. See my comments in the params file and the `transform_main()` method.

src/pudl/transform/ferc1.py
zaneselvans, line 4642: These transformer classes are set up to do a bunch of different steps in a predefined order, and to take their cues as to what transformations should be applied to which columns based on the parameters defined in `params/ferc1.py`

I think the only unusual thing we're doing here that needs to be special cased creating an additional two copies of the original `supporting_structure_type` column so that we have 3 versions in the output: the original, the one categorized by structure type, and one categorized by structure material.

Once those 3 original columns have been created, I think the actual string categorization can be handled using the existing default behavior, as directed by the parameters. This would mean creating appropriate `categorize_strings` entries in column-level parameters for each of the two columns to be categorized, (`supporting_structure_type` and `supporting_structure_material`) rather than manually applying the changes in the custom `split_supporting_structure()` method.

You can find other examples of these transform parameterizations in the `params/ferc1.py` module by searching for `categorize_strings`

Mostly I think this is just shuffling some of the existing code / parmeters around, which will probably result in a much shorter, simpler `split_supporting_structure` method, since all the actual categorization will be done by the already templated transforms.

There's a fair amount of implicit logic here so I can see how this would be confusing. You can see what the transform template looks like in `Ferc1AbstractTableTransformer`.  You're already inserting the `split_supporting_structure` at the right point in the process. The only thing that really needs to change is for the parameters to be updated so that the `categorize_strings` step automatically takes care of the categorization in a standardized way, and then paring down the `split_supporting_structure` method to do the splitting but not the categorizing.

zaneselvans (changes requested):
Hey sorry I think I may have misled you slightly about how the transforms are structured / templated.  I think it should be pretty easy to shuffle these parts around to follow the existing pattern though. See inline comments.

I'm holding off on digging into the 9000+ lines of categorization until we get all the other parts organized.

dd921:
@zaneselvans Thanks for the detailed feedback! I've tried to address your comments, but may need some assistance with the alembic version files, I was having a bit of trouble combining them like you mentioned.

src/pudl/transform/ferc1.py
```diff
@@ -4604,9 +4604,42 @@ class TransmissionLinesTableTransformer(Ferc1AbstractTableTransformer):
...
+        """
+        if "supporting_structure_type" not in df.columns:
+            return df
+
+        original_col = df["supporting_structure_type"].copy()
+
+        # Preserve original unstructured value for user interpretation
+        df["supporting_structure_type_original"] = original_col.copy()
+
+        df["supporting_structure_type"] = original_col.copy()
```
zaneselvans: Could just not assign this one, since it's the original source and we're not modifying the values.

src/pudl/transform/ferc1.py
```diff
@@ -4604,9 +4604,42 @@ class TransmissionLinesTableTransformer(Ferc1AbstractTableTransformer):
...
+            df: DataFrame with supporting_structure_type column.
+
+        Returns:
+            DataFrame with:
+            - supporting_structure_type_original: Original unstructured value
+            - supporting_structure_type: Categorized structure type (NA for ambiguous values)
+            - supporting_structure_material: Categorized material (NA for ambiguous values)
+        """
+        if "supporting_structure_type" not in df.columns:
+            return df
```
zaneselvans: I think we probably want the transform to fail if this column isn't in the table -- it'll fail at the end of the transform anyway when we try to enforce the schema. So we don't need the check + early return.

src/pudl/transform/ferc1.py
```diff
@@ -4604,9 +4604,39 @@ class TransmissionLinesTableTransformer(Ferc1AbstractTableTransformer):
...
+
+        return df
+
     def transform_main(self: Self, df: pd.DataFrame) -> pd.DataFrame:
-        """Do some string-to-numeric ninja moves."""
+        """Transform transmission lines table, splitting structure type and material."""
         df["num_transmission_circuits"] = pd.to_numeric(df["num_transmission_circuits"])
+        # Split supporting_structure_type before calling super().transform_main()
+        # which will apply categorize_strings to the new columns
+        df = self.split_supporting_structure(df)
```
krivard, line 4641: I think we probably don't need a separate method call here anymore -- just


```suggestion
        # Prepare to split supporting structure type in the categorize_strings phase
        df = df.assign(
            supporting_structure_type_original=df.supporting_structure_type.copy(),
            supporting_structure_material=df.supporting_structure_type.copy(),
        )
```

krivard:
Okay, alembic files should be cleaned up. To adjust your local schema, do:

```
$ git checkout 27dc5d3bd8e1
$ alembic downgrade 91b851ca4e39
$ git checkout ferc1-structure-type
$ git pull
$ alembic upgrade head
```

The headless checkout is needed so that you have the old migration files that know how to roll your initial set of changes back to something in `main`. The pull drops those files and replaces them with a single migration to go from `main` to support for the new columns. I verified the new migration works by successfully materializing `core_ferc1__yearly_transmission_lines_sched422`. Let me know if it gives you any trouble!

dd921:
@krivard Thank you for the review and alembic work! I addressed the non-blocking suggestion above.

Regarding the beastly dictionaries - I don't have the script I used for the initial categorization. I've taken a mix of approaches for organizing them, bouncing between scripts doing high level heuristics & pattern matching and then some Cursor-assisted refinement/reorganization of moving items into `na_category` that don't make logical sense for their assignment.

I'm trying to take a more reproducible approach moving forward. I see a few options: 
1. Start over and recreate the script that takes the initial unique values and re-analyze the distribution of categories, try to handle the punctuation and whitespace cleaning from the start with the raw data.

2. Apply your suggestion to clean punctuation and whitespace on the existing dicts and then give it another manual review on my end.
    -  I did apply this to the dictionaries in a notebook and saw ~13% reduction in unique values across both materials and structure type categorization. There were a few hundred re-categorization conflicts that emerged from this. When I looked deeper I did see a few illogical groupings that already exist in in the dictionaries, making me think I should take a bit longer to add some manual judgement to the dicts in their current state. These errors were especially prevalent in the structure type dict. I'm a bit confused if these multi-structure types belong as uncategorized or if they should just go into a category with at least one "match".
Ex:    
```
       'tower' :
            "h-frame twr",
            "h-frame, steel pole",
            "h-frame,steel pole",
```

Let me know what you think! Thanks again for all the help here.

krivard:
> I'm a bit confused if these multi-structure types belong as uncategorized or if they should just go into a category with at least one "match"

So long as the multi-structure records make up less than 10% of the total, we want them to map to `na_category`. In other words, we only want high-confidence and complete matches, with all ambiguous and partial matches relegated to NA.

> I did apply [clean punctuation and whitespace] to the dictionaries in a notebook and saw ~13% reduction in unique values

13% is not nothing, but it's not a big enough reduction to warrant starting over. Thanks for checking though!

> Regarding the beastly dictionaries

Thoughts on dumping out each dict as a YAML file? Then they won't be such monster scrollhounds in the source files. I'm checking with some colleagues about what a good way to load the files into the params structures is, but maybe you have ideas having touched the code more deeply than I have? I suppose we could technically read them in at import/definition time but that seems.... rude

dd921:
> So long as the multi-structure records make up less than 10% of the total, we want them to map to na_category. In other words, we only want high-confidence and complete matches, with all ambiguous and partial matches relegated to NA.

Let me dig into this a bit deeper - I think there are definitely some ambiguous ones currently mapped which I could manually shift to `na_category`. I agree with the sentiment of putting only high-confidence matches in the new derived columns.

> Thoughts on dumping out each dict as a YAML file? Then they won't be such monster scrollhounds in the source files. I'm checking with some colleagues about what a good way to load the files into the params structures is, but maybe you have ideas having touched the code more deeply than I have? I suppose we could technically read them in at import/definition time but that seems.... rude

I am glad you suggested this as I thought about leveraging YAML from the start of this issue, but it seemed to be an anti-pattern to what y'all have in place today and I didn't want to deviate. I can take a stab at this but would love to hear input from the others! I'm not sure I have a "graceful" solution for loading them in.

Also, is this something we'd want to align with the other dicts like `FUEL_CATEGORIES`? Seems out of scope for this work but could make this logic more readable at the cost of breaking the actual values into separate files.


My first thought was something like this maybe? definitely open to what you and your colleagues have in mind! 
```

def _load_category_dict(filename: str) -> dict[str, dict[str, set[str]]]:
    """Load categories from YAML

    Args:
        filename: Name of the YAML file in pudl.package_data.ferc1

    Returns:
        Dictionary with structure {"categories": {"category_name": set([...])}}
    """
    yaml_file = importlib.resources.files("pudl.package_data.ferc1") / filename
    with yaml_file.open() as f:
        data = yaml.safe_load(f)

    return {
        "categories": {
            cat: set(values) for cat, values in data["categories"].items()
        }
    }


SUPPORTING_STRUCTURE_TYPE_CATEGORIES: dict[str, dict[str, set[str]]] = _load_category_dict(
    "supporting_structure_type_categories.yml"
)
```

krivard:
hi hi sorry for the delay! this turned out to be a ["one chalk mark $1; knowing where to put it $10k" problem ](https://www.smithsonianmag.com/history/charles-proteus-steinmetz-the-wizard-of-schenectady-51912022/) and it took us a bit to work out where it should go :)

The machinery is about to merge into main from https://github.com/catalyst-cooperative/pudl/pull/4978 -- you can take a look there if curious but the short version is

* grab the dict that hangs off the `categories` key only
* serialize it to YAML as a dict[str, list[str]] and not as dict[str, set[str]] for better readability of the YAML, and sort the lists so it'll be easier to find & update stuff down the line. Something like

```
yaml.dump(
    {
        cat: sorted(val)
        for cat, val in categories.items()
    },
)
```
* set the `categories` key to a Path to your new file

& the TransformParams mechanisms will do the rest. That way, the file only gets read if we're actually running the transform step, and not if we're just importing something unrelated from the module.

> is this something we'd want to align with the other dicts like FUEL_CATEGORIES?

Totally, though we'll do those as separate PRs. I did a quick count and FUEL_CATEGORIES, FUEL_UNIT_CATEGORIES, PLANT_TYPE_CATEGORIES, and CONSTRUCTION_TYPE_CATEGORIES are the biggies, with 300-500 lines each. I'll drop those in a new Issue.

dd921:
@krivard

I was able to materialize `core_ferc1__yearly_transmission_lines_sched422` with the new changes. Please let me know if this is what you had in mind, and thanks again to all involved for making the updates to accommodate  federating to yaml.

I'm not sure what happened with the remaining checks getting hung up, but I'm also not sure if I can manually re-trigger them.

Also please let me know if the update I made in  [8b170a4](https://github.com/catalyst-cooperative/pudl/pull/4859/commits/8b170a49de09b0235bf3166e8eb99ece17b89b4e) to the alembic migration messed up any of your early revisions combining those files.

src/pudl/transform/params/ferc1.py
```diff
@@ -772,9696 +774,22 @@
...
-            "z.52",
-            "zpox",
-            "—",
-        },
-    }
+SUPPORTING_STRUCTURE_MATERIAL_CATEGORIES: dict[str, Path] = {
+    "categories": Path(
+        importlib.resources.files("pudl.package_data.ferc1")
+        / "supporting_structure_material_categories.yml"
+    )
```
krivard, line 792: Perfect thanks!

migrations/versions/f11b06e302fb_add_core_ferc1__yearly_transmission_.py
```diff
@@ -11,7 +11,7 @@
 
 # revision identifiers, used by Alembic.
 revision = 'f11b06e302fb'
-down_revision = 'dcbe4c9e66d4'
+down_revision = '86e3a75cf325'
```
krivard: nice and surgical ⭐

krivard:
Great work! I've approved the checks; they should complete momentarily.

Are you still working on looking over the mappings themselves, or is this ready for review & merge?

dd921:
@krivard My apologies for missing your last comment! I'm ready for review on this - I looked over the mappings again and feel good about where they are at re: ambiguity defaulting to `na_category` but I am happy to adjust anything as y'all see fit. 

Thanks!

src/pudl/package_data/ferc1/supporting_structure_material_categories.yml
```diff
@@ -0,0 +1,4831 @@
...
+  - (3) steel (a)
+  - (3) steel (c)
+  - (3) steel (d)
+  - (3)steel
+  - (c-h-f)(s-h-f)
+  - (c-h-f)(s-p)
+  - (c-p)(s-p)
+  - (c-p)(s-p)(s-t)
+  - (c-p)(s-p)(w-h-f)
+  - (c-p)(s-t)
```
krivard, line 18: looks like there's a ton of multi-material items here -- these on L13-18 are concrete-steel or concrete-steel-wood but there's a pile more below. I'm inclined to merge this to a local branch so that zane and I can use our Superior Domain Experience to shuffle out all the ones that are secretly multi-category... but before I do that, are there any of these you _did_ catch, but found analytics to justify keeping them in a single category?

src/pudl/package_data/ferc1/supporting_structure_material_categories.yml
```diff
@@ -0,0 +1,4831 @@
...
+  - (s-h-f)(s-m-p)(s-p)
+  - (s-h-f)(s-p)
+  - (s-h-f)(s-p)(s-t)
+  - (s-h-f)(s-t)
+  - (s-h-f)(w-h-f)
+  - (s-m-p)(s-p)
+  - (s-m-p)(s-t)
+  - (s-p) (s-a)
+  - (s-p) (s-t)
+  - (s-p) (w-p)
```
krivard: multi-material (steel, wood)

test/unit/transform/ferc1_test.py
```diff
@@ -1315,3 +1316,142 @@ def test_filter_for_freshest_data_xbrl(df):
...
+        "h-frame",
+        "pole",
+        "steel",
+        "2-29.55",  # Unintelligible value (number pattern) - should map to NA
+        pd.NA,
+    ]
+    test_data = pd.DataFrame(
+        {
+            "supporting_structure_type": original_values,
+            "num_transmission_circuits": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
```
krivard, line 1346: catalyst todo: when we drop `split_supporting_structure()` we'll want to set up this df to have three copies of `original_values` column

test/unit/transform/ferc1_test.py
```diff
@@ -1315,3 +1316,142 @@ def test_filter_for_freshest_data_xbrl(df):
...
+        }
+    )
+
+    result = (
+        transformer.split_supporting_structure(test_data)
+        .pipe(transformer.normalize_strings)
+        .pipe(transformer.categorize_strings)
```
krivard, line 1353: after todo: above
```suggestion
        transformer.normalize_strings(test_data)
        .pipe(transformer.categorize_strings)
```

test/unit/transform/ferc1_test.py
```diff
@@ -1315,3 +1316,142 @@ def test_filter_for_freshest_data_xbrl(df):
...
+    # Test data with uncategorized value
+    test_data = pd.DataFrame(
+        {
+            "supporting_structure_type": ["unknown_value", "steel tower"],
+            "num_transmission_circuits": [1, 2],
+        }
+    )
+
+    result = (
+        transformer.split_supporting_structure(test_data)
```
krivard, line 1442: catalyst todo: same here

src/pudl/transform/ferc1.py
krivard: catalyst todo: apply the above

Issue #4860, opened 2025-12-22
Consolidate Zenodo cache classes using `UPath` API


zaneselvans:
# Overview

* In #4856 we added support for caching PUDL's raw inputs using S3, so now the Datastore can pull from a local filesystem, the Zenodo API, GCS, or S3.
* But this could be much simpler: the `UPath` class from the `universal_pathlib` library backed by `fsspec` provides a uniform filesystem-like interface to local paths as well as GCS and S3.
* Using `UPath` as an abstraction layer would also facilitate reading data from all the other storage methods that `UPath` supports.
* We should consolidate the `GoogleCloudStorageCache`, `S3Cache` and `LocalFileCache` classes into a single class that uses `UPath` to provide access to all of them, or if necessary, create smaller, simpler custom classes that do anything specialized to the various storage backends, with more of the core functionality in the `AbstractCache` class.

## Success Criteria

* We have consolidated our cloud storage and local filesystem cache classes & logic to use a single abstraction, derived from `UPath`.

Issue #4675, opened 2025-10-22
Can't access new data via zenodo while materializing asset in dagster UI


cmgosnell:
### Describe the bug
When I'm generating assets in the dagster UI which require downloading new data from zenodo, it very frequently causes errors. Sometimes those errors are the informative zenodo timeout errors and sometimes its a ChildProcessingError.... which is uninformative.

### Bug Severity
How badly is this bug affecting you?
 - **Medium:** With some effort, I can work around the bug. The way to work around this is to run the `pudl_datastore` script before trying to materialize your assets. That's annoying but doable.  Mostly

### To Reproduce
 - remove a local pudl data file from `pudl_in` and re-run those raw assets. It fails 9 times out of 10 (ISH) for me.

### Expected behavior
I would expect the asset generation process to gracefully grab the data it needs if its not stored locally. At the very least I would expect to see the zenodo error instead of dagster's uninformative message.

zaneselvans:
With the new Zenodo cache on S3 as the default this might be better now. Next time we update a DOI we should try running the ETL without pre-downloading.

PR #4870, opened 2025-12-25
Improve and simplify Zenodo input data caching

### LINKED ISSUE #4494: Fix GitHub input caching and pull Zenodo DOIs out of Python code
# Overview

- Our GitHub Actions appear to be pulling the raw inputs for PUDL from GCS every time they run, rather than using the cached version on GitHub -- the cache key never seems to hit. This slows down the CI and also results in egress fees.
- We should debug this and get it back to where the cached inputs are working again.
- Our current method of generating the hash key is brittle and could be the issue. It greps for the Zenodo DOIs in our Python code, writes the results out to disk, and hashes that file.
- It would be very easy to instantiate the ZenodoDOI settings class from a YAML or JSON file stored in the repo -- this functionality is built into Pydantic.
- This would also be a better separation of our code and our inputs.
- Then we could just hash that JSON/YAML file to see if we need to download fresh data.
- We could also only trigger a refresh of the zenodo cache on GCS when the DOI file changes, rather than any time the `src/pudl/workspace/datastore.py` module is touched.
- Having the dataset:DOI mapping in a neat, machine readable file would also pave the way for easy automated PRs that attempt to update our well-behaved inputs (`eiaapi`, `nrelatb`, `eia930`, `epacems`, etc.)

## Success Criteria

- Our GitHub actions pull data from the cache on GitHub rather than GCS when possible.
- The DOIs that determine what inputs feed into PUDL are easily machine readable in a separate file.
- The `.github/workflows/pytest.yml` workflow has been simplified to use the new setup.

zaneselvans:
# Overview

Several improvements to how we manage our caches of raw input data from Zenodo.

- Replace the separate `LocalFIleCache`, `S3Cache`, and `GoogleCloudStorageCache` classes with a single `UPathCache` that relies on the `UPath` class to access data on a variety of storage backends.
- Pull Zenodo DOIs out of `datastore.py` and into their own settings file.
- Populate intermediate (typically S3) writable cache layers when populating the local cache layer. 
- Only run `zenodo-cache-sync` action when the new `zenodo_dois.yml` file is touched.
- In `pytest` GitHub Actions workflow, cache input data based on hash of `zenodo_dois.yml`.
- Use the warmest available GitHub Actions cache of `$PUDL_INPUT` if the cache key misses.

Closes #4494, #4860, #4675

# Testing

- Ran integration tests in CI
- Ran unit tests locally
- Built the docs
- Removed my local `eiaapi` archive and rematerialized the Dagster assets depending on it, to see if I was able to download fresh inputs automatically without interrupting the ETL.

## To-do list

- [x] Verify #4675 is fixed by removing a local archive and re-downloading in ETL.
- [x] Update release notes.
- [x] Update developer docs that talk about how to change a Zenodo archive DOI.
- [x] Do a self-review of the PR

### DIFF
--- a/src/pudl/workspace/resource_cache.py
+++ b/src/pudl/workspace/resource_cache.py
@@ -107,6 +107,10 @@ class UPathCache(AbstractCache):
             f"base_path={self._base_path}"
         )
 
+    def __repr__(self) -> str:
+        """Returns string representation of UPathCache."""
+        return f"UPathCache(storage_path={self._base_path})"
+
     def _setup_credentials(self) -> dict[str, Any]:
         """Set up backend-specific credentials and storage options.
 
@@ -311,25 +315,29 @@ class LayeredCache(AbstractCache):
         """
         for i, cache in enumerate(self._caches):
             if cache.contains(resource):
-                logger.debug(
-                    f"get:{resource} found in layer {i} ({cache.__class__.__name__})."
-                )
+                logger.debug(f"get:{resource} found in layer {i} ({cache}).")
                 content = cache.get(resource)
 
                 # Populate all closer cache layers with this content
                 for j in range(i):
                     closer_cache = self._caches[j]
                     if not closer_cache.is_read_only():
+                        if closer_cache.contains(resource):
+                            logger.debug(
+                                f"{resource} already exists in layer {j} "
+                                f"({closer_cache}). Skipping update."
+                            )
+                            continue
                         logger.debug(
                             f"Populating {resource} into closer layer {j} "
-                            f"({closer_cache.__class__.__name__}) from layer {i}"
+                            f"({closer_cache}) from layer {i}"
                         )
                         try:
                             closer_cache.add(resource, content)
                         except Exception as e:
                             logger.warning(
                                 f"Failed to populate {resource} into layer {j} "
-                                f"({closer_cache.__class__.__name__}): {e}"
+                                f"({closer_cache}): {e}"
                             )
 
                 return content
@@ -344,11 +352,15 @@ class LayeredCache(AbstractCache):
         for cache_layer in self._caches:
             if cache_layer.is_read_only():
                 continue
-            logger.debug(f"Adding {resource} to cache {cache_layer.__class__.__name__}")
+            if cache_layer.contains(resource):
+                logger.debug(
+                    f"{resource} already exists in layer {cache_layer}. "
+                    "Skipping update."
+                )
+                continue
+            logger.debug(f"Adding {resource} to cache {cache_layer}")
             cache_layer.add(resource, content)
-            logger.debug(
-                f"Added {resource} to cache layer {cache_layer.__class__.__name__})"
-            )
+            logger.debug(f"Added {resource} to cache layer {cache_layer})")
 
     def delete(self, resource: PudlResourceKey):
         """Removes resource from the cache if the cache is not in the read_only mode."""
@@ -364,9 +376,7 @@ class LayeredCache(AbstractCache):
         """Returns True if resource is present in the cache."""
         for i, cache in enumerate(self._caches):
             if cache.contains(resource):
-                logger.debug(
-                    f"contains: {resource} found in layer {i} ({cache.__class__.__name__})."
-                )
+                logger.debug(f"contains: {resource} found in layer {i} ({cache}).")
                 return True
         logger.debug(f"contains: {resource} not found in layered cache.")
         return False
@@ -376,8 +386,6 @@ class LayeredCache(AbstractCache):
         for cache_layer in self._caches:
             if cache_layer.is_read_only():
                 continue
-            logger.debug(
-                f"{resource} optimally cached in {cache_layer.__class__.__name__}"
-            )
+            logger.debug(f"{resource} optimally cached in {cache_layer}")
             return cache_layer.contains(resource)
         return False
--- a/test/unit/workspace/resource_cache_test.py
+++ b/test/unit/workspace/resource_cache_test.py
@@ -46,6 +46,7 @@ class TestLayeredCache:
         assert self.layered_cache.get(res) == b"firstLayer"
 
         # Test updating all layers
+        self.layered_cache.delete(res)
         self.layered_cache.add(res, b"newContents")
         assert self.layered_cache.get(res) == b"newContents"
         assert self.cache_1.get(res) == b"newContents"
@@ -101,6 +102,56 @@ class TestLayeredCache:
         r_new = PudlResourceKey("a", "b", "new")
         ro_cache.add(r_new, b"xyz")
         assert not ro_cache.contains(r_new)
+
+    def test_skip_update_if_closer_layer_has_resource(self, mocker):
+        """Test that get() doesn't overwrite closer layer if it already has the resource."""
+        # Setup mocks
+        layer1 = mocker.Mock(spec=UPathCache)
+        layer2 = mocker.Mock(spec=UPathCache)
+
+        # Configure layer1 (local):
+        # 1. contains() returns False (during search)
+        # 2. contains() returns True (during update attempt)
+        layer1.contains.side_effect = [False, True]
+        layer1.is_read_only.return_value = False
+
+        # Configure layer2 (remote):
+        # contains() returns True (found here)
+        layer2.contains.return_value = True
+        layer2.get.return_value = b"content"
+
+        # Create layered cache
+        lc = LayeredCache(layer1, layer2)
+        res = PudlResourceKey("a", "b", "x")
+
+        # Execute
+        lc.get(res)
+
+        # Verify
+        layer1.add.assert_not_called()
+
+    def test_add_skips_existing_layers(self, mocker):
+        """Test that add() does not write to layers that already contain the resource."""
+        layer1 = mocker.Mock(spec=UPathCache)
+        layer2 = mocker.Mock(spec=UPathCache)
+
+        layer1.is_read_only.return_value = False
+        layer2.is_read_only.return_value = False
+
+        # layer1 doesn't have it, layer2 has it
+        layer1.contains.return_value = False
+        layer2.contains.return_value = True
+
+        lc = LayeredCache(layer1, layer2)
+        res = PudlResourceKey("a", "b", "x")
+        content = b"content"
+
+        lc.add(res, content)
+
+        # layer1 should be written to
+        layer1.add.assert_called_once_with(res, content)
+        # layer2 should NOT be written to because it already has it
+        layer2.add.assert_not_called()
 
 
 class TestUPathCache:


.github/workflows/pytest.yml
```diff
@@ -233,17 +233,14 @@ jobs:
           which sqlite3
           sqlite3 --version
 
-      - name: Compile Zenodo datastore DOIs for cache invalidation
-        run:
-          grep -e '.*10\.\(5281\|5072\)/zenodo\..*' src/pudl/workspace/datastore.py
-          | sed -e 's/",*$//g' | sed -e 's/^.*"//g' | sort > datastore-dois.txt
-
```
zaneselvans, line 186: This ad-hoc compilation of the DOIs is no longer necessary since we now have a dedicated file we can hash that contains just mapping of data source IDs to DOIs.

.github/workflows/pytest.yml
```diff
@@ -233,17 +233,14 @@ jobs:
...
       - name: Restore Zenodo datastore from cache if possible
         uses: actions/cache@v5
         id: cache-zenodo-datastore
         with:
           path: ${{ env.PUDL_INPUT }}
-          key: zenodo-datastore-${{ hashFiles('datastore-dois.txt') }}
+          key: zenodo-datastore-${{ hashFiles('src/pudl/package_data/settings/zenodo_dois.yml') }}
+          restore-keys: |
+            zenodo-datastore-
```
zaneselvans, line 189: Using this more lenient fallback for the hash keys means that we'll almost always be able to use cached data for most of the datasets we rely on.  By default it will pull the most recently cached version. If it's not all the right data, it'll just need to download the datasets whose DOIs have changed.

.github/workflows/pytest.yml
```diff
@@ -233,17 +233,14 @@ jobs:
...
-          grep -e '.*10\.\(5281\|5072\)/zenodo\..*' src/pudl/workspace/datastore.py
-          | sed -e 's/",*$//g' | sed -e 's/^.*"//g' | sort > datastore-dois.txt
-
       - name: Restore Zenodo datastore from cache if possible
         uses: actions/cache@v5
         id: cache-zenodo-datastore
         with:
           path: ${{ env.PUDL_INPUT }}
-          key: zenodo-datastore-${{ hashFiles('datastore-dois.txt') }}
+          key: zenodo-datastore-${{ hashFiles('src/pudl/package_data/settings/zenodo_dois.yml') }}
```
zaneselvans, line 187: Hash the `zenodo_dois.yml` settings file directly to generate the cache key.

.github/workflows/zenodo-cache-sync.yml
```diff
@@ -7,7 +7,7 @@ on:
     - cron: "0 6 * * 1-5" # Weekdays at midnight on MST
   pull_request:
     paths:
-      - "src/pudl/workspace/datastore.py"
+      - "src/pudl/package_data/settings/zenodo_dois.yml"
```
zaneselvans, line 10: Now instead of this workflow running any time anything in the datastore module changes, it'll just refresh the cache when we actually change the DOIs (or add/remove datasets).

src/pudl/workspace/datastore.py
```diff
@@ -187,36 +189,69 @@ def get_json_string(self) -> str:
...
-    ferc1: ZenodoDoi = "10.5281/zenodo.17563678"
-    ferc2: ZenodoDoi = "10.5281/zenodo.17223540"
-    ferc6: ZenodoDoi = "10.5281/zenodo.17119798"
-    ferc60: ZenodoDoi = "10.5281/zenodo.17223513"
-    ferc714: ZenodoDoi = "10.5281/zenodo.16676145"
-    gridpathratoolkit: ZenodoDoi = "10.5281/zenodo.10892394"
-    nrelatb: ZenodoDoi = "10.5281/zenodo.12658647"
-    phmsagas: ZenodoDoi = "10.5281/zenodo.17076661"
-    sec10k: ZenodoDoi = "10.5281/zenodo.15161694"
-    vcerare: ZenodoDoi = "10.5281/zenodo.15166129"
```
zaneselvans, line 214: Rather than hard-coding the DOI defaults in the class definition, offload them into a settings file which is used by default to initialize the class.

src/pudl/workspace/datastore.py
```diff
@@ -187,36 +189,69 @@ def get_json_string(self) -> str:
...
+        # Load defaults from YAML file
+        default_path = (
+            importlib.resources.files("pudl.package_data.settings") / "zenodo_dois.yml"
+        )
+        with default_path.open() as f:
+            yaml_data = yaml.safe_load(f)
+
+        # Merge provided data with defaults (provided data takes precedence)
+        yaml_data.update(data)
+        super().__init__(**yaml_data)
```
zaneselvans, line 239: A DOI for every dataset is required to initialize the class. This allows the user to create a `ZenodoDoiSettings` that overrides a subset of the DOIs with the arguments that are passed in, while falling back to the defaults specified in the settings file for all the others.

src/pudl/workspace/datastore.py
```diff
@@ -315,21 +352,30 @@ class Datastore:
 
     def __init__(
         self,
-        local_cache_path: Path | None = None,
-        cloud_cache_path: str | None = "s3://pudl.catalyst.coop/zenodo",
+        local_cache_path: str | Path | UPath | None = None,
+        cloud_cache_path: str | UPath | None = "s3://pudl.catalyst.coop/zenodo",
```
zaneselvans, line 356: Allows a variety of types to be used to initialize the Datastore. But they are all converted to `UPath` with an explicit protocol before being handed off to the `UPathCache` class.

src/pudl/workspace/datastore.py
```diff
@@ -342,30 +388,43 @@ def __init__(
...
-                        "Unable to obtain credentials for GCS Cache at "
-                        f" {cloud_cache_path}. "
-                        f"Falling back to Zenodo if necessary. Error was: {e}"
-                    )
-            else:
+            # Convert to UPath
+            cloud_upath = UPath(cloud_cache_path)
+
+            # Get supported cloud protocols (everything except 'file')
+            supported_cloud_protocols = UPathCache.supported_protocols - {"file"}
```
zaneselvans, line 411: So it remains functional if we add support for other protocols.

.pre-commit-config.yaml
```diff
@@ -91,6 +91,16 @@ repos:
...
+      - id: sort-zenodo-dois
+        name: sort-zenodo-dois
+        stages: [pre-commit]
+        language: system
+        verbose: false
+        files: "^src/pudl/package_data/settings/zenodo_dois\\.yml$"
+        pass_filenames: true
+        always_run: false
+        entry: bash -c 'for file in "$@"; do LC_ALL=C sort "$file" -o "$file"; done' --
+
```
zaneselvans, line 103: Pre-commit hook to keep the `zenodo_dois.yml` file sorted, so we don't end up with spurious changes.  Have to specify the locale in the command to sort because different operating systems have different lexical sorting otherwise.

src/pudl/workspace/resource_cache.py
```diff
@@ -3,43 +3,16 @@
...
-from google.cloud.storage.retry import _should_retry
-from google.resumable_media.common import DataCorruption
+from upath import UPath
 
 import pudl.logging_helpers
 
 logger = pudl.logging_helpers.get_logger(__name__)
 
 
-def extend_gcp_retry_predicate(predicate, *exception_types):
```
zaneselvans, line 25: It would be nice if we could retain this more robust retry logic, but I didn't see an obvious way to force it into the `UPathCache` and we aren't depending on GCS by default any more, so I dropped it.

I retained the `AbstractCache` class since it defines the interface that's used by the `LayeredCache` as well as the `UPathCache` class.

src/pudl/workspace/resource_cache.py
zaneselvans, line 1: The changes in this file are easier to read if you look at the file as a whole -- in here the diffs are interleaved, but in practice what I did was:
- Remove the `LocalFileCache`, `S3Cache`, and `GoogleCloudStorageCache` classes.
- Add a new `UPathCache` class that can handle all 3 of those storage backends using the`UPath` class.
- The only backend-specific logic is in how credentials are obtained and managed, which is internal to the class, in the `_setup_credentials` method.

test/integration/resource_cache_test.py
```diff
@@ -0,0 +1,264 @@
...
+        assert not cache.contains(sample_resource)
+
+    @pytest.mark.xfail(
+        raises=(
+            MaxRetryError,
+            ConnectionError,
+            RetryError,
+            ResponseError,
+        )
+    )
```
zaneselvans: This pattern of xfail-on-network-flakiness was suggested by an LLM and I'm not sure how I feel about it.  It **is** annoying when spurious network hiccups cause the tests to fail, but also we won't notice if something goes wrong that we need to deal with.

test/unit/workspace/resource_cache_test.py
zaneselvans, line 1: With just the LayeredCache and UPathCache to test there's a lot of condensation in here.

.github/workflows/pytest.yml
zschira: That's nice.

src/pudl/workspace/datastore.py
```diff
@@ -315,21 +352,30 @@ class Datastore:
...
         timeout: float = 15.0,
     ):
         """Datastore manages input data retrieval for PUDL datasets.
 
+        Requires at least one of local_cache_path or cloud_cache_path to be provided.
+
         Args:
-            local_cache_path: if provided, :class:`LocalFileCache` pointed at the data
-                subdirectory of this path will be used with this Datastore.
+            local_cache_path: if provided, :class:`UPathCache` pointed at Zenodo
```
zschira, line 364: Have you tested this with both absolute and local paths? I think it would be fine to support only absolute paths, but it might be helpful to explicitly call that out if it's the case.

src/pudl/workspace/resource_cache.py
```diff
@@ -85,236 +58,211 @@ def contains(self, resource: PudlResourceKey) -> bool:
...
+            try:
+                credentials, project_id = google.auth.default()
+                logger.debug(f"Using GCP credentials with project_id={project_id}")
+                # fsspec's gcsfs will automatically use default credentials
+                # We can pass the project for requester-pays buckets
+                if project_id:
+                    storage_options["project"] = project_id
+            except Exception as e:
+                logger.warning(
+                    f"Could not load GCP credentials: {e}. Attempting anonymous access."
```
zschira, line 153: Would anonymous access would ever work for GCS? I suppose it's fine to try and you would still get errors later on, but maybe it would be simpler to just immediately raise an error.

src/pudl/workspace/resource_cache.py
```diff
@@ -85,236 +58,211 @@ def contains(self, resource: PudlResourceKey) -> bool:
...
-            return response["Body"].read()
-        except ClientError as e:
-            if e.response["Error"]["Code"] == "NoSuchKey":
-                raise KeyError(
-                    f"{resource} not found in S3 bucket {self._bucket_name}"
-                ) from e
-            raise
+            return path.read_bytes()
+        except FileNotFoundError as e:
+            raise KeyError(f"{resource} not found at {path}") from e
```
zschira, line 202: Is there a reason to catch this and explicitly raise an error? Is this just to provide a more useful error message than the default?

src/pudl/workspace/resource_cache.py
```diff
@@ -85,236 +58,211 @@ def contains(self, resource: PudlResourceKey) -> bool:
...
+        # Check if we can write (for S3/GCS without credentials)
+        if self._protocol == "s3" and self._storage_options.get("anon", False):
             raise RuntimeError(
                 "Cannot write to S3 without credentials. "
                 "Please configure AWS credentials to write to S3."
             )
 
-        key = self._get_object_key(resource)
-        logger.debug(f"Adding {resource} to S3 bucket {self._bucket_name}")
+        if self._protocol == "gs" and self._storage_options.get("token") == "anon":
```
zschira: It might be more maintainable to add a function like `def _check_not_anon(self) -> bool:` (or a better name) so we can confine the GCS/S3 specific logic for this check to a single place.

src/pudl/workspace/resource_cache.py
```diff
@@ -85,236 +58,211 @@ def contains(self, resource: PudlResourceKey) -> bool:
...
+                "Please configure GCP credentials to delete from GCS."
+            )
+
+        path = self._resource_path(resource)
+        logger.debug(f"Deleting {resource} from UPath cache at {path}")
 
-        self._s3_client.delete_object(Bucket=self._bucket_name, Key=key)
+        try:
+            path.unlink(missing_ok=True)
+        except Exception as e:
```
zschira: Same question as above, are you catching exceptions here just to provide better error messages?

src/pudl/workspace/resource_cache.py
```diff
@@ -85,236 +58,211 @@ def contains(self, resource: PudlResourceKey) -> bool:
...
-            if error_code == "404" or error_code == "NoSuchKey" or http_status == 404:
-                # Object doesn't exist - this is expected behavior
-                logger.debug(f"{resource} not found in S3 bucket {self._bucket_name}")
-                return False
-
-            # For other errors, log and re-raise
-            logger.error(f"Error checking if {resource} exists in S3: {e}")
-            raise
+            return path.exists()
+        except Exception as e:
```
zschira: Is there a reason to catch this and return `False`? It seems like this could cause confusion if the function is saying the path doesn't exist when actually there's some other underlying issue. I think at least bumping up the log level would be nice so this isn't totally hidden if you're running with log level of `INFO`.

test/integration/resource_cache_test.py
zschira: I think I would be ok doing something like this for zenodo specific tests, but I think GCS/S3 are reliable enough that it would be worth seeing all failures.

test/integration/resource_cache_test.py
```diff
@@ -0,0 +1,229 @@
...
+from urllib3.exceptions import MaxRetryError, ResponseError
+
+from pudl.workspace.datastore import Datastore, ZenodoDoiSettings
+from pudl.workspace.resource_cache import LayeredCache, PudlResourceKey, UPathCache
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture
+def gcs_test_cache_path():
```
zschira, line 21: This approach seems vulnerable to race conditions if the integration tests are being run by multiple users simultaneously. Maybe we could create a sub-directory of this cache path using a UUID and then clean up that path after tests are complete?

test/unit/workspace/resource_cache_test.py
zschira: Nice!

src/pudl/workspace/datastore.py
zaneselvans: Slapping `file://` onto the front of a relative path does seem to work fine, though the stored path can sometimes look a little funny, e.g. 

* `UPath("../pudl-input")` yields: `PosixUPath('../pudl-input')`
* If you prepend `file://` on the same path: like `UPath("file://../pudl-input")` you'll get: `FilePath('/Users/zane/code/catalyst/pudl/../pudl-input', protocol='file')` which is correct if not canonical.
* There's `UPath.resolve()` which will take a local path and make it absolute and canonical, so I'll do that before adding the `file://` protocol.
* So the input path can be relative but the value it works with internally will always be absolute and canonical and use the file protocol.

src/pudl/workspace/resource_cache.py
zaneselvans: I think it *can* work - we just don't happen to have any totally free and open GCS buckets that can be accessed without any credentials right now.

src/pudl/workspace/resource_cache.py
zaneselvans: Only reason was to provide the error message.

src/pudl/workspace/resource_cache.py
zaneselvans: Yeah, just an error message.

src/pudl/workspace/resource_cache.py
zaneselvans: I feel like there was a specific case that I ran into where `.exists()` didn't do the right thing, but I'm not remembering what it was now, and this is clearly too heavy handed, so I'll simplify and if it's a real problem we'll find out soon enough.

test/integration/resource_cache_test.py
zaneselvans: Ah yeah that's true. Adding a UUID is easy.

test/integration/resource_cache_test.py
zaneselvans: Removed the XFAIL.

src/pudl/workspace/resource_cache.py
```diff
@@ -85,236 +58,199 @@ def contains(self, resource: PudlResourceKey) -> bool:
...
-        return (self._path_prefix / resource.get_local_path()).as_posix().lstrip("/")
+        return self._base_path / resource.get_local_path()
+
+    def _is_anonymous(self) -> bool:
+        """Returns True if the cache is using anonymous access (no credentials)."""
+        if self._protocol == "s3":
+            return self._storage_options.get("anon", False)
+        if self._protocol == "gs":
+            return self._storage_options.get("token") == "anon"
+        return False
```
zaneselvans: Consolidated protocol specific checks here.

src/pudl/workspace/resource_cache.py
zaneselvans: And this logic was duplicated in 2 places too. Consolidated into the `_is_anonymous()` method.

zaneselvans:
Okay @zschira I think I addressed the issues you highlighted.

zschira (approved):
Looks good!

PR #4871, opened 2025-12-27
Migrate to using `pixi` for environment/dependency management

### LINKED ISSUE #4604: Modernize Catalyst python packaging and environment management
# Overview

* We currently rely on `conda-lock` to specify a reproducible python environment for PUDL and `mamba` to do package management.
* We also use a `Makefile` to specify shared tasks that we want to be able to run in a uniform way across the team.
* We've been treating `pudl` like an "application" rather than a "library" for a while now, but we are still publishing the `catalystcoop.pudl` package to PyPI and `conda-forge`.
* Since we set up the current dependency management and packaging system in 2023, Python packaging has evolved significantly.  The author of `conda-lock` now recommends that people use `pixi` instead if they need to manage projects within the `conda-forge` ecosystem, and in fact uses `pixi` themselves to manage `conda-lock`.  They've also provided some migration tools.

## Advantages of [`pixi`](https://pixi.sh)
* Can manage dependencies from both PyPI and `conda-forge` while still providing reproducible multi-platform lockfiles.
* Can be used to package a project for installation via `conda-forge` -- this would allow `catalystcoop.pudl` to remain installable as an "application" with a locked environment.
* Faster dependency solving.
* Ability to update single dependency versions rather than needing to re-lock everything every time.
* Can also manage tasks in a way that's similar to a `Makefile`
* Allows the definition of multiple "features" which can be composed into different python environments, which would allow us to separate our core dependencies from those that we need for development, testing, data visualization, etc.
* Better human-readable reporting of what changed in the GitHub bot PRs.
* Better information for debugging dependency issues when they come up.
* Cleaner management of dependencies between the main PUDL repo and the `pudl-archivers` repo (which already uses pixi)

## Related Changes
- Stop publishing `catalystcoop.pudl` to PyPI since it's probably not really installable from there.
- Migrate all tasks that we want to retain from the `Makefile` into `pixi` tasks.
- Publish `catalystcoop.pudl` as an application with locked dependencies to conda-forge. This will actually be installable, reflect the real dependencies / Python environment, and the way we expect the package to be used. Alternatively we could just stop publishing the package outside of the git repo / Zenodo tarball archives entirely.
- Standardize our packaging / dependency management across repos to either use `uv` (for those that don't need `conda` specific functionality) or `pixi` (for those that do) and get rid of all the complicated `tox` setups.
- Installation with `pixi` will require overriding some build commands on RTD. We might also consolidate

### `uv` compatible repos
- ✅ `ferc-xbrl-extractor`
- ✅ `eel-hole`
- ✅ `pudl-usage-metrics-dashboard`
- ✅ `open-energy-data-for-all`
- ❌ `cheshire` (template repo)
- ❌ `dbfread`

### `pixi` repos
- 🟨 `pudl-archivers` (multi-environment, depends on `pudl`; could switch entirely to pixi + hatch)
- 🟨 `pudl` (geospatial, system dependencies)
- ❌ `pudl-examples` (geospatial)

zaneselvans:
# Overview

## Tool Replacements

- Switched from using `conda-lock` to `pixi` to create dependency lockfiles. Removed lockfiles under `environments` and added `pixi.lock`.
- Replaced `setuptools`, `build` and `setuptools_scm` to `hatch`, `hatchling` and `hatch-vcs`.
- Replaced (most of) our `make` targets with `pixi` tasks and removed the Makefile.
- Updated `pytest`, `build-deploy-pudl`, and `build-deploy-docs` workflows to use pixi.
- Switched from using micromamba docker image to pixi docker image.
- Tiny updates to the `gcp_pudl_etl.sh` script required by the new docker image (change in default user)
- Updated developer documentation to reflect the new packaging/environment/task setup
- Switched to a cleaner development versioning scheme using `hatch` -- now there's no hash hanging off the back.
- Set up a small custom RTD build configuration to work with the new packaging setup

## Application, not a Package

- Removed elements of `pyproject.toml` that were only relevant to publishing PUDL on PyPI / `conda-forge`.
- Updated the data release issue template and documentation to remove the PyPI / `conda-forge` steps.
- Removed the PyPI sections from our `release.yml` workflow file.
- Retained the GitHub release, to make versioned outputs obvious on GitHub & to serve as an archive separate from Zenodo.

## MIscellaneous Stuff

- Fixed formatting errors in `dbt` `schema.yml` files for:
  - `core_eia861__yearly_distributed_generation_fuel`
  - `core_eia861__yearly_energy_efficiency`
- Moved `furo` (our Sphinx theme) into the PyPI dependencies section, since the package on `conda-forge` is very stale. This resulted in upgrading to Sphinx 9+ which resulted in finding a Sphinx or `sphinx-autoapi` bug that I needed to work around, and in chasing that down I ended up replacing a lot of bad single-backtick quotes in our docstrings with double-backtick quotes.
- Changed the rules for running `pudl-deploy-docs` to it will redeploy whenever a PR merges into `main`, so we can see that it continues to work and so the `docs.catalyst.coop/pudl` docs are actually up to date.

Closes #4604

# Reviewer Questions

## Versioning Scheme

- Hatch makes versioning schemes easier to define.
- I switched our development versions to be like `v2025.12.1.dev56` (add the number of commits since the last tagged version to the last tagged version). These versions may be duplicative on different development branches, but once they land on main, the `dev23` suffix will indicate how many PRs have been merged into `main` since the last release.
- Given that we aren't distributing python packages any more, we could also go to using a leading zero on the month which would make the versions sort chronologically.  

## Integration with `pudl-archiver` repo

- We don't intend for anybody to be using `catalystcoop.pudl` as a dependency, since it needs to have locked dependencies for a reproducible environment, and it's really an application and not a library. That's why we're not distributing it on PyPI or `conda-forge` any more.
- However **we** install it as a dependency in the `pudl-archiver` repo directly from git, which has always been a little brittle.
- The way the `pyproject.toml` and pixi are set up on this branch do not allow the package dependencies to be installed with pip, meaning that the install from a git URL will fail.
- This is because all of the dependencies are specified in `tool.pixi.dependencies` rather in the top-level `dependencies` table.
- I tried to find a way to make the dependencies visible both to `pip` and to `pixi` without duplicating them all, and without ending up with a bunch of dependencies getting installed from PyPI, and failed.
- If we use the top-level project `dependencies` table, then we get pip installed packages, some of which will have binary incompatibilities.  Also, if `catalystcoop.pudl` is installed with pip (as with the git URL) it won't really be functional, since we do have binary dependencies. The versions of dependencies installed are also almost guaranteed to be different from the locked environment, since they'll be re-solved at the time of install.
- Also, also installing from git means there's no way for the installer to know whether the package needs to be updated. There's no version. This means that if you have an existing (e.g. local) install, you need to manually remove and re-install the PUDL package to get a refreshed version, which is easy to forget, and also takes a long time, since it seems to clone the whole repo.
- If you have any suggestions for how to get everything installed from `conda-forge` when working locally, avoid duplicate specification of the packages, and still make the package installable remotely that would be great.
- Alternatively, we could consider some more architectural solutions, which I think involve either splitting or merging repositories:
  - We could finally break the data source metadata out into a separate package that both PUDL and the archiver can depend on. If we wanted to we could also break the datastore module out into that package, as we've long discussed, disentangling it and the zenodo resource caching infrastructure from PUDL.  Given the large number of data sources that now only have metadata defined in `pudl-archiver` this might be a good idea, since having all of the data source metadata in one place, along with the tool that lets you access the archives, would mean all of those data sources would be accessible through the datastore, even if they're not all being actively used in PUDL.
  - In the other direction, we could merge the `pudl-archiver` repository into the PUDL application repository -- both of them use pixi, neither of them is a package that's getting distributed, and there are a variety of different ways we could manage them in the same repo, so the archiver would always have the most recent version of pudl metadata handy.  Pixi has good support for separated and even conflicting dependency environments. They could be two separate packages under `src/` or we could just add a `src/pudl/archiver` subpackage, or use pixi to define a separate environment/feature for the archiver.
  - Also, these two strategies are not mutually exclusive -- we could both split out the metadata / datastore as its own package, and also merge `pudl` and `pudl-archiver` repos. And architecturally, I think that kinda makes the most sense as a long term solution.

# Testing

- Ran all the CI on GitHub multiple times.
- Ran a successful branch deployment with Google Batch.
- Did a test of the new `update-lockfiles` workflow and merged the changes into this branch.
- Built and reviewed the docs locally.
- Verified that the RTD builds are working w/ pixi install.

# To-do list

- [x] Update developer docs explaining how to create and manage python environment.
- [x] Run `pytest-integration` in CI and make sure it passes.
- [x] Run a branch deployment to make sure I didn't break anything
- [x] Add a CI workflow that attempts to build the PUDL conda package with `pixi build`
- [x] Add the built PUDL conda package as an attachment to our GitHub releases.
- [x] Revise release notes in light of new conda packaging setup.
- [x] Run another branch deployment to make sure that packaging updates haven't broken anything.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -494,7 +494,7 @@ doctest_optionflags = ["NORMALIZE_WHITESPACE", "IGNORE_EXCEPTION_DETAIL", "ELLIP
 
 [tool.coverage.run]
 source = ["src/pudl/", "test/integration/", "test/unit/"]
-omit = ["src/pudl/*/ferceqr.py"]
+omit = ["src/pudl/*/ferceqr.py", "src/pudl/_version.py"]
 relative_files = true
 
 [tool.coverage.report]
--- a/pixi.lock
+++ b/pixi.lock
@@ -3875,8 +3875,8 @@ packages:
   timestamp: 1741554795915
 - pypi: ./
   name: catalystcoop-pudl
-  version: 2025.12.2.dev89
-  sha256: 68c094e6ecdd1764f6b8ae7f1ccbc8947e919a3b00dbe9c18523bfa58270b62c
+  version: 2025.12.2.dev90
+  sha256: cf31ce434c54f022e5a3a71e42b04dffde1d9caa9746cac8d7596cf44d4f67be
   requires_python: '>=3.13,<3.14.0a0'
 - conda: https://conda.anaconda.org/conda-forge/noarch/catalystcoop.dbfread-3.0.0-pyhd8ed1ab_1.conda
   sha256: 7a706fefa47027ab14cac7572dbddadaf9a12fc13a1222ad1e75f43f8fd9dcfb


.github/workflows/build-deploy-docs.yml
```diff
@@ -1,10 +1,9 @@
 name: build-deploy-docs
 
 on:
-  # push:
-  #   branches:
-  #     - '!gh-pages'
-  # pull_request:
+  push:
+    branches:
+      - main
```
zaneselvans, line 6: I set this up to deploy on main so it keeps getting exercised and we can check-in on the outputs.

.github/workflows/build-deploy-pudl.yml
```diff
@@ -120,12 +120,8 @@ jobs:
         run: |-
           ./devtools/generate_batch_config.py \
             --container-image "docker.io/catalystcoop/pudl-etl@${{ steps.docker-build.outputs.digest }}" \
-            --container-command "micromamba" \
+            --container-command "pixi" \
```
zaneselvans, line 123: Remarkably little needed to change in here to update the nightly build setup, which was nice.

.github/workflows/com-dev-notify.yml
```diff
@@ -19,8 +19,8 @@ jobs:
       - name: Get username if a discussion was created
         if: ${{ (github.event_name == 'discussion') }}
         run: |
-          echo "username=${{ github.event.discussion.user.login }}" >> $GITHUB_ENV
-          echo "url=${{ github.event.discussion.html_url }}" >> $GITHUB_ENV
+          echo "username=${{ github.event.discussion.user.login }}" >> "${GITHUB_ENV}"
+          echo "url=${{ github.event.discussion.html_url }}" >> "${GITHUB_ENV}"
```
zaneselvans, line 23: Addressed shellcheck linting complaints.

.github/workflows/pytest.yml
```diff
@@ -89,35 +88,19 @@ jobs:
...
-          cache-environment: true
-
-      - name: Log environment details
-        run: |
-          conda info
-          conda list
-          conda config --show-sources
-          conda config --show
-          printenv | sort
+          locked: true
```
zaneselvans, line 94: We install the locked environment everywhere in the workflows, otherwise they will automatically update the environment.

.github/workflows/release.yml
```diff
@@ -2,109 +2,48 @@
...
-          python -m build
-          twine check --strict dist/*
-      - name: Store the distribution packages
-        uses: actions/upload-artifact@v6
-        with:
-          name: python-package-distributions
-          path: dist/
+          gh release create '${{ github.ref_name }}' \
+            --title 'PUDL ${{ github.ref_name }}' \
+            --notes "PUDL is an application that must be installed from the repository using pixi.
```
zaneselvans, line 56: This is a placeholder. Ideally we would extract the relevant release notes from `release_notes.rst` and use them here automatically, plus some basic pointers to install & data access.

.github/workflows/release.yml
zaneselvans, line 1: Mostly removing the release mechanics in here -- just retaining the GitHub release for visibility of the versioned releases and a secondary archive of the source code.

.github/workflows/update-lockfiles.yml
```diff
@@ -30,17 +30,18 @@ jobs:
...
       - name: Update pixi lockfile, pre-commit hooks, and dbt deps
         run: |
           set -o pipefail
           pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
-          pixi install
+          pixi install --locked
           pixi run pre-commit-autoupdate
-          pixi run dbt-deps-upgrade
-      - name: Make a PR to merge updated pixi lockfiles
+          PUDL_OUTPUT="/home/runner/pudl-work/output/" pixi run dbt-deps-upgrade
```
zaneselvans, line 43: Need to give it a PUDL_OUTPUT because dbt refers to it somewhere and this will fail otherwise.

.github/workflows/update-lockfiles.yml
```diff
@@ -30,17 +30,18 @@ jobs:
...
         with:
           token: ${{ steps.generate-token.outputs.token }}
+          fetch-depth: 0 # Fetch all history and tags for hatch-vcs version detection
       - uses: prefix-dev/setup-pixi@v0.9.3
         with:
           run-install: false
       - name: Update pixi lockfile, pre-commit hooks, and dbt deps
         run: |
           set -o pipefail
           pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
```
zaneselvans, line 40: There are some very minor diffs that seem to be coming up between the workflow lockfile and local lockfiles.  Not significant versioning issues, and just ~25 out of 10s of thousands of rows, but it's annoying.

.github/workflows/zenodo-data-release.yml
```diff
@@ -55,23 +55,19 @@ jobs:
...
+          locked: true
+          cache: true
+          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
 
       - name: Do a Zenodo data release
         env:
           ZENODO_SANDBOX_TOKEN_PUBLISH: ${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }}
           ZENODO_TOKEN_UPLOAD: ${{ secrets.ZENODO_TOKEN_UPLOAD }}
         run: |
-          zenodo_data_release \
```
zaneselvans, line 74: Miraculously this actually succeeded when I tested it. Maybe Zenodo is okay again?

dbt/models/eia861/core_eia861__yearly_distributed_generation_fuel/schema.yml
```diff
@@ -15,8 +15,9 @@ sources:
           - name: fuel_pct
             data_tests:
               - dbt_expectations.expect_column_values_to_be_between:
-                  min_value: 0
-                  max_value: 100
+                  arguments:
+                    min_value: 0
+                    max_value: 100
```
zaneselvans, line 20: Missing explicit arguments was creating some warnings.

dbt/models/eia861/core_eia861__yearly_energy_efficiency/schema.yml
```diff
@@ -11,10 +11,10 @@ sources:
                 partition_expr: "EXTRACT(YEAR FROM report_date)"
           - dbt_expectations.expect_column_pair_values_A_to_be_greater_than_B:
               arguments:
-              column_A: incremental_life_cycle_energy_savings_mwh
```
zaneselvans, line 14: Lack of indentation that was creating some warnings.

docker/Dockerfile
```diff
@@ -1,6 +1,8 @@
-FROM mambaorg/micromamba:2.4.0-ubuntu24.04
+FROM ghcr.io/prefix-dev/pixi:0.62.2-noble
```
zaneselvans, line 1: This is still ubuntu24.04 under the hood. They're just using the code name instead of the version number.

docker/Dockerfile
```diff
@@ -48,22 +48,13 @@ RUN mkdir -p ${PUDL_INPUT} ${PUDL_OUTPUT} ${DAGSTER_HOME} ${PUDL_REPO}
...
-# can't find libraries contained within the environment. It's unclear why!
-ENV LD_LIBRARY_PATH=${CONDA_PREFIX}/lib
-RUN ${CONDA_RUN} pip install --no-cache-dir --no-deps --editable ${PUDL_REPO}
+# Install the pixi environment (includes the editable local package)
+RUN pixi install --locked
 
-WORKDIR ${PUDL_REPO}
 # Run the unit tests:
-CMD ["micromamba", "run", "--prefix", "${CONDA_PREFIX}", "--attach", "''", "pytest", "test/unit"]
+CMD ["pixi", "run", "-v", "pytest-unit"]
```
zaneselvans, line 61: I was able to run the container and it ran the tests.

docker/gcp_pudl_etl.sh
```diff
@@ -240,7 +240,7 @@ fi
 
 # Set these variables *only* if they are not already set by the container or workflow:
 : "${PUDL_GCS_OUTPUT:=gs://builds.catalyst.coop/$BUILD_ID}"
-: "${PUDL_SETTINGS_YML:=/home/mambauser/pudl/src/pudl/package_data/settings/etl_full.yml}"
+: "${PUDL_SETTINGS_YML:=/home/ubuntu/pudl/src/pudl/package_data/settings/etl_full.yml}"
```
zaneselvans, line 243: The whole script is run wrapped in a `pixi run bash` so there's very little to change.

src/pudl/__init__.py
```diff
@@ -27,7 +25,12 @@
 __maintainer__ = "Catalyst Cooperative"
 __license__ = "MIT License"
 __maintainer_email__ = "zane.selvans@catalyst.coop"
-__version__ = importlib.metadata.version("catalystcoop.pudl")
+
+try:
+    from pudl._version import __version__
+except ImportError:
+    __version__ = "0.0.0.dev0"
+
```
zaneselvans, line 33: Switched to using `hatch` / `hatchling` to package PUDL. It's pulling a version from the git based on our tags, but it stashes that version in this `_version.py` file for import. The file is generated every time the package is installed.

.pre-commit-config.yaml
```diff
@@ -119,16 +119,7 @@ repos:
...
-        entry: pytest --cov-fail-under=0 --doctest-modules src/pudl test/unit -m "not slow"
-
-      - id: conda-lock
-        name: conda-lock
-        stages: [pre-commit]
-        language: unsupported
-        verbose: false
-        files: "pyproject.toml"
-        always_run: false
-        entry: make conda-clean conda-lock.yml
```
zaneselvans, line 131: No longer required. The `pixi install` that happens as part of the pre-commit hooks *should* fail if `pixi.lock` is not consistent with `pyproject.toml` but we'll see how that finnicky it is in practice.

.readthedocs.yaml
```diff
@@ -6,24 +6,22 @@
...
+    install:
+      - pixi install
+    build:
+      html:
+        - >-
+          mkdir -p $HOME/pudl-work/input $HOME/pudl-work/output $READTHEDOCS_OUTPUT/html/ &&
+          export PUDL_INPUT=$HOME/pudl-work/input &&
+          export PUDL_OUTPUT=$HOME/pudl-work/output &&
+          pixi run docs-build &&
+          cp -r docs/_build/html/* $READTHEDOCS_OUTPUT/html/
```
zaneselvans, line 27: This was a bit finnicky to get set up, but it works now

AGENTS.md
zaneselvans, line 1: I had the LLM update its own instructions after working through the pixi migration and it seemed to work, which is pretty weird.

pyproject.toml
```diff
@@ -146,46 +234,6 @@ pudl_null_cols = "pudl.scripts.pudl_null_cols:main"
...
-git_describe_command = [
-    "git",
-    "describe",
-    "--dirty",
-    "--tags",
-    "--long",
-    "--match",
-    "v20*",
-]
-
```
zaneselvans, line 188: Replaced setuptools with hatch.

pyproject.toml
```diff
@@ -338,30 +386,6 @@ log_cli = true
...
-google-cloud-sdk = ">=535"
-nodejs = ">=24"
-pandoc = ">=3.7"
-parallel = ">=20250822"
-pip = ">=25"
-prettier = ">=3.6,<4"
-python = ">=3.13,<3.14.0a0"
-sqlite = ">=3.51"
-zip = ">=3.0"
-
```
zaneselvans, line 364: No longer necessary.  All of these non-python dependencies have been added to the pixi dependencies above.

pyproject.toml
```diff
@@ -10,129 +25,202 @@ authors = [{ name = "Catalyst Cooperative", email = "pudl@catalyst.coop" }]
...
+
+[tool.pixi.activation.env]
+# Reusable variables for pixi tasks
+ETL_FAST_YML = "src/pudl/package_data/settings/etl_fast.yml"
+ETL_FULL_YML = "src/pudl/package_data/settings/etl_full.yml"
+PYTEST_ARGS = "--durations 20 --cov-fail-under=0"
+COVARGS = "--append"
+PIXI_LOCKED = "true"
+
+[tool.pixi.tasks]
```
zaneselvans, line 262: Pixi "tasks" take the place of our Make targets.

pyproject.toml
```diff
@@ -10,129 +25,202 @@ authors = [{ name = "Catalyst Cooperative", email = "pudl@catalyst.coop" }]
...
+# All dependencies are explicitly managed via conda-forge in [tool.pixi.dependencies] above.
+"catalystcoop.pudl" = { path = ".", editable = true, extras = [] }
+
+[tool.pixi.activation.env]
+# Reusable variables for pixi tasks
+ETL_FAST_YML = "src/pudl/package_data/settings/etl_fast.yml"
+ETL_FULL_YML = "src/pudl/package_data/settings/etl_full.yml"
+PYTEST_ARGS = "--durations 20 --cov-fail-under=0"
+COVARGS = "--append"
+PIXI_LOCKED = "true"
```
zaneselvans: Setting `PIXI_LOCKED` means that by default if you run `pixi install` it will just install the dependencies found in `pixi.lock` rather than re-solving the environment and updating the dependencies.

Resolving and updating is very fast, so it's not a hassle, but this is intended to try and limit dependency churn, and try to keep the big updates limited to the weekly `update-lockfiles` workflow.

If you explicitly run `pixi update` or `pixi add` it will still update the lockfile.

If there's an inconsistency between `pyproject.toml` and `pixi.lock` then `pixi install` will fail (and it runs every time you do any `pixi run` command)

pyproject.toml
```diff
@@ -10,129 +25,202 @@ authors = [{ name = "Catalyst Cooperative", email = "pudl@catalyst.coop" }]
...
-    "ferc form 6",
-    "ferc form 60",
-    "ferc form 714",
-    "finance",
-    "fossil fuel",
-    "natural gas",
-    "oil",
-    "policy",
-    "utility",
-]
```
zaneselvans, line 135: `classifiers` and `keywords` were meant for consumption by pypi / conda-forge.

pyproject.toml
```diff
@@ -146,46 +234,6 @@ pudl_null_cols = "pudl.scripts.pudl_null_cols:main"
 resource_description = "pudl.scripts.resource_description:main"
 zenodo_data_release = "pudl.scripts.zenodo_data_release:pudl_zenodo_data_release"
 
-[project.urls]
```
zaneselvans, line 149: Only needed for PyPI / conda-forge.

pyproject.toml
```diff
@@ -10,129 +25,202 @@ authors = [{ name = "Catalyst Cooperative", email = "pudl@catalyst.coop" }]
...
+# All optional dependencies are now managed via pixi features.
+# See [tool.pixi.feature.*.dependencies] below.
+dependencies = []
+
+[tool.pixi.workspace]
+name = "catalystcoop.pudl"
+platforms = ["linux-64", "osx-64", "osx-arm64"]
+channels = ["conda-forge"]
+
+[tool.pixi.dependencies]
```
zaneselvans, line 60: For now I've consolidated all of the dependencies into a single large environment. It solves fast and keeps things simple.  We can break it up into smaller "features" with optional dependencies / environments later if we want to.

pyproject.toml
```diff
@@ -10,129 +25,202 @@ authors = [{ name = "Catalyst Cooperative", email = "pudl@catalyst.coop" }]
...
+typos = ">=1.40"
+universal-pathlib = ">=0.3.7"
+urllib3 = ">=2.4"
+xlsxwriter = ">=3.2"
+zip = ">=3.0"
+
+[tool.pixi.pypi-dependencies]
+# Install only the local package in editable mode with no dependencies.
+# All dependencies are explicitly managed via conda-forge in [tool.pixi.dependencies] above.
+"catalystcoop.pudl" = { path = ".", editable = true, extras = [] }
```
zaneselvans: PUDL isn't actually installed from PyPI -- this is how the local editable package gets installed into the pixi environment.

If we run into dependencies that can only be installed from PyPI we can add them here, and they'll get locked just like conda dependencies. This will work fine for pure python packages (like `furo`, below) but probably won't be robust for packages with compiled extensions like `grpcio` (above).

.github/ISSUE_TEMPLATE/versioned_release.md
```diff
@@ -23,9 +23,8 @@ assignees: ""
 - [ ] Tag the release `git tag -as -m "PUDL vYYYY.M.x" vYYYY.M.x`
 - [ ] Push the release tag to `main`: `git push origin vYYYY.M.x`
 - [ ] Verify that the new [GitHub (software) release](https://github.com/catalyst-cooperative/pudl/releases) has been published
-- [ ] Verify [`catalystcoop.pudl` PyPI (software) release](https://pypi.org/project/catalystcoop.pudl/)
```
zaneselvans, line 26: Remove the tasks related to software release publication.

docs/dev/nightly_data_builds.rst
```diff
@@ -121,64 +121,56 @@ permissions.
...
 ------------------------------------------------
 You can download the outputs from a successful nightly build data directly from the
-``s3://pudl.catalyst.coop`` bucket. To do this, you'll
-need to `follow the instructions
-<https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html>`__
-for installing the AWS CLI tool.
-
-To test that the AWS CLI tool and the bucket are working properly, run:
+``s3://pudl.catalyst.coop`` bucket using the ``gcloud storage`` CLI, which can
+access both GCS and S3 storage buckets.
```
zaneselvans, line 125: Is there a reason we were having people use the AWS CLI rather than `gcloud` (which can handle both GCS and S3 and which is part of our environment and doesn't need to be installed separately)?  It seemed like maybe a holdover so I updated it all to use the google tooling.

src/pudl/analysis/record_linkage/name_cleaner.py
zaneselvans, line 1: Only docstring changes in here. Was chasing down a bug in the docs build.

src/pudl/extract/dbf.py
zaneselvans, line 1: Only docstring changes in here. Was chasing down a bug in the docs build.

src/pudl/metadata/classes.py
```diff
@@ -73,13 +73,13 @@
...
+# The 'schema' field in Resource shadows BaseModel.schema() (deprecated in Pydantic v2).
+# This is intentional to mirror the Frictionless Data standard and avoid the need to
+# define an inconvenient alias.
 warnings.filterwarnings(
     action="ignore",
-    message='Field name "schema" in "Resource" shadows an attribute in parent "PudlMeta"',
+    message=r'Field name "schema" in "Resource" shadows an attribute in parent',
     category=UserWarning,
-    module="pydantic._internal._fields",
 )
```
zaneselvans, line 83: This filter stopped working and got real annoying. Removing the (now changed) module fixed it.

src/pudl/metadata/classes.py
zaneselvans, line 1: Other than the filterwarning update, only docstring changes in here. Was chasing down a bug in the docs build.

src/pudl/metadata/constants.py
zaneselvans, line 1: Only docstring changes in here. Was chasing down a bug in the docs build.

src/pudl/transform/ferc1.py
zaneselvans, line 1: Only docstring changes in here. Was chasing down a bug in the docs build.

src/pudl/helpers.py
zaneselvans, line 1: Only docstring changes in here. Was chasing down a bug in the docs build.

pyproject.toml
```diff
@@ -1,6 +1,21 @@
...
+packages = ["src/pudl"]
+
+[tool.hatch.version]
+source = "vcs"
+
+[tool.hatch.version.raw-options]
+version_scheme = "guess-next-dev"
+local_scheme = "no-local-version"
+tag_regex = "^v(?P<version>\\d{4}\\.\\d{1,2}\\.\\d{1,2})$"
+git_describe_command = ["git", "describe", "--tags", "--long", "--match", "v20*"]
```
zaneselvans, line 17: Defines our versioning scheme.

pyproject.toml
```diff
@@ -1,6 +1,21 @@
...
-build-backend = "setuptools.build_meta"
-requires = ["setuptools>=80", "setuptools_scm[toml]>=9", "wheel"]
+build-backend = "hatchling.build"
+requires = ["hatchling", "hatch-vcs"]
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/pudl"]
+
+[tool.hatch.version]
+source = "vcs"
```
zaneselvans, line 9: `hatch-vcs` replaces `setuptools_scm` and reads versions from the git tags.

pyproject.toml
```diff
@@ -10,129 +25,205 @@ authors = [{ name = "Catalyst Cooperative", email = "pudl@catalyst.coop" }]
 requires-python = ">=3.13,<3.14.0a0"
 dynamic = ["version"]
 license = { file = "LICENSE.txt" }
-dependencies = [
```
zaneselvans, line 13: All dependencies moved under control of `pixi` below.

AGENTS.md
```diff
@@ -38,10 +42,14 @@
   tests should use the pytest.parametrize decorator to enumerate tests cases, specifying
   the appropriate success or failure or exception to be raised for each test as
   appropriate.
-- Tests must be run inside the `pudl-dev` conda environment.
-- For example, the unit tests can be run with `mamba run -n pudl-dev pytest test/unit`.
+- Tests must be run inside the pixi environment.
+- When individual tests are run, we should turn off coverage collection, since otherwise
```
jdangerx: nit: usually I just ignore the coverage failures when I run individual tests in dev :shrug:

docs/dev/nightly_data_builds.rst
jdangerx: No, I don't think we particularly prefer `aws` over `gcloud`. If it's part of our pixi env will it need to be invoked via `pixi run gcloud` though?

docs/dev/testing.rst
```diff
@@ -108,11 +77,16 @@ datastore instead by using our custom ``--tmp-data`` with ``pytest``:
...
 Running pytest Directly
 -------------------------------------------------------------------------------
-Running tests directly with ``pytest`` gives you the ability to run only tests from a
-particular test module or even a single individual test case. It's also faster because
-there's no testing environment to set up. Instead, it just uses your Python environment
-which should be the ``pudl-dev`` conda environment discussed in :doc:`/dev/dev_setup`.
-This is convenient if you're debugging something specific or developing new test cases.
+Running tests directly with ``pytest`` gives you the ability to run only
+tests from a particular test module or even a single individual test case.
+It's also faster because there's no testing environment to set up. Instead,
```
jdangerx: No longer true since we stopped using Tox, right?

docs/release_notes.rst
```diff
@@ -6,6 +6,27 @@ PUDL Release Notes
 v2026.X.x (2026-XX-XX)
 ---------------------------------------------------------------------------------------
 
+Application, not Library
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+From now on we will treat PUDL like an application rather than a library, since that's
+what it has become for all intents and purposes. There's no change in the licensing or
+openness of the project -- this is just a technical evolution, based on our experience
+that almost all users want the data, not the pipeline. Treating PUDL like a standalone
```
jdangerx: Hmm... I suspect that some users seem to use forked versions of our code to do their own custom transformations (@grgmiller maybe?) we should probably make sure this doesn't break their workflows.

jdangerx:
Notes from some digging and discussion:

Options discussed:

* :x: give up on pixi-in-pudl for now
  * pixi is worth it, we're burning time on this stuff and also on weird environment inconsistencies betweeen pudl archiver. plus easier for contribs
* try to keep the status quo working with pixi build backends
  * to try this out, we'd want to pin the dependency to the `pixi-setup` branch instead of just `main`
  * someone seems to have gotten this working (modulo a *different* issue that also got fixed) in https://github.com/prefix-dev/pixi/issues/1569
  * the pixi build backends seems to have matured a decent amount since that issue, though still marked as 'experimental'
* pull out data source metadata into own package
  * good thing to do, but we should think about this. code structure, docs and workflow impacts - probably kind of involved.
* pull pudl-archiver into pudl
  * eb: nice to keep it separated. pudl-archiver has a lot of things that aren't in pudl anymore. plus it'd be intimidating
* get rid of eager import of every submodule when we `import pudl` - which then allows us to do `from pudl.metadata.classes import Contributor, DataSource, License` without having all the fancy dependencies of `pudl` actually installed...
  * this is janky but we already install a version of `pudl` that doesn't really fully work...
  * we'd want to figure out a way to get the dependencies of `Contributor`, `DataSource`, and `License` into the `pudl-archiver` env - whether through `pudl/pyproject.toml` or through direct specification in `pudl-archiver/pyproject.toml`
  * we'd probably want to pull `Contributor`, `DataSource`, `License` out into their own submodules since `pudl.metadata.classes` includes a *bunch* of rando dependencies for the many classes and complicated behaviors defined in it.

Current plan/next steps:
* try using `pixi-build-python` pixi backend ([the docs](https://prefix-dev.github.io/pixi-build-backends/backends/pixi-build-python/) appear to assume you're using `pixi.toml` so you'll need to prefix all the section names with `tool.pixi.`) to replicate the existing pudl-archiver -> pudl dependency pattern
* if that fails, try experimenting with lazy imports in `pudl`
* if that fails, we should discuss the prospect of a bigger project to finally separate datasources into their own repo.

zaneselvans:
This took longer than I had hoped to get working, but I did learn a lot about how to make a conda package with pixi.

## Satisfying Bits

- The PUDL package now knows how to build itself as a conda package, and the pudl-archiver package is able to install and build the conda package from the git URL 🎉  PR here https://github.com/catalyst-cooperative/pudl-archiver/pull/974
- This means virtually all of the `pudl-archiver` dependencies are now installed from `conda-forge`.
- It also means that we don't need to keep the pudl-specific pins (e.g. for GDAL) in the `pudl-archiver` any more, since they will come through automatically when pixi solves the environment. This should alleviate a significant hassle we've had when those dependencies change and automate the inter-repo dependencies.

## Unsatisfying Bits

* I wasn't able to figure out how to avoid duplicating the list of runtime dependencies. So they show up in both: `[tool.pixi.dependencies]` and `[tool.pixi.package.run-dependencies]`. This seems odd to me, but I looked at some other pixi conda packages on github and they had duplicates as well. If we can figure out how to update the dependency version specs with `pixi upgrade` without blowing away all of the intentional pins and restrictions, maybe this won't be so annoying. And we don't need to change dependency pins very often. I'll handle it in the weekly updates.
* There's currently no way to obtain a dynamic version based on the git tags for consumption by pixi -- it can't read the version from `hatch-vcs` unfortunately, though there is an issue open for this functionality: https://github.com/prefix-dev/pixi/issues/2946.  for now I've hard-coded an obviously fake `v0.0.0`. The only place this should ever show up is in `pudl-archiver`.
* There's something I don't understand about how the playwright installation works, and it prevented me from adding it as a conda dependency, even though the most recent version of the package is available on `conda-forge`. @e-belfer do you understand what's going on there?

.github/workflows/pytest.yml
jdangerx, line 1: You ripped out a lot of debugging stuff here, I think that makes sense!

.github/workflows/update-lockfiles.yml
jdangerx: As I understand it, this would mean that once the update-lockfiles PR gets merged, people will pull locally and the next `pixi run` will tweak their environments slightly? Seems fine if that's the case, though obviously it would be nice if we understood why this was going on.

docker/Dockerfile
```diff
@@ -48,22 +48,13 @@ RUN mkdir -p ${PUDL_INPUT} ${PUDL_OUTPUT} ${DAGSTER_HOME} ${PUDL_REPO}
...
 
-# Copy conda-lock.yml in so we can build the conda environment and cache that layer in
-# the Docker image before installing PUDL.
-COPY environments/conda-lock.yml ${PUDL_REPO}/environments/conda-lock.yml
-# Create a conda environment based on the specification in the repo
-RUN micromamba create --prefix ${CONDA_PREFIX} --yes --file ${PUDL_REPO}/environments/conda-lock.yml && \
-    micromamba clean -afy
-
-# Copy the rest of the cloned PUDL repo into the image.
+# Copy the entire PUDL repo into the image.
```
jdangerx, line 51: nit - we could still copy the lockfile in first, so the env layer only gets rebuilt when the lockfile changes. But, if pixi is fast and we're not doing a ton of local building and re-building of the Dockerfile it's probably not a big deal.

docker/Dockerfile
jdangerx: Branch build success also indicates this Dockerfile is working!

docs/dev/data_validation_quickstart.rst
```diff
@@ -8,8 +8,8 @@ Setup
 -----
 
 The ``dbt/`` directory contains the PUDL dbt project which manages our `data tests
-<https://docs.getdbt.com/docs/build/data-tests>`__. To run dbt you'll need to have the
-``pudl-dev`` conda environment activated (see :doc:`dev_setup`).
+<https://docs.getdbt.com/docs/build/data-tests>`__ and dbt is part of the PUDL pixi
```
jdangerx:
```suggestion
<https://docs.getdbt.com/docs/build/data-tests>`__. dbt is part of the PUDL pixi
```

docs/dev/dev_setup.rst
```diff
@@ -27,29 +25,44 @@ Linux) and are already familiar with ``git``, GitHub, and the Unix shell.
...
+After installation, restart your shell or source your profile to make the ``pixi``
+command available.
+
+**Windows:**
+
+The PUDL development environment is not currently supported on Windows because some
+dependencies are not available for Windows through conda-forge.  Windows users should
+use the `Windows Subsystem for Linux (WSL)
+<https://learn.microsoft.com/en-us/windows/wsl/install>`__ and follow the Linux
+installation instructions above.
```
jdangerx: I think we should just point them at the pixi installation instructions and not vendor them here - if the install instructions change and we don't update this, it will just cause confusion.

docs/dev/dev_setup.rst
```diff
@@ -64,92 +77,97 @@ from your fork to your local computer where you'll be editing the code or docs.
...
-installed. You'll typically want to use the predefined ``make`` commands rather than
-running the individual commands they wrap. If you'd like to learn more about how
-Makefiles work, check out `this excellent Makefile tutorial
-<https://makefiletutorial.com/>`__
-
-To create the ``pudl-dev`` environment, install the local PUDL package, and activate the
-newly created environment, run:
+We use `pixi <https://pixi.sh>`__ to manage our development environment. Pixi uses a
+lockfile (``pixi.lock``) to specify particular versions of all of PUDL's direct and
+indirect software dependencies, resulting in a stable, reproducible environment.
```
jdangerx: This is duplicative with the `Install pixi` section above - if we do cut the installation instructions maybe it makes sense to just merge the two sections?

docs/conf.py
```diff
@@ -122,6 +122,12 @@
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ["_build"]
 
+# This is necessary because Sphinx 9.0+ is getting confused about class
+# attributes named type, and type[SomeClass] annotations in factory functions
+suppress_warnings = [
```
jdangerx, line 128: Might be good to include a reference to your GH issue: https://github.com/sphinx-doc/sphinx/issues/14223

docs/release_notes.rst
jdangerx: Have we talked with *any* users who want the pipeline over the data? @grgmiller and @gschivley - I'm not sure how PUDL fits into your various systems, will this cause problems for you?

.readthedocs.yaml
```diff
@@ -6,24 +6,22 @@
...
-
-# Set the version of Python and requirements required to build your docs
-python:
-  install:
-    - method: pip
-      path: .
+    python: "3.13"
+  jobs:
+    create_environment:
+      - asdf plugin add pixi
```
jdangerx, line 15: `asdf` seems cool - is it something that got here by adapting your local workflow, or were you fixing some sort of issue that was showing up without it?

AGENTS.md
jdangerx: Agree. Weird.

pyproject.toml
```diff
@@ -1,139 +1,318 @@
...
-    "finance",
-    "fossil fuel",
-    "natural gas",
-    "oil",
-    "policy",
-    "utility",
+license = "MIT"
+# NOTE: THERE ARE TWO COPIES OF MOST DEPENDENCIES.
+# Dependencies are specified here so that the pudl-archiver repo is able to install
+# PUDL with uv from the git repo URL. These are duplicates of the core runtime
```
jdangerx: Oh is *that* how the python pixi build backend works? Wild

pyproject.toml
jdangerx: If it ain't broke/slow don't fix it!

pyproject.toml
```diff
@@ -1,139 +1,318 @@
...
-    "ferc form 714",
-    "finance",
-    "fossil fuel",
-    "natural gas",
-    "oil",
-    "policy",
-    "utility",
+license = "MIT"
+# NOTE: THERE ARE TWO COPIES OF MOST DEPENDENCIES.
+# Dependencies are specified here so that the pudl-archiver repo is able to install
```
jdangerx: Slightly confusing that the list `here` is actually empty, and the deps are in `tools.pixi.*` instead.

.github/workflows/update-lockfiles.yml
zaneselvans: Yes that's' right.  If folks have the 'keep our version" git merge strategy set up, it should always prefer the local copy, since that's associated with `pixi.lock`

.github/workflows/zenodo-data-release.yml
zaneselvans: Lol, no.

pyproject.toml
zaneselvans: Based on your comment in the `pudl-archiver` PR and reviewing the pixi docs again I went ahead and removed the `PIXI_LOCKED`

docs/dev/nightly_data_builds.rst
zaneselvans: I have mixed feelings here.  If you just want to access our distributed data, then you super don't need to install our environment, and however you get `gcloud` working locally is fine.  Also if you use `pixi shell` (which I've been doing a fair bit of) then the `pixi run` prefix is unnecessary.

.github/workflows/release.yml
zaneselvans: Now that we've set up `catalystcoop.pudl` to be packaged by `pixi-build` I think we should add a workflow that attempts to run `pixi build` to verify that that's still functional. I guess we could also stash the built conda package in the GitHub release for future reference / safe keeping.

pyproject.toml
zaneselvans: This was stale. I've updated the comments to reflect the new organization of dependencies.

.readthedocs.yaml
zaneselvans: This.... is some shameless cut-and-paste.

docs/release_notes.rst
zaneselvans: I'll revisit this note given that we've now made PUDL into an installable `conda` package.

docs/dev/testing.rst
zaneselvans: I guess that's right. Will update.

docker/Dockerfile
zaneselvans: IIRC this was actually broken, because to install the environment, you need to figure out what the new version is, and that information comes from the git tags, so with just the lockfile copied in, the install failed because the `.git/` directory was missing, so `hatch-vcs` couldn't figure out the version.

docs/dev/data_validation_quickstart.rst
zaneselvans: Fixed locally.

docs/dev/dev_setup.rst
zaneselvans: Combined this and the other section you mentioned below and linked out to the installation instructions instead of repeating them here.

pyproject.toml
zaneselvans: No sorry, this was a stale comment.  But that's how pip/uv would have done the installation.

The `pixi-build-python` backend looks at the `tool.pixi.package.run-dependencies` section to see what dependencies need to be satisfied, while `pixi update` generates a `pixi.lock` based on the `tool.pixi.dependencies`

docker/Dockerfile
zaneselvans: I've added comment to the Dockerfile explaining this limitation.

PR #4878, opened 2025-12-29
Eqr batch build

### LINKED ISSUE #4847: Design batch job to run the EQR ETL and upload outputs to S3

zschira:
# Overview

Closes #4847.

This PR adds a new workflow to run the full EQR ETL as a google batch job and publish the results.

## Documentation

# Testing

I tested this by running the workflow to kick of a batch job and monitoring the results. See the job [here](https://console.cloud.google.com/batch/jobsDetail/regions/us-west1/jobs/run-ferceqr-etl-2025-12-29-1525-91fb43b40/details?project=catalyst-cooperative-pudl) and outputs at `gs://builds.catalyst.coop/ferceqr`

## To-do list

- [x] Make `generate_batch_config` script configurable
- [x] Create workflow to generate batch job using new configurable `generate_batch_config` script
- [x] Create new bash script to run EQR batch job that uses the dagster CLI
- [x] Implement automation tooling for publishing results on the completion of a job and sending slack notification
- [x] Make sure all logging is handled sufficiently (Copy raw logs to GCS and link in slack like normal batch job)
- [x] Update job to publish to AWS, not just GCS
- [x] Make build script return appropriate value based on results of dagster job


.github/workflows/build-deploy-ferceqr.yml
```diff
@@ -0,0 +1,107 @@
...
+on:
+  workflow_dispatch:
+
+env:
+  GCP_BILLING_PROJECT: ${{ secrets.GCP_BILLING_PROJECT }}
+  GCS_OUTPUT_BUCKET: gs://builds.catalyst.coop
+  BATCH_JOB_JSON: batch_job.json
+
+jobs:
+  build_and_deploy_eqr:
```
zschira, line 14: At some point all of the docker build stuff should be refactored into a reusable workflow.

.github/workflows/build-deploy-ferceqr.yml
```diff
@@ -0,0 +1,107 @@
...
+            --container-env AWS_DEFAULT_REGION=${{ secrets.AWS_DEFAULT_REGION }} \
+            --container-env AWS_SECRET_ACCESS_KEY=${{ secrets.AWS_SECRET_ACCESS_KEY }} \
+            --container-env BUILD_REF=${{ github.ref_name }} \
+            --container-env BUILD_ID=${{ env.BUILD_ID }} \
+            --container-env FERCEQR_BUILD=true \
+            --container-env GCP_BILLING_PROJECT=${{ secrets.GCP_BILLING_PROJECT }} \
+            --container-env GITHUB_ACTION_TRIGGER=${{ github.event_name }} \
+            --container-env OMP_NUM_THREADS=4 \
+            --container-env PUDL_BOT_PAT=${{ secrets.PUDL_BOT_PAT }} \
+            --disk-gb 1000 \
```
zschira, line 104: For now using the same hardware resources as nightly builds, but with 1TB of disk. I'm sure there's some tuning we could do here.

devtools/generate_batch_config.py
```diff
@@ -32,6 +32,9 @@ def to_config(
     container_env: list[list[str]],
     container_command: str,
     container_arg: str,
+    vcpu: int,
```
zschira, line 35: Adding configurable hardware resources.

docker/Dockerfile
```diff
@@ -10,7 +10,7 @@ SHELL [ "/bin/bash", "-exo", "pipefail", "-c" ]
 # Install some linux packages
 # hadolint ignore=DL3008
 RUN apt-get update && \
-    apt-get install --no-install-recommends -y git postgresql && \
+    apt-get install --no-install-recommends -y git postgresql inotify-tools psmisc && \
```
zschira: Added some CLI tools to use in the `gcp_ferceqr_etl.sh` script.

zaneselvans:
It looks like there may be a bunch of older unrelated commits hanging out on this branch, some of which I suspect are already on main, so you might want to do a `git rebase -i main` and cull the dupes.

docker/gcp_ferceqr_etl.sh
```diff
@@ -0,0 +1,46 @@
...
+
+function run_ferceqr_etl() {
+    echo "Running FERC EQR ETL"
+    # Launch dagster deployment in the background
+    initialize_postgres &&
+        authenticate_gcp &&
+        dagster dev &
+
+    # Kick off the ferceqr_etl job asynchronously
+    dagster job backfill --noprompt -j ferceqr_etl --location pudl.etl
```
zschira, line 35: When executing a job with the dagster CLI it always gets executed asynchronously in the background, so we need a way to know when the job is complete, which is why I'm using `inotifywait` to wait for a file named `SUCCESS` or `FAILURE` to appear in the `PUDL_OUTPUT` directory.

pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "coverage>=7.10",
     "dagster>=1.12.1",
     "dagster-dg-cli",
+    "dagster-webserver>=1.12.1",
```
zschira: This is needed in the EQR build to launch a local deployment.

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,145 @@
...
+
+from pudl.helpers import ParquetData
+from pudl.logging_helpers import get_logger
+from pudl.settings import ferceqr_year_quarters
+from pudl.workspace.setup import PudlPaths
+
+logger = get_logger(__name__)
+
+
+ferceqr_sensor_status = (
```
zschira, line 35: The sensor monitoring the EQR job will only ever be running if the `FERCEQR_BUILD` environment variable is set allowing us to ignore this functionality during local development.

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,145 @@
+"""Define tooling for monitoring the ferceqr_etl job during batch builds."""
```
zschira: This module implements a [dagster sensor](https://docs.dagster.io/guides/automate/sensors) and related tooling to manage monitoring the status of the EQR job and handling success/failure.

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,145 @@
...
+        attached_file_path=str(_get_etl_status_csv_path()),
+    )
+    _write_status_file("FAILURE")
+
+
+def _get_etl_status_csv_path() -> Path:
+    return PudlPaths().output_dir / "ferceqr_etl_status.csv"
+
+
+@dg.sensor(
```
zschira, line 167: This sensor runs every 60 seconds and will query the status of all the EQR assets to see if all partitions have been run, and if there were any failures. Once it detects that the job has completed, it will execute an asset to either handle a successful job or a failed one. I think the logic in this sensor could easily be generalized so we can apply this pattern to PUDL/SEC10k.

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,145 @@
...
+        return None
+    # The backfill must be complete to reach this point
+    # Check if any partitions failed
+    if (asset_statuses == dg.AssetPartitionStatus.FAILED).any(axis=1).any():
+        # Write status to CSV file
+        asset_statuses.to_csv(_get_etl_status_csv_path())
+
+        # Execute asset to send slack notification about failure
+        return dg.RunRequest(
+            run_key="ferceqr_deployment",
```
zschira, line 216: Returning a `RunRequest` with `run_key` set means that dagster won't try to launch a second run with the same key. This is important because it takes a long time to copy all of the outputs to GCS so this sensor will get executed multiple times while the `deploy_ferceqr` asset is still running.

We can make this key static because in our batch job context, dagster has no knowledge of previous runs.

zschira (commented):
## Design overview
The implementation of the EQR batch job is largely a simplified copy of the existing 
PUDL nightly build tooling, but it does all of the publishing outputs (to GCS only right now) and sending notifications in python using dagster tooling. I've decided to go this direction for a couple reasons, which I'll try to explain below:

### Reusability
The logic in the `gcp_pudl_etl.sh` script is very hard to reuse without just copy/pasting (or making it super complex by reusing the same script for different jobs and making it configurable). With my approach, the bash script is only really used to launch a dagster job, then all of the custom logic can be implemented alongside the rest of our python code. If we extended this pattern to PUDL, SEC10k, etc we could potentially have one script that just takes a dagster job name and will kick off that job. This would mean we could have a whole suite of jobs that use the same container and same top-level script with just some different arguments, then implement everything bespoke in python/dagster.
### Maintainability
We write lots of python as a team, and it's generally easier for us to maintain.

### Using the dagster CLI to run jobs
Because the dagster CLI launches jobs asynchronously, we need a way to tell when jobs have completed. This means we need the whole sensor setup to monitor the status of the job and write a file at the end that the top-level script can wait for (I'll explain more below).

## How it works
I've created a dagster sensor to monitor the status of the EQR job. This sensor will run every 60 seconds (in builds only, during local development it doesn't run at all). Once this sensor detects that the job has completed, it will check if it was successful or not and launch an asset to handle publication / notification (or just notification in the case of a failure). These assets will create a file at the end of their execution, which will signal to the top-level script that the job is complete and the script can continue with cleanup/exiting.

.github/workflows/build-deploy-ferceqr.yml
zaneselvans: From the console dashboard it seems like it's doing a great job of saturating both CPU and memory usage throughout almost the entire run, so we can probably give it more of both and reduce the processing time for the same price.

At the end there's like 30 minutes of very low resource utilization. Is this because it's uploading the outputs to GCS?  I suppose if we uploaded each file as soon as it was finished or wrote directly to an output staging bucket rather than waiting for the entire job to complete we could get rid of this little tail. But we'd need to make sure that the data is only published to the distribution location after the whole run has succeeded.  But that final bucket-to-bucket transfer could be done without a beefy VM running.

zschira: Yeah that down period is just copying data to GCS. If we wanted, we could just write parquet files directly to GCS without even writing them to disk to completely negate the copy.

As for publishing to S3 (which I'm not doing yet), we could theoretically do this with a free github runner, since it's entirely io-bound, but we'd need a way to kickoff the runner after a successful build.

zaneselvans: We're doing this in the nightly builds to run the Zenodo data release, which was similarly I/O bound (by Zenodo).  It was pretty straightforward to kick off a GitHub Action with `curl` in the shell script but we could also do it with the requests library or it looks like someone has made a [Python GitHub library](https://github.com/PyGithub/PyGithub). I could imagine that action also doing the data validation with DuckDB querying the remote data in the staging area, and then making the decision about whether to publish or not based on the results.  But maybe that is too janky and disconnected. Getting the data validation into the main Dagster workflow seems like it would be cleaner.

src/pudl/etl/__init__.py
```diff
@@ -245,7 +260,7 @@ def load_dataset_settings_from_file(setting_filename: str) -> dict:
                     }
                 }
             },
-            selection="not key:*_ferceqr*",
+            selection="not key:*ferceqr*",
```
jdangerx: nit:  If we're keying off of `groups` for the `ferceqr_etl` should we also key off of groups when we're filtering the EQR assets out? I guess here we're also filtering *out* the deployment assets, but that can also be keyed by group.

src/pudl/etl/__init__.py
```diff
@@ -7,14 +7,22 @@
...
+    AutomationConditionSensorDefinition,
+    DagsterRunStatus,
+    DefaultSensorStatus,
     Definitions,
+    JobSelector,
     define_asset_job,
     load_asset_checks_from_modules,
     load_assets_from_modules,
+    run_status_sensor,
+    sensor,
```
jdangerx: These added imports seem to not be referenced anywhere - do we still need them somehow?

src/pudl/etl/__init__.py
```diff
@@ -31,6 +39,7 @@
     ferc_to_sqlite_settings,
 )
 from pudl.settings import EtlSettings
+from pudl.workspace.setup import PudlPaths
```
jdangerx: This import also *seems* extraneous but may be sneakily needed somehow?

src/pudl/etl/__init__.py
```diff
@@ -113,7 +122,13 @@
     "out_state_demand_ferc714": [pudl.analysis.state_demand],
 }
 
-all_asset_modules = raw_module_groups | core_module_groups | out_module_groups
+automation_assets = {
```
jdangerx: `automation` seems like a vague name for this - we could just call them `ferceqr_deployment_assets` or something until we figure out what else we're putting in here...

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Don't do anything if there are still partitions that haven't finished running
+    if (in_progress_parts | not_started_parts).any():
+        logger.info("Partitions still in progress, continuing.")
+        return None
+    # The backfill must be complete to reach this point
+    # Check if any partitions failed
+    if (asset_statuses == dg.AssetPartitionStatus.FAILED).any(axis=1).any():
+        # Write status to CSV file
+        asset_statuses.to_csv(_get_etl_status_csv_path())
```
jdangerx, line 212: Nice, this seems like it was super useful in debugging :)

src/pudl/etl/ferceqr_deployment.py
jdangerx: This explanation is useful/timeless enough to put in a comment IMO!

docker/gcp_ferceqr_etl.sh
jdangerx: Cannot *believe* there's no way to query run status via CLI but here we are.

docker/Dockerfile
jdangerx: What do we get from `psmisc`?

jdangerx (commented):
Looks pretty good so far! Let me know when you've knocked out your todo list :)

.github/workflows/build-deploy-ferceqr.yml
zschira: Ah yeah seems like there's plenty of easy ways to work out kicking off a publication action. For the time being, I'm partial to just getting this PR in as is, then come back later to separate the publication out into its own action, but I don't feel too strongly if we want to just get it done now.

As for the validation piece, dagster added support for [DBT source asset checks](https://github.com/dagster-io/dagster/issues/18641), so it would be really nice if we could get that setup and then just do validation in asset checks. I tried some prototyping a while back, but wasn't quite able to get it working, but I didn't spend much time on this though and it didn't seem like it would be too hard.

src/pudl/etl/ferceqr_deployment.py
zschira: Added an extended description in the docstring!

src/pudl/etl/__init__.py
zschira: Yeah it was easier to just exclude all EQR assets, but I'm fine with changing this to groups if you feel like it would be cleaner/more consistent.

src/pudl/etl/__init__.py
zschira: Oh yeah these had been from some prototyping and I figured `ruff` would catch it, but maybe we're excluding this file? Either way, I just switched this file to do `import dagster as dg` and then using `dg.*` as that seems to be more idiomatic and importing every individual class/function is annoying.

src/pudl/etl/__init__.py
zschira: No also a hold over from prototyping. Fixed now.

src/pudl/etl/__init__.py
zschira: Changed to `ferceqr_deployment_assets`

src/pudl/etl/ferceqr_deployment.py
zschira: Hopefully! At first I was considering sending an a formatted table in the slack message, but decided it wasn't worth the hassle.

docker/gcp_ferceqr_etl.sh
zschira: Yeah there's also no super straightforward way to query the status of a backfill even with the python API, which is why I had to loop through all the assets/partitions and get the status individually.

docker/Dockerfile
zschira: This is for the `killall dagster` that happens once the job is over, which admittedly, might not be the most elegant way of shutting down dagster, but it seems to work fine.

.github/workflows/build-deploy-ferceqr.yml
zaneselvans: Before we merge, might we want to update this to deploy to the GCS mirror of our PUDL outputs, at:
`gs://pudl.catalyst.coop/ferceqr` instead of the builds bucket, to keep the S3 and GCS sources in sync?

This environment variable doesn't seem to be getting passed in to the Docker container below. Should it be?

At what point do we want to do the first distribution to S3?

devtools/generate_batch_config.py
zaneselvans: Some explicit documentation of what all these arguments mean in the docstring would be good.

docker/gcp_ferceqr_etl.sh
zaneselvans: If we wanted to only backfill some partitions, would we pass them in here, e.g. after identifying which of the archive zipfiles have a last-modified datestamp that's newer than the published parquet files? Or would we configure all of that using a sensor (?) inside Dagster that marks the right partitions as out of date based on those last-modified times?

docker/gcp_ferceqr_etl.sh
```diff
@@ -0,0 +1,72 @@
...
+echo "aws_secret_access_key = ${AWS_SECRET_ACCESS_KEY}" >>~/.aws/credentials
+set -x
+
+run_ferceqr_etl 2>&1 | tee "$LOGFILE"
+
+# This needs to happen regardless of the ETL outcome:
+pg_ctlcluster "$PG_VERSION" dagster stop 2>&1
+
+# Copy logs to GCS build directory
+gcloud storage --quiet cp "$LOGFILE" "gs://builds.catalyst.coop/ferceqr/"
```
zaneselvans: Should this URL be based on an envvar that's passed in from the GHA workflow? Or do we want to hard-code the distribution destinations in here?

Rather than (or in addition to?) dropping this in the `builds` bucket, I think we probably want to put it in the GCS distribution bucket at `gs://pudl.catalyst.coop/ferceqr` so that the GCS & S3 outputs remain in sync.

docs/release_notes.rst
```diff
@@ -11,6 +11,14 @@ Enhancements
 
 New Data
 ^^^^^^^^
+* Adds a new ETL for FERC EQR data, as well as associated infrastructure for running
+  the job and publishing outputs, which can be found at
+  ``s3://pudl.catalyst.coop/ferceqr``. There are 4 new tables which are produced by
```
zaneselvans, line 44: We should make sure this URL matches wherever we're actually writing the data.

docs/release_notes.rst
```diff
@@ -11,6 +11,14 @@ Enhancements
...
+  this ETL including, ``core_ferceqr__quarterly_identity``,
+  ``core_ferceqr__contracts``, ``core_ferceqr__quarterly_index_pub``, and
+  ``core_ferceqr__transactions``. Due to the size of this data, the tables are split
+  into a set of parquet files partitioned by year-quarter, and cannot be downloaded
+  as a single file like other PUDL tables.
```
zaneselvans:
```suggestion
  this ETL including, :ref:`core_ferceqr__quarterly_identity`,
  :ref:`core_ferceqr__contracts`, :ref:`core_ferceqr__quarterly_index_pub`, and
  :ref:`core_ferceqr__transactions`. Due to the size of this data, the tables are split
  into a set of parquet files partitioned by year-quarter, and cannot be downloaded
  as a single file like other PUDL tables.
```

src/pudl/etl/__init__.py
zaneselvans: I agree using the same selection method in both cases would feel more consistent and less likely to result in mistaken selections / misses in the future.

docker/gcp_ferceqr_etl.sh
zaneselvans: I'm not seeing the a LOGFILE in the `/ferceqr` output path. Should it be there?

src/pudl/etl/__init__.py
```diff
@@ -258,17 +262,19 @@ def load_dataset_settings_from_file(setting_filename: str) -> dict:
...
             ),
-            selection="key:*_ferceqr*",
+            selection=dg.AssetSelection.groups("raw_ferceqr")
+            | dg.AssetSelection.groups("core_ferceqr"),
         ),
     ],
+    sensors=[ferceqr_deployment.ferceqr_sensor],
 )
 
 """A collection of dagster assets, resources, IO managers, and jobs for the PUDL ETL."""
```
zaneselvans, line 280: If we want this string to annotate `defs` doesn't it need to be on the line immediately following the constant it's annotating?

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,159 @@
...
+    )
+
+
+def _write_status_file(status: Literal["SUCCESS", "FAILURE"]):
+    """Notify build script that job is complete by creating a status file."""
+    (PudlPaths().output_dir / status).touch()
+
+
+@dg.asset
+def deploy_ferceqr():
```
zaneselvans, line 109: What prevents this asset from running locally? We probably don't ever want to casually try and upload 200 GB of parquet files over a residential internet connection.

I don't see these assets when I select the `ferceqr_etl` job in the UI, but I do see their asset group and can select them manually and presumably run them, but if I try and select all EQR assets with `key:*fercqr*` in the UI they don't come up.

I think we probably want to have a developer docs page about running the FERC EQR ETL.

src/pudl/etl/ferceqr_deployment.py
zaneselvans: What does it mean for the sensor to be running vs. not running? What operations does that ultimately impact?

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,159 @@
...
+    """Publish EQR outputs to cloud storage."""
+    output_locations = [
+        "gs://builds.catalyst.coop/ferceqr",
+        "s3://pudl.catalyst.coop/ferceqr",
```
zaneselvans: The deployment URLs show up in the GHA workflow as an environment variable, as a hard-coded value in the build script, and also here.  Where do we actually want them to be defined? (also we should use `gs://pudl.catalyst.coop` not builds))

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,159 @@
...
+    ]
+    # Copy parquet files to GCS
+    logger.info("Build successful, copying ferceqr data to GCS.")
+    for output_location in output_locations:
```
zaneselvans: We've got `output_location` and `output_dir` and also `PUDL_OUTPUT` -- more descriptive names for the variables that help differentiate between these different kinds of places that data gets written would be helpful.

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,159 @@
...
+        for table in FERCEQR_TRANSFORM_ASSETS:
+            logger.info(f"Copying {table} to {output_location}.")
+            base_path = UPath(output_location) / table
+            base_path.mkdir(exist_ok=True)
+
+            # Loop through partitioned parquet files for table and write to GCS
+            for file in ParquetData(table_name=table).parquet_directory.iterdir():
+                destination_path = base_path / file.name
+                destination_path.write_bytes(file.read_bytes())
```
zaneselvans: Do we need some kind of try-except / retries here in the event that the copy fails so that we recognize that it failed and get a failure notification?

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,159 @@
...
+    an asset to handle either a successful or failed run. We need to set ``run_key``
+    in the ``RunRequest`` because dagster will only execute one run per key, so if
+    the sensor executes while one of handler assets is still in progress, dagster will
+    not try to execute the handler asset again. We can also make ``run_key`` a static
+    key, because our batch jobs have no memory of previous executions.
+    """
+    asset_statuses = {}
+
+    # Query status of all assets across all partitions
+    for asset_key in FERCEQR_TRANSFORM_ASSETS + ["extract_eqr"]:
```
zaneselvans: We should rename this multi-asset to `extract_ferceqr` since the name is showing up in other places.

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -0,0 +1,159 @@
...
+        axis=1
+    )
+    not_started_parts = asset_statuses[FERCEQR_TRANSFORM_ASSETS].isnull().any(
+        axis=1
+    ) & (asset_statuses["extract_eqr"] != dg.AssetPartitionStatus.FAILED)
+
+    # Don't do anything if there are still partitions that haven't finished running
+    if (in_progress_parts | not_started_parts).any():
+        logger.info("Partitions still in progress, continuing.")
+        return None
```
zaneselvans, line 207: This will keep running even if there are already assets that have failed, right?

Do we want to fast-fail as soon as we know that the run will not be successful? Or do we want to try and run everything even if some things have failed?

I guess it depends on what kinds of failures we end up seeing.

src/pudl/etl/ferceqr_deployment.py
zaneselvans: I don't understand the point about the key being static vs. having a memory of prior runs. Can you explain that more?

src/pudl/transform/ferceqr.py
```diff
@@ -39,11 +42,6 @@ def transform_eqr_table(
     relation_columns = table_data.columns
     if table_name == "core_ferceqr__quarterly_identity":
         relation_columns.remove("filing_quarter")
-    assert len(dtypes.keys()) == len(relation_columns), (
-        f"DuckDB dtypes has {len(dtypes.keys())} columns, but DuckDB relation has "
-        f"{len(table_data.columns)} columns in {table_name} for {year_quarter}. "
-        f"Columns in dtypes: {dtypes.keys()}. Columns in relation: {table_data.columns}."
-    )
```
zaneselvans, line 46: Was this check causing problems?

zaneselvans (changes requested):
- Rather than pushing the outputs to `gs://builds.catalyst.coop/` wouldn't we want to publish them to `gs://pudl.catalyst.coop/` as we do with the PUDL outputs now, so those two resource stay in sync? Are there any build-specific non-public outputs that we want to specifically save to the `builds` bucket?
- The S3/GCS paths to which we're deploying show up in several places and it's not clear which one is supposed to be controlling.
- Some naming / docstring requests.
- I think some dev documentation about how the EQR setup would be helpful, given that it's using a number of Dagster concepts we haven't touched before.
- The deployment asset should be aware of whether it's running in a deployment or locally, and not attempt to upload 200GB of data if it's local.
- I didn't see the `$LOGFILE` showing up in the builds bucket.

docker/gcp_ferceqr_etl.sh
zschira: I think we could explicitly pass partitions with the CLI, or we may be able to setup asset versioning to accomplish this, but we would need some kind of persistent storage between runs so dagster knows what versions were used in the previous run.

src/pudl/etl/__init__.py
zschira: I've updated this to be more consistent

devtools/generate_batch_config.py
zschira: Added help strings to the new args in the top level CLI.

src/pudl/etl/ferceqr_deployment.py
zschira: If the sensor is running, then every 60s it will get executed by Dagster to check if the EQR backfill has completed, and handle the completion appropriately. It only runs if `FERCEQR_BUILD` is set, which should only be true during batch builds. The sensor is ultimately what triggers execution of the success or failure assets, which is why we don't need to worry about executing them during local development.

I've expanded the module level docstring a bit to hopefully reflect this better.

src/pudl/etl/ferceqr_deployment.py
zschira: Basically, if we had a persistent dagster deployment, then we would need to generate a new string that is associated with each individual build, so we would only execute the success/failure asset once per build, but since we run on ephemeral infrastructure with no persistent storage, as far as dagster is aware, each build is the only one that has ever been executed, so we can just use a static key.

.github/workflows/build-deploy-ferceqr.yml
zschira: I've fixed this to pass the `GCS_OUTPUT_BUCKET` variable to the docker container, and I'm always referencing that variable when needed elsewhere. I also changed it to `gs://pudl.catalyst.coop/ferceqr`.

I've also set up the S3 path to be configured in the same way, and the next build should publish to S3, which I suppose we can do anytime now.

docker/gcp_ferceqr_etl.sh
zschira: I updated this to copy the log file after the last build, so we won't see this until we run another one. If we're ready to publish to S3 (which I guess there's no reason not to since we're not pointing to the data anyway) I can run this and verify that the logs are being handled correctly.

src/pudl/etl/__init__.py
zschira: Just removed the blank line here.

src/pudl/etl/ferceqr_deployment.py
zschira: These assets automatically get executed by the sensor during batch builds, but the sensor doesn't run in local development. You could still manually execute them through the catalog, so maybe we should not even load these assets in local development to avoid that accidentally happening.

src/pudl/etl/ferceqr_deployment.py
zschira: I've standardized this so you only need to set these URL's in the workflow, and they'll get referenced from there when used elsewhere.

src/pudl/etl/ferceqr_deployment.py
zschira: Renamed these variables.

src/pudl/etl/ferceqr_deployment.py
zschira: For now I lean towards leaving this as is and adding in error handling as it becomes necessary, but open to suggestion if there's something specific we want to look for here.

src/pudl/etl/ferceqr_deployment.py
zschira: Renamed.

src/pudl/etl/ferceqr_deployment.py
zschira: I figured it would be nice to get through all partitions before failing, that way we'll have more information to debug with when we can see exactly which partitions passed/failed.

src/pudl/transform/ferceqr.py
zschira: Early on I thought this had caused some problems, but I just added it back in and it seems to be working fine, so I've kept it in.

src/pudl/etl/ferceqr_deployment.py
zaneselvans: Or maybe load them conditionally if the `FERCEQR_BUILD` envvar is set, so you *can* make them load locally and run them if really want to for testing or something, but you have to go out of your way and can't do it accidentally.

.github/workflows/build-deploy-ferceqr.yml
```diff
@@ -5,7 +5,8 @@ on:
 
 env:
   GCP_BILLING_PROJECT: ${{ secrets.GCP_BILLING_PROJECT }}
-  GCS_OUTPUT_BUCKET: gs://builds.catalyst.coop
+  GCS_OUTPUT_BUCKET: gs://pudl.catalyst.coop
```
zaneselvans: I think the GCS bucket will also need the `/ferceqr` -- it has exactly the same structure as the `s3://pudl.catalyst.coop` bucket.

docker/gcp_ferceqr_etl.sh
zaneselvans: I was imagining we could just look at the timestamps on the raw archives, and see if they are newer than the timestamps on the existing outputs for each quarter of data.

docs/dev/ferceqr_data_builds.rst
```diff
@@ -0,0 +1,24 @@
...
+Notable differences from nightly builds
+---------------------------------------
+While the FERC EQR builds use much of the same infrastructure as nightly
+builds, there are a few notable differences. First, the data distribution
+is all done in Python using Dagster tooling. More information on how this
+works can be found in the :mod:`pudl.etl.ferceqr_deployment` module. One other
+major difference from the nightly builds is where data will be published.
+Due to the large size of the FERC EQR dataset, we do not maintain multiple
+versions of the data in S3, and instead have just one version, which will be
+overwritten each time a build is executed. This version can be found at
```
zaneselvans:
```suggestion
overwritten each time a build is executed successfully. This version can be found at
```

docs/dev/run_the_etl.rst
```diff
@@ -301,6 +301,25 @@ want to rerun the entire ETL.
...
+
+The FERC EQR ETL is contained in a Dagster job called ``ferceqr_etl``.
+Executing this job from the Dagster UI is slightly different from the main
+PUDL ETL jobs because the EQR job uses Dagster partitions. After selecting
+"Materialize All" (or "Materialize selected" for a selection of assets),
+a screen will popup allowing you to select the partitions to execute.
+From here you can select a set of year-quarter combinations. This will
+trigger a "backfill", which will execute each partition in its own "run".
+
+
```
zaneselvans, line 338: Should note in here the need to configure `dagster.yml` to use the `QueuedRunCoordinator` & set a `run_concurrency` that's appropriate to your machine's compute / RAM if you're going to materialize more than one quarter at a time.

src/pudl/transform/ferceqr.py
```diff
@@ -79,6 +82,8 @@ def core_ferceqr__quarterly_identity(
 ):
     """Apply data types to EQR ident table."""
     year_quarter = context.partition_key
+    logger.info(f"Transforming ferceqr identity table for {year_quarter}")
```
zaneselvans, line 75: The per-quarter logging output is definitely good.

.github/workflows/build-deploy-ferceqr.yml
zaneselvans:
```suggestion
  GCS_OUTPUT_BUCKET: gs://pudl.catalyst.coop/ferceqr
```

zaneselvans (approved):
Mostly minor docs tweaks. Only blocking thing is the `GCS_OUTPUT_BUCKET` URL needs the `/ferceqr` suffix just like the S3 one. But I will add it now.

docker/gcp_ferceqr_etl.sh
zaneselvans: Do we want to copy the logs to the public output bucket, or just stash them in the private builds bucket under a run-ID directory like we do for the PUDL build logs? It's pretty easy to accidentally expose a credential from an environment variable.

.github/workflows/build-deploy-ferceqr.yml
zschira: Oops, yeah good catch.

docker/gcp_ferceqr_etl.sh
zschira: That's a good point. I've changed this to publish logs in `gs://builds.catalyst.coop`.

docs/dev/run_the_etl.rst
zschira: Makes sense, added a section discussing config stuff.

src/pudl/transform/ferceqr.py
```diff
@@ -34,16 +37,6 @@ def transform_eqr_table(
 
     # Verify that all expected columns are included and that there are no extra columns
     assert set(dtypes.keys()) == set(col_expressions.keys())
-    # Can't check for identical values because we haven't transformed the table yet, but
```
zschira, line 37: Ended up removing this check after all. I realized it was causing issues for certain quarters of contracts data. The issue is that this is checking the columns from the `table_data` variable, but at this stage this just refers to the raw data, which will have different names and sometimes extra columns. The check above is also already verifying that the columns in the output table will match the defined schema.

In debugging this, I did find that the source of the issue was that all quarters of contracts data after the first 3 have an extra column called `seller_history_name` which we could keep if desired, but given that we actually have unique ID's for sellers, we probably don't need this for record linkage, so I'm not sure that it's useful.

zaneselvans: We have unique IDs, but unless we build a changelog table for the data associated with those CIDs, which will require significant changes to the CID archiver and ETL to save all historical versions, then we won't have any other versions of the attribute information associated with the sellers.  And even if we do build that changelog table, it'll only capture attribute changes going forward, since all historical version of the data have already been lost. So this could be worthwhile information to retain.

This is exactly the kind of unexpected weirdness I was trying to catch with this check -- drift in the the structure of the raw data relative to our expectations. We can't check that the names match (since it's still all the raw original names) but if the number of columns changes that should raise a red flag. So maybe it is worth explicitly skipping the specific cases in which we expect it to fail, so that any future column additions or deletions in the raw data will raise an error?

zaneselvans: (not saying this needs to be changed to merge -- but if we're not checking it we should create an issue so we don't forget and can fix it later)

zaneselvans:
`v3_final_Final_noreally.doc`

src/pudl/transform/ferceqr.py
zschira: That makes sense. I want to just get this PR merged, so I've added an issue (#4909) to track adding the column and adding a version of this check back in.

PR #4892, opened 2026-01-02
Update ferc1 data source docs

zaneselvans:
# Overview

- While doing some other maintenance I noticed that the FERC Form 1 data source docs page was pretty out of date and didn't reflect the current state of the project, so I updated it.
- Added Catalyst publications to the bibliography (presentations, software packages, etc.)
- Removed the incomplete and not really updated set of Zenodo dataset references in the bibliography, and replaced it with a general link to the Catalyst Cooperative Community, which houses all of the data archives.
- Replaced a few unnecessary hard-coded RTD URLs in the docs with RST relative links or references.
- Tried to clarify the new pixi setup docs.

## Testing

- Built the docs locally and inspected the pages I changed.

## To-do list

### DIFF
--- a/docs/dev/dev_setup.rst
+++ b/docs/dev/dev_setup.rst
@@ -80,7 +80,6 @@ our tests. Previously we used a mix of ``mamba``, ``conda-lock``, and ``make`` t
 provide this functionality. Pixi is much faster, more flexible, under very active,
 development, and lets us use one tool instead of three.
 
-
 See `the Pixi installation guide <https://pixi.prefix.dev/latest/installation/>`__.
 
 With ``pixi`` installed, to install all of PUDL's dependencies and set up the
@@ -91,19 +90,24 @@ development environment, from the root of the PUDL repository run:
     $ pixi install
 
 This will create the environment and install the PUDL package in editable mode. Pixi
-automatically activates the environment when you run commands with ``pixi run``, so you
-don't need to manually activate environments like you would with conda. If you prefer
-to enable the environment and forget about it, you can use ``pixi shell``, but note
-that this may complicate switching between different environments.
+automatically runs commands within the environment associated with your current
+directory when you prefix them with ``pixi run``, so you don't need to manually activate
+environments like you would with conda. If you prefer to enable the environment and
+forget about it, you can use ``pixi shell``, but note that you will need to run ``exit``
+to get out of that shell when switching environments.
 
-To see all available tasks defined in the project, run:
+``pixi run`` is also used to run tasks that have been defined in ``pyproject.toml``.
+To see all the available tasks defined in the project, run:
 
 .. code-block:: console
 
     $ pixi task list
 
-There's also additional information about running tests in the :doc:`testing`
-documentation.
+When defining a new task, make sure the name doesn't clash with a generic shell command,
+otherwise the named task will take precedence and make it impossible to run the shell
+command in the pixi environment.
+
+There's additional information about running tests in the :doc:`testing` documentation.
 
 -------------------------------------------------------------------------------
 Updating the PUDL Development Environment
--- a/docs/templates/ferc1_child.rst.jinja
+++ b/docs/templates/ferc1_child.rst.jinja
@@ -42,12 +42,12 @@ following criteria for three consecutive years prior to reporting:
 .. note::
 
    Publicly owned utilities and member-owned electric cooperatives are not required
-   to file the FERC Form 1. However, many of them do file the analogous USDA Forms
-   7 and 12.
+   to file the FERC Form 1. However, many of them do file the analogous USDA Rural
+   Utilities Service (RUS) Forms 7 and 12.
 
-Annual responses are due in April of the following year, with new XBRL filings being
-added to FERC's RSS feed on a rolling basis. Revisions and late filings typically
-continue to trickle in for months after the filing deadline.
+Annual responses are due in April of the following year, with new XBRL filings added to
+FERC's RSS feed on a rolling basis. Revisions and late filings typically continue to
+trickle in for months after the filing deadline.
 {% endblock %}
 
 {% block original_data %}
@@ -56,12 +56,12 @@ General Comments
 ~~~~~~~~~~~~~~~~
 
 The FERC 1 database is poorly normalized and the data itself does not appear to be
-subject to much quality control. For more detailed context and documentation on a
-table-by-table basis, look at :doc:`/data_dictionaries/ferc1_db`.
+subject to much quality control. For this reason, we strongly encourage the use of
+the processed FERC data in the PUDL database when it is available.
 
-Many of the FERC Form 1 tables in PUDL contain a ``record_id`` that indicates what
-original table and record in either the XBRL or DBF derived SQLite databases the data
-came from for forensic / data provenance purposes. The ``record_id`` is NOT in general a
+To identify the source data for each record, many of the FERC Form 1 tables in PUDL
+contain a ``record_id`` field that indicates what table and record originates from (in
+either the XBRL or DBF-derived SQLite databases). The ``record_id`` is not in general a
 unique identifier within a given PUDL database table, because in many cases the original
 data has been reshaped and normalized such that several records in the PUDL database can
 be traced back to a single record in the original data.
@@ -129,11 +129,21 @@ performed on the column inaccurate. Sometimes there are values reported for the 
 rows and not the individual plants making them difficult to simply remove. Moreover,
 these duplicate rows are incredibly difficult to identify.
 
-To improve the usability of this data we have manually compiled a mapping between the
-FERC utility IDs (which differ between the DBF and XBRL data) and EIA utility IDs,
-allowing individual utilities to be traced through all years of reporting across
-multiple datasets. We also manually map FERC plant names to EIA plant IDs, and associate
-related plant records across multiple years of reporting within the FERC data.
+To improve the usability of this data we have done a lot of work to link FERC plants
+and utilities to their EIA counterparts, and to associate related FERC plant records
+across different reporting years.
+
+* We have manually compiled a mapping between the FERC utility IDs for the DBF and XBRL
+  reporting eras.
+* We have manually compiled a mapping between those unified FERC utilities and the
+  corresponding EIA utilities.
+* We have compiled a manual mapping between the FERC plant names and corresponding EIA
+  plant IDs.
+* We have created a programmatic, probabilistic linkage between sub-plant parts
+  reporting in FERC Form 1 and the associated generators, prime movers, ownership, or
+  other slices of EIA plants.
+* We have created a programmatic, probabilistic linkage between related FERC plant
+  records, since FERC does not use plant IDs.
 
 We have only integrated a small subset of the available Form 1 tables. If there's data
 you need from Form 1 in bulk, you can `hire us <https://catalyst.coop/hire-catalyst/>`__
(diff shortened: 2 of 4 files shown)


docs/data_dictionaries/usage_warnings.rst
```diff
@@ -37,7 +37,7 @@ FERC:
   FERC data is notoriously difficult to extract cleanly, and often contains
   free-form strings, non-labeled total rows and lack of IDs. See `Notable
   Irregularities
-  <https://catalystcoop-pudl.readthedocs.io/en/latest/data_sources/ferc1.html#notable-irregularities>`_
+  <https://catalystcoop-pudl.readthedocs.io/en/latest/data_sources/ferc1.html#notable-irregularities>`__
```
zaneselvans, line 40: Was missing the double-underscore required for external links. I tried and failed to find an easy way to do an internal link to that header in the data source page. Adding a ref in RST didn't work for some reason.

docs/templates/epacems_child.rst.jinja
```diff
@@ -3,7 +3,7 @@
 Clicking on the links will show you a description of the table as well as the names and
 descriptions of each of its fields. Due to the size of the CEMS data we do not publish
 it as part of our SQLite database, but it is available through all of our other
-`access methods <https://catalystcoop-pudl.readthedocs.io/en/latest/data_access.html#>`__.
+:doc:`data access methods <../data_access>`.
```
zaneselvans, line 6: Hard-coded link will take the user to a potentially different version of the docs (`latest`) and will break if we go off RTD.

docs/catalyst_pubs.bib
```diff
@@ -33,123 +70,21 @@ @misc{PudlData
...
-@misc{PudlRawEia860,
-  author       = {Gosnell, Christina and Selvans, Zane},
-  title        = {{PUDL} Raw {EIA} Form 860},
-  year         = {2021},
-  howpublished = {Dataset archived on {Z}enodo},
-  publisher    = {{Z}enodo},
-  doi          = {10.5281/zenodo.4127026},
-  url          = {https://doi.org/10.5281/zenodo.4127026},
-  urldate      = {2021-11-01}
-}
```
zaneselvans, line 55: This list of data sources was incomplete and not really being updated, and yet also enormous. Removed in favor of a single reference to our top-level community.  If we could generate these automatically based on all the dataset DOIs that we own in Zenodo, or the local datastore config, maybe that would be worth including?

docs/catalyst_pubs.bib
```diff
@@ -168,3 +103,46 @@ @proceedings{lamb_2024_13948332
   doi          = {10.5281/zenodo.13948332},
   url          = {https://doi.org/10.5281/zenodo.13948332}
 }
+
+@misc{selvans_2019_2825635,
```
zaneselvans, line 107: Added refs to the software and presentations we have archived on Zenodo.

docs/dev/dev_setup.rst
```diff
@@ -89,7 +92,9 @@ development environment, from the root of the PUDL repository run:
 
 This will create the environment and install the PUDL package in editable mode. Pixi
 automatically activates the environment when you run commands with ``pixi run``, so you
-don't need to manually activate environments like you would with conda.
+don't need to manually activate environments like you would with conda. If you prefer
+to enable the environment and forget about it, you can use ``pixi shell``, but note
+that this may complicate switching between different environments.
```
e-belfer: Or 
```suggestion
that you'll need to run ``exit`` to exit the virtual environment.
```

docs/dev/dev_setup.rst
```diff
@@ -125,8 +131,18 @@ update just that dependency:
...
 
 Our automated GitHub Action handles weekly updates to all dependencies. You should
-not generally need to update all dependencies manually unless you're working on
-dependency management.
+not generally need to update all dependencies yourself unless you're working on
+dependency management. The command to update everything in ``pixi.lock`` is:
+
+.. code-block:: console
+
+   $ pixi update
```
e-belfer, line 143: Update updates within the specified range, upgrade upgrades regardless of pins - double-checking you mean update here?

docs/templates/ferc1_child.rst.jinja
```diff
@@ -17,40 +17,57 @@ respondent IDs, enabling a connection between the FERC and EIA data sets.
...
 * 500MWh of annual power exchanges delivered
 * 500MWh of annual wheeling for others (deliveries plus losses)
 
-Annual responses are due in April of the following year. FERC typically releases the
-new data in October.
+.. note::
+
+   Publicly owned utilities and member-owned electric cooperatives are not required
+   to file the FERC Form 1. However, many of them do file the analogous USDA Forms
+   7 and 12.
```
e-belfer: Flagging, we should eventually link to the USDA Data Source page here as part of #4889.

docs/templates/ferc1_child.rst.jinja
```diff
@@ -17,40 +17,57 @@ respondent IDs, enabling a connection between the FERC and EIA data sets.
...
+
+Annual responses are due in April of the following year, with new XBRL filings being
```
e-belfer:
```suggestion
Annual responses are due in April of the following year, with new XBRL filings
```
Non-blocking, think this is clearer.

zaneselvans:
Close and re-open to re-run CI and see if I also get the RTD failure with pixi installation.

docs/templates/ferc1_child.rst.jinja
```diff
@@ -17,40 +17,57 @@ respondent IDs, enabling a connection between the FERC and EIA data sets.
...
-As of early 2021, the data is now published as a collection of XBRL filings. Previous
-data remains in Visual FoxPro databases. The new data remains difficult to access and we
-are in the process of understanding the underlying data and integrating this new format
-into PUDL.
+Many of the FERC Form 1 tables in PUDL contain a ``record_id`` that indicates what
+original table and record in either the XBRL or DBF derived SQLite databases the data
+came from for forensic / data provenance purposes. The ``record_id`` is NOT in general a
+unique identifier within a given PUDL database table, because in many cases the original
+data has been reshaped and normalized such that several records in the PUDL database can
+be traced back to a single record in the original data.
```
e-belfer: Non-blocking. Slightly confusing wording, I would suggest something like:

To identify the origins of each record, many of the FERC Form 1 tables in PUDL contain a ``record_id`` field that indicates what original table and record in either the XBRL or DBF-derived SQLite databases the data originates from. The ``record_id`` is *not* in general a unique identifier within a given PUDL database table, because in many cases the original
data has been reshaped and normalized such that several records in the PUDL database can
be traced back to a single record in the original data.

docs/templates/ferc1_child.rst.jinja
```diff
@@ -17,40 +17,57 @@ respondent IDs, enabling a connection between the FERC and EIA data sets.
...
-.. seealso::
 
-    Explore the raw FERC Form 1 data at: https://data.catalyst.coop/{{ source.name }}
-    (old DBF) and https://data.catalyst.coop/{{ source.name }}_xbrl (new XBRL)
+General Comments
+~~~~~~~~~~~~~~~~
+
+The FERC 1 database is poorly normalized and the data itself does not appear to be
+subject to much quality control. For more detailed context and documentation on a
+table-by-table basis, look at :doc:`/data_dictionaries/ferc1_db`.
```
e-belfer: This would be a good place to add something like, "For this reason, we strongly encourage the use of the processed FERC data in the PUDL database for most use cases."

The link to the FERC1 DB here will probably send some people directly to the raw SQLite data, which most people don't want.

docs/templates/ferc1_child.rst.jinja
```diff
@@ -91,20 +121,22 @@ dropping the records with non-unique natural composite keys, we chose to preserv
 of the records and use surrogate auto-incrementing primary keys in the cloned SQLite
```
e-belfer, line 121: Non-blocking: we could point to the fact that our metadata now highlights and explains why there isn't a PK here.

docs/templates/ferc1_child.rst.jinja
```diff
@@ -91,20 +121,22 @@ dropping the records with non-unique natural composite keys, we chose to preserv
...
 
-To improve their usability, we have developed a complex system of regional mapping in
-order to create ids for each of the plants that can then be compared to PUDL ids and
-used for integration with EIA and other data. We also remove many of the duplicate rows
-and are in the midst of executing a more thorough review of the extraneous rows.
+To improve the usability of this data we have manually compiled a mapping between the
+FERC utility IDs (which differ between the DBF and XBRL data) and EIA utility IDs,
+allowing individual utilities to be traced through all years of reporting across
+multiple datasets. We also manually map FERC plant names to EIA plant IDs, and associate
+related plant records across multiple years of reporting within the FERC data.
```
e-belfer: This isn't just manual, no? There's an ML component here linking these the FERC-EIA records.

docs/catalyst_pubs.bib
e-belfer: Agree with this decision. Nothing comes to mind immediately, but maybe worth writing up a follow-up issue for?

e-belfer (approved):
While you're in here, can you update the reference to `make docs-build` in the Building Documentation page to `pixi run docs-build`?

I built the page and inspected it.

Otherwise, this mostly looks good! A few non-blocking suggestions here, and a stronger but still non-blocking suggestion to point people more forcefully at our version of the data rather than the raw SQLite DB.

docs/templates/ferc1_child.rst.jinja
zaneselvans: That's what I was thinking when I mentioned it... but nothing to link to quite yet. Will try to make sure it comes up if one greps.

docs/catalyst_pubs.bib
zaneselvans: See #4918

docs/dev/dev_setup.rst
zaneselvans: Right now in the PUDL repo we don't set upper bounds on the dependencies unless there's a known problem, so `pixi update` should upgrade to the most recent version of all the packages that aren't being held back intentionally.

There are a number of dependencies that are intentionally being held back since they have compatibility issues, so I don't want to use `pixi upgrade` in this repo since it blows away those intentional pins in `pyproject.toml` -- we could exclude them specifically, but there's enough of them that it's a bit annoying.

docs/templates/ferc1_child.rst.jinja
zaneselvans: Reworded to try and be more readable.

docs/templates/ferc1_child.rst.jinja
zaneselvans: All those direct-to-datasette links are dead right now anyway... so extra good to avoid this link.

docs/templates/ferc1_child.rst.jinja
zaneselvans: This is discussing the original SQLite DBs that are converted from DBF, for which we don't have this kind of metadata.

docs/templates/ferc1_child.rst.jinja
zaneselvans: - The mapping from FERC Plant Name to EIA Plant ID is manual.
- The mapping of individual plant parts is largely ML (these are ownership, prime mover, generator ID, etc. sub-plants)
- The association of FERC plant records across years is ML.

zaneselvans: I'll try to clarify.

PR #4905, opened 2026-01-06
Add metadata for USDA RUS forms 7 & 12

aesharpe:
Add metadata for USDA RUS forms 7 & 12 so we can make archives for them.

### DIFF
--- a/src/pudl/metadata/constants.py
+++ b/src/pudl/metadata/constants.py
@@ -237,6 +237,7 @@ CONTRIBUTORS: dict[str, dict[str, str]] = {
         "role": "contributor",
         "zenodo_role": "data collector",
         "organization": "University of Minnesota",
+        "orcid": "0000-0003-3969-505X",
     },
 }
 """PUDL Contributors for attribution.


src/pudl/metadata/sources.py
```diff
@@ -884,6 +884,75 @@
         "license_raw": LICENSES["us-govt"],
         "license_pudl": LICENSES["cc-by-4.0"],
     },
+    "usda7": {
+        "title": "USDA Form 7 -- Financial and Operating Report Electric Distribution",
+        "path": "https://www.rd.usda.gov/resources/forms/rus-forms",
```
aesharpe, line 889: I used the same link for both forms. There is no obvious webpage to use -- most link to PDFs that could easily go out of date.

zaneselvans:
@aesharpe I restarted the RTD build and it seems to be working now. I think it was just a transient error.

src/pudl/metadata/sources.py
```diff
@@ -884,6 +884,75 @@
...
+            "financial and operating report submitted by rural electric utilities "
+            "that receive funding from the USDA Rural Utilities Service (RUS) for "
+            "distribution services."
+        ),
+        "source_file_dict": {
+            "respondents": "Rural electric cooperatives borrowing from USDA RUS.",
+            "source_format": "Comma Separated Value (.csv)",
+        },
+        "working_partitions": {
+            "years": sorted(set(range(2006, 2022))),
```
zaneselvans, line 903: Given that this doesn't go through the present, do we need to add a metadata note about it not being updated? Or is the intention to update it?

src/pudl/metadata/sources.py
zaneselvans: Seems like as good a URL as there is.  I did a little bit of digging and found a couple of past versions of the blank forms and the form instructions from UEP Bulletins 1717B-2 (2002) & 1717B-3 (2006). There didn't seem to be more recent revisions.

Given that there's no public place to download the data in bulk it seems like noting in here that it was obtained by FOIA would be helpful, so folks don't go digging expecting to find the source data on the USDA site somewhere.

src/pudl/metadata/sources.py
```diff
@@ -884,6 +884,75 @@
...
+                + KEYWORDS["usda"]
+                + KEYWORDS["us_govt"]
+                + KEYWORDS["finance"]
+            )
+        ),
+        "license_raw": LICENSES["us-govt"],
+        "license_pudl": LICENSES["cc-by-4.0"],
+    },
+    "usda12": {
+        "title": "USDA Form 12 -- Financial and Operating Report Electric Power Supply",
```
zaneselvans: Given that we call "power supply" generation everywhere else, we might want to add it in parentheses here and in the description below, to help folks who are searching for that word.
```suggestion
        "title": "USDA Form 12 -- Financial and Operating Report Electric Power Supply (Generation)",
```

src/pudl/metadata/sources.py
```diff
@@ -884,6 +884,75 @@
         "license_raw": LICENSES["us-govt"],
         "license_pudl": LICENSES["cc-by-4.0"],
     },
+    "usda7": {
```
zaneselvans: Searching around for documentation of these forms, it looked like there were several different USDA sub-agency forms with these numbers in addition to the RUS Forms 7 & 12, so `usda7` and `usda12` may not be unique identifiers. We've used PHMSA (a sub-agency) rather than USDOT for the PHMSA natural gas. Maybe it makes sense to use RUS here instead of USDA? `rus7` and `rus12`?

E.g.
* [Animal and Plant Health Inspection Service (APHIS) Form 12](https://www.aphis.usda.gov/sites/default/files/APHIS12.pdf)
* [Wildlife Service Form 12](https://corpslakes.erdc.dren.mil/employees/invasive/pdfs/WS%20Form%2012A%20Mar07.pdf)
* [Food & Nutrition Service (FNS) Form 7](https://www.fns.usda.gov/form/fns-7) (Destination Data for Donated Foods)

src/pudl/metadata/constants.py
```diff
@@ -376,6 +376,14 @@
         "water energy nexus",
         "energy water nexus",
     ],
+    "usda": [
+        "usda",
+        "rus",
```
e-belfer, line 390: Non-blocking: "rural utilities service" and "rural utility" would maybe also be helpful here

src/pudl/metadata/sources.py
e-belfer: The intention is to hopefully update it, but also this note will want to live in the datasource docs, not in the dataset description.

src/pudl/metadata/sources.py
```diff
@@ -884,6 +884,75 @@
...
+            "distribution services."
+        ),
+        "source_file_dict": {
+            "respondents": "Rural electric cooperatives borrowing from USDA RUS.",
+            "source_format": "Comma Separated Value (.csv)",
+        },
+        "working_partitions": {
+            "years": sorted(set(range(2006, 2022))),
+        },
+        "contributors": [
```
e-belfer, line 905: In this context, I think we want to add the person who actually submitted the FOIAs, Matt Grimley.

src/pudl/metadata/sources.py
e-belfer: I disagree - I think referencing generation in the description makes sense but the form title itself should be 1:1 with the original form name.

src/pudl/metadata/sources.py
e-belfer: I agree, that seems like as good a link as any and adding a note about the FOIAs would be helpful here. Something like:

"This data was obtained through multiple Freedom of Information Access (FOIA) requests."

Should be fine given that this winds up in many different places.

src/pudl/metadata/sources.py
```diff
@@ -884,6 +884,75 @@
         "license_raw": LICENSES["us-govt"],
         "license_pudl": LICENSES["cc-by-4.0"],
     },
+    "usda7": {
+        "title": "USDA Form 7 -- Financial and Operating Report Electric Distribution",
```
e-belfer:
```suggestion
        "title": "USDA Form 7 -- Financial and Operating Report, Electric Distribution",
```
Or another dash would be fine, just something to break up these two parts would help. Same for Form 12.

e-belfer (changes requested):
My main request is to add Matt to the contributors (and agree with Zane, to mention the FOIA'ing in the description), otherwise minor non-blocking suggestions.

The internal code is mostly for our purposes, and technically they're not even called Form 7 and Form 12 anymore (they're just "Financial and Operating Report - Electric Distribution", etc.), but I think that calling them `rus7` and `rus12` makes sense given that its a sub-agency with a number of its own reports we might want to integrate down the line.

src/pudl/metadata/sources.py
aesharpe: hmmm that's a good point. We've also combined them in the past (epacamd) but I don't have a strong preference either way -- I'll just do `rus`

src/pudl/metadata/sources.py
aesharpe: I think I agree with Ella here. I'll make sure generation gets referenced in the description.

src/pudl/metadata/constants.py
```diff
@@ -230,6 +230,13 @@
         "zenodo_role": "producer",
         "organization": "Moment Energy Insights",
     },
+    "matt-grimley": {
+        "title": "Matt Grimley",
+        "email": "griml011@umn.edu",
+        "path": "https://www.hhh.umn.edu/directory/matthew-grimley",
+        "role": "contributor",
+        "organization": "University of Minnesota",
+    },
```
aesharpe: Questions for this: 
- Should I use his full name "Matthew"? 
- is his "role" correct? 
- should he have a "zenodo_role"?

e-belfer: 1) yes, 2) yes, I think that's the most accurate (https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/contributor/), and 3) I think zenodo_role would most accurately be "data collector", but we'll probably need to do this by hand anyways since the interface with the contributors via the API is still a bit clunky.

e-belfer (approved):
This looks good, thank you! See my non-blocking comments about the Zenodo contributor metadata.

src/pudl/metadata/constants.py
zaneselvans: Might want to check and see whether he has an ORCID so he can be associated with / credited to the archives in a machine readable way.

PR #4879, opened 2025-12-31
Add FERC EQR data source docs page

katie-lamb:
# Overview

Closes #4852

## Core changes

- Flesh out the `ferceqr` data source documentation page. See #4852 for checklist of contents.
- Add `ferceqr` page to data source docs index.
- Add historical FERC EQR data dictionaries to our docs for download / context.
- Create a [metadata-only FERC EQR archive on Zenodo](https://zenodo.org/records/18251902) to provide public access to `datapackage.json` and make use of existing infrastructure with a minimum of changes.
- Modify the `Datastore` class to ignore download requests for URLs that start with `gs://` so it doesn't barf on GCS backed metadata-only data packages.
- Add more keywords to the EQR data source.

## Datastore Housekeeping

While I was rummaging around in here trying to make the datastore work for the EQR, I fixed some other things.

- Fill in missing concept DOIs in all data source metadata.
- Display concept DOI as a link to Zenodo on all data source docs pages.
- Remove `eiawater` DOI since we don't actually touch that dataset in PUDL.
- Update `eia191` and `eia757a` DOIs to the most recent archives available.
- Do a metadata-only update to the `censusdp1tract` archive to modernize it.
- Remove legacy Datastore code that only existed to work with very old archives.

## Discussion of Zenodo Workaround

After reviewing #4911 I decided to try the "metadata only" Zenodo archive because:

- To make the docs build and other metadata access work for everyone, we needed to put the `datapackage.json` file somewhere public anyway.
- It re-uses the infrastructure we already have, requires minimal changes to the existing code, and doesn't result in semantic inconsistencies like having a "doi" that's not really a DOI.
- Having the metadata-only archive on Zenodo provides potential users a signpost to the dataset if they're used to looking for our data there already, and also gives the dataset a DOI that can be cited.
- The `datapackage.json` on Zenodo is conceptually almost perfectly fine -- the `path` URLs point to where the data really is, even if it's not accessible to the user (but we could make it requester-pays if we wanted to).

### Shortcomings

- Right now we need to remember to manually upload a new `datapackage.json` if/when we update the EQR snapshot. I'm sure we could automate this, but for now we haven't.
- There's currently no concept of dataset versioning in the FERC EQR data, but there is in the Zenodo archive, so it's possible for them to mismatch.  The file sizes and hashes will fail to match if this happens, and there might be more resources available in a newer snapshot of the data.

# Testing

How did you make sure this worked? How can a reviewer verify this?

[...]

PR #4911, opened 2026-01-09
fix: poke at datastore GCS datapackage workaround

### LINKED ISSUE #4929: Make Datastore able to fetch non-Zenodo data / metadata
# Overview

- Our data source documentation pages are dynamically generated during the docs build.
- Some of the information they display ultimately comes from the `datapackage.json` that describes the Zenodo archive for that data source.
- This information is retrieved by `Datastore.get_datapackage_descriptor()` which is called by `DataSource.add_datastore_metadats() which is called by `DataSource.to_rst()`
- Unfortunately none of this works if the data source is not archived on Zenodo, as is now the case with the FERC EQR data.
- This means we can't build the FERC EQR data source documentation page without some kind of workaround or generalization of the metadata fetching.
- We do have the necessary metadata at `gs://archives.catalyst.coop/ferceqr/published/datapackage.json` so it should be possible to generalize the functionality that's currently part of the `ZenodoFetcher` to allow it to fetch data and metadata from other kinds of data sources, like a `UPath`.
- However, the assumption that the data will be coming from Zenodo is pretty entangled with the system right now so this looks like it would take a bit of thinking and effort.

## Thoughts / Questions / Concerns

- If the metadata is being pulled from a private cloud bucket (as with the FERC EQR currently) then no outside contributor will ever be able to build the docs.
- Right now `add_datastore_metadata()` always tries to access `datapackage.json` via a `local_cache_path`. If the data isn't there, it'll go to Zenodo, which is flaky. It never tries to use cached data from S3, because the `Datastore` isn't passed in, it's created locally with a generic `PudlPaths` object.

jdangerx:
## Things done

* Add a GCS fetcher to go with the Zenodo fetcher - plus an interface they both share so that datastore can just do Fetcher things.
* ran `pixi run docs-build` and it seems to work on this branch, but not on `ferc-eqr-data-source-docs` branch, so something must have gotten fixed ;)

## Things ran into, but ignored:

* now `DatapackageDescriptor.doi` no longer exclusively refers to a DOI - this would be an easy refactor to something like `identifier` probably.
* resource fetching only works for Zenodo datasets, so I didn't add it to the `DatasetFetcher` interface. Instead I added a big ol comment around where I'd expect things to fail when someone tries to call `ds.get_resources` on a GCS dataset.


zaneselvans:
Unfortunately the `datapackage.json` that we are trying to access here is in a private bucket, so we won't be able to build on RTD, and no outside contributor will be able to build the docs either.

zaneselvans:
I ended up taking a different approach, at least for now. See discussion in #4879.

PR #4931, opened 2026-01-13
Add workaround for click issue 3110

zaneselvans:
# Overview

- Workaround for https://github.com/pallets/click/issues/3110 which was causing one of our `dbt_helper` tests to fail.
- This broke the builds because we updated to a new version of Click, exposing this regression.

# Testing

To exercise this fix run:
```console
pixi run pytest test/integration/dbt_test.py::test_validate_asset_selection
```

On `main` it will fail, but on this branch it should succeed.


krivard (approved):
🐿️

PR #4932, opened 2026-01-13
Extract RUS7 Energy Efficiency table

cmgosnell:
# Overview

Working on https://github.com/catalyst-cooperative/pudl/issues/4897.

## Documentation

# Testing

## To-do list


aesharpe (approved):
A couple small things-- 

1) I would maybe say: `residential_excluding_seasonal_invested_in_report_year`
instead of `residential_excluding_seasonal_invested_new_report_year` -- this happens in a few different columns. (in vs. new) for clarity.

2) I think what's so tricky about these chunky columns (also with the `Power Requirements` table) is the placement of the word `customers`: 

`residential_excluding_seasonal_customers_num_cumulative` should maybe be: 
`residential_customers_excluding_seasonal_num_cumulative`

3) I'm kind of torn about whether to put the word efficiency in any of the column names. It adds extra length, but also helpful context. I know this is the "efficiency" table so it should be self evident, but columns like `residential_excluding_seasonal_invested_new_report_year` are kind of hard to wrap your head around. I think this might be solved by pivoting the table in the transform so there is a column for sector (residential, etc.), a column for cumulative vs. report year, a column for value type (no. cust, investment, savings) and a column for efficiency program value that contains the actual value...

cmgosnell:
these raw column names are silly long but i super super expect we will pivot them greatly to end with a schema that looks something like this:

```python
energy_efficiency = [
    "borrower_id_rus",
    "borrower_name",
    "report_date",
    "customer_classification",  # enum w/ things like residential_excluding_seasonal, residential_seasonal	irrigation etc.
    "date_range",   # ?idk about that name enum w/ new_in_report_year, cumulative
    "customers_num",
    "invested",
    "estimated_savings_mmbtu",
]
```

so:

1. imma DO BOTH bc i think in a date_range column it'll be readable to have `new_in_report_year` / it sounds relatively like "Added this year" from the form.
2. I totally hear you that these raw cols are hard to read... but does it make more sense with my imagined schema? (aka `customer_num` will be its own column)
3. I waffle on adding `efficiency`... I can see it being helpful extra context but it would clutter the names. If its okay let's hold off on that for now.

PR #4936, opened 2026-01-13
Extract RUS 12 Financial and Operating Report - Sources and Distribution table

aesharpe:
# Overview

Extracts RUS 12 Financial and Operating Report - Sources and Distribution table. Part of #4900.

## What did you change?
- Add package data for this table


src/pudl/extract/rus12.py
```diff
@@ -83,5 +83,6 @@ def _raw_rus12__page(raw_rus12__all_dfs: dict[str, pd.DataFrame]):
     for in_page, out_page in {
         "statement_of_operations": None,
         "balance_sheet": None,
+        "financial_and_operating_report_sources_and_distribution": None,
```
aesharpe: iffy on the name

cmgosnell: oof ya thats looooong

cmgosnell: if im reading the [blank form](https://www.rd.usda.gov/sites/default/files/OpRpt_PS_2013_Current.pdf) correctly, the `financial_and_operating_report` is all of form 12. So this table could just be `sources_and_distribution`

cmgosnell (changes requested):
gosh this table is weirdly structured. good job mapping! smol suggestions
- `*_num_plants` - > `*_plants_num` bc num in this case is a unit and we always put those at the end
- i would prolly spell out `cc` and `ic` to be like the rest but that's a nit and i will not die on that hill (i assume we will pivot this table and the those plant types will be categorical values not long long parts of pudl column names)
- also non-blocking but i've been trying to put total in front of things that seem calculated so they are easier to flag. so like `net_interchange_mwh` -> `total_net_interchange_mwh` but not `total_purchased_power_mwh` -> `purchased_power_mwh`

(also these net's make sense to me and should stay bc they do mean something!)

PR #4937, opened 2026-01-13
Extract RUS 12 Renewable Summary table

aesharpe:
# Overview

Extracts RUS 12 Renewable Generation Plant Summary table. Part of #4900.

## What did you change?
- Add package data for renewable summary table
- Rows with csv quotation issues are marked with -1 in the file_map.csv and need to be revisited. Note that there are no renewable summary tables prior to 2010 so those -1 values can stay.


src/pudl/extract/rus12.py
```diff
@@ -85,5 +85,6 @@ def _raw_rus12__page(raw_rus12__all_dfs: dict[str, pd.DataFrame]):
         "statement_of_operations": None,
         "balance_sheet": None,
         "sources_and_distribution": None,
+        "renewable_summary": None,
```
cmgosnell, line 88: this is a table with plants in it? imo it should be like the other plants tables and be `renewable_plants`

cmgosnell (changes requested):
i have some thoughts about column naming mostly to try to keep the columns as similar as possible to our existing column names.

- what is your thinking about putting rus at the end of these? we don't do that with other datasets like `prime_mover_code`
   - prime_mover_name_rus
   - primary_renewable_fuel_type_id_rus
   - primary_renewable_fuel_type_name_rus
- we use `capacity_factor` not `capacity_factor_pct`
- we use `opex` for O&M

if there will be different enums or something we can always define dataset or table specific fields.

aesharpe:
> what is your thinking about putting rus at the end of these? we don't do that with other datasets like `prime_mover_code`
> * prime_mover_name_rus
> * primary_renewable_fuel_type_id_rus
> * primary_renewable_fuel_type_name_rus

I think any time we have an ID field we should specify what org that ID is coming from.
The other ones I was just thinking would provide some clarity / specificity but I'm happy to remove them too.

For now I'll just remove all the extra `rus` suffixes.

PR #4943, opened 2026-01-15
Extract RUS 12 Power Supply Loans table

aesharpe:
# Overview

Extracts RUS 12 Power Supply Loans table. Part of #4900.

## What did you change?

- Add package data for power supply loans table


aesharpe:
@cmgosnell @e-belfer I'm wondering whether we should add a `power_supply` prefix to all of the rus12 tables (and by extention a `distribution` prefix to the rus7 tables. It would definitely make them more intuitive for users.

Otherwise this table should probably get renamed to just `loans`

PR #4946, opened 2026-01-15
Extract RUS12 Lines and Stations Labor and Materials table

aesharpe:
# Overview

Extracts RUS 12 Lines and Stations Labor and Materials table. Part of #4900.

## What did you change?

- Add package data for lines and stations materials and labor table.


cmgosnell (approved):
mapping looks great!

schema-wise its so silly but I can see pulling out `employees_num` into its own smol table and stacking the rest of the table.

PR #4962, opened 2026-01-20
Handle partitioned assets in datapackage outputs

### LINKED ISSUE #4831: Update datapackage output to handle EQR partitioned parquet files

zschira:
# Overview
Closes #4831.

This PR modifies our `datapackage.json` outputs to handle tables which are distributed as partitioned parquet files. For these tables (only applies to EQR tables for now), the `path` field in for the `Resource` corresponding to the table will contain a list of paths pointing to all parquet tables which make up the table.

This allows eel-hole to understand EQR outputs making exploration of this dataset possible alongside the rest of PUDL data.

## What did you change?
I modified the `to_frictionless` methods of the `Resource`/`Package` classes to handle the partitioned path outputs. To make this work, I added a new field to the `PudlResourceDescriptor` class called `output_partition_source_key`, which specifies a specific `datasource` and `working_partitions` key that corresponds to the output partitions.

## To-do list

- [x] Add partitioned path capabilities
- [x] Update EQR resource descriptors to use new partitioned metadata outputs
- [x] Fix tests
- [x] Update the `datapckage.json` creation in nightly builds.

### DIFF
--- a/src/pudl/metadata/sources.py
+++ b/src/pudl/metadata/sources.py
@@ -695,7 +695,8 @@ SOURCES: dict[str, Any] = {
             "various required forms. Each regulatory program requires a unique CID number. These "
             "CID numbers show up in various FERC filings, such as Forms 1, 2, 6, 60, 714, and EQR."
         ),
-        "working_partitions": {},
+        # we only want the data table, not the data dictionary table
+        "working_partitions": {"data_set": "data_table"},
         "keywords": sorted(
             set(
                 ["ferc", "cid", "company identifier"]


src/pudl/metadata/classes.py
```diff
@@ -1147,6 +1148,17 @@ def serialize_df(
...
 
+    class PudlPartitionDescriptor(PudlMeta):
+        """Contains necessary information to construct paths for partitioned tables.
+
+        Certain large tables must be output as a directory of partitioned parquet
+        files to avoid massive unwieldy parquet files. This is done by specifying
+        a datasource and working partition key.
+        """
+
+        datasource: str
```
zschira: This works by getting a set of `working_partitions` from a `DataSource`. A single `Resource` can come from more than one `DataSource`, which is why we need to specify the `datasource` here even though `ferceqr` only has one source. It seems totally possible that we could have a table that's partitioned in a way that doesn't line up cleanly with the `working_partitions` of an input source, so we may need a different way to handle specifying the output partitions in the future.

src/pudl/metadata/classes.py
```diff
@@ -1635,6 +1651,23 @@ def dict_from_resource_descriptor(  # noqa: C901
                         f["encoder"] = encoder
                         break
 
+        # Construct path based on output partitions
```
zschira: Use the `PudlPartitionDescriptor` to create path(s) during the construction of a resource. This assumes that each partition in `working_partitions` will correspond to a single parquet file in the directory `$PUDL_OUTPUT/{resource_name}/{partition}.parquet`.

src/pudl/metadata/classes.py
```diff
@@ -1678,19 +1711,23 @@ def to_sql(
 
     def to_frictionless(self) -> frictionless.Resource:
         """Convert to a Frictionless Resource."""
-        schema = frictionless.Schema(
-            fields=[
-                frictionless.Field(name=f.name, description=f.description)
+        schema = {
```
zschira: After some discussion in my [issue](https://github.com/frictionlessdata/frictionless-py/issues/1761) in the `frictionless-py` repo, I realized that the `Resource` class can't have a `path` that is a `list[str]`, but there is an `extrapaths` field that is used for handling this. However, if you use the python API to load / validate a `datapackage.json` that has a list of strings for the path field it works just fine, so I've switched this so now our `Package.to_frictionless` will write the `datapackage.json` file first, then load it as a `frictionless.Package` afterwards.

jdangerx: Hmm, hitting disk seems like a weird side-effect to have here - we should probably at least document it in the docstring.

`extrapaths` seems like the more "intended" way of doing this, did you run into any problems doing it that way?

src/pudl/metadata/classes.py
jdangerx: In the spirit of moving this sort of conditional logic to the caller, I think it might be cleaner to allow the `PudlResourceDescriptor` to take `path`/`extrapaths` as a parameter, and create those in the resource definitions themselves - then we also don't have to worry that we're putting this EQR-specific logic in a general-resource-handling location.

We could default `Resource.path` to `{resource_id}.parquet` using a `default_factory`.

I don't know if you ran into issues with `extrapath` in the frictionless `Resource` constructor, but if that worked fine, we could then call `frictionless.Resource(..., path=self.paths, extrapaths=self.extrapaths)` when converting to frictionless.

src/pudl/metadata/classes.py
```diff
@@ -1635,6 +1651,23 @@ def dict_from_resource_descriptor(  # noqa: C901
...
 
+        # Construct path based on output partitions
+        if (part_key := descriptor.output_partition_source_key) is not None:
+            [source] = [
+                source
+                for source in obj["sources"]
+                if source.name == part_key.datasource
+            ]
+            obj["path"] = [
+                f"{resource_id}/{partition}.parquet"
```
jdangerx: See [comment on original issue](https://github.com/catalyst-cooperative/pudl/issues/4831#issuecomment-3770352498) but I think we should use fully qualified URLs instead of relative paths to be compliant with the spec.

src/pudl/metadata/classes.py
zschira: The issue with doing this is that when we call `to_json` on the resulting `frictionless.Package` it doesn't combine `path` and `extrapaths`, so we end up with an `extrapaths` field. We could certainly work with this and look for `extrapaths` on the eel-hole side, but given that `extrapaths` is not documented in the frictionless standard, I was hesitant to go down that route.

I also tried constructing the `frictionless.Package` from a dict rather than going to disk first, but it still ends up separating the `path`/`extrapaths`. This is why I decided to only really use the official `frictionless` package for validation after already writing to disk. In practice the only place we actually call `to_frictionless` is when we are actually writing the `datapackage.json` to disk, so this seemed like an ok arrangement.

I could see just using `frictionless` directly and accepting the whole `path`/`extrapaths` thing since the tooling really seems to be pushing us in that direction.

src/pudl/metadata/classes.py
zschira: I didn't add `path`/`extrapaths` because I didn't want to have to define all of the partition info in two places, since everything we need to construct the paths already exists in the partitions defined at the `datasource` level. This way when we integrate new quarters, we only have to remember to update the partitions in `src/pudl/metadata/classes.py`.

I'm definitely not married to this setup, but I do think it's worthwhile to avoid redefining the partitions as I could very easily see us forgetting to update the `path`/`extrapaths`, and then we would have missing partitions in eel-hole. One alternative could be to just import the big `SOURCES` dictionary in `src/pudl/metadata/resources/ferceqr.py` and pull out the `working_partitions` that way. Neither option feels super clean to me, but this alternative might at least make it a bit more explicit how the paths are constructed.

src/pudl/metadata/classes.py
zschira: I see 3 options for handling this:

1. Just use S3 URLs with the understanding that this will be a little weird for data distributed via zenodo (but probably not really cause any issues)
2. Use relative paths for everything but EQR. It's probably not ideal to mix relative / absolute paths, but also there's a certain logic to this option given that EQR will never show up in zenodo so this will be technically correct no matter where the `datapackage.json` exists
3. Don't include the EQR tables in the main `datapackage.json` and generate an entirely separate json file that exists gets distributed in `s3://pudl.catalyst.coop/ferceqr/`

I'm actually gravitating towards option 3 because it feels most correct, and also the EQR tables are produced at a different frequency from the rest of PUDL, so it makes sense that the metadata should also be updated at that frequency.

jdangerx: Option 3 is definitely the cleanest - if we aren't wedded to EQR showing up within the PUDL datapackage for any other reason we should do that. I think it makes my downstream work easier too.

src/pudl/metadata/classes.py
jdangerx: I think it's much more direct to have something like

```python
# resources/ferceqr.py

PARTITIONS = DataSource.from_id["ferceqr").working_partitions
...
"paths": [f"core_ferceqr__foo/{p}.parquet" for p in PARTITIONS],
...
```

or splitting into `path`/`extrapaths`, which would make the conversion code in `Resource` stupid simple:

```python
# resources/ferceqr.py

PARTITIONS = DataSource.from_id["ferceqr").working_partitions
...
"path": f"core_ferceqr__foo/{PARTITIONS[0]}.parquet",
"extrapaths": [f"core_ferceqr__foo/{p}.parquet" for p in PARTITIONS[1:]],
...
```


Rather than having to remember what a `PudlPartitionDescriptor` is, and then that it gets munged in this way within the big honkin `Resource` class.

src/pudl/metadata/classes.py
jdangerx: Ugh, that's frustrating - I was just noticing that while playing around with the `Resource` class. Time to rewrite `frictionless` in Rust 🦀 

I think, since we're the ones using the `datapackage.json`, and since only EQR has multiple paths, we should:

* use what `frictionless` gives us, so that we can unblock our EQR work
* go make a PR so that `frictionless` handles multiple paths properly, as the spec dictates -_-

src/pudl/metadata/classes.py
zschira: I've gone this route, however we do have to import the big `SOURCES` dict because importing `DataSource` from `classes.py` would cause a circular import. Not too big of a deal though.

src/pudl/metadata/classes.py
zschira: This seems like a reasonable path, so I've gone ahead and switched back to using `frictionless` directly. I closed my issue in the `frictionless` repo, but maybe I should just reopen it and refocus towards outputting a single `path` field.

src/pudl/metadata/classes.py
zschira: I've now separated out the main PUDL and EQR datapackages. This was a bit awkward to do because we use the big `PUDL_PACKAGE` object in a few places where we typically still want EQR resources to be included (like docs generation), so I added parameters to the `to_frictionless` method so we can specifically filter out resources when we actually generate the JSON files. Definitely open to suggestion if you have other ideas on how to handle things, but this seemed a good way to not have to touch a bunch of other stuff along with this change.

zschira:
>Finally, this behavior should be pretty straightforward to unit test by making a Package with a single-path resource and a multi-path resource and verifying that the output looks like a valid datapackage.

I've added some unit tests for the multiple vs single path behavior as well as testing the split datapackage output stuff.

Also with the new datapackage setup we'll probably want to re-run the EQR batch job once this merges to make the EQR specific datapackage available to eel-hole (we could manually upload it, but it's probably worth validating the full pipeline).

src/pudl/etl/ferceqr_deployment.py
```diff
@@ -121,19 +135,22 @@ def deploy_ferceqr():
...
+                # UPath's don't work well with requester pays buckets, so use the fsspec
+                # filesystem directly
+                fs.mkdirs(path=table_remote_path, exist_ok=True)
+                fs.put(
+                    f"{ParquetData(table_name=table).parquet_directory}/",
+                    f"{table_remote_path}/",
+                    recursive=True,
+                )
+            # Copy datapackage to distribution path
+            fs.put_file(datapackage_path, f"{distribution_path}/")
```
jdangerx: non-blocking: would [`fs.pipe`](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.pipe) make it so you don't have to write the datapackage to disk and can instead keep it as a string in memory?

src/pudl/metadata/classes.py
```diff
@@ -2348,11 +2362,24 @@ def encode(self, df: pd.DataFrame, copy: bool = True) -> pd.DataFrame:
                 )
         return encoded_df
 
-    def to_frictionless(self) -> frictionless.Package:
+    def to_frictionless(
+        self,
+        exclude_pattern: Pattern[str] | None = None,
+        include_pattern: Pattern[str] | None = None,
```
jdangerx, line 2363: Handy! Let's document what these are for, and what happens when you pass both/neither. Also that it's `match` not `search` (was just lightly confused by this when trying to make a FERC EQR datapackage for testing locally)

src/pudl/metadata/classes.py
```diff
@@ -1437,14 +1440,16 @@ class Resource(PudlMeta):
...
-    path: String | None = None
+    path: String | list[String]
     dialect: dict[str, str] | None = None
     profile: String = "tabular-data-resource"
     contributors: list[Contributor] = []
     licenses: list[License] = []
     sources: list[DataSource] = []
     keywords: list[String] = []
     encoder: Encoder | None = None
+    path: str = pydantic.Field(default_factory=lambda data: f"{data['name']}.parquet")
```
jdangerx, line 1450: `path` is double-defined here - can probably get rid of the earlier one!

src/pudl/metadata/classes.py
```diff
@@ -1638,6 +1643,10 @@ def dict_from_resource_descriptor(  # noqa: C901
                         f["encoder"] = encoder
                         break
 
+        # If path is None, completely remove it so Resource class uses default factory
+        if obj["path"] is None:
```
jdangerx: non-blocking: seems like you could call [`obj = descriptor.model_dump(by_alias=True, exclude_none=True)`](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump) and avoid having to manually exclude this `None` value.

test/unit/metadata_test.py
```diff
@@ -255,6 +258,61 @@ def test_merge_descriptions():
...
+                dummy_resource_dict
+                | {
+                    "path": paths[0],
+                    "extrapaths": paths[1:],
+                }
+            ),
+        )
+    )
+    assert multiple_path_resource.path == paths[0]
+    assert set(multiple_path_resource.extrapaths) == set(paths[1:])
```
jdangerx: Sweet, a straightforward and readable test!

non-blocking: beyond asserting that the `Resource` forms correctly, do we want to make sure that we can re-read a `Resource` from the json and get:
 
* first path as `.path`
* other paths as `.extrapaths`
* all paths as `.normpaths`

This is probably something that `frictionless-py` should test on their own... but, you know. Maybe it's worth testing to see that this somewhat undocumented behavior remains intact.

src/pudl/etl/ferceqr_deployment.py
zschira: Seems like it! I've switched to using `pipe` now.

src/pudl/metadata/classes.py
zschira: Good call, added descriptions of these params to the docstring.

src/pudl/metadata/classes.py
zschira: Oh yeah, whoops. Kinda surprised `pydantic` lets you double define a field like this.

src/pudl/metadata/classes.py
zschira: Ah, much nicer!

test/unit/metadata_test.py
zschira: Makes sense test the frictionless outputs. I've now switched this test to go from `Resource` -> `frictionless.Resource` -> `json` -> `frictionless.Resource` and checking paths on the `frictionless.Resource`

PR #4963, opened 2026-01-20
Sign PUDL user agent for datastore requests

### LINKED ISSUE #4961: Add custom user header to PUDL datastore requests
### Overview

According to Zenodo, they're currently blocking generic Python request headers:

> We are temporarily blocking requests with the default Python requests header due to excessive load
> 
> Our recommendation for programmatic access of Zenodo's REST APIs is to set a custom User-Agent header that identifies your client (e.g., "<my tool>/<tool version> (https://github.com/my/tool)"). This helps us distinguish legitimate tools from abusive traffic and allows us to identify known clients, reducing the likelihood of your requests being affected by rate-limiting measures in the future.
> 
> In your code that would be something like:
> 
> ```
> headers = {
>    "Authorization": "Bearer " + os.environ.get("ZENODO_API_KEY"),
>    "User-Agent": "<my tool>/1.4.2 (https://github.com/my/tool)",
> }
> ```

### Success Criteria

* [ ] Successfully use pudl_datastore to download data locally

### Next steps
* [ ] Add a custom user-agent header to our pudl_datastore query
* [ ] Test it out

e-belfer:
# Overview

Closes #4961 

Zenodo is rejecting all Python requests without a custom user agent, causing test failures and making the datastore unusable at the present.

## What did you change?
Added a `User-Agent` header to the datastore session.

# Testing

Run `pudl_datastore` on any dataset that isn't cached locally.

## To-do list
- [ ] Figure out how to dynamically update PUDL version? (optional)

### DIFF
--- a/docs/dev/nightly_data_builds.rst
+++ b/docs/dev/nightly_data_builds.rst
@@ -195,6 +195,16 @@ the project to be used for requester pays access through applications:
 
   gcloud auth application-default login
 
+.. tip::
+
+   If you've done all this and you are still getting "ERROR: (gcloud.storage.hash)
+   HTTPError 400: Bucket is a requester pays bucket but no user project provided."
+   errors below, try:
+
+   .. code::
+
+      gcloud config set billing/quota_project catalyst-cooperative-pudl
+
 To test whether your GCP account is set up correctly and authenticated you can run the
 following command to list the contents of the cloud storage bucket containing the PUDL
 data. This doesn't actually download any data, but will show you the versions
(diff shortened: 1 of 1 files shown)


src/pudl/workspace/datastore.py
```diff
@@ -271,6 +271,13 @@ def __init__(
         self.http = requests.Session()
         self.http.mount("http://", adapter)
         self.http.mount("https://", adapter)
+        # As of 01/2026, Zenodo rejects all Python requests with no custom user agent.
+        # We add this to note where our project's requests are originating from.
+        self.http.headers.update(
+            {
+                "User-Agent": "pudl/v2026.1.0 (https://github.com/catalyst-cooperative/pudl)"
+            }
+        )  # TODO: Better way to update version?
```
e-belfer: Curious if there's a way to simply update the version here, possibly by referencing citation.cff? But I think this is fine as is if not.

zschira: Since this repo depends on PUDL, could we just get the version from the package? Like:

```py
"User-Agent": f"pudl/{pudl.__version__} (https://github.com/catalyst-cooperative/pudl)"
```

zschira (approved):
Just one question about using the PUDL version from the package itself, but otherwise good to go!

PR #4967, opened 2026-01-21
ETL the FERC Company Identifier (CID) table

### LINKED ISSUE #4851: Add FERC company ID table to PUDL (and EQR?) Jobs
# Overview

Once we have a normal archive of the FERC CIDs (#4832) we need to add that table as an asset to both the PUDL ETL and the FERC EQR Dagster jobs. 

## Success Criteria

- We have a basic transform that creates a FERC CID table based on the most recently archived data.
- This table is available both in the PUDL ETL and the FERC EQR ETL

## Notes

- If this table is part of PUDL, we might just be able to assume that the PUDL version is always more up to date than the EQR version, and rather than processing it in both places, just use the most recent nightly version of it for anything we need.
- In the 2nd phase I think we're going to want this to be a changelog table so we can capture past company names, IDs, etc.

katie-lamb:
# Overview

Closes #4851 .

Extracts and transforms the FERC Company Identifier table.

## What did you change?

Extracted the FERC CID table using a simple custom Extractor that has partitions on "data_set". I only extract the data table itself, not the data dictionary.

## Notes from Transform
* the CID column is almost a primary key for the table except that CID = C003521 is duplicated. Both Chestnut Flats Wind Lessee, LLC and PivotGen are assigned to CID = C00351. After reading [this article](https://www.windpowerengineering.com/pivotgen-to-revitalize-38-mw-wind-project-in-pennsylvania/) about the Chestnut Flats Wind project, I elected to drop Chestnut Flats Wind Lessee, LLC in order to enforce CID as the primary key.
  * However, it's not great to drop an org name, would it be better to not have a primary key in this table?
* From Ella's poking: "261/282 FERC IDs match to the XBRL data, most of the baddies have IDs that start with an R"

## To Do

- [x] create an alembic revision
- [x] test writing core table to DB
- [x] finish resource metadata description
- [x] add more dbt validation tests
- [ ] check viability of FK constraint with other tables

<br class="Apple-interchange-newline">

<br class="Apple-interchange-newline">

## Documentation

# Testing

## To-do list

### DIFF
--- a/dbt/models/ferc/core_ferc__entity_companies/schema.yml
+++ b/dbt/models/ferc/core_ferc__entity_companies/schema.yml
@@ -9,6 +9,10 @@ sources:
           - name: company_id_ferc
             data_tests:
               - unique
+              - dbt_expectations.expect_column_values_to_match_regex:
+                  arguments:
+                    regex: "^C\\d{6}$"
+                    is_raw: True
           - name: company_name
           - name: program
           - name: company_website
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -23,6 +23,15 @@ RUS-12
 
 * Extracted data for twelve USDA RUS tables. See :issue:`4900` and PR :pr:`4916`.
 
+FERC EQR
+~~~~~~~~
+
+* Added the company identifiers (CID) table from EQR. See :issue:`4851` and
+  :pr:`4967`. Also, note that the actual FERC EQR data is available on `PUDL
+  Viewer <https://data.catalyst.coop/search?q=ferceqr>`__ as well as `on S3 for
+  direct download
+  <https://catalystcoop-pudl.readthedocs.io/en/nightly/data_access.html#ferc-eqr-experimental>`__
+
 Expanded Data Coverage
 ^^^^^^^^^^^^^^^^^^^^^^
 


src/pudl/package_data/settings/zenodo_dois.yml
```diff
@@ -18,6 +18,7 @@ ferc60: 10.5281/zenodo.17223513
 ferc6: 10.5281/zenodo.17119798
 ferc714: 10.5281/zenodo.16676145
 ferceqr: 10.5281/zenodo.18251902
+ferccid: 10.5281/zenodo.18176680
```
katie-lamb: this should be the production DOI? same as what's in [the pudl_archivers production DOIs?](https://github.com/catalyst-cooperative/pudl-archiver/blob/main/src/pudl_archiver/package_data/zenodo_doi.yaml)

e-belfer: No you want the DOI for the specific archive, whereas the other document tracks concept DOIs.

src/pudl/extract/ferccid.py
```diff
@@ -0,0 +1,63 @@
...
+from io import BytesIO
+
+import pandas as pd
+from dagster import Output, asset
+
+from pudl.extract.extractor import GenericExtractor, GenericMetadata
+from pudl.metadata.classes import DataSource
+
+
+class Extractor(GenericExtractor):
```
katie-lamb: Since the dataset had no years for working partitions and is just one table, I couldn't use the `CsvExtractor` class and made my own extractor that has pretty simple functionality.

src/pudl/extract/ferccid.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self, *args, **kwargs):
+        """Initialize the module.
+
+        Args:
+            ds (:class:datastore.Datastore): Initialized datastore.
+        """
+        self.METADATA = GenericMetadata("ferccid")
+        super().__init__(*args, **kwargs)
+
+    def source_filename(self):  # noqa: D102
```
katie-lamb: `source_filename` has to be defined to subclass `GenericExtractor`.  It could be cleaner to subclass the `CsvExtractor` and then override `load_source` and `extract` methods, but I thought this was clearer even though `source_filename` doesn't do anything.

src/pudl/metadata/fields.py
```diff
@@ -1147,6 +1147,10 @@
         ),
         "unit": "MW",
     },
+    "company_id_ferccid": {
```
katie-lamb: Maybe this should be just `cid` or `company_id`? Tried to keep with the convention of `utility_id_eia` but maybe it's confusing have CID (company ID) be redundant

src/pudl/metadata/fields.py
```diff
@@ -6005,6 +6016,20 @@
...
+        "description": "The specific program or initiative associated with the FERC organization, which can provide context for the company's activities.",
+        "constraints": {
+            "enum": [
+                "FPA (Market Based Rate) Public Utilities",
+                "FPA (Traditional Cost of Service and Market Based Rates) Public Utilities",
+                "ICA Oil Pipelines",
+                "NGA Gas Pipelines",
+                "NGPA 311 and NGA Hinshaw Gas Pipelines",
+                "Power Administrations",
+            ]
```
katie-lamb, line 6023: This is the correct way to make a column "categorical"? I guess I'm not strictly sure if these are the only categories that `program` can be, but they're the only ones that appear right now

src/pudl/transform/ferccid.py
```diff
@@ -0,0 +1,54 @@
...
+    col = (
+        col.str.replace(r"[\x00-\x1f\x7f-\x9f]", "", regex=True)
+        .str.strip()
+        .str.replace(r"\s+", " ", regex=True)
+        .replace(to_replace=r"^(?i:\s*|na|nan|none)$", value=pd.NA, regex=True)
+    )
+    return col
+
+
+@asset(  # io_manager_key="pudl_io_manager"
```
katie-lamb: uncomment this when I'm ready to make an alembic revision

src/pudl/transform/ferccid.py
```diff
@@ -0,0 +1,54 @@
...
+)
+def core_ferccid__data(raw_ferccid__data):
```
katie-lamb: how does this name seem?

src/pudl/transform/ferccid.py
```diff
@@ -0,0 +1,54 @@
...
+    # the CID values we know have duplicates that should be removed.
+    known_dupe = cid_df.loc[
+        (cid_df["company_id_ferccid"] == "C003521")
+        & (cid_df["company_name"] == "Chestnut Flats Wind Lessee, LLC")
+    ]
+    cid_df = cid_df.drop(known_dupe.index)
+    if not cid_df["company_id_ferccid"].is_unique:
+        raise ValueError(
+            f"Duplicate CIDs found in core_ferccid_data: {cid_df[cid_df['cid'].duplicated(keep=False)]}"
+        )
```
katie-lamb: maybe it's enough to just have a "unique" test on the column in the schema validation? Or is it helpful enough to just throw this ValueError here and log duplicates?

src/pudl/metadata/resources/ferccid.py
```diff
@@ -0,0 +1,29 @@
+"""Table definitions for the FERC Company Identifier table."""
+
+from typing import Any
+
+RESOURCE_METADATA: dict[str, dict[str, Any]] = {
+    "core_ferccid__data": {
+        # TODO: fill in "description":
```
katie-lamb: was a little confused about how the structure of this description key works now

src/pudl/metadata/resources/ferccid.py
```diff
@@ -0,0 +1,29 @@
...
+                "state",
+                "zip_code",
+                "zip_4",
+            ],
+            "primary_key": [
+                "company_id_ferccid",
+            ],
+        },
+        "sources": ["ferccid"],
+        "etl_group": ["ferccid", "ferceqr"],
```
katie-lamb: does it work to just throw ferceqr here too? should it be part of that etl group? or make it part of the EQR job later?

dbt/models/ferccid/core_ferccid__data/schema.yml
```diff
@@ -0,0 +1,20 @@
+version: 2
+sources:
+  - name: pudl
+    tables:
+      - name: core_ferccid__data
+        data_tests:
+          - expect_columns_not_all_null
+        columns:
+          - name: company_identifier_ferccid
```
katie-lamb: Is there any easy to enforce that it's a 7 character ID that starts with C?

katie-lamb:
Hey @zschira I left some comments if you want to get started with the review. I'm also done with the extract, but haven't finished/tested the transformation. I just mainly just need to make an alembic revision and test writing it to the DB.

src/pudl/metadata/fields.py
e-belfer: I think `company_id_ferc` would be clearest and most consistent with our other naming conventions.

src/pudl/metadata/resources/ferccid.py
e-belfer: Because this is not changing over time (seeing no date column), the table name should probably be something like `core_ferccid__entity_companies`.

I tried to clarify the new process [here](https://catalystcoop-pudl.readthedocs.io/en/latest/dev/metadata.html) but am happy to answer any questions you have about the table metadata.

src/pudl/metadata/fields.py
e-belfer: Is company here always a utility, or could it be a variety of entities? Just wondering if company is the most descriptive common name we have for all these reporters.

src/pudl/metadata/fields.py
e-belfer: The best way to do this would be to make a list in the `enums.py` and refer to it here. We generally prefer to normalize the string text, but if you're finding it's really clean maybe it's ok to leave it as is. See `model_case_eiaaeo` for an example.

src/pudl/transform/ferccid.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@asset(  # io_manager_key="pudl_io_manager"
+)
+def core_ferccid__data(raw_ferccid__data):
+    """Clean the FERC Company Identifier table."""
+    cid_df = raw_ferccid__data.copy()
+    cid_df.columns = cid_df.columns.str.lower()
+    # override with field descriptions from data dictionary
+    cid_df = cid_df.rename(
```
e-belfer, line 28: If these are the desired column names, you can just make them the header column names in the extraction CSV and they'll get renamed during that step.

src/pudl/transform/ferccid.py
e-belfer: If this is the PK of the table it will definitely expect uniqueness as part of the schema checks. This seems like a useful early warning though, and I don't see why not keep it in.

src/pudl/extract/ferccid.py
```diff
@@ -3,60 +3,21 @@
...
 
 import pandas as pd
-from dagster import Output, asset
+from dagster import asset
 
-from pudl.extract.extractor import GenericExtractor, GenericMetadata
 from pudl.metadata.classes import DataSource
 
 
-class Extractor(GenericExtractor):
```
zschira: I felt like using a full `Extractor` class for this one super simple table was a bit overkill and added unnecessary complexity, so I've gone to just a simple extract function.

src/pudl/metadata/resources/ferccid.py
```diff
@@ -3,13 +3,13 @@
 from typing import Any
 
 RESOURCE_METADATA: dict[str, dict[str, Any]] = {
-    "core_ferccid__data": {
+    "core_ferc__entity_companies": {
```
zschira, line 6: I was a bit uncertain about what to use as the data source in the table name, since we do have a `ferccid` data source defined, but that feels more confusing to me than just calling it `core_ferc` as I've done.

src/pudl/metadata/resources/ferccid.py
```diff
@@ -3,13 +3,13 @@
...
         "description": {
             "additional_summary_text": "the companies that submit required filings to FERC."
         },
         "schema": {
             "fields": [
-                "company_id_ferccid",
+                "company_id_ferc",
```
zschira, line 12: I think we should reuse the `company_id_ferc` field defined for EQR. I'm also pretty sure that `utility_id_ferc1_xbrl` is a CID, so we might want to change that column name as well.

src/pudl/metadata/fields.py
zschira: We do have other cases on defining enums in line like this and I kind of like having one fewer file to look at when it's just a handful of categories like this.

zschira (commented):
Changed some naming decisions and simplified the extract.

src/pudl/extract/ferccid.py
katie-lamb: The only other functionality I initially wanted to grab from the full `Extractor` class was checking when/if the column names changed (using the `column_map`). Does that feel important to do in the extract? Or can it just be caught/handled in the transform?

src/pudl/metadata/resources/ferccid.py
katie-lamb: Oh yep, didn't see `company_id_ferc` from EQR. I thought about renaming the `utility_id_ferc1_xbrl` column, but I think there are some values in that column which don't show up in the CID table. The missing values start with "R", so they either might be erroneously starting with "R" and should start with "C", or they're actually not in the CID table.

 Ideally we'd have an FK constraint on that column, if we can track down why there are IDs that aren't in the CID table.

katie-lamb: Also, with the appearance of some IDs that start with "R" in other tables, it might be good to enforce that the CID's in this table are 7 character strings that start with "C". Is there a good place to do that or should it just go in the transform?

src/pudl/transform/ferccid.py
```diff
@@ -0,0 +1,51 @@
...
+        col.str.replace(r"[\x00-\x1f\x7f-\x9f]", "", regex=True)
+        .str.strip()
+        .str.replace(r"\s+", " ", regex=True)
+        .replace(to_replace=r"^(?i:\s*|na|nan|none)$", value=pd.NA, regex=True)
+    )
+    return col
+
+
+@asset(io_manager_key="pudl_io_manager")
+def core_ferc__entity_companies(
```
katie-lamb, line 21: Assert somewhere in here that the CIDs are 7 character strings that start with C?

src/pudl/metadata/resources/ferccid.py
jdangerx: This could live in the dbt schema:

```diff
diff --git a/dbt/models/ferc/core_ferc__entity_companies/schema.yml b/dbt/models/ferc/core_ferc__entity_companies/schema.yml
index 8c3c12ac4..54a075676 100644
--- a/dbt/models/ferc/core_ferc__entity_companies/schema.yml
+++ b/dbt/models/ferc/core_ferc__entity_companies/schema.yml
@@ -9,6 +9,10 @@ sources:
           - name: company_id_ferc
             data_tests:
               - unique
+              - dbt_expectations.expect_column_values_to_match_regex:
+                  arguments:
+                    regex: "^C\\d{6}$"
+                    is_raw: True
           - name: company_name
           - name: program
           - name: company_website

```

zaneselvans: This column appears not only this table, but in all of the EQR tables, so it would be good to set the constraint in a non-table-specific place (the column constraints). The CIDs are also allowed to start with `D` for delegated filers (which we see in the EQR data) so we should update this to be more general.

zaneselvans:
Since one of the purposes of curating this table is ultimately to have a record of all the different values that have ever been associated with a given CID, I think we probably should not be dropping the record with the duplicate CIDs.  We can create a reconciled `out_ferc__changelog_companies` or something down the road that has more structure, but the `core_` table seems like it should contain all the data.

I think we also may be clobbering a bunch of Canaidan postal codes by enforcing inappropriate ZIP code regexes on them. I created #5019.

Issue #4883, opened 2025-12-31
Integrate USDA RUS Form 7: Distribution Top 10 Tables


aesharpe:
Top ten tables (in no particular order) as determined by stakeholders: 

- [x] **A - OpRpt_DStatementOfOperations.csv**: Operational and maintenance expenses, total cost of electric service, patronage capital at the utility level.
- [x] **C - OpRpt_DBalanceSheet.csv**: Utility assets, debts, liabilities, and credits
- [ ] **US_Distribution_Borrower.csv**: List of all borrowers
- [x] **H - OpRpt_DEmployeeHourAndPayrollStatistics.csv**: – Number of employees, hours worked, and payroll
- [x] **I - OpRpt_DPatronageCapital.csv**: Capital credits distributed and received
- [x] **M - OpRpt_DAnnualMeetingAndBoardData.csv**: date of annual meeting, number of members, number of board members, members present, members voting by mail/proxy, quorum, whether manager has a written contract, fees and expenses for board meetings.
- [x] **N - OpRpt_DLongTermDebts.csv**: existing debts from other institutions
- [x] **O - OpRpt_DPowerReqDb.csv**: power requirement database including kWh sold, revenue, and number of customers by residential sales, irrigation, commercial and industrial (above and below a threshold), street lighting, sales for resale, etc. Also includes total kWh purchased, generated, used, transmission revenue, and other revenue, cost of purchases and generation, interchange kWh-Net and peak sum.
- [x] **P -OpRpt_DEnergyEfficiencyProgramsSummary.csv**: number of customers, amount invested, MMBTU savings, for different efficiency programs added this year and to date.
- [x] **qI - OpRpt_DDistributionInvestments.csv**: Utility investments in rural development such as non-utility property, associated orgs, economic development projects, other, special funds, cash, special deposits, etc.

Issue #4887, opened 2025-12-31
Validate USDA RUS Form 7: Distribution Tables


aesharpe:
### `core_rus7__yearly_employee_statistics` ✔️ 
- [x] all employee and employee columns >0
- [x] employee hours worked should not be greater than employees * regular time schedule
- [x] added some expected change consistency

### `core_rus7__yearly_meeting_and_board` ✔️ 
- [x] report date > last annual meeting date
- [x] members present + voting < total members
- [x] numeric columns are positive

### `core_rus7__yearly_energy_efficiency` 🟡 
- [x] customer count is positive where cumulative
- [x] invested value is positive
- [x] savings is positive
- [ ] *TODO*: savings is not unreasonable - more complicated, still need to think about this - see below.

Todos:
Investigate probably wrong units for savings_mmbtu. there are 20 records that are reported by borrowers in years where the savings is reported in an int larger than the Int64 type... which really seems like a problem of scale.
```python
# Found bc of this error:
# pyarrow.lib.ArrowInvalid: ('Integer value 4989470342 not in range: -2147483648 to 2147483647', 'Conversion failed for column savings_mmbtu with type Int64')
df = defs.load_asset_value("core_rus7__yearly_energy_efficiency")
idx = ["borrower_id_rus","report_date"]
df = df.set_index(idx)
too_big_borrower_years = df[df.savings_mmbtu > 2147483647].sort_values(by="savings_mmbtu").index.unique()
df.loc[too_big_borrower_years].sort_index()
```

### `core_rus7__yearly_investments` ✔️ 
- Both included and excluded investments and the income_or_loss column often contain negative values. This seems sensible, and is not a validation problem. No additional validations added here.

### `core_rus7__yearly_long_term_debt` ✔️

- [x] Expect values are positive
- [x] Expect principal + interest sum to current year total

### `core_rus7__yearly_patronage_capital` ✔️

- [x] Numeric values are positive
- [x] Patronage current year > patronage cumulative
- Validate totals are sum of components -- skipped for now

### `core_rus7__yearly_power_requirements` ✔️

- [x] Numeric values are positive
- [x] Converted `is_coincident_peak` to boolean column from 0/1

### `core_rus7__scd_borrowers` 🟡 
- [x] Are all states actual states? - Yes, but also include territories,  "MH" (Marshall Islands), and "FM" (micronesia). Updated field description for source and added enum.
[...]

PR #4971, opened 2026-01-23
Transform/Publish 4 more RUS7 tables

cmgosnell:
# Overview

Working on #4883 

I decided to pause on the scd borrowers table..  because this one could end up having a bunch of FK relationships to the rest of the rus7 tables and i'd rather sort that out once the other first ten tables are in.

## Documentation

# Testing

## To-do list

- [x] Run `pixi run pre-commit-run` to run linters and static code analysis checks.
- [x] Run `pixi run pytest-ci` locally to ensure that the merge queue will accept your PR.

### DIFF
--- a/migrations/versions/9c66ca6fcafa_add_moar_cool_rus7_tables.py
+++ b/migrations/versions/9c66ca6fcafa_add_moar_cool_rus7_tables.py
@@ -1,7 +1,7 @@
 """Add moar cool rus7 tables
 
 Revision ID: 9c66ca6fcafa
-Revises: dcbe4c9e66d4
+Revises: fdea3b18cdb5
 Create Date: 2026-01-23 12:16:56.167671
 
 """
@@ -11,7 +11,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = '9c66ca6fcafa'
-down_revision = 'dcbe4c9e66d4'
+down_revision = 'fdea3b18cdb5'
 branch_labels = None
 depends_on = None
 


src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+    },
+    "savings_mmbtu": {
+        "type": "number",
+        "description": (
+            "The estimated amount of energy savings from energy efficiency programs. "
+            "Warning: We found values much larger than expected that we have not yet "
+            "cleaned - this is likely a reporting unit error."
+        ),
+        "unit": "MMBtu",
+    },
```
cmgosnell, line 9269: this is the column which i found the too-big int's in. I added this into https://github.com/catalyst-cooperative/pudl/issues/4887

src/pudl/package_data/rus7/column_maps/power_requirements.csv
```diff
@@ -1,4 +1,4 @@
...
...l_and_industrial_small_customers_num_december,commercial_and_industrial_small_customers_num_avg,commercial_and_industrial_small_sales_kwh,commercial_and_industrial_small_revenue,commercial_and_industrial_large_customers_num_december,commercial_and_industrial_large_customers_num_avg,commercial_and_industrial_large_sales_kwh,commercial_and_industrial_large_revenue,public_street_lighting_customers_num_december,public_street_lighting_customers_num_avg,public_street_lighting_sales_kwh,public_street_lighting_revenue,public_other_customers_num_december,public_other_customers_num_avg,public_other_sales_kwh,public_other_revenue,sales_for_resale_rus_borrowers_customers_num_december,sales_for_resale_rus_borrowers_customers_num_avg,sales_for_resale_rus_borrowers_sales_kwh,sales_for_resale_rus_borrowers_revenue,sales_for_resale_other_customers_num_december,sales_for_resale_other_customers_num_avg,sales_for_resale_other_sales_kwh,sales_for_resale_other_revenue,total_customers_num_december,total_customers_num_avg,total_sales_kwh,total_revenue,transmission_revenue,other_electric_revenue,own_use_kwh,purchased_kwh,generated_kwh,purchases_and_generation_cost,interchange_kwh,peak_kw,is_peak_coincident
```
cmgosnell, line 1: just a few small renames to get nicer column names and categorical values out of this

src/pudl/transform/rus7.py
```diff
@@ -96,8 +96,8 @@ def core_rus7__yearly_energy_efficiency(raw_rus7__energy_efficiency):
         idx_ish=["report_date", "borrower_id_rus", "borrower_name_rus"],
         data_cols=data_cols,
         pattern=rf"^({'|'.join(data_cols)})_(.+)_(cumulative|new_in_report_year)$",
-        match_names=["data_cols", "customer_classification", "date_range"],
-        unstack_level=["customer_classification", "date_range"],
+        match_names=["data_cols", "customer_class", "date_range"],
+        unstack_level=["customer_class", "date_range"],
```
cmgosnell: i decided to rename `customer_classification` to `customer_class` because 861 has a `customer_class` name. while they have different categories its effectively the same thing.

src/pudl/transform/rus.py
```diff
@@ -104,3 +104,35 @@ def multi_index_stack(
     df.columns = df.columns.map("".join)
     df = df.dropna(subset=data_cols, how="all")
     return df
+
+
+def convert_units(
```
cmgosnell, line 109: i know we've done this for other data sources but it's all a little too specific to that application it seemed. so i made a simple converter. mostly to convert kwh -> mwh but is generic enough if you can convert with just a simple multiplier.

src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+    "date_range": {
+        "type": "string",
+        "description": (
+            "The date range that any given record pertains to. Ex: 'december' implies that this record covers the month of "
+            "December only, while 'avg' implies this record pertains to the average of the reporting period."
+        ),
+        "constraints": {
+            "enum": {"avg", "december", "new_in_report_year", "cumulative"}
+        },
+    },
```
cmgosnell: i still don't love this column name... its not awful but any advice would be appreciated.

src/pudl/metadata/resources/rus7.py
```diff
@@ -116,61 +116,33 @@
...
                 "report_date",
                 "borrower_id_rus",
                 "borrower_name_rus",
-                "customer_classification",  # enum w/ things like residential_excluding_seasonal, residential_seasonal	irrigation etc.
-                "date_range",  # ?idk about that name enum w/ new_in_report_year, cumulative
+                "customer_class",
+                "date_range",
                 "customers_num",
                 "invested",
-                "estimated_savings_mmbtu",
```
aesharpe, line 168: why no estimated?

src/pudl/metadata/fields.py
aesharpe: hmm ya this is a weird one. Some other options: 
- observation_period
- timeframe

src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+    "invested": {
+        "type": "integer",
+        "description": "The amount of money invested.",
+        "unit": "USD",
+    },
```
aesharpe, line 9260: I have a field called `total_investment` for the rus12 renewable plants table that seems similar. We should see if they can be combined or not. But there is a subtle difference between investment and invested.

src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+        "description": (
+            "The estimated amount of energy savings from energy efficiency programs. "
+            "Warning: We found values much larger than expected that we have not yet "
+            "cleaned - this is likely a reporting unit error."
+        ),
+        "unit": "MMBtu",
+    },
+    "electric_sales_revenue": {
+        "type": "integer",
+        "description": "Total Revenue Received From Sales of Electric Energy. Total of lines 1c thru 9c on the original form.",
```
aesharpe, line 9272: I like that you wrote the total of the lines, however, it's possible that the line numbers could change over time so I'm not sure if it's worth it to be that specific in the column description. But I do see the desire to highlight it being a total. Instead of writing 1c through 9c it might actually behoove us to just write out the column names themselves, as annoying as that might be, it is more useful for the user.

src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+        "unit": "MWh",
+    },
+    "transmission_revenue": {
+        "type": "integer",
+        "description": "Transmission revenue.",
+        "unit": "USD",
+    },
+    "other_electric_revenue": {
+        "type": "integer",
+        "description": "Other electric revenue - other than electric_sales_revenue.",
```
aesharpe: Maybe just say: 


```suggestion
        "description": "revenue other than electric_sales_revenue.",
```

src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+        "unit": "USD",
+    },
+    "own_use_mwh": {
+        "type": "number",
+        "description": "The electricity in kWh used internally.",
```
aesharpe: If possible to clarify what internally means that would be great. Also change to MWh.


```suggestion
        "description": "The electricity in MWh used internally.",
```

src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+        "unit": "MWh",
+    },
+    "generated_mwh": {
+        "type": "number",
+        "description": "The total electricity generated.",
+        "unit": "MWh",
+    },
```
aesharpe, line 9299: this would probably be more helpful as `electricity_generated_mwh` but that's just a nit

src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+    "purchases_and_generation_cost": {
+        "type": "integer",
+        "description": "The cost of purchases and generation of electricity.",
+        "unit": "UDS",
+    },
+    "interchange_mwh": {
+        "type": "number",
+        "description": "The net interchange of electricity..",
```
aesharpe:
```suggestion
        "description": "The net interchange of electricity.",
```

aesharpe: If there's a way to explain what interchange means that might be helpful

src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+        "unit": "MWh",
+    },
+    "peak_mw": {
+        "type": "number",
+        "description": "The peak system kWh - the sum of all MW.",
+        "unit": "MW",
```
aesharpe, line 9313:
```suggestion
    "peak_mw": {
        "type": "number",
        "description": "The peak system MWh - the sum of all MW.",
        "unit": "MW",
```

src/pudl/transform/rus.py
```diff
@@ -104,3 +104,35 @@ def multi_index_stack(
...
+            dataframe.
+        converter: the float or integer you need to multiply the old values by to
+            convert the units.
+    """
+    convert_cols = df.filter(regex=rf"_{old_unit}$").columns
+    df.loc[:, convert_cols] = df.loc[:, convert_cols].astype("float") * converter
+    df = df.rename(
+        columns={
+            col_name: col_name.removesuffix(f"_{old_unit}") + f"_{new_unit}"
+            for col_name in convert_cols
```
aesharpe, line 139: One nit -- I'd like to use this when converting from `thousand_dollars` to regular USD, however, I'll likely drop the units instead of converting them. I.e., instead of `value_kw` to `value_mw` I'll be going from `value_thousand_dollars` to `value`. If there could be a way to not have a blank `_` at the end of the column when you feed in no new_unit that would be great :)

src/pudl/transform/rus7.py
aesharpe: I like it!

src/pudl/transform/rus7.py
aesharpe: I would elaborate on this doc string to explain that you are splitting up the raw table.

Also, on a more meta note, I feel like these doc strings should say "Transform the raw..." instead of the core table name because that's what they are actually doing.

src/pudl/transform/rus7.py
```diff
@@ -187,11 +194,26 @@ def core_rus7__yearly_power_requirements(
     # the rest of the table pertains to other utility functions. The totals show up
     # in the reshaped electric sales portion of the table but we also rename them
     # and include them here as well.
```
aesharpe: Explain why we are repeating columns, not just that we are.

src/pudl/metadata/fields.py
```diff
@@ -9224,6 +9224,81 @@
...
+    },
+    "is_peak_coincident": {
+        "type": "integer",
+        "description": "Whether or not the peak_kw is coincident or non-coincident peak.",
+        "unit": "boolean",
+    },
```
aesharpe, line 9319:
```suggestion
    "is_peak_coincident": {
        "type": "integer",
        "description": "Whether or not the peak_mw is coincident or non-coincident peak.",
        "unit": "boolean",
    },
```

src/pudl/metadata/resources/rus7.py
cmgosnell: i happen to search "estimated" in fields and noticed how many of the 861 column descriptions say estimated but don't have it in the column name. so i went with that.

src/pudl/metadata/fields.py
cmgosnell: hm ya the investment vs invested feels subtle. i guess i think we should veer towards the side of their verbiage... but i could change this to `investment`. that sounds fine to me! on the `total` bit this one is definitely not a total - it includes some totals in there but those are labeled by from other columns

src/pudl/metadata/fields.py
cmgosnell: i dont disagree with you that id' be nicer but i don't want to spend a ton of time on these things considering the budget

src/pudl/metadata/fields.py
cmgosnell: i like that when looking just at this column but in the context of the whole table i think it'd need to change many of these column names. and also mwh kinda implies this is electricity. so unless you feel strongly about this i'd like to keep it

src/pudl/transform/rus.py
cmgosnell: easy!

src/pudl/transform/rus7.py
cmgosnell: good point i added notes into these three doc strings about splitting the table up!

on the meta note, ya your not wrong - this was really just a shortcut

src/pudl/transform/eia.py
```diff
@@ -171,7 +171,7 @@ def occurrence_consistency(
         col_df["record_occurrences"] / col_df["entity_occurrences"]
     )
     col_df[f"{col}_is_consistent"] = col_df[f"{col}_consistent_rate"] > strictness
-    col_df = col_df.sort_values(f"{col}_consistent_rate")
+    col_df = col_df.sort_values(f"{col}_consistent_rate", ascending=False)
```
cmgosnell: oh this was a mistake to include! will remove shortly

src/pudl/metadata/resources/rus7.py
aesharpe: Hmmm I think it can't hurt to keep it in for clarity. 861 is due for a check-up anyways. But I won't die on this hill or anything.

src/pudl/metadata/fields.py
aesharpe: I also like `amount_invested` or `investment_amount`

src/pudl/metadata/fields.py
aesharpe: I don't feel strongly // I see where you're coming from!

PR #4973, opened 2026-01-26
chore: schedule the EQR batch job

jdangerx:
Just a tiny follow-up to #4878 .

### DIFF
--- a/.github/workflows/build-deploy-ferceqr.yml
+++ b/.github/workflows/build-deploy-ferceqr.yml
@@ -3,7 +3,7 @@ name: build-deploy-ferceqr
 on:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-deploy-ferceqr.yml
```diff
@@ -2,6 +2,8 @@
 name: build-deploy-ferceqr
 on:
   workflow_dispatch:
+  schedule:
+    - cron: "48 0 1 * *" # 00:48 UTC on the first of every month
```
zschira: The EQR archiver currently runs on the 1st of the month as well, so I'm wondering if we want to bump this a couple days so we catch new data as it shows up?

```suggestion
    - cron: "48 0 3 * *" # 00:48 UTC on the third of every month
```

Issue #4913, opened 2026-01-10
Some PUDL utility IDs in FERC do not have corresponding EIA utility IDs


mhnichols16:
### Describe the bug

When merging `out_eia__yearly_utilities` with `out_ferc1__yearly_income_statements_sched114` (where ferc_account == 426.4) using `utility_id_pudl` as the key, about a third of the FERC entities seem not to have a corresponding EIA utility_id. 

### To Reproduce

Here is an R code snippet reproducing what I've found. The two files referenced above, `out_eia__yearly_utilities` and `out_ferc1__yearly_income_statements_sched114`, were saved under those names. 

```r
library(tidyverse)

# Load FERC Form 1 data and limit to ferc_account == 426.4
ferc4264 <- read.csv("out_ferc1__yearly_income_statements_sched114.csv") %>%
  filter(ferc_account == "426.4") %>%
  mutate(utility_id_pudl = as.character(utility_id_pudl)) %>%
  filter(report_year <= 2015)

# Load yearly utilities data, which has both utility_id_pudl and utility_id_eia
pudl_eia_cwalk <- read.csv("out_eia__yearly_utilities.csv") %>%
  select(report_date, utility_id_eia, utility_id_pudl) %>%
  # create year variable
  mutate(report_date = substr(as.character(report_date),1, 10)) %>%
  mutate(report_year = lubridate::year(as.Date(report_date))) %>%
  distinct() %>%
  # limit date range
  filter(report_year <= 2015) %>%
  # drop duplicated utility ID for now
  filter(!(utility_id_pudl == 203 & report_year == 2009)) %>%
  distinct()

# Merge
# This is a many-to-many merge without dropping utility_id_pudl 203
ferc_eia_cwalk <- left_join(ferc4264,
                            pudl_eia_cwalk)

sum(is.na(ferc_eia_cwalk$utility_id_eia)) 
missing <- ferc_eia_cwalk %>% filter(is.na(utility_id_eia))
```

The dataset `missing` has all of the FERC utilities missing a EIA utility ID. 

More specifically, some `utility_id_pudl` identifiers in the FERC data have no corresponding `utility_id_eia`, but very similarly named utilities in `out_eia__yearly_utilities` have a different `utility_id_pudl` and a corresponding`utility_id_eia`. For example, 

[...]

PR #4975, opened 2026-01-26
Update FERC-EIA utility mapping to auto-match utilities with identical names

### LINKED ISSUE #1317: Refactor FERC1-EIA plant and utility ID association process
## Description
Design and implement a new process for linking FERC & EIA plants & utilities for programmatic use which is less labor intensive, produces high-quality associations, and can be generalized to other similar tasks that we foresee needing to tackle.

## Motivation
A big part of PUDL's unique value beyond the original data we pull in is the programmatically usable connections between different datasets. Perhaps chief among them is the linkage between utility and plant level data that's reported to both FERC and EIA. The process we've used since The Beginning involves a lot of tedious, difficult to reproduce, manual work, matching plant names, capacities, fuel types, and collections of owned/operated plants within a gigantic spreadsheet. Updating the FERC-EIA ID mapping has often been a bottleneck in our ability to integrate new years of data, and any change in our methodology requires a lot of work to be re-done.

As we look forward to doing similar work to connect natural gas utilities and their owned facilities in the FERC 2, EIA 176, and PHMSA Pipelines data, and in the integrations between and within other datasets, we want to use more robust, repeatable, (more programmatic processes while maintaining high confidence in the results.

## Some options to consider
* If [our proposed collaboration](https://drive.google.com/file/d/1kdmoYEUTDEbEEmgMvl1M10_YA-U3krKv/view?usp=sharing) with Xu Chu's data lab at Georgia Tech is funded ([through CCAI](https://www.climatechange.ai/calls/innovation_grants) or otherise) we will hopefully be tackling this type of issue through a weakly supervised ML system in 2022. See their [ML for Data Integration page](https://chu-data-lab.cc.gatech.edu/ml-for-data-integration/).
* [Open Refine](https://openrefine.org/) is an open source tool designed to make this kind of semi-automated entity matching somewhat easier, and allows the specified transformations to be exported as JSON, which can be tracked in revision control. (side note: there's [a Datasette plugin](https://github.com/drkane/datasette-reconcile) that exposed an OpenRefine API endpoint).

### Billing
* If the complexity of the refactor is modest (e.g. switching to OpenRefine) this could potentially be done as a project under the Sloan Maintenance & Refactoring category.
* If we are aiming for a more automated, generalized system along the lines of Chu's approaches, this will be a substantial amount of work and will need to be funded internally (R&D tax credits?) or through grants (unless we can find a client with a very long view of things).

### Timing / Dependencies
* It would be good to get this new system up and running **before** we need to integrate a new year of FERC 1/EIA data.
* We could also use a new dataset integration (e.g. FERC 2 + EIA 176) as an opportunity to try a new tool like OpenRefine, rather than adapting the current process, which we know we would like to migrate away from.

e-belfer:
# Overview

Closes #1317 

Right now our FERC-EIA utility map is entirely manual and missing a lot of records that should be joined (see #1317 for characterization). This attempts to add a step into our data updates workflow that auto-matches FERC and EIA utilities.

## What did you change?
Added a new script that auto-matches utilities based on name and updates the mapping spreadsheet for easy manual review of changes. Also add a notes column into the match to enable rapid identification of all automated matches.

Along the way, also fixes a few issues we encountered:
- Corrects the sorting method used in the harvesting. Previously we were sorting and keeping the least consistent value where strictness was zero (plant name, utility name, prime mover code).
- Moved one pre-existing match in the spreadsheet to the same row.

## What is the impact on matching?

In #4913 we characterized the FERC data as follows:

> There are 432 unique utility_id_ferc1 values, and they get mapped to utility_id_pudl and utility_id_eia via [a static mapping file](https://github.com/catalyst-cooperative/pudl/blob/main/src/pudl/package_data/glue/utility_id_pudl.csv). The mapping file does not establish a unique 1:1 mapping between ferc1 and pudl or between ferc1 and eia.
> 
>     All ferc1 ids have at least one pudl id
>     423 ferc1 ids have exactly one pudl id
>     9 ferc1 ids have more than one pudl id (8 have 2; 1 has 3)
>     283 ferc1 ids have no eia id (~67%) - this does seem high
>     148 ferc1 ids have exactly one eia id
>     1 ferc1 id has more than one eia id (it has 3)

With the mapping spreadsheet produced as a result of this process, we see:

- 151 ferc1 ids have no eia id (34.9%)
- 243 ferc1 IDs have one EIA ID (56%)

The remaining unmatched utilities can be found here: 
[unmatched_utilities.csv](https://github.com/user-attachments/files/24894408/unmatched_utilities.csv)

Some of these are clearly junk (e.g., "Test 423 (do not use)"), while others could clearly be handmapped with minimal additional effort - e.g., FERC ID 63 Berea College Utilities and EIA ID 1594 Berea College.

## To-do list
- [x] Add unit tests
- [x] Check how matching method change has impacted matches, if at all (slight improvement, 6 more utilities matched and no matches lost)
- [x] Fix the "this utility has >1 PUDL ID" cases identified in https://github.com/catalyst-cooperative/pudl/issues/4913
- [x] Add a false match dictionary that has EIA and FERC IDs that should *not* be matched, drop matches if they're in this list
- [x] Check for corp to corporation mapping
- [x] Make sure that utility name isn't getting merged into any output tables
- [x] Update developer docs

## Out of scope
- Handle the "new utility" case: automate adding new unmapped utilities to the spreadsheet.
- https://github.com/catalyst-cooperative/pudl/issues/4988
- We could use addresses to improve the precision, but this would require integrating FERC Form 1's identification table, which we don't currently process. The matching module is designed to be relatively easy to swap out for a different method. This will be a follow-up PR.

## Documentation

# Testing

## Testing to-do list


src/pudl/package_data/glue/utility_id_pudl_test_update.csv
e-belfer, line 1: Need to remove from final version, but this is what is currently getting produced when running this script.

src/pudl/scripts/auto_match_utilities.py
```diff
@@ -0,0 +1,280 @@
...
+
+    # Warn about duplicated utilities
+    logger.warning("The following utilities are matched to more than one FERC ID:")
+    logger.info(
+        matched_utilities[matched_utilities.utility_id_ferc1.duplicated(keep=False)]
+    )
+    logger.warning("The following utilities are matched to more than one EIA ID:")
+    logger.info(
+        matched_utilities[matched_utilities.utility_id_eia.duplicated(keep=False)]
+    )
```
cmgosnell: why swap between warning and info level? also total nit but shouldn't these alter you if there are actually duplicates found?

src/pudl/scripts/auto_match_utilities.py
```diff
@@ -0,0 +1,280 @@
...
+
+    # Get the override spreadsheet
+    overrides = get_existing_overrides()
+    ## Get matches from the CSV
+    override_matches = overrides[
+        overrides.utility_id_eia.notnull() & overrides.utility_id_ferc1.notnull()
+    ]
```
cmgosnell, line 307: i'm not sure even if there are any like this in the sheet but it is possible to link more than one EIA utility together under the same pudl id. and same with ferc1 utilities. not sure if we do this ever but it might be good to double check that because this would wipe that out

src/pudl/scripts/auto_match_utilities.py
```diff
@@ -0,0 +1,280 @@
...
+    # Check that we aren't dropping any EIA matches
+    assert drop_records[matching_utility_column].isnull().all()
+    # Check that no other plants are mapped to this ID before we drop it
+    pudl_ids = drop_records.utility_id_pudl
+    assert overrides[
+        (overrides.utility_id_pudl.isin(pudl_ids))
+        & (overrides[matching_utility_column].notnull())
+    ].empty
+    # Drop these unmatched records so we can replace them with matches.
```
e-belfer, line 151: To avoid changing the ID of everything, we just delete these records and re-add them with a new auto-incremented PUDL ID - this means there might be gaps in the PUDL ID counts where these records formerly were.

src/pudl/scripts/auto_match_utilities.py
```diff
@@ -0,0 +1,358 @@
...
+    logger.info(
+        f"Dropping {len(drop_records)} unmatched {entity} records to replace with matches."
+    )
+    return overrides.drop(drop_records.index, axis=0)
+
+
+def add_new_matches_to_spreadsheet(matches: pd.DataFrame, overrides: pd.DataFrame):
```
cmgosnell: total unimportant nit but its a csv not a spreadsheet.

this is also smol (but maybe annoying to update so ignore me if you disagree) imo think it would be more conceptually easier to hold these arg names as `matches_auto` or `matches_new` and `matches_existing`.

src/pudl/scripts/auto_match_utilities.py
```diff
@@ -0,0 +1,358 @@
...
+    a ValueError.
+
+    Once all PUDL utility IDs are assigned, we flag all records where multiple FERC and
+    multiple EIA utilities are matched together. These are uncommon cases and benefit
+    from manual review to ensure that no unexpected connections have been created.
+    """
+    final_updates = []
+    matches["utility_id_pudl"] = pd.NA
+
+    for _i, match in matches.iterrows():
```
cmgosnell: there are not a million of these so looping through this probably wont take forever but ergonomically if there are a lot of complex match errors in one batch it'd be much nicer imo to do this by doing two merges of the `overrides` onto the `matches` (on `utility_id_ferc1` and then one `utility_id_eia`) to check the match-ability.

src/pudl/scripts/auto_match_utilities.py
```diff
@@ -0,0 +1,358 @@
...
+    updates_df = pd.DataFrame(final_updates)
+    updates_df = updates_df.drop(
+        columns=["utility_id_pudl_ferc1", "utility_id_pudl_eia"]
+    )
+    # Whichever records have a blank PUDL ID have no linkage to other existing records
+    # and can be reassigned without impact.
+    # Reassign the blank PUDL IDs by auto-incrementing from the current max record value.
+    # First, assign a group ID for each group of cleaned utility names.
+    updates_df = updates_df.assign(
+        group_id=lambda x: x.groupby("cleaned_utility_name").ngroup()
```
cmgosnell, line 207: your using the `cleaned_utility_name` here as the unique key to auto-incerment by because you know that's the unique value here? that makes sense after i think about it but to me is not intuitive as i mostly know the messy messy names. can you just add to the note above that `cleaned_utility_name` is unique in the updates if true?

src/pudl/scripts/auto_match_utilities.py
```diff
@@ -0,0 +1,358 @@
...
+    )
+    # Then use this value to fill NAs in the utility ID PUDL.
+    updates_df.utility_id_pudl = updates_df.utility_id_pudl.fillna(
+        updates_df.group_id + max(overrides.utility_id_pudl)
+    )
+    updates_df = updates_df.drop(columns=["group_id", "cleaned_utility_name"])
+    # Add a note to these records to help us quickly identify which were matched automatically.s
+    updates_df["notes"] = (
+        f"These records were automatically matched based on utility name on {pd.to_datetime('today').strftime('%m/%d/%Y')}"
```
cmgosnell, line 216: 🎉

src/pudl/scripts/auto_match_utilities.py
```diff
@@ -0,0 +1,358 @@
...
+    )
+    logger.info(
+        f"Adding {len(updates_df)} utility matches to the override spreadsheet. Remember to manually review these."
+    )
+    updated_overrides = pd.concat([overrides, updates_df]).sort_values(
+        "utility_id_pudl"
+    )
```
cmgosnell: putting the `overrides` first here means that we will generally at least start with the same chunk of ids (just thinking of the diffs :-) if yes can you add a lil note here saying that so we don't change it

src/pudl/scripts/auto_match_utilities.py
```diff
@@ -0,0 +1,358 @@
...
+        csv_path = (
+            importlib.resources.files("pudl.package_data.glue")
+            / "utility_id_pudl_test_update.csv"
+        )
+    else:
+        csv_path = (
+            importlib.resources.files("pudl.package_data.glue") / "utility_id_pudl.csv"
+        )
+    logger.info(f"Writing matches to {csv_path}")
+    dataframe.to_csv(csv_path, index=False)
```
cmgosnell, line 255: if its a test run it might be good to spit this out into the `devtools/ferc1-eia-glue` directory along side the unmatached stuff

src/pudl/scripts/auto_match_utilities.py
e-belfer: I was trying to avoid having a lot of yellow screaming text, because this is in fact expected to occur at some level. Perhaps this should all be `logger.info`.

I'm not 100% following the second question - these aren't duplicates necessarily, just cases where an entity might be reporting with more than one EIA ID to a corresponding FERC ID and vice versa.

src/pudl/scripts/auto_match_utilities.py
e-belfer: We definitely do this often, and this code shouldn't wipe this out as it is only touching utilities that are completely unmatched. If there's already an ID associated with either entity, it adopts that ID rather than assigning a new one (which is what is happening in the add_new_matches_to_spreadsheet() function.

src/pudl/scripts/auto_match_utilities.py
e-belfer: Those seem like completely reasonable names, I'll move towards them. Was struggling to find clear variable names here and I like this suggestion.

src/pudl/scripts/auto_match_utilities.py
e-belfer: This only happened once in the entire batch I was working with (which surfaced the sorting problem in the harvesting) so I anticipate it will be a rare error.

src/pudl/scripts/auto_match_utilities.py
e-belfer: Yes, we've joined these matches on the utility name, so utility name becomes the functioning ID linking these records until we assign them a new PUDL ID. I can add a note to clarify this.

src/pudl/scripts/auto_match_utilities.py
e-belfer: Yes, theoretically the only things that should change in the CSV when we run it are deletion of formerly unmatched rows and addition of new matched rows at the bottom. Otherwise we expect nothing to move around, though for the first migration there will be a lot of movement because I sorted the entire CSV by utility ID PUDL to bring matches sprinkled throughout the DF into order, make it easy to auto-increment from the highest ID when adding new utilities and reduce movement in the future.

src/pudl/scripts/auto_match_utilities.py
e-belfer: Great idea, I will update the path.

e-belfer:
> the other thing i want to flag is that there are a small collection of ferc-eia glue things in the glue/ferc1-eia.py and maybe this stuff should live in there? im not married to that at all but it might be confusing to make them so separate.

Because this is a CLI tool, I want to make it installable as a script. That was my logic for putting it in the scripts folder, but I could see an argument for moving the entire script file into the `glue` folder instead. I'm not married to it!

src/pudl/scripts/auto_match_utilities.py
e-belfer: I added the two other IDs to this sort, as there was some movement when there are multiple rows for the same PUDL ID.

e-belfer:
As part of this PR, I resolved the following FERC utilities with more than one PUDL ID:

- Alcoa Power Generating Inc. (FERC PUDL ID 282) - had two PUDL IDs, just a duplicate. Removed one.
- Emera Maine / Versant Power (FERC PUDL ID 158) - a rename, changed both rows to have the same PUDL ID.
- Evergy Metro Inc (Evergy Metro), formerly Kansas City Power & Light Co (FERC PUDL ID 254) - renamed, changed both rows to have the same PUDL ID.
- MDU Resources Group and Montana-Dakota Utilities Co. (FERC PUDL ID 319) - one is a subsidiary of the other and MDU Resources Group is already joined with another Montana-Dakota Utilities Co. record. Changed to have the same PUDL ID.
- South Central MCN LLC is a subsidiary of Gridliance High Plains LLC (FERC PUDL ID 364) - changed to have the same PUDL ID.
- upper michigan energy resources company (pudl determined) and upper michigan energy resources corporation (PUDL FERC ID 368) - confirmed to be permutations of the same entity.
- el investment company and Entergy Louisiana, LLC - [confirmed](https://www.sec.gov/Archives/edgar/data/44570/000006598415000286/a05815.htm) name change, mapped to same PUDL ID.
- Luning energy holdings llc, invenergy investments / liberty utilities (calpeco electric) llc (FERC ID 241) - Confirmed [subsidiary](https://eplanning.blm.gov/public_projects/?doc=2012254%2F200476070%2F20051683%2F250057866%2FLSEP_II_FinalEA_211217_combined_508.pdf). Combined.

There are almost 400 EIA IDs mapped to >1 PUDL ID in the spreadsheet, which we should absolutely resolve but is out of scope for this PR. See https://github.com/catalyst-cooperative/pudl/issues/4988

Note that these duplications don't currently show up in the core assn tables because we are currently dropping duplicate EIA and FERC Form 1 IDs in the `glue()` function, but I've added dbt tests that check for sure that these aren't occurring in those tables.

src/pudl/package_data/glue/utility_id_pudl.csv
e-belfer, line 1: A summary of these changes:

- 17,223 rows unchanged
- 122 rows added - of these, 115 are automatic matches, and the remaining 7 are ID changes done through manual review.
- 221 rows dropped - zero of these records were matched.
- Added a notes column, which complicates standard diffing.

e-belfer:
Summary of per-table changes:

The following plants were matched to more than one PUDL utility ID in the old records, and are now matched to just one. Great!
| utility_id_pudl_x | plant_id_pudl | _merge_x | utility_id_pudl_y | _merge_y |
|---:|---:|---:|---:|---:|
| NaN | 169 | NaN | 164 | right_only |
| NaN | 175 | NaN | 164 | right_only |
| NaN | 188 | NaN | 164 | right_only |
| NaN | 226 | NaN | 164 | right_only |
| NaN | 428 | NaN | 164 | right_only |
| NaN | 459 | NaN | 164 | right_only |
| NaN | 590 | NaN | 21 | right_only |
| NaN | 1175 | NaN | 164 | right_only |
| NaN | 13607 | NaN | 5516 | right_only |

core_pudl__entity_utilities_pudl
A lot of name changes due to the change in sorting. Hand-reviewed, every utility seems to be rematched.

dbt/models/pudl/core_pudl__assn_utilities_plants/schema.yml
```diff
@@ -12,3 +12,10 @@ sources:
         columns:
           - name: utility_id_pudl
           - name: plant_id_pudl
+            tests:
+              - one_value_per_key:
+                  arguments:
+                    other_column_name: utility_id_pudl
+                  description: We expect each plant to only be assigned to one utility, but we still need to deduplicate ID assignment. See issue 4988.
+                  config:
+                    severity: warn
```
cmgosnell, line 21: neat!

e-belfer:
As a result of the [nightly build failure](https://console.cloud.google.com/storage/browser/builds.catalyst.coop/2026-02-04-1839-2038a7302-update-ferc1-eia-utility-mapping), I investigated the following failures:

`dbt_expectations_source_expect_table_row_count_to_be_between_pudl_core_sec10k__assn_exhibit_21_subsidiaries_and_eia_utilities_5564__5543:
`
- Utility name changes result in more matches. Updated the DBT test.

"out_eia923__yearly_generation_fuel_by_generator"
- Changes in here are primarily utility name changes, not value changes.
- The extra 2013 rows are plants which have found units - Cape Canaveral (plant ID 609) and Edwardsport (1004).

docs/release_notes.rst
```diff
@@ -63,6 +63,12 @@ New Data Tests & Validations
 Bug Fixes & Data Cleaning
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
+* We added an automatic script to help match FERC and EIA utilities with near-identical
+  utility names as part of our ongoing data updates. As a result, we have matched an
+  additional 115 utilities and resolved a small handful of cases where a utility was
+  mapped to more than one PUDL ID. See :issue:`1317`, :issue:`4934` and :issue:`4913`,
+  as well as PR :pr:`4975`.
```
e-belfer: To do: add note about fixing sort and impacts on both utility name and prime mover code

PR #4978, opened 2026-01-28
Transform params: permit `categories` key for StringCategories to be a Path to a YAML file

krivard:
# Overview

Add handling so that category data can be stored in YAML and loaded into params objects at construction time.

StringCategories are currently stored inline in pudl.transform.params files.

We are going to need thousands of categorizable items for an upcoming contribution, and it would be significantly more readable to store these category settings in YAML.

This PR makes is possible to specify a Path in the "categories" key for "categorize_strings" transforms.

## What did you change?

* Added a `before` field validator to `StringCategories.categories` to first check for Path values and load YAML data into the proper structure.
* Added unit tests:
  * verify StringCategories objects from inline categories match those from Path categories
  * verify `categorize_strings` output from Path categories meets existing test for inline categories. This needed some light restructuring to make a fixture for ANIMAL_CATS so that parameterization would work nicely.

## Documentation

# Testing

- ran unit tests
- built and inspected docs

## To-do list

- [x] Figure out where to document the YAML format

### DIFF
--- a/src/pudl/transform/classes.py
+++ b/src/pudl/transform/classes.py
@@ -457,7 +457,7 @@ class StringCategories(TransformParams):
                 f
             )
 
-    We recommend putting any YAML files within a dataset directory in `pudl/scr/package_data`.
+    We recommend putting any YAML files within a dataset directory in ``src/pudl/package_data``.
     """
 
     na_category: str = "na_category"


src/pudl/transform/classes.py
```diff
@@ -449,6 +451,18 @@ class StringCategories(TransformParams):
...
 
+    def __init__(self, **kwargs):
+        """Intervene in construction without sacrificing validation."""
+        # deserialize categories from disk if needed
+        if isinstance(kwargs["categories"], Path):
+            with kwargs["categories"].open() as f:
+                data = yaml.safe_load(f)
+                kwargs["categories"] = {
+                    cat: set(values) for cat, values in data.items()
+                }
```
krivard: This assumes a YAML file has been written like this, e.g.:

```
with open("categoryfile.yml", "w") as f:
    yaml.dump(
        {
            cat: sorted(val)
            for cat, val in SUPPORTING_STRUCTURE_TYPE_CATEGORIES["categories"].items()
        }, 
        f
    )
```

We should document that somewhere probably. Where is good?

cmgosnell (commented):
I think this is going in a great direction!! I like this here in the original transformer. I th

instead of overwriting init, you can use a pydantic validator instead. You can make model validators which run [before validation](https://docs.pydantic.dev/latest/concepts/validators/#field-before-validator), so it could do this get from yml situation and

another high level musing/very much not blocking is that it _could_ be nice to have one big yaml file where we could compile table/column based categories and the transformer param passed in could be a bool which then goes an opens one file and grabs the table specific cram.

src/pudl/transform/classes.py
cmgosnell: I think the place to document this would be in the argument docs for `StringCategories` - mostly bc if this is not structured accordingly the user will get a failure probably within this `StringCategories` and hopefully look for guidance here.

docs/release_notes.rst
```diff
@@ -93,6 +93,10 @@ Quality of Life Improvements
   package is stored on GCS, as is the case with FERC EQR. See the
   `FERC EQR archive on Zenodo <https://doi.org/10.5281/zenodo.18251901>`__ as an
   example. See PR :pr:`4879`.
+* Added handling to :class:`pudl.transform.classes.StringCategories` so that the
+  ``categories`` key of transform params can be specified as a Path to a YAML file.
+  This will make it possible to include large categorization sets without the params
+  Python files becoming unwieldy.
```
```suggestion
  Python files becoming unwieldy. See PR :pr:`4978`.
```

src/pudl/transform/classes.py
```diff
@@ -449,6 +467,17 @@ class StringCategories(TransformParams):
     :func:`categorize_strings` to see how it is used.
     """
 
+    @field_validator("categories", mode="before")
+    @classmethod
+    def maybe_load_categories(cls, v):
+        """If categories was specified as a Path, load it from disk."""
+        # deserialize categories from disk if needed
```
krivard:
```suggestion
```

test/unit/transform/classes_test.py
```diff
@@ -182,6 +184,24 @@
     },
 }
 
+
+@pytest.fixture
+def inline_animal_cats() -> dict[str, set[str]]:
+    """Fixture for a version of ANIMAL_CATS where the categories dict is defined inline."""
+    return ANIMAL_CATS
```
krivard, line 191: This is a little silly but it makes it possible to specify treat the inline and serialized versions the same way in a `@pytest.mark.parametrize` expression

src/pudl/transform/classes.py
```diff
@@ -439,7 +441,23 @@ class StringCategories(TransformParams):
...
+        categories: dict[str, set[str]] = {...}
+        with open("categoryfile.yml", "w") as f:
+            yaml.dump(
+                {
+                    cat: sorted(val)
+                    for cat, val in categories.items()
+                },
+                f
+            )
+
```
cmgosnell, line 459: nit/ this is my expectation but ya change if your thinking something else:

```suggestion
                f
            )

We recommend putting any YAML files within a dataset directory in `pudl/scr/package_data`.
```

src/pudl/transform/classes.py
```diff
@@ -439,7 +441,24 @@ class StringCategories(TransformParams):
...
+We recommend putting any YAML files within a dataset directory in `pudl/scr/package_data`.
```
krivard: oops
```suggestion
    We recommend putting any YAML files within a dataset directory in `pudl/scr/package_data`.
```

src/pudl/transform/classes.py
```diff
@@ -439,7 +441,24 @@ class StringCategories(TransformParams):
...
+        with open("categoryfile.yml", "w") as f:
+            yaml.dump(
+                {
+                    cat: sorted(val)
+                    for cat, val in categories.items()
+                },
+                f
+            )
+
+    We recommend putting any YAML files within a dataset directory in `pudl/scr/package_data`.
```
krivard: oops again
```suggestion
    We recommend putting any YAML files within a dataset directory in ``src/pudl/package_data``.
```

Issue #4884, opened 2025-12-31
Integrate USDA RUS Form 12: Power Supply Top 10 Tables


aesharpe:
Top ten tables (in no particular order) as determined by stakeholders:
 
- [ ] **AA - OpRpt_PSStatementOfOperations.csv**: Statement of Operations - revenues, expenses, margins, and patronage capital.
- [ ] **AB - OpRpt_PSFinancialBalanceSheet.csv**: Balance Sheet - assets, liabilities, utility plant, investments, and equity.
- [ ] **C - OpRpt_PSOperatingReportSDEnergy.csv**: summary of where energy comes from (generated, purchased, interchanged) and where it goes (sales, losses, own use). Includes plant counts, capacity, net energy received, and costs.
- [ ] **CRE - OpRpt_PSRenewableGeneratingPlantSummary.csv**: data specific to renewable energy facilities: plant names, prime movers, fuel types, capacity, generation output, capacity factors, employees, O&M costs, and RUS funding.
- [ ] **DG - B_OpRpt_PSOperatingReportPlantLabor.csv**: plant labor statistics.
- [ ] **HFIV - OpRpt_PSPowerSupplyLoans.csv**: loans for power supply.
- [ ] **HH - OpRpt_PSLongTermDebts.csv**: long-term debt and debt service.
- [ ] **HI - OpRpt_PSAnnualSupplementAnnualMeetingAndBoardData.csv**: annual meeting and board data.
- [ ] **IC - OpRpt_PSLinesAndStationsLaborAndMaterialSummary.csv**: summary of labor and materials for operation and maintenance.
- [ ] **POWER - Supply_Borrower.csv**: total number of borrowers.

PR #4979, opened 2026-01-30
Transform 7-10 of top ten RUS12 tables

### LINKED ISSUE #4901: Transform USDA Form 12: Power Supply Top 10 Tables

aesharpe:
# Overview

Closes #4901 and #4884.

Adds transform functions for the remaining 4 RUS 12 tables
- [x] "statement_of_operations"
- [ ] "balance_sheet"
- [x] "sources_and_distribution"
- [ ] "renewable_plants"
- [x] "plant_labor"
- [x] "loans"
- [ ] "long_term_debt"
- [ ] "meeting_and_board"
- [ ] lines_and_stations_labor_materials"
- [ ] "borrowers"

## What did you change?

- Add dbt `schema.yml` file for transformed tables
- Add transform function for transformed tables

TO-DO: 
- [ ] Update alembic
- [ ] Add transformed table schema to `RESOURCE_METADATA` from `DRAFT_RESOURCE_METADATA`
- [ ] Add new fields to `fields.py`
- [ ] Make dbt schemas and add some simple tests

## Documentation

# Testing

- Materialize the asset in dagster
- Run unit tests

## To-do list


src/pudl/metadata/resources/rus12.py
```diff
@@ -297,12 +301,7 @@
                 "loan_maturity_date",
                 "loan_original_amount",
                 "loan_balance",
-                "for_rural_development",  # this should maybe be is_for_rural_development
-            ],
-            "primary_key": [
-                "report_date",
-                "borrower_id_rus",
-                "loan_organization",
+                "for_rural_development",
```
aesharpe, line 300: worth changing to `is_for...` or not?

src/pudl/package_data/rus12/column_maps/sources_and_distribution.csv
```diff
@@ -1,4 +1,4 @@
...
+year_index,borrower_id_rus,borrower_name_rus,report_year,report_month,fossil_steam_plant_num,fossil_steam_capacity_kw,fossil_steam_mwh,fossil_steam_cost,nuclear_plant_num,nuclear_capacity_kw,nuclear_mwh,nuclear_cost,hydro_plant_num,hydro_capacity_kw,hydro_mwh,hydro_cost,combined_cycle_plant_num,combined_cycle_capacity_kw,combined_cycle_mwh,combined_cycle_cost,internal_combustion_plant_num,internal_combustion_capacity_kw,internal_combustion_mwh,internal_combustion_cost,other_plant_num,other_capacity_kw,other_mwh,other_cost,total_plant_num,total_plant_capacity_kw,total_plant_mwh,total_plant_cost,purchased_power_mwh,purchased_power_cost,received_into_system_gross_mwh,received_into_system_gross_cost,delivered_out_of_system_gross_mwh,delivered_out_of_system_gross_cost,net_interchange_mwh,net_interchange_cost,received_into_system_wheeling_mwh,received_into_system_wheeling_cost,delivered_out_of_system_wheeling_mwh,delivered_out_of_system_wheeling_cost,net_energy_wheeled_mwh,net_energy_wheeled_cost,energy_available_for_sale_mwh,total_sales_mwh,energy_furnished_without_charge_mwh,energy_used_by_borrower_mwh,total_energy_accounted_mwh,total_energy_losses_mwh,total_energy_losses_pct
```
aesharpe: I changed most of the suffixes back from `net_energy_mwh` to just `mwh` because, upon further inspection, it felt clear enough! Also adding `net_energy` to all the longer column names felt unnecessary, and I wanted them to be consitent.

src/pudl/transform/rus.py
```diff
@@ -103,4 +106,7 @@ def multi_index_stack(
     # remove the remaining multi-index
     df.columns = df.columns.map("".join)
     df = df.dropna(subset=data_cols, how="all")
+    if drop_zero_rows:
+        all_zero_mask = (df.fillna(0)[data_cols] == 0).all(axis=1)
+        df = df[~all_zero_mask]
```
aesharpe, line 111: I added this to remove rows that are all 0 or NA as opposed to just all NA. For right now, I made it a function parameter (i.e. something you have to explicitly set instead of something that happens automatically). Let me know what you think about this -- happy to make it just part of the function if you think that makes sense. I know it wouldn't be good for categorical values which is why I made it optional.

src/pudl/metadata/resources/rus12.py
```diff
@@ -203,15 +203,17 @@
...
                 "borrower_id_rus",
                 "borrower_name_rus",
-                "operations_type"  # enum (list below)
-                "ytd_amount",
+                "operation_expense_group",
+                "operation_expense_type",
+                "ytd"  # I think this should be more descriptive
                 "ytd_budget",
-                "report_month_amount",
+                "report_month",
```
aesharpe: I'd like to add a suffix here that makes it more clear that `ytd` and `report_month` are monetary quantities not dates. Maybe the `_amount` suffix? I can either do that in the transform function or in the package_data.

src/pudl/transform/rus12.py
```diff
@@ -128,3 +128,165 @@ def core_rus12__yearly_lines_stations_labor_materials_cost(
...
+
+    # Stack by operating revenue group and expense type
+    data_cols = ["report_month", "ytd", "ytd_budget"]
+    operation_expense_group = {
+        "oprev": "operation_revenue_and_patronage_capital",
+        "opex": "operation_expense",
+        "maintex": "maintenance_expense",
+        "ces": "cost_of_electric_service",
+        "npc": "net_patronage_capital_or_margins",
+    }
```
aesharpe: I did an interesting thing here -- I'm curious what you think. Instead of just stacking by operating expense I added another layer that represents each of the total rows.

For now I left the total values in there, but I can remove them easily and then they could be easily re-calculated because of how I've labeled each of the other columns. 

This entailed adding prefix values to the `package_data`. I thought about just writing out the full name instead of making a dictionary in transform, but some of them are annoyingly long, so I just kept the abbreviations and then map them on later.

src/pudl/metadata/resources/rus12.py
cmgosnell: should these be `opex_ytd` , ` opex_tyd_budget` and `opex_report_month`? ( i would also change the group and type to `opex_group`, `opex_type` but thats a nit.

src/pudl/metadata/resources/rus12.py
```diff
@@ -297,12 +303,65 @@
...
+            "additional_details_text": "",
+        },
+        "schema": {
+            "fields": [
+                "report_date",
+                "borrower_id_rus",
+                "borrower_name_rus",
+                "source_of_energy",
+                "cost",
+                "mwh",
```
cmgosnell: i think maybe this is not descriptive enough (i know! me!). the form says "NET ENERGY RECIEVED BY SYSTEM" (ha they did my favorite spelling i/e error). scanning through our existing columns this is similar to `net_power_exchanged_mwh` (there is some specificity in that field definition re reporting requirement for EIA i think which means if you usse this it should probably get a rus specific definition but still seems conceptually the same) or just mirror their language of `net_energy_received_mwh` which seems fine and good

src/pudl/transform/rus12.py
```diff
@@ -128,3 +128,169 @@ def core_rus12__yearly_lines_stations_labor_materials_cost(
...
+        match_names=["plant_type", "data_cols"],
+        unstack_level=[
+            "plant_type",
+        ],
+        drop_zero_rows=True,
+    )
+
+    # Make sure plant num is only int values and then convert to integer
+    assert (df.plants_num.dropna() % 1 == 0).all()
+    df.plants_num = df.plants_num.astype("Int64")
```
cmgosnell: once this is fully metadata-ed this will effectively happen during the enforce schema step

src/pudl/transform/rus12.py
```diff
@@ -128,3 +128,169 @@ def core_rus12__yearly_lines_stations_labor_materials_cost(
...
+    This function process all columns from the Sources and Distribution table
+    that are not plant type specific. The plant type specific columns are processed
+    in core_rus12__yearly_sources_and_distribution_by_plant_type.
+
+    The multi_index_stack function intentionally drops a few columns that don't
+    show up in other tables. This include:
+    - ``total_energy_losses_pct`` (calculable with other cols, dropped because pct value
+    was an outlier column and not easily stacked with other columns).
+    - ``total_plant_num`` (calculable with the sources_and_distribution_by_plant_type table).
+    - ``total_capacity_kw`` (calculable with the sources_and_distribution_by_plant_type table).
```
cmgosnell, line 246: we could add these back in by adding "total" into the `plant_types`? imo we shouldn't be dropping these totals without verifying they are actually calculable

src/pudl/transform/rus12.py
cmgosnell: i love this adding the sections into the names and adding the section as a column this is great (we basically did the exact same thing)

i'm lightly pro keeping long names in mapping sheet (could be easy to swap with find `,oprev_` etc etc) bc it might be confusing to future column mapper but nbd absolutely don't think it matters either way.

src/pudl/transform/rus12.py
```diff
@@ -128,3 +128,169 @@ def core_rus12__yearly_lines_stations_labor_materials_cost(
...
+    entirely NA through all years. It then reshapes the table by stacking expense
+    types by the name of the total column for which they are calculation components.
+    """
+    df = rus.early_transform(raw_df=raw_rus12__statement_of_operations)
+
+    # Setting this assertion before stacking the table to make sure the PK holds
+    assert (
+        df[["borrower_id_rus", "borrower_name_rus", "report_date"]].duplicated().any()
+        is not True
+    ), "Primary key violation found in core_rus12__yearly_statement_of_operations"
```
cmgosnell: nbd but you can do this if you want
```suggestion
    early_check_pk(df, pk_early=["borrower_id_rus", "borrower_name_rus", "report_date"])
```

cmgosnell (commented):
sprinkling of early comments! overall it looks great.

src/pudl/transform/rus12.py
aesharpe: I agree, I'll add them to the sheet.

src/pudl/metadata/resources/rus12.py
aesharpe: Decided to keep as is

src/pudl/transform/rus12.py
aesharpe: These shouldn't be in this table, but I can add them back to the `by_plant` table.

aesharpe: I looked at it and all `total` values are NA except for `plant_num` which makes me feel good about excluding it.

src/pudl/metadata/fields.py
```diff
@@ -43,12 +43,14 @@
     MODEL_CASES_EIAAEO,
     NERC_REGIONS,
     PLANT_PARTS,
+    PLANT_TYPE_RUS12,
     PRIME_MOVER_TYPES_RUS12,
     RELIABILITY_STANDARDS,
     RENEWABLE_FUEL_TYPES_RUS12,
```
e-belfer, line 49: `Solar - photvoltaic` is probably a typo and should be `Solar - photovoltaic` in `primary_renewable_fuel_type`

dbt/models/rus12/core_rus12__yearly_loans/schema.yml
```diff
@@ -0,0 +1,16 @@
+version: 2
+sources:
+  - name: pudl
+    tables:
+      - name: core_rus12__yearly_renewable_plants
```
e-belfer: The table name is wrong here, likely copy-paste issue.

src/pudl/transform/rus12.py
```diff
@@ -132,3 +132,186 @@ def core_rus12__yearly_lines_stations_labor_materials_cost(
...
+
+@asset(io_manager_key="pudl_io_manager")
+def core_rus12__yearly_loans(raw_rus12__loans):
+    """Transform the raw_rus12__loans table."""
+    df = rus.early_transform(raw_df=raw_rus12__loans)
+    df.loan_maturity_date = pd.to_datetime(df.loan_maturity_date, format="mixed")
+    df.for_rural_development = df.for_rural_development.astype("boolean")
+
+    # Make sure loan balance isn't more than original loan amount
+    loan_diff = df.loan_original_amount - df.loan_balance
```
e-belfer: Reasonable, but why not write this in DBT?

src/pudl/transform/rus12.py
```diff
@@ -132,3 +132,186 @@ def core_rus12__yearly_lines_stations_labor_materials_cost(
...
+    )
+    # Convert units
+    df = rus.convert_units(
+        df,
+        old_unit="kw",
+        new_unit="mw",
+        converter=0.001,
+    )
+
+    # # Make sure plant num is only int values and then convert to integer
```
e-belfer: Looks like there are still open todos here - are they leftover or out of scope for this PR?

src/pudl/metadata/fields.py
```diff
@@ -7045,6 +7089,11 @@
             "Indicates whether the generator is part of a solid fuel gasification system"
         ),
     },
+    "source_of_energy": {
+        "type": "string",
+        "description": "The source of energy (not plant type).",
+        # "constraints": {"enum": SOURCE_OF_ENERGY_RUS12},
```
e-belfer: Was this something in progress that got abandoned? This should be constrained.

src/pudl/metadata/fields.py
```diff
@@ -3304,6 +3318,10 @@
             "Indicates whether the boiler is capable of re-injecting fly ash."
         ),
     },
+    "for_rural_development": {
+        "type": "boolean",
+        "description": "Whether or not the investment is for rural development.",
```
e-belfer: Maybe we can make this slightly more specific? The USDA docs say:

Identify investments in rural economic development by placing an X in column e. Include investments in any/all types of projects or products that were made to improve the economy and/or quality of life in your area.

src/pudl/metadata/fields.py
```diff
@@ -2659,6 +2661,18 @@
         "type": "integer",
         "description": "Number of employees.",
     },
+    "employees_full_time_num": {
+        "type": "integer",
+        "description": "Number of full time employees.",
```
```suggestion
        "description": "Number of employees hired full-time for normal operations of the system.",
```

e-belfer: Non-blocking but maybe slightly more specific, leaning on the USDA docs.

src/pudl/metadata/fields.py
```diff
@@ -2659,6 +2661,18 @@
...
+    },
+    "employees_part_time_num": {
+        "type": "integer",
+        "description": "Number of part time employees.",
```
e-belfer: Non-blocking: USDA defines as:

"the number of employees regularly employed on a part-time basis. Exclude employees hired for short periods of time to complete special jobs."

src/pudl/metadata/resources/rus12.py
```diff
@@ -295,60 +297,80 @@
...
+                "The primary key should be report_date, borrower_id_rus, plant_name_rus, "
+                "and plant_type, but this table did not report plant_type before 2009 and "
+                "there are respondents who report multiple rows per plant pre-2009. "
+                "The data cannot be backfilled because there is no way to distinguish between "
+                "duplicate rows pre-2009."
+            ),
+            "usage_warnings": ["experimental_wip"],
+            "additional_source_text": "(Parts D, E, F, G - Section B)",
+            "additional_details_text": (
+                "There are also some plants (at least one, Walter Scott) that are reported "
```
e-belfer: Some plants or just one? This is slightly confusing.

src/pudl/transform/rus12.py
e-belfer: If that is the case, we should be asserting this explicitly. Why drop the total plant_num if there are non-zero values?

src/pudl/transform/rus12.py
```diff
@@ -132,3 +132,186 @@ def core_rus12__yearly_lines_stations_labor_materials_cost(
...
+        unstack_level=[
+            "source_of_energy",
+        ],
+        drop_zero_rows=True,
+    )
+    return df
+
+
+@asset(io_manager_key="pudl_io_manager")
+def core_rus12__yearly_statement_of_operations(raw_rus12__statement_of_operations):
```
e-belfer, line 282: Noting that there are a number of negative opex values here - possibly out of scope for this round of integration.

aesharpe: sadly, yes

src/pudl/metadata/fields.py
aesharpe: good catch, that's a typo in the raw data

dbt/models/rus12/core_rus12__yearly_loans/schema.yml
aesharpe: Another good catch, thanks!

src/pudl/transform/rus12.py
aesharpe: fair, done!

src/pudl/transform/rus12.py
aesharpe: It has to do with the way the table is pivoted. It's easier to exclude all total values than just the non `plant_num` values that are NA. And because `plant_num` is already calculable and the rest of the total fields are unreliable I figured we should just drop them.

src/pudl/metadata/resources/rus12.py
aesharpe: I went ahead and fixed this in the code.

dbt/tests/data_tests/generic_tests/expect_col1_greater_or_equal_to_col2.sql
```diff
@@ -0,0 +1,10 @@
+{% test expect_col1_greater_or_equal_to_col2(model, col1, col2) %}
```
e-belfer: There's already the dbt-expectations test that does this: [expect_column_pair_values_A_to_be_greater_than_B](https://github.com/calogica/dbt-expectations/blob/main/macros/schema_tests/multi-column/expect_column_pair_values_A_to_be_greater_than_B.sql)

You can use row_condition to specify not null values.

dbt/tests/data_tests/generic_tests/expect_positive_values.sql
```diff
@@ -0,0 +1,14 @@
+{% test expect_positive_values(model, positive_columns) %}
```
e-belfer: Same thing, you could already do this using the dbt-expectations test [expect_column_min_to_be_between](https://github.com/calogica/dbt-expectations?tab=readme-ov-file#expect_column_min_to_be_between) with the min value set as 0 and no max value or max value also set as 0.

See core_eia176__yearly_gas_disposition as one example of many uses of this pattern.

dbt/tests/data_tests/generic_tests/schema.yml
```diff
@@ -289,3 +289,22 @@ macros:
       - name: max_differing_fuel_rate
         type: float
         description: Maximum acceptable fraction of units mapping to multiple fuels. (optional; default 0.01)
+
+  - name: test_expect_col1_greater_or_equal_to_col2
```
e-belfer: See my comments above - there are already patterns to do these things, which are widely implemented elsewhere, even if they're slightly awkward. What's the motivation to write new tests?

src/pudl/transform/rus12.py
```diff
@@ -155,6 +155,18 @@ def core_rus12__yearly_plant_labor(raw_rus12__plant_labor):
...
 
+    # Remove duplicate Walter Scott plant entries.
+    exclude_cols = ["borrower_id_rus", "borrower_name_rus"]
+    dupe_mask = (
+        df["borrower_id_rus"].isin(["IA0083", "IA0084"])
+        & (df["plant_name_rus"] == "Walter Scott")
+        & df.drop(columns=exclude_cols).duplicated(keep=False)
+        & (
+            df["borrower_id_rus"] == "IA0083"
+        )  # dropping the IA0083 and keeping the IA0084 so both borrowers show up in the data.
```
e-belfer, line 166: Is there any reason you chose one over the other? I could imagine co-ownership where both owners are taking out separate loans for their portion of the plant, but if one of the entries has no data or you did some research and verified this is incorrect it'd be helpful to know. Otherwise my instinct would not be to drop but to note it in the description as you had before.

src/pudl/metadata/resources/rus12.py
e-belfer: See my comments there - my instinct is not to drop one or the other unless you have a clear rationale for doing so, just to more clearly explain whether this is a unique case or an example of a broader problem.

I see both are reporting identical values for all years, but even without this question each of those borrowers is reporting more than one row per plant/year with different values. So the PK question to me is bigger than just whether one or the other is the correct borrower.

src/pudl/transform/rus12.py
aesharpe: All the numbers are the same (it's clearly a duplicate), so I think dropping it makes sense.

aesharpe: I explained in the comment that I drop IA0083 so that IA0084 shows up in the data. It's a little arbitrary, but I don't think it's a big deal. The table is about plant labor more than it is what utility is reporting plant labor so I think it's more important not to have duplicate rows than it is to preserve the two utilities reporting the same data.

src/pudl/metadata/resources/rus12.py
aesharpe: The reason there is more than one row per plant/year besides just the Walter Scott issue is because there is no `plant_type` field pre-2009. That feels like a different type of primary key error than the Walter Scott one which feels like a spot fix to me.

dbt/tests/data_tests/generic_tests/expect_col1_greater_or_equal_to_col2.sql
aesharpe: To be honest, I haven't really worked with dbt before, and I didn't realize there were default dbt tests, so I was just going off the existing tests in pudl.

aesharpe: I fixed it to the dbt one!

src/pudl/transform/rus12.py
aesharpe: Also, I thought that we usually remove total columns if they are calculable?

src/pudl/metadata/resources/rus12.py
```diff
@@ -295,60 +297,78 @@
...
             ],
         },
         "sources": ["rus12"],
         "etl_group": "rus12",
         "field_namespace": "rus",
     },
 }
-
-
-operations_type_enum = [
```
e-belfer, line 313: Did this get moved elsewhere or is it no longer needed?

src/pudl/transform/rus12.py
e-belfer: I think I have the opposite instinct but that's ok, your argument makes sense to me and I don't want to hold up this PR. It would probably be helpful to note that you did this in the additional table description so that if someone is looking by borrower ID for the dropped borrower they aren't totally confused about why there's no data there.

src/pudl/metadata/resources/rus12.py
aesharpe: no longer needed

PR #4991, opened 2026-02-04
Improve core EIA930 performance

zschira:
# Overview

While working on #4985 I found that the new EIA930 data was the straw that broke the camel's back in terms of actually being able to run the `core` assets (specifically the multi-asset `core_eia930__hourly_operations_assets`).

This PR uses duckdb to improve the performance of these assets to make them once again runnable on my local desktop.

# Testing

I haven't been able to get the script `devtools/check_against_nightly.py` to pass and rather than trying to get this working, I decided to just use duckdb to spot check these tables. To do this I used queries like the following:

```sql
WITH new AS (SELECT * FROM 'core_eia930__hourly_net_generation_by_energy_source.parquet')
WITH remote AS (SELECT * FROM 's3://pudl.catalyst.coop/nightly core_eia930__hourly_net_generation_by_energy_source.parquet')
SELECT * FROM new
JOIN remote
USING(datetime_utc, balancing_authority_code_eia, generation_energy_source)
WHERE abs(new.net_generation_imputed_eia_mwh - remote.net_generation_imputed_eia_mwh) > 0.1;
```

And this should return 0 rows. I repeated this for all `float` cols.

### DIFF
--- a/src/pudl/transform/eia930.py
+++ b/src/pudl/transform/eia930.py
@@ -1,5 +1,7 @@
 """Module to perform data cleaning functions on EIA930 data tables."""
 
+import re
+
 import duckdb
 from dagster import AssetOut, Output, asset, multi_asset
 
@@ -77,27 +79,23 @@ def _transform_hourly_operations(
     table: duckdb.DuckDBPyRelation, conn: duckdb.DuckDBPyConnection
 ) -> ParquetData:
     """Transform the eia930 hourly operations table."""
+    nondata_cols = [
+        "datetime_utc",
+        "balancing_authority_code_eia",
+    ]
+    data_col_pattern = r"(demand|interchange|net_generation_total)"
+
+    # Grab columns and rename as needed
     operations = table.select(
-        duckdb.ColumnExpression("datetime_utc"),
-        duckdb.ColumnExpression("balancing_authority_code_eia"),
-        duckdb.ColumnExpression("net_generation_total_reported_mwh").alias(
-            "net_generation_reported_mwh"
-        ),
-        duckdb.ColumnExpression("net_generation_total_adjusted_mwh").alias(
-            "net_generation_adjusted_mwh"
-        ),
-        duckdb.ColumnExpression("net_generation_total_imputed_mwh").alias(
-            "net_generation_imputed_eia_mwh"
-        ),
-        duckdb.ColumnExpression("interchange_reported_mwh"),
-        duckdb.ColumnExpression("interchange_adjusted_mwh"),
-        duckdb.ColumnExpression("interchange_imputed_mwh").alias(
-            "interchange_imputed_eia_mwh"
-        ),
-        duckdb.ColumnExpression("demand_reported_mwh"),
-        duckdb.ColumnExpression("demand_adjusted_mwh"),
-        duckdb.ColumnExpression("demand_imputed_mwh").alias("demand_imputed_eia_mwh"),
-        duckdb.ColumnExpression("demand_forecast_mwh"),
+        *[
+            duckdb.ColumnExpression(name).alias(
+                name.replace("net_generation_total_", "net_generation_").replace(
+                    "imputed_mwh", "imputed_eia_mwh"
+                )
+            )
+            for name in table.columns
+            if name in nondata_cols or re.search(data_col_pattern, name)
+        ],
     )
     return persist_table_as_parquet(operations, "core_eia930__hourly_operations")
 


src/pudl/transform/eia930.py
```diff
@@ -31,72 +122,23 @@ def core_eia930__hourly_operations_assets(
...
+        conn.sql("SET memory_limit = '16GB';")
+        netgen_parquet = _transform_netgen_by_source(tbl, conn)
+        operations = _transform_hourly_operations(tbl, conn)
 
     # NOTE: currently there are some BIG differences between the calculated totals and
     # the reported for net generation.
     return (
         Output(
-            value=netgen_by_source,
+            value=lf_from_parquet(netgen_parquet),
```
zschira, line 135: Read as polars then use the io-manager to write back to parquet. This is just because we have at least some basic schema enforcement setup for Polars, but is otherwise not necessary.

src/pudl/transform/eia930.py
```diff
@@ -1,16 +1,107 @@
...
+    return persist_table_as_parquet(
+        netgen_by_source, "core_eia930__hourly_net_generation_by_energy_source"
+    )
+
+
+def _transform_hourly_operations(
+    table: duckdb.DuckDBPyRelation, conn: duckdb.DuckDBPyConnection
+) -> ParquetData:
+    """Transform the eia930 hourly operations table."""
+    operations = table.select(
```
zschira, line 89: I tried to use [column expressions](https://duckdb.org/docs/stable/sql/expressions/star#renaming-columns-with-regular-expressions-in-a-columns-expression) to do the select/renaming here, but it got complex with two rounds of renaming (`net_generation_total`->`net_generation` + `imputed_mwh` -> `imputed_eia_mwh`). Just listing out all the columns in the `select` and aliasing the ones that need a name change is verbose, but much more readable.

src/pudl/transform/eia930.py
```diff
@@ -31,72 +122,23 @@ def core_eia930__hourly_operations_assets(
...
-            "net_generation_imputed_mwh": "net_generation_imputed_eia_mwh",
-            "demand_imputed_mwh": "demand_imputed_eia_mwh",
-            "interchange_imputed_mwh": "interchange_imputed_eia_mwh",
-        }
-    )
+    with duckdb_relation_from_parquet(raw_eia930__balance) as (
+        tbl,
+        conn,
+    ):
+        conn.sql("SET memory_limit = '16GB';")
```
krivard, line 127: Does 16 have a more specific justification than "that's the size machine we're trying to target as a minimum spec"?

src/pudl/transform/eia930.py
```diff
@@ -1,16 +1,107 @@
...
+                for long, short in data_cols
+            ]
+        )
+    ).to_view("raw")
+    conn.query(
+        # Transform wide table to long
+        # category column will contain values formatted like ``{energy_source}_{adjusted|imputed|reported}``
+        "UNPIVOT raw "
+        f"ON {', '.join([short for _, short in data_cols])} "
+        "INTO NAME category VALUE net_generation_mwh",
```
krivard, line 53: nice use of UNPIVOT!

src/pudl/transform/eia930.py
```diff
@@ -1,16 +1,107 @@
...
+        ),
+        duckdb.ColumnExpression("interchange_reported_mwh"),
+        duckdb.ColumnExpression("interchange_adjusted_mwh"),
+        duckdb.ColumnExpression("interchange_imputed_mwh").alias(
+            "interchange_imputed_eia_mwh"
+        ),
+        duckdb.ColumnExpression("demand_reported_mwh"),
+        duckdb.ColumnExpression("demand_adjusted_mwh"),
+        duckdb.ColumnExpression("demand_imputed_mwh").alias("demand_imputed_eia_mwh"),
+        duckdb.ColumnExpression("demand_forecast_mwh"),
```
krivard: Potential regression opportunity: the old code selected columns based on name regex, which means it did not need to be updated when new columns were added, old ones deleted, or for minor name changes after the prefix. By contrast, this version hard-codes all column names exactly.

I'm guessing it's like this because there's no easy way to build a list of columns programmatically, in which case, we should file a followup issue to look at this in more depth when we're not on a deadline.

src/pudl/extract/eia930.py
```diff
@@ -99,10 +91,8 @@ def extract_page(
         f"SELECT * FROM {view_name}"  # noqa: S608 (we trust this view name)
         for view_name in individual_views
     )
-    output_path = PudlPaths().parquet_path(f"raw_eia930__{page}")
     all_partitions = con.query(union_query)
-    all_partitions.to_parquet(str(output_path))
-    return output_path
+    return persist_table_as_parquet(all_partitions, table_name=f"raw_eia930__{page}")
```
krivard, line 95: yes good

krivard (approved):
No major blockers; a curious note or two

src/pudl/transform/eia930.py
zschira: No that's pretty much it. If we actually wanted it to run on a 16GB machine we'd probably need a lower limit, but this seemed like a reasonable balance to me.

src/pudl/transform/eia930.py
zschira: After looking at this again, I realized it wasn't actually too hard to build the column list programmatically, so I just took care of that so we don't have to worry about another issue.

Issue #5003, opened 2026-02-09
Create Standalone data deployment pipeline


jdangerx:
### Overview

Following on #4775, which decoupled our nightly builds from Zenodo, it will be nice to be able to fully actualize the dream of "we can distribute *already built* artifacts instead of running the ETL for every release."

### Success criteria

* [ ] `deploy_pudl` workflow can be triggered manually, for both nightly/stable and with staging/non-staging environments
* [ ] Workflow replicates existing behavior:
  * [x] compress SQLite files
  * [x] bundle Parquet files
  * [x] move Parquet files and associated datapackage.json to top level
  * [x] remove files/dirs we *don't* want to distribute
  * [ ] nightly flow:
    * [x] upload to `nightly` and `eel-hole`
    * [x] reboot pudl viewer
    * [x] trigger Zenodo upload
    * [x] `nightly` branch updated to point at `nightly-YYYY-MM-DD`
  * [ ] stable flow:
    * [x] upload to `stable` and `vXXXX`
    * [x] temporary hold on GCS versioned output
    * [x] trigger Zenodo upload
    * [x] `stable` branch updates to point at `vXXXX`
* [ ] Workflow notifies Slack on failure or success - if failure, notes what stages succeeded and failed
* [ ] Workflow can work from build artifacts uploaded to GCS
* [ ] There is staging behavior:
  * [ ] when run on a staging environment, distributes data to `/staging/{what_the_path_would_have_been}` so we can test that this logic works
  * [x] `staging` paths auto-delete after 7 days
  
### Next steps
* [ ] make stub workflow and put it in `main` so we can start testing
* [ ] write the dang code and trigger it via workflow
* [ ] test that we can use it to distribute a set of build artifacts stored on GCS, to both nightly-staging and stable-staging (use most recent nightly and versioned builds so we don't have to re-run any ETLs)
* [ ] update `build-deploy-pudl` workflow to not try to deploy PUDL anymore
* [ ] make the new `deploy-pudl` workflow trigger on nightly build success (via a tag push trigger?)
* [ ] update release process

PR #5004, opened 2026-02-09
feat: add stub deploy-pudl workflow

jdangerx:
# Overview

A small piece of #5003 .

To split out the deployment we need a separate workflow that we can test. GHA needs *some* stub in `main` so that the branch versions will show up in the workflow dispatch UI.

## What did you change?

Add stub workflow.


krivard (approved):
legit!

PR #5008, opened 2026-02-11
Add FERC Form 1 identification table to PUDL

e-belfer:
# Overview

Addresses first step in #5150 

In #4975, we updated the utility matching method to auto-match utilities when they hear identical (cleaned) utility names. We want to extend this method to consider utility addresses in matching.

The first step to doing this will be to bring in the FERC Form 1 identification table, which includes data on utility addresses necessary for matching.

## What did you change?

- Add FERC Form 1 identification table as `core_ferc1__yearly_identification_and_certification`
- Determine method for pre-processing of addresses on FERC side
- Updated the FERC transform process to handle cases where we don't want to default concat all the tables.
- Added the most basic data validations and metadata  

## Documentation

- [x] Update the [release notes]

# Testing

## To-do list

- [x] Run `pixi run pre-commit-run` to run linters and static code analysis checks.
- [x] Run `pixi run pytest-ci` locally to ensure that the merge queue will accept your PR.

### DIFF
--- a/src/pudl/extract/ferc1.py
+++ b/src/pudl/extract/ferc1.py
@@ -214,6 +214,8 @@ TABLE_NAME_MAP_FERC1: dict[str, dict[str, str]] = {
         "xbrl": [
             "identification_001",
             "corporate_officer_certification_001",
+        ],
+    },
     "core_ferc1__yearly_other_regulatory_assets_sched232": {
         "dbf": "f1_othr_reg_assets",
         "xbrl": [
--- a/migrations/versions/4f252e9e2ce3_add_identification_and_certification_.py
+++ b/migrations/versions/4f252e9e2ce3_add_identification_and_certification_.py
@@ -1,7 +1,7 @@
 """Add identification and certification table
 
 Revision ID: 4f252e9e2ce3
-Revises: cfbb9479cc73
+Revises: 6b25097da735
 Create Date: 2026-04-14 14:30:23.354100
 
 """
@@ -11,7 +11,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = '4f252e9e2ce3'
-down_revision = 'cfbb9479cc73'
+down_revision = '6b25097da735'
 branch_labels = None
 depends_on = None
 


src/pudl/extract/ferc1.py
```diff
@@ -205,6 +205,13 @@
         "dbf": "f1_othr_reg_liab",
         "xbrl": "other_regulatory_liabilities_account_254_278",
     },
+    "core_ferc1__yearly_identification_certification": {
+        "dbf": "f1_ident_attsttn",
+        "xbrl": [
+            "identification_001",
+            "corporate_officer_certification_001",
+        ],
```
cmgosnell, line 224: ow double xbrl tables! neat

src/pudl/transform/params/ferc1.py
```diff
@@ -3289,6 +3289,61 @@
...
+    "core_ferc1__yearly_identification_certification": {
+        "rename_columns_ferc1": {
+            "dbf": {
+                "columns": {
+                    "respondent_id": "utility_id_ferc1_dbf",
+                    # "filing_name": "filing_name",
+                    "cntc_phone": "contact_phone",
+                    "respondent_name": "utility_name_ferc1",
+                    "cntc_name": "contact_name",
+                    "chg_dt_prev_name": "prior_utility_name_ferc1",  # or "company_name_old"?
```
cmgosnell: i like prior better than old but that's light personal preference

src/pudl/transform/params/ferc1.py
```diff
@@ -3289,6 +3289,61 @@
             }
         ],
     },
+    "core_ferc1__yearly_identification_certification": {
+        "rename_columns_ferc1": {
+            "dbf": {
+                "columns": {
+                    "respondent_id": "utility_id_ferc1_dbf",
+                    # "filing_name": "filing_name",
```
cmgosnell: why #-ed?

src/pudl/transform/params/ferc1.py
```diff
@@ -3289,6 +3289,61 @@
...
+            "xbrl": {
+                "columns": {
+                    "entity_id": "utility_id_ferc1_xbrl",
+                    # "filing_name": "filing_name",
+                    "start_date": "start_date",
+                    "end_date": "end_date",
+                    "telephone_of_contact_person": "contact_phone",
+                    "respondent_legal_name": "utility_name_ferc1",
+                    "migrated_data": "is_migrated_data",
+                    "vendor_name": "filing_software_vendor_name",
```
cmgosnell, line 3417: is this for who actually does the submissions to form 1? that's bogus i hate they both knew the utilities would have to hire someone to deal with XBRL and the XBRL peeps wanted to know who was serving this new market they created.... (obviously noting to do here just screaming into the void)

src/pudl/transform/ferc1.py
```diff
@@ -3071,6 +3078,133 @@ def reconcile_table_calculations(
...
+        df["prior_utility_name_ferc1"] = df["prior_utility_name_ferc1"].replace(
+            pattern, pd.NA, regex=True
+        )
+
+        df[
+            ["office_street_address", "office_city", "office_state", "office_zip_code"]
+        ] = pd.DataFrame(
+            df["office_street_address"].apply(parse_address).tolist(),
+            index=df.index,
+        )
```
katie-lamb: a few suggested cleaning steps:
* it might be nice to make them lowercase or title case, but not a big deal. will be done in splink/during match anyways
* expand address suffixes to their full name and stripping punctuation, or vice versa. e.g. st. -> street. this could also happen within a separate cleaning in the match step. i believe i do this for the SEC addresses and can dig it up

katie-lamb: taking these cleaning steps also depends on how useful address actually is for the match. maybe zip or state is really more useful

src/pudl/helpers.py
```diff
@@ -2573,3 +2574,45 @@ def normalize_year_fragments(
...
+            parsed.get("OccupancyType", ""),
+            occupancy,  # Only add if occupancy type exists
+        ]
+        street_address = " ".join([p for p in street_parts if pd.notna(p)]).strip()
+
+        return (
+            None if street_address == "" else street_address,
+            parsed.get("PlaceName", None),
+            parsed.get("StateName", None),
+            parsed.get("ZipCode", None),
```
katie-lamb, line 2614: it seems like we can maybe do the match without state and zip, but it's probably nice to have those columns in this table anyways.

src/pudl/transform/ferc1.py
```diff
@@ -3071,6 +3078,133 @@ def reconcile_table_calculations(
...
+        """
+        # df = self.reconcile_table_calculations(df).pipe(self.enforce_schema)
+        # if df.empty:
+        #     raise ValueError(f"{self.table_id.value}: Final dataframe is empty!!!")
+        # for col in df:
+        #     if df[col].isna().all():
+        #         raise ValueError(
+        #             f"{self.table_id.value}: Column {col} is entirely NULL!"
+        #         )
+        return df
```
cmgosnell: presumably you'll delete this when you add metadata?

src/pudl/transform/ferc1.py
```diff
@@ -3071,6 +3078,133 @@ def reconcile_table_calculations(
...
+            # Sort to avoid dependence on the ordering of rename_columns.
+            # Doing the sorting here because we have a particular ordering
+            # hard coded for the DBF primary keys.
+            pk_cols += sorted(col for col in cols if col.endswith("_axis"))
+        return pk_cols
+
+    @cache_df(key="dbf")
+    def drop_unused_original_columns_dbf(self, df: pd.DataFrame) -> pd.DataFrame:
+        """Remove residual DBF specific column."""
+        return df
```
cmgosnell: you need this? or are you specifically wanting to not drop these? if yes then just add comment plzzz

src/pudl/helpers.py
katie-lamb: oh actually, state and zip are pretty useful to have, for what's called a "blocking rule", basically we filter for pairs that have an exact match on state and zip.

src/pudl/transform/ferc1.py
katie-lamb: Seems like there will be a normalization process on most of these columns during the match / inside the match module, so i think it's fine to keep them not as normalized here.

e-belfer: If any of those steps seem like they'd make sense to migrate into the actual table transform, definitely let me know.

src/pudl/transform/params/ferc1.py
e-belfer: This column doesn't exist in this table.

src/pudl/transform/params/ferc1.py
e-belfer: Yes, I believe so.

src/pudl/transform/ferc1.py
e-belfer: I updated the docstring, these columns don't exist in this table.

src/pudl/transform/ferc1.py
katie-lamb: I think the current steps in this transform seem good for getting this table to a state of readability. The pre-match transform can take on cleaning steps that are more specific to making it viable to match

src/pudl/transform/ferc1.py
```diff
@@ -2586,6 +2600,42 @@ def drop_duplicate_rows_dbf(
             df = drop_duplicate_rows_dbf(df, params=params)
         return df
 
+    # TODO: Ok not to use cache decorator??
```
cmgosnell: its fine.  if you want I'd use `@cache_df("dbf")`/`@cache_df("xbrl")` respectively. adding these would basically mean if this step passes but the next bit doesn't you could access it. also you can access the various stages along the way with these cached dfs. but that's a long winded way of saying this is fine without the caching this step is pretty reconstructable if need be

src/pudl/transform/ferc1.py
```diff
@@ -2586,6 +2600,42 @@ def drop_duplicate_rows_dbf(
...
+        return pd.concat(list(raw_dbf_dfs.values()))
+
+    # TODO: Ok not to use cache decorator??
+    def preprocess_xbrl(self, raw_xbrl_dfs: dict[str, pd.DataFrame]) -> pd.DataFrame:
+        """Pre-process XBRL inputs into one dataframe. Grab freshest data and concat by default."""
+        raw_xbrls = {
+            table_name: filter_for_freshest_data_xbrl(
+                df,
+                get_primary_key_raw_xbrl(
+                    table_name.removeprefix("raw_ferc1_xbrl__"), "ferc1"
```
cmgosnell, line 2619: not blocking at all but his whole pass in the dictionary of named dfs is totally acceptable but my preference personally would be to pass in lists of dfs. which i thiiiink could be done real easy if it was `raw_dbf_dfs` and `raw_xbrl_instant_dfs` and `raw_xbrl_duration_dfs` and this one lil guy changed. 

```suggestion
                    self.table_name.value.removeprefix("core_ferc1_xbrl__"), "ferc1"
```

src/pudl/transform/ferc1.py
```diff
@@ -3071,6 +3121,204 @@ def reconcile_table_calculations(
...
+        )
+        # Rather than concatenating, we merge these tables on their primary key.
+        # This is because we want to join these two tables into one horizontally
+        # rather than vertically.
+        # We sort to consistently assign the table name to identification.
+        duration_df_list = [
+            df
+            for key, df in sorted(raw_xbrls.items(), reverse=True)
+            if key.endswith("_duration")
+        ]
```
cmgosnell, line 3166: hm my suggestion of pass in list of dfs would junk this up a bit. ignore if that's complicating. ( i suppose you could drop the ones with the whatever values in there but i don't feel strongly.

src/pudl/metadata/resources/ferc1.py
```diff
@@ -336,6 +336,17 @@
...
     },
+    "yearly_identification_certification": {
+        "additional_summary_text": "respondent identification and corporate officer certification information.",
+        "additional_source_text": "(Part I and II)",
+        "usage_warnings": ["free_text"],
+        "additional_details_text": (
+            "This table contains information identifying a utility's contact information, "
+            "addresses and individuals responsible for accounting and certification. Note that "
+            "the entity_id_gleif field does not conform to the expected format and largely includes "
+            "the names of respondents."
```
cmgosnell, line 375: 🤦🏻

src/pudl/transform/ferc1.py
e-belfer: I can use it for dbf but not xbrl because it expects a dataframe, not a tuple. I thought it might be confusing to have one but not the other cached, and I didn't want to refactor even further - what do you think?

src/pudl/transform/ferc1.py
e-belfer: I think this would require some more refactoring of the preprocess_xbrl step, so I'm going to leave a note about it as a future TODO but not do it in this PR.

e-belfer:
@cmgosnell Because this touches all of the FERC transforms, I'm going to kick off a build to make sure nothing unexpected happens.

src/pudl/metadata/fields.py
```diff
@@ -11108,6 +11138,44 @@
             )
         }
     },
+    "core_ferc1__yearly_identification_certification": {
+        "contact_name": {"description": ("Name of contact person for the respondent.")},
+        "contact_title": {
+            "description": "Title of contact person for the respondent.",
+        },
+        # TODO 04/26: Would require additional cleaning to validate, some
```
e-belfer, line 11174: @katie-lamb There are some Ontario postal codes and also non-standardized zips in here that would require more data rescue to figure out, so for now I am not constraining. This could be an additional round of improvements that might rescue more matches.

PR #5015, opened 2026-02-16
Duckdb helper script

zschira:
# Overview

This PR adds a new helper script which generates a duckdb file with views pointing at local and remote parquet files. I've been using this locally during development to have an easy way to spot check data and compare new versions of a table to nightly or stable outputs.

The script will create 3 schemas: `stable`, `nightly`, and `local`. Within each of these schemas there will be view for every table that exists in the corresponding location (if you haven't materialized an asset locally duckdb would error if you tried to create a view pointing to the nonexistent parquet file). This allows you to run queries like:

```sql
SELECT * FROM {local|nightly|stable}.{table_name};
```

I've also found that using this with the [duckdb ui](https://duckdb.org/docs/stable/core_extensions/ui) can be really useful for doing data updates and comparing the updated version of a table to the previous nightly build, because duckdb will automatically generate some useful column level stats like # of nulls, # of unique values, a histogram of values, etc.


src/pudl/scripts/generate_pudl_duckdb.py
```diff
@@ -0,0 +1,59 @@
...
+        [
+            conn.sql(
+                f"CREATE OR REPLACE VIEW {schema}.{table} AS "  # noqa: S608
+                + f"(SELECT * FROM '{base_path}/{table}/*.parquet')".replace(  # noqa: S608
+                    "nightly", "ferceqr"
+                )
+            )
+            for table in PARTITIONED_TABLES
+            for schema, base_path in schema_path_map.items()
+            if schema != "stable" and UPath(f"{base_path}/{table}/", anon=True).exists()
```
krivard, line 55: why do we need `schema != stable` here?

zschira: I've changed the comment to refer to make it clear that this only applies to EQR tables, and the reason it does the `schema != stable` thing is because there's only one copy of EQR on s3, so this just adds EQR to the `nightly` schema, and not the `stable` one.

PR #5016, opened 2026-02-17
Separate data distribution from nightly builds

### LINKED ISSUE #5003: Create Standalone data deployment pipeline

jdangerx:
# Overview

Closes #5003.

## Documentation

# Testing

## To-do list


.github/workflows/deploy-pudl.yml
```diff
@@ -2,11 +2,120 @@
...
+          - nightly
+          - stable
+      git_tag:
+        description: "Git tag to merge into nightly/stable branch and trigger Zenodo workflow for"
+        required: true
+        type: string
+      docker_tag:
+        description: "Dockerhub tag name"
+        type: string
+      gcs_build_location:
```
zaneselvans: It seems like we could generate the list of builds available for deployment programmatically by listing the bucket contents, and keeping only those builds that match the expected pattern (starts with `YYYY-MM-DD-HHMM-` and ends with either `-main` or `-vYYYY.M.x`) and make this into a dropdown somehow? Which would avoid typos and accidental deployment of branch or otherwise inappropriate outputs.  There are typically only ~20 outputs since we only keep them for 30 days.

.github/workflows/deploy-pudl.yml
```diff
@@ -2,11 +2,120 @@
...
+        required: true
+        type: choice
+        options:
+          - nightly
+          - stable
+      git_tag:
+        description: "Git tag to merge into nightly/stable branch and trigger Zenodo workflow for"
+        required: true
+        type: string
+      docker_tag:
```
zaneselvans: What's the scenario where we would need a specific docker tag, and how would we even know what it should be?

.github/workflows/deploy-pudl.yml
```diff
@@ -2,11 +2,120 @@
...
+    # inputs: deploy target (nightly/stable), staging or not, GCS build location (bucket is mounted to batch job with gcsfuse)
+    inputs:
+      deploy_target:
+        description: "Deploy target (nightly/stable)"
+        required: true
+        type: choice
+        options:
+          - nightly
+          - stable
+      git_tag:
```
zaneselvans, line 6: Is this the normal operation that you're envisioning?

- Nightly builds happen on a schedule as they are now.
- At the end of a nightly build, if it was successful, it kicks off a nightly data distribution.
- When we want to do a versioned release, we push a `v20YY.M.x` tag to some previously successful nightly build commit.
- This tag triggers the distribution of the corresponding nightly build outputs as a stable versioned release.

As it's outlined right now, it seems like we'd be able to accidentally distribute the wrong data, e.g to send a random branch build to the stable outputs. Or to deploy data from a build that completed but didn't pass data validations. Or maybe that didn't even complete.

.github/workflows/deploy-pudl.yml
```diff
@@ -2,11 +2,120 @@
...
+      - id: "auth"
+        uses: "google-github-actions/auth@v3"
+        with:
+          workload_identity_provider: "projects/345950277072/locations/global/workloadIdentityPools/gh-actions-pool/providers/gh-actions-provider"
+          service_account: "deploy-pudl-github-action@catalyst-cooperative-pudl.iam.gserviceaccount.com"
+
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@v3
+
+      - name: Make GCP Batch config file
```
zaneselvans, line 74: Is it actually necessary to spin up a Batch job? Or could we do all of this deployment from a GitHub runner?

devtools/generate_batch_config.py
zaneselvans, line 1: Seems like this is just re-org of existing contents, no new functionality?

src/pudl/deploy/pudl.py
```diff
@@ -0,0 +1,258 @@
...
+        gcs_fs.put(f"{source_dir}/*", gcs_path, recursive=True)
+
+        s3_path = f"s3://pudl.catalyst.coop/{suffix}/"
+        logger.info(f"Uploading outputs to {s3_path}")
+        s3_fs.put(f"{source_dir}/*", s3_path, recursive=True)
+
+    logger.info(f"Upload complete for {len(path_suffixes)} path(s)")
+
+
+def update_git_branch(tag: str, branch: str, staging: bool) -> None:
```
zaneselvans, line 147: Should do some input validation here to ensure that ONLY nightly tags can ever be merged into `nightly` and ONLY versioned tags can be merged into `stable`. Or alternatively, given that we can infer what branch the tag should be getting merged into based on its value, maybe we should not have the branch parameter at all.

src/pudl/scripts/deploy.py
```diff
@@ -0,0 +1,194 @@
...
+        path_suffixes=path_suffixes,
+    )
+
+    update_git_branch(tag=git_tag, branch="nightly", staging=staging)
+    if not staging:
+        trigger_zenodo_release(
+            build_ref=git_tag,
+            env="sandbox",
+            source_dir="s3://pudl.catalyst.coop/nightly/",
+            ignore_regex="",
```
zaneselvans: The default `ignore_regex` that we have right now is `(^.*\.parquet$|^pudl_parquet_datapackage\.json$)` is there a reason that we don't want to continue with that?

src/pudl/deployment/deploy_outputs.py
```diff
@@ -0,0 +1,258 @@
...
+
+import gcsfs
+import requests
+import s3fs
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def prepare_outputs_for_distribution(output_dir: Path) -> None:
```
zaneselvans: With the `builds.catalyst.coop` bucket mounted to the VM, do these filesystem operations not affect the bucket contents?  In the current system, we copy everything to `builds.catalyst.coop` and then mess with the (now disposable) local filesystem, leaving the build outputs as they appeared at the end of the build. Seems like we want to keep the build outputs pristine, in case we need to debug something, or attempt to distribute the same outputs again when something goes wrong.

src/pudl/scripts/deploy.py
```diff
@@ -0,0 +1,194 @@
...
+
+    Orchestrates the full deployment workflow:
+    1. Prepare outputs (compress SQLite, create parquet archive)
+    2. Upload to cloud storage (GCS and S3)
+    3. Update git branches (if not staging)
+    4. Set GCS temporary hold for versioned releases (stable only, not staging)
+    5. Trigger Zenodo release (if not staging)
+    6. Update Cloud Run service (nightly only, not staging)
+    """
+    # TODO 2026-02-11: munge GCS to local - pull out later
```
zaneselvans: Can we pull this out later? Given that we modify the filesystem layout?

src/pudl/scripts/deploy.py
```diff
@@ -0,0 +1,194 @@
...
+    "--staging",
+    type=bool,
+    default=False,
+    help=(
+        "Upload to staging locations for validation. Skips Zenodo triggers "
+        "and Cloud Run deployments."
+    ),
+    show_default=True,
+)
+def pudl_deploy(
```
zaneselvans, line 143: Somewhere very early in the process we should be checking for the `success` file in the source directory and bailing if it's not present, since that means we're looking at a failed build, and we should not distribute the contents.

src/pudl/scripts/deploy.py
```diff
@@ -0,0 +1,194 @@
...
+    "deploy_type",
+    type=click.Choice(["nightly", "stable"], case_sensitive=False),
+)
+@click.argument("source_path", type=str)
+@click.option(
+    "--git-tag",
+    type=str,
+    required=True,
+    help="Git tag to merge into branch (e.g., nightly-2026-02-09 or v2026.2.9).",
+)
```
zaneselvans: If I understand correctly:

- The git tag is assumed to already exist prior to attempting data distribution.
- The git tag already implies a specific git commit / hash.
- We can identify what build outputs are associated with that hash based on the path within `builds.catalyst.coop`
- If we automatically identify what build outputs to distribute based on the tag/hash, then we have fewer input params to manage.
- This would also make it impossible to accidentally distribute the wrong build outputs.
- The format of the git tag can also be used to infer whether it's appropriate to distribute the build outputs to `nightly` or `stable` so maybe we don't need that parameter either.

src/pudl/scripts/deploy.py
zschira, line 143: I decided to implement the EQR deployment using dagster constructs, because I think there are benefits to moving towards moving as much stuff as possible into dagster. With the setup I have for EQR, you can run the deployment assets locally, and also include them in the EQR job, so they get automatically executed during batch builds. I think this achieves the same flexibility of having a standalone script, but also means we could more easily migrate to a full dagster deployment in the future if we wanted to. I can see pros and cons to both approaches, but I think it would be good to decide on a single way of doing deployments to the extent possible.

src/pudl/etl/deploy_outputs.py
```diff
@@ -0,0 +1,258 @@
...
+    logger.info(f"Created parquet archive: {archive_path}")
+
+    logger.info("Removing dbt database.")
+    test_db = output_dir / "pudl_dbt_tests.duckdb"
+    test_db.unlink()
+
+    logger.info("Output preparation complete")
+
+
+def _run(cmd: list[str]) -> None:
```
zschira: A lot of these `_run` calls are running things that could theoretically be done with a python API instead of kicking off a subprocess. Might be cleaner to moving python API's directly when possible.

src/pudl/scripts/deploy.py
jdangerx: I think this TODO is out of date. Agree that since we modify the filesystem, we shouldn't just go mutating the builds.catalyst.coop bucket.

.github/workflows/deploy-pudl.yml
```diff
@@ -2,11 +2,120 @@
 name: deploy-pudl
 on:
   workflow_dispatch:
+    # inputs: deploy target (nightly/stable), staging or not, GCS build location (bucket is mounted to batch job with gcsfuse)
```
jdangerx: No longer mounted with gcsfuse - need to update.

.github/workflows/deploy-pudl.yml
jdangerx: This is a shortcut for debugging the workflow - I didn't want to wait 8 min for the docker image to rebuild every time if all I had changed was `deploy-pudl.yml`. We can take this out or label more explicitly.

.github/workflows/deploy-pudl.yml
jdangerx: Yeah, that's the normal operation - you're right that we'd have to be careful about sending the right build outputs. I guess right now we have to be careful about pushing the `vXXXX` tag to the right ref, but that feels easier. We *could* use the ref to reconstruct the GCS URI (find the most recent build that corresponds to this ref)...

.github/workflows/deploy-pudl.yml
jdangerx: The egress of downloading the bucket contents is like $5 :( so I decided on a Batch job.

devtools/plans/2026-02-09-distribution-refactor-design.md
jdangerx, line 1: Not sure how useful it is to actually include this design doc... IME the agent didn't follow this exactly anyways.

devtools/generate_batch_config.py
jdangerx: Yes, this was reorged because I wanted to just mount the bucket through `gcsfuse` - but then decided against it. It's slower than just accessing GCS normally. The main benefit of `gcsfuse` is "make this bucket look like the local FS." Since we're making a local copy anyways that doesn't really help us.

src/pudl/deployment/deploy_outputs.py
jdangerx: Correct! I also realized this and so made it make a local copy instead, but left an out-of-date misleading comment that showed up directly at the top of this PR 😓

src/pudl/etl/deploy_outputs.py
jdangerx: Sure, I didn't want to add an explicit dependency on `gitpython` for the git stuff, but could probably use `google-cloud-sdk` for the Cloud Run and Storage interactions.

src/pudl/scripts/deploy.py
jdangerx: No, I think this is just a copy-paste error.

src/pudl/scripts/deploy.py
jdangerx: Yes, I think that's all true, though we would have to rename the build paths within `builds.catalyst.coop` since the build ref doesn't make it into the name. And if we have multiple builds off the same ref, we'll have to figure out which one is the "correct" one to distribute.

src/pudl/scripts/deploy.py
jdangerx: Yeah, I think standardizing is a good idea! Happy to use dagster-land.

jdangerx:
@zaneselvans @zschira 

Thanks for taking a look - definitely helpful to get some feedback before things get *too* baked :bread: 

Here's my synthesis of all the feedback - is blocking/non-blocking listed right? Am I missing anything?

### Blocking

* [x] we should derive a bunch of things from the triggering tag:
  * [x] nightly/stable
  * [x] target git branch
  * [x] GCS source
    * [ ] this will require some naming tweaks for the GCS builds to facilitate...
* [x] we should check for ETL success before actually deploying
* [x] we should probably *NOT* allow manual overrides in case there's some sort of error in the above logic
* [x] we should remove debugging stuff / hanging changes from experimentation:
  * [x] docker image specification
  * [x] generate_batch_config.py doesn't need to be changed
  * [x] ignore_regex for trigger_zenodo_deploy
  * [x] bad comments


### Non-blocking
* [ ] use Python clients for git, cloud run, storage
* [ ] Dagsterification of deployment assets

.github/workflows/deploy-pudl.yml
zaneselvans: And I guess what we finally end up sending out to S3 is considerably smaller than what we need to download from the build outputs since it's been compressed, and we don't actually upload everything, so this makes sense.

.github/workflows/deploy-pudl.yml
zaneselvans: At the top level of the `workflow_dispatch` and anything we would be accessing on a day to day basis, I think I'd prefer to minimize the number of parameters and make it as impossible as we can to deploy the wrong outputs, which I think means that everything can just be inferred from the tag that is being deployed.

In the event that we tag a commit which doesn't have successful build outputs, I guess we can either fail, or kick off a build?

The underlying functions could be more flexible, but I just know if we make it possible to specify incorrect combinations of tags, commits, and deployment destinations, we will eventually screw it up.

.github/workflows/deploy-pudl.yml
zaneselvans: I think making it clear that this not meant for production use somehow would be good.

src/pudl/scripts/deploy.py
zaneselvans: Maybe I'm misunderstanding something. How does the git ref play into this?  Looking at our current crop of build outputs, each one has a date-time string, a short commit hash, and what I think is a git ref? Sometimes a branch and sometimes a tag (for releases):

```
gs://builds.catalyst.coop/2026-01-27-0606-5f3305073-main/
gs://builds.catalyst.coop/2026-01-28-0607-e2eda4620-main/
gs://builds.catalyst.coop/2026-01-29-0615-f610f8e9e-main/
gs://builds.catalyst.coop/2026-01-31-0610-446351f93-main/
gs://builds.catalyst.coop/2026-02-03-0617-e671af3a9-main/
gs://builds.catalyst.coop/2026-02-04-0617-3a9204038-main/
gs://builds.catalyst.coop/2026-02-04-1839-2038a7302-update-ferc1-eia-utility-mapping/
gs://builds.catalyst.coop/2026-02-05-0620-28d4f13ec-main/
gs://builds.catalyst.coop/2026-02-05-1849-c97fec2fb-test-quit-if-cleanup-fails/
gs://builds.catalyst.coop/2026-02-05-1849-fee37eaef-hotfix-clean-up-outputs/
gs://builds.catalyst.coop/2026-02-06-0618-997a95750-main/
gs://builds.catalyst.coop/2026-02-07-0612-dac8c8ea0-main/
gs://builds.catalyst.coop/2026-02-09-2239-ded6e0ff7-update-eia860-923-q1-2026/
gs://builds.catalyst.coop/2026-02-11-0625-4677d8d0c-main/
gs://builds.catalyst.coop/2026-02-12-0624-88da11a6b-main/
gs://builds.catalyst.coop/2026-02-12-1401-88da11a6b-v2026.2.0/
gs://builds.catalyst.coop/2026-02-14-0614-139d043d9-main/
gs://builds.catalyst.coop/2026-02-18-0623-a2a9489d5-main/
gs://builds.catalyst.coop/2026-02-19-0623-88c049712-main/
gs://builds.catalyst.coop/2026-02-23-0624-3ba23ca9a-main/
```

I suspect that the issue of having multiple builds on exactly the same commit is going to come up rarely, and if it does come up, it'll probably be the case that only one of them will have a `success` file.  Right now I think it typically only happens when we have a nightly build and a versioned release build on the same commit. But the idea here is that we wouldn't ever do that going forward, right?

zaneselvans:
Things I'm unclear about:
- What additional information we need into the GCS build output paths
- What kind of manual overrides you have in mind (for `workflow_dispatch`? Or deeper in the code?) and the potential for building a footgun

jdangerx:
Some notes from sync call:

* re: additional info in the builds.catalyst.coop bucket: I had erroneously thought that the hash in the path was somehow not a git hash. It is. No problem here.
* re: manual interventions if the code does the wrong thing somehow - probably not worth building this into workflow dispatch trigger. If you need to do something weird or different, just change the dang code.
  * the only downside of this is slower response time if we *do* need to do something weird. but that seems like an acceptable downside, since the upside is that things shouldn't break as much anyways

Adjusted scoping list in comment above.

src/pudl/deployment/__init__.py
```diff
@@ -0,0 +1 @@
+"""Package for deploying PUDL data."""
```
zschira: Now that we have a few deployment specific modules, I thought it would make sense to create a dedicated subdirectory rather than putting it all in `src/etl`

.github/workflows/deploy-pudl.yml
zschira: This only takes `git_tag` and `staging` as an input now. It would be nice if we could have a dropdown of tags, but I don't think there's an easy way to accomplish this, but I've included validation so it should only work with tags that are properly formatted `nightly`/`stable` tags

src/pudl/deploy/pudl.py
```diff
@@ -0,0 +1,324 @@
...
+
+    datapackage = parquet_dir / "pudl_parquet_datapackage.json"
+    if datapackage.exists():
+        shutil.move(str(datapackage), str(output_dir / datapackage.name))
+
+    shutil.rmtree(parquet_dir)
+
+    # Compress SQLite databases
+    sqlite_files = list(output_dir.glob("*.sqlite"))
+    for sqlite_file in sqlite_files:
```
zschira, line 72: I think as an eventual direction, it would be really nice if we produced the zipped `sqlite` / `parquet` files from the main ETL, then didn't have to copy everything over when doing distribution.

src/pudl/deploy/pudl.py
```diff
@@ -0,0 +1,329 @@
...
+    - Creates parquet archive (no compression, already compressed)
+    - Removes test databases and temporary directories
+
+    In general, we want to know if these files don't exist, so
+    FileNotFoundErrors are OK and we don't need to pre-emptively try to avoid
+    them.
+
+    Args:
+        local_path: Path on local filesystem where we will prep outputs for distribution.
+        build_path: Remote path containing raw build outputs.
```
zschira, line 47: I thought it was a bit more clear to pass the `build_path` in here and copy the raw build outputs within this function.

src/pudl/deploy/pudl.py
```diff
@@ -0,0 +1,329 @@
...
+    try:
+        git_ref = _run(["git", "rev-parse", "--short", f"{tag}^{{}}"]).strip()
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(f"Can't find git tag: {tag}") from e
+
+    # Loop through all builds associated with git ref and find most recent one
+    most_recent_build_dt = datetime.min
+    most_recent_build_path = None
+    build_path_pattern = re.compile(r"(\d{4}-\d{2}-\d{2}-\d{4})-([a-z|0-9]{9})-(.+)")
+    for build_path in build_bucket.glob(f"*-{git_ref}-*"):
```
zschira, line 302: This function will always get the most recent build associated with a tag by parsing the datetime from the build directory name. We'd need to change something if we wanted to make it possible to deploy from a build other than the most recent one, but hopefully we can avoid that corner case.

src/pudl/scripts/deploy.py
```diff
@@ -0,0 +1,190 @@
...
+
+def _get_zenodo_release_source_dir(deploy_type: DeploymentType, git_tag: str) -> str:
+    if deploy_type == DeploymentType.NIGHTLY:
+        source_dir = "s3://pudl.catalyst.coop/nightly/"
+    else:
+        source_dir = f"s3://pudl.catalyst.coop/{git_tag}/"
+    return source_dir
+
+
+def _deploy_outputs(
```
zschira, line 82: I consolidated the multiple `deploy_outputs` functions.

src/pudl/scripts/deploy.py
zschira: This will now get done by `get_build_from_tag`

src/pudl/deploy/pudl.py
```diff
@@ -256,3 +271,54 @@ def set_gcs_temporary_hold(gcs_path: str) -> None:
...
+    build_bucket = UPath("gs://builds.catalyst.coop")
+    try:
+        git_ref = _run(["git", "rev-parse", "--short", f"{tag}^{{}}"]).strip()
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(f"Can't find git tag: {tag}") from e
+
+    # Loop through all builds associated with git ref and find most recent one
+    most_recent_build_dt = datetime.min
+    most_recent_build_path = None
+    build_path_pattern = re.compile(r"(\d{4}-\d{2}-\d{2}-\d{4})-([a-z|0-9]{9})-(.+)")
```
zaneselvans: A total nit, but aren't the hashes hexidecimal? In which case this regex would use `[a-f|0-9]`?

src/pudl/deploy/pudl.py
zaneselvans: Like, prepare all of the files we ultimately want to push to **anywhere** and have them stashed in the build output path? That would be much simpler. Probably almost double our storage costs for the bucket, but it would be nice to just have them all ready to go off the shelf.

src/pudl/deployment/__init__.py
zaneselvans: Agree that if we're going to have several different deployment modules for different jobs or destinations, it would be nice to organize them in a subpackage here. Can we standardize the naming though?  If they're already under `deployement` then maybe the modules don't need to say "deploy" or "deployment" at all and can just be the name of the thing being deployed... or if we have helper modules that contain logic for deploying to a particular target, use the name of the target?

Also, I feel like we're using "deploy" and "distribute" interchangeably in different contexts.  To me "deploy" has the connotation of a software/service (like we would deploy a new version of the PUDL Data Viewer) while "distribute" means "put static output somewhere" (like a versioned release to Zenodo, or nightlies to S3). And I think here we're only talking about data distribution? Am I the only one with these connotations?

So might it look like:

```
# datasets with their own distinct distribution workflows
src/pudl/distribute/ferceqr.py
src/pudl/distribute/pudl.py

# distribution targets, which have their own potentially reusable logic
# in case we are distributing different datasets to the same destinations
src/pudl/distribute/upath.py
src/pudl/distribute/zenodo.py
src/pudl/distribute/huggingface.py
```

Having another intuitive verb at the top level of the PUDL module tree also feels good. `extract`, `transform`, `distribute`, `convert`

src/pudl/deploy/pudl.py
```diff
@@ -63,29 +79,29 @@ def prepare_outputs_for_distribution(output_dir: Path) -> None:
...
-    archive_path = output_dir / "pudl_parquet.zip"
+    parquet_files = list(local_path.glob("*.parquet"))
+    assert len(parquet_files) > 0, f"No parquet files in {local_path}."
+    archive_path = local_path / "pudl_parquet.zip"
     with zipfile.ZipFile(archive_path, "w", zipfile.ZIP_STORED) as zf:
         for parquet_file in parquet_files:
             zf.write(parquet_file, arcname=parquet_file.name)
 
-        datapackage = output_dir / "pudl_parquet_datapackage.json"
+        datapackage = local_path / "pudl_parquet_datapackage.json"
```
zaneselvans, line 89: In messing around with the agents, I realized that having this exhaustive datapackage publicily available in a queryable location would be really helpful for machine navigation of the metadata, for folks that don't have the repo cloned, so if we should be sure we keep distributing it and don't zip it up.

src/pudl/scripts/deploy.py
```diff
@@ -1,112 +1,118 @@
...
+use the git tag to identify builds associated with the tag, and determine whether
+the deployment is intended to be a nightly or stable deployment. It expects nightly
+deployments to have tags conforming to the pattern 'nightly-YYYY-MM-DD' and stable
+deployments 'vYYYY.M.D'.
 
 Examples:
     Deploy nightly build to production:
-        pudl_deploy nightly gs://builds.catalyst.coop/2025-02-05-1234-abc123-main
-            --git-tag nightly-2025-02-05
+        pudl_deploy nightly-2025-02-05
```
zaneselvans, line 14: I love how concise this command is! If these are ultimately Dagster assets then we can probably get away without even needing a script, and maybe it looks something like...

```console
pixi run dg launch --job pudl_deploy --config-json {"git_tag": "nightly-2025-02-05}
```

which maybe we wrap in a [templated pixi task](https://pixi.prefix.dev/latest/workspace/advanced_tasks/#minijinja-templating-for-task-arguments) and it ends up being

```console
pixi run pudl-deploy nightly-2025-02-05
```

src/pudl/deployment/__init__.py
zschira: I've done a first round of renames, although I think there's more refactor that might be best in a follow up pr. I stuck with the name `deploy`, since there's slightly more than just copying files (like updating the git branch), but I don't feel too strongly about this and I see the case for `distribute` as well.

Right now I've created:

- `src/pudl/deploy/pudl.py`
- `src/pudl/deploy/ferceqr.py`

Right now these modules do very different things with the `ferceqr` one having some dagster automation stuff + assets for handling deployment, while `pudl` has mostly utility functions, which get used by `src/pudl/scripts/deploy.py`. I think there's several followup steps that are probably better for separate PR(s) that would make these more consistent. First, I think there's some shared functionality that could be refactored out like moving local outputs to s3/gcs. I think it would make sense to split this functionality out into separate modules like `upath.py` / `zenodo.py`. The second change would be standardizing the high level deployment orchestration. I think making both `pudl.py` and `ferceqr.py` define dagster jobs that use shared functionality from other modules within the `deploy` package would make sense.

src/pudl/deploy/pudl.py
zschira: Yeah true about the storage requirements. Something to consider for the future, but definitely not right now

src/pudl/deploy/pudl.py
zschira: Good call, fixed.

src/pudl/deploy/pudl.py
```diff
@@ -0,0 +1,329 @@
...
+
+    # Check that we found a build
+    if most_recent_build_path is None:
+        raise RuntimeError(
+            f"Can't find a build associated with tag: {tag}, ref: {git_ref}"
+        )
+    logger.info(
+        f"Most recent build associated with tag {tag}: {most_recent_build_path.as_uri()}"
+    )
+    return check_build_success(build_path)
```
zaneselvans: 🛑 Blocking: I was getting a unit test failure for this function in the `True-True` case but it wasn't happening in the CI on GitHub, and looking at this it seems like maybe we're supposed to be returning `most_recent_build_path`? Maybe they were just showing up in a different order in MacOS vs. Linux filesystems?

```suggestion
    return check_build_success(most_recent_build_path)
```

src/pudl/deploy/pudl.py
```diff
@@ -0,0 +1,329 @@
...
+        git_ref = _run(["git", "rev-parse", "--short", f"{tag}^{{}}"]).strip()
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(f"Can't find git tag: {tag}") from e
+
+    # Loop through all builds associated with git ref and find most recent one
+    most_recent_build_dt = datetime.min
+    most_recent_build_path = None
+    build_path_pattern = re.compile(r"(\d{4}-\d{2}-\d{2}-\d{4})-([a-f|0-9]{9})-(.+)")
+    for build_path in build_bucket.glob(f"*-{git_ref}-*"):
+        match = build_path_pattern.search(str(build_path))
```
zaneselvans: 🛑 Blocking: let's explicitly define the behavior `if match is None`? Should it just `continue` or `raise` something? Seems like that's indicative of unexpected stuff hanging out in the builds bucket, so probably an error?

test/unit/deploy/test_deploy_pudl.py
```diff
@@ -0,0 +1,208 @@
...
+                call(["git", "checkout", "nightly"], **kwargs),
+                call(["git", "merge", "--ff-only", "nightly-2026-02-09"], **kwargs),
+            ]
+        )
+
+
+@pytest.mark.parametrize(
+    "create_builds,build_successful",
+    [
+        (True, True),
```
zaneselvans, line 160: 🛑 Blocking: The `True-True` case was failing for me locally, but not in GitHub CI. Returning the most recent build (see above) fixed it. I think that's what we actually want to do?

```
>               assert get_build_from_tag(example_tag) == expected_path
E               AssertionError: assert PosixPath('/private/var/folders/ps/6jyqvztj5fq_tvhwx5h59d7w0000z8/T/pytest-of-zane/pytest-154/test_get_build_from_tag_True_T0/2026-02-04-0530-401c66a99-main') == PosixPath('/private/var/folders/ps/6jyqvztj5fq_tvhwx5h59d7w0000z8/T/pytest-of-zane/pytest-154/test_get_build_from_tag_True_T0/2026-02-04-1230-401c66a99-main')
E                +  where PosixPath('/private/var/folders/ps/6jyqvztj5fq_tvhwx5h59d7w0000z8/T/pytest-of-zane/pytest-154/test_get_build_from_tag_True_T0/2026-02-04-0530-401c66a99-main') = get_build_from_tag('example_tag')
```

src/pudl/deploy/pudl.py
```diff
@@ -0,0 +1,329 @@
...
+            "update",
+            f"{gcs_path}*",
+            "--temporary-hold",
+        ]
+    )
+
+    logger.info(f"Temporary hold set on {gcs_path}")
+
+
+def check_build_success(build_path: UPath):
```
zaneselvans: This name reads like a function that will return a `bool` so I think it's worth annotating.

```suggestion
def check_build_success(build_path: UPath) -> UPath:
```

src/pudl/scripts/deploy.py
```diff
@@ -0,0 +1,190 @@
...
+    source_dir: Path,
+    deploy_type: DeploymentType,
+    git_tag: str,
+    staging: bool,
+    github_token: str,
+):
+    """Execute stable deployment workflow.
+
+    Deploys to versioned and stable paths, updates stable branch, sets GCS temporary
+    hold on versioned release, and triggers production Zenodo release (unpublished).
```
zaneselvans: 🛑 Blocking: It looks like this function actually does both stable & nightly outputs, and does publish the nightly outputs on Zenodo, in which case we should update the docstring to reflect that.

src/pudl/scripts/deploy.py
```diff
@@ -0,0 +1,190 @@
...
+This CLI orchestrates deployment of completed PUDL ETL builds to public cloud
+storage (GCS and S3), git branch updates, Zenodo releases, and Cloud Run deployments.
+
+The script takes a git tag, and an optional staging flag for safe testing. It will
+use the git tag to identify builds associated with the tag, and determine whether
+the deployment is intended to be a nightly or stable deployment. It expects nightly
+deployments to have tags conforming to the pattern 'nightly-YYYY-MM-DD' and stable
+deployments 'vYYYY.M.D'.
+
+Examples:
```
zaneselvans, line 12: 🛑 Blocking: We should update these examples to include the `GITHUB_TOKEN` if it's a required argument.

test/unit/deploy/test_deploy_pudl.py
zschira: Yeah that definitely should've been returning `most_recent_build_path`. I guess the path order on linux was randomly hiding the error.

src/pudl/scripts/deploy.py
zschira: Ah yeah forgot to update the docstring after consolidating these methods. Fixed.

src/pudl/scripts/deploy.py
zschira: True. I considered making `GITHUB_TOKEN` an environment variable, but it feels a bit premature if we might still change the orchestration to use dagster or something.

src/pudl/deploy/pudl.py
zschira: At first I thought about just logging a warning, but decided to elevate this to an error because it should be a very unlikely scenario and we probably want to guarantee some human intervention if it ever happens.

src/pudl/scripts/deploy.py
zaneselvans: It would be great if we can avoid needing to specify the token on the CLI if the user is already authenticated (since I imagine we'll want to use it locally too) And also it's easy to mess up and have it get printed in the logs somewhere when we're running it on GH or in the builds. But for the moment I guess we can leave it.

src/pudl/deployment/__init__.py
zaneselvans: This is a great start, and I agree on the direction for future PRs. If you want to write up a couple of issues reflecting that direction that would be great:
- Factor our deployment-target specific functionality into shared modules. (`deploy/upath.py` and `deploy/zenodo.py` initially) so we'll have both `deploy/{target}.py` and `deploy/{data_product}.py` modules in here.  Would it make sense to pull existing functionality into a `deploy/eel_hole.py` module too? For wiring up to dagster in the future?
- dagsterize the PUDL deployment along the lines of the FERC EQR deployment.

zaneselvans:
@zschira also:
- Update the release notes
- Delete the agent planning documents, unless we have a strong reason to keep them around.

docs/dev/nightly_data_builds.rst
```diff
@@ -69,6 +69,20 @@ Platform (GCP) permissions to start and update the GCE instance. We use Workflow
...
+takes a git tag as an input and will find build outputs, and determine the deployment
+type (stable or nightly) from the tag. It will then upload outputs from the build to
+GCS and S3, update the git branch associated with the deployment type, and trigger
+a zenodo release. This action can also take an optional ``staging`` flag will upload
+outputs to a dedicated staging area, and will not update the git branch or trigger a
+zenodo release.
+
+Eventually, the deployment functionality will be removed from the ``build-deploy-pudl``
+action and it will instead push a tag after a successful build, which will trigger this
+workflow.
```
zaneselvans: This tag-push kickoff will work for the versioned releases, but for the nightly builds we push the tag at the beginning of the build, to mark what commit we *tried* to build that night, even if it doesn't succeed.

docs/release_notes.rst
```diff
@@ -10,6 +10,10 @@ v2026.4.0 (unreleased)
 
 Enhancements
 ^^^^^^^^^^^^
+* Added a new standalone data deployment workflow, ``deploy-pudl.yml``. This is
+  still in testing, but will allow us to separate deployment from builds, enabling
+  deployment from an existing build and creating more modular and reusable
+  infrastructure.
```
zaneselvans: Always gotta reference the issues and PRs in here.
```suggestion
* Added a new standalone data deployment workflow, ``deploy-pudl.yml``. This is
  still in testing, but will allow us to separate deployment from builds, enabling
  deployment from an existing build and creating more modular and reusable
  infrastructure. See issue :issue`5003` and PR :pr:`5016`.
```

src/pudl/scripts/deploy.py
```diff
@@ -0,0 +1,192 @@
...
+@click.argument(
+    "git-tag",
+    type=str,
+)
+@click.option(
+    "--github-token",
+    type=str,
+    required=True,
+    help="GitHub token for authentication.",
+)
```
zaneselvans: We'll need to remove this guy right?  Or could make it optional for local CLI usage if you don't happen to call your local token envvar `GITHUB_TOKEN`

docs/dev/nightly_data_builds.rst
zschira: Fixed to just mention that it will trigger the action after a successful build.

src/pudl/scripts/deploy.py
zschira: I went ahead and removed it for now, but we could add it back as optional in the future if that becomes useful. I'm guessing we'll mostly want to run this through the action though as you'd otherwise have to copy all of the outputs to your local machine.

zschira:
I've created a new epic (#5113) to track the remaining steps for fully decoupling builds from deployment. I also ran a successful [deployment job](https://console.cloud.google.com/batch/jobsDetail/regions/us-east1/jobs/deploy-outputs-23358000007-1/logs?project=catalyst-cooperative-pudl) earlier using this new setup. Going to merge now so we can move to next steps.

Issue #4888, opened 2025-12-31
Validate USDA RUS Form 12: Power Supply Tables


aesharpe:
### `core_rus12__yearly_meeting_and_board` ✔️ 
- [x] report date > last annual meeting date
- [x] members present + voting < total members
- [x] numeric columns are positive

### `core_rus12__yearly_long_term_debt` ✔️ 
- [x] Expect values are positive
- [x] Expect principal + interest sum to current year total

### `core_rus12__yearly_lines_stations_labor_materials_cost` ✔️
- [x] Cost >0

### `core_rus12__yearly_loans` ✔️
- [x] Loan original amount >= loan balance
- [x] Loan maturity date >= reporting date
- [x] Numeric values are positive

### `core_rus12__yearly_plant_labor` ✔️ 
- [x] Numeric values are positive
- [x] If time, compare hours to employees

### `core_rus12__yearly_renewable_plants` ✔️ 
- [x] Numeric values are positive
- [x] Percentages are between 0-100
- [x] Amount invested is > RUS funding

### `core_rus12__yearly_sources_and_distribution` ✔️ 
- [x] `net_energy_received_mwh` is positive
- [x] cost is positive where energy is received

### `core_rus12__yearly_sources_and_distribution_by_plant_type` ✔️ 
- [x] `net_energy_received_mwh` is positive
- [x] cost is positive
- [x] plant_num is positive
- [x] capacity is positive

### `core_rus12__yearly_statement_of_operations` 🟡 
- [ ] TODO: Check that opex spending is positive - about 3% of the data violates this across records, investigate more?
- [ ] TODO: Check that monthly spending is ~1/12 of the YTD budget (5-20% range) - 11% of the data violates this constraint, is this still useful to check?
- [ ] TODO: Check that budgeted spending is roughly equal to the YTD budget (75-125% range) - 12% of the data violates this constraint, is this still useful to check?
- [ ] TODO: Any other constraints here that are clear to check?

### `core_rus12__yearly_investments` ✔️ 
- Both included and excluded investments and the income_or_loss column often contain negative values. This seems sensible, and is not a validation problem. No additional validations added here.

e-belfer:
Validations are partially but not fully complete.

PR #5017, opened 2026-02-17
Validation first pass for RUS Form 7 and 12 tables

### LINKED ISSUE #4887: Validate USDA RUS Form 7: Distribution Tables

e-belfer:
# Overview

Closes #4887 and #4888 

- Add a suite of validation tests for the RUS Form 7 and 12 tables
- Fix dtype issues identified during validation
- Fix other column issues identified during validation

## What did you change?
- Wrote DBT models
- Wrote generic DBT tests
- Fixed field metadata and schemas as needed, which involved recreating the migrations from scratch

## Out of scope
- Name harvesting
- Foreign key checks for the RUS IDs
- Intertable value checks (e.g., is information consistent across tables when reported in more than one place)?
- Totals validations against sub-components

## Documentation

# Testing

Recreate your DB and update the data. Then, you can run `dbt_helper validate --asset-select group:core_rus12` (and the same for rus7) to run all these checks against the updated tables.

## To-do list

### DIFF
--- a/dbt/models/rus12/core_rus12__yearly_statement_of_operations/schema.yml
+++ b/dbt/models/rus12/core_rus12__yearly_statement_of_operations/schema.yml
@@ -34,20 +34,23 @@ sources:
               - dbt_expectations.expect_column_min_to_be_between:
                   arguments:
                     min_value: 0.0
+                  row_condition: "opex_type != 'interest_charged_to_construction_credit'"
                   config:
-                    error_if: ">650" # TODO: ~3.7% of data violates this. Worth checking?
+                    error_if: ">307" # TODO: ~1.8% of data violates this. Worth checking?
           - name: opex_ytd
             data_tests:
               - dbt_expectations.expect_column_min_to_be_between:
                   arguments:
                     min_value: 0.0
+                  row_condition: "opex_type != 'interest_charged_to_construction_credit'"
                   config:
-                    error_if: ">560" # TODO: ~3.2% of data violates this. Worth checking?
+                    error_if: ">135" # TODO: ~0.7% of data violates this. Worth checking?
           - name: opex_ytd_budget
             data_tests:
               - dbt_expectations.expect_column_min_to_be_between:
                   arguments:
                     min_value: 0.0
+                  row_condition: "opex_type != 'interest_charged_to_construction_credit'"
                   config:
-                    error_if: ">320" # TODO: ~1.8% of data violates this. Worth checking?
+                    error_if: ">43" # TODO: ~0.03% of data violates this. Worth checking?
           - name: is_total
--- a/dbt/models/rus7/core_rus7__yearly_statement_of_operations/schema.yml
+++ b/dbt/models/rus7/core_rus7__yearly_statement_of_operations/schema.yml
@@ -30,18 +30,27 @@ sources:
           - name: opex_type
           - name: opex_group
           - name: opex_report_month
-            # data_tests:
-            #   - dbt_expectations.expect_column_min_to_be_between:
-            #       arguments:
-            #         min_value: 0.0 # TODO: This is widely violated. Keep or adjust expectation?
+            data_tests:
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
+                  config:
+                    error_if: ">6750" # TODO: About 3.4% of the data violates this, even with the broad categorical exclusions below. Investigate or refine further.
+                  row_condition: "opex_type != 'interest_charged_to_construction_credit' and opex_group != 'patronage_capital_or_margins'"
           - name: opex_ytd
-            # data_tests:
-            #   - dbt_expectations.expect_column_min_to_be_between:
-            #       arguments:
-            #         min_value: 0.0 # TODO: This is widely violated. Keep or adjust expectation?
+            data_tests:
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
+                  config:
+                    error_if: ">1600" # TODO: About 0.8% of the data violates this, even with the broad categorical exclusions below. Investigate or refine further.
+                  row_condition: "opex_type != 'interest_charged_to_construction_credit' and opex_group != 'patronage_capital_or_margins'"
           - name: opex_ytd_budget
-            # data_tests:
-            #     - dbt_expectations.expect_column_min_to_be_between:
-            #         arguments:
-            #           min_value: 0.0 # TODO: This is widely violated. Keep or adjust expectation?
+            data_tests:
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
+                  config:
+                    error_if: ">810" # TODO: About 0.4% of the data violates this. Investigate or refine further the broad categorical exclusions below.
+                  row_condition: "opex_type != 'interest_charged_to_construction_credit' and opex_group != 'patronage_capital_or_margins'"
           - name: is_total


dbt/macros/expect_column_ratio.sql
```diff
@@ -0,0 +1,52 @@
...
+
+  {% set column_list = adapter.get_columns_in_relation(model) | map(attribute="name") | list %}
+
+  WITH ratio_check AS (
+    SELECT
+      *,
+      CASE
+        WHEN {{ denominator_column }} IS NULL OR {{ denominator_column }} = 0
+        THEN NULL
+        ELSE {{ numerator_column }} / NULLIF({{ denominator_column }}, 0)
```
cmgosnell: total nit but does this `NULLIF` need to be here if you've already `{{ denominator_column }} = 0` above?

dbt/macros/expect_columns_ratio.sql
cmgosnell, line 1: 1000% nit but i think this would be more clear if it was `expect_columns_ratio`

dbt/models/rus7/core_rus7__yearly_balance_sheet_assets/schema.yml
cmgosnell, line 1: should we add a threshold or a `expect_column_min_to_be_between` for `balance`? or i suppose balance is just probably expected to be negative sometimes. (same question for the liabilities).

This should be totally oos but I also wonder if the finance peeps at RMI could give us some insight into expected proportional values for particular `asset_type`'s compared to others (like whatever `utility_plant_in_service` shouldn't normally be more than X% of `total`

dbt/models/rus7/core_rus7__yearly_energy_efficiency/schema.yml
```diff
@@ -3,12 +3,30 @@ sources:
...
+                  arguments:
+                    min_value: 0.0
           - name: savings_mmbtu
+            data_tests:
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
+                  config:
+                    error_if: ">8"
+                  description: "KS0059 has negative values which don't follow from prior years of data." #TODO
```
cmgosnell, line 32: this is the column that probably imu has unit errors embedded in there.

dbt/macros/expect_summed_columns_not_exceed_threshold.sql
cmgosnell, line 1: total naming nit that you can take or leave but i think this would be easier to understand as `expect_summed_columns_not_exceed_threshold`

cmgosnell (commented):
this is looking amazing. took me a little longer than i'd desire to parse through the sql but that's my level of sql not what's going on in there. I see your todo list in the issue and i think this all looks like its moving in the right direction.

On a meta level, it seems we should compile all of the investigate/TODO's and organize them in rough priority and share them with RMI/RPC. Many of them imho seems super low priority, just nice to know.

dbt/models/rus7/core_rus7__yearly_balance_sheet_assets/schema.yml
e-belfer: Absolutely - I need to fix the column names and dtypes so I haven't actually finished validating this table yet.

dbt/models/rus7/core_rus7__yearly_energy_efficiency/schema.yml
e-belfer: Could check this against total energy demand reported in another table.

dbt/macros/expect_column_ratio.sql
e-belfer: No not necessarily, it's just a safety guard that can get removed.

dbt/models/rus7/core_rus7__yearly_balance_sheet_assets/schema.yml
e-belfer: Open question as to what the useful validations would be here, as this is very often negative.

dbt/models/rus12/core_rus12__yearly_sources_and_distribution_by_plant_type/schema.yml
```diff
@@ -11,6 +11,24 @@ sources:
...
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
           - name: net_energy_received_mwh
+            data_tests:
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
+                  config:
+                    error_if: ">24" #TODO: Almost entirely explained by GA0109 (hydro plant)
```
cmgosnell, line 34: non-blocking but maybe we should exclude plant_type hydro from this?

dbt/models/rus12/core_rus12__yearly_statement_of_operations/schema.yml
```diff
@@ -3,13 +3,51 @@ sources:
...
           - name: borrower_name_rus
           - name: opex_type
           - name: opex_group
           - name: opex_report_month
+            data_tests:
+              - dbt_expectations.expect_column_min_to_be_between:
+                  arguments:
+                    min_value: 0.0
+                  config:
+                    error_if: ">650" # TODO: ~3.7% of data violates this. Worth checking?
```
cmgosnell: in my general understanding it is possible - albeit rare - to have negative opex values. and there are clearly patterns here for which fields are negative:

Image

Image

cmgosnell: it seems like at the very least we should add this row condition to all of these checks and reduce the # of exceptions:
`row_condition: "opex_type != 'interest_charged_to_construction_credit'"`


Image

dbt/models/rus7/core_rus7__yearly_statement_of_operations/schema.yml
cmgosnell, line 1: oof this table looks confusing to test to say the least.

looking at `opex_ytd` here in the same fashion as i was for the rus12 version of this table - the # of negative records: 
Image
and the portion of negative records:

Image

If we want to keep these expect not negative tests, then we should almost certainly remove `opex_group IN ("patronage_capital_or_margins", "patronage_and_operating_margins")` and it seems also `opex_type != "interest_charged_to_construction"` but even for the `opex_ytd` that would still leave 659 negative values :-/ _only_ 149 for `opex_ytd_budget`

it might just be that this table requires some deeper documentation or we should ask deeper subject experts what are more nuanced expectations of this table

dbt/models/rus12/core_rus12__yearly_sources_and_distribution_by_plant_type/schema.yml
e-belfer: The vast majority of hydro data (at least from the 25th percentile) is positive, so I think the flag here is weird plant rather than weird plant type.

dbt/models/rus12/core_rus12__yearly_statement_of_operations/schema.yml
e-belfer: For all three numeric fields the 75th percentile of this opex type is negative or 0, so I agree with this and have excluded it from the test.

dbt/models/rus7/core_rus7__yearly_statement_of_operations/schema.yml
e-belfer: Looking at the percentiles of negative data, I'm seeing that ~23% of patronage_capital_or_margins data is negative, and ~55% of the interest charged to construction. For now, I'm going to adopt your suggestion but I'll leave a to do in here to look into this further.

Issue #4886, opened 2025-12-31
Integrate USDA Form 12: Power Supply Remaining Tables


aesharpe:
There are 19 remaining Form 12 tables and 8 remaining total tables. Note that we still need to decide how to handle the totals tables. 
 
- [ ] **DA - OpRpt_PSOperatingReportPlantSteamUnits.csv**: detailed operating data for fossil fuel steam plants: boiler/turbine performance, fuel consumption (coal, oil, gas), operating hours, generation statistics.
- [ ] **DD - OpRpt_PSOperatingReportPlantSteamCost.csv**: same but cost.
- [ ] **DG - C_OpRpt_PSOperatingReportPlantFactorsAndMaxDemand.csv**: plant max demand.
- [ ] **EA - OpRpt_PSOperatingReportPlantHydroUnits.csv**: hydroelectric facility operations: generating unit performance, hydraulic data (pool/tail race elevations), water spilled, labor.
- [ ] **ED - OpRpt_PSOperatingReportPlantHydroCost.csv**: same but production cost.
- [ ] **FCCA - OpRpt_PSOperatingReportPlantCombinedCycleUnits.csv**: combined cycle facility operations: similar to steam and IC plants but for combined cycle technology, including fuel use, generation, efficiency.
- [ ] **FCCD - OpRpt_PSOperatingReportPlantCombinedCycleCost.csv**: same but cost.
- [ ] **FICA - OpRpt_PSOperatingReportPlantInternalCombustionUnits.csv**: internal combustion engine generator data: fuel consumption, operating hours, generation, BTU efficiency, labor.
- [ ] **FICD - OpRpt_PSOperatingReportPlantInternalCombustionCost.csv**: same but cost
- [ ] **GA - OpRpt_PSOperatingReportPlantNuclearUnits.csv**: nuclear power plant operations: reactor unit performance, operating hours, scheduled/unscheduled outages, fuel expenses, maintenance, labor.
- [ ] **GD - OpRpt_PSOperatingReportPlantNuclearCost.csv**: same but cost.
- [ ] **HA - OpRpt_PSASUtilityPlant.csv**: utility plant additions and retirements.
- [ ] **HB - OpRpt_PSASUtilityPlantAccumulatedProvisionForDeprAndAmort.csv**:  accumulated depreciation and amortization.
- [ ] **HC - OpRpt_PSNonUtilityPlant.csv**: non-utility property.
[...]

PR #5018, opened 2026-02-17
Transform next tranche of RUS 12 tables

cmgosnell:
# Overview

Working on #4886.

## Documentation

# Testing

## To-do list

### DIFF
--- a/docs/templates/access_examples/core_epacems__hourly_emissions.rst.jinja
+++ b/docs/templates/access_examples/core_epacems__hourly_emissions.rst.jinja
@@ -1,4 +1,4 @@
-* `Browse and query this table online <https://data.catalyst.coop/search?&q=name:{{ resource.name }}>`__
+* `Browse and query this table online <https://data.catalyst.coop/preview/pudl/{{ resource.name }}>`__
 * `Download this table as a Parquet file <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/{{ resource.name }}.parquet>`__
 
 
--- a/docs/templates/access_examples/core_ferceqr__contracts.rst.jinja
+++ b/docs/templates/access_examples/core_ferceqr__contracts.rst.jinja
@@ -1,3 +1,4 @@
+* `Browse and query this table online <https://data.catalyst.coop/preview/pudl/{{ resource.name }}>`__
 
 .. note::
 
--- a/docs/templates/access_examples/core_ferceqr__quarterly_identity.rst.jinja
+++ b/docs/templates/access_examples/core_ferceqr__quarterly_identity.rst.jinja
@@ -1,3 +1,4 @@
+* `Browse and query this table online <https://data.catalyst.coop/preview/pudl/{{ resource.name }}>`__
 
 .. note::
 
--- a/docs/templates/access_examples/core_ferceqr__quarterly_index_pub.rst.jinja
+++ b/docs/templates/access_examples/core_ferceqr__quarterly_index_pub.rst.jinja
@@ -1,3 +1,4 @@
+* `Browse and query this table online <https://data.catalyst.coop/preview/pudl/{{ resource.name }}>`__
 
 .. note::
 
--- a/docs/templates/access_examples/core_ferceqr__transactions.rst.jinja
+++ b/docs/templates/access_examples/core_ferceqr__transactions.rst.jinja
@@ -1,3 +1,4 @@
+* `Browse and query this table online <https://data.catalyst.coop/preview/pudl/{{ resource.name }}>`__
 
 .. note::
 
--- a/docs/templates/access_examples/default.rst.jinja
+++ b/docs/templates/access_examples/default.rst.jinja
@@ -1,4 +1,4 @@
-* `Browse and query this table online <https://data.catalyst.coop/search?&q=name:{{ resource.name }}>`__
+* `Browse and query this table online <https://data.catalyst.coop/preview/pudl/{{ resource.name }}>`__
 * `Download this table as a Parquet file <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/{{ resource.name }}.parquet>`__
 
 .. tabs::
--- a/docs/templates/access_examples/out_vcerare__hourly_available_capacity_factor.rst.jinja
+++ b/docs/templates/access_examples/out_vcerare__hourly_available_capacity_factor.rst.jinja
@@ -1,4 +1,4 @@
-* `Browse and query this table online <https://data.catalyst.coop/search?&q=name:{{ resource.name }}>`__
+* `Browse and query this table online <https://data.catalyst.coop/preview/pudl/{{ resource.name }}>`__
 * `Download this table as a Parquet file <https://s3.us-west-2.amazonaws.com/pudl.catalyst.coop/nightly/{{ resource.name }}.parquet>`__
 
 .. warning::
--- a/docs/templates/data_source_parent.rst.jinja
+++ b/docs/templates/data_source_parent.rst.jinja
@@ -42,7 +42,7 @@ the names and descriptions of each of its fields.
    {%- for resource in source_resources %}
    * - :ref:`{{ resource.sphinx_ref_name }}`
    {%- block browse_online scoped %}
-     - https://data.catalyst.coop/search?q=name:{{ resource.name }}
+     - https://data.catalyst.coop/preview/pudl/{{ resource.name }}
    {% endblock %}
    {%- endfor %}
 
@@ -63,7 +63,7 @@ existing tables.
    {%- for resource in extra_resources %}
    * - :ref:`{{ resource.sphinx_ref_name }}`
    {%- block browse_extra_online scoped %}
-     - https://data.catalyst.coop/search?q=name:{{ resource.name }}
+     - https://data.catalyst.coop/preview/pudl/{{ resource.name }}
    {% endblock %}
    {%- endfor %}
 {%- endif %}


src/pudl/metadata/resources/rus12.py
```diff
@@ -376,3 +376,143 @@
...
+        "description": {
+            "additional_summary_text": (
+                "plant operational data including fuel consumption and operational hours by unit."
+            ),
+            "usage_warnings": ["experimental_wip"],
+            "additional_source_text": "(Part )",
+            "additional_primary_key_text": (
+                "This table has no primary key because there are a handful of plants that "
+                "have duplicate records. The primary key of this table "
+                "otherwise would be: [`report_date`, `borrower_id_rus`, `plant_name_rus`, `plant_name_rus`, `unit_id_rus`, `plant_type`,`is_fully_owned_by_borrower`]."
```
cmgosnell: bummer. and they really all have different values :-/
```python
df = defs.load_asset_value("core_rus12__yearly_plant_operations")
pk = ["report_date", "borrower_id_rus", "plant_name_rus", "plant_name_rus", "unit_id_rus", "plant_type","is_fully_owned_by_borrower"]
print(len(df_all[df_all.duplicated(pk,keep=False)].set_index(pk).sort_index()))
df_all[df_all.duplicated(pk,keep=False)].plant_name_rus.value_counts()
```
```
52
plant_name_rus
Apache         32
Limon          12
Deer Creek      6
St. Francis     2
Name: count, dtype: int64
```

src/pudl/metadata/resources/rus12.py
```diff
@@ -376,3 +376,143 @@
...
+                "borrower_id_rus",
+                "borrower_name_rus",
+                "plant_name_rus",
+                "unit_id_rus",
+                "plant_type",
+                "capacity_kw",
+                "gross_generation_mwh",
+                "borrower_ownership_pct",
+                "is_fully_owned_by_borrower",
+                "is_partly_owned_by_borrower",  # was not reported till 2009
```
cmgosnell: not a huge deal, but this is not what I would have expected overall... I would have expected them 
to be if fully = True then partly = False and vice versa

```python
df = defs.load_asset_value("core_rus12__yearly_plant_operations")
df[["is_fully_owned_by_borrower", "is_partly_owned_by_borrower"]].value_counts(dropna=False).sort_index()
```

```
is_fully_owned_by_borrower  is_partly_owned_by_borrower
False                       NaN                             211
                            False                             1
                            True                            967
True                        NaN                            1111
                            False                          4290
                            True                            969
Name: count, dtype: int64
```

src/pudl/metadata/resources/rus12.py
```diff
@@ -376,3 +376,143 @@
...
+                "fuel_consumption_coal_1000_lbs",  # TODO convert these units? or just rename to klbs, ? and kgal
+                "fuel_consumption_gas_1000_cubic_feet",
+                "fuel_consumption_oil_1000_gals",
```
cmgosnell: @e-belfer  thoughts on whether to convert these units?

src/pudl/metadata/resources/rus12.py
```diff
@@ -376,3 +376,143 @@
...
 }
+
+DRAFT_RESOURCE_METADATA = {
+    "core_rus12__yearly_plant_costs": {
+        "description": {
+            "additional_summary_text": ("costs of net energy generated by plant."),
+            "usage_warnings": ["experimental_wip"],
+            "additional_source_text": "(Part F - Section D)",
+            "additional_primary_key_text": (
+                "This table has no primary key because there is one plant (named Walter "
```
e-belfer: @cmgosnell We should check in about this plant (which possibly has two borrowers)? @aesharpe and I had a long conversation about it in #4979 , and whatever we decide we should be consistent about how we treat this across our tables.

src/pudl/metadata/resources/rus12.py
e-belfer: I think it would be generally consistent with our effort to purge thousands units (e.g., #4301).

src/pudl/transform/rus12.py
```diff
@@ -324,3 +324,115 @@ def core_rus12__yearly_statement_of_operations(raw_rus12__statement_of_operation
...
+    # this total flag is a little different than the others. it grabs the high-level
+    # cost group totals and it flags the cost_type's of net and less. Because these
+    # are clearly calculated fields from looking at the form.
+    df["is_total"] = df.cost_type.str.contains("total|net|less") | (
+        df.cost_group == "total"
+    )
+    # TODO: settle on names and do this rename in the column maps
+    df_all = df_all.rename(
+        columns={
+            "amount": "cost",
```
e-belfer: Elsewhere we are using cost, this seems broadly consistent.

src/pudl/metadata/resources/rus12.py
e-belfer: Image

This has to do with some very weird data structure which is also affecting the PKs for this table, and we should think about how to deal with this. Perhaps we actually want to split these tables into _by_plant and _by_borrower based on these categoricals.

src/pudl/metadata/resources/rus12.py
```diff
@@ -376,3 +376,143 @@
...
+            "fields": ["report_date", "borrower_id_rus", "borrower_name_rus"],
+            "primary_key": ["report_date", "borrower_id_rus"],
+        },
+        "sources": ["rus12"],
+        "etl_group": "rus12",
+        "field_namespace": "rus",
+    },
+}
+
+plant_cost_type_enum = {
```
e-belfer: Will want to move this to enums.py once all is migrated to alembic.

src/pudl/metadata/resources/rus12.py
cmgosnell: we flagged this plant with RMI/RPC for possible cleaning/validation and landed on putting fixing these records at a lower priority.

docs/templates/rus12_child.rst.jinja
```diff
@@ -29,6 +29,10 @@ What data is available through PUDL?
 {% block availability %}
 We've integrated all RUS 12 tables with the exception of eight totals
 tables containing aggregated versions of values present in other tables.
+RUS data is only available through 2021 because this is the last year a
+FOIA request was submitted. Future updates of this data will require
+additional FOIA requests. If you are interesting in supporting this
+please contact pudl@catalyst.coop.
```
cmgosnell, line 35: this was a thing we thought about adding after #5028 was merged in

cmgosnell:
i'm converting this t ready for review @e-belfer  even though i haven't done the migration (i don't want to have to do it over and over again so once this is reviewed i will translate into real tables

src/pudl/transform/rus12.py
```diff
@@ -324,3 +324,228 @@ def core_rus12__yearly_statement_of_operations(raw_rus12__statement_of_operation
...
+    This transform takes all of the plant production cost tables, processes
+    them similarly and combines them into one plant cost table.
+    """
+    plant_cost_tables = {
+        "combined_cycle": raw_rus12__combined_cycle_plant_costs,
+        "hydroelectric": raw_rus12__hydroelectric_plant_costs,
+        "internal_combustion": raw_rus12__internal_combustion_plant_costs,
+        "nuclear": raw_rus12__nuclear_plant_costs,
+        "steam": raw_rus12__steam_plant_costs,
+    }
```
cmgosnell, line 324: okay i only now am seeing that there are two other lists of plant types in rus12.

```python
               {
                    "combined_cycle",
                    "hydroelectric",
                    "internal_combustion",
                    "nuclear",
                    "steam",
                },
```

#### **core_rus12__yearly_sources_and_distribution_by_plant_type**
```python
plant_types = [
        "fossil_steam",
        "hydro",
        "internal_combustion",
        "combined_cycle",
        "nuclear",
        "other",
    ]
```

#### enumed in **core_rus12__yearly_plant_labor**
```python
PLANT_TYPE_RUS12: set[str] = [
    "Combined Cycle",
    "Steam",
    "Hydro",
    "Internal Combustion",
    "Nuclear",
]
```


I'd like to standardize them all to:

```python
plant_types = [
        "steam",
        "hydro",
        "internal_combustion",
        "combined_cycle",
        "nuclear",
        "other",
    ]
```

src/pudl/metadata/resources/rus12.py
cmgosnell: okay i broke the records here out into two separate tables. The BorrowerShared column wasn't reported before 2009 and there wasn't any old docs I could find. I added a big section in the resource metadata explaining this plus the pk weirdness

src/pudl/metadata/resources/rus12.py
```diff
@@ -1,5 +1,25 @@
...
+    "There are two boolean columns used to delineate which records are associated "
+    "with the borrowers' share vs the whole plant - which is documented in "
+    "``_OR_PowerSupply Plant File Documentation.rtf`` in the newer years in the "
+    "RUS 12 archive. One of these two fields - ``is_partly_owned_by_borrower`` - "
+    "was not reported before 2009. For the pre-2009 years, we assume that all records "
+    "that report TRUE for is_full_ownership_portion should end up in the by-plant table "
+    "while all records should end up in the by-borrower portion of the table."
+    "Like the post-2009 records, this involves records from the original tables ending "
+    "up in both of these PUDL tables."
+)
```
e-belfer, line 40:
```suggestion
    "The data in this table comes from five different portions of RUS 12 "
    "corresponding to different plant types (steam, hydroelectric, "
    "combined_cycle, internal_combustion and nuclear).\n"
    "The original form "
    "includes plant operations data for each plant type reported in multiple ways: with "
    "records corresponding to the portion of plants that borrowers own as well as the "
    "whole plant. We split these records into two tables, "
    ":ref:`core_rus12__yearly_plant_operations_by_borrower`, which contains records "
    "reported at the borrower level, and  "
    ":ref:`core_rus12__yearly_plant_operations_by_plant`, which contains records reported "
    "at the plant level."
    "Records that are wholly owned by one borrower show up in both "
    ":ref:`core_rus12__yearly_plant_operations_by_borrower` and "
    ":ref:`core_rus12__yearly_plant_operations_by_plant`.\n\n"
    "There are two boolean columns used to delineate which records are associated "
    "with the borrowers' share vs the whole plant - which is documented in "
    "``_OR_PowerSupply Plant File Documentation.rtf`` in the newer years in the "
    "RUS 12 archive. One of these two fields - ``is_partly_owned_by_borrower`` - "
    "was not reported before 2009. For the pre-2009 years, we assume that all records "
    "that report TRUE for is_full_ownership_portion should end up in the by-plant table "
    "while all records should end up in the by-borrower portion of the table."
    "Like the post-2009 records, this involves records from the original tables ending "
    "up in both of these PUDL tables."
)
```

I think this gives a bit more context about the transformation here.

src/pudl/metadata/resources/rus12.py
```diff
@@ -375,4 +395,138 @@
...
+            "additional_primary_key_text": (
+                "This table has no primary key because there is one plant (named Walter "
+                "Scott) that has duplicate records every year. The primary key of this table "
+                "otherwise would be: ['report_date', 'borrower_id_rus', 'plant_name_rus', 'cost_group', 'cost_type']."
+            ),
+            "additional_details_text": (
+                "The cost column in this table is expected to be largely non-null, the "
+                "cost_per_mwh and cost_per_mmbtu columns only apply to some cost_type's "
+                "and even plant_type's and thus are expected to contain many nulls "
+                "(64% and 90% respectively)"
```
e-belfer: If this number is pretty likely to change with a new year of data, we probably want to say this more generally to avoid it immediately being out of date.

src/pudl/metadata/resources/rus12.py
```diff
@@ -375,4 +395,138 @@
...
+                "borrower_id_rus",
+                "borrower_name_rus",
+                "plant_name_rus",
+                "unit_id_rus",
+                "plant_type",
+                "capacity_mw",
+                "gross_generation_mwh",
+                "ownership_pct",
+                "is_full_ownership_portion",
+                "is_partly_owned_by_borrower",
```
e-belfer, line 490: I could imagine us dropping or combining these columns now that we've used them to split up this table, as they're no longer that useful here. But this is possibly OOS for this current PR.

src/pudl/metadata/resources/rus12.py
```diff
@@ -375,4 +395,138 @@
...
+        "etl_group": "rus12",
+        "field_namespace": "rus",
+    },
+    "core_rus12__yearly_plant_operations_by_plant": {
+        "description": {
+            "additional_summary_text": (
+                "whole plant operational data including fuel consumption and operational hours."
+            ),
+            "usage_warnings": ["experimental_wip"],
+            "additional_source_text": "(Part D, E, F (CC), F (IC) & G - Section A)",
```
e-belfer, line 512: Is there a reason this differs from the _by_borrower table source text for Part F?

src/pudl/transform/rus12.py
```diff
@@ -154,6 +155,7 @@ def core_rus12__yearly_loans(raw_rus12__loans):
 def core_rus12__yearly_plant_labor(raw_rus12__plant_labor):
     """Transform the raw_rus12__plant_labor table."""
     df = rus.early_transform(raw_df=raw_rus12__plant_labor)
+    df = cleanstrings_snake(df, ["plant_type"])
 
     # Remove duplicate Walter Scott plant entries.
```
e-belfer: See my note above.

src/pudl/metadata/resources/rus12.py
```diff
@@ -375,4 +395,138 @@
...
+                "This data is for the total plant and does not contain any data for the "
+                "Borrower’s Share; this data can be  matched up with the data in the "
+                "following two files (where “YYYY” is the data year) that contain data for "
+                "Sections B and C for all plants; however you should use caution when using "
+                "total plant data since there are cases where more than one Borrower shares "
+                "units at the same plant which means that you will be getting duplicate plant "
+                "total records (and there is no guarantee that the total plant records entered "
+                "by two borrowers for the same plant will be identical):\n\n"
+                "OpRpt_PS_YYYY__US_dg_B_OpRpt_PSOperatingReportPlantLabor.csv\n"
+                "OpRpt_PS_YYYY__US_dg_C_OpRpt_PSOperatingReportPlantFactorsAndMaxDemand.csv"
```
e-belfer: I think including this is fine, but we'll want to rewrite it rather than copying reference to the CSV which is no longer visible. I think something like:

`Plant-level data can be matched to the :ref:`core_rus12__yearly_plant_labor` and forthcoming ``core_rus12__yearly_plant_factors_and_maximum_demand``, which report Section B and C data for all plants. The RUS instructions note that:

"you should use caution when using total plant data since there are cases where more than one Borrower shares units at the same plant which means that you will be getting duplicate plant total records (and there is no guarantee that the total plant records entered by two borrowers for the same plant will be identical)."


Notably though, I think this warning actually wants to live with the `yearly_plant_labor` table, as the warning here about duplicate data actually refers to those two tables (see our discussion of Walter Scott in #4979 - and we'll probably want to not drop the duplicated data by borrower in that table, which we are currently doing (see my comment below)

src/pudl/metadata/fields.py
```diff
@@ -9588,6 +9589,102 @@
...
+        "description": (
+            "Gallons of oil consumed for fuel."
+            "This field is only reported for plant_type's combined_cycle, combined_cycle and steam."
+        ),
+        "unit": "gal",
+    },
+    "fuel_consumption_other": {
+        "type": "number",
+        "description": (
+            "Other fuel consumed. Neither units nor type of fuel are documented."
```
e-belfer: Oof.

src/pudl/metadata/fields.py
```diff
@@ -9588,6 +9589,102 @@
...
+        ),
+    },
+    "is_partly_owned_by_borrower": {
+        "type": "boolean",
+        "description": (
+            "Whether or not the plant record is partially owned by the borrower. "
+            "This column was not reported before 2009."
+        ),
+    },
+    "fuel_consumption_coal_lbs": {
```
e-belfer, line 9630: Just to confirm, these are all annual? May be helpful to note (but nit and non-blocking).

docs/release_notes.rst
```diff
@@ -12,6 +12,11 @@ Enhancements
 New Data
 ^^^^^^^^
 
+RUS-12
+~~~~~~
+
+* Transformed and published USDA RUS tables. See :issue:`4886` and PR :pr:`5018`.
```
e-belfer: Do you want to say which, or at least "additional"?

src/pudl/metadata/fields.py
```diff
@@ -9588,6 +9589,102 @@
...
+        "constraints": {"enum": {"capex", "opex", "total"}},
+    },
+    "cost_type": {
+        "type": "string",
+        "description": "Detailed category of cost type.",
+        "constraints": {"enum": PLANT_COST_TYPES_RUS12},
+    },
+    "cost_per_mwh": {
+        "type": "number",
+        "description": "Cost per MWh",
```
e-belfer: Non-blocking nit: Is there a description here that isn't verbatim the name of the column? E.g., "Unit cost of energy production in cost per MWh."

src/pudl/helpers.py
```diff
@@ -1427,7 +1427,7 @@ def cleanstrings_snake(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
     for col in cols:
         df.loc[:, col] = (
             df[col]
-            .astype(str)
+            .astype(pd.StringDtype())
```
cmgosnell, line 1430: I updated this because i wanted to use it in `core_rus12__yearly_plant_labor` to snake-case the plant types and it was converting nulls to `"nan"` and that seemed wrong. I looked at all of the instances of where this is used and it seemed okay. i am running full etl and will run the validaiton checks just in case though

src/pudl/transform/rus12.py
```diff
@@ -252,7 +254,7 @@ def core_rus12__yearly_sources_and_distribution(
     df = rus.early_transform(raw_df=raw_rus12__sources_and_distribution)
     # Remove plant type columns handled in sources_and_distribution_by_plant_type function
     plant_types = [
```
e-belfer: Why not just import and reference the enum here? Seems like this leaves less room for error.

src/pudl/transform/rus12.py
```diff
@@ -196,7 +198,7 @@ def core_rus12__yearly_sources_and_distribution_by_plant_type(
     data_cols = ["capacity_kw", "cost", "plant_num", "net_energy_received_mwh"]
 
     plant_types = [
```
e-belfer: Why not just import and reference the enum here?

src/pudl/transform/rus12.py
```diff
@@ -324,3 +326,232 @@ def core_rus12__yearly_statement_of_operations(raw_rus12__statement_of_operation
...
+        df["plant_type"] = plant_type
+        df_outs[plant_type] = df
+    df_all = pd.concat(df_outs.values())
+    # this total flag is a little different than the others. it grabs the high-level
+    # cost group totals and it flags the cost_type's of net and less. Because these
+    # are clearly calculated fields from looking at the form.
+    df["is_total"] = df.cost_type.str.contains("total|net|less") | (
+        df.cost_group == "total"
+    )
+    # TODO: settle on names and do this rename in the column maps
```
e-belfer: Is this still an open todo?

src/pudl/transform/rus12.py
```diff
@@ -324,3 +326,232 @@ def core_rus12__yearly_statement_of_operations(raw_rus12__statement_of_operation
...
+            "per_mmbtu": "cost_per_mmbtu",
+        }
+    )
+    return df_all
+
+
+def _drop_bad_ownership_plant(df):
+    """Drop 1 plant record with unexpected ownership label and duplicate data.
+
+    There is a Wisdom steam plant record that is labeled to be both fully owned by
```
e-belfer, line 368: Do you want to add a note on this in the table or docs description somewhere?

src/pudl/transform/rus12.py
```diff
@@ -324,3 +326,232 @@ def core_rus12__yearly_statement_of_operations(raw_rus12__statement_of_operation
...
+            "is_full_ownership_portion": pd.BooleanDtype(),
+            "is_partly_owned_by_borrower": pd.BooleanDtype(),
+        }
+    ).pipe(_drop_bad_ownership_plant)
+
+    # for old years, there is no is_partly_owned_by_borrower column and no
+    # accompanying documentation. We assume if its a full ownership portion,
+    # it should go in both. if not full ownership portion  it should go in by borrower.
+
+    # Older years do not have a record for
```
e-belfer: Did this comment get cut-off?

src/pudl/transform/rus12.py
```diff
@@ -324,3 +326,232 @@ def core_rus12__yearly_statement_of_operations(raw_rus12__statement_of_operation
...
+    null_partly_owned_mask = df.report_date.dt.year.isin([2006, 2007, 2008])
+
+    # From _OR_PowerSupply Plant File Documentation.rtf in 2021 archive
+    # TO FOCUS ONLY ON DATA FOR THE BORROWERS’ SHARE OF THE PLANTS
+    # FullOwnershipScope    BorrowerShared
+    # FALSE                 TRUE
+    # TRUE                  FALSE
+    df_by_borrower = df[
```
e-belfer, line 506: Before this split, is it worth asserting that all records have these values filled out? If there are NAs in here other than the expected ones for earlier records we might be losing records without knowing.

We could also do this afterwards by checking that all rows got assigned somewhere, if that's simpler.

src/pudl/metadata/resources/rus12.py
```diff
@@ -375,4 +395,138 @@
...
+            "additional_details_text": (
+                "The cost column in this table is expected to be largely non-null, the "
+                "cost_per_mwh and cost_per_mmbtu columns only apply to some cost_type's "
+                "and even plant_type's and thus are expected to contain many nulls "
+                "(64% and 90% respectively)"
+            ),
+        },
+        "schema": {
+            "fields": [
+                "report_date",
```
e-belfer, line 447: You're dropping the `is_total` field here.

src/pudl/metadata/resources/rus12.py
cmgosnell: ah no i just expanded it but in only one place 🙃  good catch

cmgosnell:
the `unit_id_rus` as string was actually somewhat intentional bc I noticed some non-numeric values:

image

but I didn't realize at first there are only two of them and it looks like i can convert them to their numeric versions confidently:
image

and on 
> What is the distinction between cost types power and power_cost?

This was just an oopsies in the column mapping - it's fixed and now condensed into just `power` ⚡

src/pudl/transform/rus12.py
cmgosnell: ah yes. i re-wrote this overall explanation at least once... but the above comment is the one i'll remove this guy

src/pudl/transform/rus12.py
cmgosnell: great point! i think its simpler to do it after the fact so i added one in there.

src/pudl/metadata/resources/rus12.py
cmgosnell: agreed! i think combining them and keeping it makes sense to me because its hard to wrap my brain around the double columns. but i'd like to keep this oos for now.

src/pudl/metadata/resources/rus12.py
cmgosnell: okay i used this better edited text for this table and also added a modified version to the plant_labor table.

src/pudl/metadata/resources/rus12.py
```diff
@@ -1,5 +1,44 @@
...
+    "  Wisdom steam plant record that is labeled to be both fully owned by borrower "
+    "  and partly owned for one year. Which is an unexpected combo based on the "
+    "  `_OR_PowerSupply Plant File Documentation.rst` documentation file in the rus12 "
+    "  archive. Luckily this plant has exactly the same records as the other Wisdom steam"
+    "  plant that year with more expected ownership labels."
+    "* **Replaced two string values for unit_id_rus**: There are two instances of "
+    "  unit_id_rus's that have string values in them - ``WSL GT 12`` and ``WSL ST 10``. "
+    "  Based on pre-cleaned data, we were able to clearly identify that we can use just "
+    "  the numeric values in these bad strings - 12 and 10 respectively. This enables us "
+    "  to have an integer type for this unit_id_rus column."
```
cmgosnell, line 39: i added notes about the two bespoke cleaning steps here.

src/pudl/metadata/resources/rus12.py
```diff
@@ -34,7 +73,7 @@
             "additional_summary_text": (
                 "assets and other debts from the balance sheet."
             ),
-            "usage_warnings": ["experimental_wip"],
+            "usage_warnings": ["experimental_wip", "aggregation_hazard"],
```
cmgosnell, line 76: i also went through and added `aggregation_hazard` to all tables with an `is_total` column

src/pudl/metadata/resources/rus12.py
```diff
@@ -316,13 +355,21 @@
...
+                "Plant-level data from :ref:`core_rus12__yearly_plant_operations_by_plant` "
+                "can be matched to data from this table, but you must also join this data using"
+                "the ``borrower_id_rus`` column as well as the ``report_date`` and ``plant_name``. "
+                "This is because the :ref:`core_rus12__yearly_plant_operations_by_plant` table "
+                "includes records from each plant owner. "
+                "The RUS instructions note that:\n\n"
+                "you should use caution when using total plant data since there are cases where more than one "
+                "Borrower shares units at the same plant which means that you will be getting duplicate plant "
+                "total records (and there is no guarantee that the total plant records entered by two borrowers "
+                "for the same plant will be identical)."
```
cmgosnell, line 372: i removed the Walter Scott note because i removed this deletion in the transforms based on our new understanding. Plus added an altered version of the merge-with-caution note,

src/pudl/metadata/fields.py
```diff
@@ -9588,6 +9589,102 @@
         "description": "The year-to-date budget for amount of operational expense, cost or income.",
         "unit": "USD",
     },
+    "unit_id_rus": {
+        "type": "integer",
+        "description": "RUS-assigned unit identification code.",
+    },
```
cmgosnell, line 9595: switched this base to integer after removing the `WSL` bad strings (see `transform.rus12.fix_string_unit_id_rus`)

src/pudl/transform/rus12.py
```diff
@@ -154,18 +156,7 @@ def core_rus12__yearly_loans(raw_rus12__loans):
...
-    exclude_cols = ["borrower_id_rus", "borrower_name_rus"]
-    dupe_mask = (
-        df["borrower_id_rus"].isin(["IA0083", "IA0084"])
-        & (df["plant_name_rus"] == "Walter Scott")
-        & df.drop(columns=exclude_cols).duplicated(keep=False)
-        & (
-            df["borrower_id_rus"] == "IA0083"
-        )  # dropping the IA0083 and keeping the IA0084 so both borrowers show up in the data.
-    )
-    df = df.loc[~dupe_mask]
```
cmgosnell, line 168: deleted bc we learned from integrating the `core_rus12__yearly_plant_operations_by_plant` table  and the rst docs in the archive that this table should have one record per borrower-owner. see resources for detailed note

src/pudl/transform/rus12.py
```diff
@@ -324,3 +300,260 @@ def core_rus12__yearly_statement_of_operations(raw_rus12__statement_of_operation
...
+    # we want to check whether or not we are loosing any records:
+    both = pd.concat(
+        [df_by_borrower.set_index(idx_check), df_by_plant.set_index(idx_check)],
+        axis="index",
+        join="outer",
+        copy=False,
+    )
+    # There is some overlap between the by_borrower and by_plant records but we
+    # should have all of the original df records
+    assert df.set_index(idx_check).index.difference(both.index, sort=True).empty
```
cmgosnell, line 544: this is how im checking if we are not loosing any records with the above complicated masks

src/pudl/metadata/resources/rus12.py
```diff
@@ -17,7 +23,20 @@
     "that report TRUE for is_full_ownership_portion should end up in the by-plant table "
     "while all records should end up in the by-borrower portion of the table."
     "Like the post-2009 records, this involves records from the original tables ending "
-    "up in both of these PUDL tables."
+    "up in both of these PUDL tables.\n\n"
+    "Also there are two cleaning steps that we performed that alter the original data "
+    "slightly:\n\n"
+    "* **One dropped with unexpected ownership label and duplicate data**: There is a "
+    "  Wisdom steam plant record that is labeled to be both fully owned by borrower "
+    "  and partly owned for one year. Which is an unexpected combo based on the "
```
e-belfer: Non-blocking nit: For some reason suggest changes isn't working, but ", which is an unexpected" would make this more grammatically correct.

src/pudl/metadata/resources/rus12.py
e-belfer: Just for my own sanity understanding - are those the values this same entity also has in other years? Or are we batch correcting over all records over time?

src/pudl/metadata/resources/rus12.py
e-belfer: Non-blocking: You resolved this but didn't change anything - was that intentional?

e-belfer (approved):
Looks great! Two non-blocking questions/nits for you to consider or ignore as you wish.

src/pudl/metadata/resources/rus12.py
cmgosnell: good question! i will update the note to make that clear. but see the [comment above](https://github.com/catalyst-cooperative/pudl/pull/5018#issuecomment-3986724789)  - its just one year

src/pudl/metadata/resources/rus12.py
cmgosnell: i took out the #s altogether!

e-belfer: Oops, was looking at the wrong commit! Great.

Issue #5022, opened 2026-02-18
Switch `pandera` schema checks to use `polars`


zschira:
#### Table Schema
Currently we use `pandera` to validate the schema of our tables when assets are materialized. These schema checks are run in `asset_checks` immediately after any asset with a defined schema is materialized. This works well, but there's a growing list of tables which are too big to be loaded into `pandas` and checked by `pandera`. Given our current architecture, there are two likely options for addressing table schema validation that could work across the board:

1. Migrate schema checks to `dbt`
    - This would require modifying the `dbt_helper` script to automatically modify our `dbt` schema definitions to include datatypes, column level checks, and ideally configure `dbt` tests to run as `asset_checks` to maintain the immediate feedback loop.
    - This option has the benefit of making `dbt` schema definitions a single place to see everything we expect about a table (both schema and contents)
    - Leverages duckdb for maximum performance
2. Switch `pandera` to use the `polars` backend
    - This is would hopefully be a relatively easy change that could use a lot of existing infrastructure
    - While we have found `pandera` to be less memory efficient than `duckdb`, the `core_ferceqr__transactions` table is likely the only one that `polars` wouldn't be able to handle, and given that this is a partitioned asset, the asset check should run on individual partitions rather than the full table.
    - Would mean `dbt` does not become the definitive place to see table expectations

#### DB schema
Another type of schema validation that we do is foreign key checks, which happen at the database level. Currently we use `SQLite` for foreign key checks, which means there's an even larger set of tables for which we can't do these checks. `pandera` does not provide any way to implement checks across multiple tables, so `dbt` would likely be our best option for this.

### Success Criteria

We can implement the following types of checks on even the largest tables:

* [ ] validate datatypes
* [ ] validate column level expectations like categoricals, ranges, etc
* [ ] validate primary key uniqueness
* [ ] validate foreign key relationships

Issue #4853, opened 2025-12-17
Standardize bad FERC EQR categorical values


zaneselvans:
# Overview

There are a handful of categorical values in the FERC EQR data which aren't fixed simply by standardizing the case. They are easy to identify and map to the appropriate canonical values:

### `core_ferceqr__contracts.product_name`
```py
{
    "NEGOTIATED RATE TRANSMISSION": "NEGOTIATED-RATE TRANSMISSION",
}
```

### `core_ferceqr__transactions.product_name`
```py
{
    "NEGOTIATED RATE TRANSMISSION": "NEGOTIATED-RATE TRANSMISSION",
}
```

### `core_ferceqr__transactions.timezone`
```py
{
    "EST": "ES",
    "MST": "MS",
    "CDT": "CD",
    "MPT": "MP",
    "PST": "PS",
    "MDT": "MD",
    "EPT": "EP",
    "EDT": "ED",
    "CPT": "CP",
    "PDT": "PD",
    "PPT": "PP",
    "CST": "CS",
}
```

PR #5027, opened 2026-02-24
Add ability to apply schema to tables created with duckdb

zschira:
# Overview
This PR adds tooling to properly apply schemas to tables created with `duckdb`. Specifically, it adds the ability to properly handle columns with `enum` constraints

This enables work on #5022, as it enables us to produce tables with `duckdb` that exactly conform to their defined schema meaning that we can apply schema checks to these tables.

## What did you change?
- Updated `Resource.to_duckdb_dtypes` to handle `enum` types
- Updated `ferceqr` transforms to apply dtypes and make categorical columns conform to constraints

# Testing

I've tested these changes with the new `polarized` asset checks, which will be included in a follow up pr.

### DIFF
--- a/src/pudl/metadata/sources.py
+++ b/src/pudl/metadata/sources.py
@@ -747,9 +747,7 @@ SOURCES: dict[str, Any] = {
                 f"{year}q{quarter}"
                 for year in range(2013, 2026)
                 for quarter in range(1, 5)
-                if not (
-                    (year == 2013 and quarter < 3) or (year == 2025 and quarter > 3)
-                )
+                if not (year == 2013 and quarter < 3)
             ],
         },
         "license_raw": LICENSES["us-govt"],
--- a/src/pudl/transform/ferceqr.py
+++ b/src/pudl/transform/ferceqr.py
@@ -1,4 +1,26 @@
-"""Transform FERC EQR data."""
+"""Transform FERC Electric Quarterly Report (EQR) data.
+
+This module implements the transformation stage of the PUDL pipeline for FERC EQR data.
+Raw EQR data is ingested as quarterly-partitioned Apache Parquet files and transformed
+into clean, typed core tables that are written back out as Parquet.
+
+The module is structured in two layers:
+
+Reusable DuckDB transformation helpers operate on :class:`duckdb.DuckDBPyRelation`
+objects and are composed together inside each Dagster asset definition.
+
+Private expression factory functions that accept a column name (and optional parameters)
+and return a :class:`duckdb.Expression` suitable for use inside
+:func:`apply_column_transforms`.
+
+Dagster assets apply these helpers to produce four core FERC EQR tables, each of which
+is partitioned by ``year_quarter``:
+
+- :ref:`core_ferceqr__quarterly_identity`
+- :ref:`core_ferceqr__transactions`
+- :ref:`core_ferceqr__contracts`
+- :ref:`core_ferceqr__quarterly_index_pub`
+"""
 
 from collections.abc import Callable
 
@@ -22,12 +44,16 @@ def apply_duckdb_dtypes(
     table_name: str,
     conn: duckdb.DuckDBPyConnection,
 ):
-    """Cast columns to dtype as defined in schema.
+    """Cast each column to the dtype declared in the PUDL metadata schema for the table.
+
+    Column types are looked up from the :class:`pudl.metadata.classes.Resource` for
+    ``table_name``. Any custom enum types required by the schema are created in ``conn``
+    before the cast is applied.
 
     Args:
-        table_data: Duckdb table to transform.
-        table_name: Name of table.
-        conn: Connection to duckdb database, which is required to create custom enum types.
+        table_data: DuckDB relation whose columns will be cast.
+        table_name: PUDL table name used to look up the schema from the metadata.
+        conn: DuckDB connection used to register custom enum types as needed.
     """
     dtypes = Resource.from_id(table_name).to_duckdb_dtypes(conn)
 
@@ -44,11 +70,11 @@ def apply_duckdb_dtypes(
 def rename_duckdb_columns(
     table_data: duckdb.DuckDBPyRelation, mapping: dict[str, str]
 ) -> duckdb.DuckDBPyRelation:
-    """Rename columns of a duckdb table relation and return.
+    """Rename one or more columns in a DuckDB relation, passing all others through unchanged.
 
     Args:
-        table_data: Duckdb table to transform.
-        mapping: Maps column names to new names
+        table_data: DuckDB relation containing the columns to rename.
+        mapping: Maps existing column names to their new names.
     """
     return table_data.select(
         duckdb.StarExpression(exclude=mapping.keys()),
@@ -64,13 +90,17 @@ def apply_column_transforms(
     columns: list[str],
     transform: Callable[[str], duckdb.Expression],
 ) -> duckdb.DuckDBPyRelation:
-    """Apply a single transformation to a set of columns in a duckdb table.
+    """Apply a DuckDB expression factory to a set of columns, replacing each in place.
+
+    The ``transform`` callable is invoked once per column name and must return a
+    :class:`duckdb.Expression` whose result will be aliased back to the original column
+    name. All columns not listed in *columns* are passed through unchanged.
 
     Args:
-        table_data: Duckdb table to transform.
-        columns: List of columns to apply transform to.
-        transform: Callable which expects a column name and returns an Expression
-            defining the transform.
+        table_data: DuckDB relation containing the columns to transform.
+        columns: Names of the columns to which ``transform`` will be applied.
+        transform: Callable that accepts a column name and returns a DuckDB Expression
+            defining the transformation for that column.
     """
     return table_data.select(
         duckdb.StarExpression(exclude=columns),
@@ -78,7 +108,12 @@ def apply_column_transforms(
     )
 
 
-def _yn_to_bool(col: str):
+def _yn_to_bool(col: str) -> duckdb.Expression:
+    """Return a DuckDB expression that converts ``'Y'``/``'N'`` strings to booleans.
+
+    The comparison is case-insensitive. Any value other than ``'Y'`` or ``'N'`` is
+    mapped to ``NULL``.
+    """
     return duckdb.SQLExpression(f"""
 CASE
     WHEN UPPER({col}) = 'Y' THEN TRUE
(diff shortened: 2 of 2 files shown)


src/pudl/metadata/classes.py
```diff
@@ -638,12 +638,22 @@ def from_id(cls, x: str) -> "Field":
...
         return cls(**cls.dict_from_id(x))
 
-    def to_duckdb_dtype(self) -> duckdb.sqltypes.DuckDBPyType:
+    def to_duckdb_dtype(
+        self, conn: duckdb.DuckDBPyConnection
+    ) -> duckdb.sqltypes.DuckDBPyType:
         """Return duckdb data type."""
+        if self.constraints.enum:
+            quoted_enum = [f"'{val}'" for val in self.constraints.enum]
+            conn.execute(
```
zschira, line 647: To use `enum` types in `duckdb`, we actually have to first create the type in the current `duckdb` connection.

src/pudl/metadata/fields.py
```diff
@@ -10563,17 +10563,30 @@
                     "AP",
                     "AS",
                     "CD",
+                    "CDT",
```
zschira: Now that we're actually checking `enum` constraints on these tables, it exposed some unaccounted for categories.

pyproject.toml
```diff
@@ -115,7 +115,7 @@ numpy = ">=2.3"
 packaging = ">=25"
 pandas = ">=2.3"
 pandoc = ">=3.7"
-pandera-core = ">=0.27"
```
zschira, line 118: This change actually doesn't need to be included here as it we're not updating the `pandera` checks in this PR, but I accidentally included it and figured it would be fine to just get this in preemptively.

src/pudl/transform/ferceqr.py
```diff
@@ -15,38 +17,63 @@
 logger = get_logger(__name__)
 
 
-def transform_eqr_table(
-    table_name: str,
+def apply_duckdb_dtypes(
```
zschira, line 42: There's a bit of refactor here that made it easier to update these transforms so the resulting tables actually conform to schema (specifically due to actually checking `enum` types). The changes can be summarized as:

- Pull functionality out of old `transform_eqr_table` into two standalone functions:
  - `apply_duckdb_dtypes`
  - `rename_duckdb_columns`
- Create helper function `apply_column_transforms` that applies a single transformation to a set of columns

This refactor creates a more flexible and modular approach to transforming these tables that made it easier to add things like spot fixes to categorical columns.

src/pudl/metadata/fields.py
zaneselvans: These are invalid values according to the FERC data dictionary, so if we are going to enforce the ENUM constraint, we should standardize them rather than adding them to the allowed values (as you've already done for `product_name` below). The standard values are listed in #4853

src/pudl/metadata/classes.py
```diff
@@ -638,12 +638,22 @@ def from_id(cls, x: str) -> "Field":
...
+                f"CREATE TYPE {self.name}_type as ENUM ({','.join(quoted_enum)})"
+            )
+            return conn.dtype(f"{self.name}_type")
```
zaneselvans, line 650: In the case where the same field appears in multiple tables with the same ENUM constraints, will this try and create the same type multiple times with the same name? Does that cause issues?

pyproject.toml
zaneselvans: Will we really no longer depend on `pandera-core` at all? Like it'll only be needed as a dependency of `pandera-polars`?  Or should we now have both of them?

src/pudl/transform/ferceqr.py
```diff
@@ -15,38 +17,63 @@
...
+    # Cast columns and implicitly drop any columns not defined in the schema
+    return table_data.select(
+        *[
+            duckdb.ColumnExpression(col).cast(dtype).alias(col)
+            for col, dtype in dtypes.items()
+        ]
+    )
+
+
+def rename_duckdb_columns(
```
zaneselvans, line 70: If this function is being used to rename columns, then should we still have the `.alias(col)` in the dtype enforcement function above? Or should those two operations be entirely separated?  Oh, or is what's happening that a column in the selection is being defined as a function of some other column which yeah might have the same name, but it's getting modified.

src/pudl/transform/ferceqr.py
```diff
@@ -66,42 +93,42 @@ def _na_to_null(col_name: str) -> duckdb.Expression:
...
-                ),
+        table_data = apply_column_transforms(
+            table,
+            ["transactions_reported_to_index_price_publishers"],
+            _yn_to_bool,
+        )
+        table_data = rename_duckdb_columns(
+            table_data,
+            {
+                "company_identifier": "company_id_ferc",
```
zaneselvans, line 187: I know it's not syntactically required, but I think adding the keyword argument names in here would make this more readable.

src/pudl/transform/ferceqr.py
```diff
@@ -66,42 +93,42 @@ def _na_to_null(col_name: str) -> duckdb.Expression:
...
             },
         )
 
+        return persist_table_as_parquet(
+            table_name=table_name,
+            table_data=apply_duckdb_dtypes(table_data, table_name, conn),
```
zaneselvans: Sneaking in a last transform in the return statement rather than alongside the other transforms above makes it less clear what all transformations are being applied to the table.

src/pudl/transform/ferceqr.py
```diff
@@ -110,174 +137,165 @@ def core_ferceqr__transactions(context, raw_ferceqr__transactions: ParquetData):
...
-                ),
-                "standardized_price": duckdb.ColumnExpression("standardized_price"),
-                "total_transmission_charge": duckdb.ColumnExpression(
-                    "total_transmission_charge"
-                ),
-                "total_transaction_charge": duckdb.ColumnExpression(
-                    "total_transaction_charge"
-                ),
+    ) as (table, conn):
+        table_data = apply_column_transforms(
```
zaneselvans, line 211: Would it be easy to make this into a chain of operations rather than repeated clobbering of the `table_data` variable? Purely aesthetic I guess.

src/pudl/transform/ferceqr.py
```diff
@@ -110,174 +137,165 @@ def core_ferceqr__transactions(context, raw_ferceqr__transactions: ParquetData):
...
+                "product_name",
+                "rate_units",
+            ],
+            _na_to_null,
+        )
+        # Normalize categorical string
+        table_data = apply_column_transforms(
+            table_data,
+            ["product_name"],
+            lambda _: duckdb.SQLExpression(
```
zaneselvans: Maybe another generic transform function to handle this case and the timezone remapping?

src/pudl/metadata/classes.py
zschira: This will create the type multiple times, but that doesn't cause a problem and is actually necessary because the duckdb connection used here is in-memory, so it only exists for the duration of a single asset run.

src/pudl/transform/ferceqr.py
zschira: This is basically equivalent to doing something like `SELECT CAST(i AS VARCHAR) AS i` because by default duckdb will create a new column called `CAST(i AS VARCHAR)` or something. So, this is just maintaining the column names that the table starts with at the beginning of the function rather than changing them.

src/pudl/transform/ferceqr.py
zschira: Fixed

src/pudl/transform/ferceqr.py
zschira: Fixed

src/pudl/transform/ferceqr.py
zschira: I also wanted to try to chain these, but there's no equivalent to pandas `pipe` that would allow this in the `duckdb` API. I could see creating a wrapper class to enable piping if we end up doing more duckdb stuff, but felt like overkill for just these transforms.

src/pudl/transform/ferceqr.py
zschira: Good call, I added a new `_replace_strings` function that is used here and for the timezone fixes.

pyproject.toml
zschira: We still use `pandera` with `pandas` in a couple places, but `pandera-core` is implicitly included with `pandera-polars`. We could definitely make it explicit though if preferred.

src/pudl/metadata/fields.py
zschira: Fixed now.

zschira:
> I like the improved modularity, but one side-effect is that it's once again unclear in the context of the transform functions what all columns we expect the returned tables to have in them. Are there (still?) programmatic guard rails in place that ensure that all columns specified in the metadata are present in the outputs? And that there are never any extra columns? And that the specified columns are non-null?

This will all be offloaded to the `pandera` schema checks in my followup PR, which I've tested with these tables and works well. I just changed the `apply_duckdb_dtypes` function to no longer implicitly drop columns, so we're forced to explicitly drop columns or the schema checks will fail.

src/pudl/metadata/fields.py
```diff
@@ -10569,6 +10569,7 @@
                     "AS",
                     "CD",
                     "CP",
+                    "CPD",
```
zaneselvans: `CPD` is invalid according to the data dictionary, but I'm not sure if it should get mapped to `CP` (Central Prevailing) or `CD` (Central Daylight).  Is there any clue from the data itself? If not we should probably set it to NULL.

src/pudl/transform/ferceqr.py
```diff
@@ -66,41 +95,61 @@ def _na_to_null(col_name: str) -> duckdb.Expression:
...
 
+def _parse_datetimes(col_name: str, fmt: str) -> duckdb.Expression:
+    """Return a duckdb expression to parse datetimes from strings."""
+    return duckdb.SQLExpression(f"TRY_STRPTIME({col_name}, '{fmt}')")
+
+
+def _replace_strings(
+    col_name: str, replace_mapping: dict[str, str]
+) -> duckdb.Expression:
+    """Return a duckdb expression to replace a substrings in a column with specified values."""
```
zaneselvans: This is more complex than I expected. When I was thinking about these fixes initially I had imagined this being a CASE statement, with each case indicating a mapping from an existing incorrect value to the corresponding corrected value. Can you add some more documentation / an example here? I'm having trouble understanding what this is actually doing.

By replacing "substrings" do you mean that it will replace only the portion of the input string that matches with the target? I can see how that would be useful in other contexts, but it also seems more prone to unexpected behavior, when the current application is just "If exactly this bad value, set to this good value."

What role is `base` playing? How does `reduce()` work here? Is it creating a nested chain of substring replacements?  Would that mean that with particular sets of input replacement mappings, you could get order-dependent outputs because the output of one replacement matches the pattern targeted by a subsequent replacement?

Would `dict.popitem()` permanently change `replace_mapping` meaning that if you tried to use the same mapping on multiple columns it might not be the same each time?

src/pudl/metadata/fields.py
zaneselvans: Querying the EQR outputs that I have locally with
```sql
SELECT * FROM 'core_ferceqr__transactions/*.parquet'
WHERE timezone = 'CPD';
```
I get zero results.  Are you seeing this value in your outputs?  If not maybe it was a typo / accident when I compiled these values.

src/pudl/transform/ferceqr.py
zaneselvans: An alternative might look like

```python
def _recode_categorical(
    col_name: str, replace_mapping: dict[str, str]
) -> duckdb.Expression:
    when_clauses = "\n    ".join(
        f"WHEN {col_name} = '{bad}' THEN '{good}'"
        for bad, good in replace_mapping.items()
    )
    return duckdb.SQLExpression(f"""
CASE
    {when_clauses}
    ELSE {col_name}
END""")
```

zschira: Switched to a `CASE` statement as recommended. Definitely simpler this way.

src/pudl/metadata/fields.py
zschira: I'm not seeing this value locally, or in nightly builds, so I think this is probably a typo. Removed now.

src/pudl/metadata/sources.py
```diff
@@ -747,9 +747,7 @@
                 f"{year}q{quarter}"
                 for year in range(2013, 2026)
                 for quarter in range(1, 5)
-                if not (
-                    (year == 2013 and quarter < 3) or (year == 2025 and quarter > 3)
-                )
+                if not (year == 2013 and quarter < 3)
```
zaneselvans, line 750: I noticed that the working partitions had not been updated to include last quarter, so I added it.

src/pudl/transform/ferceqr.py
```diff
@@ -1,4 +1,28 @@
-"""Transform FERC EQR data."""
+"""Transform FERC Electric Quarterly Report (EQR) data.
```
zaneselvans, line 1: Made the docstrings all less terse.

zaneselvans (approved):

PR #5029, opened 2026-02-25
Fix metadata on noncompliant tables

krivard:
# Overview

We want to highlight the following tables in our topic index, but their description summaries were borked because their names don't comply with our naming conventions. 

* out_ferc714__georeferenced_respondents
* out_eia923__fuel_receipts_costs
* out_censusdp1tract__states
* out_censusdp1tract__counties
* out_censusdp1tract__tracts

## What did you change?

* Added overrides to type codes and/or summary text for affected tables, so that summaries render with a complete sentence.
* Added bonus warnings to censusdp1tract tables, because the dp1 data is no longer reported in the easy format and thus our data is old
* **Renamed `core_eia923__monthly_fuel_receipts_costs` to `core_eia923__fuel_receipts_costs`** as it is not aggregated monthly and does not belong with our timeseries tables
  * Updated associated details text to describe aggregation gotchas
* Updated metadata description compliance test to scream if any table renders its summary as a fragment instead of a complete sentence.

## Documentation

# Testing

* Verified using `resource_description --name [name]`
* Ran `pudl_check_fks`
* Running pytest integrations in GHA

## To-do list

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,8 @@ The Public Utility Data Liberation Project (PUDL)
 .. |rtd| image:: https://img.shields.io/readthedocs/catalystcoop-pudl?style=flat&logo=readthedocs
    :target: https://catalystcoop-pudl.readthedocs.io/en/nightly/
    :alt: Read the Docs Build Status
+.. |oc| image:: https://opencollective.com/pudl/tiers/badge.svg
+   :target: https://opencollective.com/pudl
 .. |ruff| image:: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json
    :target: https://github.com/astral-sh/ruff
 .. |pre-commit-ci| image:: https://results.pre-commit.ci/badge/github/catalyst-cooperative/pudl/main.svg
@@ -44,15 +46,12 @@ The Public Utility Data Liberation Project (PUDL)
 .. |youtube| image:: https://img.shields.io/badge/YouTube-%23FF0000.svg?logo=YouTube&logoColor=white
    :target: https://youtube.com/@CatalystCooperative
    :alt: Catalyst Cooperative on YouTube
-.. |twitter| image:: https://img.shields.io/badge/X-%23000000.svg?logo=X&logoColor=white
-   :target: https://x.com/CatalystCoop
-   :alt: Catalyst Cooperative on Twitter
 .. |aws| image:: https://img.shields.io/badge/Amazon_AWS-FF9900?style=flat&logo=amazonaws&logoColor=white
    :target: https://registry.opendata.aws/catalyst-cooperative-pudl/
    :alt: PUDL in the AWS Open Data Registry
 
-|repo-status| |pytest| |codecov| |rtd| |ruff| |pre-commit-ci| |zenodo-doi|
-|office-hours| |mastodon| |linkedin| |bluesky| |kaggle| |slack| |youtube| |twitter|
+|repo-status| |pytest| |codecov| |rtd| |oc| |ruff| |pre-commit-ci|
+|zenodo-doi| |office-hours| |mastodon| |linkedin| |bluesky| |kaggle| |slack| |youtube|
 |aws|
 
 What is PUDL?
@@ -327,7 +326,6 @@ Contact Us
 * Follow us on BlueSky:  `@catalyst.coop <https://bsky.app/profile/catalyst.coop>`__
 * `Follow us on LinkedIn <https://www.linkedin.com/company/catalyst-cooperative/>`__
 * `Follow us on HuggingFace <https://huggingface.co/catalystcooperative>`__
-* Follow us on Twitter: `@CatalystCoop <https://twitter.com/CatalystCoop>`__
 * `Follow us on Kaggle <https://www.kaggle.com/catalystcooperative/>`__
 * `Follow us on YouTube <https://youtube.com/@CatalystCooperative>`__
 * `Join our Slack <https://join.slack.com/t/catalystcooperative/shared_invite/zt-2yg1v2sb7-GsoGlA9Ojc_LCJ00vPWKbQ>`__


src/pudl/metadata/resources/censusdp1tract.py
```diff
@@ -191,14 +191,19 @@
     "dp0230002",
 ]
 
+CENSUS_CAVEAT = (
+    "Based on the 2010 Census. Used "
+    "as a convenient source of geospatial information within PUDL, but should "
+    "not be treated as authoritative."
+)
```
krivard, line 198: all three tables had the same text here so I factored it out

src/pudl/metadata/resources/censusdp1tract.py
```diff
@@ -191,14 +191,19 @@
...
-        "description": (
-            "State-level summarized US Census demographic information, along with "
-            "a geometry describing state boundaries. Based on the 2010 Census. Used "
-            "as a convenient source of geospatial information within PUDL, but should "
-            "not be treated as authoritative."
-        ),
+        "description": {
+            "additional_summary_text": "State-level summarized US Census demographic information, along with a geometry describing state boundaries.",
+            "additional_details_text": CENSUS_CAVEAT,
+            "usage_warnings": ["discontinued_data"],
```
krivard, line 205: the alternative warning is "discontinued_pudl" which would emphasize that the data is still being collected, we're just not integrating it. I picked "discontinued_data" because the issue is the format of how the data are currently being reported.

src/pudl/metadata/resources/eia923.py
```diff
@@ -428,6 +428,8 @@
                     "plants and utilities"
                 ),
                 "usage_warnings": ["harvested"],
+                "table_type_code": "timeseries",
+                "timeseries_resolution_code": "monthly",
```
krivard: (we really should discontinue all these weird-named duplicates of monthly timeseries tables. but since we have them, and since people use them, they might as well document themselves properly)

zaneselvans: What's our criteria for calling something a timeseries?  The records in this table have "monthly" resolution, in that they are all associated with a particular month (by `report_date`) but it's not possible to create a neat data series with the month as part of the primary key without doing a bunch of aggregation and losing information. Each record is more like an individual transaction between a fuel supplier and a plant. But there are cases where the same fuel supplier delivers the same fuel to a given plant multiple times in a month. And then there are many cases in which there isn't a continuous sequence of deliveries to a plant each month from a given supplier, or maybe even from any supplier.  So is it really a time series?  It seems qualitatively different from e.g. the hourly demand or generation time series, or even the monthly generation & fuel consumption totals for a given generator / boiler.

Which is why this has remained a weirdo. The `core_ferceqr__transactions` table is kind of similar.

krivard: A timeseries is anything you can plot using time as the x axis.

I'm not convinced this one is really so weird, given that we publish `out_eia923__monthly_fuel_receipts_costs` and `out_eia923__monthly_fuel_receipts_costs` and neither of them have a primary key either (& so presumably have the same weird/not-weird status and justification)

zaneselvans: Looking at the data directly, `out_eia923__monthly_fuel_receipts_costs` table is an aggregated version of `out_eia923__fuel_receipts_costs`. It has a unique PK which is:
```py
[
    "report_date",
    "plant_id_eia",
    "fuel_type_code_pudl",
]
```
and only contains the subset of columns which can be meaningfully aggregated -- mostly numeric, plus a few categoricals, and attributes which are uniquely determined within each PK like the Plant Name. I assume the `yearly` version is the same, but aggregated to the year.

So we should update the description to differentiate between the aggregated and original versions of this data, and I think add PKs to both the monthly & annually aggregated versions of the FRC tables. The description we currently have is explicitly associated with the `monthly` version of the table, but it really describes the original version, where each record is an individual delivery.

I think our current table naming conventions can't really distinguish between the original and monthly aggregated versions of this table, since they both contain the same type of information, and nominally have "monthly" reporting frequency, which is why this weirdo is still here.

zaneselvans: Looking at how the metadata is defined and the tables are named right now, I think it's confusing.

* `core_eia923__monthly_fuel_receipts_costs` is the ultimate source for all of the FRC tables, and the output table which is most directly analogous is `out_eia923__fuel_receipts_costs` which to me suggests that they should have the same table name, with the exception of the `core_` vs. `out_` prefixes.
* Then we have the monthly/yearly aggregated output versions of the table, which need to have a different description that indicates they are aggregated / simplified versions of that source data, and by virtue of that aggregation, do have primary keys (this aggregated data would also be good for plotting fuel deliveries by fuel type over time to a plant in the plant summary dashboard)
* However, currently all 4 of these tables share the same description, the one which was originally written for the core table.

I'd propose the following 4 tables:
* `core_eia923__fuel_receipts_costs` -- which uses the current description.
* `out_eia923__fuel_receipts_costs` -- which shares the above description, but with the additional notes that come from being an output table.
* `out_eia923__monthly_fuel_receipts_costs` and `out_eia923__yearly_fuel_receipts_costs` which have defined primary keys, and refer back to the `core_eia923__fuel_receipts_costs` table description (but don't include it) for most of the context of what data is in the table, and explain that they are time aggregated, and only contain the columns which can be meaningfully aggregated (mostly numerical with a few categoricals) plus plant-level attributes that are homogeneous within each row.

I'm happy to make these changes on this branch.

krivard: If I understand correctly, that means the following changes:

* `core_eia923__monthly_fuel_receipts_costs` -> `core_eia923__fuel_receipts_costs` name change
  * add `table_type_code:timeseries`
* `out_eia923__fuel_receipts_costs` - add notes to `additional_details_text` if needed (I'm okay with just the default layer text but if there's particular things added that need more explanation, Additional Details is where that would go)
* `out_eia923__*ly_fuel_receipts_costs` - 
  * define primary keys, and drop `additional_primary_key_text`
  * update `additional_details_text` to explain aggregation and add cross-reference to unaggregated table

That sounds great! Go for it. Should I plan an extra announcement of the core table name change?

I'm going to add a new `timeseries_resolution_code` for "irregular", to be used for these kinds of chronological logs that don't follow any kind of equal spacing.

zaneselvans: That all looks right.

I still feel weird about calling these tables where there's no way to select a unique data point at a given timestep timeseries. Like there's no dependent variable which is (in the mathematical sense) a function of time, because it's not guaranteed to be single-valued.  Whereas for the aggregated tables, all of the variables can be selected on the basis of time in combination with plant ID and fuel type -- so it's a collection of many individual time series.

krivard: Alternately and since it seems like there's less of a pattern here than I thought, we can handle these as one-offs -- any table that doesn't have a valid asset type but does have `additional_summary_text` that is a complete sentence starting with a capital letter will display correctly.

I've added an explicit check for this and fixed the other tables that didn't meet it.

I think we should still do the remaining frc changes, but they become:

* core_eia923__monthly_fuel_receipts_costs -> core_eia923__fuel_receipts_costs name change
  * make sure additional_summary_text is a complete sentence starting with a capital letter
* out_eia923__fuel_receipts_costs - add notes to additional_details_text if needed (I'm okay with just the default layer text explaining denormalization but if there's particular things added that need more explanation, Additional Details is where that would go)
* out_eia923__*ly_fuel_receipts_costs -
  * define primary keys, and drop additional_primary_key_text
  * update additional_details_text to explain aggregation and add cross-reference to unaggregated table

zaneselvans: I will work on implementing this today.

zaneselvans: Okay I think this is implemented and I've pushed.

zaneselvans:
@krivard I've renamed the FRC tables as we discussed and updated all the instances I could find of `core_eia923__monthly_fuel_receipts_costs` to `core_eia923__fuel_receipts_costs`.  I created a new alembic migration and:

* ✅ Re-materialized everything up and down stream of the FRC tables and it seems to have worked fine.
* ✅ Ran `pixi run dbt build` and got no novel errors.
* ✅ Ran `pixi run pudl_check_fks` to make sure the new PK / FK constraints are being respected.

src/pudl/metadata/resources/eia860.py
```diff
@@ -247,7 +247,7 @@
                     "core_eia923__monthly_boiler_fuel",
                     "out_eia923__generation",
                     "out_eia923__monthly_generation",
-                    "core_eia923__monthly_fuel_receipts_costs",
+                    "out_eia923__monthly_fuel_receipts_costs",
```
krivard, line 252: this was a good catch ⭐

src/pudl/metadata/resources/eia860.py
```diff
@@ -396,15 +396,15 @@
                     "core_eia923__monthly_boiler_fuel",
                     "out_eia923__boiler_fuel",
                     "out_eia923__monthly_boiler_fuel",
-                    "out_eia923__fuel_receipts_costs",
                     "out_eia923__monthly_fuel_receipts_costs",
+                    "core_eia923__fuel_receipts_costs",
+                    "out_eia923__fuel_receipts_costs",
```
krivard, line 402: actually do we need either of these if they're not aggregated? or is about the granularity of the date column?

src/pudl/metadata/resources/eia923.py
```diff
@@ -58,8 +58,8 @@
 complex."""
         ),
     },
-    "core_eia923__monthly_fuel_receipts_costs": {
-        "additional_summary_text": "fuel deliveries to power plants.",
+    "core_eia923__fuel_receipts_costs": {
+        "additional_summary_text": "Individual fuel deliveries to power plants, organized by fuel type and supplier.",
```
krivard, line 62: yes good

src/pudl/metadata/resources/eia860.py
zaneselvans: It's just about the granularity of the date column found in association with the `plant_id_eia` columns -- unless we exclude it, it'll automatically try to generate a FK constraint that uses both columns, and will fail, because all but January months will not be found in the SCD table.

I'd love to set up year-only FK checks for these tables in `dbt`.

zaneselvans: Ahhh, I see I got confused in here thinking all of these changes were part of the same resource, but see now that they're split across several.

zaneselvans:
For some reason the CI in the merge queue was going verrrrrrry slow and it got pulled from the queue after like 2 hours. It was still running 3h15m in, when usually it only takes 50 minutes. So I cancelled and resubmitted. Not sure what's going on.

